PR #22584, opened 2025-12-02
chore(core): Credential resolver interface and decorator

afitzek:
## Summary

Implements ICredentialResolver interface and CredentialResolverMetadata for the dynamic credential resolution system (PAY-4207, PAY-4208). Adds @CredentialResolver() decorator for auto-discovery, registry service for resolver type management, and custom error classes for resolver operations.

closes https://linear.app/n8n/issue/PAY-4207/define-resolver-interface
closes https://linear.app/n8n/issue/PAY-4208/define-resolver-metadata-interface

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/@n8n/decorators/src/credential-resolver/credential-resolver-metadata.ts
+++ b/packages/@n8n/decorators/src/credential-resolver/credential-resolver-metadata.ts
@@ -1,0 +1,52 @@
+import { Container, Service } from '@n8n/di';
+
+import { CredentialResolverClass } from './credential-resolver';
+
+type CredentialResolverEntry = {
+	class: CredentialResolverClass;
+};
+
+/**
+ * Registry service for credential resolver type discovery and instantiation.
+ * Resolver classes decorated with @CredentialResolver() are automatically registered.
+ */
+@Service()
+export class CredentialResolverEntryMetadata {
+	private readonly credentialResolverEntries: Set<CredentialResolverEntry> = new Set();
+
+	/** Registers a credential resolver class. Called automatically by @CredentialResolver() decorator. */
+	register(credentialResolverEntry: CredentialResolverEntry) {
+		this.credentialResolverEntries.add(credentialResolverEntry);
+	}
+
+	/** Returns all registered resolver entries as [index, entry] tuples. */
+	getEntries() {
+		return [...this.credentialResolverEntries.entries()];
+	}
+
+	/** Returns all registered resolver classes. */
+	getClasses() {
+		return [...this.credentialResolverEntries.values()].map((entry) => entry.class);
+	}
+}
+
+/**
+ * Decorator to mark a class as a credential resolver.
+ * Automatically registers the resolver for discovery and enables dependency injection.
+ *
+ * @example
+ * @CredentialResolver()
+ * class MyResolver implements ICredentialResolver { ... }
+ */
+export const CredentialResolver =
+	<T extends CredentialResolverClass>() =>
+	(target: T) => {
+		// Register resolver class for discovery by registry
+		Container.get(CredentialResolverEntryMetadata).register({
+			class: target,
+		});
+
+		// Enable dependency injection for the resolver class
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-return
+		return Service()(target);
+	};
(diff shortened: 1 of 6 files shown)


packages/@n8n/decorators/src/credential-resolver/credential-resolver.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Configuration object passed to resolver methods. Structure is defined by resolver type's metadata.options.
+ */
+export type CredentialResolverConfiguration = Record<string, unknown>;
+
+/**
+ * Metadata describing a credential resolver type for UI integration and discovery.
+ */
+export interface CredentialResolverMetadata {
+	/** Unique identifier for the resolver type */
+	name: string;
```
BGZStephen, line 18: Should this be an enum type that can be validated by zod?

packages/@n8n/decorators/src/credential-resolver/credential-resolver-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+	private readonly credentialResolverEntries: Set<CredentialResolverEntry> = new Set();
+
+	/** Registers a credential resolver class. Called automatically by @CredentialResolver() decorator. */
+	register(credentialResolverEntry: CredentialResolverEntry) {
+		this.credentialResolverEntries.add(credentialResolverEntry);
+	}
+
+	/** Returns all registered resolver entries as [index, entry] tuples. */
+	getEntries() {
+		return [...this.credentialResolverEntries.entries()];
```
BGZStephen, line 24: Is there a reason we don't just use a set as the data structure for `credentialResolverEntries`? To avoid producing multiple copies of the credentialResolverEntries with a spread operator?

For `getClasses` you could also make use of a set that is added to as part of the `register` command:

```
	register(credentialResolverEntry: CredentialResolverEntry) {
		this.credentialResolverEntries.add(credentialResolverEntry);
        this.credentialResolverClasses.add(credentialResolverEntry.class)
	}

   	getEntries() {
		return this.credentialResolverEntries;
    }

    getClasses() {
		return this.credentialResolverClasses;
	}
```

packages/@n8n/decorators/src/credential-resolver/credential-resolver.ts
afitzek: An enum type would mean we need a central list of available resolvers, but this should not be the case, since in a far future I imagine, that we can also register resolvers using some kind of plugin mechanism.

packages/@n8n/decorators/src/credential-resolver/credential-resolver-metadata.ts
afitzek: We could also do that, I basically followed the patterns for context establishment hooks and the module registry. 

https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/decorators/src/context-establishment/context-establishment-hook-metadata.ts
https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/decorators/src/module/module-metadata.ts

afitzek: I think the original reason for this was, to not expose the set to avoid potential mutations of the internal set structure. So all users of the registry need to be aware, that they are not allowed to mutate the handed out sets.

BGZStephen: Yeah, I thought it'd be something like that, there's a real lack of immutability in javascript :(

packages/@n8n/decorators/src/credential-resolver/credential-resolver.ts
BGZStephen: I see what you're saying, playing devil's advocate, wouldn't it be nicer to start restrictive, and relax it *if* / *when* the need arises?

afitzek: Actually this will never be parsed, but an implementation will have this value hard coded. For example a resolver can be defined as:

```
class Resolver implements IResolver {
  metadata: {
    name: "resolver.oauth.1.0",
    ...
  }
}
```

So I'm not sure if we would benefit from restricting this type using an enum.

BGZStephen (approved):

packages/@n8n/decorators/src/credential-resolver/credential-resolver.ts
BGZStephen: That makes sense if each resolver is instantiated independently, I guess there is a question over uniqueness, but I'm sure that'll figure its self out :D

PR #22433, opened 2025-11-28
docs: Add component specification and usage examples for N8nPopover

cstuncsik:
## Summary

Add API specification for N8nPopover migration from Element+ to Reka UI

## Related Linear tickets, Github issues, and Community forum posts

PAY-4193

## Review / Merge checklist

- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #22591, opened 2025-12-02
chore(editor): Add Reka UI based popover component

cstuncsik:
## Summary

Refactor/rebuild N8nPopover component with Reka UI

PAY-4194

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/design-system/src/v2/components/Popover/Popover.vue
```diff
@@ -0,0 +1,233 @@
...
+
+defineOptions({
+	inheritAttrs: false,
+});
+
+const props = withDefaults(defineProps<N8nPopoverProps>(), {
+	trigger: 'click',
+	placement: 'bottom',
+	teleported: true,
+	showArrow: true,
```
konstantintieber, line 15: Here we have the `showArrow` prop that's specific to elementPlus but not needed by like in the tooltip component the AI agent generated.

konstantintieber (commented):
I did a quick code search for `<N8nPopover` and found that there are only three popovers left in our codebase that don't use the new `N8nPopoverReka`:
- OperatorSelect
- ResourceLocatorDropdown
- CanvasNodeStickyColorSelector

In the component spec PR [you proposed](https://github.com/n8n-io/n8n/pull/22433#issuecomment-3589666441) to migrate both the existing `N8nPopover` and `N8nPopoverReka` to the new AI generated one.

I would expect it's less effort to migrate those three to `N8nPopoverReka`, than to migrate both of them (meaning 14 popover instances in total) to the AI generated component. With the ones that are using `N8nPopoverReka` today, we at least don't have to do manual testing again.
And we get the benefits of a human-authored implementation.

What do you think?

packages/frontend/@n8n/design-system/src/v2/components/Popover/Popover.vue
cstuncsik: In `OperatorSelect`, `showArrow` is actually set to false.

I'm also fine with a migration where we also change the consumer side but initially I just wanted a fully compatible drop-in replacement

Basically not touching the editor code but only replacing the components in the design system package

konstantintieber: While a drop-in replacement sounds nice, I'd still want to check each usage of the component manually when migrating.

Which in this case would mean testing 14 usages manually when going for the AI-generated one, vs just testing three manually when going with the existing `N8nPopoverReka`.

I'm not challenging the "drop-in migration" approach in general, just seeing an opportunity to migrate the popover with less manual effort and high confidence in no regressions.

cstuncsik: Makes sense but I also need to modify a bit the `N8nPopoverReka`, at least to have the `showArrow` prop.

I'll make another PR in which I alos do the migration and let's see.

konstantintieber (approved):
Component looks fine.

Whichever engineer ends up doing the migration, can decide whether they want to go the route of migrating all usages to the v2 reka Popover, having to test all existing usages manually, or migrating the remaining element+ popovers to the existing `N8nPopoverReka`, meaning they only need to test the former element+ popovers for regressions. In the latter case, this v2 popover component would be deleted again.

r00gm (approved):

PR #18928, opened 2025-08-28
refactor(core): Consolidate task result validation in Code node

ivov:
## Summary

For reasons lost to time, we have [two copies](https://github.com/n8n-io/n8n/pull/11109/files#diff-2b6a5a9ccce937cb51a63c7c2ff1c04c38f561a00b39177db017b7994b9830c2) of result validation, one n8n-side for `vm2` and `pyodide`, and another runner-side. To prevent adding a third copy for the Python runner, this PR consolidates task result validation n8n-side for legacy libs and for all runner types. This means: 

- Runners always return raw output, less coupled to n8n items
- Occasional waste from sending back invalid output items
- No need to reimplement validation for every runner type

Keep in mind now a task may succeed, and later on still end in failure n8n-side. From my testing, this does not affect the end result.

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/CAT-1096

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #22595, opened 2025-12-02
feat(core): Introduce native Python code tool for AI agent

ivov:
## Summary

Introduce native Python as an option for the Code Tool that can be invoked by the AI Agent node.

https://linear.app/n8n/issue/CAT-1848

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/task-runner-python/src/message_serde.py
+++ b/packages/@n8n/task-runner-python/src/message_serde.py
@@ -50,6 +50,7 @@ def _parse_task_settings(d: dict) -> BrokerTaskSettings:
         workflow_id = settings_dict.get("workflowId", "Unknown")
         node_name = settings_dict.get("nodeName", "Unknown")
         node_id = settings_dict.get("nodeId", "Unknown")
+        query = settings_dict.get("query")
     except KeyError as e:
         raise ValueError(f"Missing field in task settings message: {e}")
 
@@ -64,6 +65,7 @@ def _parse_task_settings(d: dict) -> BrokerTaskSettings:
             workflow_id=workflow_id,
             node_name=node_name,
             node_id=node_id,
+            query=query,
         ),
     )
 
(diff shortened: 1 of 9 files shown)


packages/@n8n/task-runner-python/src/pipe_reader.py
```diff
@@ -79,9 +79,6 @@ def _validate_pipe_message(self, msg) -> PipeMessage:
         if has_result and has_error:
             raise InvalidPipeMsgContentError("Msg has both 'result' and 'error' keys")
 
-        if has_result and not isinstance(msg["result"], list):
-            raise InvalidPipeMsgContentError("'result' must be a list")
-
```
ivov, line 84: The result for a code tool call is a plain string rather than items, so two options:

1. Introduce a `tool` mode for the runner and conditionally skip list validation in the runner.
2. Skip list validation in the runner and rely on the TypeScript side validating items via `validateRunCodeAllItems` and `validateRunCodeEachItem` in the Code node. 

Opted for the second. At #18928 we centralized all validation n8n-side, so this lingering runner-side check is redundant and couples the runner to n8n more than needed.

tomi:
We need this in the `v2` branch as well

ivov:
> We need this in the `v2` branch as well

Mention here: https://linear.app/n8n/issue/CAT-1848/introduce-native-python-code-tool-for-ai-agent#comment-38fcaab1

packages/@n8n/task-runner-python/src/message_serde.py
```diff
@@ -64,6 +65,7 @@ def _parse_task_settings(d: dict) -> BrokerTaskSettings:
             workflow_id=workflow_id,
             node_name=node_name,
             node_id=node_id,
+            query=query,
```
tomi, line 68: How is query different from code?

tomi (approved):

packages/@n8n/task-runner-python/src/message_serde.py
ivov: `query` is the input string that the AI Agent passes to the tool when it calls it

schrothbn (approved):
Looks great

PR #22602, opened 2025-12-02
feat: Stop bumping `versionId` on every workflow change (no-changelog)

dariacodes:
## Summary

This PR solves the problem with bumping `versionId` on any workflow changes for conflict detection.

- Implement workflow checksum for conflict detection using only meaningful workflow fields  (nodes/connections/settings/etc.), excluding ids/timestamps.
- Bump and save workflow versions in workflow history only on `nodes` or `connections` changes.

Closes [ADO-4481](https://linear.app/n8n/issue/ADO-4481/feature-be-fix-problem-with-having-duplicate-workflow-versions)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/workflow/src/utils.ts
+++ b/packages/workflow/src/utils.ts
@@ -18,8 +18,19 @@ BigInt.prototype.toJSON = function () {
 	return this.toString();
 };
 
+/**
+ * Type guard for plain objects suitable for key-based traversal/serialization.
+ *
+ * Returns `true` for objects whose prototype is `Object.prototype` (object literals)
+ * or `null` (`Object.create(null)`), and `false` for arrays and non-plain objects
+ * such as `Date`, `Map`, `Set`, and class instances.
+ */
 export function isObject(value: unknown): value is Record<string, unknown> {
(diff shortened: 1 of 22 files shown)


packages/workflow/src/utils.ts
```diff
@@ -18,6 +18,10 @@ BigInt.prototype.toJSON = function () {
 	return this.toString();
 };
 
+export function isObject(value: unknown): value is Record<string, unknown> {
+	return value !== null && typeof value === 'object' && !Array.isArray(value);
```
scdekov: nit: `Object.prototype.toString.call({}) === '[object Object]'` this seems to be the preferred way  to check if the value is an object as there are a bunch more non primitives in js returning 'object' for typeof value like Dates, Sets etc

scdekov (dismissed):
Great job!

packages/workflow/src/utils.ts
dariacodes: Good point 👍  I've just pushed an improvement

scdekov (approved):

PR #22605, opened 2025-12-02
fix(editor): Fix displaying of workflow publish info (no-changelog)

scdekov:
## Summary

The workflow publish history was not retuned on the workflows detail endpoint and the activate one but it was needed to be displayed in the UI when publishing respectively from the canvas and the wf history.

https://www.notion.so/n8n/Publish-author-and-date-not-correct-2b75b6e0c94f8006b1dcfb801a246072?source=copy_link

## Review / Merge checklist

- [ ] Tests included. 


CharlieKolb (dismissed):
Ahh we only update the history afterwards, that makes sense. Thanks for debugging this and great work! 🙇

CharlieKolb (approved):

PR #22609, opened 2025-12-02
feat(core): Add credential resolver entity for dynamic credential module

guillaumejacquart:
## Summary

This PR adds the credential resolver entity and db table.
This table will be used by the admin users to create credential resolvers by name, resolver type and configuration (key / value pair). The configuration is stored encrypted in the db

https://linear.app/n8n/issue/PAY-4212/create-resolver-database-entity

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
+++ b/packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
@@ -1,0 +1,20 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const tableName = 'dynamic_credential_resolver';
+
+export class CreateDynamicCredentialResolverTable1764682447000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(tableName)
+			.withColumns(
+				column('id').varchar(16).primary.notNull,
+				column('name').varchar(128).notNull,
+				column('type').varchar(128).notNull,
+				column('config').text.notNull,
+			)
+			.withTimestamps.withIndexOn('type');
+	}
+
+	async down({ schemaBuilder: { dropTable } }: MigrationContext) {
+		await dropTable(tableName);
+	}
+}
--- a/packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
+++ b/packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
@@ -6,10 +6,12 @@ export class CreateDynamicCredentialResolverTable1764682447000 implements Revers
 	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
 		await createTable(tableName)
 			.withColumns(
-				column('id').varchar(16).primary.notNull,
+				column('id').varchar(16).primary,
 				column('name').varchar(128).notNull,
 				column('type').varchar(128).notNull,
-				column('config').text.notNull,
+				column('config').text.notNull.comment(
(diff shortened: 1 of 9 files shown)


packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
```diff
@@ -0,0 +1,20 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const tableName = 'dynamic_credential_resolver';
+
+export class CreateDynamicCredentialResolverTable1764682447000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(tableName)
+			.withColumns(
+				column('id').varchar(16).primary.notNull,
+				column('name').varchar(128).notNull,
```
afitzek, line 10: Just to keep in mind, I think sqlite is not enforcing this length check nor is the annotation on the entity enforcing the length limit.

afitzek (dismissed):
Looks good to me 👍 thanks.

packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
```diff
@@ -0,0 +1,20 @@
...
+				column('type').varchar(128).notNull,
+				column('config').text.notNull,
```
ivov: `.comment()` to clarify what this column holds? Also curious if this should be a JSON column instead?

packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
```diff
@@ -0,0 +1,20 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const tableName = 'dynamic_credential_resolver';
+
+export class CreateDynamicCredentialResolverTable1764682447000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(tableName)
+			.withColumns(
+				column('id').varchar(16).primary.notNull,
```
ivov: Redundant to mark both primary and not null.

packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
ivov, line 10: 1. Curious if this name is set by the user?
2. Maybe or maybe not, should name have a unique constraint? To prevent confusion even if we have UUIDs?

guillaumejacquart: 1. Yes
2. I don't think we should, the same as multiple credentials can have the same name. I'm not against it, but I'm not sure it's needed. Maybe we could, as we do for credentials, suggest unique name on creation (adding a ` (Number)` suffix)

packages/@n8n/db/src/migrations/common/1764682447000-CreateCredentialResolverTable.ts
guillaumejacquart: I added the comment. It cannot be a JSON because it's gonna be encrypted (as it can contain sensitive information like client id / secret for an OAuth resolver). That's why I chose text, like credential data field

ivov (dismissed):
🚀

geemanjs (approved):

afitzek (approved):

BGZStephen (approved):

ivov (approved):

PR #22106, opened 2025-11-20
feat: Reduce unauthentication information in settings endpoint further

ireneea:
## Summary

Change the public setting to only contains properties that are absolutely needed by the non-authenticated pages

## Related Linear tickets, Github issues, and Community forum posts

resolves PAY-3939

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [x] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

---

> [!NOTE]
> Reduces unauthenticated settings to essentials, updates frontend to skip side effects in public mode, moves user/banners init to authenticated flow, and makes logo release channel optional with safer favicon handling.
> 
> - **Backend (CLI)**:
>   - **Public settings**: Redefines `PublicFrontendSettings` to a minimal schema and updates `getPublicSettings()` to return only `settingsMode`, `previewMode`, `authCookie.secure`, `userManagement` auth/setup fields, minimal `sso` flags/URLs, and `enterprise` SSO flags.
> - **Frontend**:
>   - **Settings store**: On `settingsMode: 'public'`, only `setSettings()` is applied; skips root store/telemetry side effects; removes dedicated `setPublicSettings()`.
>   - **App init**: `initializeCore()` no longer initializes users or static banners; `initializeAuthenticatedFeatures()` now initializes `usersStore` (if not preview) and pushes static banners (`NON_PRODUCTION_LICENSE`, `V1`).
>   - **Logo**: `releaseChannel` prop is optional; only customizes logo/favicon for non-`stable` channels; guards favicon with `URL.createObjectURL` check.
> - **Tests**:
>   - Updated unit tests across CLI and editor UI to match minimal public settings and the new initialization/banners flow; router test ensures `usersStore.initialize()` runs during authenticated feature init.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1a7b332b7bfa540e085f9f79fe76ba1c330e8cf2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #22610, opened 2025-12-02
feat(core): Add settings endpoint to public API

konstantintieber:
## Summary

By only returning the `versionCli` property on the existing `/rest/settings` endpoint for authenticated users in https://github.com/n8n-io/n8n/pull/22106, we no longer have a way to fetch the version of an instance via the API, since our internal API doesn't support authorization via API keys.

This PR adds a new `GET /api/v1/settings` endpoint that returns a single `version` property. A `X-N8N-API-KEY` header with a valid API key is required to access this endpoint.

image

**How to test**
- Start cli package locally
- Execute `GET http://localhost:5678/api/v1/settings`

PAY-4252

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/constants.ts
```diff
@@ -21,6 +21,11 @@ const n8nPackageJson = jsonParse<n8n.PackageJson>(readFileSync(packageJsonPath,
 export const N8N_VERSION = n8nPackageJson.version;
 export const N8N_RELEASE_DATE = statSync(packageJsonPath).mtime;
 
+const ROOT_DIR = resolve(__dirname, '../../..');
+const rootPackageJsonPath = join(ROOT_DIR, 'package.json');
+const n8nRootPackageJson = jsonParse<n8n.PackageJson>(readFileSync(rootPackageJsonPath, 'utf8'));
+export const N8N_ROOT_VERSION = n8nRootPackageJson.version;
```
konstantintieber: The existing `N8N_VERSION` constant points to the version inside `packages/cli/package.json`.
There can be a docker image tagged `1.123.3`, where the version of the cli might still be `1.123.2`, because the cli package was not changed in the patch updates from 1.123.2->123.3.
This is why this new `N8N_ROOT_VERSION` constant was added here.
This one will always be in sync with the version of the installed docker image.

I assume this is also why the `GET /rest/settings` response names the property containing the `N8N_VERSION` as `versionCli`, rather than just `version`.

packages/cli/src/public-api/v1/handlers/settings/spec/schemas/settings.yml
```diff
@@ -0,0 +1,10 @@
+type: object
+additionalProperties: false
+required:
+  - version
+properties:
+  version:
```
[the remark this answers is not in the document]
konstantintieber: Fixed in https://github.com/n8n-io/n8n/pull/22610/commits/ede3915a8308e9a266fd473e6011da6c55a03d3c

packages/cli/src/constants.ts
konstantintieber: As was correctly pointed out by the AI review in https://github.com/n8n-io/n8n/pull/22610#discussion_r2581341240, we can't rely on the folder hierarchy of our source code to get the root package.json version.

When running...
- ~/ pnpm run build:docker
- ~/packages/testing/containers pnpm run stack:sqlite

We get this error:
`Error: ENOENT: no such file or directory, open '/usr/local/lib/package.json'`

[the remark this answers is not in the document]
konstantintieber: This is correct. We need a different way to get the tag of the installed docker image.
See https://github.com/n8n-io/n8n/pull/22610#discussion_r2581459813

konstantintieber: Switched to the `N8N_VERSION` in https://github.com/n8n-io/n8n/pull/22610/commits/8f1b9cdf8d9ce8291cdc618b8e9a5fc6c6a70173 for now until we decide how we want to support returning the "root package.json version" in this endpoint.

packages/cli/src/public-api/v1/handlers/settings/settings.handler.ts
```diff
@@ -0,0 +1,13 @@
+import type express from 'express';
+
+import type { SettingsRequest } from '../../../types';
+import { N8N_VERSION } from '@/constants';
+
+export = {
+	getSettings: [
+		async (_req: SettingsRequest.Get, res: express.Response): Promise<express.Response> => {
+			// TODO: expose version from root package.json here, rather than the one from /packages/cli/package.json
```
konstantintieber, line 9: I think the long-term solution for this TODO would be to change the [bump-versions.mjs](https://github.com/n8n-io/n8n/blob/master/.github/scripts/bump-versions.mjs#L54) to increment the version number inside `packages/cli/package.json` on every release by default.

In the vast majority of our releases today the cli package is bumped anyway, so I don't see this as blocking here.

geemanjs (approved):

BGZStephen:
I'm a little confused, is there a reason we don't add the version to `GET http://localhost:5678/rest/settings`, it's an unauthenticated API endpoint?

image

konstantintieber:
@BGZStephen We just removed the version from the unauthenticated response, because of security reasons: https://linear.app/n8n/issue/PAY-3939/reduce-unauthentication-information-in-settings-endpoint

This is why this new public API endpoint is now being added (only accessible with API key).

guillaumejacquart:
I agree with @BGZStephen , if feel like what's already available in the unauthenticated payload for /rest/settings. is already more sensible than the n8n version. I think we could re-add it to the response, no ?

konstantintieber:
Closing in favor of https://github.com/n8n-io/n8n/pull/22804

EDIT: https://github.com/n8n-io/n8n/pull/22804 was reverted.
We'll solve this in a different manner.
Follow-up ticket: https://linear.app/n8n/issue/PAY-4272/

Issue #15556, opened 2025-05-20
Subworkflow does not return last node's output when using "Wait for Webhook call back"


bragaus:
### Bug Description

When using a Subworkflow with a `Wait for Webhook call back` node inside it, the main workflow does not receive the final output of the Subworkflow once it completes. This differs from the expected behavior, where the last node’s output in the Subworkflow is returned to the main workflow.

### To Reproduce

1. Create a main workflow that calls a Subworkflow via the "Execute Workflow" node.
2. Inside the Subworkflow, use a "Wait for Webhook call back" node.
3. Trigger the Subworkflow by running the main workflow.
4. Once the webhook is called, the Subworkflow resumes and finishes as expected.
5. The main workflow resumes, but the output of the Subworkflow is not returned (i.e., the last node's data is missing).

### Expected behavior

The output of the last node in the Subworkflow should be returned to the main workflow after the Subworkflow finishes, even when using `Wait for Webhook call back`.

[Teste2.json](https://github.com/user-attachments/files/20354365/Teste2.json)
[Teste1.json](https://github.com/user-attachments/files/20354364/Teste1.json)

### Operating System

Ubuntu 20.04

### n8n Version

1.93.0

### Node.js Version

v22.14.0

### Database

SQLite (default)

### Execution mode

main (default)

Joffcom:
Hey @bragaus,

We have created an internal ticket to look into this which we will be tracking as "GHC-2089"

Issue #13135, opened 2025-02-07
Calling a sub-workflow with a wait by webhook node, don't return the data of the last node, but the node before the wait.


Arnaud-MARIE:
## Describe the problem/error/question
When calling a sub-workflow containing a Wait (Webhook) node, the return data is incorrect:
- When called as Agent Tool: Returns data from node before Wait
- When called as Workflow: Returns initial input data
- Expected: Should return data from last node after webhook completion

## Steps to Reproduce
1. Create sub-workflow with:
  - Initial API call
  - Wait (Webhook) node
  - Final data processing
2. Call this sub-workflow from main workflow
3. Wait for webhook completion
4. Check returned data in main workflow

## Current Behavior
- When called as Agent Tool: Returns data from the node before the Wait node
- When called as Workflow: Returns only the initial input data
- Webhook executes successfully but final data is not returned

## Expected Behavior
Should return the processed data from the last node after webhook completion

## Technical Details
- Using Wait node in 'Webhook' mode
- Webhook successfully triggers and completes
- Sub-workflow executes completely
- Issue occurs in data return to parent workflow

## Please share your workflow/screenshots/recording

![Image](https://github.com/user-attachments/assets/6bd57a42-c4c6-4528-a61a-9b244722022d)

![Image](https://github.com/user-attachments/assets/fdf86c08-e4ed-4fc2-a926-601635486c7e)

![Image](https://github.com/user-attachments/assets/49dc216e-2399-4b2c-9098-4f8e22a04f64)

```

Sub workflow : 
{
  "nodes": [
    {
      "parameters": {
        "resume": "webhook",
        "httpMethod": "POST",
        "responseMode": "lastNode",
        "limitWaitTime": true,
        "resumeAmount": 10,
        "resumeUnit": "minutes",
        "options": {}
      },
      "type": "n8n-nodes-base.wait",
      "typeVersion": 1.1,
      "position": [
        400,
        0
      ],
      "id": "ff89d4fa-5c3b-48ba-81cb-abb6554445ac",
      "name": "Wait response",
      "webhookId": "7e72cb0f-e29c-4e74-9509-95cc75e8e836"
    },
    {
      "parameters": {
        "assignments": {
          "assignments": [
            {
              "id": "e2347a4a-cc19-4049-8a19-e11a411b648d",
              "name": "test",
[...]

PR #22611, opened 2025-12-02
fix(core): Make sub-workflows with waits return correct data to parents

despairblue:
## Summary

Fixes parent workflows receiving incorrect data from child workflows that contain Wait nodes. Parents now receive data from the child's **last node** after the wait completes.

Also fixes a race condition where multiple child executions (in "run once for each item" mode) could resume the same parent execution multiple times.

## Related Linear Issues
- https://linear.app/n8n/issue/CAT-1801/subworkflow-doesnt-return-correct-data-after-waiting
- https://linear.app/n8n/issue/PAY-2720/community-issue-calling-a-sub-workflow-with-a-wait-by-webhook-node

## Related GitHub Issues
- https://github.com/n8n-io/n8n/issues/13135
- https://github.com/n8n-io/n8n/issues/15556

## Breaking Change

This is a **breaking change** for v2 release. The behavior changes from returning no data to returning data after the wait completes, which is the expected behavior but differs from historical functionality.

### Core Fix
1. **New helper function** `updateParentExecutionWithChildResults()` in `workflow-helpers.ts`
   - Updates parent's `nodeExecutionStack[0].data` with child's final output
   - Applied in both execution paths: time-based waits (WaitTracker) and webhook-based waits (WebhookHelpers)

### Race Condition Fix
2. **Atomic status updates** in `ExecutionRepository.updateExistingExecution()`
   - Added optional `requireStatus` parameter
   - Uses WHERE clause: `WHERE id = ? AND status = ?` for atomic check-and-update
   - Returns boolean indicating success/failure

3. **Exception handling** for concurrent resume attempts
   - New `ExecutionAlreadyResumingError` exception
   - `ActiveExecutions.add()` throws when another process already resumed execution
   - `WaitTracker.startExecution()` catches and handles gracefully

## How It Works

**Before (Buggy Behavior)**:
1. Child hits Wait node → Parent received no data from the sub-execution
2. Child completes after wait → Parent again received no data from the sub-execution
3. Result: Parent gets wrong data (the input of the "Execute workflow" node)

**After (Fixed Behavior)**:
1. Child hits Wait node → Parent pauses
2. Child completes after wait → Updates parent's execution data in DB
3. Parent resumes → Execute Workflow node returns updated data

**Race Condition Prevention**:
- Multiple children try to resume parent → Only first succeeds (atomic DB update)
- Others get `ExecutionAlreadyResumingError` → Caught and handled
- Parent executes only once with correct data

**E2E Test (1)**:
- `packages/testing/playwright/tests/ui/subworkflow-wait.spec.ts`
  - Full flow with real workflows and webhook triggers
  - Verifies parent receives `'child - after'` not `'child - before'`

## Review Notes

- This fixes long-standing community issue with 30+ comments
- Promise.race semantics maintained (parent resumes on first child completion)
- Race condition in "run once for each item" mode is documented and acceptable
- No performance impact - single additional DB query per child completion

Co-Authored-By: Claude <noreply@anthropic.com>


zlebandit:
Hi, I've just tested a use case with the v2.0.0-rc.0 and it did not work....

I show you my workflow : 

`{
  "nodes": [
    {
      "parameters": {
        "assignments": {
          "assignments": [
            {
              "id": "87abd7a2-a07e-454f-9497-314567bd5f35",
              "name": "tableName",
              "value": "=scrapping_{{ $now.toMillis() }}_{{ $executionId.replace(/-/g, '_') }}",
              "type": "string"
            }
          ]
        },
        "options": {}
      },
      "type": "n8n-nodes-base.set",
      "typeVersion": 3.4,
      "position": [
        32,
        -112
      ],
      "id": "4f0e092f-b854-4ad2-998d-ec77b29f661e",
      "name": "Nom Table"
    },
    {
      "parameters": {
        "operation": "executeQuery",
        "query": "CREATE TABLE {{ $json.tableName }} (\n    id BIGSERIAL PRIMARY KEY,\n    link TEXT NOT NULL,\n    data TEXT,\n    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP\n);",
        "options": {}
      },
      "type": "n8n-nodes-base.postgres",
      "typeVersion": 2.5,
      "position": [
        176,
        -112
      ],
      "id": "637357a5-4d0b-4169-bf28-e94aea673fef",
      "name": "Create Table",
      "credentials": {
        "postgres": {
          "id": "NxNKscv1VnZmx1FB",
          "name": "Postgres (n8n_data)"
        }
      }
    },
    {
      "parameters": {
        "mode": "chooseBranch",
        "useDataOfInput": 2
      },
      "type": "n8n-nodes-base.merge",
      "typeVersion": 3.2,
      "position": [
        320,
        0
      ],
      "id": "4d1144d5-9514-418d-8467-ceba99fa32c8",
      "name": "Merge"
    },
    {
      "parameters": {
        "workflowId": {
          "__rl": true,
          "value": "sIgdYNA7vFhVl2sg",
          "mode": "list",
          "cachedResultUrl": "/workflow/sIgdYNA7vFhVl2sg",
          "cachedResultName": "-> Scrape and summarize TEST"
        },
        "workflowInputs": {
          "mappingMode": "defineBelow",
          "value": {
            "link": "={{ $('Start').item.json.link }}",
            "tableName": "={{ $('Nom Table').first().json.tableName }}",
            "totalLignes": "={{ $('Start').all().length }}",
            "webhookUrl": "={{ $execution.resumeUrl }}"
          },
          "matchingColumns": [
            "link"
          ],
          "schema": [
            {
              "id": "link",
              "displayName": "link",
              "required": false,
              "defaultMatch": false,
              "display": true,
              "canBeUsedToMatch": true,
              "type": "string",
              "removed": false
            },
            {
              "id": "tableName",
              "displayName": "tableName",
              "required": false,
              "defaultMatch": false,
              "display": true,
              "canBeUsedToMatch": true,
              "type": "string",
              "removed": false
            },
            {
              "id": "totalLignes",
              "displayName": "totalLignes",
              "required": false,
              "defaultMatch": false,
              "display": true,
              "canBeUsedToMatch": true,
              "type": "number",
              "removed": false
            },
            {
              "id": "webhookUrl",
              "displayName": "webhookUrl",
              "required": false,
              "defaultMatch": false,
              "display": true,
              "canBeUsedToMatch": true,
              "type": "string",
              "removed": false
            }
          ],
          "attemptToConvertTypes": false,
          "convertFieldsToString": true
        },
        "mode": "each",
        "options": {
          "waitForSubWorkflow": false
        }
      },
      "type": "n8n-nodes-base.executeWorkflow",
      "typeVersion": 1.2,
      "position": [
        -256,
        224
      ],
      "name": "Call Sous-agent Scrapping",
      "id": "af6ad2dc-42b8-4f38-b398-1aefd0f0831f"
    },
    {
      "parameters": {},
      "type": "n8n-nodes-base.limit",
      "typeVersion": 1,
      "position": [
        -112,
        -112
      ],
      "id": "1beddc00-b393-487b-8b79-dd5d1ad450b5",
      "name": "Limit"
    },
    {
      "parameters": {},
      "type": "n8n-nodes-base.limit",
      "typeVersion": 1,
      "position": [
        -112,
        224
      ],
      "id": "6f7f1e7c-daed-4d86-afbe-dad7fb91cfc8",
      "name": "Limit1"
    },
    {
      "parameters": {
        "workflowInputs": {
          "values": [
            {
              "name": "link"
            }
          ]
        }
      },
      "type": "n8n-nodes-base.executeWorkflowTrigger",
      "typeVersion": 1.1,
      "position": [
        -256,
        0
      ],
      "id": "032ef014-593d-4fff-8c8f-3702308717b8",
      "name": "Start"
    },
    {
      "parameters": {
        "resume": "webhook",
        "options": {}
      },
      "type": "n8n-nodes-base.wait",
      "typeVersion": 1.1,
      "position": [
        32,
        224
      ],
      "id": "75578d81-447e-4591-b990-4d5a240a275d",
      "name": "Wait",
      "webhookId": "1edd2d75-e26b-4059-b6e5-3b6df1f0698a"
    },
    {
      "parameters": {
        "operation": "executeQuery",
        "query": "DROP TABLE IF EXISTS {{ $('Nom Table').first().json.tableName }};",
        "options": {}
      },
      "type": "n8n-nodes-base.postgres",
      "typeVersion": 2.5,
      "position": [
        320,
        224
      ],
      "id": "b0ab44f5-26d0-413f-ad49-afce00e7d0e4",
      "name": "Delete Table",
      "credentials": {
        "postgres": {
          "id": "NxNKscv1VnZmx1FB",
          "name": "Postgres (n8n_data)"
        }
      }
    },
    {
      "parameters": {
        "operation": "executeQuery",
        "query": "SELECT json_agg(json_build_object('link', link, 'data', data)) AS items\nFROM {{ $('Nom Table').first().json.tableName }};",
        "options": {}
      },
      "type": "n8n-nodes-base.postgres",
      "typeVersion": 2.5,
      "position": [
        176,
        224
      ],
      "id": "bcc0909a-854b-406b-b3d6-dee997f680e4",
      "name": "Get data",
      "retryOnFail": false,
      "maxTries": 2,
      "waitBetweenTries": 500,
      "credentials": {
        "postgres": {
          "id": "NxNKscv1VnZmx1FB",
          "name": "Postgres (n8n_data)"
        }
      }
    },
    {
      "parameters": {
        "jsCode": "return $('Get data').first().json.items.map(item => ({\n  json: {\n    link: item.link,\n    data: item.data\n  }\n}));"
      },
      "type": "n8n-nodes-base.code",
      "typeVersion": 2,
      "position": [
        464,
        224
      ],
      "id": "5f8739a0-bdcb-4229-b5be-abc7c3c6a59c",
      "name": "Output"
    }
  ],
  "connections": {
    "Nom Table": {
      "main": [
        [
          {
            "node": "Create Table",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Create Table": {
      "main": [
        [
          {
            "node": "Merge",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Merge": {
      "main": [
        [
          {
            "node": "Call Sous-agent Scrapping",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Call Sous-agent Scrapping": {
      "main": [
        [
          {
            "node": "Limit1",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Limit": {
      "main": [
        [
          {
            "node": "Nom Table",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Limit1": {
      "main": [
        [
          {
            "node": "Wait",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Start": {
      "main": [
        [
          {
            "node": "Limit",
            "type": "main",
            "index": 0
          },
          {
            "node": "Merge",
            "type": "main",
            "index": 1
          }
        ]
      ]
    },
    "Wait": {
      "main": [
        [
          {
            "node": "Get data",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Delete Table": {
      "main": [
        [
          {
            "node": "Output",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Get data": {
      "main": [
        [
          {
            "node": "Delete Table",
            "type": "main",
            "index": 0
          }
        ]
      ]
    }
  },
  "pinData": {},
  "meta": {
    "templateCredsSetupCompleted": true,
    "instanceId": "250454b1fbe6ec327f9e0fa5f7c54818b4e831f2b98aa45ea794fa241534bd84"
  }
}`

despairblue:
@zlebandit v2.0.0-rc.0 does not include this fix yet. It needs to be reviewed and merged first.
I did try your workflow as a parent-workflow and also as a sub-workflow and it does work with this branch. So once this is merged and a new rc (release candidate) is released you can try that one. It should work then.

Thanks for checking ❤️

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -446,28 +450,58 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+				const whereCondition: { id: string; status?: ExecutionStatus } = { id: executionId };
+				if (requireStatus) whereCondition.status = requireStatus;
+
+				const result = await tx.update(ExecutionEntity, whereCondition, executionInformation);
+				executionTableAffectedRows = result.affected ?? 0;
+
+				// If requireStatus was set and the update failed, don't update executionData
+				// This prevents modifying data for an execution another process is handling
+				if (requireStatus && executionTableAffectedRows === 0) {
+					return;
```
mfsiega: `return false`?

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -446,28 +450,58 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
 				await tx.update(ExecutionData, { executionId }, executionData);
 			}
 		});
+
+		// When requireStatus is set, only return true if the execution table update succeeded
+		if (requireStatus) {
+			return executionTableAffectedRows > 0;
+		}
+
+		// When no status requirement, any update counts as success (backward compatibility)
```
mfsiega: It looks like we were previously returning void - what do we need to be backward compatible with?

More generally, what are the semantics of what we're returning here? Maybe add it as a comment on the method.

packages/cli/src/__tests__/wait-tracker.test.ts
```diff
@@ -145,109 +146,199 @@ describe('WaitTracker', () => {
...
-						startedAt: parentExecution.startedAt,
-					},
+					expect.any(Object),
 					false,
 					false,
 					parentExecution.id,
 				);
 			});
 
-			it('should not resume parent execution when shouldResume is false', async () => {
```
mfsiega, line 216: I can't say for sure with the diff but it looks like we're losing test coverage for the case where this `shouldResume` flag is false? It's used by the ExecuteWorkflow node to indicate whether the parent is waiting for the child(ren), and therefore whether the child(ren) should attempt to resume the parent e.g., [here](https://github.com/n8n-io/n8n/pull/22611/files?diff=unified&w=0#diff-02229804d7f90397a4e6a449387f36d5c4b1f16bb7b3a6f3c5f00aba57d9c66cR148).

Maybe I'm missing something that's changing here? But if not, maybe just add this test back?

mfsiega (commented):
A few small comments, nothing major. Main fix looks 👍 .

packages/@n8n/db/src/repositories/execution.repository.ts
despairblue: Whatever is returned here will be the return value of the `await this.manager.transaction()`, which we don't use:

https://github.com/n8n-io/n8n/blob/2d2427ac0d51b710851db76b2629f541fb68bce8/packages/@n8n/db/src/repositories/execution.repository.ts#L478-L489

This line is what makes it return false:

https://github.com/n8n-io/n8n/blob/2d2427ac0d51b710851db76b2629f541fb68bce8/packages/@n8n/db/src/repositories/execution.repository.ts#L498-L501

The early return in the transaction is just aborting the transaction earlier so we don't update the execution data if we failed to update the execution.

Hope that makes sense.

mfsiega: Makes perfect sense, was just skimming and didn't read it properly. Sorry about that!

packages/cli/src/__tests__/wait-tracker.test.ts
despairblue: Good catch. Never trust AI.
I restored them and only fixed them up by returning a non-mocked IRun.

2026d5ea72f49e732865e2c0267f99cbec46274f

packages/@n8n/db/src/repositories/execution.repository.ts
despairblue: Added a doc string: c20550b702743065c2a6216b9edea1b695e532a9

packages/@n8n/db/src/repositories/execution.repository.ts
despairblue: All good, it was confusing, re-reading I was able to simplify it: a181e79e49db8e155e38bc6e1a01c90aa968b686

packages/cli/src/workflow-helpers.ts
```diff
@@ -217,6 +218,60 @@ export function shouldRestartParentExecution(
...
+	subworkflowResults: IRun,
+): Promise<void> {
+	const lastExecutedNodeData = getDataLastExecutedNodeData(subworkflowResults);
+	if (!lastExecutedNodeData?.data) return;
+	const parent = await executionRepository.findSingleExecution(parentExecutionId, {
+		includeData: true,
+		unflattenData: true,
+	});
+
+	if (!parent || parent.status !== 'waiting') {
```
mfsiega, line 250:
```suggestion
	if (parent?.status !== 'waiting') {
```

packages/cli/src/workflow-helpers.ts
```diff
@@ -217,6 +218,60 @@ export function shouldRestartParentExecution(
...
+		return;
+	}
+
+	const parentWithSubWorkflowResults = { data: { ...parent.data } };
+
+	if (
```
mfsiega, line 256:
```suggestion
    const nodeExecutionStack = parentWithSubWorkflowResults.data.executionData?.nodeExecutionStack;
	if (!nodeExecutionStack || nodeExecutionStack.length === 0) {
	    return;
	}
```

packages/cli/src/workflow-helpers.ts
```diff
@@ -217,6 +218,60 @@ export function shouldRestartParentExecution(
...
+		!parentWithSubWorkflowResults.data.executionData?.nodeExecutionStack ||
+		parentWithSubWorkflowResults.data.executionData.nodeExecutionStack.length === 0
+	) {
+		return;
+	}
+
+	// Copy the sub workflow result to the parent execution's Execute Workflow node inputs
+	// so that the Execute Workflow node returns the correct data when parent execution is resumed
+	// and the Execute Workflow node is executed again in disabled mode.
+	parentWithSubWorkflowResults.data.executionData.nodeExecutionStack[0].data =
```
mfsiega, line 266:
```suggestion
	nodeExecutionStack[0].data =
```

packages/cli/src/workflow-helpers.ts
```diff
@@ -217,6 +218,60 @@ export function shouldRestartParentExecution(
...
+	if (
+		!parentWithSubWorkflowResults.data.executionData?.nodeExecutionStack ||
+		parentWithSubWorkflowResults.data.executionData.nodeExecutionStack.length === 0
+	) {
+		return;
+	}
+
+	// Copy the sub workflow result to the parent execution's Execute Workflow node inputs
+	// so that the Execute Workflow node returns the correct data when parent execution is resumed
+	// and the Execute Workflow node is executed again in disabled mode.
```
mfsiega, line 265: Just a small note - I'm basically just trusting that this is how everything's structured e.g., that the most recent node is at index 0. I wonder if it would help the code health to introduce classes etc that encapsulate those details.

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -423,7 +423,19 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
 	}
 
-	async updateExistingExecution(executionId: string, execution: Partial<IExecutionResponse>) {
+	/**
+	 * Update an existing execution in the database.
+	 *
+	 * @param executionId - The ID of the execution to update
+	 * @param execution - Partial execution data to update
+	 * @param requireStatus - Optional status requirement. If provided, update only succeeds if execution has this status
+	 * @returns true if update succeeded, false if requireStatus condition was not met
```
mfsiega, line 432: Strictly speaking we would also return false if the executionId didn't exist, right?

I would marginally prefer to just return true if any rows were affected, but it could be that I'm missing some reason that that's not good. In any case I don't feel _so_ strongly.

mfsiega (approved):
Just a few small comments, LGTM.

packages/cli/src/workflow-helpers.ts
despairblue: For some things that could make sense, e.g. in this case to clear up on the stack semantics. 

We use `shift` to get the next node from the stack:

https://github.com/n8n-io/n8n/blob/a181e79e49db8e155e38bc6e1a01c90aa968b686/packages/core/src/execution-engine/workflow-execute.ts#L1501-L1502

Which means the first one in there should be the one that waits.

zlebandit:
> @zlebandit v2.0.0-rc.0 does not include this fix yet. It needs to be reviewed and merged first. I did try your workflow as a parent-workflow and also as a sub-workflow and it does work with this branch. So once this is merged and a new rc (release candidate) is released you can try that one. It should work then.
> 
> Thanks for checking ❤️

Hi, just tested the rc1 and..... IT WORKED !!
OUAHOUUUU !

packages/cli/src/workflow-helpers.ts
despairblue: addressed in: https://github.com/n8n-io/n8n/pull/22745

packages/cli/src/workflow-helpers.ts
despairblue: addressed in: https://github.com/n8n-io/n8n/pull/22745

packages/cli/src/workflow-helpers.ts
despairblue: addressed in: https://github.com/n8n-io/n8n/pull/22745

packages/@n8n/db/src/repositories/execution.repository.ts
despairblue: addressed in: https://github.com/n8n-io/n8n/pull/22745

PR #22620, opened 2025-12-02
fix(core): Fix workflow statistics constraint

guillaumejacquart:
## Summary

This PR modifies the workflow_statistics table to preserve statistics when workflows are deleted, instead of cascading deletes.

### The problem
Currently, the on delete cascade on the workflow statistics `workflowId` field means than when a workflow is deleted, the linked statistics are deleted too. This goes against the purpose of the statistics, which is to be an always increasing cumulative number (especially used for licensing quotas)

### Key Changes:

- Schema Changes:
  - Added auto-incrementing id as new primary key (required for nullable FK)
  - Removed FK constraint of workflow ID, to allow orphaned statistics linking to non existent workflow (deleted workflow).  This is to ease debugging the number (on top of having the workflow name)
  - Added workflowName column to add more readable information to the statistics, especially on workflow deletion
  - Changed count and rootCount from INT to BIGINT for large cumulative values
  - Added unique index on (workflowId, name)
- Tests Added/Updated:
  - Migration tests verifying behavior
  - Integration tests verifying workflowName population

## Review / Merge checklist

- [X] Tests included. 


packages/@n8n/db/src/migrations/mysqldb/1764669632000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,99 @@
...
+		);
+
+		// Step 6: Populate workflowName from existing workflows
+		await queryRunner.query(
+			`UPDATE ${tablePrefix}workflow_statistics ws JOIN ${tablePrefix}workflow_entity we ON ws.\`workflowId\` = we.\`id\` SET ws.\`workflowName\` = we.\`name\``,
+		);
+
+		// Step 7: Add unique index on (workflowId, name) to maintain data integrity
+		await queryRunner.query(
+			`CREATE UNIQUE INDEX \`IDX_${tablePrefix}workflow_statistics_workflow_name\` ON ${tablePrefix}workflow_statistics (\`workflowId\`, \`name\`)`,
```
afitzek: Do we need to make sure that this condition is true in all possible cases? For example if a workflow is deleted, the workflowId becomes null, creating a new workflow with the same name, could this be a conflict in this table?

packages/@n8n/db/src/migrations/mysqldb/1764669632000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,99 @@
...
+		await queryRunner.query(
+			`ALTER TABLE ${tablePrefix}workflow_statistics DROP FOREIGN KEY \`fk_${tablePrefix}workflow_statistics_workflow_id\``,
+		);
+
+		// Step 2: Drop the primary key (workflowId, name)
+		await queryRunner.query(`ALTER TABLE ${tablePrefix}workflow_statistics DROP PRIMARY KEY`);
+
+		// Step 3: Add a new auto-generated id column as primary key
+		await queryRunner.query(
+			`ALTER TABLE ${tablePrefix}workflow_statistics ADD COLUMN \`id\` INT AUTO_INCREMENT PRIMARY KEY FIRST`,
```
afitzek: Do we expect large numbers here as well, that could overflow the INT datatype. I'm not sure how the workflow_statistics table is used.

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,107 @@
...
+
+		// Step 5: Populate workflowName from existing workflows
+		await queryRunner.query(
+			`UPDATE ${tablePrefix}workflow_statistics ws SET "workflowName" = we."name" FROM ${tablePrefix}workflow_entity we WHERE ws."workflowId" = we."id"`,
+		);
+
+		// Step 6: Add unique constraint on (workflowId, name) to maintain data integrity
+		// Note: Must be a regular unique index (not partial) for ON CONFLICT to work in upsertWorkflowStatistics
+		await queryRunner.query(
+			`CREATE UNIQUE INDEX "IDX_${tablePrefix}workflow_statistics_workflow_name" ON ${tablePrefix}workflow_statistics ("workflowId", "name")`,
```
afitzek, line 42: Do we need to make sure that this condition is true in all possible cases? For example if a workflow is deleted, the workflowId becomes null, creating a new workflow with the same name, could this be a conflict in this table?

packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,82 @@
...
+ * - Add auto-incrementing id as primary key
+ * - Make workflowId nullable for SET NULL to work
+ * - Add unique index on (workflowId, name)
+ */
+export class ChangeWorkflowStatisticsFKToNoAction1764669632000 implements ReversibleMigration {
+	async up({ queryRunner, tablePrefix }: MigrationContext) {
+		// Create new table with id primary key, workflowName column, and SET NULL foreign key
+		await queryRunner.query(`
+			CREATE TABLE "${tablePrefix}TMP_workflow_statistics" (
+				"id" INTEGER PRIMARY KEY AUTOINCREMENT,
```
afitzek, line 18: Is this datatype big enough? I'm not sure what the multiplying factor for this table is going to be.

packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,82 @@
...
+		await queryRunner.query(`DROP TABLE "${tablePrefix}workflow_statistics"`);
+
+		// Rename new table
+		await queryRunner.query(
+			`ALTER TABLE "${tablePrefix}TMP_workflow_statistics" RENAME TO "${tablePrefix}workflow_statistics"`,
+		);
+
+		// Create unique index on (workflowId, name) to maintain data integrity
+		await queryRunner.query(
+			`CREATE UNIQUE INDEX "IDX_${tablePrefix}workflow_statistics_workflow_name" ON "${tablePrefix}workflow_statistics" ("workflowId", "name")`,
```
afitzek, line 46: Do we need to make sure that this condition is true in all possible cases? For example if a workflow is deleted, the workflowId becomes null, creating a new workflow with the same name, could this be a conflict in this table?

packages/@n8n/db/src/utils/transformers.ts
```diff
@@ -35,3 +35,12 @@ const jsonColumn: ValueTransformer = {
...
 
 export const sqlite = { jsonColumn };
+
+/**
+ * Transformer for bigint columns that returns strings from PostgreSQL.
+ * Converts string values to numbers on read.
+ */
+export const bigintStringToNumber: ValueTransformer = {
+	to: (value: number): number => value,
+	from: (value: string | number): number => (typeof value === 'string' ? Number(value) : value),
```
afitzek, line 45: I'm not sure that number is the approriate js type to store bigint. afaik there is a bigint type as well in js. Especially if we expect very high numbers, this might overflow the js number type.

afitzek (commented):
I'm not sure about mainly two things here:
1. Is the unique constraint good and will it not prohibit us from inserting values, in case workflows get deleted and have the same name, or have the same name in different projects, etc.
2. Is the js data type of number enough for handling BigInt from Postgres, the ranges definitely differ a lot, but I'm not sure what the value ranges are that we are expecting.

packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: There is no bigint in sqlite. INTEGER type max value is 2⁶³

packages/@n8n/db/src/utils/transformers.ts
guillaumejacquart: I thought about this, but it implies lots of change to handle js bigint (transformers for typeorm, typing in the typeorm find methods,  all comparisons with numbers in tests, ...). Plus typeorm sqlite driver converts INTEGER to number anyway (although sqlite integers can be bigger than javascript ones). So ideally, it would require storing as text instead...
But it should be fine as JS handles 2^53-1 max, which is more than enough for us. WDYT ?

packages/@n8n/db/src/migrations/mysqldb/1764669632000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: name here is the name of the metric (production_executions, ...). SQLite and postgres are both fine with multiple null values, so duplicated (null, 'production_executions') index values is working.
This makes me think that it would be nice to have the name of the workflow as well in the history for when it's deleted

packages/@n8n/db/src/migrations/mysqldb/1764669632000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: I deleted the mysql one as it's not a supported engine anymore. Postgres use serial and sqlite use integer (with 2^63 max value)

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: name here is the name of the metric (production_executions, ...). SQLite and postgres are both fine with multiple null values, so duplicated (null, 'production_executions') index values is working.
This makes me think that it would be nice to have the name of the workflow as well in the history for when it's deleted

packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: Should be, 2^63-1 is quite large

packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: Same here

packages/@n8n/db/src/utils/transformers.ts
guillaumejacquart: Yes, I thought about that, but changing to bigint is quite impacting (dto schemas, types in the typeorm repos find functions, strict comparison with integers, typeorm transformer same on tests, ...). I was afraid of missing a comparison somewhere. But I can do a followup PR though
But in JS, max safe int is 2^53-1, which seem more than enough

packages/@n8n/db/src/migrations/mysqldb/1764669632000-ChangeWorkflowStatisticsFKToNoAction.ts
afitzek: I'm wondering if it makes sense to also store the workflow id, even when the workflow is deleted, so that we can still distinguish, between workflows with the same name and also group the metrics correctly even if there are workflow name conflicts. If the name here is the name of the metric, we might not be able to group together the correct statistics to a workflow instance, if we remove the id?

packages/@n8n/db/src/utils/transformers.ts
afitzek: 2^53-1 is a lot, just wanted to bring it up. 👍

packages/@n8n/db/src/migrations/mysqldb/1764669632000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: Hum, that's a good point, although for now we do not group by workflow id , the stats are used for license metrics but not by workflow (I'll double check though).
To do what you suggest, it would mean removing the foreign key constraint (as there is no ON DELETE KEEP or something like that).
I'll see if that makes sense. Maybe the foreign key constraint is not relevant for that use case 🤔 . Any opinion ?

guillaumejacquart: I updated the PR to remove the FK, as it feels contradictory to what we want to do as you pointed out. We actually want to keep the statistics with proper workflow id / name, so we do not want to have the FK constraint

afitzek (dismissed):
Nice looks good, thank you for addressing the comments! 🚀

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
ivov, line 1: Description says:
> Foreign Key Behavior: Changed ON DELETE CASCADE → ON DELETE SET NULL so statistics are retained when a workflow is deleted

But we're actually dropping the FK!

packages/cli/test/migration/1767018516000-change-workflow-statistics-fk-to-no-action.test.ts
```diff
@@ -0,0 +1,432 @@
...
+		]);
+	}
+
+	it('should preserve statistics after workflow deletion (no FK constraint) and CASCADE after rollback', async () => {
+		// Debug: Check schema BEFORE migration
+		const contextBefore = createTestMigrationContext(dataSource);
+		const schemaBefore = await contextBefore.queryRunner.query(
+			"SELECT sql FROM sqlite_master WHERE type='table' AND name='workflow_statistics'",
+		);
+		console.log('Schema BEFORE migration:', JSON.stringify(schemaBefore, null, 2));
```
ivov:
```suggestion

```

packages/cli/test/migration/1767018516000-change-workflow-statistics-fk-to-no-action.test.ts
```diff
@@ -0,0 +1,432 @@
...
+
+		// Create context AFTER migration runs (since runSingleMigration reinitializes the connection)
+		dataSource = Container.get(DataSource);
+		let context = createTestMigrationContext(dataSource);
+
+		// Debug: Check if FK constraint exists
+		const schema = await context.queryRunner.query(
+			"SELECT sql FROM sqlite_master WHERE type='table' AND name='workflow_statistics'",
+		);
+		console.log('Schema AFTER migration:', JSON.stringify(schema, null, 2));
```
ivov:
```suggestion

```

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,102 @@
...
+		);
+
+		// Step 2: Drop the unique index
+		await queryRunner.query(
+			`DROP INDEX IF EXISTS "IDX_${tablePrefix}workflow_statistics_workflow_name"`,
+		);
+
+		// Step 3: Delete any orphaned statistics rows (workflowId IS NULL)
+		await queryRunner.query(
+			`DELETE FROM ${tablePrefix}workflow_statistics WHERE "workflowId" IS NULL`,
```
ivov: Do I understand correctly that orphans would have an ID pointing to nothing, i.e. we're _not_ setting them to `NULL` in the up migration?

packages/cli/test/migration/1767018516000-change-workflow-statistics-fk-to-no-action.test.ts
```diff
@@ -0,0 +1,432 @@
...
+		await context.queryRunner.query(`DELETE FROM ${tableName} WHERE ${idColumn} = ${placeholder}`, [
+			workflowId,
+		]);
+	}
+
+	it('should preserve statistics after workflow deletion (no FK constraint) and CASCADE after rollback', async () => {
+		// Debug: Check schema BEFORE migration
+		const contextBefore = createTestMigrationContext(dataSource);
+		const schemaBefore = await contextBefore.queryRunner.query(
+			"SELECT sql FROM sqlite_master WHERE type='table' AND name='workflow_statistics'",
```
ivov: This won't work on postgres?

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,102 @@
...
+		);
+		await queryRunner.query(
+			`ALTER TABLE ${tablePrefix}workflow_statistics ALTER COLUMN "rootCount" TYPE BIGINT`,
+		);
+	}
+
+	async down({ queryRunner, tablePrefix }: MigrationContext) {
+		// Step 1: Change count and rootCount columns back to INTEGER
+		await queryRunner.query(
+			`ALTER TABLE ${tablePrefix}workflow_statistics ALTER COLUMN "count" TYPE INTEGER`,
```
ivov, line 65: This can overflow?

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
```diff
@@ -0,0 +1,102 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+/**
+ * Removes the foreign key constraint from workflow_statistics to allow keeping
+ * statistics for deleted workflows. The workflowId becomes an orphaned reference,
```
ivov, line 5: Why do we keep orphaned refs indefinitely instead of cleaning them up?

ivov (commented):
Quick pass

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: Yes, we allow orphans with real ID pointing to nothing (cause the workflow was deleted).
This line actually is not very useful, as there should be no line with workflowId null. But it's just a security to make sure there are no null values for when we re-introduce the FK with on delete cascade

packages/cli/test/migration/1767018516000-change-workflow-statistics-fk-to-no-action.test.ts
guillaumejacquart: Sorry, debugging lines for test that I forgot to remove

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: I updated the description, thanks. We want to drop the FK

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: we want to keep the orphaned workflow statistics because they shoud count for the computation of license metrics. If we delete them on workflow deletion (like we do now), then the cumulated production execution computed by the license metrics service would decrease, which is not supposed to happen (an leads to issues on the license server)

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: Ah yes true... I wonder what we should do in that case. Maybe the migration should be just irreversible 🤔  ?

guillaumejacquart: I updated the overflowing value to 0, "resetting" the count for those workflow stats. I think it's mostly "fine", because license server has a mechanism in place to handle data reset

packages/@n8n/db/src/migrations/postgresdb/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
guillaumejacquart: I removed the line and kept the workflow NOT NULLABLE (as it was before). so that we are sure there are no lines with null workflowId.
Also added the deletion of orphans before resetting the FK on down

packages/@n8n/db/src/entities/workflow-statistics.ts
```diff
@@ -1,26 +1,33 @@
...
-	@PrimaryColumn({ length: 128 })
+	@Column({ length: 128 })
 	name: StatisticsNames;
 
-	@ManyToOne('WorkflowEntity', 'shared')
-	workflow: WorkflowEntity;
+	// workflowId is kept as an orphaned reference when workflows are deleted
+	// No FK constraint in database - allows keeping statistics for deleted workflows
+	@Column({ type: 'varchar', length: 36, nullable: true })
+	workflowId: string | null;
```
ivov: Migration says this is not nullable on both DBs?

ivov (commented):
Thanks for addressing everything, found a schema mismatch, else let's bring it in!

ivov (dismissed):

ivov (approved):

PR #22625, opened 2025-12-02
chore(ai-builder): Telemetry updates for pinned data

mike12345567:
## Summary

This addresses the linear ticket: https://linear.app/n8n/issue/AI-1751/telemetry-add-workflow-id-etc-ids-to-some-existing-events

The following events have had some extra properties added to them: `user_clicked_pin_data_icon`, `ndv_data_pinning_success` and `user_opened_ndv_edit_state`

They have each had the workflow, instance, user and node ID added to them.

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/composables/usePinnedData.ts
```diff
@@ -215,6 +217,10 @@ export function usePinnedData(
 			data_size: stringSizeInBytes(data.value),
 			view: displayMode,
 			run_index: runIndex,
+			workflow_id: workflowsStore.workflowId,
+			user_id: usersStore.currentUserId ?? undefined,
+			instance_id: rootStore.instanceId,
```
mutdmour: why are we adding user_id and instance_id? Are not those added as traits in every telemetry call?

mutdmour: here https://github.com/n8n-io/n8n/blob/master/packages/cli/src/telemetry/index.ts#L227-L232

mike12345567: ah good question - I'll check in with Harshal to see why he asked for them to be added!

mutdmour: looks like user_id is not added by default. but instance_id is.
let's add `user_id` there so it's by default for all events?

packages/cli/src/telemetry/index.ts
```diff
@@ -225,6 +225,7 @@ export class Telemetry {
 		const updatedProperties = {
 			...properties,
 			instance_id: instanceId,
+			user_id: user_id ?? undefined,
```
mutdmour, line 228: can you add some tests for this? there's `telemetry.test.ts` file which already covers a lot of this component

packages/frontend/editor-ui/src/app/composables/usePinnedData.ts
```diff
@@ -215,6 +217,8 @@ export function usePinnedData(
 			data_size: stringSizeInBytes(data.value),
 			view: displayMode,
 			run_index: runIndex,
+			workflow_id: workflowsStore.workflowId,
```
mutdmour, line 218: optional - there's a test called `should trigger telemetry on successful data setting` maybe we can extend that to check the values sent are correct, not just that the event was fired

mutdmour (approved):

PR #22634, opened 2025-12-02
fix(editor): Update the value of updated at on data changes

nikhilkuria:
## Summary
https://www.loom.com/share/4cb89baaa6e845b098ee0947c196993b

currently, only table renames changed updatedat

https://linear.app/n8n/issue/ADO-4484/last-updated-and-created-date-always-the-same-on-data-table-cards

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/data-table/data-table.repository.ts
+++ b/packages/cli/src/modules/data-table/data-table.repository.ts
@@ -41,7 +41,7 @@ export class DataTableRepository extends Repository<DataTable> {
 		await withTransaction(this.manager, trx, async (em) => {
 			await em.update(DataTable, { id: dataTableId }, { updatedAt: new Date() });
 		}).catch((error) => {
-			this.logger.debug('Failed to update DataTable timestamp', { dataTableId, error });
+			this.logger.warn('Failed to update DataTable timestamp', { dataTableId, error });
 		});
 	}
 
--- a/packages/cli/src/modules/data-table/data-table.repository.ts
+++ b/packages/cli/src/modules/data-table/data-table.repository.ts
@@ -3,6 +3,7 @@ import {
 	type DataTableCreateColumnSchema,
 	type ListDataTableQueryDto,
 } from '@n8n/api-types';
+import { Logger } from '@n8n/backend-common';
 import { GlobalConfig } from '@n8n/config';
 import { Project, withTransaction } from '@n8n/db';
 import { Service } from '@n8n/di';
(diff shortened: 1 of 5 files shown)


packages/cli/src/modules/data-table/data-table.service.ts
```diff
@@ -171,7 +171,11 @@ export class DataTableService {
 	async addColumn(dataTableId: string, projectId: string, dto: AddDataTableColumnDto) {
```
nikhilkuria, line 171: Alternative I considered is having the `touchUpdatedAt` as part of `data-table-column-repository` 
The reason for having it here is for isolation of responsibilities.

packages/cli/src/modules/data-table/data-table.repository.ts
```diff
@@ -24,10 +25,26 @@ export class DataTableRepository extends Repository<DataTable> {
...
+	 * This is used to track when the table's content (rows/columns) has changed.
+	 *
+	 * Note: This method logs but does not throw errors to ensure that timestamp
+	 * update failures don't affect the primary data operations.
+	 */
+	async touchUpdatedAt(dataTableId: string, trx?: EntityManager) {
+		await withTransaction(this.manager, trx, async (em) => {
+			await em.update(DataTable, { id: dataTableId }, { updatedAt: new Date() });
+		}).catch((error) => {
+			this.logger.debug('Failed to update DataTable timestamp', { dataTableId, error });
```
scdekov: Should we log this with higher log level? I'm not sure what log levels we're logging eg for cloud instances but it sounds like this should be warning/error

packages/cli/src/modules/data-table/data-table.repository.ts
```diff
@@ -24,10 +25,26 @@ export class DataTableRepository extends Repository<DataTable> {
...
+	/**
+	 * Updates the updatedAt timestamp for a data table without modifying any other fields.
+	 * This is used to track when the table's content (rows/columns) has changed.
+	 *
+	 * Note: This method logs but does not throw errors to ensure that timestamp
+	 * update failures don't affect the primary data operations.
+	 */
+	async touchUpdatedAt(dataTableId: string, trx?: EntityManager) {
+		await withTransaction(this.manager, trx, async (em) => {
+			await em.update(DataTable, { id: dataTableId }, { updatedAt: new Date() });
```
scdekov, line 42: Interestingly when this fails while the updatedAt had value beforehand, the result is that updatedAt is set to null where I would expect it to stay with the latest value assuming that transaction will be reverted. Anyways it's probably not worth changing anything as the main operation is successful (at least on sqlite) 🤞

scdekov (dismissed):

scdekov (dismissed):

packages/cli/src/modules/data-table/data-table.repository.ts
nikhilkuria: How did you notice that? With a transaction, failure should mean the older value is retained.

packages/cli/src/modules/data-table/data-table.repository.ts
nikhilkuria: Good point, making this warn

packages/cli/src/modules/data-table/data-table.repository.ts
scdekov: I tried setting a random string to updatedAt there just by hardcoding it but it looks like the behavior may be specific to sqlite as it uses strings under the hood for dates. Did you actually tried failing the transaction? I was wondering if the catch needs to be within the withTransaction callback or the way you've added it and that's why I tried to break it.

scdekov (approved):

packages/cli/src/modules/data-table/data-table.repository.ts
nikhilkuria: I tried to break it like this
```
async touchUpdatedAt(dataTableId: string, trx?: EntityManager) {
		await withTransaction(this.manager, trx, async (em) => {
			throw Error('Test Error');
			await em.update(DataTable, { id: dataTableId }, { updatedAt: new Date() });
		}).catch((error) => {
			this.logger.warn('Failed to update DataTable timestamp', { dataTableId, error });
		});
	}
```

In this case, the earlier value was preserved.

PR #22643, opened 2025-12-03
feat(core): Block access to env in code and expressions by default

tomi:
## Summary

Changes the default value of `N8N_BLOCK_ENV_ACCESS_IN_NODE` to true. This means Code node and expressions don't have access to `process.env` anymore unless explicitly enabled by setting `N8N_BLOCK_ENV_ACCESS_IN_NODE=false`

https://linear.app/n8n/issue/CAT-691/block-env-access-from-code-node-by-default

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/Code/PythonSandbox.ts
+++ b/packages/nodes-base/nodes/Code/PythonSandbox.ts
@@ -11,7 +11,7 @@ type PythonSandboxContext = {
 
 type PyodideError = Error & { type: string };
 
-const envAccessBlocked = process.env.N8N_BLOCK_ENV_ACCESS_IN_NODE === 'true';
+const envAccessBlocked = process.env.N8N_BLOCK_ENV_ACCESS_IN_NODE !== 'false';
 
 export class PythonSandbox extends Sandbox {
 	private readonly context: PythonSandboxContext;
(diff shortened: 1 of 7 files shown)


packages/nodes-base/nodes/Code/PythonSandbox.ts
```diff
@@ -11,7 +11,7 @@ type PythonSandboxContext = {
 
 type PyodideError = Error & { type: string };
 
-const envAccessBlocked = process.env.N8N_BLOCK_ENV_ACCESS_IN_NODE === 'true';
+const envAccessBlocked = process.env.N8N_BLOCK_ENV_ACCESS_IN_NODE !== 'false';
```
ivov, line 14: One issue I see with this is that we're comparing against the env value, so it's incosistent with our normal config behavior where `1` is `true`, `0` is `false`, casing insensitive for `true` and `false`, etc.

ivov (approved):

packages/nodes-base/nodes/Code/PythonSandbox.ts
tomi: Good point. We should move it to the normal config like rest of the things. I created https://linear.app/n8n/issue/CAT-1863/move-n8n-block-env-access-in-node-to-config-package

PR #22213, opened 2025-11-24
fix(core): Prevent credentials saving if missing required properties

guillaumejacquart:
## Summary

This PR adds logic to prevent saving credentials which lacks mandatory parameters

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/PAY-4098/credentials-saved-successfully-without-mandatory-fields

## Review / Merge checklist

- [X] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [X] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #22645, opened 2025-12-03
fix(MCP Client Node): Make "Use Dynamic Client Registration" toggle not required

RomanDavydchuk:
## Summary

After a change in #22213, which prevents credentials from being saved if there are missing required fields, MCP OAuth2 credentials broke. This is because `useDynamicClientRegistration` field is `required` and has `default: true`. This only happens in this particular case, so if `default` is `false` or the field is not of type `boolean` it still works with a `default` value. Since it doesn't make much sense to have a toggle as `required` (and given we expect that it _can_ be turned off), I removed `required: true`

#### To test
1. Try connecting to any MCP server using OAuth2 with DCR, such as `https://mcp.notion.com/mcp`
2. Observe the error

https://linear.app/n8n/issue/NODE-4006/mcp-clients-oauth-is-not-working-with-gleancom#comment-ff76e915

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/credentials/McpOAuth2Api.credentials.ts
+++ b/packages/@n8n/nodes-langchain/credentials/McpOAuth2Api.credentials.ts
@@ -15,7 +15,6 @@ export class McpOAuth2Api implements ICredentialType {
 			name: 'useDynamicClientRegistration',
 			type: 'boolean',
 			default: true,
-			required: true,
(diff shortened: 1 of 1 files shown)


Joffcom (approved):

ShireenMissi (approved):
I tested locally and it works ✅

PR #22651, opened 2025-12-03
feat(core): Add more context window limits for our supported Chat models (no-changelog)

Cadiac:
## Summary

To prevent users from running into context window limits like

image

add more limits for our supported providers & add a safety margin of 95% to not meet the limits. 

Since on the screenshot we saw this happen with 211k / 200k on Opus 4.5 Haiku which already had the limit of 200k in place it does seem like the truncatenation doesn't somehow get applied here as expected 🤔 I'll take a look on the agent node and open a second PR about that if I find something, perhaps we don't truncate the tokens originating from tool call results..

## Review / Merge checklist

- [ ] Tests included. 


autologie (approved):
Thanks!

PR #22656, opened 2025-12-03
Fix Remove Duplicates node not showing new RSS items after first run

Mr-KRAMA:
Problem
New RSS items were marked as processed during the same run
So on next run they looked like duplicates and were removed

Root Cause
In deduplication helper ts
New item hashes were added to processed data inside the loop

Solution
Collect all new item hashes first
Add them to processed data after the loop is complete

Result
New items show correctly on every run
Duplicate items still filtered
No breaking changes

Files Updated
deduplication helper ts and related tests


Mr-KRAMA:
Branch updated. The fix and tests are complete. This PR is ready for review.

PR #22662, opened 2025-12-03
chore(core): Implement generic credential storage provider

afitzek:
## Summary

Implements generic credential storage provider with database-backed persistence for entity-specific credential data. The storage layer uses a composite primary key (credential_id, subject_id, resolver_id) with CASCADE DELETE foreign key constraints to ensure data integrity. Includes comprehensive integration tests verifying CRUD operations, upsert functionality, cascade deletes, and entry isolation.

closes https://linear.app/n8n/issue/PAY-4225/implement-generic-credential-storage-provider

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
+++ b/packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
@@ -1,0 +1,31 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const tableName = 'dynamic_credential_entry';
+
+export class AddDynamicCredentialEntryTable1764689388394 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(tableName)
+			.withColumns(
+				column('credential_id').varchar(16).primary.notNull,
+				column('subject_id').varchar(16).primary.notNull,
+				column('resolver_id').varchar(16).primary.notNull,
+				column('data').text.notNull,
+			)
+			.withTimestamps.withForeignKey('credential_id', {
+				tableName: 'credentials_entity',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withForeignKey('resolver_id', {
+				tableName: 'dynamic_credential_resolver',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withIndexOn(['credential_id'])
+			.withIndexOn(['resolver_id']);
+	}
+
+	async down({ schemaBuilder: { dropTable } }: MigrationContext) {
+		await dropTable(tableName);
+	}
+}
--- a/packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
+++ b/packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
@@ -21,7 +21,7 @@ export class AddDynamicCredentialEntryTable1764689388394 implements ReversibleMi
 				columnName: 'id',
 				onDelete: 'CASCADE',
 			})
-			.withIndexOn(['credential_id'])
+			.withIndexOn(['subject_id'])
 			.withIndexOn(['resolver_id']);
 	}
 
--- a/packages/@n8n/decorators/src/module/module.ts
+++ b/packages/@n8n/decorators/src/module/module.ts
@@ -16,13 +16,18 @@ export interface BaseEntity {
 	reload(): Promise<void>;
 }
 
-export interface TimestampedEntity {
+export interface TimestampedIdEntity {
 	id: string;
 	createdAt: Date;
 	updatedAt: Date;
 }
 
-export type EntityClass = new () => BaseEntity | TimestampedEntity;
+export interface TimestampedEntity {
+	createdAt: Date;
+	updatedAt: Date;
+}
+
+export type EntityClass = new () => BaseEntity | TimestampedIdEntity | TimestampedEntity;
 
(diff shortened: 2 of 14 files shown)


packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
```diff
@@ -0,0 +1,31 @@
...
+				tableName: 'credentials_entity',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withForeignKey('resolver_id', {
+				tableName: 'dynamic_credential_resolver',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withIndexOn(['credential_id'])
```
guillaumejacquart: I **think** postgres and sqlite can use the composite primary key as index when querying on a single field of the composite key, which would make the dedicated indexes redundant

packages/@n8n/decorators/src/module/module.ts
```diff
@@ -16,13 +16,18 @@ export interface BaseEntity {
...
 
-export interface TimestampedEntity {
+export interface TimestampedIdEntity {
 	id: string;
 	createdAt: Date;
 	updatedAt: Date;
 }
 
-export type EntityClass = new () => BaseEntity | TimestampedEntity;
+export interface TimestampedEntity {
```
guillaumejacquart, line 25: Not sure why you need that. BaseEntity already accepts not having id I think, and WithTimestamps extends BaseEntity

packages/cli/src/modules/dynamic-credentials.ee/database/entities/dynamic-credential-entry.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	@PrimaryColumn({
+		name: 'resolver_id',
+	})
+	resolverId: string;
+
+	@Column('text')
+	data: string;
+
+	@ManyToOne(() => CredentialsEntity)
```
guillaumejacquart: Does this work without setting the other way OnyToMany on the CredentialsEntity ?

guillaumejacquart (commented):
LOoks good ! some comments

packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
afitzek: Yes, already addressed 😄

packages/@n8n/decorators/src/module/module.ts
afitzek: Base entity is an interface defined here, that requires an id, I imagine because of the hasId() method.

packages/cli/src/modules/dynamic-credentials.ee/database/entities/dynamic-credential-entry.ts
afitzek: I think so, but I'll write a test to check 👍

afitzek: I added tests to filter and fetch the relations on the entity. Seems to work fine.

packages/@n8n/decorators/src/module/module.ts
guillaumejacquart: I don't see it requiring an id 🤔 
```typescript
export interface BaseEntity {
	hasId(): boolean;
	save(options?: unknown): Promise<this>;
	remove(options?: unknown): Promise<this>;
	softRemove(options?: unknown): Promise<this>;
	recover(options?: unknown): Promise<this>;
	reload(): Promise<void>;
}
```

packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
guillaumejacquart: Ah yes, the AI was faster :D 
Same goes for subject_id and resolver_id though, no ?

packages/@n8n/decorators/src/module/module.ts
afitzek: I checked the BaseEntity interface here is off I assume, because we define the BaseEntity here (https://github.com/n8n-io/n8n/blob/5f3ef83d9e10ee93a911f751e3a0838681f8ee45/packages/%40n8n/db/src/entities/abstract-entity.ts#L101) as empty class and then just mixin properties to it.

afitzek: So we need to change the interface, because we are not using the BaseEntity from typeorm in the module to define the entity, but a mixed in version of our own BaseEntity.

packages/@n8n/db/src/migrations/common/1764689388394-AddDynamicCredentialEntryTable.ts
afitzek: No, because it is a composite index, the DB can only use the first entry in the index as a solo index lookup. If we want to query only for subject_id or resolver_id the PK index does not help the DB.

```
[id1, id2, id3]
[1, 1, 1]
[1, 2, 1]
[1, 2, 2]
[2, 1, 1]
[2, 1, 2]
[2, 2, 2]
[2, 2, 2]
```

So for example this would be the index, it is sorted by id1 first, then by id2, then by id3. I can use this to quickly find all with id1 == 1, because these are grouped together. It is tricky to find all id2 == 1, since they are scattered around

guillaumejacquart (dismissed):
Thanks for taking the time to address the comments

guillaumejacquart (approved):

PR #22678, opened 2025-12-03
fix(editor): Chat UI feedback (no-changelog)

autologie:
## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
+++ b/packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
@@ -48,6 +48,7 @@ const systemPrompt = ref('');
 const selectedModel = ref<ChatHubBaseLLMModel | null>(null);
 const isSaving = ref(false);
 const isDeleting = ref(false);
+const isOpened = ref(false);
 const tools = ref<INode[]>([]);
 const nameInputRef = useTemplateRef('nameInput');
 
@@ -92,6 +93,10 @@ const canSelectTools = computed(
 	() => selectedAgent.value?.metadata.capabilities.functionCalling ?? false,
 );
 
+modalBus.value.once('opened', () => {
+	isOpened.value = true;
+});
+
 // If the agent doesn't support tools anymore, reset tools
 watch(
 	selectedAgent,
@@ -122,12 +127,11 @@ watch(
 );
 
 watch(
-	[isLoadingAgent, nameInputRef],
-	async ([isLoading, name]) => {
-		if (!isLoading) {
+	[isOpened, isLoadingAgent, nameInputRef],
+	async ([opened, isLoading, name]) => {
+		if (opened && !isLoading) {
 			// autofocus attribute doesn't work in modal
 			// https://github.com/element-plus/element-plus/issues/15250
-			await new Promise((r) => setTimeout(r, 0));
 			name?.focus();
 		}
 	},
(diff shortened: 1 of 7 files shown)


packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessage.vue
autologie, line 1: Showing copy button for code blocks is a bit tricky because it's rendered by markdown library. In this PR implemented via

1. Markdown lib plugin to collect code block contents and inserts action wrapper div
2. mouse event handler to show/hide actions
3. `<Teleport />` to render actions in the wrapper div in 1

packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
```diff
@@ -120,6 +121,19 @@ watch(
 	{ immediate: true },
 );
 
+watch(
+	[isLoadingAgent, nameInputRef],
+	async ([isLoading, name]) => {
+		if (!isLoading) {
+			// autofocus attribute doesn't work in modal
+			// https://github.com/element-plus/element-plus/issues/15250
+			await new Promise((r) => setTimeout(r, 0));
```
Cadiac: not sure if relevant here, but if you ever need to do something on modals after the modal is fully open & the animation is complete, there's `modalBus.on('opened'` https://github.com/n8n-io/n8n/blob/master/packages/frontend/editor-ui/src/app/components/WhatsNewModal.vue#L68.

that might be relevant if you need this to position a menu that opens or something correctly.

I'm fine with the setTimeout here too.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessage.vue
Cadiac: Based on my previous experiences with other markdown libraries I was expecting we'd be somehow able to tell the markdown rendered "render  blocks using this custom component", passing it a custom component that would do the copying logic etc.

But I guess that wasn't the case with vueMarkdownit, or could we have just created our own component with copying logic here and drawn that component instead of the pre + div, with the code to render as a prop to that custom component?

packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
autologie: Thanks, I could get rid of setTimeout using the event bus. [0de7ae7](https://github.com/n8n-io/n8n/pull/22678/commits/0de7ae76a4c381ed003c3880683561251600f659)

packages/frontend/editor-ui/src/features/ai/chatHub/components/ModelSelector.vue
```diff
@@ -100,7 +102,7 @@ const menu = computed(() => {
 			? []
 			: [...agents.value['custom-agent'].models, ...agents.value['n8n'].models].map((agent) => ({
 					id: stringifyModel(agent.model),
-					title: agent.name,
+					title: truncateBeforeLast(agent.name, MAX_AGENT_NAME_CHARS_MENU),
```
Cadiac: Image

These seem to get a different avatar color when the name is truncated 😅

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessage.vue
autologie: Yes some libraries has the API to override render function for each element type, but I couldn't find such with markdown-it

Cadiac: Then this is the way to go for us, nice workaround.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ModelSelector.vue
autologie: Great catch, fixed [90043ba](https://github.com/n8n-io/n8n/pull/22678/commits/90043ba337fcc622863eaa1e08435178446492fd)

Cadiac (approved):
🚀

PR #22687, opened 2025-12-03
feat(editor): Add hotkey for publish

nikhilkuria:
## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -773,6 +774,29 @@ async function onSaveWorkflow() {
 	}
 }
 
+async function onPublishWorkflow() {
+	const workflowIsArchived = workflowsStore.workflow.isArchived;
```
scdekov, line 778: Moving the handler here make sense conceptually but rn it breaks the loading state when we're saving before opening the modal (check the `autoSaveForPublish` prop in the old version). I guess the better option may be to move this loading into the store but if we want to merge this for the 2.0 release I'd suggest just moving it back to the WorkflowHeaderDraftPublishActions component unless there was a functional problem with it being there.

scdekov (changes requested):

packages/frontend/editor-ui/src/app/views/NodeView.vue
scdekov: That's how it works in v2

https://github.com/user-attachments/assets/c8c70ef4-aae7-4f54-b16c-d11c1b063a93

scdekov: And that's in this branch 

https://github.com/user-attachments/assets/f4f23a05-7d5b-42a2-a820-b051b4e53ef2

PR #22712, opened 2025-12-04
feat(core): Add resolvable fields to credential entity

guillaumejacquart:
## Summary

This PR adds fields on the CredentialEntity to support dynamic credential resolution:
- **resolverId**: varchar foreign key (nullable) linking to a credential resolver
- **isResolvable**: boolean to determine whether to try and resolve the credentials at runtime
- **resolvableAllowFallback**: boolean to determine whether credentials should fallback to static credential data if the resolver fails

### Design Decisions

**TypeORM Relationship**: We intentionally omit the `@ManyToOne` decorator because the linked entity is defined in the dynamic credentials module. TypeORM treats `resolverId` as a regular text field while the database handles foreign key constraints. This approach is acceptable since we don't need to load `credential.resolver` via TypeORM—the credential module can independently query the resolver repository with the resolver ID.

**Field Independence**: `isResolvable` can be `true` even when `resolverId` is `null`. This handles cases where the resolver is configured at a higher level in the hierarchy (workflow/project/instance) rather than on the individual credential. This also allows customers to "disable" the resolving mechanism temporarily (e.g., for debugging) while keeping the resolver ID reference.

**Fallback Behavior**: `resolvableAllowFallback` only falls back to static credential data, not to alternative resolvers. The fallback path for credentials goes from the most specific resolver (on the credential) to the least specific (on the workflow, project, or instance). The last possible option is to allow fallback to the static credential data. This supports edge cases like LLM credentials where some users might not have personal credentials and can fall back to a company key.

### Migration Safety

The migration includes `transaction = false` for SQLite to prevent [data loss from SQLite foreign key quirk](https://n8nio.slack.com/archives/C035KBDA917/p1764321309535179?thread_ts=1764258843.340829&cid=C035KBDA917) when dropping foreign keys.

https://linear.app/n8n/issue/PAY-4214/add-resolvable-fields-to-credential-entity

## Review / Merge checklist

- [X] Tests included.

### DIFF
--- a/packages/@n8n/db/src/entities/credentials-entity.ts
+++ b/packages/@n8n/db/src/entities/credentials-entity.ts
@@ -42,6 +42,25 @@ export class CredentialsEntity extends WithTimestampsAndStringId implements ICre
 	@Column({ default: false })
 	isGlobal: boolean;
 
+	/**
+	 * Whether the credential can be dynamically resolved by a resolver.
+	 */
+	@Column({ default: false })
+	isResolvable: boolean;
+
+	/**
+	 * Whether the credential resolver should allow falling back to static credentials
+	 * if dynamic resolution fails.
+	 */
+	@Column({ default: false })
+	resolvableAllowFallback: boolean;
+
+	/**
+	 * ID of the dynamic credential resolver associated with this credential.
+	 */
+	@Column({ type: 'varchar', nullable: true })
+	resolverId?: string;
+
 	toJSON() {
 		const { shared, ...rest } = this;
 		return rest;
--- a/packages/@n8n/db/src/entities/credentials-entity.ts
+++ b/packages/@n8n/db/src/entities/credentials-entity.ts
@@ -59,7 +59,7 @@ export class CredentialsEntity extends WithTimestampsAndStringId implements ICre
 	 * ID of the dynamic credential resolver associated with this credential.
 	 */
 	@Column({ type: 'varchar', nullable: true })
-	resolverId?: string;
+	resolverId: string | null;
 
 	toJSON() {
 		const { shared, ...rest } = this;
--- a/packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
+++ b/packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
@@ -1,0 +1,42 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const credentialsTableName = 'credentials_entity';
+const resolverTableName = 'dynamic_credential_resolver';
+const FOREIGN_KEY_NAME = 'credentials_entity_resolverId_foreign';
+
+export class AddResolvableFieldsToCredentials1765459448000 implements ReversibleMigration {
+	async up({ schemaBuilder: { addColumns, addForeignKey, column } }: MigrationContext) {
+		// Add isResolvable, resolvableAllowFallback, and resolverId columns to credentials_entity
+		await addColumns(credentialsTableName, [
+			column('isResolvable').bool.notNull.default(false),
+			column('resolvableAllowFallback').bool.notNull.default(false),
+			column('resolverId').varchar(16),
+		]);
+
+		// Add foreign key constraint
+		await addForeignKey(
+			credentialsTableName,
+			'resolverId',
+			[resolverTableName, 'id'],
+			FOREIGN_KEY_NAME,
+			'SET NULL',
+		);
+	}
+
+	async down({ schemaBuilder: { dropColumns, dropForeignKey } }: MigrationContext) {
+		// Drop foreign key constraint
+		await dropForeignKey(
+			credentialsTableName,
+			'resolverId',
+			[resolverTableName, 'id'],
+			FOREIGN_KEY_NAME,
+		);
+
+		// Drop columns from credentials_entity
+		await dropColumns(credentialsTableName, [
+			'isResolvable',
+			'resolvableAllowFallback',
+			'resolverId',
+		]);
+	}
+}
--- a/packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
+++ b/packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
@@ -6,14 +6,12 @@ const FOREIGN_KEY_NAME = 'credentials_entity_resolverId_foreign';
 
 export class AddResolvableFieldsToCredentials1765459448000 implements ReversibleMigration {
 	async up({ schemaBuilder: { addColumns, addForeignKey, column } }: MigrationContext) {
-		// Add isResolvable, resolvableAllowFallback, and resolverId columns to credentials_entity
 		await addColumns(credentialsTableName, [
 			column('isResolvable').bool.notNull.default(false),
 			column('resolvableAllowFallback').bool.notNull.default(false),
 			column('resolverId').varchar(16),
 		]);
 
-		// Add foreign key constraint
 		await addForeignKey(
 			credentialsTableName,
 			'resolverId',
@@ -24,7 +22,6 @@ export class AddResolvableFieldsToCredentials1765459448000 implements Reversible
 	}
 
 	async down({ schemaBuilder: { dropColumns, dropForeignKey } }: MigrationContext) {
-		// Drop foreign key constraint
 		await dropForeignKey(
 			credentialsTableName,
 			'resolverId',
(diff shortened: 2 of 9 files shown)


afitzek (dismissed):
Lgtm 🚀

BGZStephen (dismissed):

BGZStephen (dismissed):

packages/@n8n/db/src/entities/credentials-entity.ts
```diff
@@ -42,6 +42,25 @@ export class CredentialsEntity extends WithTimestampsAndStringId implements ICre
...
+	 * if dynamic resolution fails.
+	 */
+	@Column({ default: false })
+	resolvableAllowFallback: boolean;
+
+	/**
+	 * ID of the dynamic credential resolver associated with this credential.
+	 */
+	@Column({ type: 'varchar', nullable: true })
+	resolverId?: string;
```
ivov: Shouldn't this be `string | null`?

packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
ivov, line 1: None of the comments in this migration add value.

ivov, line 1: ⚠️  If this runs in sqlite, I expect [this quirk](https://n8nio.slack.com/archives/C035KBDA917/p1764321309535179?thread_ts=1764258843.340829&cid=C035KBDA917) applies and can result in data loss. Please double check this.

packages/cli/src/modules/mcp/__tests__/webhook-utils.test.ts
```diff
@@ -25,6 +25,8 @@ const mockCredentialsService = (
 				shared: [] as SharedCredentials[],
 				isManaged: false,
 				isGlobal: false,
+				isResolvable: false,
+				resolvableAllowFallback: false,
```
ivov, line 30: Should we mock `resolverId` as well?

packages/cli/test/integration/dynamic-credentials/credential-resolver.repository.test.ts
```diff
@@ -0,0 +1,112 @@
...
+			expect(credential.resolverId).toBeNull();
+		});
+
+		it('should set resolverId to null on resolver deletion (CASCADE SET NULL)', async () => {
+			const credential = await credentialsRepository.save(
+				credentialsRepository.create({
+					name: 'Linked',
+					type: 'oauth2',
+					data: '',
+					resolverId: resolver.id,
```
ivov, line 107: If we're passing a `resolverId` then `isResolvable` should be `true`? Or not necessarily?

packages/cli/src/modules/dynamic-credentials.ee/database/repositories/__tests__/credential-resolver.repository.integration.test.ts
```diff
@@ -0,0 +1,112 @@
...
+				credentialsRepository.create({
+					name: 'Linked',
+					type: 'oauth2',
+					data: '',
+					resolverId: resolver.id,
+				}),
+			);
+
+			await resolverRepository.delete(resolver.id);
+			const orphaned = await credentialsRepository.findOne({ where: { id: credential.id } });
```
ivov: Or `findOneBy`

packages/@n8n/db/src/entities/credentials-entity.ts
```diff
@@ -42,6 +42,25 @@ export class CredentialsEntity extends WithTimestampsAndStringId implements ICre
...
+	/**
+	 * Whether the credential resolver should allow falling back to static credentials
+	 * if dynamic resolution fails.
+	 */
+	@Column({ default: false })
+	resolvableAllowFallback: boolean;
+
+	/**
+	 * ID of the dynamic credential resolver associated with this credential.
+	 */
```
ivov, line 60: Without `@ManyToOne` decorator I expect `typeorm` can't resolve the relationship so we can't load `credential.resolver`, so the DB schema and typeorm drift apart. 

If that's true, it makes me wonder if dynamic credentials should be an optional module or not.

packages/@n8n/db/src/entities/credentials-entity.ts
```diff
@@ -42,6 +42,25 @@ export class CredentialsEntity extends WithTimestampsAndStringId implements ICre
 	@Column({ default: false })
 	isGlobal: boolean;
 
+	/**
+	 * Whether the credential can be dynamically resolved by a resolver.
+	 */
+	@Column({ default: false })
+	isResolvable: boolean;
```
ivov, line 49: Not familiar with the details of this project, can we rely on `resolverId` being `null` or not `null` for this?

packages/@n8n/db/src/entities/credentials-entity.ts
```diff
@@ -42,6 +42,25 @@ export class CredentialsEntity extends WithTimestampsAndStringId implements ICre
...
+
+	/**
+	 * Whether the credential resolver should allow falling back to static credentials
+	 * if dynamic resolution fails.
+	 */
+	@Column({ default: false })
+	resolvableAllowFallback: boolean;
```
ivov, line 56: So we always fall back to static creds and never e.g. to another resolver?

packages/@n8n/db/src/entities/credentials-entity.ts
guillaumejacquart: Indeed, thanks

packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
guillaumejacquart: true dat

packages/cli/test/integration/dynamic-credentials/credential-resolver.repository.test.ts
guillaumejacquart: In theory and most use cases yes, but with the full picture in mind, it's possible that at some point a customer sets a credential to use a resolver, but "disable" the resolving mechanism (maybe in case of bug)

packages/@n8n/db/src/entities/credentials-entity.ts
guillaumejacquart: actually no, because the resolverId can be null, and isResolvable true. It's when the resolver has been setup for the workflow instead of the credentials

packages/@n8n/db/src/entities/credentials-entity.ts
guillaumejacquart: Yes that's true. we thought it would be fine to have that, as we're not using typeorm for migration generation. But I see your point 🤔 
I think it could be possible to extends the credential entity from the module itself (no sure, have to test). Not sure if it makes sense either though

packages/@n8n/db/src/entities/credentials-entity.ts
guillaumejacquart: That's how we've been designing it yes. I don't think there is a use case for a fallback to another resolver, as it would mean we're not sure how the extracted identity token would be resolved or could be resolved in multiple ways

packages/@n8n/db/src/entities/credentials-entity.ts
guillaumejacquart: Otherwise I think it's ok that TypeORM is not "aware" of the relationship. the field can be considered as a normal string field. The module owns the relationship

packages/@n8n/db/src/entities/credentials-entity.ts
afitzek: This would be the last fallback to the credential data, the fallback path for credentials goes from the most specific resolver (on the credential) to the least specific (on the workflow), maybe later on the project or instance as well. The last possible option is to allow fallback to the static credential data. There are edge cases like credentials for LLMs, that can use the user specific ones, but some users might not have them, so they can use the company key to access the LLM for example.

packages/@n8n/db/src/entities/credentials-entity.ts
afitzek: Yes we cannot load the resolver based on the credential.resolver, via TypeORM, but this is not necessary either. I would expect typeorm to just treat this as a text, it would be the DBs responsibility to set it to null in the FK constraint. Usually the credential entity is already fetched from the DB, when we need to determine if we want the specific resolver or a different one. So at this point, we would directly query the resolver repository with the resolver id stored on this credential.

packages/@n8n/db/src/migrations/common/1765459448000-AddResolvableFieldsToCredentials.ts
guillaumejacquart: Woooo, that's a big one. Thanks ! I added the transaction = false for sqlite and testing locally, not linked tables deleted 👍

afitzek (dismissed):
🚀

BGZStephen (dismissed):

RicardoE105 (approved):
Do not have a lot of context, but code looks good and ran the migration locally and it worked fine.

PR #22715, opened 2025-12-04
fix(editor): Don't create duplicate placeholder nodes on agent failure

schrothbn:
## Summary
Fixes an issue where AI Agent nodes appeared duplicated in the execution logs view when the agent execution finished with an error.

When the full AI Agent execution fails or doesn't complete, the agent node itself is NOT in the execution's `runData`, but its child nodes (LLM, tools, etc.) ARE present. Each child node independently tried to insert the parent AI Agent as a placeholder node, resulting in the agent appearing multiple times in the logs tree.

- Added deduplication logic in `createLogTreeRec` to track and filter duplicate placeholder entries (nodes without actual task data)
- Updated `mergeStartData` to better handle duplicate runs by checking for exact matches (same executionIndex AND startTime) to prevent duplicates from AI agent cases
- Added comprehensive test coverage for the duplication scenario

## Review / Merge checklist

- [ ] Tests included. 


mike12345567 (approved):
LGTM!

PR #22725, opened 2025-12-04
fix(core): Address chat feedback items (no-changelog)

Cadiac:
## Summary

Address some chat feedback items:
- Place configure credentials button at the bottom
- Disable new chat button at the collapsed menu
- Add a notice on chat trigger
- Add current time on base llm chats system prompt

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
+++ b/packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
@@ -468,14 +468,21 @@ export class ChatHubWorkflowService {
 		};
 	}
 
+	getSystemMessageMetadata(timeZone: string) {
+		const now = DateTime.now().setZone(timeZone).toISO({
+			includeOffset: true,
+		});
+
+		return `The user's current local date and time is: ${now} (timezone: ${timeZone}).
+When you need to reference “now”, use this date and time.`;
+	}
+
 	private getBaseSystemMessage(timeZone: string) {
 		const now = DateTime.now().setZone(timeZone).toISO({
 			includeOffset: true,
 		});
 
-		return `You are a helpful assistant.
-The user's current local date and time is: ${now} (timezone: ${timeZone}).
-When you need to reference “now”, use this date and time.`;
+		return 'You are a helpful assistant.\n' + this.getSystemMessageMetadata(timeZone);
 	}
 
 	private buildToolsAgentNode(
--- a/packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
+++ b/packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
@@ -1,5 +1,6 @@
 import { ChatHubConversationModel, ChatSessionId, type ChatHubInputModality } from '@n8n/api-types';
 import { Logger } from '@n8n/backend-common';
+import { GlobalConfig } from '@n8n/config';
 import {
 	SharedWorkflow,
 	SharedWorkflowRepository,
@@ -9,6 +10,7 @@ import {
 } from '@n8n/db';
 import { Service } from '@n8n/di';
 import { EntityManager } from '@n8n/typeorm';
+import { DateTime } from 'luxon';
 import {
 	AGENT_LANGCHAIN_NODE_TYPE,
 	CHAT_TRIGGER_NODE_TYPE,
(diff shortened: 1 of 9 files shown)


packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -463,6 +466,17 @@ export class ChatHubWorkflowService {
 		};
 	}
 
+	private getBaseSystemMessage() {
+		const timeZone = this.globalConfig.generic.timezone;
```
autologie: Minor, but I think this should be taken from the browser instead of the instance global config

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -474,7 +488,7 @@ export class ChatHubWorkflowService {
 						model.provider !== 'n8n' && model.provider !== 'custom-agent'
 							? getMaxContextWindowTokens(model.provider, model.model)
 							: undefined,
-					systemMessage,
+					systemMessage: systemMessage ?? this.getBaseSystemMessage(),
```
autologie: For custom agents to be able to recognize current time in the future... maybe expression support? 😱

Cadiac: maybe, we could make that field an expression but would that open a window for some security holes, as these user's can't really execute workflows 🤔

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatConversationHeader.vue
```diff
@@ -44,6 +44,10 @@ const showOpenWorkflow = computed(() => {
 	);
 });
 
+const isNewSession = computed(() => {
```
autologie: I'd pass this down from ChatView

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: 🤷 perhaps, but would we send the user's time on every chat request or make it configurable by user or..?

Cadiac: and I do agree that this isn't maybe expected that this is read from server and not from the client, but at least I made it use the n8n's concept of timezone (defined by `GENERIC_TIMEZONE` env variable)

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatConversationHeader.vue
Cadiac: sure

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
autologie: > send the user's time on every chat request

Maybe out of scope but I think this is most effortless as well as intuitive for the user

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
autologie: Good point about the security. I guess a simple solution is just concatenate user-provided instructions and time info.

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: Changed the frontend to send the timezone to backend, and use that on both custom agents & base LLMs prompt.

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: I made a solution that concatenates the user's timezone to the system message for custom agents too - perhaps we would like to include some metadata of the user / n8n instance there too but didn't do that yet.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatConversationHeader.vue
Cadiac: Fixed in [e701ce7](https://github.com/n8n-io/n8n/pull/22725/commits/e701ce715119cdb5c3fe952b0b20104065f27ab4)

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
autologie: Thanks, did a quick test locally and worked well

autologie (approved):
Thank you for addressing comments!

PR #22737, opened 2025-12-04
refactor(core): Mark native Python as stable

ivov:
https://linear.app/n8n/issue/CAT-1119


tomi (approved):
Should we include this in `v2` (2.0.0) alraedy?

PR #22747, opened 2025-12-04
test: Enable orchestration for e2e tests

seemewalkin:
## Summary

- Switch to `pwc-p` orchestration with batch size 5                                                                                                                                                                 
- Disable `fullyParallel` for container tests (improves dashboard accuracy with workers=1)                                                                                                                             

https://linear.app/n8n/issue/CAT-1699/add-orchestration-setup-for-playwright-tests-for-faster-execution-on

## Review / Merge checklist

- [ ] Tests included. 


packages/testing/playwright/currents.config.ts
```diff
@@ -3,6 +3,14 @@ import type { CurrentsConfig } from '@currents/playwright';
 const config: CurrentsConfig = {
 	recordKey: process.env.CURRENTS_RECORD_KEY ?? '',
 	projectId: process.env.CURRENTS_PROJECT_ID ?? 'I0yzoc',
+	ciBuildId:
```
shortstacked: Isn't this automatically pulled when running in GH anyway?

.github/workflows/playwright-test-reusable.yml
```diff
@@ -104,12 +103,19 @@ jobs:
...
+          else
+            ${{ inputs.test-command }} \
+              --shard=${{ matrix.shard }}/${{ strategy.job-total }} \
+              --workers=${{ env.PLAYWRIGHT_WORKERS }}
+          fi
         env:
+          RESET_E2E_DB: true
           N8N_DOCKER_IMAGE: ${{ inputs.test-mode == 'docker-build' && 'n8nio/n8n:local' || inputs.docker-image }}
           CURRENTS_RECORD_KEY: ${{ secrets.CURRENTS_RECORD_KEY }}
+          CURRENTS_PROJECT_ID: I0yzoc
```
shortstacked: Same as below I don't think we need these right?

packages/testing/playwright/playwright-projects.ts
```diff
@@ -64,7 +64,7 @@ export function getProjects(): Project[] {
 					testDir: './tests/e2e',
 					grepInvert: new RegExp(grepInvertPatterns.join('|')),
 					timeout: name === 'standard' ? 60000 : 180000, // 60 seconds for standard container test, 180 for containers to allow startup etc
-					fullyParallel: true,
+					fullyParallel: false,
```
shortstacked: We should make this based off the worker count so it doesn't interfere with local runs etc.
e.g if worker count is 1, fullyParallel = false

packages/testing/playwright/playwright.config.ts
```diff
@@ -94,6 +95,7 @@ export default defineConfig<CurrentsFixtures, CurrentsWorkerFixtures>({
 				['junit', { outputFile: process.env.PLAYWRIGHT_JUNIT_OUTPUT_NAME ?? 'results.xml' }],
 				['html', { open: 'never' }],
 				['json', { outputFile: 'test-results.json' }],
+				...(!process.env.CURRENTS_RECORD_KEY ? [['github'] as const] : []),
```
shortstacked: We shouldn't need/add the Github reporter back since it was causing noise with it's annotations, is it needed for orchestration/currents?

packages/testing/playwright/currents.config.ts
shortstacked: https://docs.currents.dev/guides/ci-build-id

.github/workflows/playwright-test-reusable.yml
```diff
@@ -104,12 +103,19 @@ jobs:
 
       - name: Run Tests
         run: |
-          ${{ inputs.test-command }} \
-            --shard=${{ matrix.shard }}/${{ strategy.job-total }} \
-            --workers=${{ env.PLAYWRIGHT_WORKERS }}
+          if echo "${{ inputs.test-command }}" | grep -q "pwc-p"; then
+            ${{ inputs.test-command }}
```
shortstacked: Do we need to remove the shard/workers here. Will orchestration break if we leave the test command as is with the shard in it, or will it just ignore it? That way we could reuse the same command without changing this part

packages/testing/playwright/currents.config.ts
seemewalkin: https://github.com/n8n-io/n8n/actions/runs/20141520929/job/57810225518?pr=22747#step:8:27 unfortunately fails without, remember facing that on initial tries. will explicitly use github_run_id though!

packages/testing/playwright/playwright-projects.ts
seemewalkin: or this option? https://github.com/n8n-io/n8n/pull/22747/files#diff-aef2c13acdeca77c6adbad9648e66222dcc3c0668d88be8943028afeff3d60d6R37 if no currents key then we can run parallel (both local/community)? should also work 🙏

packages/testing/playwright/playwright.config.ts
seemewalkin: was thinking to add that solely for community PRs, we don't need that/nor currents

packages/testing/playwright/playwright-projects.ts
```diff
@@ -32,6 +32,10 @@ const CONTAINER_CONFIGS: Array<{ name: string; config: N8NConfig }> = [
 	{ name: 'multi-main', config: { queueMode: { mains: 2, workers: 1 } } },
 ];
 
+// Enable fullyParallel for standard Playwright runs (community PRs, local)
+// Disable for Currents orchestration
+const enableParallelism = !process.env.CURRENTS_RECORD_KEY;
```
shortstacked: Can we do this on the CURRENTS_BATCH_SIZE instead? The coverage tests use a Currents Record Key but they run with 4 workers/parallel

.github/workflows/playwright-test-reusable.yml
seemewalkin: unfortunately https://github.com/n8n-io/n8n/actions/runs/20164603765/job/57884945124?pr=22747 --shard is being processed by Playwright before Currents orchestration, prefilters the tests and then nothing runs. will have to bring the if condition. (they briefly mention it here https://docs.currents.dev/guides/ci-optimization/playwright-orchestration).

shortstacked (dismissed):

shortstacked (approved):
Looks great!

PR #22749, opened 2025-12-04
fix(core): Hide migration rule issues not relevant to cloud

guillaumejacquart:
## Summary

This PR exclude the task runner docker image rule and task runner settings detection from cloud environment, because cloud deployment handles the docker image.
Also fixes the SettingsFilePermissionsRule to check if the user explicitly set the related env variable, and thus is aware of the impact.

https://linear.app/n8n/issue/PAY-4265/hide-migration-rule-issues-not-relevant-to-cloud

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/breaking-changes/rules/v2/settings-file-permissions.rule.ts
+++ b/packages/cli/src/modules/breaking-changes/rules/v2/settings-file-permissions.rule.ts
@@ -1,4 +1,4 @@
-import { InstanceSettingsConfig } from '@n8n/config';
+import { GlobalConfig } from '@n8n/config';
 import { Service } from '@n8n/di';
 
 import type {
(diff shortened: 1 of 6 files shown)


packages/cli/src/modules/breaking-changes/rules/v2/settings-file-permissions.rule.ts
```diff
@@ -28,9 +28,18 @@ export class SettingsFilePermissionsRule implements IBreakingChangeInstanceRule
...
+			return {
+				isAffected: false,
+				instanceIssues: [],
+				recommendations: [],
+			};
+		}
+
+		// If N8N_ENFORCE_SETTINGS_FILE_PERMISSIONS is explicitly set to any value, users are not affected
+		// because they've already handled the configuration and are aware of this setting.
+		if (process.env.N8N_ENFORCE_SETTINGS_FILE_PERMISSIONS) {
```
ivov, line 42: Interesting case where our own config is useless to detect this.

ivov (approved):

PR #22763, opened 2025-12-04
fix: Don't show unavailable models in model selector (no-changelog)

autologie:
## Summary

This PR introduces `available` field in model metadata to filter out models that cannot be used in the chat.

The `available` flag is set to `false` for those which failed to respond in an experiment and the returned error seemed permanent, such as
- Model not found
- Multi-turn conversations are not supported
- Invalid request

but not for errors such as
- Need to accept terms of use
- Need to upgrade

because they may yield different result depending on the API key used.

## Review / Merge checklist

- [ ] Tests included. 


Cadiac (approved):
👍

PR #22767, opened 2025-12-04
fix(core): Only resolve the filepath once

mfsiega:
## Summary

Make sure we only resolve filepaths once, so we know that when we read them they haven't been changed to link somewhere else.

## Review / Merge checklist

- [ ] Tests included. 


packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -74,34 +76,59 @@ export const getFileSystemHelperFunctions = (node: INode): FileSystemHelperFunct
 		}
 
 		try {
-			await fsAccess(filePath);
+			await fsAccess(resolvedFilePath);
```
mfsiega, line 79: Just a note: the NodeJS [docs](https://nodejs.org/api/fs.html#fspromisesaccesspath-mode) say: `Using fsPromises.access() to check for the accessibility of a file before calling fsPromises.open() is not recommended.`. Rather you should just try to open the file, and handle any potential failure. We should think about fixing this, potentially in this PR; but it's a bit orthogonal to the issue at hand.

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -35,18 +36,19 @@ const getAllowedPaths = () => {
 	return allowedPaths;
 };
 
-export async function isFilePathBlocked(filePath: string): Promise<boolean> {
-	const allowedPaths = getAllowedPaths();
-	let resolvedFilePath = '';
+const resolvePath = async (path: PathLike): Promise<ResolvedFilePath> => {
```
tomi: nit: Rest of the functions in this file have been defined with `function`. Maybe we could keep it consistent?

packages/workflow/src/interfaces.ts
```diff
@@ -694,12 +694,19 @@ export interface BaseHelperFunctions {
 	returnJsonArray(jsonData: IDataObject | IDataObject[]): INodeExecutionData[];
 }
 
+const __brand = Symbol('brand');
```
tomi: Or:
```suggestion
const __brand = Symbol('resolvedFilePath');
```

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -74,34 +76,59 @@ export const getFileSystemHelperFunctions = (node: INode): FileSystemHelperFunct
...
+		});
+
+		return await new Promise<ReturnType<typeof createReadStream>>((resolve, reject) => {
+			stream.once('error', (error) => {
+				if ((error as NodeJS.ErrnoException).code === 'ELOOP') {
+					reject(
+						new NodeOperationError(
+							node,
+							`The file "${String(resolvedFilePath)}" could not be opened.`,
+							{ level: 'warning' },
```
tomi: Could we attach the original error as `cause`?

packages/workflow/src/interfaces.ts
```diff
@@ -694,12 +694,19 @@ export interface BaseHelperFunctions {
...
+};
+
 export interface FileSystemHelperFunctions {
-	isFilePathBlocked(filePath: string): Promise<boolean>;
-	createReadStream(path: PathLike): Promise<Readable>;
+	resolvePath(path: PathLike): Promise<ResolvedFilePath>;
+	isFilePathBlocked(filePath: ResolvedFilePath): boolean;
+	createReadStream(filePath: ResolvedFilePath): Promise<Readable>;
 	getStoragePath(): string;
 	writeContentToFile(
```
tomi, line 717: Could we add docs to all these functions with mention that "Use `resolvePath` to resolve the path"?

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -74,34 +76,59 @@ export const getFileSystemHelperFunctions = (node: INode): FileSystemHelperFunct
...
+			throw new NodeOperationError(
+				node,
+				`The file "${String(resolvedFilePath)}" is not writable.`,
+				{
+					level: 'warning',
+				},
+			);
 		}
-		return await fsWriteFile(filePath, content, { encoding: 'binary', flag });
+		return await fsWriteFile(resolvedFilePath, content, { encoding: 'binary', flag });
```
tomi: Doesn't this still have the issue? Can we make this so that `fsWriteFile` doesn't follow symlinks?

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -35,18 +36,19 @@ const getAllowedPaths = () => {
 	return allowedPaths;
 };
 
-export async function isFilePathBlocked(filePath: string): Promise<boolean> {
-	const allowedPaths = getAllowedPaths();
-	let resolvedFilePath = '';
+const resolvePath = async (path: PathLike): Promise<ResolvedFilePath> => {
 	try {
-		resolvedFilePath = await fsRealpath(filePath);
+		return (await fsRealpath(path)) as ResolvedFilePath; // apply brand, since we know it's resolved now
```
Cadiac, line 41: Branded types seem like a smart way to enforce this 👍

Cadiac (dismissed):
Seems like a smart way to solve this 👍

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
mfsiega: Done.

packages/workflow/src/interfaces.ts
mfsiega: Done.

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
mfsiega: Done. (I think this is what you meant?)

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
mfsiega: Good call! I think I'm handling the flags correctly here.

Cadiac (dismissed):

tomi (dismissed):
Love the usage of branded types 🫶 🚀

packages/workflow/src/interfaces.ts
mfsiega: Done.

Cadiac (approved):

PR #22769, opened 2025-12-04
fix(editor): mcp design tweaks (no-changelog)

RobTF9:
## Summary
Made some design adjustments based on the notion feedback table

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/mcpAccess/components/header/connectPopover/MCPAccessTokenPopoverTab.vue
```diff
@@ -149,6 +150,9 @@ onMounted(async () => {
 					</N8nTooltip>
 				</template>
 			</ConnectionParameter>
+			<N8nNotice>
```
MiloradFilipovic: I think we should show the notice only when key can be copied:
```suggestion
			<N8nNotice v-if="!isKeyRedacted">
```

But with this we don't know how to size the loading skeletons (it's not certain if the notice will be there after loading), so we'll have some layout shifts.
It's really a nitpick but it's quite nice how nothing moves in the current version.
Alternatively, we could just show warning icon next to the access token with a tooltip (similar to parameter input with issues). We have some horizontal real-estate to do this since the `rotate` button is not visible when key is copiable.

MiloradFilipovic (approved):
Looks good, thanks for addressing these. Left a minor comment about the notice. Let me know what you think.

MiloradFilipovic (approved):
👌🚢

PR #22774, opened 2025-12-04
fix(editor): Bind color-scheme to app theme setting

Tuukkaa:
## Summary
Fixes `light-dark()` CSS function resolving based on OS preference instead of n8n's theme setting.

**Problem**
When a user sets n8n to "light" theme while their OS is in dark mode (or vice versa), canvas elements using `light-dark()` would display with the wrong colors because `color-scheme: light dark` on `:root` defers to OS preference.

**Solution**
Bind `color-scheme` explicitly to the `data-theme` attribute on `body`, so `light-dark()` always resolves to match the app's actual theme.

https://linear.app/n8n/issue/AI-1774/fix-light-dark-css-function-not-respecting-app-theme-setting

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/@n8n/design-system/src/css/_tokens.scss
+++ b/packages/frontend/@n8n/design-system/src/css/_tokens.scss
@@ -709,8 +709,6 @@
 	@include theme;
 }
 
-/* Bind color-scheme to n8n's theme setting so light-dark() CSS function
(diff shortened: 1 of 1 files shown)


tomi (approved):
🚀

PR #22805, opened 2025-12-05
feat: Include active version id in `workflowActivated` broadcast event (no-changelog)

dariacodes:
## Summary

- Add `activeVersionId` to broadcasted workflow activated event and update the state correctly
    - update checksum in case there are no ui changes so that user can proceed with other actions straightaway

Closes [ADO-4501](https://linear.app/n8n/issue/ADO-4501/feature-queue-mode-include-active-version-id-in-workflowactivated)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/active-workflow-manager.ts
+++ b/packages/cli/src/active-workflow-manager.ts
@@ -565,10 +565,24 @@ export class ActiveWorkflowManager {
 	) {
 		const added = { webhooks: false, triggersAndPollers: false };
 
+		const dbWorkflow = existingWorkflow ?? (await this.workflowRepository.findById(workflowId));
+
+		if (!dbWorkflow) {
+			throw new WorkflowActivationError(`Failed to find workflow with ID "${workflowId}"`, {
+				level: 'warning',
+			});
+		}
+
 		if (this.instanceSettings.isMultiMain && shouldPublish) {
+			if (!dbWorkflow?.activeVersionId) {
+				throw new UnexpectedError('Active version ID not found for workflow', {
+					extra: { workflowId },
+				});
+			}
+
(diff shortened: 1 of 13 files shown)


packages/cli/src/active-workflow-manager.ts
```diff
@@ -565,10 +565,24 @@ export class ActiveWorkflowManager {
...
+		const dbWorkflow = existingWorkflow ?? (await this.workflowRepository.findById(workflowId));
+
+		if (!dbWorkflow) {
+			throw new WorkflowActivationError(`Failed to find workflow with ID "${workflowId}"`, {
+				level: 'warning',
+			});
+		}
+
 		if (this.instanceSettings.isMultiMain && shouldPublish) {
+			if (!dbWorkflow?.activeVersionId) {
```
scdekov, line 577: Does adding a workflow without publishing make sense anymore? Maybe something to check for a follow up cleanup

scdekov (approved):

packages/cli/src/active-workflow-manager.ts
dariacodes: Could you elaborate? This check is just a sanity one in case some code called this on an inactive workflow (it should never happen, but if it ever does, we need to shout about it). Before it was also impossible to call this on inactive workflows, because by this moment they should be already updated in the db

scdekov: I was referring to the `shouldPublish` parameter. It sounds like it should always be true if we're adding a wf here

dariacodes: Ah, no, it's about broadcasting events, not our publish

PR #22807, opened 2025-12-05
chore(ai-builder): Use isAIMessage for LangGraph v1 compatibility (no-changelog)

OlegIvaniv:
## Summary
- Replace `instanceof AIMessage` with `isAIMessage()` function in legacy workflow's `shouldContinue` edge
- Fixes Langsmith evaluation failures after LangGraph v1 upgrade


mike12345567 (approved):
Thanks for this!

PR #22822, opened 2025-12-05
fix(editor): Improve chat error handling (no-changelog)

autologie:
## Summary

Tested sending many attachment files. Server responded with status 413, which is expected, but the chat UI wasn't able to present any feedback because `streamRequest` function didn't call `onError` callback function. Also adding more user-friendly messages.

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -411,7 +411,18 @@
 	"chatHub.message.actions.executionId": "Execution ID",
 	"chatHub.message.edit.cancel": "Cancel",
 	"chatHub.message.edit.send": "Send",
-	"chatHub.message.error.unknown": "Error: Unknown error occurred",
+	"chatHub.message.error.unknown": "Something went wrong. Please try again.",
+	"chatHub.error.payloadTooLarge": "Your message is too large. Try using shorter text or fewer files.",
+	"chatHub.error.badRequest": "Something went wrong with your message. Please try again.",
+	"chatHub.error.forbidden": "You don't have permission to do this. Please check your settings.",
+	"chatHub.error.serverError": "We're having trouble right now. Please try again in a moment.",
```
mutdmour: Would prefer error messages that are less personal, which is generally how the errors are in n8n

Avoid using `your` or `we`. `Something went wrong.` instead of `Something went wrong with your message`. `Could not reach service` instead of `We're having trouble`

autologie: I did so to make them more understandable to chat users, but you're right that they don't match the tone of the rest. Updated https://github.com/n8n-io/n8n/pull/22822/commits/e3ead2cd07a25847e6f43eb8b6705a411d165aa1

mutdmour (approved):

Keepstacking22:
Honestly I don't have any knowledge of what I'm supposed to do with any of
this I'm so confused with all of it I get so over whelmed and jus sign off
that's where I'm at with this so if anyone can help me get this done that
wohld be a blessing from God

Dawn Mendez

PR #22837, opened 2025-12-05
feat(core): Add 'Project Chat User' role for chat only access use case (no-changelog)

Cadiac:
## Summary

image

Add new `Project Chat User` role and new workflow permission `workflow:chat` that this role (and other stronger roles) grant.

This role is now the new acecss requirement to chatting with n8n workflows on Chat, as the previous approach of requiring `workflow:read` for chat access was too broad, as it also allows reading all workflow's executions which could reveal data from other users, which feels unexpected / security issue in "Chat only users" use case.

We should add support for this new permission on custom roles, but perhaps lets do that as a followup?

https://n8nio.slack.com/archives/C09JFRUU7GC/p1764940531014169

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/permissions/src/roles/all-roles.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
konstantintieber, line 1: What naming in this file requires disabling this eslint rule?

konstantintieber (commented):
PR looks good to me. Just wondering about one eslint rule before approving.

packages/@n8n/permissions/src/roles/all-roles.ts
Cadiac: It's the `eslint[@typescript-eslint/naming-convention](https://typescript-eslint.io/rules/naming-convention)` rule we have on warning level disliking our scope names, and if one has warnings visible our codebase is pretty yellow - I've been trying to either fix or ignore issues as I go...

to be honest that's a rule we should probably just disable as we appear to be breaking it in many places.

image

konstantintieber (approved):
Nice improvement to only expose triggering the execution of a workflow from chat hub, without any other access to the workflow.

packages/frontend/editor-ui/src/app/stores/roles.store.ts
```diff
@@ -1,27 +1,42 @@
...
+	const settingsStore = useSettingsStore();
 
 	const roles = ref<AllRolesMap>({
 		global: [],
 		project: [],
 		credential: [],
 		workflow: [],
 	});
-	const projectRoleOrder = ref<string[]>(['project:viewer', 'project:editor', 'project:admin']);
+	const projectRoleOrder = ref<string[]>([
```
autologie, line 24: The original ordering was narrower-first, so we should probably preserve the pattern?

packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
```diff
@@ -408,7 +408,7 @@ export class ChatTrigger extends Node {
 			},
 			{
 				displayName:
-					'Your n8n users will be able to use this agent in <a href="/home/chat/" target="_blank">Chat</a> once this workflow is published. Make sure to share this workflow with at least viewer access to all users who should use it. Currently, only streaming response mode is supported.',
+					'Your n8n users will be able to use this agent in <a href="/home/chat/" target="_blank">Chat</a> once this workflow is published. Make sure to share this workflow with at least \'workflow:execute-chat\' access to all users who should use it. Currently, only streaming response mode is supported.',
```
autologie, line 411: I think permission name is not widely used in user-facing context. "Chat user" could be more straightforward

packages/frontend/editor-ui/src/app/stores/roles.store.ts
Cadiac: I'm not sure - I felt like chat user would be kind of a special case anyways, with viewer probably being the more common option. I did consider putting chat user up there first as the most narrow role, but then decided against it 🤔

packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
Cadiac: perhaps, but as custom roles are now exposing those permission names I made it visible here.
but I might change this as our less technical users surely won't have any idea of the actual permission name.

packages/frontend/editor-ui/src/app/stores/roles.store.ts
autologie: Most intuitive would be to make the list upside down, but that would be a bigger discussion

PR #22860, opened 2025-12-05
fix(core): Update migration that activate workflows with executeWorkflowTrigger

RicardoE105:
## Summary

Fixes the `ActivateExecuteWorkflowTriggerWorkflows` migration to properly handle all valid Execute Workflow Trigger configurations, including:
- Version 1 nodes with empty parameters
- Version 1.1 nodes with missing `type` fields in workflowInputs (which default to "string")

## Root cause

The migration logic was too strict in validating Execute Workflow Trigger nodes:

1. **Version 1 nodes**: These nodes have no `inputSource` parameter and use empty `parameters: {}`. The migration was not recognizing empty parameters as valid, failing to activate workflows with v1 Execute Workflow Trigger nodes.

2. **Version 1.1 nodes with missing types**: The Execute Workflow Trigger v1.1 has a default type of "string" for workflow inputs. When users don't explicitly set a type, it's not included in the JSON. The migration's `hasValidWorkflowInputs()` validation required the `type` field to exist, incorrectly rejecting valid configurations.

### Migration ([1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts](packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts))

1. **Version 1 support** (lines 99-102): Added check for empty or missing parameters object. If `!params || Object.keys(params).length === 0`, the workflow is considered valid and will be activated.

2. **Version 1.1 missing type support** (lines 195-196): Updated `hasValidWorkflowInputs()` to treat missing `type` field as valid:
   ```typescript
   // type is optional (defaults to 'string' in version 1.1)
   (!('type' in value) || (typeof value.type === 'string' && value.type.length > 0))

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
+++ b/packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
@@ -93,21 +93,28 @@ export class ActivateExecuteWorkflowTriggerWorkflows1763048000000 implements Irr
 					const shouldActivateByInputSource =
 						inputSource === 'passthrough' || inputSource === 'jsonExample';
 
-					// For legacy nodes (no inputSource), check if they have parameters defined
-					let hasLegacyParameters = false;
+					// For nodes without inputSource (version 1 or legacy version 1.1)
+					let hasLegacyParametersOrIsVersion1 = false;
 					if (!inputSource) {
-						const workflowInputs = executeWorkflowTriggerNode.parameters?.workflowInputs;
-						hasLegacyParameters = Boolean(
-							workflowInputs &&
-								typeof workflowInputs === 'object' &&
-								'values' in workflowInputs &&
-								Array.isArray(workflowInputs.values) &&
-								workflowInputs.values.length > 0 &&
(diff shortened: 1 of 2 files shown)


packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
```diff
@@ -93,21 +93,28 @@ export class ActivateExecuteWorkflowTriggerWorkflows1763048000000 implements Irr
 					const shouldActivateByInputSource =
 						inputSource === 'passthrough' || inputSource === 'jsonExample';
 
-					// For legacy nodes (no inputSource), check if they have parameters defined
-					let hasLegacyParameters = false;
+					// For nodes without inputSource (version 1 or legacy version 1.1)
```
tomi, line 96: For future: I think this would easier to follow if some of these extracted into their own functions. No need to do anymore now

tomi (approved):

RicardoE105:
@tomi had to sync with master because the e2e were stuck 😞

tomi (approved):

PR #22861, opened 2025-12-05
fix(core): Workflow diff preview errors out in push dialog for workflow that's newly created

konstantintieber:
## Summary

Fixes the error that's currently happening when trying to view the diff of a workflow that is newly created and to be pushed.

How to reproduce (before fix):
- On an instance that's connected to a git repo
- Create a new workflow
- Save it
- Click "Push" in the sidebar
- Click on "Compare versions" on the newly created workflow

### Before

When clicking "Compare versions":

Screenshot 2025-12-05 at 23 00 26

### After

When clicking "Compare versions":

Screenshot 2025-12-05 at 23 00 05

closes PAY-4107

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/workflows/workflowDiff/WorkflowDiffModal.vue
```diff
@@ -62,6 +68,11 @@ const isClosed = ref(false);
 
 const remote = useAsyncState<{ workflow?: IWorkflowDb; remote: boolean } | undefined, [], false>(
 	async () => {
+		if (props.data.direction === 'push' && props.data.workflowStatus === 'created') {
```
alphaautomator, line 71: Hard-coding status strings makes future terminology changes much harder, because we don’t always know every place that needs to be updated. This increases the risk of missing references and causes issues to appear only at runtime.

Using SourceControlledFileStatus (the Zod enum) is a better approach because it gives us a single source of truth. If any key in the enum changes, TypeScript will throw build-time errors, and VS Code will surface those issues in the Problems tab. This makes refactoring much safer and easier to track.

thoughts ??

konstantintieber: We have type checking for these strings, so that they always match a valid git file status. So it's all good.

packages/frontend/editor-ui/src/app/composables/useWorkflowDiffRouting.ts
```diff
@@ -48,6 +49,7 @@ export function useWorkflowDiffRouting() {
 
 	const handleDiffModal = (
 		diffWorkflowId: string | undefined,
+		diffWorkflowStatus: SourceControlledFileStatus | undefined,
```
cstuncsik, line 52: Just somtype nitpicks

```suggestion
		diffWorkflowStatus?: SourceControlledFileStatus,
```

Also this prop is defined as required in `WorkflowDiffModal.vue`, I don't know why the type checking didn't catch this (maybe because of the way we use/define our `Modal` component)

packages/frontend/editor-ui/src/features/workflows/workflowDiff/WorkflowDiffModal.vue
```diff
@@ -36,7 +37,12 @@ import {
 	N8nText,
 } from '@n8n/design-system';
 const props = defineProps<{
-	data: { eventBus: EventBus; workflowId: string; direction: 'push' | 'pull' };
+	data: {
+		eventBus: EventBus;
+		workflowId: string;
+		direction: 'push' | 'pull';
+		workflowStatus: SourceControlledFileStatus;
```
cstuncsik: As I mentioned earlier, this is required here but can be `undefined`

cstuncsik (dismissed):

packages/frontend/editor-ui/src/app/composables/useWorkflowDiffRouting.ts
konstantintieber: Yeah it's because we only reference the modal by its, key not the actual instance.
Good find.

I addressed this here: https://github.com/n8n-io/n8n/pull/22861/commits/c7833a348a059766c5347f7b7ef3281fffcf2d47

cstuncsik (approved):

PR #22862, opened 2025-12-05
feat(core): Dynamic credential resolvers API routes

guillaumejacquart:
## Summary

Adds CRUD controller routes and schemas for credential resolvers

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
@@ -1,4 +1,9 @@
-import { CreateCredentialResolverDto, UpdateCredentialResolverDto } from '@n8n/api-types';
+import {
+	CreateCredentialResolverDto,
+	CredentialResolver,
+	credentialResolverSchema,
+	UpdateCredentialResolverDto,
+} from '@n8n/api-types';
 import { AuthenticatedRequest } from '@n8n/db';
 import {
 	Body,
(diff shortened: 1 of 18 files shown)


packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
```diff
@@ -0,0 +1,144 @@
...
+	constructor(private readonly service: DynamicCredentialResolverService) {}
+
+	@Get('/')
+	@GlobalScope('credentialResolver:list')
+	async listResolvers(
+		_req: AuthenticatedRequest,
+		_res: Response,
+	): Promise<Array<CredentialResolver>> {
+		try {
+			return (await this.service.findAll()).map((resolver) =>
```
BGZStephen: It may be nicer to have a schema for the array of resolvers for parsing, otherwise a single schema failure for a resolver will hide potential issues with other resolvers

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
```diff
@@ -0,0 +1,144 @@
...
+import { DynamicCredentialResolverNotFoundError } from './errors/credential-resolver-not-found.error';
+import { DynamicCredentialResolverService } from './services/credential-resolver.service';
+
+@RestController('/credential-resolvers')
+export class CredentialResolversController {
+	constructor(private readonly service: DynamicCredentialResolverService) {}
+
+	@Get('/')
+	@GlobalScope('credentialResolver:list')
+	async listResolvers(
```
BGZStephen, line 35: Do we think we may need pagination in here?

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
```diff
@@ -0,0 +1,144 @@
...
+			if (e instanceof Error) {
+				throw new InternalServerError(e.message, e);
+			}
+			throw e;
+		}
+	}
+
+	@Delete('/:id')
+	@GlobalScope('credentialResolver:delete')
+	async deleteResolver(
```
BGZStephen, line 125: Do we need to be checking for credentials that are using the resolver? and possibly not deleting if it's in use?

BGZStephen (dismissed):
Some minor comments

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
guillaumejacquart: Maybe at some point, but for now I don't think customers will create more that a handful of resolvers

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
guillaumejacquart: Good point ! I'm just waiting on the workflow settings to add the resolver to check on that for deletion. I'll create a task

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
guillaumejacquart: Good point, done !

BGZStephen (approved):

PR #22912, opened 2025-12-08
fix(editor): Fix workflow list endpoint triggerNodeType filter query (no-changelog)

scdekov:
## Summary

The option to filter by triggerNodeType was added specifically for the wfs selector in the "Execute Sub-workflow" node. The way it should work is that it should show workflows that 1) if active has "When executed by other workflow" trigger in their active version and 2) if not active has "When executed by other workflow" trigger in their last saved version.

## Review / Merge checklist

- [ ] Tests included. 


nikhilkuria (dismissed):

nikhilkuria:
Good catch 👏

nikhilkuria (approved):

PR #22915, opened 2025-12-08
fix(core): Handle misformed nodes in credential indexing (no-changelog)

mfsiega:
## Summary

We cover a few cases:
- Credential id is null
- Missing node type
- Webhook doesn't have a path

These are all malformed nodes in a sense, but we prefer to tolerate them without throwing errors.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/workflow-index/workflow-index.service.ts
+++ b/packages/cli/src/modules/workflow-index/workflow-index.service.ts
@@ -94,9 +94,6 @@ export class WorkflowIndexService {
 		const dependencyUpdates = new WorkflowDependencies(workflow.id, workflow.versionCounter);
 
 		workflow.nodes.forEach((node) => {
-			if (node.disabled) {
-				return; // skip disabled nodes.
-			}
 			this.addNodeTypeDependencies(node, dependencyUpdates);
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/workflow-index/workflow-index.service.ts
```diff
@@ -94,6 +94,9 @@ export class WorkflowIndexService {
 		const dependencyUpdates = new WorkflowDependencies(workflow.id, workflow.versionCounter);
 
 		workflow.nodes.forEach((node) => {
+			if (node.disabled) {
```
tomi: Do we actually want to skip these? I would imagine that I am still able to find workflows that have a specific node, even if it disabled

tomi (commented):
I think we should keep the disabled nodes in the index, WDYT?

packages/cli/src/modules/workflow-index/workflow-index.service.ts
mfsiega: I don't have a strong preference - happy to include these.

tomi (approved):

PR #22922, opened 2025-12-08
fix(core): Files are not included in memory in chat (no-changelog)

autologie:
## Summary

This PR fixes the issue in ChatHub where models cannot generate response based on attachment files in the conversation history. The fix is implemented by including files as content blocks when injecting memory.

Related, but not fully address https://linear.app/n8n/issue/AI-1648/attached-files-are-not-included-in-memory

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
```diff
@@ -349,17 +349,13 @@ export function createSessionFromStreamingState(streaming: ChatStreamingState):
 	};
 }
 
-export function createMimeTypes(modalities: ChatHubInputModality[]): string | undefined {
-	if (modalities.length === 0) {
-		return undefined;
-	}
-
+export function createMimeTypes(modalities: ChatHubInputModality[]): string {
```
autologie, line 352: Text attachments are sent as text content block (like normal chat message), so we can always accept text files

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -698,6 +694,111 @@ When you need to reference “now”, use this date and time.`;
...
+			};
+		}
+
+		if (attachment.mimeType.startsWith('text/')) {
+			const buffer = await this.chatHubAttachmentService.getAsBuffer(attachment);
+			const content = buffer.toString('utf-8');
+
+			return {
+				type: 'text',
+				text: `File: ${attachment.fileName ?? 'attachment'}\nContent: \n${content}`,
```
autologie: Same format as https://github.com/n8n-io/n8n/blob/fix-files-in-memory/packages/%40n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/common.ts#L119

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -698,6 +694,111 @@ When you need to reference “now”, use this date and time.`;
...
+		});
+
+		if (currentTotalSize + url.length > maxTotalPayloadSize) {
+			return {
+				type: 'text',
+				text: `File: ${attachment.fileName ?? 'attachment'}\n(Content omitted due to size limit)`,
+			};
+		}
+
+		return { type: 'image_url', image_url: url };
```
autologie: At least for models I tested this worked for PDF as well

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -698,6 +694,111 @@ When you need to reference “now”, use this date and time.`;
...
+	private async buildContentBlockForAttachment(
+		attachment: IBinaryData,
+		provider: ChatHubProvider,
+		currentTotalSize: number,
+		maxTotalPayloadSize: number,
+	): Promise<ContentBlock> {
+		if (currentTotalSize >= maxTotalPayloadSize) {
+			return {
+				type: 'text',
+				text: `File: ${attachment.fileName ?? 'attachment'}\n(Content omitted due to size limit)`,
```
autologie: For models to not hallucinate, clearly state that file is omitted.

Response looks like this:
Screenshot 2025-12-09 at 11 46 04

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -698,6 +694,111 @@ When you need to reference “now”, use this date and time.`;
...
+		const typeMap: Record<string, MessageRecord['type']> = {
+			human: 'user',
+			ai: 'ai',
+			system: 'system',
+		};
+
+		const messageValues: MessageRecord[] = [];
+		let currentTotalSize = 0;
+
+		for (const message of history.slice().reverse()) {
```
autologie: Traversing messages from last to prioritize newer attachments (in case the limit is reached)

packages/cli/src/modules/chat-hub/chat-hub.attachment.service.ts
```diff
@@ -126,6 +130,35 @@ export class ChatHubAttachmentService {
...
+	): Promise<string> {
+		if (binaryData.data.startsWith('data:')) {
+			return binaryData.data;
+		}
+
+		if (!mustBeDataUrl) {
+			const baseUrl = this.urlService.getInstanceBaseUrl();
+
+			if (baseUrl.startsWith('https://')) {
+				const token = this.binaryDataService.createSignedToken(binaryData, '1 hour');
```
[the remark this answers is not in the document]
autologie: For memory mode, `data` field should be populated with base64-encoded data, so if neither data nor id is missing, it means there's a system error.

packages/cli/src/modules/chat-hub/chat-hub.attachment.service.ts
```diff
@@ -126,6 +130,35 @@ export class ChatHubAttachmentService {
...
+		{ mustBeDataUrl }: { mustBeDataUrl: boolean },
+	): Promise<string> {
+		if (binaryData.data.startsWith('data:')) {
+			return binaryData.data;
+		}
+
+		if (!mustBeDataUrl) {
+			const baseUrl = this.urlService.getInstanceBaseUrl();
+
+			if (baseUrl.startsWith('https://')) {
```
Cadiac: is there a limitation that https is required?

packages/cli/src/modules/chat-hub/chat-hub.attachment.service.ts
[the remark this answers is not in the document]
Cadiac: is there even memory mode for binary data anymore in v2?

packages/cli/src/modules/chat-hub/chat-hub.attachment.service.ts
autologie: As long as I tested, OpenAI requires https. But having https scheme doesn't guarantee the instance is reachable by external model providers. Maybe we should always use data URL for now..

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -706,6 +703,122 @@ ${this.getSystemMessageMetadata(timeZone)}`;
...
+		maxTotalPayloadSize: number,
+	): Promise<ContentBlock> {
+		class TotalFileSizeExceededError extends Error {}
+
+		try {
+			if (currentTotalSize >= maxTotalPayloadSize) {
+				throw new TotalFileSizeExceededError();
+			}
+
+			if (attachment.mimeType.startsWith('text/')) {
```
Cadiac: this should also handle `application/json`, didn't we have some "is text mimetype" helper somewhere that we might be able to use here?

autologie: Fixed in [fc22e3f](https://github.com/n8n-io/n8n/pull/22922/commits/fc22e3faa006edb8b9a3cda5b025acd2361b338d) to do the same check as tools agent does.

packages/cli/src/modules/chat-hub/chat-hub.attachment.service.ts
autologie: Did a second thought and I think best to always use data URL for now. Updated [52c9d71](https://github.com/n8n-io/n8n/pull/22922/commits/52c9d716c87a5a58e33dfb238d1e3b8de8ea55df)

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: Should a comment say this also in code? 😅

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -819,6 +804,17 @@ ${this.getSystemMessageMetadata(timeZone)}`;
 		}
 	}
 
+	private isTextFile(mimeType: string): boolean {
```
Cadiac, line 809: Great, this was the logic I remembered seeing somwhere else, good that we now cover the other types as well. 👍

packages/cli/src/modules/chat-hub/chat-hub.attachment.service.ts
```diff
@@ -130,24 +126,11 @@ export class ChatHubAttachmentService {
 		);
 	}
 
-	async getPubliclyAccessibleUrl(
-		binaryData: IBinaryData,
-		{ mustBeDataUrl }: { mustBeDataUrl: boolean },
-	): Promise<string> {
+	async getPubliclyAccessibleUrl(binaryData: IBinaryData): Promise<string> {
```
Cadiac: Nit: Isn't the name of this function now a bit misleading? Should we just call it `getDataUrl`?

Cadiac (commented):
Minor comments about naming but nice, appears to work

Cadiac (approved):

PR #22926, opened 2025-12-08
feat(editor): Implement new DropdownMenu component (no-changelog)

MiloradFilipovic:
## Summary
Implemented new `DropdownMenu` component:
image

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.typeguards.ts
```diff
@@ -0,0 +1,7 @@
+import type { Align, Side } from './DropdownMenu.types';
+
+const VALID_SIDES: Side[] = ['top', 'bottom', 'left', 'right'];
+const VALID_ALIGNS: Align[] = ['start', 'end', 'center'];
+
+export const isSide = (value: string): value is Side => VALID_SIDES.includes(value as Side);
+export const isAlign = (value: string): value is Align => VALID_ALIGNS.includes(value as Align);
```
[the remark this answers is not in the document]
MiloradFilipovic: This is the typeguard

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.typeguards.ts
```diff
@@ -0,0 +1,7 @@
+import type { Align, Side } from './DropdownMenu.types';
+
+const VALID_SIDES: Side[] = ['top', 'bottom', 'left', 'right'];
+const VALID_ALIGNS: Align[] = ['start', 'end', 'center'];
+
+export const isSide = (value: string): value is Side => VALID_SIDES.includes(value as Side);
```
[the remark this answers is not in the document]
MiloradFilipovic: This is the typeguard

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/composables/useMenuKeyboardNavigation.ts
r00gm, line 1: AFAIK reka already takes care of keyboard navigation, is this really needed?

MiloradFilipovic: For searchable menus, we want to keep the focus in the search input (design requirement) so built-in navigation doesn't work. If the menu is not searchable, we let Reka handle the keys

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
```diff
@@ -0,0 +1,375 @@
...
+	padding: var(--spacing--4xs);
+}
+
+.item {
+	font-size: var(--font-size--2xs);
+	line-height: 1;
+	border-radius: var(--radius);
+	display: flex;
+	align-items: center;
+	min-height: var(--spacing--lg);
```
RobTF9: This should always be `32px` don't allow multiple lines

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.vue
```diff
@@ -0,0 +1,355 @@
...
+		cursor: not-allowed;
+	}
+}
+
+.activator-emoji {
+	font-size: var(--font-size--md);
+	line-height: 1;
+}
+
+.content {
```
RobTF9, line 335: The wrapper for the `MenuItems` needs padding of 4px all the way around. Not just on the top.

There should also be a 2px gap between each menut item

Image

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
```diff
@@ -0,0 +1,375 @@
...
+.item-check,
+.item-trailing {
+	margin-left: auto;
+	flex-shrink: 0;
+}
+
+.separator {
+	height: 1px;
+	background-color: var(--color--foreground);
+	margin: var(--spacing--4xs) 0;
```
RobTF9: If add a gap to the content wrapper of 2px then will need to adjust this margin to 2px as well.

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
```diff
@@ -0,0 +1,375 @@
...
+		margin-bottom: 0;
+	}
+}
+
+.empty-state {
+	padding: var(--spacing--2xs) var(--spacing--xs);
+	color: var(--color--text--tint-1);
+	font-size: var(--font-size--sm);
+	text-align: center;
+}
```
RobTF9, line 389: Empty state isn't spaced properly. That text should be vertically centered between the bottom and the search border (though maybe addressing the padding and height issues fix it?)

Image

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
```diff
@@ -0,0 +1,375 @@
...
+	user-select: none;
+	color: var(--color--text--shade-1);
+	gap: var(--spacing--3xs);
+	outline: none;
+
+	&:not([data-disabled]) {
+		&:hover,
+		&[data-highlighted],
+		&.highlighted {
+			background-color: var(--color--background--light-1);
```
RobTF9: Use `var(--color--foreground--tint-1)` the current color isn't strong enough in light mode

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/composables/useMenuKeyboardNavigation.ts
r00gm: i think that is against accessibility, we would be actively preventing the user to navigate with the keyboard?

or do you mean that it has to be focused at first?

MiloradFilipovic: Visually this behaves the same as non-searchable fields but it does have some accessibility gaps, mainly the `aria` attributes. Unless I am missing something more, these shouldn't be hard to add. Do you have any more suggestions?
Also @RobTF9 what do you think about this?

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuSearch.vue
```diff
@@ -0,0 +1,126 @@
...
+	</div>
+</template>
+
+<style module>
+.search-container {
+	display: flex;
+	align-items: center;
+	padding: var(--spacing--4xs) var(--spacing--2xs);
+	border-bottom: var(--border);
+	margin-bottom: var(--spacing--4xs);
```
RobTF9: There will be too much spacing here after you add the padding around the items so can remove this margin-bottom.

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuSearch.vue
```diff
@@ -0,0 +1,126 @@
...
+
+<template>
+	<div :class="$style['search-container']">
+		<Icon
+			v-if="showIcon"
+			:class="$style['search-icon']"
+			icon="search"
+			size="large"
+			color="text-light"
+		/>
```
RobTF9: We don't ever want to allow an icon in this search field. Will look too similar to one of the menu items if we do so this can be removed.

RobTF9 (changes requested):
Nice work! Some style tweaks below.

One thing that we can address later but I thought of while reviewing - how would lazy loading and virtualisation work here? Thinking if this ever needs to display tags, projects or workflows that might become an issue

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/composables/useMenuKeyboardNavigation.ts
RobTF9: Are we talking about the ability to type in the field search even one of the menu item is selected?

That's a pretty common pattern, reka even do it in the combobox I think?

Also how dropdowns work in apps like Linear, Notion, figma etc

MiloradFilipovic:
> Nice work! Some style tweaks below.
> 
> One thing that we can address later but I thought of while reviewing - how would lazy loading and virtualisation work here? Thinking if this ever needs to display tags, projects or workflows that might become an issue

@RobTF9 Regarding lazy loading: The component itself doesn't handle any loading logic but it exposes enough for it to be handled from the outside. We didn't however had events for opening individual submenus so lazy loading can be triggered for them individually. I added that just now and updated storybook with an entry that showcases lazy loading.
Regarding virtual scroll: There is no virtual scroll support in Reka UI `DropdownMenu` so if we want it we're going to implement it ourselves.

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.typeguards.ts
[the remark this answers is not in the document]
MiloradFilipovic: These are typeguards

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.typeguards.ts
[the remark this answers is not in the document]
MiloradFilipovic: These are typeguards

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.test.ts
```diff
@@ -0,0 +1,666 @@
...
+				props: {
+					items: createItems(3),
+				},
+			});
+
+			const trigger = container.querySelector('button')!;
+			await userEvent.click(trigger);
+
+			const { dropdown } = await getDropdownContent();
+			expect(dropdown).toBeInTheDocument();
```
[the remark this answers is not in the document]
MiloradFilipovic: The test name doesn't mention loading state

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuSearch.vue
```diff
@@ -0,0 +1,108 @@
...
+	padding: var(--spacing--4xs) var(--spacing--2xs);
+	border-bottom: var(--border);
+}
+
+.search-input {
+	flex: 1;
+	min-width: 0;
+	border: none;
+	background: transparent;
+	outline: none;
```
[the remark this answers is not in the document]
MiloradFilipovic: We have the highlighed class for this

r00gm (approved):
only a nit pick.

'empty-state' is probably better to use camelCase css, given that we are using css modules.

i wont block

RobTF9 (approved):
This is awesome, great work!

PR #22937, opened 2025-12-08
fix: Allow negative numbers when parsing objects from string

dlavrenuek:
## Summary
When mapping data in n8n, if mapping is not specified as valid JSON, any fields containing negative numbers (both floats and integers) are lost in the resulting mapped data. Use the attached workflow to reproduce.

[SetNodeIssue.json](https://github.com/user-attachments/files/24037286/SetNodeIssue.json)

A workaround is to specify the mapping as valid json string.

https://linear.app/n8n/issue/NODE-4048/negative-numbers-lost-when-mapped-to-object-enterprise-client-qualcomm

## Review / Merge checklist

- [x] Tests included. 


packages/workflow/src/utils.ts
```diff
@@ -87,6 +87,12 @@ function syntaxNodeToValue(expression?: SyntaxNode | null): unknown {
 			return expression.value;
 		case Syntax.ArrayExpression:
 			return expression.elements.map((exp) => syntaxNodeToValue(exp));
+		case Syntax.UnaryExpression:
```
dlavrenuek: In contrast to positive numbers like `123` which are parsed as `Literal`, negative numbers like `-123` are returned as `UnaryExpression` and contain an `expression.operator` and the value in `expression.argument`. The argument itself can again be any type of node, so I have reused `syntaxNodeToValue(...)` here. This is to address any "invalid" syntax we don't support to prevent exceptions instead of assuming that `expression.argument.value` will be a number. For example `{test: -Date.now()}` as input - will be correctly parsed as `{test: undefined}` (same behavior as before with `{test: Date.now()}`.

packages/workflow/src/utils.ts
```diff
@@ -87,6 +87,12 @@ function syntaxNodeToValue(expression?: SyntaxNode | null): unknown {
...
+			const value = syntaxNodeToValue(expression.argument);
+			if (typeof value === 'number' && expression.operator === '-') {
```
DawidMyslak, line 92: Do we ever need to worry about the case `expression.operator === '+'`?

DawidMyslak (changes requested):
Looks good @dlavrenuek 👏 ! Just wanted to clarify minor thing and then I'm happy to approve.

packages/workflow/src/utils.ts
```diff
@@ -87,6 +87,12 @@ function syntaxNodeToValue(expression?: SyntaxNode | null): unknown {
 			return expression.value;
 		case Syntax.ArrayExpression:
 			return expression.elements.map((exp) => syntaxNodeToValue(exp));
+		case Syntax.UnaryExpression:
+			const value = syntaxNodeToValue(expression.argument);
```
DawidMyslak, line 91: ESLint is unhappy here:
Image

It can be fixed by wrapping it with `{}`:
```ts
		case Syntax.UnaryExpression: {
			const value = syntaxNodeToValue(expression.argument);
			if (typeof value === 'number' && expression.operator === '-') {
				return -value;
			}
			return value;
		}
```

packages/workflow/src/utils.ts
dlavrenuek: Technically what is done here is parsing JS syntax into language components and it's a fallback if the data is not valid JSON. What was implemented here so far also only supports what json supports: string (`Identifier`), number (`Literal`), object (`ObjectExpression`), array (`ArrayExpression`) and boolean(`Literal`), just the negative numbers are parsed not as `Literal`. So technically we could encounter a unaryexpression for `!true` or `!1`, but that's not what we support in my opinion.
Long story short: as I see it this js-object parsing fallback supports the same data types as JSON - in case of numbers if it's not negative, then it must be positive, so I don't see a need for special handling of `+`

packages/workflow/src/utils.ts
dlavrenuek: Interesting, when running `pnpm lint` there are no errors and I also don't see any errors in the IDE. Also the [CI lint passes](https://github.com/n8n-io/n8n/actions/runs/20035479397/job/57456415977?pr=22937). Do you may be have special linting rules and not the ones specified in the repository?

dlavrenuek: PS: I've applied the suggestion

DawidMyslak: Weird 🤔...

The rule comes from here: https://github.com/n8n-io/n8n/blob/master/packages/workflow/eslint.config.mjs#L18

Approving 👌

DawidMyslak (approved):

PR #22947, opened 2025-12-08
feat(core): Make n8n workflows on Chat require streaming response mode (no-changelog)

Cadiac:
## Summary

Chat will initially ship with just `streaming`  response mode support, and before we've been forcing the workflows to be run in this response mode, which has led to some confusion.

Instead of forcing the response mode to `streaming` make the `streaming` mode a requirement, making it so that the agent can't be used on Chat unless the correct streaming mode is selected.

To not make this clunky to use make it so that if the `availableInChat` toggle is enabled the available Response Mode options change to only `streaming`, and the default changes. Since `availableInChat` is a new field in 1.4 version of the node this shouldn't cause regressions, as the field is default: false.

Also make it possible to have `streaming` response mode in non public workflows, and improve Chat error handling on 400 errors, displaying the error text on the error toast.

https://www.notion.so/n8n/2955b6e0c94f80298ea6e47f0abdd9dc?v=2955b6e0c94f8169af97000c22abcddb&p=2ac5b6e0c94f8074b992fbde609ed7b9&pm=s

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
+++ b/packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
@@ -532,7 +532,7 @@ export class ChatTrigger extends Node {
 						displayName: 'Response Mode',
 						name: 'responseMode',
 						type: 'options',
-						options: [lastNodeResponseMode, respondNodesResponseMode],
+						options: [lastNodeResponseMode, respondNodesResponseMode, streamingResponseMode],
 						default: 'lastNode',
 						description: 'When and how to respond to the chat',
 					},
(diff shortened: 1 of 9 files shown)


autologie:
Initial value is set to lastNode, but I guess this isn't a quick fix

https://github.com/user-attachments/assets/86ce586d-87e8-40d9-a87b-ebbdd97a69b9

Cadiac:
> Initial value is set to lastNode, but I guess this isn't a quick fix
> 
>  Screen.Recording.2025-12-09.at.09.18.16.mov

Let me check that case, it should know to pick the correct default there i think but yeah, this was tricky. at least you should be seeing an error...

Cadiac:
> > Initial value is set to lastNode, but I guess this isn't a quick fix
> > Screen.Recording.2025-12-09.at.09.18.16.mov
> 
> Let me check that case, it should know to pick the correct default there i think but yeah, this was tricky. at least you should be seeing an error...

In case there are multiple options with the same identifier the initial default when adding the options is from the first one on the list, so if I swap them around like this streaming gets chosen - which at least is present on both lists

image

I'll see if I could fix the logic, but I don't think we want to change the order here as then `streaming` would become the value the options pick normally.

autologie:
Is there a fixed text content type that we can show instead of type: options when availableInChat is set to true?

autologie (dismissed):

autologie (dismissed):
Worked like a charm!

autologie (approved):

PR #22949, opened 2025-12-08
fix(Microsoft OneDrive Node): Don't overwrite filename from node parameters when uploading binary file

### LINKED ISSUE #22758: Microsoft Onedrive fileupload ignores settings filename when binary is selected
### Bug Description

When you upload a file to OneDrive the filename from the settings is ignored, and the filename from the binary file itself is used. 
I can't currently make a PR for this, but the error is here:

https://github.com/n8n-io/n8n/blob/f3fa3f9c30ea3493f013f353c53fbbf49a77d512/packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts#L209-L226

the encodedFilename var is set from the settings, but then gets overridden if there is a filename in the binary file. I would suggest switching the order of these two if statements. Check for filename in the binary file first , and overwrite if there is a filename set in the settings. 

### To Reproduce

1. have a binary file
2. Upload to Microsoft Onedrive
3. Set filename in settings, and select 'binary file'
4. Execute step
5. Filename on Onedrive has binary file filename, not the settings one

### Expected behavior

Expected the filename from the settings to be used when filled out

### Debug Info

Not needed. Not an error, just a wrong filename

### Operating System

Docker

### n8n Version

1.122.4

### Node.js Version

unknown

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

krisn0x:
The node always used the binary.fileName even when there was a file name provided in the node parameters. 
This adds v1.1 of the node and fixes the IF statements flow, so the file name from the node parameters is prioritised.

https://linear.app/n8n/issue/GHC-5780/community-issue-microsoft-onedrive-fileupload-ignore-settings-filename
fixes #22758

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
+++ b/packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
@@ -218,15 +218,18 @@ export class MicrosoftOneDrive implements INodeType {
 							const body = await this.helpers.getBinaryDataBuffer(i, binaryPropertyName);
 							let encodedFilename;
 
-							if (nodeVersion == 1 && binaryData.fileName !== undefined) {
-								// pre-fix behavior, filename provided in node parameters is always ignored
-								encodedFilename = encodeURIComponent(binaryData.fileName);
+							if (nodeVersion >= 1.1) {
+								if (fileName !== '') {
+									encodedFilename = encodeURIComponent(fileName);
+								} else if (binaryData.fileName !== undefined) {
+									encodedFilename = encodeURIComponent(binaryData.fileName);
+								}
 							} else {
 								if (fileName !== '') {
 									encodedFilename = encodeURIComponent(fileName);
 								}
 
-								if (fileName == '' && binaryData.fileName !== undefined) {
+								if (binaryData.fileName !== undefined) {
 									encodedFilename = encodeURIComponent(binaryData.fileName);
 								}
 							}
--- a/packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
+++ b/packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
@@ -19,7 +19,7 @@ export class MicrosoftOneDrive implements INodeType {
 		name: 'microsoftOneDrive',
 		icon: 'file:oneDrive.svg',
 		group: ['input'],
-		version: 1,
+		version: [1, 1.1],
 		subtitle: '={{$parameter["operation"] + ": " + $parameter["resource"]}}',
 		description: 'Consume Microsoft OneDrive API',
 		defaults: {
@@ -63,6 +63,7 @@ export class MicrosoftOneDrive implements INodeType {
 		const items = this.getInputData();
 		const returnData: INodeExecutionData[] = [];
 		const length = items.length;
+		const nodeVersion = this.getNode().typeVersion;
 		let responseData;
 		const resource = this.getNodeParameter('resource', 0);
 		const operation = this.getNodeParameter('operation', 0);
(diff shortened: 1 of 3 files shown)


packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
```diff
@@ -217,12 +218,17 @@ export class MicrosoftOneDrive implements INodeType {
...
 								encodedFilename = encodeURIComponent(binaryData.fileName);
+							} else {
+								if (fileName !== '') {
+									encodedFilename = encodeURIComponent(fileName);
+								}
+
+								if (fileName == '' && binaryData.fileName !== undefined) {
+									encodedFilename = encodeURIComponent(binaryData.fileName);
+								}
 							}
```
RomanDavydchuk: I think this whole `if-else` statement is a bit confusing. When it comes to versioning, it's quite confusing as-is, so we try to make conditions related to `nodeVersion` as simple as possible. This particular code combines conditions related both to the _node version_ and the _parameter values_. I'd suggest rewriting it similar to this:
```typescript
if (nodeVersion >= 1.1) {
  if (fileName !== '') {
    encodedFilename = encodeURIComponent(fileName);
  } else if (binaryData.fileName !== undefined) {
    encodedFilename = encodeURIComponent(binaryData.fileName);
  }
} else {
  if (fileName !== '') {
    encodedFilename = encodeURIComponent(fileName);
  }
  
  if (binaryData.fileName !== undefined) {
    encodedFilename = encodeURIComponent(binaryData.fileName);
  }
}
```
So we have an `if-else` statement on the top level that handles the version and then each branch is version-specific behavior. In the `true` branch it's the correct behavior: `binaryData.fileName` is used **only when `fileName` is empty**, while the `false` branch is exactly the same as old version

There are cases where we can simplify these conditions, but IMO for this case it's better this way, even though it's more verbose and nested. I think you code is correct, but I had to think quite a bit to make sure it actually is 😅

packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
```diff
@@ -217,12 +218,17 @@ export class MicrosoftOneDrive implements INodeType {
 							const body = await this.helpers.getBinaryDataBuffer(i, binaryPropertyName);
 							let encodedFilename;
 
-							if (fileName !== '') {
-								encodedFilename = encodeURIComponent(fileName);
-							}
-
-							if (binaryData.fileName !== undefined) {
+							if (nodeVersion == 1 && binaryData.fileName !== undefined) {
```
RomanDavydchuk: We always use strict equals operation (`===`) unless we need to use loose equals (`==`) specifically. It's better to use `===` here:
```suggestion
							if (nodeVersion === 1 && binaryData.fileName !== undefined) {
```

packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
```diff
@@ -217,12 +218,17 @@ export class MicrosoftOneDrive implements INodeType {
...
+								// pre-fix behavior, filename provided in node parameters is always ignored
 								encodedFilename = encodeURIComponent(binaryData.fileName);
+							} else {
+								if (fileName !== '') {
+									encodedFilename = encodeURIComponent(fileName);
+								}
+
+								if (fileName == '' && binaryData.fileName !== undefined) {
```
RomanDavydchuk: Same here
```suggestion
								if (fileName === '' && binaryData.fileName !== undefined) {
```

packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
```diff
@@ -217,12 +218,17 @@ export class MicrosoftOneDrive implements INodeType {
...
+									encodedFilename = encodeURIComponent(binaryData.fileName);
+								}
```
RomanDavydchuk, line 234: This can also be simplified. The first `if` statement will execute only when `fileName` **is not** an empty string and the second one will execute only when `fileName` **is** an empty string (and some other condition). So you can rewrite it like so for better readability:
```typescript
if (fileName !== '') {
  encodedFilename = encodeURIComponent(fileName);
} else if (binaryData.fileName !== undefined) {
  encodedFilename = encodeURIComponent(binaryData.fileName);
}
```
which is exactly what I did in the code I proposed in my first comment

RomanDavydchuk (changes requested):
Hey, I just looked through the code for now and left some comments. I'd say only the first one is actually about changes I'm requesting and the other 3 are more like tips for the future. Will test manually after the changes. But the tests look good though, so you'll be able to tell if it still works after making the changes 🙌

packages/nodes-base/nodes/Microsoft/OneDrive/MicrosoftOneDrive.node.ts
krisn0x: Yeah, I thought hard about how it would not look confusing and everything was pretty bad hahah, I agree with the readability improvement here, I missed it as I already knew how it works.

RomanDavydchuk (approved):
LGTM and works as expected

PR #22951, opened 2025-12-08
feat(Data Table Node): Add data table crud operations to data table node

RicardoE105:
## Summary
  Adds CRUD operations for data tables to the DataTable node, allowing users to manage data tables programmatically through workflows.

  ### Changes
  - Added new "Table" resource to DataTable node alongside existing "Row" resource
  - Implemented 4 table operations:
    - **Create**: Create new data tables with configurable columns and types
    - **Delete**: Delete existing data tables (with warning notice)
    - **List**: List all data tables with filtering and sorting options
    - **Update**: Rename existing data tables

  ## Related Linear issue
  https://linear.app/n8n/issue/ADO-4516

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/DataTable/actions/table/Table.resource.ts
```diff
@@ -0,0 +1,90 @@
...
+			{
+				name: 'Update',
+				value: update.FIELD,
+				description: 'Update a data table name',
+				action: 'Update a data table',
+			},
+		],
+		default: 'list',
+	},
+	{
```
CharlieKolb, line 50: [nit] would be nice to reference a shared object for all properties besides `displayOptions`

packages/nodes-base/nodes/DataTable/actions/table/create.operation.ts
```diff
@@ -0,0 +1,133 @@
...
+	{
+		displayName: 'Options',
+		name: 'options',
+		type: 'collection',
+		placeholder: 'Add Option',
+		default: {},
+		displayOptions,
+		options: [
+			{
+				displayName: 'Reuse Existing Tables',
```
CharlieKolb, line 81: Are you sure we need this?

If the default behavior was to throw an error on duplicate name we'd already have the node setting of continuing on error to cover this functionality without duplication.

packages/nodes-base/nodes/DataTable/actions/table/create.operation.ts
```diff
@@ -0,0 +1,133 @@
...
+	},
+];
+
+export async function execute(
+	this: IExecuteFunctions,
+	index: number,
+): Promise<INodeExecutionData[]> {
+	const tableName = this.getNodeParameter('tableName', index) as string;
+	const columnsData = this.getNodeParameter('columns.column', index, []) as Array<{
+		name: string;
```
CharlieKolb: Can we reuse existing types here?

packages/nodes-base/nodes/DataTable/actions/table/create.operation.ts
```diff
@@ -0,0 +1,133 @@
...
+		type: col.type,
+		index: idx,
+	}));
+
+	const result = await aggregateProxy.createDataTable({
+		name: tableName,
+		columns,
+	});
+
+	return [{ json: result as unknown as IDataObject }];
```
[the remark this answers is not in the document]
CharlieKolb: This compiles without the cast, please remove it.

packages/nodes-base/nodes/DataTable/actions/table/delete.operation.ts
```diff
@@ -0,0 +1,44 @@
...
+const displayOptions: IDisplayOptions = {
+	show: {
+		resource: ['table'],
+		operation: [FIELD],
+	},
+};
+
+export const description: INodeProperties[] = [
+	{
+		displayName:
```
CharlieKolb, line 22: love that

packages/nodes-base/nodes/DataTable/actions/table/list.operation.ts
```diff
@@ -0,0 +1,148 @@
...
+		name: 'limit',
+		type: 'number',
+		default: 50,
+		description: 'Max number of results to return',
+		typeOptions: {
+			minValue: 1,
+		},
+		displayOptions: {
+			show: {
+				resource: ['table'],
```
CharlieKolb: this should do `...displayOptions.show` instead

packages/nodes-base/nodes/DataTable/actions/table/list.operation.ts
```diff
@@ -0,0 +1,148 @@
...
+		],
+	},
+];
+
+export async function execute(
+	this: IExecuteFunctions,
+	index: number,
+): Promise<INodeExecutionData[]> {
+	const returnAll = this.getNodeParameter('returnAll', index) as boolean;
+	const limit = this.getNodeParameter('limit', index, 50) as number;
```
CharlieKolb: [nit] we have a constants file for these https://github.com/n8n-io/n8n/blob/a2aca5e7a498dfddcde4bd1f2ba29a4fbf98b560/packages/nodes-base/nodes/DataTable/common/constants.ts#L6

packages/nodes-base/nodes/DataTable/actions/table/list.operation.ts
```diff
@@ -0,0 +1,148 @@
...
+		sortDirection?: 'asc' | 'desc';
+	};
+
+	const aggregateProxy = await getDataTableAggregateProxy(this);
+
+	const queryOptions: ListDataTableOptions = {};
+
+	if (options.sortField && options.sortDirection) {
+		queryOptions.sortBy =
+			`${options.sortField}:${options.sortDirection}` as ListDataTableOptions['sortBy'];
```
CharlieKolb, line 107: We can get rid of this as by doing

```
export type ListDataTableOptionsSortByKey = 'name' | 'createdAt' | 'updatedAt' | 'sizeBytes';

export type ListDataTableOptions = {
	filter?: Record<string, string | string[]>;
	sortBy?: `${ListDataTableOptionsSortByKey}:asc` | `${ListDataTableOptionsSortByKey}:desc`;
	take?: number;
	skip?: number;
};
```

in https://github.com/n8n-io/n8n/blob/c55f67e4615e1c6ff03d609923754ef42e9c6507/packages/workflow/src/data-table.types.ts#L31

and then using that instead of string above. We can then also do `satisfies { name: string, value: ListDataTableOptionsSortByKey }` for the option definition above which is valuable.

It's overkill for this, but probably nice to have in general :D

CharlieKolb (commented):
Lots of nitpicks - please at least get rid of the `as unknown` casts, those are really dangerous

packages/nodes-base/nodes/DataTable/actions/table/create.operation.ts
RicardoE105: I think this was discussed here https://n8nio.slack.com/archives/C04CPHD7VDW/p1765266787830699?thread_ts=1765232200.466179&cid=C04CPHD7VDW

CharlieKolb (approved):

PR #22964, opened 2025-12-09
fix(core): Don't resume parent workflow when the child workflow goes into waiting

mfsiega:
## Summary

A child workflow resolves its execution promise when it completes *or* goes into `waiting`. In the latter case, we shouldn't resume the parent.

https://linear.app/n8n/issue/CAT-1929/

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/__tests__/wait-tracker.test.ts
+++ b/packages/cli/src/__tests__/wait-tracker.test.ts
@@ -487,6 +487,41 @@ describe('WaitTracker', () => {
 					childExecution.id,
 				);
 			});
+
+			it('should not resume parent execution when child execution status is "waiting"', async () => {
+				// ARRANGE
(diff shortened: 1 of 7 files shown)


ivov:
Thanks for the quikc fix! Looks like some e2e tests need ironing out. Unassigning until ready for review

mfsiega:
> Thanks for the quikc fix! Looks like some e2e tests need ironing out. Unassigning until ready for review

@ivov tests are passing - for what it's worth they were always passing locally, I think they've become very flaky recently. I don't _think_ this change made them worse (except that it added another test with the possibility to flake).

ivov (dismissed):
Tried to break it in regular and scaling mode, with and without offloading, all cases good, nice work!

ivov (approved):
🚀

PR #22974, opened 2025-12-09
fix: Add version history records when importing workflows

dariacodes:
## Summary

This PR adds workflow history records during workflow import. If it's an existing workflow with an existing record, it updates it.

This was causing migration problems for users when adding `activeVersionId`, so I also updated the migration to backfill workflow history again (same way we did it in `BackfillMissingWorkflowHistoryRecords1762763704614`).

To test it, I had to temporarily enable `PRAGMA foreign_keys = ON` in the migration runner to catch the FK constraint violation during testing.

Closes [ADO-4517](https://linear.app/n8n/issue/ADO-4517/db-migration-on-update-to-11225-fails-when-using-source-control) and [ADO-4520](https://linear.app/n8n/issue/ADO-4520/workflow-cannot-be-activated-in-prod-when-pushed-from-dev-env-via-git)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
+++ b/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
@@ -21,13 +21,17 @@ import {
 	WorkflowRepository,
 	WorkflowTagMappingRepository,
 	WorkflowPublishHistoryRepository,
+	WorkflowHistoryRepository,
 } from '@n8n/db';
 import { Service } from '@n8n/di';
 import { PROJECT_OWNER_ROLE_SLUG } from '@n8n/permissions';
 // eslint-disable-next-line n8n-local-rules/misplaced-n8n-typeorm-import
 import { In } from '@n8n/typeorm';
+import { QueryDeepPartialEntity } from '@n8n/typeorm/query-builder/QueryPartialEntity';
 import glob from 'fast-glob';
+import isEqual from 'lodash/isEqual';
 import { Credentials, ErrorReporter, InstanceSettings } from 'n8n-core';
+import type { IWorkflowBase } from 'n8n-workflow';
 import { ensureError, jsonParse, UnexpectedError, UserError } from 'n8n-workflow';
 import { readFile as fsReadFile } from 'node:fs/promises';
 import path from 'path';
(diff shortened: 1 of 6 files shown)


packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -1263,4 +1273,80 @@ export class SourceControlImportService {
...
+
+				if (nodesChanged || connectionsChanged) {
+					this.logger.debug(
+						`Updating workflow history for versionId ${importedWorkflow.versionId}`,
+					);
+
+					// Fetch user for author info
+					const user = await this.userRepository.findOne({ where: { id: userId } });
+
+					await this.workflowHistoryService.updateVersion(
```
guillaumejacquart, line 1313: In that case, should we maybe create a new version (not sure)

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -154,6 +159,8 @@ export class SourceControlImportService {
 		instanceSettings: InstanceSettings,
 		private readonly sourceControlScopedService: SourceControlScopedService,
 		private readonly workflowPublishHistoryRepository: WorkflowPublishHistoryRepository,
+		private readonly workflowHistoryService: WorkflowHistoryService,
```
guillaumejacquart, line 161: Just wondering if we could have only one of those dependencies (ideally the service I think)

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
guillaumejacquart: ah no, it's because the source version has changed, and we want to keep it the same version in prod as well ?

guillaumejacquart (commented):
minor comments

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
dariacodes: If we do create a new version, then we need to update it on workflow as well. To be honest, I'm not even sure this case is possible.
Would you prefer to just always create a new version?

guillaumejacquart: No, i think it's fine to keep it I guess

dariacodes: Alright. We can always change this later if we want to

guillaumejacquart (approved):
Looks good, thanks !

PR #22986, opened 2025-12-09
fix: Reduce OAuth authorization request by excluding shared projects

ireneea:
## Summary

Reduce the `GET /oauth2-credential/auth` and `GET /oauth1-credential/auth` request size by excluding the sharedProjects from the query parameter.

The `sharedProjects` property was not needed anyway and the values were sent as a serialized `"[Object, Object]"` string anyway.  

fixes PAY-4267

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
+++ b/packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
@@ -1020,7 +1020,10 @@ async function oAuthCredentialAuthorize() {
 	const types = parentTypes.value;
 
 	try {
-		const credData = { id: credential.id, ...credentialData.value };
+		// We exclude sharedWithProjects because it's not needed for the authorization and it causes the request to be too large
+		const { sharedWithProjects, ...sanitizedCredData } = credentialData.value;
+		const credData = { id: credential.id, ...sanitizedCredData };
+
 		if (credentialTypeName.value === 'oAuth2Api' || types.includes('oAuth2Api')) {
 			if (isValidCredentialResponse(credData)) {
 				url = await credentialsStore.oAuth2Authorize(credData);


packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -1020,7 +1020,10 @@ async function oAuthCredentialAuthorize() {
 	const types = parentTypes.value;
 
 	try {
-		const credData = { id: credential.id, ...credentialData.value };
+		// We exclude sharedWithProjects because it's not needed for the authorization and it causes the request to be too large
+		const { sharedWithProjects, ...sanitizedCredData } = credentialData.value;
+		const credData = { id: credential.id, ...sanitizedCredData };
```
ireneea, line 1025: I just focused on fixing `431 Request Header Fields Too Large` issue with this. But I suspect that most of the properties we sent in the request query are not needed.

Do you guys think it's worth revisiting this and be more explicit about the query data as part of a separate ticket?

konstantintieber: It's unfortunate that credentialData doesn't have a well defined type but seems to be able to contain any key value pairs.

I tried to trace back where this spread of `credentialData.value` was added to get more info as to why we send all of it, but it was added in a large PR from 4 years ago: https://github.com/n8n-io/n8n/pull/2154

ireneea: Yeah and I suspect that at the time credential data had a very limited set of props

konstantintieber: Quickly looking at the backend code, I only see it requiring the credential ID to return the credential: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/oauth/oauth.service.ts#L82

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -1020,7 +1020,10 @@ async function oAuthCredentialAuthorize() {
 	const types = parentTypes.value;
 
 	try {
-		const credData = { id: credential.id, ...credentialData.value };
+		// We exclude sharedWithProjects because it's not needed for the authorization and it causes the request to be too large
+		const { sharedWithProjects, ...sanitizedCredData } = credentialData.value;
```
konstantintieber, line 1024: Would be nice to have a unit test for this sanitization logic.

konstantintieber (approved):

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
ireneea: That's my assumption  as well but didn't want to make that change here in case there are some unforseen side effect to removing that data from the query

konstantintieber: Yeah, your fix is good to remedy the problem quickly without having to dig too deep right away 👍

PR #22988, opened 2025-12-09
feat(Google Gemini Node): Ability to create a file store and upload files to it

DawidMyslak:
## Summary

In this PR we're introducing new operations related to File Search store in Gemini node.

**Loom 3 (LATEST)**: [Google Gemini File Search](https://www.loom.com/share/4d6f0f9c03c84ea58b2603a08bc4fd22)

**Loom 1 (OLD)**: [Google Gemini File Search - Upload file via URL](https://www.loom.com/share/07e516c6f1e54d3f8ff924328112278f)
**Loom 2 (OLD)**: [Google Gemini File Search - Upload file via form](https://www.loom.com/share/604be0acef054374b76b2ee39da9e614)

https://linear.app/n8n/issue/NODE-4027/gemini-node-add-ability-to-create-a-file-store-and-upload-files-to-it

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
```diff
@@ -165,3 +233,73 @@ export async function transferFile(
...
+
+	const operationName = uploadResponse.body.name;
+	let operation = (await apiRequest.call(
+		this,
+		'GET',
+		`/v1beta/${operationName}`,
+	)) as FileSearchOperation;
+
+	while (!operation.done) {
+		await new Promise((resolve) => setTimeout(resolve, 3000));
```
DawidMyslak: File indexing might take some time, I didn't find clear recommendations in Google API docs how long should be the interval.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/actions/fileSearch/listStores.operation.ts
```diff
@@ -0,0 +1,49 @@
...
+		default: 10,
+		typeOptions: {
+			minValue: 1,
+			maxValue: 20,
+		},
+	},
+	{
+		displayName: 'Page Token',
+		name: 'pageToken',
+		type: 'string',
```
DawidMyslak, line 22: LOL? This error is here because the field is named `name: 'pageToken'`... I think it's because it contains "token"?

DawidMyslak: I might just drop it? When I think about it now, I don't expect users to need a pagination token inside the workflow?

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
```diff
@@ -88,48 +105,65 @@ export async function uploadFile(this: IExecuteFunctions, fileContent: Buffer, m
...
+		return {
+			stream: downloadResponse.data as Stream,
+			mimeType,
+		};
 	}
 
-	const uploadInitResponse = (await apiRequest.call(this, 'POST', '/upload/v1beta/files', {
+	const binaryPropertyName = this.getNodeParameter('binaryPropertyName', i, 'data');
+	if (!binaryPropertyName) {
+		throw new NodeOperationError(this.getNode(), 'Binary property name is required', {
```
dlavrenuek: Optional suggestion: I can see that the same error was already used here but the error message is not fully correct. Since there is no dedicated check if `downloadUrl` is an empty string, the error here should state `Binary property name or download URL is required`. 

```suggestion
		throw new NodeOperationError(this.getNode(), 'Binary property name or download URL is required', {
```

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
```diff
@@ -165,3 +225,100 @@ export async function transferFile(
 
 	return { fileUri: file.uri, mimeType: file.mimeType };
 }
+
+export async function createFileSearchStore(
+	this: IExecuteFunctions,
+	displayName: string,
+): Promise<IDataObject> {
```
dlavrenuek: Response type declaration should not be needed here since the response is casted as a specific type.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
dlavrenuek: 3 seconds is a long time and it also means that the node will take at least 3 seconds for any document that is not chunked immediately even if the operation needed 100ms. In other places in the same file 1 second is used as timeout. Also since 1000 is used multiple times in this file (assuming that here 3000 will be changed to 1000), I suggest defining a variable for it, for example `OPERATION_CHECK_INTERVAL` or similar

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
```diff
@@ -165,3 +225,100 @@ export async function transferFile(
...
+	}
+
+	const uploadResponse = (await uploadStream.call(this, stream, {
+		endpoint: `/upload/v1beta/${fileSearchStoreName}:uploadToFileSearchStore`,
+		mimeType,
+		body: { displayName, mimeType },
+	})) as { body: { name: string } };
+
+	const operationName = uploadResponse.body.name;
+	let operation = (await apiRequest.call(
```
dlavrenuek, line 276: There are a lot of type casts, here it would be better to define the type like

```ts
let operation: FileSearchOperation = await apiRequest.call(...);
```

This also removes the need to cast in https://github.com/n8n-io/n8n/pull/22988/files#diff-600f79f07af7d7d6a14c5c378f09e1c1b39b11900b8b833f48208fb687263347R285

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
```diff
@@ -165,3 +225,100 @@ export async function transferFile(
...
+): Promise<IDataObject> {
+	const qs: IDataObject = {};
+	if (pageSize !== undefined) {
+		qs.pageSize = pageSize;
+	}
+	if (pageToken) {
+		qs.pageToken = pageToken;
+	}
+
+	return (await apiRequest.call(this, 'GET', '/v1beta/fileSearchStores', { qs })) as IDataObject;
```
dlavrenuek, line 313: There is actually no need to await or cast here since the return value is already defined as `Promise<IDataObject>`

```suggestion
	return apiRequest.call(this, 'GET', '/v1beta/fileSearchStores', { qs });
```

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
```diff
@@ -165,3 +225,100 @@ export async function transferFile(
...
+	this: IExecuteFunctions,
+	name: string,
+	force?: boolean,
+): Promise<IDataObject> {
+	const qs: IDataObject = {};
+	if (force !== undefined) {
+		qs.force = force;
+	}
+
+	return (await apiRequest.call(this, 'DELETE', `/v1beta/${name}`, { qs })) as IDataObject;
```
dlavrenuek, line 326: Also here
```suggestion
	return apiRequest.call(this, 'DELETE', `/v1beta/${name}`, { qs });
```

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/actions/router.ts
```diff
@@ -31,6 +32,9 @@ export async function router(this: IExecuteFunctions) {
 		case 'file':
 			execute = file[googleGeminiTypeData.operation].execute;
 			break;
+		case 'fileSearch':
+			execute = fileSearch[googleGeminiTypeData.operation].execute;
```
dlavrenuek, line 36: Just a comment and there is no need for refactoring because the structure was already there before your changes: I was wondering why such a long syntax `googleGeminiTypeData.operation` was used multiple times here, it would be shorter to define following on line 18:

```ts
	const {
		resource,
		operation,
	} = {
		resource: this.getNodeParameter('resource', 0),
		operation: this.getNodeParameter('operation', 0),
	} as GoogleGeminiType;
```

instead of

```ts
	const resource = this.getNodeParameter('resource', 0);
	const operation = this.getNodeParameter('operation', 0);
	const googleGeminiTypeData = {
		resource,
		operation,
	} as GoogleGeminiType;
```

and then do

```ts
switch (resource) {
...
  execute = audio[operation].execute;
...
  execute = file[operation].execute;
```

dlavrenuek (changes requested):
Nice work, I did test the implementation successfully but not with all file types. Uploading an image did result in a 500 error from google gemini api, which looks like an error on their side and not an implementation problem here.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
dlavrenuek: Or the cast is not needed

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: We use this exact pattern in other nodes inside `nodes-langchain` - that's why I followed it.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: Bumped.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: We use this exact pattern in other nodes inside `nodes-langchain` - that's why I followed it.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: Hmm 🤔 , but `apiRequest.call` won't guarantee me the type - it returns `any`.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/actions/router.ts
DawidMyslak: Fair point. I guess we could adopt it when writing new nodes.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: Applied.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: Removed the type declaration. Cast is needed.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
dlavrenuek: That's why you have to define the type of the variable `let operation: FileSearchOperation` In the end it does the same but without having to type cast in 2 places.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
dlavrenuek: What do you think about using a constant for all of the 1000?

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
dlavrenuek: The existing pattern is fine for me but on the other hand also does not mean that we must keep it if there is room for improvement. I have experienced that often engineers don't realize that

```js
async function do() {
    return await request();
}
```

is the same as

```js
const do = () => request();
```

Given `request()` returns a promise. But it's also arguable if the second one is better :D Some people find arrow functions (justifiably) confusing, which gets more confusing when they are chained together.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: This is what I get when I try it:
Image

dlavrenuek: Interesting, definitely now what I see :D Also I think that I did try the changes with successful pnpm build, but may be I am just imagining it. May be it's because of the ()? Anyway it's not an issue worth investing time in, so just ignore my comment and let's move on

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/helpers/utils.ts
DawidMyslak: Added `OPERATION_CHECK_INTERVAL`.

dlavrenuek (approved):
Thank you for the implementation 🚀  and for addressing my picky comments :D

PR #7597, opened 2023-11-03
fix(editor): Hide not supported node options

janober:
Currently Trigger-Nodes and Sub-Nodes display options that do not make sense or work like:
- Always Output Data
- Execute Once
- On Error

This PR hides those options for nodes that can not make use of them anyway to not confuse users.

Github issue / Community forum post (link here to close automatically):

PR #23009, opened 2025-12-10
fix(editor): Show node settings in protected view for read-only review

r00gm:
## Summary

Previously, the `createCommonNodeSettings` function took an `isExecutable` parameter.
In protected view (source control environment), `isExecutable` is false, which caused the node settings (like "Always Output Data", "Execute Once", "Retry on Fail", etc.) to be hidden entirely.

Node settings will now be visible in protected view (read-only state). The settings themselves will still be read-only because other parts of the UI handle the disabled/read-only state - this change just ensures the settings are displayed so users can review what's configured.

[PAY-2519](https://linear.app/n8n/issue/PAY-2519/node-settings-missing-in-protected-view)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/ndv/settings/components/NodeSettings.vue
+++ b/packages/frontend/editor-ui/src/features/ndv/settings/components/NodeSettings.vue
@@ -457,7 +457,7 @@ const populateHiddenIssuesSet = () => {
 };
 
 const nodeSettings = computed(() =>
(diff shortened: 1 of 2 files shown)


autologie (approved):
I'm not familiar with the reason why the condition exists, looks like it has been so since https://github.com/n8n-io/n8n/pull/7597 But anyway this will solve the linked issue.

PR #23024, opened 2025-12-10
fix: Fix checksum hashing when WebCrypto is unavailable

dariacodes:
## Summary

We've been experiencing problems with calling WebCrypto `crypto.subtle` in insecure contexts (http). To make sure the fix is a proper one, I moved all `calculateWorkflowChecksum` calls to the server, as we need to update it anyway when the data is updated, so it makes sense to not call it on the client. At the same time, I updated the function to fall back to a pure-JS SHA-256 when WebCrypto is unavailable.

To test this:
1. Start the app locally (HTTP).
`N8N_SECURE_COOKIE=false pnpm dev`
2. Find your LAN IP (e.g. `ifconfig | grep "inet " | grep -v 127.0.0.1`).
3. Open `http://<LAN_IP>:<port>` (e.g. `http://192.168.1.23:5678`)
4. See no error anymore

Closes [ADO-4527](https://linear.app/n8n/issue/ADO-4527/community-issue-saving-workflows-fails-in-n8n-20-when-running-over) and [GHC-5843](https://linear.app/n8n/issue/GHC-5843/community-issue-200-could-not-find-workflow-cannot-read-properties-of)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
@@ -22,7 +22,7 @@ import {
 	N8nTooltip,
 } from '@n8n/design-system';
 import type { WorkflowSettings } from 'n8n-workflow';
-import { calculateWorkflowChecksum, deepCopy } from 'n8n-workflow';
+import { deepCopy } from 'n8n-workflow';
 import { useSettingsStore } from '@/app/stores/settings.store';
 import { useRootStore } from '@n8n/stores/useRootStore';
 import { useWorkflowsEEStore } from '@/app/stores/workflows.ee.store';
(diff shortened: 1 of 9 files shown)


packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
```diff
@@ -483,7 +483,9 @@ const saveSettings = async () => {
 	try {
 		const workflowData = await workflowsStore.updateWorkflow(String(route.params.name), data);
 		workflowsStore.setWorkflowVersionId(workflowData.versionId);
-		workflowsStore.setWorkflowChecksum(await calculateWorkflowChecksum(workflowData));
+		if (workflowData.checksum) {
```
scdekov, line 486: Should we rather throw here(and on the other files where we have the same check) if there's no checksum? It seems like updateWorkflow request should always return it so keeping the code like this may hide a bug.

scdekov (approved):
I'm thinking if we should have the checksum as an extra parameter to the other update methods eg to `workflowStore.setWorkflowVersionId` rather than a separate setter. Updating props like the versionId, the active version etc that are parameters of the checksum should always be done in combination with updating the checksum so keeping this in a single function may be cleaner and less error prone. If we want to do this it should probably be separate of this fix though.

PR #23029, opened 2025-12-10
fix(core): Resolve file path correctly when serving node icons from N8N_CUSTOM_EXTENSIONS

maspio:
## Summary - [Fixes NODE-3843](https://linear.app/n8n/issue/NODE-3843/community-issue-bug-custom-node-icon-path-includes-absolute-container) 

Fix node icons not loading when using "N8N_CUSTOM_EXTENSIONS" to load custom nodes from directories other than "/home/node/.n8n/custom".

```bash
# Test docker image with PR fixes
docker pull ghcr.io/n8n-io/n8n:branch-resolve-filePath-correctly-when-serving-node-icons
# For amd64 architectures
docker pull --platform linux/amd64 ghcr.io/n8n-io/n8n:branch-resolve-filePath-correctly-when-serving-node-icons
````

### Related Github Issues

* [Bug: Custom Node icon path includes absolute container path (/icons/CUSTOM//home/… Icon Not Loading/Broken #21360](https://github.com/n8n-io/n8n/issues/21360)
* [When placing custom nodes into the ~/.n8n/custom folder, the node icon cannot be displayed. #22452](https://github.com/n8n-io/n8n/issues/22452)
### Context

Resolves the node icon file path when loaded from `/icons/${packageName}/${iconPath}`.

### Problem

Using `N8N_CUSTOM_EXTENSIONS`, nodes can be loaded from any directory outside of `CWD='$N8N_USER_FOLDER/.n8n/'`.

Custom nodes are loaded by `custom-directory-loader.ts` using an absolute path, different from the default `package-directory-loader.ts`. The icon loading logic for custom nodes seems a bit broken, because icons are resolved by absolute paths encoded in URLs.

**Examples when served from `/icons/${packageName}/${iconPath}`:**
- `/icons/CUSTOM//home/node/.n8n-custom`
- `/icons/CUSTOM/C:/User/name/.n8n-custom`

### Implementation

`resolveIcon()` has a special `path.resolve()` strategy for custom nodes considering:
- An absolute Linux file path is encoded in the URL using `//`
- `//` in URLs can be normalized to `/` by proxies, load balancers, etc.
- A Windows file path starts with drive letters like `C:` not `/`

### Future Improvements

> [!NOTE]
> Instead of fixing the broken custom node loading strategy in `resolveIcon()`, make `custom-directory-loader.ts` also use relative paths. Besides having different icon loading strategies, encoding an absolute path in URLs seems a security risk.

## Review / Merge checklist

- [ ] Tests included. 

### CODE REVIEW

File: packages/cli/src/security-audit/risk-reporters/nodes-risk-reporter.ts [hunk 1]
```diff
@@ -1,5 +1,6 @@
 import { Service } from '@n8n/di';
 import glob from 'fast-glob';
+import { CUSTOM_NODES_PACKAGE_NAME } from 'n8n-core/src/nodes-loader/constants';
```
RomanDavydchuk: This import doesn't work and causes n8n to not start up. There are a couple of ways to fix this:
1. change the import to `n8n-core/dist/nodes-loader/constants`
2. make this new constant exported straight from the `n8n-core` package
3. do not import this constant in the `cli` package at all

packages/cli/src/load-nodes-and-credentials.ts
```diff
@@ -16,6 +17,7 @@ import {
 	UnrecognizedNodeTypeError,
 	ExecutionContextHookRegistry,
 } from 'n8n-core';
+import { CUSTOM_NODES_PACKAGE_NAME } from 'n8n-core/src/nodes-loader/constants';
```
RomanDavydchuk: Same here

packages/cli/src/__tests__/load-nodes-and-credentials.test.ts
```diff
@@ -3,6 +3,7 @@ import watcher from '@parcel/watcher';
 import fs from 'fs/promises';
 import { mock } from 'jest-mock-extended';
 import type { DirectoryLoader } from 'n8n-core';
+import { CUSTOM_NODES_PACKAGE_NAME } from 'n8n-core/src/nodes-loader/constants';
```
RomanDavydchuk: Same here

packages/@n8n/utils/src/files/path.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Fast check if file path starts with a windows drive letter, e.g. 'C:/' or 'C:\\'
+ */
+export function isWindowsFilePath(str: string) {
+	return /^([a-zA-ZA-Z]:[/|\\])/.test(str);
```
RomanDavydchuk: A couple of comments about this regex:
1. It would also return `true` for strings like `C:|`. You don't need the pipe symbol `|` between the slashes
2. `a-zA-ZA-Z` part has duplication: `a-zA-Z` would be enough
3. I don't think the capture group is necessary

I'd suggest using this regex instead:
`/^[a-zA-Z]:[\\/]/`

packages/cli/src/load-nodes-and-credentials.ts
```diff
@@ -152,13 +154,43 @@ export class LoadNodesAndCredentials {
...
 		}
+
+		const resolvePath = (iconPath: string) => {
+			return path.resolve(loader.directory, iconPath);
+		};
+
+		const resolvePathCustom = (path: string) => {
+			if (isWindowsFilePath(path)) return path;
+			return path.startsWith('/') ? path : '/' + path;
+		};
```
RomanDavydchuk, line 189: nit: I think it's better to move these out of the `resolveIcon` function, also not sure if it's necessary to extract `resolvePath` function

RomanDavydchuk (changes requested):
Left some comments. The import issue causes n8n to not start up at all (observed the same error on the image provided in PR description), but after resolving it locally and testing manually it seems to work as expected

packages/@n8n/utils/src/files/path.ts
maspio: haha right... that's what happens when you don't double check AI

packages/cli/src/load-nodes-and-credentials.ts
maspio: Yeah not ideal but I created these only because it's hard to understand why this code needs to exist at all. I left these within the function scope on purpose, because they only fix the issue within and should be removed once the root cause is fixed, left as todo comment:

https://github.com/n8n-io/n8n/blob/ca86648a4b32de940a1bd2aba696ed138bad7889/packages/cli/src/load-nodes-and-credentials.ts#L172

Anyway, that's why I did it that way. If you want to change it, feel free. I would keep it that way.

RomanDavydchuk: Makes sense, thanks for the explanation, I'll leave it as-is

RomanDavydchuk (approved):
LGTM

PR #23032, opened 2025-12-10
feat(core, editor): Better distinguish personal and workflow agents in chat (no-changelog)

autologie:
## Summary

This PR improves custom agents in ChatHub by introducing icons and reorganizing pages and model selector.

You can select icon or emoji when creating a personal agent. For workflow agent, icon/emoji is taken from the project it belongs to.
Screenshot 2025-12-11 at 15 48 37

Personal agents and workflow agents are distinguished by different page and menu
Screenshot 2025-12-11 at 15 49 47

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/api-types/src/chat-hub.ts
```diff
@@ -305,7 +321,6 @@ export class ChatHubSendMessageRequest extends Z.class({
 	),
 	tools: z.array(INodeSchema),
 	attachments: z.array(chatAttachmentSchema),
-	agentName: z.string().optional(),
```
autologie, line 309: ~~Instead of adding `agentIcon` here, this PR implements server-side logic to resolve name and icon from workflowId/agentId for better data integrity.~~

Edit: this is not the case anymore. agentName still sent from FE.

packages/frontend/@n8n/design-system/src/components/N8nNavigationDropdown/NavigationDropdown.vue
```diff
@@ -202,12 +202,6 @@ defineExpose({
 			}
 		}
 	}
-
-	& hr {
```
autologie, line 206: This didn't seem to match

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatAgentCard.vue
```diff
@@ -22,14 +23,17 @@ const i18n = useI18n();
...
-	];
+	return agent.model.provider === 'custom-agent'
+		? [{ id: 'delete' as const, label: i18n.baseText('chatHub.agent.card.menu.delete') }]
+		: [];
 });
 
+const canEdit = computed(
+	() =>
+		agent.model.provider === 'custom-agent' ||
+		hasPermission(['rbac'], { rbac: { scope: ['workflow:read'] } }),
```
autologie, line 34: Don't show edit button if the user is a chat user

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatSidebarContent.vue
```diff
@@ -111,15 +115,15 @@ useIntersectionObserver(
 	loadMoreTrigger,
 	([{ isIntersecting }]) => {
 		if (isIntersecting) {
-			void chatStore.fetchMoreSessions();
+			void chatStore.fetchMoreSessions({ minLoadingTime: 250 });
```
autologie, line 118: minLoadingTime because skeleton UI disappearing too quickly doesn't look nice

packages/@n8n/db/src/migrations/common/1765361177092-AddIconToAgentTable.ts
```diff
@@ -0,0 +1,19 @@
...
+const agentsTableName = 'chat_hub_agents';
+const sessionsTableName = 'chat_hub_sessions';
+
+export class AddIconToAgentTable1765361177092 implements ReversibleMigration {
+	async up({ schemaBuilder: { addColumns, column } }: MigrationContext) {
+		// Add icon column to agents table (nullable)
+		await addColumns(agentsTableName, [column('icon').json]);
+
+		// Add agentIcon column to sessions table (nullable)
+		await addColumns(sessionsTableName, [column('agentIcon').json]);
```
Cadiac: I'm questioning duplicating this field on the sessions table a bit - while I get that this can be handy can we just do a join?

autologie: For personal agents it should be straightforward, for workflow agents it's more tricky because the icon is actually in the project table. But we can also make another query and merge data in service.

Cadiac: ah right true.
probably worth keeping it here then, we can maybe drop this column later on if we feel like normalizing data.

packages/cli/src/modules/chat-hub/chat-hub.models.service.ts
```diff
@@ -728,15 +728,36 @@ export class ChatHubModelsService {
...
+			activeWorkflows.map((workflow) => workflow.id),
+		);
+	}
+
+	async fetchAgentWorkflowsAsModelsByIds(ids: string[]): Promise<ChatModelDto[]> {
 		const workflows = await this.workflowRepository.find({
-			select: { id: true, name: true },
-			where: { id: In(activeWorkflows.map((workflow) => workflow.id)) },
-			relations: { activeVersion: true },
+			select: {
```
Cadiac, line 736: Project icon on workflow agents feels unexpected to me, was that what Rob suggested?

autologie: Yes, we discussed adding new parameter on chat trigger but the icon/emoji component isn't available as parameter input.

Cadiac: Yeah, we'd have to add a new parameter type to support that nicely.
Okay, I guess this doesnt have to be a permanent solution..

packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
```diff
@@ -371,3 +369,13 @@ export function createMimeTypes(modalities: ChatHubInputModality[]): string {
...
 }
+
+export const personalAgentDefaultIcon: AgentIconOrEmoji = {
+	type: 'icon',
+	value: 'message-square' satisfies IconName,
+};
+
+export const workflowAgentDefaultIcon: AgentIconOrEmoji = {
+	type: 'icon',
+	value: 'bot' satisfies IconName,
```
Cadiac, line 384: I'm not a big fan of seeing just tons of `bot` icons on my history as the default instead of the `N8nAvatar` components :l I have a feeling that most WFs are going to be missing the project icon in real production.

autologie: You'll see tons of layer icons instead :)
Screenshot 2025-12-11 at 16 44 54

But how it will look by default is a good point, let me check with Rob

Cadiac: ah right, that was the default icon actually :D

packages/cli/src/modules/chat-hub/__tests__/chat-hub.service.integration.test.ts
```diff
@@ -733,4 +744,135 @@ describe('chatHub', () => {
...
+				provider: 'openai',
+				model: 'gpt-4',
+				tools: [],
+			});
+
+			// Set project icon for the member's personal project
+			const projectRepository = Container.get(ProjectRepository);
+			const personalProject = await projectRepository.getPersonalProjectForUserOrFail(member.id);
+			const projectIcon = { type: 'emoji' as const, value: '🚀' };
+			await projectRepository.update(personalProject.id, { icon: projectIcon });
```
Cadiac: I guess this test isn't fully realistic, as personal project's don't allow you to set the project icon?

Doesn't really matter though

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatAgentCard.vue
Cadiac: We should probably check for the actual permission editing an agent requires,

```
hasPermission(['rbac'], { rbac: { scope: ['chatHubAgent:update'] } })
```

There's also `hasRole(['global:chatUser'])` but the actual agent update permission seems best here.

Cadiac: Or wait, are chat users not allowed to edit the personal agents they create..? That doesn't seem right, they have this permission to do so?

autologie: The 2nd line is intended for workflow agent, and I assumed chat users can edit their agents (1st line)

Cadiac: Right, my bad. Got confused just reading the code.

`workflow:read` makes sense here, could also be `workflow:update`.

packages/@n8n/db/src/migrations/common/1765361177092-AddIconToAgentTable.ts
autologie: "Just do a join" worked. Well, not a join but some joins. And it also solves the problem of icons in sessions table get out of sync when it is changed after the conversation. Thank you for the suggestion. [a061d6b](https://github.com/n8n-io/n8n/pull/23032/commits/a061d6b890a700e9492e605cb8d91b292aef7c71)

packages/cli/src/modules/chat-hub/__tests__/chat-hub.service.integration.test.ts
autologie: That's true, let me try if I can make it more realistic without too complicated setups.

packages/@n8n/db/src/migrations/common/1765361177092-AddIconToAgentTable.ts
```diff
@@ -1,19 +1,49 @@
...
+		const escapedTableName = escape.tableName(tableName);
+		const escapedIdColumn = escape.columnName('id');
+
 		// Add icon column to agents table (nullable)
-		await addColumns(agentsTableName, [column('icon').json]);
+		await addColumns(tableName, [column('icon').json]);
 
-		// Add agentIcon column to sessions table (nullable)
-		await addColumns(sessionsTableName, [column('agentIcon').json]);
+		// Change agents.id from UUID to varchar to match foreign key references
```
Cadiac: Right, we have a mix of uuid and varchar fields.
Should we change the fields referencing the agents.id from `varchar(36)` -> `uuid` instead?

Also it looks like we're missing actual foreign key from `chat_hub_sessions.agentId` -> `chat_hub_agents.id` currently?

autologie: Thanks, changed to UUID and added FK constraint [6fdb3b0](https://github.com/n8n-io/n8n/pull/23032/commits/6fdb3b0b98ad455f9480b1298147feb303dc9717)

packages/@n8n/db/src/migrations/common/1765361177092-AddIconToAgentTable.ts
```diff
@@ -1,49 +1,53 @@
...
-			await runQuery(`ALTER TABLE ${escapedTableName} MODIFY COLUMN \`id\` VARCHAR(255)`);
 		}
+
+		// Clean up orphaned agentId references before adding foreign key constraint
+		await runQuery(
+			`UPDATE ${escape.tableName(table.sessions)} SET "agentId" = NULL WHERE "agentId" IS NOT NULL AND "agentId" NOT IN (SELECT id FROM ${escape.tableName(table.agents)})`,
+		);
+
+		// Add foreign key constraint for agentId in sessions table
+		await addForeignKey(table.sessions, 'agentId', [table.agents, 'id'], undefined, 'SET NULL');
```
Cadiac: Image

There's also agentId on messages, typed `varchar` and similarly missing the FK.
Do we want to fix that now or leave it as is? I don't think we really use that at the moment, but would be nice to have that properly set up?

Cadiac (commented):
Otherwise this appears to work and I'm happy with it, but we could consider fixing that other FK issue too now - or just do it in the future, it's not important for monday's release.

packages/@n8n/db/src/migrations/common/1765361177092-AddIconToAgentTable.ts
autologie: Done [9e4a2c0](https://github.com/n8n-io/n8n/pull/23032/commits/9e4a2c0a907884d347441cc8eea598cbbfc2a7cc)

packages/@n8n/db/src/migrations/common/1765361177092-AddIconToAgentTable.ts
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	async down({
+		schemaBuilder: { dropColumns, dropForeignKey },
+		runQuery,
+		isPostgres,
+		escape,
+	}: MigrationContext) {
+		// Drop foreign key constraints
+		await dropForeignKey(table.sessions, 'agentId', [table.agents, 'id']);
```
Cadiac: This doesn't appear to drop the FKs on sqlite correctly 🤔 

on both sessions and messages tables if I go back and forth the migration

```
pnpm start
pnpm start -- db:revert
```

more fk's accumulate. Perhaps giving a name for the FK could help?

Image

Cadiac: ...though this PR has timestamp in its name lower than `BackfillMissingWorkflowHistoryRecords1765448186933`, which is marked as `IrreversibleMigration`, so I guess we don't care much as one can't migrate down psat it anyways

Cadiac (dismissed):
Everything appears to work, so awaiting for @n8n-io/migrations-review to have a look at this and we should be good to go 👍

tomi (commented):
Could we separate the sessions table migration to a separate migration and ~~put it in the postgresdb specific migrations folder~~? Now the name of the migration is a bit misleading as it does two things

Edit: Looks like the migration does only small part for Postgres. In this case it's okay to keep them in the same "common" migration

autologie:
> Could we separate the sessions table migration to a separate migration and ~put it in the postgresdb specific migrations folder~? Now the name of the migration is a bit misleading as it does two things
> 
> Edit: Looks like the migration does only small part for Postgres. In this case it's okay to keep them in the same "common" migration

Updated the migration to only add `icon` column in this PR. We'll handle adding FK constraints in a separate PR later.

tomi (dismissed):
Only looked at the migration. LGTM 🚀

Cadiac:
> Updated the migration to only add icon column in this PR. We'll handle adding FK constraints in a separate PR later.

Today right after this one I hope?

Cadiac (dismissed):
Maybe `no-changelog` this PR as this is a new feature launching this week?

Cadiac (dismissed):

tomi (approved):

PR #23036, opened 2025-12-10
feat: Autosave workflows

dariacodes:
## Summary

- Autosave workflow with a debounce if state is dirty
- Remove manual save functionality (save button, command bar save, save shortcut)
- Introduce save animation
- Update e2e
    - remove save functionality testing where possible (ideally, we would test it separately, but many tests unfortunately rely on that)
    - Fixed several failing e2e (`tests/e2e/sharing/sharing.spec.ts` and `subworkflows/workflow-selector.spec.ts`) with the help of @shortstacked, who spotted a race condition happening during workflow import.

Closes [ADO-428](https://linear.app/n8n/issue/ADO-4280/feature-fe-autosave-workflows-on-changes)

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.test.ts
scdekov, line 650: Are those still relevant? We're now creating workflows with real ids and do not show this toast.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.vue
```diff
@@ -438,13 +439,20 @@ const handleImportWorkflowFromFile = () => {
 	}
 };
 
+const handleWorkflowSaved = (data: { isFirstSave: boolean }) => {
```
scdekov, line 422: Seems like this is not needed anymore since workflows are persisted on creation

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.vue
```diff
@@ -438,13 +439,20 @@ const handleImportWorkflowFromFile = () => {
...
+};
+
 onMounted(() => {
 	nodeViewEventBus.on('importWorkflowFromFile', handleImportWorkflowFromFile);
 	nodeViewEventBus.on('archiveWorkflow', handleArchiveWorkflow);
 	nodeViewEventBus.on('unarchiveWorkflow', handleUnarchiveWorkflow);
 	nodeViewEventBus.on('deleteWorkflow', handleDeleteWorkflow);
 	nodeViewEventBus.on('renameWorkflow', onNameToggle);
 	nodeViewEventBus.on('addTag', onTagsEditEnable);
+	canvasEventBus.on('saved:workflow', handleWorkflowSaved);
```
scdekov, line 435: probably not needed neither

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -54,14 +52,42 @@ const isWorkflowSaving = computed(() => {
...
 
+/**
+ * Cancel autosave if scheduled or wait for it to finish if in progress
+ * Save immediately if autosave idle or cancelled
+ */
+const saveBeforePublish = async () => {
+	if (autosaveStore.autoSaveState === AutoSaveState.InProgress && autosaveStore.pendingAutoSave) {
+		autoSaveForPublish.value = true;
+		try {
+			await autosaveStore.pendingAutoSave;
```
scdekov, line 114: Wouldn't that save make the state not dirty at the end which should cause the next check to not work as expected. It actually seems to be working correctly when I test it but just by reading this code it seems potentially problematic.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -54,14 +52,42 @@ const isWorkflowSaving = computed(() => {
...
+		} finally {
+			autoSaveForPublish.value = false;
+		}
+	} else if (autosaveStore.autoSaveState === AutoSaveState.Scheduled) {
+		cancelAutoSave();
+	}
+
+	if (uiStore.stateIsDirty || props.isNewWorkflow) {
```
scdekov: Lets double check if `isNewWorkflow` is ever `true` and if not simplify this? I guess it may've been forgotten when making wfs creation persistent

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -54,14 +52,42 @@ const isWorkflowSaving = computed(() => {
...
+		autoSaveForPublish.value = true;
+		try {
+			const saved = await saveCurrentWorkflow({}, true);
+			if (!saved) {
+				throw new Error('Failed to save workflow before publish');
+			}
+		} finally {
+			autoSaveForPublish.value = false;
+		}
+	} else {
```
scdekov: ✂️

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -54,14 +52,42 @@ const isWorkflowSaving = computed(() => {
...
+	} else if (autosaveStore.autoSaveState === AutoSaveState.Scheduled) {
+		cancelAutoSave();
+	}
+
+	if (uiStore.stateIsDirty || props.isNewWorkflow) {
+		autoSaveForPublish.value = true;
+		try {
+			const saved = await saveCurrentWorkflow({}, true);
+			if (!saved) {
+				throw new Error('Failed to save workflow before publish');
```
scdekov: Throwing here may not be fully correct as we're also returning saved=false when clicking cancel in the override changes modal

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
```diff
@@ -420,9 +435,49 @@ export function useWorkflowSaving({
...
+			}
+
+			autosaveStore.setAutoSaveState(AutoSaveState.InProgress);
+
+			const savePromise = (async () => {
+				try {
+					await saveCurrentWorkflow({}, true, false, true);
+				} finally {
+					if (autosaveStore.autoSaveState === AutoSaveState.InProgress) {
+						autosaveStore.setAutoSaveState(AutoSaveState.Idle);
```
scdekov: call `autosaveStore.reset()`

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.vue
dariacodes: This is needed to show the success toast when it is the first save. We persist workflows on backend only after you save it for the first time (i.d. workflow is changed)

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.vue
dariacodes: also related to the same thing

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: What do you mean by "not work as expected"? If we had a pending autosave, then we just wait for it to finish and we don't need to save again. And if it was scheduled, we just unschedule it and the state should still be dirty.

scdekov: but it's possible to have changes between the save started and clicking the publish button(which calls `saveBeforePublish`), right? In that case there will be a pending autosave that will be awaited but we'll still need another save after that, no?

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.test.ts
scdekov: ignore the above

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
scdekov: ignore the above

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: Yeah, you're right. I fixed this [here](https://github.com/n8n-io/n8n/pull/23036/commits/7f99cc407ded60422d3469919fc40942e334ce27)

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: I fixed this [here](https://github.com/n8n-io/n8n/pull/23036/commits/743af4b92a308bad70467d9db8178c950aab28a0). After this fix, if there are new unsaved changes, the UI state won't be cleaned up

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -57,10 +57,13 @@ const importFileRef = computed(() => actionsMenuRef.value?.importFileRef);
  * Save immediately if autosave idle or cancelled
  */
 const saveBeforePublish = async () => {
+	let saved = true;
```
scdekov: nit: start with `saved = false` and set it to true within the try block after `await autosaveStore.pendingAutoSave;`

scdekov: I'd even say it's bigger than a nit. It seems a little confusing that we start with saved = true and then we may not do any saving and will return saved = true. This seems to be working fine as `onPublishButtonClick` expects saved to be returned as true in order to show the modal but it feels a little confusing. I'd say we have the check if state is dirty in onPublishButtonClick and if it's not not call saveBeforePublish at all. Pretty much how it was working before.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -70,26 +73,19 @@ const saveBeforePublish = async () => {
...
 const onPublishButtonClick = async () => {
-	// If there are unsaved changes, save the workflow first
-	if (uiStore.stateIsDirty || props.isNewWorkflow) {
-		try {
-			await saveBeforePublish();
-		} catch (error) {
-			return;
-		}
+	// Save the workflow first
+	const saved = await saveBeforePublish();
```
scdekov: I think it's good to keep the condition

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
```diff
@@ -416,7 +428,10 @@ export function useWorkflowSaving({
 			}
 
 			uiStore.removeActiveAction('workflowSaving');
-			uiStore.markStateClean();
+			// Only mark state clean if no new changes were made during the save
```
scdekov, line 437: It's weird that we're doing this 2 times. I see that it was the case before so maybe we just leave a todo here to check later if we need to clean 2 times?

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: I did it so that I can remove the checks from the `onPublishButtonClick` - if state is not dirty, it means it is saved already. But sure, I can put the check back and start with `false` if it'll be less confusing

dariacodes: upd.: please check [this commit](https://github.com/n8n-io/n8n/pull/23036/commits/ab0d28c1c0d4499820dee40dd12d642cac8f030a)

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
dariacodes: Good catch! I removed the first one

scdekov: Awesome, let's just make sure we check that there wasn't a weird reason that actually required this.

scdekov (dismissed):

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
dariacodes: This code is too long, so it's quite possible that someone didn't notice the duplication. All the operations in-between are just syncing local state and it's logical to have it at the same place with other post-save operations, so I think it is a good call.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.vue
scdekov: This event doesn't seem to be emitted from WorkflowHeaderDraftPublishActions so it should be possible to drop the handler here

scdekov: @claude please handle this

RicardoE105 (approved):

RicardoE105 (dismissed):

.github/workflows/test-visual-chromatic.yml
```diff
@@ -41,7 +41,9 @@ jobs:
         !startsWith(github.event.pull_request.head.ref, 'release/') &&
         !startsWith(github.event.pull_request.head.ref, 'release-pr/')
       )
-    runs-on: blacksmith-2vcpu-ubuntu-2204
+    runs-on: blacksmith-4vcpu-ubuntu-2204
```
shortstacked, line 44: You only need one or the other. 4vcpu = 16gb ram (so we don't need the max old space size) or 2vcpu + max old space size (since it onlyu has 6gb ram).

dariacodes: Got it! I left just the cpu bump

shortstacked (approved):
LGTM: Ship it 🚀

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -80,54 +70,179 @@ const containsTrigger = computed((): boolean => {
...
+	});
+};
+
+const publishButtonConfig = computed(() => {
+	// Handle permission-denied state first
+	if (!props.workflowPermissions.publish) {
+		return {
+			text: i18n.baseText('workflows.publish'),
+			enabled: false,
+			showIndicator: false,
```
konstantintieber, line 156: I stumbled onto this change while working on https://github.com/n8n-io/n8n/pull/24111

Hardcoding `showIndicator` and `indicatorClass` to false and empty string prevents users who don't have permission to publish a workflow to see that someone else may have published this workflow.

They'd need to first click on the workflow history to see the green indicator next to the published version.

This is why I think we should adjust the logic here to something like this:
```ts
const publishButtonConfig = computed(() => {
  if (!props.workflowPermissions.publish) {
		const defaultConfigForNoPermission = {
			text: i18n.baseText('workflows.publish'),
			enabled: false,
			showIndicator: false,
			indicatorClass: '',
			tooltip: i18n.baseText('workflows.publish.permissionDenied'),
			showVersionInfo: false,
		};
		const isWorkflowPublished = workflowsStore.workflow.activeVersion;
		if (isWorkflowPublished) {
			defaultConfigForNoPermission.showIndicator = true;
			defaultConfigForNoPermission.indicatorClass = 'published';
			return {
				...defaultConfigForNoPermission,
				showIndicator: true,
				indicatorClass: 'published',
			};
		} else {
			return defaultConfigForNoPermission;
		}
	}
    /* ... */
}
```

@dariacodes @RobTF9 what do you think?

Here's a photo of what that would look like with no publish permissions (used a photo rather than a screenshot so that the hover state is visible):
image

PR #23038, opened 2025-12-10
feat(editor): Show plus button between tool and AI agent

yehorkardash:
## Summary

This PR adds a new plus button between the main node and the tool node, which shows a list of HITL nodes.

For this new logic I had to update node positioning code and add a new separate node view. It would be better to add HITL as a category for Tools subcategory, however that would require more changes to view stack logic. The downside of this approach is that there will be several different views for HITL nodes

https://github.com/user-attachments/assets/e0cfa005-e5c1-4a86-96b5-8f79071a8fb5

https://linear.app/n8n/issue/NODE-3885/add-ability-to-add-hitl-tool-in-between-agent-and-tool

## Review / Merge checklist

- [x] Tests included. 


elsmr (commented):
Works as expected 🎉 , fixed some import issues

elsmr (approved):

PR #23039, opened 2025-12-10
chore(editor): Refactor wf checksum setting (no-changelog)

scdekov:
## Summary

Workflow checksum needs to be updated on the front end with every update on workflow nodes, version, or other parameters of the checksum function. Because of that it's safer to keep the checksum update within those update operations rather than as a separate one.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
@@ -482,10 +482,10 @@ const saveSettings = async () => {
 
 	try {
 		const workflowData = await workflowsStore.updateWorkflow(String(route.params.name), data);
-		workflowsStore.setWorkflowVersionId(workflowData.versionId);
-		if (workflowData.checksum) {
-			workflowsStore.setWorkflowChecksum(workflowData.checksum);
+		if (!workflowData.checksum) {
+			throw new Error('Failed to update workflow');
 		}
+		workflowsStore.setWorkflowVersionId(workflowData.versionId, workflowData.checksum);
 	} catch (error) {
 		toast.showError(error, i18n.baseText('workflowSettings.showError.saveSettings3.title'));
 		isLoading.value = false;
--- a/packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
@@ -481,11 +481,7 @@ const saveSettings = async () => {
 	data.expectedChecksum = workflowsStore.workflowChecksum;
 
 	try {
-		const workflowData = await workflowsStore.updateWorkflow(String(route.params.name), data);
-		if (!workflowData.checksum) {
-			throw new Error('Failed to update workflow');
-		}
-		workflowsStore.setWorkflowVersionId(workflowData.versionId, workflowData.checksum);
+		await workflowsStore.updateWorkflow(String(route.params.name), data);
 	} catch (error) {
 		toast.showError(error, i18n.baseText('workflowSettings.showError.saveSettings3.title'));
 		isLoading.value = false;
--- a/packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
@@ -847,10 +847,11 @@ export function useWorkflowHelpers() {
 		}
 
 		const workflow = await workflowsStore.updateWorkflow(workflowId, data);
-		workflowsStore.setWorkflowVersionId(workflow.versionId);
-		if (workflow.checksum) {
-			workflowsStore.setWorkflowChecksum(workflow.checksum);
+		if (!workflow.checksum) {
+			throw new Error('Failed to update workflow');
 		}
+
+		workflowsStore.setWorkflowVersionId(workflow.versionId, workflow.checksum);
 
 		if (isCurrentWorkflow) {
 			workflowState.setActive(workflow.activeVersionId);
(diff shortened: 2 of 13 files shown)


packages/frontend/editor-ui/src/app/stores/workflows.store.ts
```diff
@@ -1637,6 +1628,14 @@ export const useWorkflowsStore = defineStore(STORES.WORKFLOWS, () => {
 			data as unknown as IDataObject,
 		);
 
+		if (!updatedWorkflow.checksum) {
+			throw new Error('Failed to update workflow');
+		}
+
+		if (id === workflow.value.id) {
+			workflowChecksum.value = updatedWorkflow.checksum;
```
dariacodes: Why, by the way, here we're not setting `versionId` at the same time? Can't we use `setWorkflowVersionId` as well?
I see that in many files that are using this method, we're setting the state separately, but then e.g. in `WorkflowsView` we don't need to set these. I wonder if it is possible to refactor this method to update everything correctly without the need to do this in every place where we're calling it.

packages/frontend/editor-ui/src/app/stores/workflows.store.ts
```diff
@@ -1717,10 +1719,6 @@ export const useWorkflowsStore = defineStore(STORES.WORKFLOWS, () => {
 
 		// Update local store state to reflect the change
 		if (isCurrentWorkflow) {
-			setWorkflowVersionId(updated.versionId);
-			if (updated.checksum) {
-				setWorkflowChecksum(updated.checksum);
-			}
 			setWorkflowSettings(updated.settings ?? {});
```
dariacodes, line 1723: If we're removing this, we should definitely set `versionId` in `updateWorkflow` as well, right? (see my prev comment)

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
```diff
@@ -847,20 +847,26 @@ export function useWorkflowHelpers() {
...
 
 		const workflow = await workflowsStore.updateWorkflow(workflowId, data);
-		workflowsStore.setWorkflowVersionId(workflow.versionId);
-		if (workflow.checksum) {
-			workflowsStore.setWorkflowChecksum(workflow.checksum);
+		if (!workflow.checksum) {
+			throw new Error('Failed to update workflow');
 		}
 
+		workflowsStore.setWorkflowVersionId(workflow.versionId, workflow.checksum);
```
dariacodes: If we move this to `updateWorkflow`, then we can remove this from here, right?

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
```diff
@@ -847,20 +847,26 @@ export function useWorkflowHelpers() {
...
 			uiStore.stateIsDirty = false;
 		}
 
 		if (workflow.activeVersion) {
-			workflowsStore.setWorkflowActive(workflowId, workflow.activeVersion, isCurrentWorkflow);
+			workflowsStore.setWorkflowActive(
+				workflowId,
+				workflow.activeVersion,
+				isCurrentWorkflow,
+				workflow.checksum,
```
dariacodes: I'm not sure about this one, because we need to update the checksum this way only when it's the only thing that changed, right? But here the whole workflow was already updated and we updated the checksum in the store method, so we don't need to set the checksum again.
Also, it's not possible anymore to update "active" unless we call "activate"/"deactivate" endpoints, right?

packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
```diff
@@ -482,10 +482,10 @@ const saveSettings = async () => {
 
 	try {
 		const workflowData = await workflowsStore.updateWorkflow(String(route.params.name), data);
-		workflowsStore.setWorkflowVersionId(workflowData.versionId);
-		if (workflowData.checksum) {
-			workflowsStore.setWorkflowChecksum(workflowData.checksum);
+		if (!workflowData.checksum) {
+			throw new Error('Failed to update workflow');
 		}
+		workflowsStore.setWorkflowVersionId(workflowData.versionId, workflowData.checksum);
```
dariacodes: Same here (see other comments), we can move this to `updateWorkflow`

packages/frontend/editor-ui/src/app/stores/workflows.store.ts
scdekov: We can use the `setWorkflowVersionId` here fore sure. I was mainly focused on the checksum and that's why I didn't do it. Will update it now

packages/frontend/editor-ui/src/app/stores/workflows.store.ts
scdekov: This method is supposed to update settings only in which we shouldn't be changing the version hence I think an explicit setting of the version here could be confusing.

dariacodes: But it's a backend decision, right? So I'd leave it to backend and have the same logic of updating the state to not even have a need to think about it on frontend. Wdyt?

scdekov: Make sense if we move that versions setting in the updateWorkflow method which I'll do.

dariacodes: Yeah, exactly! Thank you!

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
scdekov: You're right, we should be able to activate or deactivate from here. I'll drop this code. As for the checksum param in setWorkflowActive my goal was to make it harder to forget to update the checksum when activating since we have an explicit setter for it. I'm still not able to fully understand why we have all of the explicit setters for the different wf props. Anyways the `setWorkflowActive` is used only which happens only from one place and in there we're also using the `setWorkflowVersionId` so I guess I can drop the param.

dariacodes (approved):
Good job! I tested it and it seems to be working correctly 👍

PR #23045, opened 2025-12-10
fix(core): Reload external secrets settings and providers from db on pubsub event

guillaumejacquart:
## Summary

This PR solves this problem:
- In multi main setup, when an instance changes external secret configuration (add, changes, delete a provider), it saves it to the db and pushes a pubsub event
- other instance listen to this event, but reload only the provider that were already in cache
- this means that all new providers were not reloaded at added to the provider cache of leader instances

The solution:
- on pubsub event listener, reloads settings from the db
- reload new and existing providers configuration
- disconnect and remove old providers from cache

https://linear.app/n8n/issue/PAY-4295/external-secrets-provider-list-disappears-at-random#comment-6e331eec

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
```diff
@@ -86,11 +84,24 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
...
+		this.cachedSettings = (await this.getDecryptedSettings()) ?? {};
+
+		const newProviders = new Set(Object.keys(this.cachedSettings));
+		const existingProviders = Object.keys(this.providers);
+
+		// Disconnect and remove providers that are no longer in settings
+		for (const provider of existingProviders) {
+			if (!newProviders.has(provider)) {
+				this.logger.debug(`Removing provider ${provider} - no longer in settings`);
+				await this.providers[provider].disconnect().catch(() => {});
```
MarcL: Should we want to log an error here if there's an error on disconnection?

guillaumejacquart: Good point. I added a warning log (don't want to make it an error because it could be quite legit - provider was removed on the other main and underlying service is off). I added similar logs anywhere we catch the disconnect

MarcL (approved):

PR #23050, opened 2025-12-10
Added support for NTLM SMTP authentication

Braidon:
## Summary

Added NTLM switch at SMTP parameters form to turn on NTLM auth, it uses the same login/password arguments.


shortstacked:
Closing as massively out of sync.

PR #23058, opened 2025-12-11
fix(editor): Model selector dropdown menu in chat closes too eagerly

autologie:
## Summary

This PR fixes the bug in model selector dropdown in chat where it closes when a top-level menu item is clicked.

I initially thought that the behavior comes from ElementPlus, but it turned out that the culprit was `onClickOutside` which I added to ensure closing the menu when it's opened programmatically.

## Review / Merge checklist

- [ ] Tests included. 


Cadiac (approved):
Nice that we could fix it this simply 👍

PR #23060, opened 2025-12-11
feat(ai-builder): Add support for versioning in ai-builder

mutdmour:
## Summary

https://github.com/user-attachments/assets/71310cc2-dca0-4efa-bcc5-f0e31e5bcad4

## Flow Diagrams

New User Message Flow:

1. User types message
2. Frontend saves workflow → gets versionId
3. Frontend sends message to /ai/build with versionId in payload
4. Backend stores versionId in message metadata (additional_kwargs)
5. AI processes and responds
6. (No version save after response - already saved before)

Session Loading Flow:

1. Frontend calls /ai/sessions to get messages
2. Extract versionIds from messages with additional_kwargs.versionId
3. Call POST /workflow-history/workflow/:id/versions with versionIds
4. Backend returns only existing versions
5. Frontend filters out messages referencing non-existent versions
6. Display version cards only for existing versions

Restore Flow:

1. User clicks restore on version card (version card is derived from user message with versionId)
2. Show RestoreVersionConfirm dialog
3. On confirm:
   a. restore workflow version using workflow history feature
   b. Backend finds user message with this messageId in metadata, removes that message AND all messages after it
   c. Frontend truncates local messages similarly

## Review / Merge checklist

- [X] Tests included. 


packages/frontend/editor-ui/src/features/ai/assistant/builder.store.ts
```diff
@@ -776,6 +852,70 @@ export const useBuilderStore = defineStore(STORES.BUILDER, () => {
...
+			false,
+		);
+
+		// version id is important to update, because otherwise the next time user saves,
+		// "overwrite" prevention modal shows, because the version id on the FE would be out of sync with latest on the backend
+		workflowState.setWorkflowProperty('versionId', updatedWorkflow.versionId);
+		workflowState.setWorkflowProperty('updatedAt', updatedWorkflow.updatedAt);
+
+		// 2. Truncate messages in backend session (removes message with messageId and all after)
+		await truncateBuilderMessages(rootStore.restApiContext, workflowId, messageId);
```
mike12345567, line 888: One thing I noted using this locally, the truncation happening after the restore looks a little odd, it might make sense to do both the truncation and restoring of the workflow at the same time with `Promise.all([...])`

packages/cli/src/workflows/workflow-history/__tests__/workflow-history.service.test.ts
```diff
@@ -127,4 +128,111 @@ describe('WorkflowHistoryService', () => {
...
+			// Arrange
+			const workflowId = '123';
+			const versionIds = ['version1'];
+			const workflow = getWorkflow({ addNodeWithoutCreds: true });
+			workflow.id = workflowId;
+			workflowFinderService.findWorkflowForUser.mockResolvedValueOnce(workflow);
+
+			const createdAt = new Date('2024-01-01');
+			const mockVersions = [{ versionId: 'version1', createdAt }];
+			workflowHistoryRepository.find.mockResolvedValueOnce(mockVersions as any);
```
mike12345567: use of `as any` in this file - should be possible to type this correctly or cast it to the correct type.

packages/frontend/@n8n/design-system/src/components/AskAssistantChat/messages/TextMessage.test.ts
```diff
@@ -0,0 +1,431 @@
+import { render, fireEvent } from '@testing-library/vue';
+import { createPinia, setActivePinia } from 'pinia';
+import { nextTick } from 'vue';
+
+import TextMessage from './TextMessage.vue';
```
mike12345567, line 5: Thanks for adding this test set that was missing!

mike12345567 (dismissed):
LGTM! One minor comment, I noticed it when playing around with it, the truncation/restore happening at separate times felt a little odd.

Great feature, very well implemented, can't wait for users to get it!

packages/frontend/editor-ui/src/features/ai/assistant/builder.store.ts
mutdmour: Ya maybe UX could be improved. But if restore API request fails for some reason, I think it would be better to keep the context.

mutdmour: Maybe a loading state is needed. Or we fade the messages or something. 🤔 I would say out of scope for now though.

mike12345567 (dismissed):
LGTM!

mike12345567 (approved):

PR #23066, opened 2025-12-11
chore(core): Workflow credential status endpoint

afitzek:
## Summary

Add workflow status endpoint, to check if a workflow can be executed using a specific access token.

Example request:
```
curl -H "Authorization: Bearer test" http://localhost:5678/rest/workflows/NWwk9GB1F2uPBKfX/execution-status
```

Response if workflow is not ready to execute
```
{
  "data": {
    "workflowId": "NWwk9GB1F2uPBKfX",
    "readyToExecute": false,
    "credentials": [
      {
        "credentialId": "nS6qcOJq8G07WM8k",
        "credentialStatus": "missing",
        "credentialType": "httpBearerAuth",
        "authorizationUrl": "/credentials/nS6qcOJq8G07WM8k/authorize?resolverId=t9wgNJoZYuhtrum6"
      }
    ]
  }
}
```

Response if workflow is ready to execute
```
{
  "data": {
    "workflowId": "NWwk9GB1F2uPBKfX",
    "readyToExecute": true,
    "credentials": [
      {
        "credentialId": "nS6qcOJq8G07WM8k",
        "credentialStatus": "configured",
        "credentialType": "httpBearerAuth",
        "authorizationUrl": "/credentials/nS6qcOJq8G07WM8k/authorize?resolverId=t9wgNJoZYuhtrum6"
      }
    ]
  }
}
```

closes https://linear.app/n8n/issue/PAY-4237/implement-workflow-credential-status-endpoint

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/api-types/src/dto/workflows/workflow-execution-status.dto.ts
+++ b/packages/@n8n/api-types/src/dto/workflows/workflow-execution-status.dto.ts
@@ -1,0 +1,18 @@
+import { z } from 'zod';
+
+export const WorkflowExecutionStatusDtoSchema = z.object({
+	workflowId: z.string(),
+	credentials: z
+		.array(
+			z.object({
+				credentialId: z.string(),
+				credentialType: z.string(),
+				credentialStatus: z.enum(['missing', 'configured']),
+				authorizationUrl: z.string().optional(),
+			}),
+		)
+		.optional(),
+	readyToExecute: z.boolean(),
+});
+
+export type WorkflowExecutionStatusDto = z.infer<typeof WorkflowExecutionStatusDtoSchema>;
--- a/packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
@@ -1,0 +1,137 @@
+import { CredentialsRepository, In, WorkflowRepository } from '@n8n/db';
+import { DynamicCredentialResolverRegistry } from './credential-resolver-registry.service';
+import { DynamicCredentialResolverRepository } from '../database/repositories/credential-resolver.repository';
+import { Cipher } from 'n8n-core';
+import { jsonParse } from 'n8n-workflow';
+import { ICredentialResolver } from '@n8n/decorators';
+import { Service } from '@n8n/di';
+
+@Service()
+export class CredentialResolverWorkflowService {
+	constructor(
+		private readonly workflowRepository: WorkflowRepository,
+		private readonly credentialRepository: CredentialsRepository,
+		private readonly resolverRegistry: DynamicCredentialResolverRegistry,
+		private readonly resolverRepository: DynamicCredentialResolverRepository,
+		private readonly cipher: Cipher,
+	) {}
+
+	private async getResolver(resolverId: string) {
+		const resolver = await this.resolverRepository.findOneBy({ id: resolverId });
+		if (!resolver) {
+			throw new Error('Credential resolver not found');
+		}
+		const resolverInstance = this.resolverRegistry.getResolverByName(resolver.type);
+		if (!resolverInstance) {
+			throw new Error('Credential resolver implementation not found');
+		}
+		try {
+			const decryptedConfig = this.cipher.decrypt(resolver.config);
+			const resolverConfig = jsonParse<Record<string, unknown>>(decryptedConfig);
+
+			return {
+				resolverInstance,
+				resolverConfig,
+			};
+		} catch (error) {
+			throw new Error('Failed to decrypt or parse resolver configuration');
+		}
+	}
+
+	async getWorkflowStatus(workflowId: string, identityToken: string) {
+		const workflow = await this.workflowRepository.get({
+			id: workflowId,
+		});
+
+		if (!workflow) {
+			throw new Error('Workflow not found');
+		}
+
+		const resolverId = workflow.settings?.credentialResolverId;
+
+		let workflowResolverInstance: ICredentialResolver | null = null;
+		let workflowResolverConfig: Record<string, unknown> | null = null;
+
+		if (resolverId) {
+			const { resolverInstance, resolverConfig } = await this.getResolver(resolverId);
+			workflowResolverInstance = resolverInstance;
+			workflowResolverConfig = resolverConfig;
+		}
+
+		const credentialsToCheck: Array<{ credentialId: string }> = [];
+
+		for (const node of workflow.nodes ?? []) {
+			for (const credentialName in node.credentials ?? {}) {
+				const credentialData = node.credentials?.[credentialName];
+				// Fetch credentials from the DB using the credentialData.id
+				if (credentialData?.id) {
+					credentialsToCheck.push({ credentialId: credentialData.id });
+				}
+			}
+		}
+
+		const credentials = await this.credentialRepository.find({
+			where: {
+				id: In(credentialsToCheck.map((c) => c.credentialId)),
+				isResolvable: true,
+				// TODO: filter on isResolvable true only once we have a migration to add the column with default false
+			},
+		});
+		// Get all credentials from the DB based on the ids collected above and resolvable being true
+		//
+		const credentialsStatuses: Array<{
+			credentialId: string;
+			resolverId: string;
+			credentialType: string;
+			status: 'missing' | 'configured';
+		}> = [];
+
+		for (const credential of credentials) {
+			if (credential.isResolvable) {
+				let resolverInstance: ICredentialResolver | null = workflowResolverInstance;
+				let resolverConfig: Record<string, unknown> | null = workflowResolverConfig;
+				const credentialResolverId = credential.resolverId ?? resolverId;
+				if (credentialResolverId) {
+					if (credentialResolverId !== resolverId) {
+						const {
+							resolverInstance: credentialResolverInstance,
+							resolverConfig: credentialResolverConfig,
+						} = await this.getResolver(credentialResolverId);
+						resolverInstance = credentialResolverInstance;
+						resolverConfig = credentialResolverConfig;
+					}
+
+					if (resolverConfig && resolverInstance) {
+						try {
+							await resolverInstance.getSecret(
+								credential.id,
+								{ identity: identityToken, version: 1 },
+								{
+									configuration: resolverConfig,
+									resolverName: resolverInstance.metadata.name,
+									resolverId: credentialResolverId,
+								},
+							);
+							credentialsStatuses.push({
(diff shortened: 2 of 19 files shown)


packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+import { Service } from '@n8n/di';
+
+type CredentialStatus = {
+	credentialId: string;
+	resolverId: string;
+	credentialType: string;
+	status: 'missing' | 'configured';
+};
+
+function isCredentialStatus(obj: unknown): obj is CredentialStatus {
```
BGZStephen, line 16: Is there a reason you chose this approach over using Zod?

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+		private readonly resolverRepository: DynamicCredentialResolverRepository,
+		private readonly cipher: Cipher,
+	) {}
+
+	private async getResolver(resolverId: string) {
+		const resolver = await this.resolverRepository.findOneBy({ id: resolverId });
+		if (!resolver) {
+			throw new Error('Credential resolver not found');
+		}
+		const resolverInstance = this.resolverRegistry.getResolverByName(resolver.type);
```
BGZStephen: A bit nitpicky, I would expect `getResolverByName` to take the `name` of a resolver, not the `type`, not sure if it's worth reworking it to be a little clearer

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+export class CredentialResolverWorkflowService {
+	constructor(
+		private readonly workflowRepository: WorkflowRepository,
+		private readonly credentialRepository: CredentialsRepository,
+		private readonly resolverRegistry: DynamicCredentialResolverRegistry,
+		private readonly resolverRepository: DynamicCredentialResolverRepository,
+		private readonly cipher: Cipher,
+	) {}
+
+	private async getResolver(resolverId: string) {
```
BGZStephen: We're missing a return type here, which would be ideal to have

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+		let workflowResolverInstance: ICredentialResolver | null = null;
+		let workflowResolverConfig: Record<string, unknown> | null = null;
+
+		if (resolverId) {
+			const { resolverInstance, resolverConfig } = await this.getResolver(resolverId);
+			workflowResolverInstance = resolverInstance;
+			workflowResolverConfig = resolverConfig;
+		}
+
+		const credentialsToCheck: Array<{ credentialId: string }> = [];
```
BGZStephen: I don't think this needs to be an array of objects as you just map it to credential ids from what I can see?

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+
+		const credentials = await this.credentialRepository.find({
+			where: {
+				id: In(credentialsToCheck.map((c) => c.credentialId)),
+				isResolvable: true,
+			},
+		});
+		// Get all credentials from the DB based on the ids collected above and resolvable being true
+
+		const credentialStatusPromises = credentials.map(async (credential) => {
```
BGZStephen, line 120: Might be nice to split out this anonymous function to make this `getWorkflowStatus` a bit smaller, and I'm assuming make this code a little easier to unit test in isolation?

packages/@n8n/api-types/src/dto/workflows/workflow-execution-status.dto.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod';
+
+export const WorkflowExecutionStatusDtoSchema = z.object({
```
guillaumejacquart: I think for response types, we do not use the "DTO" terminology. We use the schema folder (at least that's what I've been doing :D)

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+					resolverConfig = credentialResolverConfig;
+				}
+
+				if (resolverConfig && resolverInstance) {
+					try {
+						await resolverInstance.getSecret(
+							credential.id,
+							{ identity: identityToken, version: 1 },
+							{
+								configuration: resolverConfig,
```
guillaumejacquart: Something to keep in mind, that that the resolver config could contain expressions (see this PR: https://github.com/n8n-io/n8n/pull/23110) 
But then resolved config could depend on runtime data 🤔 , meaning on this call, some credentials that need runtime data to resolve will always be considered as non configured ?

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+							},
+						);
+						return {
+							credentialId: credential.id,
+							resolverId: credentialResolverId,
+							status: 'configured',
+							credentialType: credential.type,
+						};
+					} catch (error) {
+						// Handle error (e.g., log it, collect status, etc.
```
guillaumejacquart: Differentiating between missing resolved creds and them having an error could be interesting here

packages/cli/src/modules/dynamic-credentials.ee/workflow-status.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Request, Response } from 'express';
+
+import { BadRequestError } from '@/errors/response-errors/bad-request.error';
+import { CredentialResolverWorkflowService } from './services/credential-resolver-workflow.service';
+import { WorkflowExecutionStatusDto } from '@n8n/api-types';
+
+const BEARER_TOKEN_REGEX = /^[Bb][Ee][Aa][Rr][Ee][Rr]\s+(.+)$/;
+
+@RestController('/workflows')
+export class WorkflowStatusController {
```
guillaumejacquart, line 12: File name do not match class

packages/cli/src/modules/dynamic-credentials.ee/workflow.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+	@Get('/:workflowId/execution-status', { skipAuth: true })
+	async checkWorkflowForExecution(
+		req: Request,
+		_res: Response,
+	): Promise<WorkflowExecutionStatusDto> {
+		const workflowId = req.params['workflowId'];
+		const headerValue = req.headers['authorization']?.toString();
+
+		if (!headerValue) {
+			throw new BadRequestError('Authorization header missing');
```
guillaumejacquart: Why not unauthenticated error here ? also maybe not expose the missing header

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,164 @@
...
+	resolverId: string;
+	credentialType: string;
+	status: 'missing' | 'configured';
+};
+
+function isCredentialStatus(obj: unknown): obj is CredentialStatus {
+	if (typeof obj !== 'object' || obj === null) {
+		return false;
+	}
+	return (
```
guillaumejacquart, line 20: Do we really need this type checking here ? I think obj is not supposed to be unknown, since we're building it in the credentialStatusPromises promise map

guillaumejacquart (commented):
Some comments

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
afitzek: These objects are only build internally, the typeguard exists for filtering out null values in a typescript safe way. Using zod for parsing is more expensive and not necessary in my opinion.

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
afitzek: Good point, I'll change the function name to ..ByTypename

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
afitzek: Ah yes, I wanted to store more values in there first 😅

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
afitzek: Yes we will have to add the dynamic resolution here as well. As discussed, we will keep it to only static data.

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
afitzek: I thought, if we do a type guard, lets make it work with unknown. :)

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver-workflow.service.ts
```diff
@@ -0,0 +1,184 @@
...
+			workflowResolverInstance: ICredentialResolver | null;
+			workflowResolverConfig: Record<string, unknown> | null;
+			resolverId: string | undefined;
+			identityToken: string;
+		},
+	): Promise<CredentialStatus | null> {
+		let resolverInstance: ICredentialResolver | null = options.workflowResolverInstance;
+		let resolverConfig: Record<string, unknown> | null = options.workflowResolverConfig;
+		const credentialResolverId = credential.resolverId ?? options.resolverId;
+		if (credentialResolverId) {
```
BGZStephen, line 144: A bit nitpicky, but you could invert this check and return null early in no credential resolver id is available, avoids nesting ifs

BGZStephen (approved):

PR #23074, opened 2025-12-11
chore(core): Allow marking fields in credentials as resolvable

afitzek:
## Summary

Introduces a mechanism to mark credential fields as resolvable via the new `resolvableField` property on credential schema properties. Fields defined in the schema without `resolvableField: true` are treated as static and sourced from the credential entity, while fields marked as resolvable or not defined in the schema are dynamically resolved. This enables proper separation between static credential configuration (e.g., `clientId`, `clientSecret` in OAuth2) and dynamic user-specific data (e.g., `accessToken`, `user`, `password`)

closes https://linear.app/n8n/issue/PAY-4301/solve-shared-fields-for-write-path-for-dynamic-credentials

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
@@ -83,13 +83,13 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 			return this.handleMissingContext(credentialsResolveMetadata, staticData);
 		}
 
-		const credentialType = this.loadNodesAndCredentials.getCredential(
-			credentialsResolveMetadata.type,
-		);
-
-		const sharedFields = extractSharedFields(credentialType.type);
-
 		try {
+			const credentialType = this.loadNodesAndCredentials.getCredential(
+				credentialsResolveMetadata.type,
+			);
+
+			const sharedFields = extractSharedFields(credentialType.type);
+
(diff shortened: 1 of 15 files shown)


packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -80,6 +84,12 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 		}
 
 		try {
+			const credentialType = this.loadNodesAndCredentials.getCredential(
```
guillaumejacquart, line 87: nit: why not put it on line 114 just before it's used (for cognitive load)

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -102,6 +112,13 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 				identity: credentialContext.identity,
 			});
 
+			// Remove shared fields from dynamic data to avoid conflicts
+			for (const field of sharedFields) {
+				if (field in dynamicData) {
```
guillaumejacquart, line 117: maybe the if is useless, as delete an object key that does not exist has no effect

guillaumejacquart (approved):
Looks good. Just some nit comments

PR #23079, opened 2025-12-11
fix(task-runner-python): allow standard library imports by default (f…

ppandit12:
Fixes regression #23072 where Python standard library modules (e.g., 
re
, zlib) were blocked by default in the Python task runner. The issue was caused by 
stdlib_allow
 defaulting to an empty string in 
TaskRunnerConfig
. This change updates the default to * (allow all), restoring the intended behavior.


ppandit12 (commented):
i have checked and it is working

PR #23094, opened 2025-12-11
fix: Move webhook duplication checks to backend and make them more reliable

### LINKED ISSUE #21708: Conflicting Webhook Path - Logic Validation Broken
### Bug Description

When declaring Webhook nodes in two separate workflows where each Webhook uses the same `Path` value of `:foobar`, n8n incorrectly blocks activating the second workflow because of a false conflict.  This is because global `Path` validation doesn't take into account the full URL of each unique Webhook.

For example:
Webhook 1 has declared: https://n8n.domain.com/webhook/0ad14bed-cf33-445b-8378-db0cc41a165b/:filename
Webhook 2 has declared: https://n8n.domain.com/webhook/06222c6b-f3fb-4012-bfb7-aec6284b4c2e/:filename

However, upon activating the workflow containing Webhook 2, n8n throws a `Conflicting Webhook Path` error and states the duplicative URL of:
https://n8n.domain.com/webhook/:filename

... the validation logic completely omits the UUID prefix in the URI path, which is why this validation is faulty.

### To Reproduce

1. Create workflow 1 with a single Webhook node and specify a `Path` value of `:filename`.
2. Activate workflow 1
3. Create workflow 2 with a single Webhook node and specify a `Path` value of `:filename`.
4. Activate workflow 2 -- see error.

### Expected behavior

I would expect the Conflicting Webhook Path validation logic to properly factor in the UUID portion of the URI for each Webhook before throwing a subsequent error.

### Debug Info

# Debug info

## core

- n8nVersion: 1.118.2
- platform: npm
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: postgres
- concurrency: -1
- consumerId: ac2e881d-ae53-4eb9-94a6-32bb3d9f0fe3

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 168 hours
- maxCount: 100000 executions

## client

- isTouchDevice: false

Generated at: 2025-11-11T02:06:09.662Z

### Operating System

Google Cloud Run

### n8n Version

1.118.2

### Node.js Version

22.21.0

### Database

PostgreSQL

### Execution mode

queue

### Hosting

self hosted

dlavrenuek:
## Summary

Implements webhook url conflict validation checks in the backend.
Conflict checks are implemented in webhook.service class, so the same check can be reused to check for conflicts on webhook creation.

- https://linear.app/n8n/issue/NODE-4077
- https://linear.app/n8n/issue/NODE-3862/community-issue-conflicting-webhook-path-logic-validation-broken

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/components/WorkflowActivationConflictingWebhookModal.vue
```diff
@@ -83,7 +83,8 @@ const onClick = async () => {
 			<div :class="$style.container">
 				<div>
 					<N8nText color="text-base"> You can deactivate </N8nText>
-					<N8nLink :to="workflowUrl" :underline="true"> {{ data.workflowName }} </N8nLink>
+					<N8nLink :to="workflowUrl" :underline="true">{{ data.workflowName }}</N8nLink>
+					{{ ' ' }}
```
dlavrenuek, line 87: There was a white space missing in the text. Before the change
Bildschirmfoto 2025-12-11 um 17 31 19
After the change
Bildschirmfoto 2025-12-11 um 17 32 48

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -296,6 +302,76 @@ export class WebhookService {
...
+			webhooks: IWebhookData[];
+		}[],
+	) {
+		const conflicts: {
+			trigger: INode;
+			conflict: Partial<WebhookEntity>;
+		}[] = [];
+
+		while (checkEntries.length > 0) {
+			const { node, webhooks } = checkEntries.pop()!;
```
dlavrenuek: To check for conflicts locally (same webhook within one workflow), all of the webhooks must be checked with all other webhooks in the workflow. To reduce the operation from O(n^2) to O(n log n) I did implement "pop"ing the last item and checking it with the remaining ones.

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -296,6 +302,76 @@ export class WebhookService {
...
+					webhooks.some((webhook) =>
+						checkWebhooks.some(
+							(checkWebhook) =>
+								this.getWebhookPath(checkWebhook) === this.getWebhookPath(webhook) &&
+								checkWebhook.httpMethod === webhook.httpMethod,
+						),
+					),
+			);
+			if (localConflict) {
+				const firstConflict = localConflict.webhooks[0];
```
dlavrenuek: While it is possible that there are multiple conflicts for one node (form claims POST and GET routes) only one conflict is returned for one node here. At the moment It is sufficient to determine if there is a conflict and display where the conflict is in the frontend - I also don't see a use case right now where all of the conflicts per node would be needed/used

packages/frontend/editor-ui/src/app/composables/useWorkflowActivate.ts
```diff
@@ -206,24 +209,70 @@ export function useWorkflowActivate() {
...
 		}
 	};
 
+	const handleWebhookConflictError = async (error: ResponseError) => {
+		const { trigger, conflict } = parseWebhookConflictError(error)?.pop() || {};
+		let workflowName = conflict?.workflowId;
+		try {
+			if (conflict?.workflowId) {
+				const conflictingWorkflow = await workflowsStore.fetchWorkflow(conflict?.workflowId);
+				workflowName = conflictingWorkflow.name;
```
dlavrenuek: If the user does not have access to the conflicting workflow the id will be displayed instead of the name

Bildschirmfoto 2025-12-11 um 17 56 42.

In the same scenario right now there will be an unhandled error resulting in no message to the user at all

mfsiega (commented):
Some high-level thoughts:
- Nitpick: add a doc for `findWebhookConflicts`, since it's exported?
- Is there a way to convert each webhook to something hashable, and put them in a set to check for conflicts? Something like:
```
webhookSet = {}
for each node:
    if node.webhook in webhookSet: return ...
    webhookSet.add(node.webhook)
```
I think with popping it's still O(N^2) fwiw. I'm not _too_ worried about N^2 since a workflow should only have so many trigger nodes, but still.
- Nitpick: it looks like we're returning a list of conflicts, but we only ever return a single conflict. Should we try to find them all, or just return the first? (I think just returning the first is fine.)

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -296,6 +302,85 @@ export class WebhookService {
...
+				for (const webhook of webhooks) {
+					const path = this.getWebhookPath(webhook);
+					const potentialConflict = await this.findWebhook(webhook.httpMethod, path);
+
+					if (potentialConflict && potentialConflict.workflowId !== workflow.id) {
+						conflicts.push({
+							trigger: node,
+							conflict: potentialConflict,
+						});
+						break;
```
dlavrenuek: todo: remove break - the intention of this method is to return all conflicting webhooks


```suggestion
```

packages/cli/src/workflows/__tests__/workflow.service.test.ts
mfsiega, line 1: Might be worth adding a unit test here that just verifies that if `findWebhookConflicts` returns non-empty, then we throw.

packages/cli/src/webhooks/webhook.service.ts
dlavrenuek: Improved the check to have O(n) time complexity

packages/cli/src/workflows/__tests__/workflow.service.test.ts
dlavrenuek: Added an integration tests for that and also e2e tests. Also improved the check algorithm to use a hash map resulting in a O(n) time complexity - this is possible since it's sufficient to find all nodes that have conflicts and not all conflicts per node

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -570,6 +613,8 @@ export class WorkflowService {
 			await this._detectConflicts(workflow, options.expectedChecksum);
 		}
 
+		await this._detectWebhookConflicts(workflow);
```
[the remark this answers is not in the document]
dlavrenuek: Good find, fixed

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -179,6 +179,12 @@ export class WebhookService {
 		return path.startsWith(':') || path.includes('/:');
 	}
 
+	getWebhookPath(webhook: IWebhookData): string {
+		return [webhook.path.includes(':') ? webhook.webhookId : undefined, webhook.path]
+			.filter((part) => !!part)
+			.join('/');
```
michael-radency, line 185: would paths defined as expressions be resolved at this point?

Image

dlavrenuek: I would expect it being the case because webhooks are generated using [`this.getNodeWebhooks()`](https://github.com/n8n-io/n8n/blob/f88cb9707fcc9f2908f50e777291c57ae78e3c48/packages/cli/src/webhooks/webhook.service.ts#L191) in the same service, which does resolve expressions. To be sure I did debug the flow (breakpoint where [the comparison key is built](https://github.com/n8n-io/n8n/blob/f88cb9707fcc9f2908f50e777291c57ae78e3c48/packages/cli/src/webhooks/webhook.service.ts#L325)) with a workflow containing following expressions for a wehook `ok-{{ new Date().toDateTime().toISO() }}` which is correctly resolved as.

Image

michael-radency: nice, thanks for checking :+1:

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -296,6 +302,82 @@ export class WebhookService {
...
+
+		// store processed webhooks in a map -> O(1) remaining webhooks local conflict checks
+		const processedWebhooks: Map<string, IWebhookData> = new Map();
+		const webhookToKey = (webhook: IWebhookData) =>
+			`${webhook.httpMethod} ${this.getWebhookPath(webhook)}`;
+
+		for (const { node, webhooks } of checkEntries) {
+			for (const webhook of webhooks) {
+				const webhookKey = webhookToKey(webhook);
+				if (processedWebhooks.has(webhookKey)) {
```
mfsiega: Nit: this could be re-written like:

```
const conflict = processedWebhooks.get(webhookKey);
if (conflict) {
    ...
}
```

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -296,6 +302,82 @@ export class WebhookService {
...
+						trigger: node,
+						conflict: {
+							workflowId: workflow.id,
+							webhookPath: conflict.path,
+							method: conflict.httpMethod,
+							node: conflict.node,
+							webhookId: conflict.webhookId,
+						},
+					});
+					continue;
```
mfsiega, line 338: Do we want `continue` here and below? I'm trying to think of what would happen if there were a conflict locally *and* with another workflow. Not sure about this one.

packages/testing/playwright/tests/e2e/workflows/editor/workflow-actions/publish.spec.ts
```diff
@@ -31,32 +31,93 @@ test.describe('Workflow Publish @fixme', () => {
...
+					transform: ({ id, ...workflow }) => ({
+						...workflow,
+						id: Date.now().toString(),
+					}),
+				});
+			cleanupWorkflowIds.push(workflowId2);
+
+			await expect(
+				api.workflows.activate(workflowId2, createdWorkflow2.versionId!),
+			).rejects.toThrow();
```
mfsiega: Nit: is it good to include an assertion about the error message? To make sure it's failing for the reason we expect.

packages/testing/playwright/tests/e2e/workflows/editor/workflow-actions/publish.spec.ts
```diff
@@ -31,32 +31,93 @@ test.describe('Workflow Publish @fixme', () => {
...
+		test('Rejects publishing a workflow containing local conflicts', async ({ api }) => {
+			const { workflowId, createdWorkflow } = await api.workflows.importWorkflowFromFile(
+				'webhook-publish-local-conflict.json',
+				{ makeUnique: false },
+			);
+			cleanupWorkflowIds.push(workflowId);
+
+			await expect(
+				api.workflows.activate(workflowId, createdWorkflow.versionId!),
+			).rejects.toThrow();
```
mfsiega: Here for the error message verification as well?

mfsiega (commented):
Just a few small comments, LGTM!

I did not review the frontend changes, I think adding a separate reviewer for those would be good.

michael-radency (dismissed):
LGTM

packages/cli/src/webhooks/webhook.service.ts
dlavrenuek: Since the current goal is to determine IF there is a conflict for each webhook, finding the first conflict is enough. If the requirement would change to find all conflicts, then we would need to change the logic here (remove continue but also add all webhooks to the hashmap and not only one per webhook key.

packages/cli/src/webhooks/webhook.service.ts
dlavrenuek: good suggestion, added

packages/testing/playwright/tests/e2e/workflows/editor/workflow-actions/publish.spec.ts
dlavrenuek: fixed

packages/testing/playwright/tests/e2e/workflows/editor/workflow-actions/publish.spec.ts
dlavrenuek: fixed

michael-radency (approved):

PR #23106, opened 2025-12-12
fix(core): Get chat title even after stopping and give more time for title generation (no-changelog)

Cadiac:
## Summary

We weren't fetching the conversation title if user stopped generation. Also the max 3 seconds period was too short sometimes, lets just make it way longer while happening less often?

## Review / Merge checklist

- [ ] Tests included. 


autologie (approved):
👍

PR #23110, opened 2025-12-12
feat(core): First shot in resolving the credential resolver config expressions

guillaumejacquart:
## Summary

This PR adds expression resolution support for credential resolver configurations in the dynamic credentials system. It allows users to use n8n expressions (like $execution.id, $vars.*, $secrets.*) within credential resolver configs.

 ### Key Changes

New **ResolverConfigExpressionService** (packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts)
  - Handles resolving expressions in resolver configs at runtime and validation time
  - Supports $execution, $vars, and $secrets expressions

Type-safe resolver configuration
  - Changed CredentialResolverConfiguration from Record<string, unknown> to INodeParameters for proper type checking when executing expression resolutions
  - Added isNodeParameters type guard in packages/workflow for validating config format

Updated credential resolution flow
  - resolveIfNeeded now receives additionalData, mode, and canUseExternalSecrets instead of separate executionContext and workflowSettings
  - This provides full context needed for expression resolution

https://linear.app/n8n/issue/PAY-4302/support-expressions-in-resolver-config

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/credentials/credential-resolution-provider.interface.ts
+++ b/packages/cli/src/credentials/credential-resolution-provider.interface.ts
@@ -1,8 +1,6 @@
 import type {
 	ICredentialDataDecryptedObject,
-	IExecutionContext,
 	IWorkflowExecuteAdditionalData,
-	IWorkflowSettings,
 	WorkflowExecuteMode,
 } from 'n8n-workflow';
 
@@ -27,9 +25,7 @@ export interface ICredentialResolutionProvider {
 	 *
 	 * @param credentialsResolveMetadata The credential resolve metadata
 	 * @param staticData The decrypted static credential data
-	 * @param executionContext Optional execution context containing credential context
-	 * @param workflowSettings Optional workflow settings containing resolver ID fallback
-	 * @param additionalData Additional workflow execution data for expression resolution
+	 * @param additionalData Additional workflow execution data for expression resolution and context and settings
 	 * @param mode Workflow execution mode
 	 * @param canUseExternalSecrets Whether the credential can use external secrets for expression resolution
 	 * @returns Resolved credential data (either dynamic or static)
@@ -37,8 +33,6 @@ export interface ICredentialResolutionProvider {
 	resolveIfNeeded(
 		credentialsResolveMetadata: CredentialResolveMetadata,
 		staticData: ICredentialDataDecryptedObject,
-		executionContext?: IExecutionContext,
-		workflowSettings?: IWorkflowSettings,
 		additionalData?: IWorkflowExecuteAdditionalData,
 		mode?: WorkflowExecuteMode,
 		canUseExternalSecrets?: boolean,
--- a/packages/cli/src/credentials/credential-resolution-provider.interface.ts
+++ b/packages/cli/src/credentials/credential-resolution-provider.interface.ts
@@ -1,4 +1,8 @@
-import type { ICredentialDataDecryptedObject, IWorkflowExecuteAdditionalData } from 'n8n-workflow';
+import type {
+	ICredentialDataDecryptedObject,
+	IExecutionContext,
+	IWorkflowSettings,
+} from 'n8n-workflow';
 
 export type CredentialResolveMetadata = {
 	id: string;
@@ -28,7 +32,8 @@ export interface ICredentialResolutionProvider {
 	resolveIfNeeded(
 		credentialsResolveMetadata: CredentialResolveMetadata,
 		staticData: ICredentialDataDecryptedObject,
-		additionalData?: IWorkflowExecuteAdditionalData,
+		executionContext?: IExecutionContext,
+		workflowSettings?: IWorkflowSettings,
 		canUseExternalSecrets?: boolean,
 	): Promise<ICredentialDataDecryptedObject>;
 }
--- a/packages/cli/src/credentials/credential-resolution-provider.interface.ts
+++ b/packages/cli/src/credentials/credential-resolution-provider.interface.ts
@@ -1,7 +1,9 @@
 import type {
 	ICredentialDataDecryptedObject,
 	IExecutionContext,
+	IWorkflowExecuteAdditionalData,
 	IWorkflowSettings,
+	WorkflowExecuteMode,
 } from 'n8n-workflow';
 
 export type CredentialResolveMetadata = {
@@ -27,6 +29,9 @@ export interface ICredentialResolutionProvider {
 	 * @param staticData The decrypted static credential data
 	 * @param executionContext Optional execution context containing credential context
 	 * @param workflowSettings Optional workflow settings containing resolver ID fallback
+	 * @param additionalData Additional workflow execution data for expression resolution
+	 * @param mode Workflow execution mode
+	 * @param canUseExternalSecrets Whether the credential can use external secrets for expression resolution
 	 * @returns Resolved credential data (either dynamic or static)
 	 */
 	resolveIfNeeded(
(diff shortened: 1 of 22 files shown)


packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts
```diff
@@ -0,0 +1,128 @@
...
+			position: [0, 0],
+			parameters: {} as INodeParameters,
+		};
+
+		const mockNodesData: INodeTypeData = {
+			mock: {
+				sourcePath: '',
+				type: {
+					description: { properties: [] as INodeProperties[] },
+				} as INodeType,
```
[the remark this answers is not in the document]
guillaumejacquart: That's OK as it's a mock one, and we know what properties we need

packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts
```diff
@@ -0,0 +1,128 @@
+import { CredentialResolverConfiguration } from '@n8n/decorators';
```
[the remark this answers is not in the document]
guillaumejacquart: This service is a wrapper around other services, and has close to no business logic inside

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver.service.ts
```diff
@@ -143,18 +146,35 @@ export class DynamicCredentialResolverService {
...
 		config: CredentialResolverConfiguration,
+		canUseExternalSecrets?: boolean,
 	): Promise<void> {
 		const resolverImplementation = this.registry.getResolverByTypename(type);
 		if (!resolverImplementation) {
 			throw new CredentialResolverValidationError(`Unknown resolver type: ${type}`);
 		}
 
-		await resolverImplementation.validateOptions(config);
+		// Resolve expressions in the config with mock/empty values to validate syntax
```
afitzek: If we resolve expressions with incorrect values, wouldn't this fail, for example to validate metadata, or clientid and client secret?

afitzek (commented):
I'm not sure if we should build it like this. 🤔 If we resolve for validation in a different way than for execution, the resolver config will be quite different. There might be different values in there right? Does it make sense to only have one method available, but this method can only access "static" data, no runtime data. For example external secrets and global variables. Happy to discuss

packages/workflow/src/node-parameters/node-parameter-value-type-guard.ts
```diff
@@ -0,0 +1,167 @@
...
+	if (isFilterValue(value)) {
+		return true;
+	}
+
+	if (isAssignmentCollectionValue(value)) {
+		return true;
+	}
+
+	// Check arrays
+	if (Array.isArray(value)) {
```
geemanjs: Maybe a slight lack of understanding - but would it be valid to have a mix of valid types within the array?

guillaumejacquart: I think it should make it invalid, thus the .every checks.
I shortened it a bit though, it was quite necessary verbose IMO

packages/cli/src/credentials/credential-resolution-provider.interface.ts
```diff
@@ -25,14 +21,14 @@ export interface ICredentialResolutionProvider {
...
+	 * @param additionalData Additional workflow execution data for context and settings
+	 * @param canUseExternalSecrets Whether the credential can use external secrets for expression resolution
 	 * @returns Resolved credential data (either dynamic or static)
 	 */
 	resolveIfNeeded(
 		credentialsResolveMetadata: CredentialResolveMetadata,
 		staticData: ICredentialDataDecryptedObject,
-		executionContext?: IExecutionContext,
-		workflowSettings?: IWorkflowSettings,
+		additionalData?: IWorkflowExecuteAdditionalData,
```
BGZStephen: just a general comment, `additionalData` is more ambiguous than the previous parameter names `executionContext` and `workflowSettings`, maybe we can't avoid it, but just wanted to comment

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver.service.ts
```diff
@@ -143,18 +146,32 @@ export class DynamicCredentialResolverService {
 
 	/**
 	 * Validates the config against the resolver type's schema.
+	 * Resolves expressions
 	 * @throws {CredentialResolverValidationError} When the resolver type is unknown or config is invalid
 	 */
 	private async validateConfig(
 		type: string,
 		config: CredentialResolverConfiguration,
+		canUseExternalSecrets?: boolean,
```
BGZStephen: nit: might be nicer to use a default parameter here rather than `canUseExternalSecrets ?? false`

packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts
```diff
@@ -0,0 +1,62 @@
...
+import { NodeTypes } from '@/node-types';
+import { getBase } from '@/workflow-execute-additional-data';
+
+/**
+ * Service for resolving expressions in credential resolver configurations.
+ * Uses global data only (secrets, variables) without runtime execution context.
+ */
+@Service()
+export class ResolverConfigExpressionService {
+	private readonly mockNode: INode;
```
BGZStephen: is this mock node left over from testing? seems a little odd to be here

packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts
```diff
@@ -0,0 +1,62 @@
...
+			nodeTypes: this.nodeTypes,
+		});
+
+		const additionalData = await getBase();
+		const additionalKeys = getNonWorkflowAdditionalKeys(additionalData, {
+			secretsEnabled: canUseExternalSecrets,
+		});
+
+		// If config is not INodeParameters, return as is
+		if (!isNodeParameters(config)) {
```
BGZStephen: nit, why not do this aft the very beginning of this function call?

BGZStephen (dismissed):
A few small nitpick comments, but nothing massive.

packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts
guillaumejacquart: Yeah, so it's not a mock for testing, it's a "sample" node because the expression engine (currently) requires a node to run. I removed the private field and put the node directly in the call with a comment

packages/cli/src/credentials/credential-resolution-provider.interface.ts
guillaumejacquart: I agree, it's not necessary, i reverted that

packages/cli/src/modules/dynamic-credentials.ee/services/resolver-config-expression.service.ts
```diff
@@ -0,0 +1,59 @@
...
+		const additionalKeys = getNonWorkflowAdditionalKeys(additionalData, {
+			secretsEnabled: canUseExternalSecrets,
+		});
+
+		return workflow.expression.getComplexParameterValue(
+			// Use a mock node (mandatory) to resolve expressions in the config
+			{
+				id: '1',
+				name: 'Mock Node',
+			} as INode,
```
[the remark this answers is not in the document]
guillaumejacquart: Hard to do differently for now, considering the expression resolving types (INodeParameters)

BGZStephen (approved):

PR #23113, opened 2025-12-12
fix: Make sure duplicating workflows creates them as unpublished (no-changelog)

dariacodes:
## Summary

When duplicating a published workflow, 2 problems were happening:
1. Frontend was sending a create payload with all the active fields set,
2. Backend was ignoring `active` and `activeVersionId` fields, but not `activeVersion`, so if `activeVersion` was set in the payload, workflow was created as published.

This PR fixes both problems.

Closes [ADO-4533](https://linear.app/n8n/issue/ADO-4533/bug-duplicating-workflows-doesnt-properly-update-active-version)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/workflows/workflows.controller.ts
+++ b/packages/cli/src/workflows/workflows.controller.ts
@@ -109,7 +109,9 @@ export class WorkflowsController {
 			);
 
 			// @ts-expect-error: We shouldn't accept this
-			req.body.activeVersionId = undefined;
+			delete req.body.activeVersionId;
+			// @ts-expect-error: We shouldn't accept this
+			delete req.body.activeVersion;
 			req.body.active = false;
 		}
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/workflows/workflows.controller.ts
```diff
@@ -214,12 +216,6 @@ export class WorkflowsController {
 				transactionManager,
 			);
 
-			const shouldActivate = req.body.active === true;
-			if (shouldActivate) {
-				workflow.activeVersionId = workflow.versionId;
-				await transactionManager.save(workflow);
-			}
-
```
dariacodes, line 222: This is just a clean-up. This code was never reachable as we are setting `active=false` above

packages/cli/src/workflows/workflows.controller.ts
autologie, line 106: ~~I think we should consider changing to info/debug or even remove, since this can happen in a happy path scenario of duplicating a workflow~~ saw removed from request in FE, looks good

autologie (approved):
Tried locally and seemed to work 👍

PR #23114, opened 2025-12-12
fix(API): Fixes how toJsonSchema creates dependant allOf checks when multiple fields depend on same field

guillaumejacquart:
## Summary
Fixed Public API credential creation failing for credentials with conditional field requirements (e.g., JWT Auth).

## Problem
Creating JWT Auth credentials via Public API failed with schema validation errors because the `toJsonSchema` function incorrectly grouped all properties depending on the same field into a single conditional, regardless of their specific trigger values.

**Example:** When `keyType: 'pemKey'`, it incorrectly required `secret`, `privateKey`, AND `publicKey` instead of just `privateKey` and `publicKey`.

## Solution
Changed the dependency key from just the field name to `fieldName:value` combination to create separate conditionals for each unique field+value pair.

https://linear.app/n8n/issue/PAY-3849/community-issue-jwt-auth-creation-via-api-fails-schema-validation-for

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
+++ b/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
@@ -215,11 +215,10 @@ export function toJsonSchema(properties: INodeProperties[]): IDataObject {
 				dependantValue = displayOptionsValues[0];
 			}
 
-			if (propertyRequiredDependencies[dependantName] === undefined) {
-				propertyRequiredDependencies[dependantName] = {};
-			}
+			// Create a unique key for each dependant name and value combination
+			const dependencyKey = `${dependantName}:${JSON.stringify(dependantValue)}`;
 
-			if (!resolveProperties.includes(dependantName)) {
+			if (!resolveProperties.includes(dependencyKey)) {
 				let conditionalValue;
 				if (typeof dependantValue === 'object' && dependantValue._cnd) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
```diff
@@ -215,11 +215,10 @@ export function toJsonSchema(properties: INodeProperties[]): IDataObject {
 				dependantValue = displayOptionsValues[0];
 			}
 
-			if (propertyRequiredDependencies[dependantName] === undefined) {
-				propertyRequiredDependencies[dependantName] = {};
-			}
+			// Create a unique key for each dependant name and value combination
+			const dependencyKey = `${dependantName}:${JSON.stringify(dependantValue)}`;
```
ireneea, line 221: [nit] Would it be worth adding an extra comment to explain "why" we need this. I understood why because I had the ticket context but without it I think that it might be a bit tricky to git it

ireneea (dismissed):
Nice tests 🚀 , looks good to me,  just left nit-picking comment so feel free to ignore

packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
guillaumejacquart: Good point. I added a longer comment to explain the reasoning behind that

ireneea (approved):

PR #23133, opened 2025-12-12
fix: Escape special characters in queries

yehorkardash:
## Summary

Escape special characters in queries

https://linear.app/n8n/issue/NODE-4036/

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/MySql/v2/helpers/utils.ts
```diff
@@ -575,3 +576,26 @@ export function replaceEmptyStringsByNulls(
...
+	if (!Array.isArray(whereClausesValues)) {
+		return [];
+	}
+	const someInvalid = whereClausesValues.some((clause) => !isWhereClause(clause));
+	if (someInvalid) {
+		throw new NodeOperationError(ctx.getNode(), 'Invalid where clause', {
+			itemIndex,
+		});
+	}
+	return whereClausesValues as WhereClause[];
```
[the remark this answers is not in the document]
dlavrenuek: all clauses are valid, otherwise an exception is thrown

packages/nodes-base/nodes/MySql/v2/helpers/utils.ts
```diff
@@ -575,3 +576,26 @@ export function replaceEmptyStringsByNulls(
 
 	return returnData;
 }
+
+const conditionSet = new Set(operatorOptions.map((option) => option.value));
+
+export const isWhereClause = (clause: unknown): clause is WhereClause => {
```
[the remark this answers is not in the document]
dlavrenuek: value is optional, the other properties are validated

packages/nodes-base/nodes/Microsoft/Sql/GenericFunctions.ts
```diff
@@ -114,14 +114,30 @@ export function configurePool(credentials: IDataObject) {
...
+	if (identifier.startsWith('[') && identifier.endsWith(']')) {
+		identifier = identifier.slice(1, -1);
+	}
+
+	return `[${identifier.replaceAll(']', ']]')}]`;
+}
+
+export function escapeTableName(table: string) {
 	table = table.trim();
 	if (table.startsWith('[') && table.endsWith(']')) {
```
yehorkardash, line 127: Some issues with it highlighted [here](https://linear.app/n8n/issue/NODE-4036/#comment-0256a723)

packages/nodes-base/nodes/Postgres/v2/actions/database/select.operation.ts
```diff
@@ -125,7 +129,8 @@ export async function execute(
 
 		const returnAll = this.getNodeParameter('returnAll', i, false);
 		if (!returnAll) {
-			const limit = this.getNodeParameter('limit', i, 50);
+			const limitRaw = this.getNodeParameter('limit', i, 50);
+			const limit = tryToParseNumber(limitRaw);
 			query += ` LIMIT ${limit}`;
```
DawidMyslak: Maybe a better fix would be to rely on built-in `QueryValues` in SQL query:
```ts
  const limit = this.getNodeParameter('limit', i, 50);
  query += ' LIMIT ?';
  values.push(limit);
```

It's overall safer to rely on `?` and bind values later.

packages/nodes-base/nodes/Postgres/v2/helpers/utils.ts
```diff
@@ -626,6 +627,38 @@ export const convertArraysToPostgresFormat = (
 	return newData;
 };
 
+// operations use 'equal' instead of '=' because of the way expressions are handled
+// manually add '=' to allow entering it instead of 'equal'
+const conditionSet = new Set(operatorOptions.map((option) => option.value)).add('=');
+
+export const isWhereClause = (clause: unknown): clause is WhereClause => {
+	if (typeof clause !== 'object' || clause === null) return false;
+	if (typeof clause !== 'object' || clause === null) return false;
```
DawidMyslak, line 635: Duplicated condition check.

packages/nodes-base/nodes/Microsoft/Sql/test/utils.test.ts
```diff
@@ -242,4 +342,34 @@ describe('MSSQL tests', () => {
...
+			expect(escapeTableName('test.test')).toEqual('[test.test]');
+			expect(escapeTableName('[test].[test]')).toEqual('[test].[test]');
+			expect(escapeTableName('[test]--.ok].[[test]]')).toEqual('[test]]--.ok].[[test]]]');
+			expect(escapeTableName("test] SET mytext='1' where id=2; -- -")).toEqual(
+				"[test]] SET mytext='1' where id=2; -- -]",
+			);
+			expect(escapeTableName("[[test] (id,text) values(1,'123'); DROP TABLE users; -- ]")).toEqual(
+				"[[test]] (id,text) values(1,'123'); DROP TABLE users; -- ]",
+			);
+		});
```
DawidMyslak, line 375: Let's add 2 more tests for partially bracketed input:
```ts
  expect(escapeTableName('schema.[table]')).toEqual('[schema.[table]]]');
  expect(escapeTableName('[schema].table')).toEqual('[[schema]].table]');
```
The question is however, do we think the output is correct?

In SQL Server, we only need brackets when an identifier contains:
- spaces: [Order Details]
- special characters: [my-table]
- reserved words: [User], [Index] etc.

So users might naturally write:
```sql
dbo.[Order Details]     -- schema is simple, table has space
sys.[user]              -- schema is simple, table is reserved word
MyDB.dbo.[My Table]     -- only the table needs escaping
```

DawidMyslak (changes requested):

packages/nodes-base/nodes/Postgres/v2/helpers/utils.ts
dlavrenuek: fixed

packages/nodes-base/nodes/Postgres/v2/actions/database/select.operation.ts
dlavrenuek: great idea, added. The syntax is a bit different than I would normally expect it

packages/nodes-base/nodes/Microsoft/Sql/test/utils.test.ts
dlavrenuek: Added the tests.

As for the format parsing: MS supports escaped and unescaped path, detecting the different parts would require parsing the table name character by character to find unescaped closing brackets. The current solution is a compromise between simplicity while still offering the functionality to use full schema path while requiring each part to be enclosed in [] `[MyDB].[dbo].[My Table]`

DawidMyslak: Makes sense. I just hope we won't break any existing workflows 🙏.

DawidMyslak (approved):
Epic work 👏

PR #23135, opened 2025-12-12
fix(ai-builder): Fix error message for large context windows

r00gm:
## Summary

There might be cases when our estimate of input length in tokens falls bellow the threshold, but after sending it to LLM the real token length is more than allowed. We want to show more user-friendly message in such cases, so we override the original error message from model provider.

image

[AI-1805](https://linear.app/n8n/issue/AI-1805/bug-prompt-is-too-long-error-from-builder)

## Review / Merge checklist

- [ ] Tests included. 


burivuhster (approved):
🚀

PR #12756, opened 2025-01-21
fix(core): Update execution entity and execution data in transaction

ivov:
https://n8nio.slack.com/archives/C069HS026UF/p1737393139794529

PR #23146, opened 2025-12-12
fix(core): Fix for execution history when flow includes wait node

DawidMyslak:
## Summary

The workflow execution start time is set via `setRunning` function, which is fired from multiple places. When the workflow is resumed after waiting, the same function is fired which results with overwritten `const startedAt = new Date();` - causing the bug with wrong execution time.

**[Loom Recording](https://www.loom.com/share/f319c9cf5d8a4160a1be65f4ee3d2fd9)**

https://linear.app/n8n/issue/CAT-1854/execution-history-incorrect-when-flow-includes-wait-node

## Review / Merge checklist

- [x] Tests included. 


shortstacked (dismissed):
Looks great, thanks!

DawidMyslak:
@tomi Can I get your thoughts on these changes please 🙏 ?

Explanation:
[Loom Recording](https://www.loom.com/share/f319c9cf5d8a4160a1be65f4ee3d2fd9)

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -418,9 +418,20 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
 	async setRunning(executionId: string) {
 		const startedAt = new Date();
 
-		await this.update({ id: executionId }, { status: 'running', startedAt });
+		// Check if execution already has a startedAt timestamp
```
tomi: I would like us to avoid making an extra query just to check this. It unnecessary extra load and could also cause a race condition (theoretical, this shouldn't be called twice anywhere). We can do this e.g. by coalescing the startedAt timestamp:

```sql
UPDATE execution_entity
SET
  startedAt = COALESCE(startedAt, :startedAt),
  status = 'running'
WHERE id = :id
```

or by ignoring it if it's already running

```sql
UPDATE execution_entity
SET
  startedAt = :startedAt,
  status = 'running'
WHERE id = :id AND status <> 'running'
```

tomi (commented):
Thank you for fixing this 🫶 Really appreciate the loom as well! Made it much easier to review. One comment about avoiding the extra query

packages/@n8n/db/src/repositories/execution.repository.ts
DawidMyslak: Hey @tomi, sorry for a late reply. I like your idea.

The problem I'm having is that even if I optimize the update query, I still need to return the correct `startedAt` inside this function. For example SQLite doesn't allow me to update and select data inside the same query, so I will end up having 2 queries anyway: first one doing the update, second one getting the real `startedAt`. I think PostgreSQL is better in this regard, so technically a single query can do the job. Transaction can also help with grouping 2 queries.

What do you think?

tomi: @DawidMyslak Gotcha, makes sense. In that case I think we could do it in a trx and update it first and query the value afterwards.

DawidMyslak: Thanks again @tomi 🙏 . I've pushed the change.

Just FYI: Using the update first requires a bit more code with `COALESCE`.

If we decide to go for the select first, then potentially the code will be a bit shorter:
```ts
		const fetchThenUpdate = async (manager: EntityManager) => {
			// Check if execution already has a startedAt timestamp
			const execution = await manager.findOne(ExecutionEntity, {
				select: ['startedAt'],
				where: { id: executionId },
			});

			// Only set startedAt if it's not already set (preserves original for resumed executions)
			if (execution?.startedAt) {
				await manager.update(ExecutionEntity, { id: executionId }, { status: 'running' });
				return execution.startedAt;
			} else {
				await manager.update(ExecutionEntity, { id: executionId }, { status: 'running', startedAt });
				return startedAt;
			}
		};
```

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -422,10 +423,34 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+			// Update status, set startedAt only if not already set (preserves original for resumed executions)
+			await manager
+				.createQueryBuilder()
+				.update(ExecutionEntity)
+				.set({
+					status: 'running',
+					startedAt: () => `COALESCE(startedAt, '${startedAt.toISOString()}')`,
+				})
+				.where('id = :id', { id: executionId })
+				.execute();
```
tomi, line 438: We should actually check the result and make sure it affected a row. Then again since we query it in the next step, we can change that to `findOneOrFail` so it throws if there's no execution

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -422,10 +423,34 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+			const result = await manager.findOne(ExecutionEntity, {
+				select: ['startedAt'],
+				where: { id: executionId },
+			});
 
-		await this.update({ id: executionId }, { status: 'running', startedAt });
+			return result?.startedAt ?? startedAt;
+		};
+
+		if (dbType === 'sqlite' && sqliteConfig.poolSize === 0) {
```
tomi: Do we actually need to handle these separately? I would imagine the same code would work for both

DawidMyslak: We have the same code in 2 other places in this file, that why I've used it:
```ts
if (dbType === 'sqlite' && sqliteConfig.poolSize === 0) {
  // TODO: Delete this block of code once the sqlite legacy (non-pooling) driver is dropped.
  // In the non-pooling sqlite driver we can't use transactions, because that creates nested transactions under highly concurrent loads, leading to errors in the database
  // ...
}
```

Sample PR introducing it:
https://github.com/n8n-io/n8n/pull/12756/changes#diff-c8cbb62ca9ab2ae45e5f565cd8c63fff6475809a6241ea0b90acc575615224afR398

packages/@n8n/db/src/repositories/execution.repository.ts
DawidMyslak: Good idea. Pushed.

I can also add something like:
```ts
const { affected } = await manager
  .createQueryBuilder()
 // ...

if (!affected) {
  throw new UnexpectedError('Execution not found', { extra: { executionId } });
}
```

But technically `findOneByOrFail` covers this case as it throws `EntityNotFoundError` if execution doesn't exist.

packages/@n8n/db/src/repositories/execution.repository.ts
tomi: Ah yes. When we still had the SQLite legacy driver, its trx support was completely messed up. That has now been removed in v2, but those places still need cleaning up. Let's unify this and use transactions for both.

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -422,10 +423,31 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+				.update(ExecutionEntity)
+				.set({
+					status: 'running',
+					startedAt: () => `COALESCE(startedAt, '${startedAt.toISOString()}')`,
+				})
+				.where('id = :id', { id: executionId })
+				.execute();
+
+			// Fetch the actual startedAt
+			const result = await manager.findOneByOrFail(ExecutionEntity, { id: executionId });
```
tomi: Since we only need `startedAt` we could select only that

tomi (dismissed):
Thank you addressing all the feedback 🫶 🚀

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -423,9 +424,27 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
 
-		await this.update({ id: executionId }, { status: 'running', startedAt });
+		return await this.manager.transaction(async (manager) => {
+			// Update status, set startedAt only if not already set (preserves original for resumed executions)
+			await manager
+				.createQueryBuilder()
+				.update(ExecutionEntity)
+				.set({
+					status: 'running',
+					startedAt: () => `COALESCE(startedAt, '${startedAt.toISOString()}')`,
```
DawidMyslak: @tomi just a final note before merging, I want to be extra careful.

TypeORM, which we used initially with `await this.update({ id: executionId }, { status: 'running', startedAt });` has a different approach for parsing dates.

From my local DB I can see the following examples:
Image

I think it's using `DateUtils.mixedDateToUtcDatetimeString()` under the hood.

The function I've used, which normally should be fine (`toISOString()`), will lead to a different format:
- TypeORM's `mixedDateToUtcDatetimeString`: `2026-01-15 12:30:45.123` (space separator)
- JavaScript's `toISOString()`: `2026-01-15T12:30:45.123Z (T separator, Z suffix)`

Both are valid ISO 8601 formats. SQLite's datetime functions accept both. The key difference is `T` vs space separator and the trailing `Z`.

I think the safer option would be to go for, mainly for consistency:
```ts
startedAt: () => `COALESCE(startedAt, '${DateUtils.mixedDateToUtcDatetimeString(startedAt)}')`,
```

tomi: 👍 for being thorough! Yeah I think it's good to follow the same convention.

While we are at it, maybe we could also move the startedAt to be a parameter instead of inlined. That way the statement the can get cached better

tomi (approved):
Thank you addressing all the feedback 🫶 🚀

PR #23149, opened 2025-12-12
ci: Optimize Docker image build process

shortstacked:
## Summary

Improves the efficiency of Docker image builds by implementing a whitelist approach in `.dockerignore`. This reduces the build context size significantly, resulting in faster build times.

Removes yarn specific configuration from the base Dockerfile.

Addresses Fortify scan that detects Yarn in image and reports vulnerability. Also Yarn isn't needed so it's safe to remove.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,21 +1,27 @@
-# We want to include the THIRD_PARTY_LICENSES.md file in the Docker image,
-# but not other .md files
-**/*.md
-!**/THIRD_PARTY_LICENSES.md
-**/.env
-.cache
-assets
-node_modules
-packages/node-dev
-packages/**/node_modules
-packages/**/dist
-packages/**/.turbo
-packages/**/*.test.*
-.git
-.github
-!.github/scripts
-*.tsbuildinfo
-docker/compose
-docker/**/Dockerfile
-.vscode
-packages/testing
+# Whitelist approach: ignore everything, then allow only what Docker builds need
+# This reduces build context from ~900MB to just what's required
+
+# Ignore everything first
+*
+
+# === n8n main image (docker/images/n8n/Dockerfile) ===
+!compiled
+!compiled/**
+!THIRD_PARTY_LICENSES.md
+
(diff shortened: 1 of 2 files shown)


docker/images/n8n-base/Dockerfile
```diff
@@ -43,6 +42,8 @@ FROM node:${NODE_VERSION}-alpine
 
 COPY --from=builder / /
 
+RUN rm -rf /opt/yarn* && apk del apk-tools
```
tomi, line 45: Removing this from base image is _technically_ a breaking change. I would argue it shouldn't have been there in the first place, so let's just remove it

.dockerignore
```diff
@@ -1,21 +1,27 @@
...
-.git
-.github
-!.github/scripts
-*.tsbuildinfo
-docker/compose
-docker/**/Dockerfile
-.vscode
-packages/testing
\ No newline at end of file
+# Whitelist approach: ignore everything, then allow only what Docker builds need
```
tomi, line 1: nit: 
```suggestion
# Allowlist approach: ignore everything, then allow only what Docker builds need
```

.dockerignore
```diff
@@ -1,21 +1,27 @@
...
+# This reduces build context from ~900MB to just what's required
+
+# Ignore everything first
+*
+
+# === n8n main image (docker/images/n8n/Dockerfile) ===
```
tomi, line 7: As a follow-up, we could also have [separate ignore files](https://docs.docker.com/build/buildkit/dockerfile-release-notes/#110) for different dockerfiles. The difference is probably negligible, so probably not worth it

tomi (approved):
Awesome! 😍 🚀 Just couple comments

docker/images/n8n-base/Dockerfile
frankywahl: This also breaks custom Docker image builds as highlighted in the documentation https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.executecommand/#run-curl-command as `apk` does not exist anymore

PR #23153, opened 2025-12-12
fix(editor): Allow space key in input fields when holding shift

cstuncsik:
## Summary

This PR fixes an issue where pressing the space bar while holding shift did not insert a space character in:
- Rename node dialogue
- Workflow name input on canvas

The root cause was the canvas panning mode keyboard handler which would activate panning mode on any space key event, including Shift+Space, preventing the space character from being typed in input fields.

**Fix:** Added a check in the switchToPanningMode() function to detect if the user is currently typing in an input field and skip panning mode activation in those cases.

- https://linear.app/n8n/issue/ADO-4216
- https://github.com/bcakmakoglu/vue-flow/issues/1999

## Changes
- Modified Canvas.vue to check if active element is an input field before switching to panning mode
- Ensures space bar works correctly in all input contexts while maintaining panning functionality

## Review / Merge checklist
- [ ] Docs updated or follow-up ticket created (no docs needed for this UX fix)
- [ ] Tests included (keyboard interaction fix, manual testing performed)


packages/frontend/editor-ui/src/features/workflows/canvas/components/Canvas.vue
```diff
@@ -211,6 +211,17 @@ const selectionKeyCode = ref<string | true | null>(isMobileDevice ? 'Shift' : tr
 const isInPanningMode = ref(false);
 
 function switchToPanningMode() {
+	// Don't switch to panning mode if user is typing in an input field
+	const activeElement = document.activeElement;
```
konstantintieber: During my testing with @MiloradFilipovic, the only required code change to fix the bug was the `@keydown.space.stop` on the input element.

I removed this added code in `switchToPanningMode` and still couldn't reproduce the issue in renaming the workflow title or a node name.

Which case does this code cover?

packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -841,6 +841,11 @@ async function onOpenRenameNodeModal(id: string) {
 		let shouldSaveAfterRename = false;
 
 		const handleKeyDown = (e: KeyboardEvent) => {
+			// Stop propagation for space key to prevent VueFlow from intercepting it
+			// when modifier keys (like Shift) are pressed. See: https://github.com/bcakmakoglu/vue-flow/issues/1999
+			if (e.key === ' ') {
```
konstantintieber, line 846: I couldn't trigger the `onOpenRenameNodeModal` function to be called.
What do I need to do in the UI to make this modal open?
I tried referencing a node via a `{{ $('HTTP Request').item.json.data }}` expression, and then renaming the `HTTP Request` node, but the expression just gets auto-updated.

cstuncsik: That opens when you select a node on the canvas and press the Space button (one, short press)

packages/frontend/editor-ui/src/features/workflows/canvas/components/Canvas.vue
cstuncsik: You're right, this is not needed anymore

packages/frontend/editor-ui/src/app/views/NodeView.vue
konstantintieber: Verified that this change is needed for the space key to register in the input in the dialog while shift is pressed.
Nice catch

konstantintieber (approved):
Great stuff

packages/frontend/editor-ui/src/app/views/NodeView.vue
cstuncsik: It is, but kudos go to the AI agent ;)
I'm not sure I could have found it

PR #23155, opened 2025-12-12
fix: Sharepoint file selection correctly applies filter

dlavrenuek:
## Summary

For operations like File -> Download
- Fixes folder search by implementing client-side filtering
- Improves file search by changing filter conditions used for Sharepoint API from `eq 'name'` which does full filename comparison to `startswith(name, 'file')`, which does partial comparison ignoring the upper/lowercase.

To test connect to a sharepoint instance/site which contains over 200 files (can be 0 bytes) and over 200 folders.

Also added escaping for the filter parameters, previously it was possible to "escape" the filter and add other filter params, which should not provide any security issues (only filters what the user has already access to) but it would prevent the filter to find entities with `'` in the name

https://linear.app/n8n/issue/NODE-3882/sharepoint-node-filter-and-pagination-do-not-work

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Microsoft/SharePoint/methods/listSearch.ts
```diff
@@ -34,7 +34,7 @@ export async function getFiles(
 			$select: 'id,name,file',
 		};
 		if (filter) {
-			qs.$filter = `name eq '${filter}'`;
+			qs.$filter = `startswith(name, '${filter.replaceAll("'", "''")}')`;
```
dlavrenuek: see docs: https://learn.microsoft.com/en-us/graph/filter-query-parameter?tabs=http

packages/nodes-base/nodes/Microsoft/SharePoint/methods/listSearch.ts
```diff
@@ -137,7 +134,7 @@ export async function getItems(
 			$select: 'id,fields',
 		};
 		if (filter) {
-			qs.$filter = `fields/Title eq '${filter}'`;
+			qs.$filter = `fields/Title eq '${filter.replaceAll("'", "''")}'`;
```
yehorkardash: let's move `replaceAll("'", "''")` to a helper function

yehorkardash: do we need a similar escaping mechanism when uploading files or other operations inside sharepoint node?

dlavrenuek: Good suggestions, I have moved it to a separate helper function.
I can not spot any other place where it should be required to escape the input, do you?

yehorkardash (approved):

PR #23166, opened 2025-12-12
feat(core): Use draft sub-workflow version for manual execution, published for production

RicardoE105:
## Summary

  This PR implements sub-workflow versioning support to enable iterating on sub-workflows without requiring publication for manual/chat executions, while ensuring production workflows only reference published versions.

  ### Key Changes

  #### 1. Draft Sub-Workflows for Manual/Chat Executions

  - **Feature**: Manual and chat executions now use draft (unpublished) sub-workflow versions
  - **Why**: Enables developers to test and iterate on sub-workflows without publishing them first
  - **Implementation**:
    - Added `useDraftSubWorkflows` flag to `IWorkflowExecuteAdditionalData`
    - Flag is set to `true` for manual/chat execution modes, `false` for production (webhook/trigger)
    - Execute Workflow node respects this flag when fetching sub-workflow data

  #### 2. Sub-Workflow Reference Validation on Publishing

  - **Feature**: Validates that all sub-workflow references are published before publishing a workflow
  - **Why**: Prevents runtime errors in production when a workflow references unpublished sub-workflows
  - **Behavior**:
    - ✅ Allows self-referencing workflows (workflow calling itself)
    - ✅ Skips validation for expressions (e.g., `={{ $json.workflowId }}`)
    - ✅ Skips validation for non-database sources (URL, parameter, localFile)
    - ❌ Blocks publishing if referencing non-existent workflows
    - ❌ Blocks publishing if referencing draft-only workflows
  - **Validation scope**: Direct sub-workflow references (1-level deep)
    - Does not recursively validate sub-workflows that call other sub-workflows
    - This is intentional to keep implementation simple; can be extended later if needed

  #### 3. Frontend Improvements

  - **Resource Locator**: Fixed alignment issues in dropdown component
  - **Workflow Selector**: Updated to work with new validation
  - **i18n**: Updated error messages for sub-workflow validation

  ### How It Works

  **Manual/Chat Execution Flow:**
  User clicks "Test Workflow" → useDraftSubWorkflows = true →
  Execute Workflow node fetches latest draft version → Executes with draft data

  **Production Execution Flow:**
  Webhook/Trigger fires → useDraftSubWorkflows = false →
  Execute Workflow node fetches published version → Executes with published data

  **Publishing Flow:**
  User clicks "Publish" → Validates all sub-workflow references are published →
  If valid: publishes workflow | If invalid: shows error with specific missing workflows

  ### Comparison with V1

  | Feature | V1 (pre-versioning) | V2 (this PR) |
  |---------|---------------------|--------------|
  | Sub-workflow validation | None - runtime errors | Validates at publish time |
  | Draft testing | Not applicable | Manual/chat use drafts |
  | Production execution | Always used latest | Always uses published version |
  | Terminology | "Activate" workflow | "Publish" workflow |

  ### User Experience

  **Before this PR:**
  - Developers had to publish sub-workflows to test them in parent workflows
  - Runtime errors in production if sub-workflow wasn't published
  - No validation at publish time

  **After this PR:**
  - Test button uses draft sub-workflows (fast iteration)
  - Publish button validates all sub-workflows are published (prevents runtime errors)
  - Clear error messages guide users to publish dependencies first

  https://linear.app/n8n/issue/ADO-4535/cannot-use-a-sub-worfklow-in-the-same-workflow

  ## Review / Merge checklist

  - [x] Tests included (469 new test cases)
  - [ ] PR Labeled with `release/backport` (if applicable)


packages/cli/src/workflow-runner.ts
```diff
@@ -253,6 +253,9 @@ export class WorkflowRunner {
 		// TODO: set this in queue mode as well
 		additionalData.restartExecutionId = restartExecutionId;
 		additionalData.streamingEnabled = data.streamingEnabled;
+		// Use draft sub-workflows for manual/chat executions to enable
+		// iterating on sub-workflows without requiring them to be published
+		additionalData.useDraftSubWorkflows = ['manual', 'chat'].includes(data.executionMode);
```
dariacodes: I see we're using this already in two places - would be nice to put this in a util

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -940,4 +941,44 @@ export class WorkflowService {
...
+					{ relations: [] },
+				);
+
+				if (!subWorkflow) {
+					return { exists: false, isPublished: false };
+				}
+
+				return {
+					exists: true,
+					isPublished: subWorkflow.activeVersionId !== null,
```
dariacodes: what if this sub-workflow calls another sub-workflow? should we recursively validate all of them? we could introduce a set of `visitedWorkflows` or  a limit for `depth`

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -26,4 +40,105 @@ export class WorkflowValidationService {
...
 	}
+
+	/**
+	 * Validates that all sub-workflow references in a workflow are published.
+	 *
+	 * This prevents publishing a parent workflow that references draft-only sub-workflows,
+	 * which would cause runtime errors when the workflow is triggered in production.
+	 *
+	 */
+	async validateSubWorkflowReferences(
```
dariacodes, line 54: I think you need to mock this in integration tests, something like
`workflowValidationService.validateSubWorkflowReferences.mockResolvedValue({ isValid: true });`

dariacodes (commented):
Nice changes! I only reviewed the code so far, ping me once we agree on the final implementation and I'll test it

packages/cli/src/workflows/workflow.service.ts
RicardoE105: I do not think we should do this just yet. Will overcomplicate things and this is basically a new behavior. I will merge this one and then think about the other one.

dariacodes: Actually, it might be even unnecessary as it's impossible to publish a workflow that is calling an unpublished sub-workflow, so we should never have a situation where a sub-workflow is published and calling an unpublished sub-workflow inside at the same time.

dariacodes (dismissed):
I checked that when running manually, the latest version is used and during prod execution the published version is used 🚢

packages/cli/src/workflow-execute-additional-data.ts
```diff
@@ -158,9 +174,16 @@ export async function executeWorkflow(
 ): Promise<ExecuteWorkflowData> {
 	const activeExecutions = Container.get(ActiveExecutions);
 
+	// Use draft sub-workflows for manual/chat executions to enable
+	// iterating on sub-workflows without requiring them to be published
+	const useDraftVersion = options.executionMode
```
mfsiega: Nit: maybe just `options.executionMode && isManualOrChatExecution(options.executionMode)`. Or even simpler, make `isManualOrChatExecution` handled undefined executionMode by returning false?

packages/cli/src/workflow-execute-additional-data.ts
```diff
@@ -94,12 +95,20 @@ export function getRunData(
...
+ *
+ * When `useDraftVersion` is false (production executions):
+ * - Requires the workflow to have an active (published) version
+ * - Uses nodes/connections from the activeVersion in WorkflowHistory
  */
 export async function getWorkflowData(
 	workflowInfo: IExecuteWorkflowInfo,
 	parentWorkflowId: string,
 	parentWorkflowSettings?: IWorkflowSettings,
+	options?: { useDraftVersion?: boolean },
```
mfsiega: Nit: another option would be to get rid of the boolean param, and have two functions (`getDraftWorkflowData` and `getPublishedWorkflowData`), and switch between them at the call-site. The rest of the fn could be pulled into a private helper.

I think booleans that switch the behaviour are generally a bit less readable than two functions, but I don't feel super strongly.

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -571,6 +571,7 @@ export class WorkflowService {
 		}
 
 		this._validateNodes(workflowId, versionToActivate.nodes);
+		await this._validateSubWorkflowReferences(workflowId, versionToActivate.nodes);
```
mfsiega, line 574: Two high-level thoughts:
1. It feels a _bit_ weird that a workflow can pass this validation, but then one of the sub-workflows gets de-activated and it fails during execution anyway. I could also imagine a case where you have some branch of your workflow that you know won't get executed in prod but you don't want to delete it, or something. I wonder if this should be a UI warning instead of completely preventing publication?
2. Haven't tested this myself but it seems like a disconnected (unreachable) or disabled ExecuteWorkflow node would still fail validation, even though the workflow could execute perfectly fine. Should we tolerate those cases? Disabled nodes are easy, but it's not as simple to find unreachable nodes (you have to find the ancestors, and check if there's at least one trigger, I guess?). Maybe this should be a follow-up at least?

RicardoE105: Good questions. Will run it by product and decide.

RicardoE105: Ok product decided to keep it as it's for now.

https://n8nio.slack.com/archives/C04CPHD7VDW/p1766089080784779

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -979,4 +980,44 @@ export class WorkflowService {
...
 	}
+
+	/**
+	 * Validates that all sub-workflow references in a workflow are published.
+	 * Prevents publishing a parent workflow that references draft-only sub-workflows.
+	 */
+	private async _validateSubWorkflowReferences(workflowId: string, nodes: INode[]) {
+		const validation = await this.workflowValidationService.validateSubWorkflowReferences(
+			nodes,
+			async (subWorkflowId: string) => {
```
mfsiega: It seems slightly odd to pass this callback in, instead of just putting the code in the workflow validation service, was there a particular reason? It also seems like maybe this should be called from `validateNodes`, rather than alongside it? but not so important.

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -979,4 +980,44 @@ export class WorkflowService {
 
 		return Object.keys(changes).length > 0 ? changes : undefined;
 	}
+
+	/**
+	 * Validates that all sub-workflow references in a workflow are published.
+	 * Prevents publishing a parent workflow that references draft-only sub-workflows.
+	 */
```
mfsiega, line 996: It might be worth mentioning the cases where a published workflow could end up referencing a draft-only sub-workflow:
- Sub-workflow gets un-published
- Workflow id is an expression
- Others?

and also mention that we further enforce the invariant when the workflow is executed, to cover those cases. 

Otherwise a reader might wonder about these things (as I did).

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -26,4 +40,113 @@ export class WorkflowValidationService {
...
+			const source: string | undefined =
+				typeof node.parameters?.source === 'string' ? node.parameters.source : undefined;
+
+			if (this.shouldSkipSubWorkflowValidation(workflowId, source)) {
+				continue;
+			}
+
+			const status = await getWorkflowStatus(workflowId!);
+
+			if (!status.exists) {
```
mfsiega: Nit: I think this would be marginally more readable as `if (status.doesNotExist || status.draftOnly)` (would involve a small change to the getWorkflowStatus fn to reverse the conditions).

packages/testing/playwright/tests/e2e/api/subworkflow-version-resolution.spec.ts
mfsiega, line 1: Really nice tests 👏

packages/testing/playwright/tests/e2e/api/subworkflow-version-resolution.spec.ts
```diff
@@ -0,0 +1,504 @@
...
+				workflowData: parentWorkflow,
+				triggerToStartFrom: { name: 'Manual Trigger' },
+			},
+		});
+
+		expect(manualExecResponse.ok()).toBe(true);
+		const manualExecution = await manualExecResponse.json();
+
+		// Wait for execution to complete
+		await new Promise((resolve) => setTimeout(resolve, 2000));
```
mfsiega: I think we have a helper for this already, `waitForExecution`? I see you're using it below, it seems like it would be applicable here.

packages/testing/playwright/tests/e2e/api/subworkflow-version-resolution.spec.ts
```diff
@@ -0,0 +1,504 @@
...
+
+		// Get and verify execution result
+		const executionDetails = await api.workflows.getExecution(manualExecution.data.executionId);
+		const executionData = helpers.parseExecutionData(executionDetails.data);
+
+		const actualValue = helpers.extractValueFromExecutionData(executionData);
```
mfsiega: I wonder if it would be easier to use the Respond to Webhook node? You can configure it to just send back the output of the sub-workflow - so you just get the webhook response. It would be nice to get rid of the complicated parsing you had to do here.

I attached some sample workflows for what I had in mind:
[Child (1).json](https://github.com/user-attachments/files/24261657/Child.1.json)
[Parent (2).json](https://github.com/user-attachments/files/24261658/Parent.2.json)

packages/testing/playwright/tests/e2e/api/subworkflow-version-resolution.spec.ts
mfsiega: Also: there's already a subdirectory for subworkflow-related tests (`packages/testing/playwright/tests/e2e/workflows/editor/subworkflows`), maybe this belongs there?

packages/testing/playwright/tests/e2e/api/subworkflow-version-resolution.spec.ts
```diff
@@ -0,0 +1,504 @@
...
+ * 4. Self-referencing workflows can be published
+ */
+
+// Helper functions to reduce duplication
+const helpers = {
+	/**
+	 * Creates a sub-workflow with Execute Workflow Trigger and Set Value nodes
+	 */
+	async createSubWorkflow(api: ApiHelpers, name: string, value: string) {
+		return await api.workflows.createWorkflowFromDefinition({
```
mfsiega: A lot of other workflow-related tests put the workflow definitions in a separate file - would it make sense to apply that pattern here? You can check some of the other sub-workflow tests and see if it looks nicer.

packages/testing/playwright/tests/e2e/api/subworkflow-version-resolution.spec.ts
```diff
@@ -0,0 +1,504 @@
...
+				},
+			},
+		});
+	},
+
+	/**
+	 * Extracts the 'value' field from the complex flattened execution data structure
+	 * returned by n8n's Execute Workflow node
+	 */
+	extractValueFromExecutionData(executionData: unknown[]): string {
```
mfsiega: I'm *hoping* we can get rid of this :)

mfsiega (commented):
Just some style/readability/test comments, generally lgtm!

packages/cli/src/workflow-execute-additional-data.ts
mfsiega, line 1: It would be nice to have unit test coverage on `executeWorkflow` - then we could verify that we're using the draft/published version appropriately based on the execution mode; doesn't need to block this since it way already missing, but we should try to make sure we add it as a follow-up.

packages/testing/playwright/tests/e2e/workflows/editor/subworkflows/subworkflow-version-resolution.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+			data: {
+				workflowData: parentWorkflow,
+				triggerToStartFrom: { name: 'When clicking Test workflow' },
+			},
+		});
+
+		expect(runResponse.ok()).toBe(true);
+
+		const execution = await api.workflows.waitForExecution(parentWorkflowId, 10000, 'manual');
+		expect(execution.status).toBe('success');
```
mfsiega, line 122: Cubic's comment here _seems_ valid.

You could do something like:
- Duplicate the parent workflow definition that you're using the prod test below, but set `"active": false` (when this is true, importWorkflowFromFile automatically publishes the workflow).
- When you execute it manually, instead of an execution id you'll get back a response like `{ needsWebhook: true }`.
- Then you can trigger the test webhook endpoint which should be treated the same as a manual execution, and the rest of the test looks basically the same.

There might be simpler/cleverer options, but that seems the most straightforward.

If you've manually verified that this works as intended (for both manual executions and test webhook executions) it's probably okay to leave as-is and improve the test in a follow-up. We're also missing good unit test coverage here, so adding more tests *somewhere* would be good.

mfsiega (dismissed):
A few testing follow-ups, but I don't think they have to block this PR. LGTM!

packages/testing/playwright/tests/e2e/workflows/editor/subworkflows/subworkflow-version-resolution.spec.ts
mfsiega: I like what you did here, nice!

mfsiega (approved):
Thank you for bearing with me! 🚢

PR #23187, opened 2025-12-13
fix: Improve webhook trigger node tests

shortstacked:
Refactors webhook trigger node tests for better reliability and efficiency.
- Removes unnecessary deregistration of all webhooks, deregistering only the relevant webhook after test completion, preventing interference with other running workflows.

- Introduces `WebhookApiHelper` to handle webhook triggering with retry logic for asynchronous registration scenarios and connection errors.
- Adds debounce to parameter input to ensure the parameter persists before webhook registration.
- Consolidates HTTP method testing and removes skipped tests.

## Summary

https://linear.app/n8n/issue/CAT-1961/fix-webhook-timing-issues-in-e2e-tests-for-multi-main-mode

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/webhooks/test-webhooks.ts
```diff
@@ -500,9 +500,10 @@ export class TestWebhooks implements IWebhookManager {
 			if (staticData) workflow.staticData = staticData;
 
 			await this.webhookService.deleteWebhook(workflow, webhook, 'internal', 'update');
-		}
 
-		await this.registrations.deregisterAll();
+			// Deregister only this webhook, not all webhooks from other running workflows
+			await this.registrations.deregister(webhook);
```
mfsiega, line 505: Just want to double-check that this is a safe change - it seems great to me. Would be nice to add an e2e test:
- Create two workflows with webhooks.
- Execute them both manually (this should start listening on the test webhook endpoint).
- Call one of the endpoints, and then the other. Both should succeed.
It seems that today after you call one endpoint, the other indeed returns a 404 because it's been deregistered.

Feel free to add in a follow-up or just open a ticket as well to track.

mfsiega (approved):
LGTM!

One nit: the fix in test-webhooks is a real fix as far as I can tell (it's real code used for "test executions", not test code), and it's kind of sneaking through with the actual test-only fixes. We might want to highlight that change, or even pull it into a separate PR.

shortstacked:
> LGTM!
> 
> One nit: the fix in test-webhooks is a real fix as far as I can tell (it's real code used for "test executions", not test code), and it's kind of sneaking through with the actual test-only fixes. We might want to highlight that change, or even pull it into a separate PR.

I had thought of that, I mentioned it in the PR description too. Part of why I'm bringing them in together is that the e2e tests also test the fix as well, since they run in parallel with several test webhooks the issue would be caught in this PR.

PR #23213, opened 2025-12-15
fix(editor): Workflow sharing modal 'no users found' message

burivuhster:
## Summary

Changed message in workflow sharing modal when no results found.

- https://linear.app/n8n/issue/PAY-4153/incorrect-copy-in-the-search-input-when-sharing-a-workflow#comment-b421af18

## Review / Merge checklist

- [ ] Tests included. 


konstantintieber (dismissed):

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -3777,7 +3776,7 @@
 	"projects.settings.member.added.title": "Member added successfully",
 	"projects.sharing.allUsers": "All users and projects",
 	"projects.sharing.noMatchingProjects": "There are no available projects",
-	"projects.sharing.noMatchingUsers": "No matching users or projects",
+	"projects.sharing.noMatchingUsers": "No matching users",
```
ireneea: The naming is misleading but the default value is correct 😅. I think the actual fix would be to pass the right`emptyOptionsText` property to the `<ProjectSharing>`  component when used in `<WorkflowShareModal>`

ireneea (commented):
Just had a small comment

packages/frontend/@n8n/i18n/src/locales/en.json
burivuhster: Good point! I updated the fix

ireneea (approved):

PR #23228, opened 2025-12-15
feat(editor): Add unpublish to workflow list

RobTF9:
## Summary
Add the ability to unpublish a workflow directly from the workflow list page.

## Review / Merge checklist

- [ ] Tests included. 


RobTF9:
@claude review

packages/frontend/editor-ui/src/app/components/WorkflowCard.vue
```diff
@@ -56,6 +59,7 @@ const WORKFLOW_LIST_ITEM_ACTIONS = {
 	MOVE_TO_FOLDER: 'moveToFolder',
 	ENABLE_MCP_ACCESS: 'enableMCPAccess',
 	REMOVE_MCP_ACCESS: 'removeMCPAccess',
```
nikhilkuria, line 59: Then let's add publish here as well

RobTF9: I think publish is more complicated because we'd need to either:
- Only publish the last published version, if there is none throw an error
- Publish the current version even if not previously published, or
- Display a select that allows a user to pick any version to publish (complicated)

Unpublish is the most important thing to be able to access quickly and what we've gotten feedback on

nikhilkuria (approved):
Nice work 👏

PR #23234, opened 2025-12-15
feat: Autosave Collaboration - Lock editing for spectators

dariacodes:
### Description
Implement write locks for users editing workflows. The lock is released after 20 seconds of inactivity.

- Fixed the sharing e2e by opening workflows page after each test (this releases the write lock)

Explanation of some solutions/decisions:
1. The `workflowUpdated` event is handled directly in the collaboration store instead of going through the `usePushConnection/handlers` pattern because it needs collaboration-specific state (like write locks, spectator mode and canvas refresh callbacks). It also has a different lifecycle - it’s tied to the `CollaborationPane` component rather than being global.

2. The collaboration store doesn’t talk to the canvas directly. Instead, it exposes a callback that the canvas hooks into. This avoids circular dependencies and keeps a clean split between business logic (the store) and view logic (the component). The store decides when a refresh should happen, and the NodeView handles how it happens, like restoring UI state or reopening the NDV.

3. Fix stuck workflow write locks with TTL + heartbeat + expiration detection
When a writer’s tab crashes/goes offline, the workflow write lock could previously remain held indefinitely, blocking collaborators until manual backend intervention. So these were introduced:
    - Backend write-lock TTL: 120s (2 minutes i.e. allows ~3 missed heartbeats) - all write locks in cache expire automatically unless renewed.
    - Writer heartbeat: every 30s the lock holder sends a `writeAccessHeartbeat` message to renew the TTL.
    - Spectator polling: every 20s read-only clients poll lock state and clear stale `currentWriterId` when the backend lock has expired.

4. We send telemetry for audit in `log-streaming.event-relay` where we could put `workflowUpdated`, but as it is only needed for collaboration to work correctly, it is not needed here. 

Closes [ADO-4284](https://linear.app/n8n/issue/ADO-4284/feature-collaboration-lock-editing-for-spectators)

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/composables/useActivityDetection.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function useActivityDetection() {
+	const collaborationStore = useCollaborationStore();
+
+	const recordActivity = useDebounceFn(() => {
+		console.log('[ActivityDetection] 🖱️ User activity detected, recording...');
+		collaborationStore.recordActivity();
+	}, 100);
+
+	const events = ['mousedown', 'keydown', 'touchstart'];
```
dariacodes, line 12: FYI We might want to expand this

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -6,11 +6,10 @@ import type { IWorkflowDb } from '@/Interface';
 import type { PermissionsRecord } from '@n8n/permissions';
```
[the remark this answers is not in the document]
dariacodes: `saveCurrentWorkflow` never throws, it just returns `false`

packages/cli/src/collaboration/collaboration.message.ts
```diff
@@ -1,6 +1,11 @@
 import { z } from 'zod';
 
-export type CollaborationMessage = WorkflowOpenedMessage | WorkflowClosedMessage;
+export type CollaborationMessage =
+	| WorkflowOpenedMessage
+	| WorkflowClosedMessage
+	| WriteAccessRequestedMessage
```
scdekov, line 6: Should we call this `RequestWriteAccessMessage`,

packages/cli/src/collaboration/collaboration.message.ts
```diff
@@ -1,6 +1,11 @@
...
+	| WriteAccessReleaseRequestedMessage
```
scdekov, line 7: Do we have to "request" the release? I guess that could be just `ReleaseWriteAccessMessage`

packages/cli/src/collaboration/collaboration.message.ts
```diff
@@ -16,15 +21,49 @@ export const workflowClosedMessageSchema = z
 	})
 	.strict();
 
+export const writeAccessRequestedMessageSchema = z
+	.object({
+		type: z.literal('writeAccessRequested'),
+		workflowId: z.string().min(1),
+		userId: z.string().min(1),
```
scdekov: Do we need to send the `userId`? I guess that should be already in the request context data?

packages/cli/src/collaboration/collaboration.service.ts
```diff
@@ -101,4 +122,146 @@ export class CollaborationService {
 
 		this.push.sendToUsers({ type: 'collaboratorsChanged', data: msgData }, userIds);
 	}
+
+	private async handleWriteAccessRequested(userId: User['id'], msg: WriteAccessRequestedMessage) {
+		const { workflowId, userId: requestedUserId } = msg;
+
+		if (userId !== requestedUserId) {
```
scdekov: Can we drop the `userId` from the message and use the `userId` param?

packages/cli/src/collaboration/collaboration.service.ts
```diff
@@ -101,4 +122,146 @@ export class CollaborationService {
...
+		await this.sendWriteAccessAcquiredMessage(workflowId, userId);
+	}
+
+	private async handleWriteAccessReleaseRequested(
+		userId: User['id'],
+		msg: WriteAccessReleaseRequestedMessage,
+	) {
+		const { workflowId } = msg;
+
+		if (!(await this.accessService.hasReadAccess(userId, workflowId))) {
```
scdekov: Do we need this check? I guess if that user have the lock they do have read(and write) access so we can spare a few db calls here in favor of the cache check which should be much cheaper.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -35,7 +35,24 @@ const props = defineProps<{
...
 
 const actionsMenuRef = useTemplateRef<InstanceType<typeof ActionsDropdownMenu>>('actionsMenu');
-const locale = useI18n();
+
+const readOnlyForPublish = computed(() => {
+	if (props.isNewWorkflow) return props.readOnly;
+	return (
+		props.readOnly ||
+		props.isArchived ||
+		(!props.workflowPermissions.update && !props.workflowPermissions.publish)
```
scdekov: This doesn't seem right. I guess if either is false we should disabled publish rather than both having to be false.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -178,35 +275,45 @@ defineExpose({
...
+					:disabled="!publishButtonConfig.enabled || readOnlyForPublish"
 					type="secondary"
 					data-test-id="workflow-open-publish-modal-button"
 					@click="onPublishButtonClick"
 				>
-					{{ locale.baseText('workflows.publish') }}
+					<div :class="[$style.flex]">
+						<span
+							v-if="publishButtonConfig.showIndicator"
+							:class="[
```
scdekov: nit: use an object instead of an array

packages/frontend/editor-ui/src/app/composables/useActivityDetection.ts
```diff
@@ -0,0 +1,30 @@
+import { onMounted, onUnmounted } from 'vue';
+import { useDebounceFn } from '@vueuse/core';
+import { useCollaborationStore } from '@/features/collaboration/collaboration/collaboration.store';
+
+export function useActivityDetection() {
+	const collaborationStore = useCollaborationStore();
+
+	const recordActivity = useDebounceFn(() => {
+		console.log('[ActivityDetection] 🖱️ User activity detected, recording...');
+		collaborationStore.recordActivity();
```
scdekov, line 9: I think it may be a good idea to lock for smaller period of time instead of full new 20 seconds on those actions. Added a feature suggestion in the notion page.

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
```diff
@@ -147,6 +148,19 @@ export function useWorkflowSaving({
 		const parentFolderId = getQueryParam(router.currentRoute.value.query, 'parentFolderId');
 		const uiContext = getQueryParam(router.currentRoute.value.query, 'uiContext');
 
+		// Prevent concurrent saves - if a save is already in progress, skip this one
+		// for autosaves (they will be rescheduled), or wait for non-autosaves
+		if (uiStore.isActionActive.workflowSaving) {
```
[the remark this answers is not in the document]
scdekov: That's fair but 2 concurrent manual saves shouldn't be possible from the UI while save + autosave is easily possible. I think it may not be worth adding the complexity here to handle 2 manual saves.

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
```diff
@@ -203,11 +217,7 @@ export function useWorkflowSaving({
 				workflowState.setWorkflowName({ newName: workflowData.name, setStateDirty: false });
 			}
 
-			if (tags) {
-				const createdTags = (workflowData.tags || []) as ITag[];
-				const tagIds = createdTags.map((tag: ITag): string => tag.id);
-				workflowState.setWorkflowTagIds(tagIds);
-			}
+			workflowState.setWorkflowTagIds(convertWorkflowTagsToIds(workflowData.tags));
```
scdekov: That worked slightly differently in the previous version ie no setWorkflowTagIds call if tags are not passed so we may have to double check it doesn't introduce a bug.

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -1,17 +1,26 @@
+/* eslint-disable import-x/extensions */
```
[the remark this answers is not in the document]
dariacodes: This code was already there, I didn't touch it, but I'll test the behavior

packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -1869,10 +1880,33 @@ watch(
...
 );
 
+// Acquire write access when user makes first edit, and trigger auto-save
 watch(
 	() => uiStore.dirtyStateSetCount,
-	() => {
-		void workflowSaving.autoSaveWorkflow();
+	(dirtyStateSetCount) => {
+		if (dirtyStateSetCount > 0) {
+			collaborationStore.requestWriteAccess();
```
scdekov, line 1888: It seems possible that `requestWriteAccess` fails and we still trigger an autoSave. The proper approach here is probably to use http endpoint for requesting write access that returns if the access is granted or not. However this can't really cause big problems as the backend will still stop the save request if someone else is editing so I guess we don't have to fix this immediately, but maybe log it somewhere?

packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -1869,10 +1880,33 @@ watch(
...
+
+			// Trigger auto-save (debounced) for writers only
+			// Skip auto-save when AI Builder is streaming to keep version history clean
+			if (!collaborationStore.shouldBeReadOnly && !builderStore.streaming) {
```
scdekov: Do we need this `!collaborationStore.shouldBeReadOnly` condition? It shouldn't be possible to trigger this watcher if in read only mode. I'd either remove it or make it throw an error if the state is readonly at this point.

packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -1869,10 +1880,33 @@ watch(
...
+);
+
+// Trigger auto-save when AI Builder streaming ends with unsaved changes
+watch(
+	() => builderStore.streaming,
+	(isStreaming, wasStreaming) => {
+		// Only trigger when streaming just ended (was streaming, now not)
+		if (wasStreaming && !isStreaming) {
+			// Trigger auto-save if there are unsaved changes
+			if (uiStore.stateIsDirty && !collaborationStore.shouldBeReadOnly) {
```
scdekov, line 1906: It's technically possible be in readonly mode here if the builder thought for too long so in that case we should probably show an error toast or something instead of silently skipping the save

packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -1938,6 +1972,24 @@ onMounted(() => {
...
+	// Register callback for collaboration store to refresh canvas when workflow updates arrive
+	collaborationStore.setRefreshCanvasCallback(async (workflow) => {
+		const activeNodeName = ndvStore.activeNode?.name;
+
+		// Refresh the canvas with updated workflow
+		await initializeWorkspace(workflow);
+
+		// Reopen NDV if it was open (with a small delay to ensure workflow is fully initialized)
+		if (activeNodeName) {
+			setTimeout(() => {
```
scdekov: Is this really needed? Looking at the UX right now the NDV stays open the whole time while I'm an observer and someone else is editing that node

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+		lastActivityTime.value = Date.now();
+	}
+
+	function requestWriteAccess() {
+		if (shouldBeReadOnly.value) {
+			console.log('[Collaboration] ❌ Write access denied - another user is writing', {
+				currentWriter: currentWriterId.value,
+				requestingUser: usersStore.currentUserId,
+			});
+			return false;
```
scdekov: Should we throw here instead of returning false? There shouldn't be a case where we try to request access while the state is readonly.

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+		});
+
+		try {
+			pushStore.send({
+				type: 'writeAccessRequested',
+				workflowId: workflowsStore.workflowId,
+				userId: usersStore.currentUserId,
+			});
+		} catch (error) {
+			console.error('[Collaboration] ❌ Failed to send writeAccessRequested message:', error);
```
scdekov: return false/rethrow

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+		}
+
+		return true;
+	}
+
+	function releaseWriteAccess() {
+		if (!isCurrentUserWriter.value) {
```
scdekov: Shouldn't be needed as we're checking this in checkInactivity

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+		if (!isCurrentUserWriter.value) return;
+
+		// Don't release write lock while AI Builder is streaming
+		if (builderStore.streaming) {
+			console.log('[Collaboration] ⏱️ Skipping inactivity check - AI Builder is streaming');
+			return;
+		}
+
+		const timeSinceActivity = Date.now() - lastActivityTime.value;
+		const timeoutThreshold = 20 * TIME.SECOND;
```
scdekov: Move to a constant on top of the file

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+
+	async function handleWorkflowUpdate() {
+		if (isCurrentUserWriter.value) {
+			return;
+		}
+
+		// Only skip updates if the user has write access AND has unsaved changes
+		// Readers should always receive updates since they can't make changes
+		if (!shouldBeReadOnly.value && uiStore.stateIsDirty) {
+			console.log('[Collaboration] ⚠️ Skipping workflow update - local changes exist');
```
scdekov: That shouldn't be possible. Throw an error

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+		try {
+			console.log('[Collaboration] 📥 Fetching updated workflow...');
+			// Fetch the latest workflow data
+			const updatedWorkflow = await workflowsStore.fetchWorkflow(workflowsStore.workflowId);
+
+			// Refresh the canvas with the new workflow data
+			if (refreshCanvasCallback) {
+				refreshCanvasCallback(updatedWorkflow);
+			} else {
+				// Fallback to just updating the store
```
scdekov: Is this a valid case?

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -53,25 +68,333 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
...
+				event.type === 'writeAccessAcquired' &&
+				event.data.workflowId === workflowsStore.workflowId
+			) {
+				currentWriterId.value = event.data.userId;
+				const writer = collaborators.value.find((c) => c.user.id === event.data.userId);
+				console.log(
+					'[Collaboration] 🔓 Write access acquired by:',
+					writer?.user.email || event.data.userId,
+				);
+				recordActivity();
```
scdekov: This should also be done only if the currentUser is the writer

scdekov (changes requested):

packages/frontend/editor-ui/src/app/views/NodeView.vue
scdekov: Actually I just saw that we'll not be releasing the log if the builder is streaming so we may be good here.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: Good catch! I think I was refactoring and this is how it was working before, but I see that project editor has both roles, so I think it'll be safe to have "or" here

packages/frontend/editor-ui/src/app/composables/useActivityDetection.ts
dariacodes: I thought 20 might be a good number, because sometimes you need to think for at least 5 seconds + you might have a slow connection

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: Yes, I wanted not to hinder reader's action during the observation process, so this ensured that when updates come, the NDV will be still opened. Otherwise, the UI will be glitching a lot. Do you have concerns with this approach?

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
dariacodes: Yeah, already done

packages/cli/src/collaboration/collaboration.message.ts
dariacodes: I'd rather not, because there was already a pattern of naming messages (`WorkflowOpenedMessage`) using past tense

packages/cli/src/collaboration/collaboration.message.ts
dariacodes: I named it this way intentionally to distinguish between client commands and server responses:
```
- WriteAccessRequestedMessage → 'writeAccessRequested'
- WriteAccessReleaseRequestedMessage → 'writeAccessReleaseRequested'
```

packages/cli/src/collaboration/collaboration.message.ts
dariacodes: Good point!

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: actually, when parameters are updated, this change is not reflected when the ndv is opened for the observer 🫠

scdekov: Actually.. :D looks like for some nodes parameters are updated for observers and for others not 🫠🫠

scdekov: No concerns I was just wondering if we need this code at all as I see there's this timeout for re-opening the ndv but I don't see any flashing or anything so I guessed it's not being closed at all.

dariacodes: Even better lol
Could you remove it and see what happens (I can also do that, a bit later)? I remember I needed it, but the code is changing so fast that who knows, maybe we can get rid of it? 😆

scdekov: I just removed it and seems to be working the same (the node stays open for the observer)

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
dariacodes: Tbh, I can't remember what I was trying to solve with this 😄 Maybe some edge-case where you started editing, your save was in progress, but someone else got the write lock first, but I think it's safe to remove this. If needed, we can address edge-cases later, this only overwhelms the code right now

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
dariacodes: We have the condition for it inside `recordActivity` function

scdekov: Yep it's a refactoring suggestion rather than anything breaking. It's weird in the handler for write access acquired (could be by anyone) to read `recordActivity()` without conditions. Having the condition within the `recordActivity` also sounds weird as you shouldn't be calling recordActivity if the user is not a writer.

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
dariacodes: I agree that this is unnecessary

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
dariacodes: Thanks for bringing it up! I pushed even a better [improvement](https://github.com/n8n-io/n8n/pull/23234/commits/da8c610dbe893e744e23043532b4819bc3f38e1d) - to only attach listeners for writers

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: Is it possible that with this code the node sometimes gets updated, sometimes not, but without this, it never gets updated?

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: yeah, let's leave it then

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
dariacodes: Tbh, I have no idea why we are updating the state with name and tags only if name or tags were sent. I'll try to investigate this

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: Good point! I'll see if I can quickly fix this

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: Nope, we don't

packages/frontend/editor-ui/src/app/views/NodeView.vue
scdekov: I can look a little more but it seems more like with or without this code eg for Set node updates are never broadcasted and for if node or webhook updates are always broadcasted.

packages/frontend/editor-ui/src/app/views/NodeView.vue
dariacodes: I have the code, but it's too much for this PR and it needs to be properly tested. Let's leave it this way, merge all of it and I'll open a new PR with these changes. Nice catch!

dariacodes: I also made it so that no write access is requested when there is only one collaborator: [1c2a07f](https://github.com/n8n-io/n8n/pull/23234/commits/1c2a07fa7e048531ffb6ff78b1d110652f60f40f)

scdekov: Isn't it better to keep this. Otherwise if a user makes a change on wf that only he's looking at and on the next second someone opens it they'll be able to immediately take control 🤔

dariacodes: Which is fine no? 😆

dariacodes: If someone joined the session (and they have access to this workflow), I bet it's fine to let them edit straightaway

dariacodes: But if the writer continues to edit, they'll get the lock immediately.

scdekov: Well no as the first user may be in the middle of a change. It could be even in a middle of editing a single input and they get locked out of nowhere

dariacodes: Actually, if we were tracking activity when typing, it could be that the write lock would be gotten immediately (the second user can't be faster in this case), but we track activity only after going out of inputs, so yeah, this might have been a bad idea. I'll write this down and revert

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
dariacodes: upd.: I just put the condition back to avoid any regressions, but I still haven't checked why we have these

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: ✅

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -41,15 +41,17 @@ const readOnlyForPublish = computed(() => {
...
+		!props.workflowPermissions.publish
 	);
 });
 
 const shouldHidePublishButton = computed(() => {
 	return (
 		props.readOnly ||
 		props.isArchived ||
-		(!props.workflowPermissions.publish && !props.workflowPermissions.update)
+		!props.workflowPermissions.publish ||
```
scdekov: For this one I'm not sure tbh we may have to check with Rob. I guess we may want to show disabled button if the user has update but no publish permission

packages/frontend/editor-ui/src/app/composables/useActivityDetection.ts
```diff
@@ -1,27 +1,49 @@
...
+	};
+
+	onMounted(() => {
+		// Attach listeners if user is writer
+		if (collaborationStore.isCurrentUserWriter) {
+			attachListeners();
+		}
+
+		// Watch for writer status changes
+		watch(
```
scdekov: Shouldn't this be outside of the onMounted?

scdekov (approved):

PR #23238, opened 2025-12-15
feat(core): Introduce WorkflowHistory compaction service

CharlieKolb:
## Summary

Introduce a compaction service to remove older (auto-saved) versions of workflows.

Currently only ever removes versions with redundant information, e.g. if a workflow history goes from changing a parameter of the same node from `'the quick'` to `'the quick brown'` to `'the quick brown fox'`, then we can remove the first two versions without losing any data. If a version was ever published or named we will never delete it currently.

Also includes https://github.com/n8n-io/n8n/pull/23282

https://linear.app/n8n/issue/ADO-4327

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/repositories/workflow-history.repository.ts
+++ b/packages/@n8n/db/src/repositories/workflow-history.repository.ts
@@ -69,7 +69,7 @@ export class WorkflowHistoryRepository extends Repository<WorkflowHistory> {
 				startDate,
 			})
 			.groupBy('wh.workflowId')
-			.getRawMany<WorkflowHistory>(); // need raw due to `distinct(true)`
+			.getRawMany<{ workflowId: string }>();
 
 		return result.map((x) => x.workflowId);
 	}
--- a/packages/@n8n/db/src/repositories/workflow-history.repository.ts
+++ b/packages/@n8n/db/src/repositories/workflow-history.repository.ts
@@ -1,11 +1,16 @@
 import { Service } from '@n8n/di';
-import { DataSource, LessThan, Repository } from '@n8n/typeorm';
+import { DataSource, In, LessThan, Repository } from '@n8n/typeorm';
+import { GroupedWorkflowHistory, groupWorkflows, RULES } from 'n8n-workflow';
 
 import { WorkflowHistory, WorkflowEntity } from '../entities';
+import { WorkflowPublishHistoryRepository } from './workflow-publish-history.repository';
 
 @Service()
 export class WorkflowHistoryRepository extends Repository<WorkflowHistory> {
-	constructor(dataSource: DataSource) {
+	constructor(
+		dataSource: DataSource,
+		private readonly workflowPublishHistoryRepository: WorkflowPublishHistoryRepository,
+	) {
 		super(WorkflowHistory, dataSource.manager);
 	}
 
@@ -40,5 +45,71 @@ export class WorkflowHistoryRepository extends Repository<WorkflowHistory> {
 			.andWhere(`versionId NOT IN (${currentVersionIdsSubquery})`)
 			.andWhere(`versionId NOT IN (${activeVersionIdsSubquery})`)
 			.execute();
+	}
+
+	private makeSkipActiveAndNamedVersionsRule(activeVersions: string[]) {
+		return (
+			prev: GroupedWorkflowHistory<WorkflowHistory>,
+			_next: GroupedWorkflowHistory<WorkflowHistory>,
+		): boolean =>
+			prev.to.name !== null ||
+			prev.to.description !== null ||
+			activeVersions.includes(prev.to.versionId);
+	}
+
+	async getWorkflowIdsInRange(startDate: Date, endDate: Date) {
+		const result = await this.manager
+			.createQueryBuilder(WorkflowHistory, 'wh')
+			.select('wh.workflowId', 'workflowId')
+			.distinct(true)
+			.where('wh.createdAt <= :endDate', {
+				endDate,
+			})
+			.andWhere('wh.createdAt >= :startDate', {
+				startDate,
+			})
+			.groupBy('wh.workflowId')
+			.getRawMany<WorkflowHistory>(); // need raw due to `distinct(true)`
+
+		return result.map((x) => x.workflowId);
+	}
+
+	/**
+	 * @returns The amount of seen and deleted versions
+	 */
+	async pruneHistory(
+		workflowId: string,
+		startDate: Date,
+		endDate: Date,
+	): Promise<{ seen: number; deleted: number }> {
+		const workflows = await this.manager
+			.createQueryBuilder(WorkflowHistory, 'wh')
+			.where('wh.workflowId = :workflowId', { workflowId })
+			.andWhere('wh.createdAt <= :endDate', {
+				endDate,
+			})
+			.andWhere('wh.createdAt >= :startDate', {
+				startDate,
+			})
+			.orderBy('wh.createdAt', 'ASC')
+			.getMany();
+
+		// Group by workflowId
+
+		const versionsToDelete = [];
+		const publishedVersions =
+			await this.workflowPublishHistoryRepository.getPublishedVersions(workflowId);
+		const grouped = groupWorkflows<WorkflowHistory>(
+			workflows,
+			[RULES.mergeAdditiveChanges],
+			[this.makeSkipActiveAndNamedVersionsRule(publishedVersions.map((x) => x.versionId))],
+		);
+		for (const group of grouped) {
+			for (const wf of group.groupedWorkflows) {
+				versionsToDelete.push(wf.versionId);
+			}
+		}
+		await this.delete({ versionId: In(versionsToDelete) });
+		return { seen: workflows.length, deleted: versionsToDelete.length };
 	}
 }
--- a/packages/cli/src/services/pruning/workflow-history-compaction.service.ts
+++ b/packages/cli/src/services/pruning/workflow-history-compaction.service.ts
@@ -1,0 +1,176 @@
+import { Logger } from '@n8n/backend-common';
+import { WorkflowHistoryCompactionConfig } from '@n8n/config';
+import { Time } from '@n8n/constants';
+import { DbConnection, WorkflowHistoryRepository } from '@n8n/db';
+import { OnLeaderStepdown, OnLeaderTakeover, OnShutdown } from '@n8n/decorators';
+import { Service } from '@n8n/di';
+import { InstanceSettings } from 'n8n-core';
+import { ensureError, sleep } from 'n8n-workflow';
+import { strict } from 'node:assert';
+
+/**
+ * Responsible for compacting auto saved workflow history entries in the database.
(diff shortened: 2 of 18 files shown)


packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,132 @@
...
+	 * Delete all soft-deleted executions and their binary data.
+	 */
+	private async compactHistories(): Promise<void> {
+		const startDate = new Date();
+		startDate.setHours(
+			startDate.getHours() - this.minimumCompactAgeHours - this.compactingTimeRangeHours,
+		);
+
+		const endDate = new Date();
+		endDate.setHours(endDate.getHours() - this.minimumCompactAgeHours);
```
dariacodes:
```suggestion
    const now = Date.now();
    
    const startDate = new Date(
        now - (this.minimumCompactAgeHours + this.compactingTimeRangeHours) * Time.hours.toMilliseconds,
    );

    const endDate = new Date(
        now - this.minimumCompactAgeHours * Time.hours.toMilliseconds,
    );
```

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,132 @@
...
+			startDate,
+			endDate,
+		);
+
+		this.logger.debug(
+			`Found ${workflowIds.length} workflows with versions between ${startDate.toISOString()} and ${endDate.toISOString()}`,
+		);
+
+		let seenSum = 0;
+		for (let i = 0; i < workflowIds.length; ++i) {
```
dariacodes: [nit] no strong opinion
It might make the code a bit more compact if we do this:
- `for (const [index, workflowId] of workflowIds.entries())`
- precompute this:
```js
const startIso = startDate.toISOString();
const endIso = endDate.toISOString();
```

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,132 @@
...
+			}
+
+			if (seenSum > this.batchSize) {
+				this.logger.warn(
+					`Encountered more than ${this.batchSize} workflow versions, waiting ${this.batchDelayMs * Time.milliseconds.toSeconds} second(s) before continuing.`,
+				);
+				this.logger.warn(
+					`Compacted ${i} of ${workflowIds.length} workflows with versions between ${startDate.toISOString()} and ${endDate.toISOString()}`,
+				);
+				await new Promise((res) => setTimeout(res, this.batchDelayMs));
```
dariacodes: we have `sleep` in utils

dariacodes (commented):
Nice work 🚀 I just left a few minor suggestions, I'll give it a go a bit later!

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,134 @@
...
+ */
+@Service()
+export class WorkflowHistoryCompactionService {
+	private compactingInterval: NodeJS.Timeout | undefined;
+
+	private readonly minimumCompactAgeHours = 24;
+	private readonly compactingTimeRangeHours = 1;
+
+	private readonly batchSize = 1_000;
+	private readonly batchDelayMs = 1_000;
```
RicardoE105: We should probably make this configurable
```

  workflowHistory: {
    compaction: {
      enabled: boolean;
      minimumAgeHours: number;
      batchSize: number;
    }
  }
```

packages/@n8n/db/src/repositories/workflow-history.repository.ts
```diff
@@ -41,4 +46,70 @@ export class WorkflowHistoryRepository extends Repository<WorkflowHistory> {
...
+	/**
+	 * @returns The amount of seen and deleted versions
+	 */
+	async pruneHistory(
+		workflowId: string,
+		startDate: Date,
+		endDate: Date,
+	): Promise<{ seen: number; deleted: number }> {
+		const workflows = await this.manager
+			.createQueryBuilder(WorkflowHistory, 'wh')
```
RicardoE105, line 86: We can add some indexes to the table to make this query faster

  CREATE INDEX idx_workflow_history_workflow_created
  ON workflow_history(workflowId, createdAt);

  CREATE INDEX idx_workflow_publish_history_workflow_version
  ON workflow_publish_history(workflowId, versionId);

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,134 @@
...
+					workflowId,
+					startDate,
+					endDate,
+				);
+				seenSum += seen;
+
+				this.logger.debug(
+					`Deleted ${deleted} of ${seen} versions of workflow ${workflowId} between ${startIso} and ${endIso}`,
+				);
+			} catch (error) {
```
RicardoE105, line 145: if we are having issues with the pruneHistory method (it throws) seenSum is not updated, and will keep failing forever as it will never reach the batchSize. At that point we are gonna overwhelm the database and crash the instance? Maybe something like below? or too cautious? 

```
  let consecutiveErrors = 0;
  const MAX_CONSECUTIVE_ERRORS = 5;

  for (const [index, workflowId] of workflowIds.entries()) {
      try {
          const { seen, deleted } = await this.workflowHistoryRepository.pruneHistory(...);
          seenSum += seen;
          consecutiveErrors = 0; // Reset on success
      } catch (error) {
          this.logger.error(...);
          consecutiveErrors++;

          if (consecutiveErrors >= MAX_CONSECUTIVE_ERRORS) {
              this.logger.error(`Too many consecutive errors (${consecutiveErrors}), stopping compaction`);
              break; // Stop processing
          }

          // Sleep after error to back off
          await sleep(this.batchDelayMs);
      }

      if (seenSum > this.batchSize) {
          await sleep(this.batchDelayMs);
          seenSum = 0;
      }
  }
```

packages/@n8n/db/src/repositories/workflow-history.repository.ts
```diff
@@ -41,4 +46,70 @@ export class WorkflowHistoryRepository extends Repository<WorkflowHistory> {
...
+		return result.map((x) => x.workflowId);
+	}
+
+	/**
+	 * @returns The amount of seen and deleted versions
+	 */
+	async pruneHistory(
+		workflowId: string,
+		startDate: Date,
+		endDate: Date,
```
RicardoE105, line 83: Let's validate this range is correct. Same for `getWorkflowIdsInRange`

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	@OnShutdown()
+	shutdown(): void {
+		this.isShuttingDown = true;
+		this.stopCompacting();
+	}
+
+	/**
+	 * Delete all soft-deleted executions and their binary data.
```
RicardoE105: wrong description

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
```diff
@@ -0,0 +1,134 @@
...
+	 */
+	private async compactHistories(): Promise<void> {
```
RicardoE105, line 98: nitpick: improve logging for debugging

```
private async compactHistories(): Promise<void> {
		const compactionStartTime = Date.now();
		const now = Date.now();

		const startDate = new Date(
			now -
				(this.minimumCompactAgeHours + this.compactingTimeRangeHours) * Time.hours.toMilliseconds,
		);
		const endDate = new Date(now - this.minimumCompactAgeHours * Time.hours.toMilliseconds);

		const startIso = startDate.toISOString();
		const endIso = endDate.toISOString();

		this.logger.info('Starting workflow history compaction', {
			dateRange: { start: startIso, end: endIso },
			minimumAgeHours: this.minimumCompactAgeHours,
			rangeHours: this.compactingTimeRangeHours,
		});

		const workflowIds = await this.workflowHistoryRepository.getWorkflowIdsInRange(
			startDate,
			endDate,
		);

		this.logger.debug(
			`Found ${workflowIds.length} workflows with versions between ${startDate.toISOString()} and ${endDate.toISOString()}`,
		);

		let seenSum = 0;
		let totalDeleted = 0;
		let errorCount = 0;

		for (const [index, workflowId] of workflowIds.entries()) {
			try {
				const { seen, deleted } = await this.workflowHistoryRepository.pruneHistory(
					workflowId,
					startDate,
					endDate,
				);
				seenSum += seen;
				totalDeleted += deleted;

				this.logger.debug(
					`Deleted ${deleted} of ${seen} versions of workflow ${workflowId} between ${startIso} and ${endIso}`,
				);
			} catch (error) {
				errorCount++;
				this.logger.error(`Failed to prune version history of workflow ${workflowId}`, {
					error: ensureError(error),
				});
			}

			if (seenSum > this.batchSize) {
				this.logger.warn(
					`Encountered more than ${this.batchSize} workflow versions, waiting ${this.batchDelayMs * Time.milliseconds.toSeconds} second(s) before continuing.`,
				);
				this.logger.warn(
					`Compacted ${index} of ${workflowIds.length} workflows with versions between ${startIso} and ${endIso}`,
				);
				await sleep(this.batchDelayMs);
				seenSum = 0;
			}
		}

		const compactionDuration = Date.now() - compactionStartTime;

		this.logger.info('Workflow history compaction complete', {
			workflowsProcessed: workflowIds.length,
			totalVersionsSeen: seenSum,
			totalVersionsDeleted: totalDeleted,
			errorCount,
			durationMs: compactionDuration,
			durationSeconds: Math.round(compactionDuration / 1000),
			dateRange: { start: startIso, end: endIso },
		});
	}

```

RicardoE105 (changes requested):
Nice work 🚀 Will do some manual testing now

packages/@n8n/db/src/repositories/workflow-history.repository.ts
CharlieKolb: We have the index on the publish history already - I planned ahead :D 

I can't really judge if the one on the workflow history itself is worth it? We have one on workflowId already, and this will be called once per workflow per hour at most.
Bear in mind this index would make writing, i.e. the auto-saves themselves slower.

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
CharlieKolb: True - we deemed this out of scope for the initial ticket, do you think we should do this now?

packages/cli/src/services/pruning/workflow-history-compaction.service.ts
CharlieKolb: Note that we don't loop based on the sum, we just take a break if we hit a certain count. I can see an argument to stop trying tho! Will work this into the solution, thanks!

RicardoE105 (approved):
🚀

PR #23251, opened 2025-12-15
feat(core): Add workflow activation events to log streaming

MarcL:
## Summary

Connect `workflow-activated` and `workflow-deactivated` events to the log streaming relay. These events were already emitted by WorkflowService but not streamed to external destinations.

**Changes:**
- Add `n8n.audit.workflow.activated` and `n8n.audit.workflow.deactivated` audit event names
- Add `activeVersionId` to audit event payload (enables tracking version restores)
- Add listeners and handlers in `LogStreamingEventRelay`
- Add unit tests for both events

https://linear.app/n8n/issue/PAY-2934

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/eventbus/event-message-classes/index.ts
```diff
@@ -81,6 +81,8 @@ export const eventNamesAudit = [
 	'n8n.audit.workflow.updated',
 	'n8n.audit.workflow.archived',
 	'n8n.audit.workflow.unarchived',
+	'n8n.audit.workflow.activated',
```
konstantintieber, line 84: Should we call these `published` and `unpublished` instead now?

guillaumejacquart (approved):
Looks good to me :)

packages/cli/src/eventbus/event-message-classes/index.ts
MarcL: @konstantintieber - I did wonder this too but the `workflow-activated` and `workflow-deactivated` events already existed. If we change it, it would alter any existing log ingestion that users have so maybe it alters their expectations. What do you think?

konstantintieber: Ah I see, so there's these "audit" events and "regular" events and for the regular ones we still use the old naming.

Then I think it's better to stick with the the same naming in the audit ones as you did here 👍

MarcL: Had a look through the code and we'd get some inconsistencies if we change it:

* telemetry would still be activated
* event messages currently map to their internal names
* DB references active not published

Should we leave as is?

konstantintieber (approved):

PR #23264, opened 2025-12-16
fix: Only support specified git config keys in Git node

dlavrenuek:
## Summary

Only support specified configuration keys in `Git` node

https://linear.app/n8n/issue/NODE-4089

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/Git/Git.node.ts
+++ b/packages/nodes-base/nodes/Git/Git.node.ts
@@ -35,7 +35,7 @@ export class Git implements INodeType {
 		name: 'git',
 		icon: 'file:git.svg',
 		group: ['transform'],
-		version: 1,
+		version: [1, 1.1],
 		description: 'Control git.',
 		defaults: {
 			name: 'Git',
@@ -357,9 +357,10 @@ export class Git implements INodeType {
 
 					const key = this.getNodeParameter('key', itemIndex, '') as string;
 					const value = this.getNodeParameter('value', itemIndex, '') as string;
+					const securityConfig = Container.get(SecurityConfig);
+					const enableGitNodeAllConfigKeys = securityConfig.enableGitNodeAllConfigKeys;
 					let append = false;
-
-					if (!ALLOWED_CONFIG_KEYS.includes(key)) {
+					if (!enableGitNodeAllConfigKeys && !ALLOWED_CONFIG_KEYS.includes(key)) {
 						throw new NodeOperationError(
 							this.getNode(),
 							`The provided git config key '${key}' is not allowed`,
--- a/packages/nodes-base/nodes/Git/Git.node.ts
+++ b/packages/nodes-base/nodes/Git/Git.node.ts
@@ -18,6 +18,7 @@ import { URL } from 'url';
 import {
 	addConfigFields,
 	addFields,
+	ALLOWED_CONFIG_KEYS,
 	cloneFields,
 	commitFields,
 	logFields,
@@ -357,6 +358,13 @@ export class Git implements INodeType {
 					const key = this.getNodeParameter('key', itemIndex, '') as string;
 					const value = this.getNodeParameter('value', itemIndex, '') as string;
 					let append = false;
+
+					if (!ALLOWED_CONFIG_KEYS.includes(key)) {
+						throw new NodeOperationError(
+							this.getNode(),
+							`The provided git config key '${key}' is not allowed`,
+						);
+					}
 
 					if (options.mode === 'append') {
 						append = true;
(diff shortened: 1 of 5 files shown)


packages/nodes-base/nodes/Git/descriptions/AddConfigDescription.ts
```diff
@@ -1,17 +1,22 @@
 import type { INodeProperties } from 'n8n-workflow';
 
+export const ALLOWED_CONFIG_KEYS = ['user.email', 'user.name', 'remote.origin.url'];
+
 export const addConfigFields: INodeProperties[] = [
 	{
 		displayName: 'Key',
 		name: 'key',
-		type: 'string',
+		type: 'options',
```
DawidMyslak, line 9: I think this is a breaking change and requires versioning.

What if an existing node has a key which is not listed in `ALLOWED_CONFIG_KEYS`?

It will most likely break and make some users unhappy. Unless that's what we want?

DawidMyslak (changes requested):
Looks good, just need to clarify something.

packages/nodes-base/nodes/Git/descriptions/AddConfigDescription.ts
dlavrenuek: Added version handling for 1 and 1.1 in the form to ensure backwards compatibility

dlavrenuek: The check can be bypassed on self-hosted environments by setting env `N8N_GIT_NODE_ENABLE_ALL_CONFIG_KEYS=true`. Otherwise we only want to allow the specified keys per default.

packages/nodes-base/nodes/Git/Git.node.ts
```diff
@@ -34,7 +35,7 @@ export class Git implements INodeType {
 		name: 'git',
 		icon: 'file:git.svg',
 		group: ['transform'],
-		version: 1,
+		version: [1, 1.1],
```
DawidMyslak, line 38: We usually also set `defaultVersion` version param. I think it's not mandatory, but setting it to `1.1` is probably a good practice.

packages/nodes-base/nodes/Git/Git.node.ts
```diff
@@ -356,7 +357,15 @@ export class Git implements INodeType {
 
 					const key = this.getNodeParameter('key', itemIndex, '') as string;
 					const value = this.getNodeParameter('value', itemIndex, '') as string;
+					const securityConfig = Container.get(SecurityConfig);
+					const enableGitNodeAllConfigKeys = securityConfig.enableGitNodeAllConfigKeys;
 					let append = false;
+					if (!enableGitNodeAllConfigKeys && !ALLOWED_CONFIG_KEYS.includes(key)) {
```
DawidMyslak, line 363: I think this part also should check the version. We want to throw only for `version >= 1.1 `.

DawidMyslak (changes requested):

packages/nodes-base/nodes/Git/Git.node.ts
dlavrenuek: I did only find it on y few nodes that have multiple versions configured, f.e. [`Code` node](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Code/Code.node.ts#L46). On most nodes it is actually unset, f.e. [`Form` node](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Form/Form.node.ts#L272).

I don't see how it would be beneficial since the latest version is used per default. If the version is set explicitly and we add a new version, there is a chance that someone forgets to update default version. So I would rather not add the default version here

packages/nodes-base/nodes/Git/Git.node.ts
dlavrenuek: This check is actually correct. The provided git config key must be always checked and only ignored if `enableGitNodeAllConfigKeys` is configured. The node version check is only for the UI because otherwise existing workflows would break in the UI due to having a value that does not exist in the select list - for this case we keep the plain input field

DawidMyslak (approved):
LGTM 👍

packages/nodes-base/nodes/Git/descriptions/AddConfigDescription.ts
tomi: I think it should be more flexible instead to allow users to allowlist which config options they want to support instead of enabling of all them. That would be more secure as well

IA-domcom:
Gracias 😊

PR #23270, opened 2025-12-16
feat: Binary data merge and expressions simplification

michael-radency:
## Summary

This PR introduces a new **combined binary data mode** for n8n workflows that changes how binary data is accessed and where it is included in items. Instead of storing binary data separately from the item's JSON data under a `binary` key, the combined mode automatically includes binary entries directly within the item's JSON data under a `_files` property, simplifying expressions for data access. Binary entries can be included at any level of nesting in the item's JSON data and can also be placed inside arrays. Items will be displayed in the UI in Schema and Table views with options to preview and download binary items.

image
image
image
image

### 1. New Binary Data Mode

**New Binary Mode Setting**
- Added `BINARY_IN_JSON_PROPERTY = '_files'` constant
- New workflow setting: `binaryMode: 'separate' | 'combined'`
- When `binaryMode: 'combined'`, binary data is automatically converted to be accessible via `$json._files` instead of `$binary`

### 2. UI/UX Improvements

**Binary Data Viewer Components**
- `BinaryDataViewModal.vue`: Modal for previewing binary files (images, videos, PDFs, JSON, HTML, markdown, text)
- `BinaryEntryDataTable.vue`: Table row component for displaying binary file metadata with download and preview actions
- `VirtualSchemaItem.vue`: Enhanced schema view to recognize and display binary data embedded in JSON

**Workflow Settings**
- New execution logic selector with 3 modes:
  - **v0 (legacy)**: Original execution order
  - **v1 (recommended)**: Sequential branch execution
  - **v2 (experimental)**: Sequential with combined binary mode enabled
- Selecting v2 sets `binaryMode: 'combined'`, which enables simplified expressions and inclusion of binaries into JSON

### 3. Expression Simplification

**Workflow Data Proxy** (`packages/workflow/src/workflow-data-proxy.ts`)
- `$item` now behaves like `$json` and `$data` in combined mode, returning the JSON portion of items directly
- `$input.item`, `$input.first()`, `$input.last()`, and `$input.all()` return JSON data directly in combined mode
- `$('nodeName').first()`, `$('nodeName').last()`, `$('nodeName').all()`, and `$('nodeName').pairedItem()` return JSON data in combined mode
- New optional `fullItem` parameter for `$('nodeName', fullItem)` to force returning full item structure even in combined mode
- Maintains backward compatibility: separate mode returns full item structures as before

**Code Node Autocomplete**
- New type declarations for combined mode: `n8n-once-for-all-items-combined.d.ts`, `n8n-once-for-each-item-combined.d.ts`
- Autocomplete now omits the `.json` suggestion for item data access in combined mode
- Suppresses `await` warnings, as it was falsely warning that `await` could not be used inside Code nodes

### 4. Backend Changes

**Binary Data Conversion**
- Updated execution logic to process binaries based on the workflow's `binaryMode` setting
- Added helper function `convertBinaryData` that moves binaries to the JSON `_files` property, automatically converting base64 data to entries based on filesystem mode

**Helper Functions**
- Enhanced `getBinaryDataBuffer()` to work with both modes
- `assertBinaryData()` now support binary data access by path from both `binary` and `json` properties

https://linear.app/n8n/issue/NODE-3854/add-binarydata-component-for-binaries-included-in-json-schema-view

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/design-system/src/components/N8nTree/Tree.vue
```diff
@@ -59,7 +108,10 @@ const N8nTree = getCurrentInstance()?.type;
 
 <template>
 	<div v-if="isObject(value)" class="n8n-tree">
-		<div v-for="(label, i) in Object.keys(value)" :key="i" :class="classes">
+		<div v-if="isBinary(value)">
+			<slot name="binary" v-bind:value="value" v-bind:path="path" />
+		</div>
+		<div v-else v-for="(label, i) in Object.keys(value)" :key="i" :class="classes">
```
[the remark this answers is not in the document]
michael-radency: will not be reordered

packages/frontend/editor-ui/src/features/ndv/runData/components/RunDataBinary.vue
```diff
@@ -35,10 +43,18 @@ async function downloadBinaryData(index: number, key: string | number) {
...
 
 <template>
 	<div :class="$style.component">
+		<N8nNotice v-if="isLegacyBinaryMode" :class="$style.info" theme="info">
+			Try the new "combined" binary mode by selecting the corresponding option in the workflow
+			<N8nLink size="small" @click="openWorkflowSettings">settings</N8nLink>.
+		</N8nNotice>
 		<N8nText v-if="binaryData.length === 0" align="center" tag="div">
 			{{ i18n.baseText('runData.noBinaryDataFound') }}
 		</N8nText>
```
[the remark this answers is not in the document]
michael-radency: will not be reordered

packages/frontend/editor-ui/src/features/shared/editors/plugins/codemirror/typescript/worker/typescript.worker.ts
```diff
@@ -141,16 +144,21 @@ export const worker: LanguageServiceWorkerInit = {
...
+						? runOnceForAllItemsCombinedTypes
+						: runOnceForAllItemsTypes
+					: isCombined
+						? runOnceForEachItemCombinedTypes
+						: runOnceForEachItemTypes;
+
+			updateFile(TYPESCRIPT_FILES.MODE_TYPES, modeTypes);
+		};
+
+		watch([mode, binaryMode], updateModeTypes, { immediate: true });
```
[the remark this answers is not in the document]
michael-radency: hmm, not sure if this is true, check again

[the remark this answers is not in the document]
michael-radency: but changing binaryMode is not possible with active NDV, so worker has to be restarted/reinitialized, mode is node parameter so it does make sense to watch it

packages/frontend/editor-ui/src/features/ndv/shared/ndv.store.ts
```diff
@@ -216,6 +216,14 @@ export const useNDVStore = defineStore(STORES.NDV, () => {
 
 	const isNDVOpen = computed(() => activeNodeName.value !== null);
 
+	const binaryDataAccessTooltip = computed(() => {
```
ShireenMissi: this feels wrong here, maybe define a flag to switch between the modes here and add the text on the component using it
also the text should be in locales

packages/frontend/editor-ui/src/features/shared/editors/plugins/codemirror/typescript/worker/type-declarations/n8n.d.ts
```diff
@@ -79,6 +79,58 @@ declare global {
 	const $runIndex: number;
 	const $now: DateTime;
 	const $today: DateTime;
+	type IHttpRequestMethods = 'DELETE' | 'GET' | 'HEAD' | 'PATCH' | 'POST' | 'PUT';
```
ShireenMissi, line 97: this change seems unrelated to json/binary merge, could you explain why it's added? and update the PR description or even extract it to a separate PR

packages/nodes-base/nodes/Form/utils/utils.ts
```diff
@@ -430,18 +437,28 @@ export async function prepareFormReturnItem(
...
-			}
-
-			returnItem.binary![binaryPropertyName] = await context.nodeHelpers.copyBinaryFile(
+			const binaryData = await context.nodeHelpers.copyBinaryFile(
 				file.filepath,
 				file.originalFilename ?? file.newFilename,
 				file.mimetype,
 			);
 
+			if (binaryMode === 'combined') {
```
ShireenMissi: could we hide the details of binary mode handling on the node level? we will still need to make changes on the node level but the actual changes can happen in one place

packages/frontend/editor-ui/src/app/composables/useRunWorkflow.ts
```diff
@@ -152,6 +152,19 @@ export function useRunWorkflow(useRunWorkflowOpts: {
...
 			const workflowData = await workflowHelpers.getWorkflowDataToSave();
 
+			if (
+				rootStore.binaryDataMode === 'default' &&
+				workflowData.settings?.binaryMode === 'combined'
+			) {
+				toast.showMessage({
+					title: 'Unsupported Binary Mode',
+					message:
+						'Binary mode "combined" is not supported when filesystem mode is "default". Please change the binary mode to "separate" in workflow settings and update expressions that are referencing binary data accordingly.',
```
ShireenMissi: since default is deprecated in v2, do we still need this?

ShireenMissi (commented):
I haven’t fully tested or reviewed this yet, but from some manual testing I can see how useful this is 👏. 
I’ve left a few comments for now and will continue reviewing the PR.

packages/frontend/editor-ui/src/features/shared/editors/plugins/codemirror/typescript/worker/type-declarations/n8n.d.ts
michael-radency: it's for autocomplete of helpers exposed in code node

packages/frontend/editor-ui/src/app/composables/useRunWorkflow.ts
michael-radency: maybe not, I'll check

packages/nodes-base/nodes/Form/utils/utils.ts
michael-radency: probably not, we do have general binary handling in `convertBinary` helper, but logic here is specific to the node

packages/frontend/editor-ui/src/app/composables/useRunWorkflow.ts
michael-radency: we do need this, 'default' mode is not removed, I could still set env and use it
image

packages/core/src/execution-engine/node-execution-context/utils/__tests__/binary-helper-functions.test.ts
```diff
@@ -329,6 +330,346 @@ describe('test binary data helper methods', () => {
...
+					],
+				],
+			};
+
+			await expect(
+				getBinaryDataBuffer(inputData, 0, 'file', 0, BINARY_MODE_COMBINED),
+			).rejects.toThrow('Provided parameter is not a string or binary data object.');
+		});
+
+		it('should throw error when path is undefined in combined mode', async () => {
```
dlavrenuek: The test name is a bit misleading because `path` is `'data.file'` and also defined, but the value is undefined. May be something like?


```suggestion
		it('should throw error when path resolves to undefined in combined mode', async () => {
```

packages/core/src/execution-engine/node-execution-context/utils/__tests__/binary-helper-functions.test.ts
```diff
@@ -450,6 +791,106 @@ describe('detectBinaryEncoding', () => {
...
 		expect(detectBinaryEncoding(emptyBuffer)).toBeDefined();
 	});
+
+	it('should detect ASCII encoding', () => {
+		const asciiBuffer = Buffer.from('Simple ASCII text 123');
+		const encoding = detectBinaryEncoding(asciiBuffer);
+		expect(encoding).toBeDefined();
+		// ASCII is often detected as UTF-8, ASCII, or similar encodings
+		expect(['UTF-8', 'ASCII', 'ascii', 'windows-1252', 'ISO-8859-1']).toContain(encoding);
+	});
```
dlavrenuek, line 800: Optional suggestion: The test file is saved in utf8, so all text inside is also utf8. Ascii can be enforced the same way it is done in other tests with different encodings but there is also an easy way to just use Buffer.from().toString('ascii'). The used `chardet` library uses only [uppercase for the charset](https://github.com/runk/node-chardet/blob/master/src/encoding/ascii.ts#L6). Also the readme does not state that it supports ascii, it obviously does support it as [`ASCII`](https://github.com/runk/node-chardet/blob/master/src/encoding/ascii.ts#L6). Does the following also work on your machine?

```suggestion
	it('should detect ASCII encoding', () => {
		const asciiBuffer = Buffer.from(Buffer.from('Simple ASCII text 123').toString('ascii'));
		const encoding = detectBinaryEncoding(asciiBuffer);
		expect(encoding).toBeDefined();
		expect(encoding).toBe("ASCII");
	});
```

packages/core/src/execution-engine/node-execution-context/utils/__tests__/binary-helper-functions.test.ts
```diff
@@ -329,6 +330,346 @@ describe('test binary data helper methods', () => {
 			).rejects.toThrow('Provided parameter is not a string or binary data object.');
 		});
 	});
+
+	describe('getBinaryDataBuffer with combined binary mode', () => {
```
dlavrenuek, line 334: Nice test coverage 🏅

packages/core/src/execution-engine/node-execution-context/utils/binary-helper-functions.ts
```diff
@@ -88,32 +86,49 @@ export function assertBinaryData(
...
-			node,
-			`The item has no binary field '${parameterData}' [item ${itemIndex}]`,
-			{
-				itemIndex,
-				description:
-					'Check that the parameter where you specified the input binary field name is correct, and that it matches a field in the binary input',
-			},
-		);
-	}
+	if (binaryMode === 'separate' || binaryMode === undefined) {
```
dlavrenuek: With this PR we will have two modes: "separate" and "combined". This condition suggests that "separate" is the default behavior if unset or specificly defined as "separate", otherwise it will be handled as "conbined". This is confusing in my opinion especially if the available modes are extended in the future.

The bahaviour should be more clear when the condition is reversed as 

```
if (binaryMode === BINARY_MODE_COMBINED) {..} else {...}
```

or if used with switch instead of if:

```
switch (binaryMode) {
	case BINARY_MODE_COMBINED: { ... }
	case "separate":
    default: { ... }
}
```

What do you think?

packages/core/src/execution-engine/node-execution-context/utils/binary-helper-functions.ts
```diff
@@ -88,32 +86,49 @@ export function assertBinaryData(
...
+				},
+			);
+		}
 
-	return binaryPropertyData;
+		return binaryPropertyData;
+	} else {
+		const itemData = inputData.main[inputIndex]![itemIndex].json;
+		const binaryData = get(itemData, parameterData);
+		if (binaryData === undefined || !isBinaryValue(binaryData)) {
```
dlavrenuek: `isBinaryValue` already checks internally if the passed value is an object, so we can safely remove the check if it's undefined

packages/core/src/execution-engine/node-execution-context/utils/binary-helper-functions.ts
```diff
@@ -126,13 +141,21 @@ export async function getBinaryDataBuffer(
...
-	if (isBinaryData(parameterData)) {
+	if (isBinaryValue(parameterData)) {
 		binaryData = parameterData;
-	} else if (typeof parameterData === 'string') {
+	} else if (typeof parameterData === 'string' && binaryMode !== BINARY_MODE_COMBINED) {
 		binaryData = inputData.main[inputIndex]![itemIndex].binary![parameterData];
+	} else if (typeof parameterData === 'string') {
+		const itemData = inputData.main[inputIndex]![itemIndex].json;
+		const data = get(itemData, parameterData);
+		if (data === undefined || !isBinaryValue(data)) {
```
dlavrenuek: also here the check if data is undefined can be omitted since `isBinaryValue` returns false if it is undefined

packages/core/src/execution-engine/node-execution-context/utils/binary-helper-functions.ts
```diff
@@ -73,8 +70,9 @@ export function assertBinaryData(
 	itemIndex: number,
 	parameterData: string | IBinaryData,
 	inputIndex: number,
+	binaryMode?: WorkflowSettingsBinaryMode,
```
dlavrenuek: Without having a lot of context about the binary merge feature it is difficult to understand what the default behavior is without looking through the whole code. The same behavior but with more clarity (and type safety) can be achieved by not marking this parameter optional but instead use a default value


```suggestion
	binaryMode: WorkflowSettingsBinaryMode = 'separate',
```

packages/core/src/execution-engine/node-execution-context/utils/binary-helper-functions.ts
```diff
@@ -126,13 +141,21 @@ export async function getBinaryDataBuffer(
...
```
dlavrenuek: Also here a default instead of optional param declaration would bring more clarity and type safety


```suggestion
	binaryMode: WorkflowSettingsBinaryMode = 'separate',
```

packages/core/src/execution-engine/node-execution-context/node-execution-context.ts
```diff
@@ -163,6 +163,10 @@ export abstract class NodeExecutionContext implements Omit<FunctionsBase, 'getCr
 		return null;
 	}
 
+	getWorkflowSettings() {
+		return this.workflow.settings;
```
dlavrenuek: Exposing the workflow settings to nodes is dangerous since it is an object and objects are passed by reference. Imagine a code inside a node like

```js
const settings = this.getWorkflowSettings();
settings.binaryMode = "combined";

doSomething(settings);
```

This will change the `binaryMode` globally in the workflow object while it stays in the memory and will affect all nodes in the execution chain. Normally you could simple return `structuredClone(this.workflow.settings)`, but as this code will be executed on every node execution (if I see it correctly) the additional computation for deep cloning the settings should not be neglected. Looking at what is already used in this file - there is the `@Memoized` - this could be used to limit running `structuredClone` only once. But then the memoized object is still reused across the execution of different nodes, so the only solution I see is to make it read only with `Object.freeze`. What do you think?

```ts
	@Memoized
	get workflowSettings() {
		return Object.freeze(structuredClone(this.workflow.settings));
	}

	getWorkflowSettings() {
		return this.workflowSettings;
	}
```

Also adding a test to ensure that modifying workflow setting throws an error would also be good

packages/frontend/@n8n/design-system/src/components/N8nTree/Tree.vue
```diff
@@ -59,7 +62,10 @@ const N8nTree = getCurrentInstance()?.type;
 
 <template>
 	<div v-if="isObject(value)" class="n8n-tree">
-		<div v-for="(label, i) in Object.keys(value)" :key="i" :class="classes">
+		<div v-if="isBinary(value)">
+			<slot name="binary" v-bind:value="value" v-bind:path="path" />
```
dlavrenuek: `v-bind:xxx` should be [shorthanded](https://vuejs.org/guide/essentials/template-syntax#shorthand) to `:xxx`


```suggestion
			<slot name="binary" :value="value" :path="path" />
```

packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
```diff
@@ -493,6 +536,26 @@ const saveSettings = async () => {
...
+			message: h('span', [
+				i18n.baseText('workflowSettings.executionLogic.changed.description'),
+				h(
+					N8nLink,
+					{
+						to: 'https://docs.n8n.io/data/binary-data/',
+						size: 'small',
+						newWindow: true,
+					},
+					() => 'Learn more',
```
dlavrenuek: Great to see that translation is considered. This text was forgotten


```suggestion
					() => i18n.baseText('generic.learnMore'),
```

packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
```diff
@@ -512,6 +575,18 @@ const updateTimeSavedPerExecution = (value: string) => {
 			: numValue;
 };
 
+const onExecutionLogicModeChange = (value: string) => {
+	if (value === 'v0' || value === 'v1') {
+		workflowSettings.value.binaryMode = 'separate';
```
dlavrenuek: The keyword `separate` is used across backend and frontend while combined mode uses `BINARY_MODE_COMBINED` constant. It might be worth defining a constant for `separate` - not that it's needed but more for consistency

packages/frontend/editor-ui/src/features/ndv/runData/components/BinaryEntryDataTable.vue
```diff
@@ -0,0 +1,205 @@
...
+const downloadBinaryData = async () => {
+	try {
+		const response = await fetch(fileUrl.value);
+		if (!response.ok) throw new Error('Error downloading file');
+		const blob = await response.blob();
+		saveAs(blob, fileName.value);
+	} catch (error) {
+		useToast().showMessage({
+			title: 'Error downloading file',
+			message: 'File could not be downloaded',
```
dlavrenuek: We should add these text to localization

packages/frontend/editor-ui/src/features/ndv/runData/components/BinaryEntryDataTable.vue
```diff
@@ -0,0 +1,205 @@
...
+};
+
+const containerStyle = computed(() => ({
+	marginLeft: props.depth ? `${props.depth * 15}px` : '0',
+}));
+
+const isDownloadHovered = ref(false);
+
+const downloadIconColor = computed(() => (isDownloadHovered.value ? 'primary' : 'text-base'));
+
```
dlavrenuek: This is a lot of complexity for a hover effect. We should simply use the CSS hover instead of using vue state+rendering logic.


```suggestion
```

packages/frontend/editor-ui/src/features/ndv/runData/components/BinaryEntryDataTable.vue
```diff
@@ -0,0 +1,205 @@
...
+			<div v-if="fileMeta" :class="$style.meta">{{ fileMeta }}</div>
+		</div>
+		<div
+			:class="$style.download"
+			@click="downloadBinaryData"
+			@mouseenter="isDownloadHovered = true"
+			@mouseleave="isDownloadHovered = false"
+		>
+			<N8nIcon icon="download" size="large" :color="downloadIconColor" />
+		</div>
```
dlavrenuek: color and hover color can be defined in the css class. Also I suggest using either a button or a link here to make it accessible


```suggestion
		<a :class="$style.download" @click="downloadBinaryData">
			<N8nIcon icon="download" size="large" />
		</a>
```

packages/frontend/editor-ui/src/features/ndv/runData/components/BinaryEntryDataTable.vue
```diff
@@ -0,0 +1,205 @@
...
+.download {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	width: 48px;
+	height: 48px;
+	cursor: pointer;
+	flex-shrink: 0;
+	opacity: 0;
+	transition: opacity 0.2s ease-in-out;
```
dlavrenuek: If this is a button or link, cursor is not needed. To keep the colors as you had them following will work:


```suggestion
	flex-shrink: 0;
	opacity: 0;
	transition: all 0.2s ease-in-out;
	color: var(--color-text-base);

	&:hover {
		color: var(--color--primary--shade-1);
	}
```

It would be even better to use a separate component "icon button" if such one exists

packages/frontend/editor-ui/src/features/ndv/runData/components/RunDataBinary.vue
```diff
@@ -1,16 +1,35 @@
 <script setup lang="ts">
```
dlavrenuek, line 1: TODO: continue review here

dlavrenuek (changes requested):
That's a great and huge piece of work @michael-radency 💪 ! Thank you for implementing the bg improvements!

Adding my review so far, will continue tomorrow.

packages/core/src/execution-engine/node-execution-context/utils/binary-helper-functions.ts
michael-radency: yes, make sense, updated, thanks :+1:

packages/core/src/execution-engine/node-execution-context/node-execution-context.ts
michael-radency: thanks @dlavrenuek , that's a great suggestion, updated :+1:

packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
```diff
@@ -594,10 +669,14 @@ onMounted(async () => {
 	if (workflowSettingsData.executionOrder === undefined) {
 		workflowSettingsData.executionOrder = 'v0';
 	}
+	if (workflowSettingsData.binaryMode === undefined) {
+		workflowSettingsData.binaryMode = BINARY_MODE_SEPARATE;
+	}
```
dlavrenuek, line 678: Just a note and there is no need to change this code since it aligns with the other code in this file: there is a [`??=`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) operator, which can reduce the block to

```js
workflowSettingsData.binaryMode ??= BINARY_MODE_SEPARATE;
```

It's not 100% same since `??=` checks for `undefined` or `null` but in most cases it should work the same if the property can not be null anyway

packages/frontend/editor-ui/src/features/ndv/runData/components/VirtualSchemaItem.vue
```diff
@@ -23,13 +31,53 @@ type Props = {
...
+
 const emit = defineEmits<{
 	click: [];
 }>();
+
+const isDownloadHovered = ref(false);
+const isViewHovered = ref(false);
+
+const downloadIconColor = computed(() => (isDownloadHovered.value ? 'primary' : 'text-base'));
+const viewIconColor = computed(() => (isViewHovered.value ? 'primary' : 'text-base'));
```
dlavrenuek: changing the color on hover should be done via plain css - using vue rendering logic with js mouseover/out events is a very heavy operation for such a task

packages/frontend/editor-ui/src/features/ndv/shared/ndv.store.ts
dlavrenuek: A good place for view related reusable logic is a hook, for example `useBinaryDataAccessTooltip`, which should be outside of the store.

packages/frontend/editor-ui/src/features/shared/editors/plugins/codemirror/typescript/worker/type-declarations/n8n.d.ts
```diff
@@ -79,6 +79,58 @@ declare global {
...
+
+	interface IHttpRequestOptions {
```
dlavrenuek, line 99: Where are these type declarations used?

packages/workflow/src/interfaces.ts
```diff
@@ -12,7 +12,14 @@ import type { Readable } from 'stream';
 import type { SecureContextOptions } from 'tls';
 import type { URLSearchParams } from 'url';
 
-import type { CODE_EXECUTION_MODES, CODE_LANGUAGES, LOG_LEVELS } from './constants';
+import type {
```
dlavrenuek, line 15: I'm wondering how this does not throw an error during compilation because all of these are constants and not types 🤔 . No need to change it tho, since it seems to work and could break without `import type`

dlavrenuek (changes requested):
Some comments and suggestions after a full review.

During testing have noticed that (based on the code I see) you can return `[{ json: {...}, binary: {...}]` from a node/code, but in the code node it shows ts error. With `// @ts-disable` it works. Is it an intended use case? If yes, then the type checks should also allow 

```js

// @ts-ignore
return [{json:{ok: "test", buffer: Buffer.from("test")}, binary:{ok:{id:123}, ko: $input.first()._files.data, buffer2: Buffer.from("test")}}];
```

packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
michael-radency: let's leave it for now as is, for consistency with surrounding code

packages/frontend/editor-ui/src/features/shared/editors/plugins/codemirror/typescript/worker/type-declarations/n8n.d.ts
michael-radency: we do expose some helpers in code node but completions was missing, those meant to add them

packages/workflow/src/interfaces.ts
michael-radency: I'm not sure either, just follow pattern here

packages/frontend/editor-ui/src/features/ndv/runData/components/BinaryEntryDataTable.vue
michael-radency: having pointer was design requirement

dlavrenuek: Thank you for making it a button! `<button>` and `<a>` (with href=) elements have a pointer per default, so there is no need to specify it per css unless it's removed via css in some other place.

PS: currently it has `transition: opacity 0.2s ease-in-out;`, so the icon color changes without transition. To also affect the icon color you can use `transition: all 0.2s ease-in-out;`

dlavrenuek (approved):
Thank you for the huge improvements 🥇 Also it's great to see how quickly you did address all of the review suggestions 💪

PR #23275, opened 2025-12-16
chore(core): Set foreign key constraint on ChatHub agentId

autologie:
## Summary

This PR follows up #23032 to introduce foreign key constraint on `agentId` column in the ChatHub session and message table.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1765886667897-AddAgentIdForeignKeys.ts
+++ b/packages/@n8n/db/src/migrations/common/1765886667897-AddAgentIdForeignKeys.ts
@@ -7,8 +7,6 @@ const table = {
 } as const;
 
 export class AddAgentIdForeignKeys1765886667897 implements ReversibleMigration {
-	transaction = false as const;
-
 	async up({ schemaBuilder: { addForeignKey }, runQuery, escape }: MigrationContext) {
 		// Clean up orphaned agentId references before adding foreign key constraint
 		await runQuery(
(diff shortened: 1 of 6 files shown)


packages/@n8n/db/src/migrations/common/1765886667897-AddAgentIdForeignKeys.ts
```diff
@@ -0,0 +1,50 @@
...
+			`UPDATE ${escape.tableName(table.messages)} SET "agentId" = NULL WHERE "agentId" IS NOT NULL AND "agentId" NOT IN (SELECT id FROM ${escape.tableName(table.agents)})`,
+		);
+
+		// Add foreign key constraint for agentId in sessions table
+		await addForeignKey(
+			table.sessions,
+			'agentId',
+			[table.agents, 'id'],
+			'FK_chat_hub_sessions_agentId',
+			'SET NULL',
```
guillaumejacquart, line 27: Why not cascade delete ? Is there a possibility that the agent is deleted and we should still keep the messages ?

guillaumejacquart (commented):
one question

packages/@n8n/db/src/migrations/common/1765886667897-AddAgentIdForeignKeys.ts
autologie: I think we want to keep sessions and messages, as these are historical data to browse even when the user doesn't need the agent anymore for new conversations.

guillaumejacquart (approved):
looks good to me

packages/@n8n/db/src/migrations/common/1765886667897-AddAgentIdForeignKeys.ts
Cadiac: Yeah, seems better & more intuitive than deletion

Cadiac (approved):

PR #23285, opened 2025-12-16
chore(core): Update workflow status endpoint to use absolute URL

afitzek:
## Summary

Updates the workflow status endpoint to return absolute authorization URLs instead of relative paths. The controller now injects `UrlService` to retrieve the instance base URL and constructs fully qualified URLs (e.g., `https://n8n.example.com/rest/credentials/{id}/authorize?resolverId=...`) for credential authorization endpoints. This ensures clients receive complete URLs that can be used directly without requiring additional context about the server's base path.

closes https://linear.app/n8n/issue/PAY-4334/update-workflow-status-endpoint-to-include-full-url

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/dynamic-credentials.ee/workflow-status.controller.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/workflow-status.controller.ts
@@ -6,12 +6,14 @@ import { CredentialResolverWorkflowService } from './services/credential-resolve
 import { WorkflowExecutionStatus } from '@n8n/api-types';
 import { getBearerToken } from './utils';
 import { UrlService } from '@/services/url.service';
+import { GlobalConfig } from '@n8n/config';
 
 @RestController('/workflows')
 export class WorkflowStatusController {
 	constructor(
 		private readonly credentialResolverWorkflowService: CredentialResolverWorkflowService,
 		private readonly urlService: UrlService,
+		private readonly globalConfig: GlobalConfig,
 	) {}
 
 	/**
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/dynamic-credentials.ee/workflow-status.controller.ts
```diff
@@ -37,14 +39,16 @@ export class WorkflowStatusController {
...
+
 		const executionStatus: WorkflowExecutionStatus = {
 			workflowId,
 			readyToExecute: isReady,
 			credentials: status.map((s) => ({
 				credentialId: s.credentialId,
 				credentialStatus: s.status,
 				credentialType: s.credentialType,
-				authorizationUrl: `/credentials/${s.credentialId}/authorize?resolverId=${encodeURIComponent(s.resolverId)}`,
+				authorizationUrl: `${basePath}/rest/credentials/${s.credentialId}/authorize?resolverId=${encodeURIComponent(s.resolverId)}`,
```
guillaumejacquart: I think we should use this: https://github.com/n8n-io/n8n/blob/master/packages/@n8n/config/src/configs/endpoints.config.ts#L93

guillaumejacquart (commented):
small comment

packages/cli/src/modules/dynamic-credentials.ee/workflow-status.controller.ts
afitzek: Very good point. Thank you 👍

guillaumejacquart (approved):

PR #23289, opened 2025-12-16
feat(core): Add dedicated workflow:publish project scope to use instead of the generic workflow:update

konstantintieber:
## Summary

See https://n8nio.slack.com/archives/C0A1FPDDV53/p1765982214066069 for manual testing confirmation.

In order to give more granular control on custom project roles, we need a dedicated project scope for publishing/unpublishing a workflow.
Until now, only workflow:update was required to do this.

On API token level, there already existed workflow:activate and workflow:deactivate, but these are really only used for requests sent to our public API via an API token. For this new `workflow:publish` project scope, we're using the new naming from the start.

A migration is added (with unit tests) that ensures that any project roles that have the `workflow:update` scope also get the `workflow:publish` scope so that we don't introduce a breaking change.

| Before  | After |
| ------------- | ------------- |
| Screenshot 2025-12-16 at 15 20 41 | Screenshot 2025-12-16 at 15 20 16 |

Screenshot of new tooltip shown when `workflow:publish` permission is missing:
image

Msql migration is no longer needed, since we dropped support for it in 2.0.

PAY-4199

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/permissions/src/roles/scopes/workflow-sharing-scopes.ee.ts
```diff
@@ -3,6 +3,7 @@ import type { Scope } from '../../types.ee';
 export const WORKFLOW_SHARING_OWNER_SCOPES: Scope[] = [
 	'workflow:read',
 	'workflow:update',
+	'workflow:publish',
```
konstantintieber, line 6: I didn't look into how workflow sharing scopes work yet.
Wondering whether we need to migrate existing "shared workflows" in the DB that used to grant `workflow:update` to now also include `workflow:publish`.

packages/cli/src/workflows/workflows.controller.ts
```diff
@@ -529,7 +529,7 @@ export class WorkflowsController {
 	}
 
 	@Post('/:workflowId/activate')
-	@ProjectScope('workflow:update')
+	@ProjectScope('workflow:publish')
```
konstantintieber, line 532: This is the key change that switches from checking `workflow:update` to `workflow:publish` for activating/deactivating a workflow.

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -143,7 +143,7 @@ defineExpose({
 				<N8nIcon icon="circle-check" color="success" size="xlarge" :class="$style.icon" />
 			</N8nTooltip>
 		</div>
-		<div v-if="!isArchived && workflowPermissions.update" :class="$style.publishButtonWrapper">
+		<div v-if="!isArchived && workflowPermissions.publish" :class="$style.publishButtonWrapper">
```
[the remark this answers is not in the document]
konstantintieber: Added tests in https://github.com/n8n-io/n8n/pull/23289/commits/7200b13583b483888263521c7fee91368b6e6701

packages/@n8n/db/src/migrations/postgresdb/1765892113000-AddWorkflowPublishScopeToProjectRoles.ts
```diff
@@ -0,0 +1,3 @@
+import { AddWorkflowPublishScopeToProjectRoles1765892113000 as BaseMigration } from '../common/1765892113000-AddWorkflowPublishScopeToProjectRoles';
+
+export class AddWorkflowPublishScopeToProjectRoles1765892113000 extends BaseMigration {}
```
afitzek: I think this class is unused?

packages/@n8n/db/src/migrations/sqlite/1765892113000-AddWorkflowPublishScopeToProjectRoles.ts
```diff
@@ -0,0 +1,3 @@
+import { AddWorkflowPublishScopeToProjectRoles1765892113000 as BaseMigration } from '../common/1765892113000-AddWorkflowPublishScopeToProjectRoles';
+
+export class AddWorkflowPublishScopeToProjectRoles1765892113000 extends BaseMigration {}
```
afitzek: I think this class is unused?

packages/@n8n/permissions/src/roles/scopes/workflow-sharing-scopes.ee.ts
afitzek: I think this should be fine 👍, afaik these should be migrated to the roles, and therefore updated.

afitzek (commented):
It looks good to me, I think we don't need the sqlite and postgres specific migration classes, and we should remove them. Other than that, nicely done. 👍

packages/@n8n/db/src/migrations/sqlite/1765892113000-AddWorkflowPublishScopeToProjectRoles.ts
konstantintieber: True that :D
Removed in https://github.com/n8n-io/n8n/pull/23289/commits/c1be58c408eb1fe7b89a23b7c2bfda6261ab0876

packages/@n8n/db/src/migrations/postgresdb/1765892113000-AddWorkflowPublishScopeToProjectRoles.ts
konstantintieber: True that :D
Removed in https://github.com/n8n-io/n8n/pull/23289/commits/c1be58c408eb1fe7b89a23b7c2bfda6261ab0876

afitzek (approved):
🚀

packages/cli/src/public-api/v1/handlers/workflows/workflows.handler.ts
```diff
@@ -329,7 +329,7 @@ export = {
 	],
 	activateWorkflow: [
 		apiKeyHasScope('workflow:activate'),
```
guillaumejacquart, line 331: should we rename workflow:activate api scope as well at some point ?

guillaumejacquart (approved):
I focused the review on the migrations and public API, did not find anything fishy. Well done !

packages/cli/src/public-api/v1/handlers/workflows/workflows.handler.ts
konstantintieber: I think we should. But at the same time I'm thinking we could introduce a new `workflow:publish` API scope while deprecating these old ones.

Mainly for the reason that I worry we're inviting customers to shoot themselves in the foot by having dedicated scopes for `publish` and `unpublish`. From a product perspective I think having a single scope that controls access to both publish & unpublish makes more sense.

What do you think?

PR #23291, opened 2025-12-16
refactor(core): Drop self-referencing entity fields from ChatHubMessage entity to solve deep type instantiation issues

Cadiac:
## Summary

`chat_hub_message` table represents a graph / tree structure, and it references to itself to keep track of previous messages of each message & retries / revisions. This works, but sometimes appears to lead into

image

```
Type instantiation is excessively deep and possibly infinite
```

issues at unrelated code changes. Sometimes these go away with `pnpm reset`, but now they're blocking unrelated PRs from building on CI.

Since we never load these entity relations on the backend and instead just construct the message graph at FE based no the IDs (and trust on the foreign keys to keep the data structure sane) dropping these entity relations until we have a real use for them could hopefully help with these random build issues.

Lets add these back later if we need them.

https://n8nio.slack.com/archives/C03MZF137FV/p1763993338525809

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/chat-hub/chat-hub-message.entity.ts
+++ b/packages/cli/src/modules/chat-hub/chat-hub-message.entity.ts
@@ -5,7 +5,6 @@ import {
 	Entity,
 	ManyToOne,
 	JoinColumn,
-	OneToMany,
 	type Relation,
 	PrimaryGeneratedColumn,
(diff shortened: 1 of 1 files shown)


autologie (approved):
Good idea, thank you for the fix

PR #23298, opened 2025-12-16
fix(editor): Previous message does not exist error when sending chat message

autologie:
## Summary

This PR fixes the bug in sending chat message that causes all subsequent attempts of sending message to fail after one failed message submission, due to inconsistent message history between client and server. This happens typically when attachment file size is too large or the client is temporarily offline.

This PR implements a fix by deferring message insertion in chat store until the first chunk of streaming is received (which guarantees the message is already stored in DB).

This change also makes it possible for the user to modify the message after a failed submission and re-submit it.

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.test.ts
autologie, line 1: Mostly restoring from an old branch I didn't manage to finish

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.test.ts
```diff
@@ -0,0 +1,625 @@
...
+			const rendered = renderComponent({ pinia });
+
+			expect(await rendered.findByText(/Failed to load conversation/i)).toBeInTheDocument();
+
+			await vi.waitFor(() => expect(mockRouterPush).toHaveBeenCalledWith({ name: 'chat' }));
+		});
+
+		it.todo(
+			'handles when the agent selected for the conversation is not available anymore',
+			async () => {
```
autologie, line 354: Realized this is broken because chatStore.getAgent() was changed to never return null. I'll make a fix in a separate PR.

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
```diff
@@ -482,7 +483,9 @@ function handleRegenerateMessage(message: ChatHubMessageDto) {
 
 	const messageToRetry = message.id;
 
-	chatStore.regenerateMessage(
+	editingMessageId.value = undefined;
```
autologie, line 489: Unrelated bug, but editing a message should be canceled when regeneration begins, similar to when submitting a new message

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
autologie, line 535: With changes made in this PR, it becomes unnecessary to generate message ID in frontend, but I guess there's no strong reason to move UUID generation to backend.

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
```diff
@@ -391,6 +394,18 @@ export const useChatStore = defineStore(CHAT_STORE, () => {
 			return;
 		}
 
+		if (!streaming.value.retryOfMessageId) {
```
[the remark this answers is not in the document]
autologie: When editing AI message, streaming response doesn't happen and thus this code is not reachable.

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
Cadiac: Not right now I think, we'd need a convenient way to relay back the ID I guess 🤔 and the other response modes might complicate that further.

Cadiac: Should this telemetry happen regardess if streaming successfully starts / be tracked immediately before we attempt to start the edit request stream? same applies to regenerate and sending chat messages

Cadiac (dismissed):
Seems to work better.
There's still ways to get into this state, for instance by disabling the chat provider you were using and trying to chat with it,

Image

this happens after second message - we never get to the streaming and instead the response gets rejected with 400 bad request error 🤔 

But this PR as is is an improvement 👍

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
autologie: I think the telemetry is to track the intent to send, so I think yes (We should probably also track how often the user encounters an error, but we can discuss later). Fixed [7da5b1e](https://github.com/n8n-io/n8n/pull/23298/commits/7da5b1e52763c52a80187a05d53fc507f6468c90)

Cadiac:
> interesting, do we start streaming before persisting user's message?

I think in this case we don't even persist the message, as it fails on the validation checks & the operation gets rolled back 🤔

Cadiac (approved):

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
autologie: I think `previousMessageId` in the first chunk metadata corresponds to the user's message ID if we want, but agreed not necessary to change now

Cadiac: ah that's true.

I'll consider changing this on the other response modes PR, I'm a bit unsure on how to do `waiting` state for instance at the moment still..

autologie:
> > interesting, do we start streaming before persisting user's message?
> 
> I think in this case we don't even persist the message, as it fails on the validation checks & the operation gets rolled back 🤔

The implemented fix is intended to catch any errors that happen before starting streaming.. if not something is flawed 🤔

Cadiac:
This must've been me confusing branches - I can't replicate the issue anymore.

PR #23304, opened 2025-12-16
feat(core): Add TLS support to Syslog client

geemanjs:
## Summary

Adds support for TLS on syslog log streaming messages.

This was a little trickier than expected as the original library in use was un-updated for 5 years.

The syslog-client was cobbled together using the original, claude code and human sprinkles to tidy things up and make it all work. 

I wanted to validate end-to-end how this would work - so I added an integration test where we stand up a server to receive tls messages, stand up a test n8n server, register an eventbus destination with syslog (pointing at our tls server), then trigger an "event" and validate it was received by the tls server. 

All Certificates / Keys in this PR are "duds" and not used by anything other than the tests. 

resolves PAY-2930

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/eventbus/message-event-bus-destination/message-event-bus-destination-syslog.ee.ts
```diff
@@ -82,18 +92,16 @@ export class MessageEventBusDestinationSyslog
...
-						? syslog.Severity.Error
-						: syslog.Severity.Debug,
+					severity: msg.eventName.toLowerCase().endsWith('error') ? Severity.Error : Severity.Debug,
 					msgid: msg.id,
 					timestamp: msg.ts.toJSDate(),
 				},
 				async (error) => {
 					if (error?.message) {
-						this.logger.debug(error.message);
+						this.logger.error(error.message);
```
geemanjs, line 102: I am assuming bumping this to error is correct.. Do we have logging standards?

I feel like a bigger / better change would be to "test" the "destination" when it's created to validate the certificate works. I will raise a ticket to do this as it is somewhat out of scope here. 

It's also possible though that the certificate could expire in which case this will fire.. Do we have a way of sort of "alerting" the customer that something isn't working?

packages/cli/test/integration/eventbus/syslog-tls.test.ts
```diff
@@ -0,0 +1,125 @@
+import { Logger } from '@n8n/backend-common';
```
geemanjs, line 1: This file is probably the best bit of this PR - it shows the full e2e of setting up and sending an event using tls.

It also shows what happens when the certificate is invalid / out of date.

packages/@n8n/syslog-client/src/schemas.ts
```diff
@@ -0,0 +1,108 @@
...
+	Transport.Udp, // 2
+	Transport.Tls, // 3
+	Transport.Unix, // 4
+] as const;
+
+/**
+ * Valid Facility values.
+ * Must be manually maintained because const enums don't exist at runtime.
+ */
+const FACILITY_VALUES = [
```
afitzek: Could we maybe build the Facility type based on the constant array. We are doing something similar in packages/@n8n/permissions/src/types.ee.ts with the Scopes.

packages/@n8n/syslog-client/test/fixtures/certificate.pem
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
afitzek, line 1: This certificate is valid till `Dec 16 13:24:01 2026 GMT` does it make sense to make this valid much longer 25 years? Otherwise I would imagine the test will start failing on that date next year.

packages/cli/src/eventbus/message-event-bus-destination/message-event-bus-destination-syslog.ee.ts
```diff
@@ -47,21 +48,30 @@ export class MessageEventBusDestinationSyslog
...
-		this.facility = options.facility ?? syslog.Facility.Local0;
+		this.facility = options.facility ?? Facility.Local0;
 		this.app_name = options.app_name ?? 'n8n';
 		this.eol = options.eol ?? '\n';
 		this.expectedStatusCode = options.expectedStatusCode ?? 200;
 
-		this.client = syslog.createClient(this.host, {
+		if (this.protocol === 'tls') {
+			if (!options.tlsCa) {
+				console.log('Syslog - No TLS CA set');
```
afitzek: We shouldn't have console log, this can either be a log message or an error.

packages/cli/src/eventbus/message-event-bus-destination/message-event-bus-destination-syslog.ee.ts
afitzek: I think the error should be fine here.

packages/cli/test/integration/eventbus/support/certificate.pem
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
afitzek, line 1: This certificate will also expire on `Dec 16 13:24:01 2026 GMT`, which will make the tests suddenly fail. Shall we extend the validity to much longer?

afitzek (commented):
Are we supposed to also update the UI to allow customers to configure the TLS option for syslog log streaming, or will this happen in a follow up PR?

The integration looks good, just some small comments.

packages/cli/test/integration/eventbus/support/certificate.pem
geemanjs: Should be 2053 now 😄

packages/@n8n/syslog-client/test/fixtures/certificate.pem
geemanjs: Should be 2053 now 😄

konstantintieber (dismissed):
Great work!

Spun up a local syslog server using docker to test the previously supported TCP-only use case and it worked as expected:

image

konstantintieber (approved):

PR #23309, opened 2025-12-16
fix: Extracts data from PDF without memory issues

shortstacked:
## Summary

The previous implementation had memory issues when extracting data from PDF files, this is a hopeful fix as I'm unable to reproduce this issue without the hardware in question.

This commit addresses this by:
- Adding a polyfill for DOMMatrix to support environments without canvas.
- Removing unnecessary dependency installation.

Fixes CAT-1945

https://linear.app/n8n/issue/CAT-1945/community-issue-out-of-memory-when-using-extract-from-file-node-with
https://github.com/n8n-io/n8n/issues/19359

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/docker/images/n8n/Dockerfile
+++ b/docker/images/n8n/Dockerfile
@@ -32,8 +32,6 @@ RUN cd /usr/local/lib/node_modules/n8n && \
     mkdir -p /home/node/.n8n && \
     chown -R node:node /home/node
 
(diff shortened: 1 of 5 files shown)


tomi (commented):
Can we build a custom image from this and ask the affected users to try it out before mergin?

shortstacked:
> Can we build a custom image from this and ask the affected users to try it out before mergin?

Good idea, I've built the image and reference it in the GH issue for now.

tomi (approved):
🚀

PR #23333, opened 2025-12-17
fix(editor): Fix wf nodes not updated correctly on workflow activated/deactivated events

scdekov:
## Summary

When n8n is running in multi main mode there are socket messages being sent when workflow is activated/deactivated. That is also the the case when settings are being changed which is triggering deactivated and then activate messages. The way we were updating the local state of the workflow in the handlers of those messages were not correct as they were not setting the node parameter defaults.

https://linear.app/n8n/issue/ADO-4553/bug-setting-workflow-settings-breaks-node-configuration

## Review / Merge checklist

- [ ] Tests included. 


CharlieKolb (approved):

dariacodes (approved):
Checked that it's fixed now! Good job! 🚀

PR #23337, opened 2025-12-17
chore(core): Refactor external secrets

guillaumejacquart:
## Summary

Refactored monolithic ExternalSecretsManager into specialized services for better separation of concerns and testability.

### Architecture Changes

**Before**: Single manager class handling everything
**After**: 6 specialized services
- SettingsStore - Encrypted settings persistence with caching
- ProviderRegistry - Provider collection management
- ProviderLifecycle - Init/connect/disconnect operations
- RetryManager - Exponential backoff retry mechanism
- SecretsCache - Secrets refresh and caching
- ExternalSecretsManager - Orchestrates all services

### Key Technical Fixes

- **Retry mechanism**: Made connectProvider throw on failure (was returning result objects), enabling proper retry flow
- **Provider state**: Added setState('initializing') on disconnect
- **Retry cleanup**: Added cancelRetry calls to prevent zombie retries on reload/disconnect
- **Error handling**: Added try-catch in updateProvider controller endpoint

https://linear.app/n8n/issue/PAY-4314/external-secrets-provider-connection-doesnt-recover-on-worker-if

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
+++ b/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
@@ -1,273 +1,114 @@
-import { LicenseState, Logger } from '@n8n/backend-common';
-import { SettingsRepository } from '@n8n/db';
+import { Logger } from '@n8n/backend-common';
+import { Time } from '@n8n/constants';
 import { OnPubSubEvent } from '@n8n/decorators';
 import { Service } from '@n8n/di';
-import { Cipher, type IExternalSecretsManager } from 'n8n-core';
-import { jsonParse, type IDataObject, ensureError, UnexpectedError } from 'n8n-workflow';
+import { type IExternalSecretsManager } from 'n8n-core';
+import { UnexpectedError, type IDataObject } from 'n8n-workflow';
 
+import { NotFoundError } from '@/errors/response-errors/not-found.error';
 import { EventService } from '@/events/event.service';
 import { Publisher } from '@/scaling/pubsub/publisher.service';
 
-import {
-	EXTERNAL_SECRETS_DB_KEY,
-	EXTERNAL_SECRETS_INITIAL_BACKOFF,
-	EXTERNAL_SECRETS_MAX_BACKOFF,
-} from './constants';
 import { ExternalSecretsProviders } from './external-secrets-providers.ee';
 import { ExternalSecretsConfig } from './external-secrets.config';
+import {
+	ProviderConnectResult,
+	ExternalSecretsProviderLifecycle,
+} from './provider-lifecycle.service';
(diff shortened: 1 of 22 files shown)


packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
```diff
@@ -1,352 +1,347 @@
...
+	// Event Handlers
+	// ========================================
+
+	@OnPubSubEvent('reload-external-secrets-providers')
+	async reloadAllProviders(): Promise<void> {
+		this.logger.debug('Reloading all external secrets providers');
+
+		const newSettings = await this.settingsStore.reload();
+
+		// Reload/add providers from new settings
```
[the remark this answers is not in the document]
guillaumejacquart: This is done on the reloadProvider function already. the check here was a duplicate

packages/cli/src/modules/external-secrets.ee/external-secrets.controller.ee.ts
```diff
@@ -71,13 +71,13 @@ export class ExternalSecretsController {
...
-		if (resp) {
-			res.statusCode = 200;
-		} else {
+		try {
+			await this.secretsService.updateProvider(providerName);
+			return { updated: true };
+		} catch (error) {
 			res.statusCode = 400;
+			return { updated: false };
 		}
```
ireneea, line 85: [optional] I can see that there is no logs at all at the moment, but I think it could be useful to add some extra logs

packages/cli/src/modules/external-secrets.ee/provider-lifecycle.service.ts
```diff
@@ -0,0 +1,134 @@
...
+	async disconnect(provider: SecretsProvider): Promise<void> {
+		try {
+			await provider.disconnect();
+			provider.setState('initializing');
+			this.logger.debug(`Disconnected secrets provider ${provider.displayName}`);
+		} catch (error) {
+			this.logger.warn(`Error disconnecting provider ${provider.displayName}`, {
+				error: ensureError(error),
+			});
+		}
```
ireneea, line 121: Should we set the provider state to 'error' here?

guillaumejacquart: I don't think so, because a provider that could not disconnect does not mean it's in an error state. It just means it could not disconnect. Anyway in practice, the disconnect should not fail (most of them are doing nothing, only the vault one is stopping a timeout)

ireneea: Makes sense 👍

ireneea (approved):

PR #23349, opened 2025-12-17
fix(ai-builder): Connection validation respects node version and type

mike12345567:
## Summary

Updating connection validation to make sure that we are checking the version as well as the node type. Nodes with multiple versions can have very different connection expectations.

Without this change the workflow builder can often get stuck on building agent nodes as the connections are all considered invalid.

## Review / Merge checklist

- [ ] Tests included. 


burivuhster (approved):
🚀

PR #23350, opened 2025-12-17
feat(core): Add user email parameter when activating EULA license

cstuncsik:
## Summary

This PR enhances license activation to capture the signed-in user's email when they accept a EULA license, enabling better tracking and compliance for license agreements.

https://linear.app/n8n/issue/IAM-31/add-signed-in-user-email-when-activating-a-eula-license

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/license/license.service.ts
```diff
@@ -109,9 +109,13 @@ export class LicenseService {
 		return this.license.getManagementJwt();
 	}
 
-	async activateLicense(activationKey: string, eulaUri?: string) {
+	async activateLicense(activationKey: string, eulaUri?: string, userEmail?: string) {
```
konstantintieber: What is the intention behind this param being optional?
As I understand the [ticket requirement](https://linear.app/n8n/issue/PAY-4069/add-signed-in-user-email-when-activating-a-eula-license), we should rather fail activation if the email is not known.

cstuncsik: Yes, thanks
Fixed

konstantintieber: I see you commited this: https://github.com/n8n-io/n8n/pull/23350/commits/0779d1c261610c049eff38f14cb96f5528ec4e60

But I don't understand why the implementation should ever support the userEmail to not be defined.
As I understand this does not exist as business case.

cstuncsik: It's either just activationKey or activationKey with both eulaUri and userEmail
In the function overload types it is not optional anymore so in theory the compiler would stop if you didn't pass but there is a runtime check as well just in case

konstantintieber: Last I heard the product requirment is that we always want to know the userEmail of the person who accepted the EULA. Is that not correct?

konstantintieber: The backend returns a 400 error when no user email is provided with a eula license: https://github.com/n8n-io/license-management/pull/164/files#diff-f21acad4b54ce32e74fb833f1805cca6b4cc50c6b1bc8ee130fd755641f43e78R117

konstantintieber: Ah now I get it. If a license is activated only with a key and no eulaUri, there is no need for the email. And the overloads correctly relfect that 👍 
Sorry for the confusion.

konstantintieber (dismissed):

packages/cli/src/license/license.service.ts
cstuncsik: Yes, that can't happen now, as we prevent calling the License SDK activation when the `activationKey` is provided, but either the `eulaUri` or the `userEmail` is missing.

And now seeing that in the license management  repo, maybe I shouldn't have done anything, as the HTTP error would have been forwarded anyway

konstantintieber: I think for easier understanding of the functionality for other code authors, it's good to reflect the business logic in the types.

package.json
```diff
@@ -89,6 +89,7 @@
       "sqlite3"
     ],
     "overrides": {
+      "@codemirror/state": "6.5.3",
```
konstantintieber: Why is this override necessary with these changes? Is it related to a change inside `@n8n_io/license-sdk`?

cstuncsik: No, I think it was accidental somehow when resolving some conflicts
Thanks for catching, fixed

konstantintieber (approved):

PR #23353, opened 2025-12-17
refactor(core): Remove oversized payloads from logs

ivov:
While debugging internal this week, I found logs getting flooded by binary buffers and registered webhooks.

### DIFF
--- a/packages/cli/src/webhooks/webhook-request-handler.ts
+++ b/packages/cli/src/webhooks/webhook-request-handler.ts
@@ -23,7 +23,6 @@ import {
 	isWebhookResponse,
 	isWebhookStreamResponse,
 } from '@/webhooks/webhook-response';
-import { WebhookService } from '@/webhooks/webhook.service';
 import type {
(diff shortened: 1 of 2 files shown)


packages/cli/src/webhooks/webhook-request-handler.ts
```diff
@@ -81,10 +80,7 @@ class WebhookRequestHandler {
 			const logger = Container.get(Logger);
 
 			if (e instanceof WebhookNotFoundError) {
-				const currentlyRegistered = await Container.get(WebhookService).findAll();
-				logger.error(`Received request for unknown webhook: ${e.message}`, {
-					currentlyRegistered: currentlyRegistered.map((w) => w.display()),
-				});
+				logger.error(`Received request for unknown webhook: ${e.message}`);
```
mfsiega, line 83: This also seems like a big improvement from a security standpoint, to not reveal all of the registered webhooks.

mfsiega (approved):
LGTM.

PR #23357, opened 2025-12-17
fix: Don't encrypt state for hooks calls

BGZStephen:
## Summary

This PR fixes an issue with n8n cloud hooks expecting a base64 string for hooks calls, when we actually have an encrypted state, fixed by allowing hooks to use the base64 state

## Review / Merge checklist

- [x] Tests included. 


guillaumejacquart (approved):
Nicely done. Might be worth adding a comment on l385 to explain why we pass the unencrypted state

PR #23360, opened 2025-12-17
fix(core): Fix orchestration init modules

guillaumejacquart:
## Summary

Add re-registering of the pubsub handlers after modules initialization on the worker.
This is to fix the fact that modules running on worker instances and registering pubsub handlers with the OnPubsubEvent decorator were initialized after the pubsub handler initialization, leading to the modules handler never registered.
We use the same technique as for start and webhook commands, which both re-register on the server.ts file

This PR "just" changes the order so that the initOrchestration is called after the modules are initialized

https://linear.app/n8n/issue/PAY-4314/external-secrets-provider-connection-doesnt-recover-on-worker-if

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/commands/worker.ts
+++ b/packages/cli/src/commands/worker.ts
@@ -117,6 +117,10 @@ export class Worker extends BaseCommand<z.infer<typeof flagsSchema>> {
 
 		await this.moduleRegistry.initModules(this.instanceSettings.instanceType);
 
+		// Re-register pubsub event handlers after modules have been initialized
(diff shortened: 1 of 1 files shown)


mfsiega (commented):
The change looks basically reasonable to me, and it seems safe enough.

For my context:
- Is there an associated bug/issue that this is addressing?
- What testing did you do? I think we would want to test that we can bring up a multi-main deployment - do we have any automated tests that cover that setup?

mfsiega (approved):
Let's ship this :)

PR #23365, opened 2025-12-17
fix(editor): Save only once on publish

dariacodes:
## Summary

This PR introduces a separate store for autosave to be able to track its state in different components. During publish we check whether autosave is already running or not, and save the workflow immediately only if needed.

Console logs will be removed later.

Closes [ADO-4558](https://linear.app/n8n/issue/ADO-4558/bug-save-on-publishing-does-not-cancel-scheduled-autosave)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
@@ -5,7 +5,8 @@ import type { FolderShortInfo } from '@/features/core/folders/folders.types';
 import type { IWorkflowDb } from '@/Interface';
 import type { PermissionsRecord } from '@n8n/permissions';
 import { computed, onBeforeUnmount, onMounted, ref, useTemplateRef } from 'vue';
-import { WORKFLOW_PUBLISH_MODAL_KEY } from '@/app/constants';
+import { storeToRefs } from 'pinia';
+import { WORKFLOW_PUBLISH_MODAL_KEY, AutoSaveState } from '@/app/constants';
 import { N8nButton, N8nIcon, N8nTooltip } from '@n8n/design-system';
 import { useI18n } from '@n8n/i18n';
 import { useUIStore } from '@/app/stores/ui.store';
@@ -14,6 +15,7 @@ import TimeAgo from '@/app/components/TimeAgo.vue';
 import { getActivatableTriggerNodes } from '@/app/utils/nodeTypesUtils';
 import { useWorkflowSaving } from '@/app/composables/useWorkflowSaving';
 import { useRouter } from 'vue-router';
+import { useWorkflowAutosaveStore } from '@/app/stores/workflowAutosave.store';
 import {
 	getLastPublishedVersion,
 	generateVersionName,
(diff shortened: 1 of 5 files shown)


packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -48,16 +54,63 @@ const isWorkflowSaving = computed(() => {
...
+		return false;
+	}
+	console.log('[Publish] ✅ Save completed');
+	return true;
+};
+
 const onPublishButtonClick = async () => {
 	// If there are unsaved changes, save the workflow first
 	if (uiStore.stateIsDirty || props.isNewWorkflow) {
 		autoSaveForPublish.value = true;
```
scdekov, line 76: Should we just move this in `saveBeforePublish`?

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -48,16 +54,63 @@ const isWorkflowSaving = computed(() => {
...
-		const saved = await saveCurrentWorkflow({}, true);
-		autoSaveForPublish.value = false;
-		if (!saved) {
-			// If save failed, don't open the modal
-			return;
+
+		// Cancel autosave if if scheduled or wait for it to finish if in progress
+		// Save immediately if autosave idle or cancelled
+		switch (autoSaveState.value) {
```
scdekov: Can we simplify this to something like
```
if (autoSaveState.value === AutoSaveState.InProgress && pendingAutoSave.value) {
  await pendingAutoSave.value
} else if (autoSaveState.value === AutoSaveState.Scheduled) {
  cancelAutoSave();
}

await saveBeforePublish()
```

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
```diff
@@ -54,6 +58,9 @@ export function useWorkflowSaving({
 	const { getWorkflowDataToSave, checkConflictingWebhooks, getWorkflowProjectRole } =
 		useWorkflowHelpers();
 
+	const autosaveStore = useWorkflowAutosaveStore();
+	const { autoSaveState, pendingAutoSave } = storeToRefs(autosaveStore);
```
scdekov: Why do we need this?

scdekov (commented):
One small UI thing, while having a scheduled save and click publish both the publish button and the history one are loading. Can fix it so that only the publish one does in that case? 

https://github.com/user-attachments/assets/e8dffeda-1566-4ba2-a6f2-126334d0cb90

dariacodes:
@scdekov They are spinning together, because after clicking on "Publish" we start saving immediately and wait until it finishes before we open the modal. One way to not show it would be to watch the autosave store changes in `WorkflowHistoryButton` component and animate it only if autosave is in progress (and not "manual" autosave invoked by publish). But I thought we want this icon to spin whenever the workflow is being saved.
A cleaner way could be to open the modal straightaway and block "Publish" until save is completed. If there are any errors during save, we could also show it right in the modal.

scdekov (approved):

PR #23366, opened 2025-12-17
fix(core): Use Redis prefix everywhere (no-changelog)

mfsiega:
## Summary

Make sure we apply the `N8N_REDIS_KEY_PREFIX` whenever we publish or subscribe to Redis PubSub.

If two deployments of n8n share a single Redis, we try to support isolating them by supplying a `N8N_REDIS_KEY_PREFIX` environment variable.

However, we don't apply this prefix for certain messages -- worker status, webhook activation. This means these messages can end up at the wrong instance, causing errors and confusion. For example, the `/settings/workers` endpoint returns all of the workers of all instances using the Redis server.

This change ensures that the prefix is applied for all of these usages.

https://linear.app/n8n/issue/CAT-1761

## Review / Merge checklist

- [ ] Tests included. 


tomi (approved):
🚀

PR #23372, opened 2025-12-17
feat(core): Add log streaming events for executions

MarcL:
## Summary

Adds two new log streaming events for audit logging:
- `n8n.audit.user.execution.deleted`: Emitted when user deletes executions
- `n8n.audit.user.workflow.executed`: Emitted when user manually runs or retries a workflow

Both events include user context for audit tracking.

https://linear.app/n8n/issue/PAY-2936

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/events/maps/relay.event-map.ts
+++ b/packages/cli/src/events/maps/relay.event-map.ts
@@ -125,12 +125,22 @@ export type RelayEventMap = {
 		userIdList: string[];
 	};
 
-	'workflow-executed-by-user': {
-		user: UserLike;
+	'workflow-executed': {
+		user?: UserLike;
 		workflowId: string;
 		workflowName: string;
 		executionId: string;
-		executionMode: 'manual' | 'retry';
+		source:
+			| 'user-manual'
+			| 'user-retry'
+			| 'webhook'
+			| 'trigger'
+			| 'error'
+			| 'cli'
+			| 'integrated'
+			| 'internal'
+			| 'evaluation'
+			| 'chat';
 	};
 
 	// #endregion
--- a/packages/cli/src/events/maps/relay.event-map.ts
+++ b/packages/cli/src/events/maps/relay.event-map.ts
@@ -125,7 +125,7 @@ export type RelayEventMap = {
 		userIdList: string[];
 	};
 
-	'workflow-executed': {
+	'workflow-executed-by-user': {
 		user: UserLike;
 		workflowId: string;
 		workflowName: string;
--- a/packages/cli/src/events/maps/relay.event-map.ts
+++ b/packages/cli/src/events/maps/relay.event-map.ts
@@ -130,17 +130,16 @@ export type RelayEventMap = {
 		workflowId: string;
 		workflowName: string;
 		executionId: string;
-		source:
-			| 'user-manual'
-			| 'user-retry'
-			| 'webhook'
-			| 'trigger'
-			| 'error'
-			| 'cli'
-			| 'integrated'
-			| 'internal'
-			| 'evaluation'
-			| 'chat';
+		source: 'user-manual' | 'user-retry';
+		// TODO: To be added in future PR
+		// | 'webhook'
+		// | 'trigger'
+		// | 'error'
+		// | 'cli'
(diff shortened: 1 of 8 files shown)


packages/cli/src/events/maps/relay.event-map.ts
```diff
@@ -125,6 +125,14 @@ export type RelayEventMap = {
 		userIdList: string[];
 	};
 
+	'workflow-executed': {
+		user: UserLike;
+		workflowId: string;
+		workflowName: string;
+		executionId: string;
+		executionMode: 'manual' | 'retry';
```
konstantintieber: Is there really only "manual" and "retry" modes?
Does this event not concern itself with workflows that are executed by a schedule trigger or webhook trigger etc.?

MarcL: @konstantintieber I went back and forth on this because we wouldn't have user info for the events that are triggered via other triggers. I chose to just go with the events that were just for the user, as the ticket suggest just needing to know about user triggered events.

Would it be better to have a `workflow-executed` event with all execution modes, including `trigger` and `error` I think - will double-check this, and then have optional user context? WDYT?

konstantintieber: Tricky question indeed.

The audit event name is `n8n.audit.user.workflow.executed`. From the `user`, I might infer that this is about manual execution. I guess a non-manual equivalent would be `n8n.audit.workflow.executed` (following the names we have today)?

The thing is, on the `workflow-relay-event-map`, the context of the audit event names is missing. So I'm thinking we maybe should use a more specific key there like `workflow-executed-manual`.

MarcL: I've renamed it to `workflow-executed-by-user` - does that make more sense now?

BGZStephen: Just chucking my opinion in for what it's worth. I'm more a fan of `n8n.audit.workflow.executed`, and distinguishing the `source` as being a user (or other), otherwise we may end up with numerous event types just for a workflow execution.

I'd rather have 1 event type, but with multiple filterable fields (source, type, etc) than have multiple event types to wade through.

MarcL: @BGZStephen How about I refactor this to go back to the original workflow executed event and add the source? I might just scope this PR to the user-related changes and then add a 2nd PR to do the non-user events. Thoughts?

And should we do the source like:

`'user' | 'schedule' | 'webhook' | 'error' | 'trigger'`

And then keep the execution mode? `'manual' | 'retry'`

Or use a specific source for manual or retry and no execution mode?
E.g. `'user-manual' | 'user-retry' | 'schedule' | 'webhook' | 'error' | 'trigger'`

BGZStephen: I like this, `user-manual` and `user-retry` make sense to me,

konstantintieber: Sounds good to me too

packages/cli/src/events/maps/relay.event-map.ts
```diff
@@ -125,12 +125,22 @@ export type RelayEventMap = {
...
-		user: UserLike;
+	'workflow-executed': {
+		user?: UserLike;
 		workflowId: string;
 		workflowName: string;
 		executionId: string;
-		executionMode: 'manual' | 'retry';
+		source:
+			| 'user-manual'
+			| 'user-retry'
```
konstantintieber: I understand that this typing includes values for `source` that are not implemented yet, right?
Is it planned do still add these imminently in a follow-up?
Otherwise keeping them here might lead to confusion on consumers of log streams seeing these values here, but never in the logs.

An alternative I see would be to leave them commented out, while not implemented.

MarcL: Good point @konstantintieber.
My plan was to quickly do a 2nd PR afterwards. I was trying to avoid making this PR too big.
I'll comment them out in this PR and add them back in the next.

konstantintieber (approved):

PR #23373, opened 2025-12-17
feat: Add PostgreSQL advisory lock for migrations

shortstacked:
## Summary

Ensures only one n8n instance runs database migrations and initializes auth roles service in multi-main setups to prevent race conditions.
Implements transaction-level advisory locks for PostgreSQL to avoid orphaned locks and improve reliability.
Retries lock acquisition with timeout and provides detailed error message with instructions for manual intervention.
Disables advisory lock for SQLite databases.
Runs main instances in parallel during testing to improve efficiency.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/connection/__tests__/db-connection.test.ts
+++ b/packages/@n8n/db/src/connection/__tests__/db-connection.test.ts
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/unbound-method */
 import type { Logger } from '@n8n/backend-common';
 import type { DatabaseConfig } from '@n8n/config';
-import { DataSource, type DataSourceOptions } from '@n8n/typeorm';
+import { DataSource, type DataSourceOptions, type QueryRunner } from '@n8n/typeorm';
 import { mock, mockDeep } from 'jest-mock-extended';
 import type { BinaryDataConfig, ErrorReporter } from 'n8n-core';
 import { DbConnectionTimeoutError, OperationalError } from 'n8n-workflow';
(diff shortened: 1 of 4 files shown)


tomi (commented):
This has still one issue: It does not work when pool size is configured to be one:

```
➜ DB_TYPE=postgresdb DB_POSTGRESDB_PASSWORD=postgres DB_POSTGRESDB_POOL_SIZE=1 DB_LOGGING_ENABLED=true DB_LOGGING_OPTIONS=all node bin/n8n
Initializing n8n process
query: SELECT * FROM current_schema()
query: CREATE EXTENSION IF NOT EXISTS "uuid-ossp"
query: SELECT version();
n8n ready on ::, port 5678
query: START TRANSACTION
query: SELECT pg_try_advisory_xact_lock(1850)
Database connection timed out
Database connection timed out
503 Database is not ready!
query: ROLLBACK
There was an error running database migrations
timeout exceeded when trying to connect
query: SELECT 1
query: SELECT 1
query: SELECT 1
Database connection recovered
```

This is because the transaction for the advisory lock reserves one DB connection, and there's no connection left in the pool for the transactions. So we would need to move this to the library side if we wanted to use a trx level lock.

This is also why the Postgres tests timed out. We should not allow merging of the PR in case a test times out. I created [CAT-1985](https://linear.app/n8n/issue/CAT-1985/pr-should-not-be-mergeable-if-tests-timeout) for this.

So I would suggest that we revert back to the session level lock with the timeout. In the long run we should move this locking logic into our typeorm fork.

PR #23392, opened 2025-12-18
fix: Handle nodes with missing type field in ActivateExecuteWorkflowTriggerWorkflows migration

OlegIvaniv:
## Summary
The ActivateExecuteWorkflowTriggerWorkflows1763048000000 migration crashes when encountering workflows with malformed nodes that are missing the type field.

Root cause: The migration iterates over workflow nodes and calls this.isTriggerNode(node.type) without checking if node.type exists. When a node has no type field, this results in:

Migration "ActivateExecuteWorkflowTriggerWorkflows1763048000000" failed, error: Cannot read properties of undefined (reading 'includes')

Fix: Add a defensive check node.type && before calling isTriggerNode() to skip malformed nodes gracefully.

## Review / Merge checklist

- [ ] Tests included. 


CharlieKolb (approved):

RicardoE105 (approved):
🚀

PR #23397, opened 2025-12-18
refactor(core): Make Chat db queries faster

Cadiac:
## Summary

Making Chat hub DB queries faster:
1) Removing unnecessary selects / skip loading unused relations / and otherwise simplifying the queries
2) Not making things run in transactions unnecessarily (`withTransaction` helper's optional parameter `false` skips creating transaction if one isn't passed, but still allows the function to be used within a transaction if we want to)
3) DB indices, turning `lastMessageAt` into a required column to avoid COALESCE & enabling index use

With this, the main query that happens when loading conversations drops on my local postgres with 100k sessions (only 10 users) DB from ~25ms to 1ms and it only does index scans. Same on sqlite.

Other big win would be on credential access checks and at finding Chat workflows (with enabled chat triggers), but both of those turned out to be nontrivial so saving those for a followup.

https://linear.app/n8n/issue/CHA-81/make-chat-queries-faster

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
+++ b/packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
@@ -1,0 +1,49 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class AddChatMessageIndices1766068346315 implements ReversibleMigration {
+	async up({ schemaBuilder: { addNotNull }, runQuery, escape, tablePrefix }: MigrationContext) {
+		const sessionsTable = escape.tableName('chat_hub_sessions');
+		const idColumn = escape.columnName('id');
+		const createdAtColumn = escape.columnName('createdAt');
+		const ownerIdColumn = escape.columnName('ownerId');
+		const lastMessageAtColumn = escape.columnName('lastMessageAt');
+
+		const messagesTable = escape.tableName('chat_hub_messages');
+		const sessionIdColumn = escape.columnName('sessionId');
+
+		// Backfill lastMessageAt for existing rows to allow adding a NOT NULL constraint
+		await runQuery(
+			`UPDATE ${sessionsTable}
+			SET ${lastMessageAtColumn} = ${createdAtColumn}
+			WHERE ${lastMessageAtColumn} IS NULL`,
+		);
+
+		await addNotNull('chat_hub_sessions', 'lastMessageAt');
+
+		// Index intended for faster sessionRepository.getManyByUserId queries
+		await runQuery(
+			'CREATE INDEX IF NOT EXISTS ' +
+				escape.tableName(`idx_${tablePrefix}chat_hub_sessions_owner_lastmsg_id`) +
+				` ON ${sessionsTable}(${ownerIdColumn}, ${lastMessageAtColumn} DESC, ${idColumn})`,
+		);
+
+		// Index intended for faster sessionRepository.getOneByIdAndUserId queries and joins
+		await runQuery(
+			'CREATE INDEX IF NOT EXISTS ' +
+				escape.tableName(`idx_${tablePrefix}chat_hub_messages_sessionId`) +
+				` ON ${messagesTable}(${sessionIdColumn})`,
+		);
+	}
+
+	async down({ schemaBuilder: { dropNotNull }, runQuery, escape, tablePrefix }: MigrationContext) {
+		await runQuery(
+			'DROP INDEX IF EXISTS ' +
+				escape.tableName(`idx_${tablePrefix}chat_hub_sessions_owner_lastmsg_id`),
+		);
+		await runQuery(
+			'DROP INDEX IF EXISTS ' + escape.tableName(`idx_${tablePrefix}chat_hub_messages_sessionId`),
+		);
+
+		await dropNotNull('chat_hub_sessions', 'lastMessageAt');
+	}
+}
--- a/packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
+++ b/packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
@@ -1,7 +1,7 @@
 import type { MigrationContext, ReversibleMigration } from '../migration-types';
 
 export class AddChatMessageIndices1766068346315 implements ReversibleMigration {
-	async up({ schemaBuilder: { addNotNull }, runQuery, escape, tablePrefix }: MigrationContext) {
+	async up({ schemaBuilder: { addNotNull }, runQuery, escape }: MigrationContext) {
 		const sessionsTable = escape.tableName('chat_hub_sessions');
 		const idColumn = escape.columnName('id');
 		const createdAtColumn = escape.columnName('createdAt');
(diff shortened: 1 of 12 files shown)


packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
```diff
@@ -0,0 +1,49 @@
...
+			SET ${lastMessageAtColumn} = ${createdAtColumn}
+			WHERE ${lastMessageAtColumn} IS NULL`,
+		);
+
+		await addNotNull('chat_hub_sessions', 'lastMessageAt');
+
+		// Index intended for faster sessionRepository.getManyByUserId queries
+		await runQuery(
+			'CREATE INDEX IF NOT EXISTS ' +
+				escape.tableName(`idx_${tablePrefix}chat_hub_sessions_owner_lastmsg_id`) +
```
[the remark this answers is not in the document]
Cadiac: changing this to escape.indexName

packages/cli/src/modules/chat-hub/chat-session.repository.ts
```diff
@@ -13,49 +13,66 @@ export class ChatHubSessionRepository extends Repository<ChatHubSession> {
...
-		});
+		return await withTransaction(
+			this.manager,
+			trx,
+			async (em) => {
+				await em.update(ChatHubSession, { id }, updates);
+				return await em.findOneOrFail(ChatHubSession, {
+					where: { id },
+					relations: ['messages'],
+				});
```
autologie: Looks like we don't use this returned value

packages/cli/src/modules/chat-hub/chat-session.repository.ts
```diff
@@ -13,49 +13,66 @@ export class ChatHubSessionRepository extends Repository<ChatHubSession> {
...
-				where: { id },
-				relations: ['messages'],
-			});
-		});
+	async updateLastMessageAt(id: string, trx?: EntityManager) {
+		return await withTransaction(
+			this.manager,
+			trx,
+			async (em) => {
+				await em.update(ChatHubSession, { id }, { lastMessageAt: new Date() });
```
autologie: Repository is persistence layer, isn't it better to pass `lastMessageAt` as an argument?

packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
```diff
@@ -0,0 +1,44 @@
...
+		await runQuery(
+			`CREATE INDEX IF NOT EXISTS ${escape.indexName('chat_hub_sessions_owner_lastmsg_id')}
+			ON ${sessionsTable}(${ownerIdColumn}, ${lastMessageAtColumn} DESC, ${idColumn})`,
+		);
+
+		// Index intended for faster sessionRepository.getOneByIdAndUserId queries and joins
+		await runQuery(
+			`CREATE INDEX IF NOT EXISTS ${escape.indexName('chat_hub_messages_sessionId')}
+			ON ${messagesTable}(${sessionIdColumn})`,
+		);
```
ivov, line 33: Can use the DSL here

packages/cli/src/modules/chat-hub/chat-session.repository.ts
Cadiac: I'll drop the select & probably just get rid of the specialized versions of the update

packages/cli/src/modules/chat-hub/chat-session.repository.ts
Cadiac: I'm dropping this and `updateChatTitle`, moving towards just having `updateChatSession`

autologie: Good idea

Cadiac: Addressed in [7fc1f0a](https://github.com/n8n-io/n8n/pull/23397/commits/7fc1f0ac5f0252c1f10a8d411d30e1d5a5375ad3), together with fixing merge conflicts

packages/cli/src/modules/chat-hub/chat-session.repository.ts
Cadiac: Addressed in [7fc1f0a](https://github.com/n8n-io/n8n/pull/23397/commits/7fc1f0ac5f0252c1f10a8d411d30e1d5a5375ad3) while solving merge conflicts, oops.

packages/cli/src/modules/chat-hub/chat-hub-agent.repository.ts
ivov, line 1: We could reduce the indirection from `withTransaction` with an optional arg:

```ts
 async deleteAgent(id: string, entityManager?: EntityManager) {
      const em = entityManager ?? this.manager;
      return await em.delete(ChatHubAgent, { id });
  }
```

autologie (approved):
Looks good from my side, also thank you for measuring. Looks like a great improvement 🚀

packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
```diff
@@ -0,0 +1,44 @@
...
+		// Index intended for faster sessionRepository.getManyByUserId queries
+		await runQuery(
+			`CREATE INDEX IF NOT EXISTS ${escape.indexName('chat_hub_sessions_owner_lastmsg_id')}
+			ON ${sessionsTable}(${ownerIdColumn}, ${lastMessageAtColumn} DESC, ${idColumn})`,
+		);
+
+		// Index intended for faster sessionRepository.getOneByIdAndUserId queries and joins
+		await runQuery(
+			`CREATE INDEX IF NOT EXISTS ${escape.indexName('chat_hub_messages_sessionId')}
+			ON ${messagesTable}(${sessionIdColumn})`,
```
ivov, line 32: Is this correct? Looking at `getManyBySessionId` I'd expect `(sessionId, createdAt ASC, id DESC)`

packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
```diff
@@ -0,0 +1,44 @@
...
+
+		await addNotNull('chat_hub_sessions', 'lastMessageAt');
+
+		// Index intended for faster sessionRepository.getManyByUserId queries
+		await runQuery(
+			`CREATE INDEX IF NOT EXISTS ${escape.indexName('chat_hub_sessions_owner_lastmsg_id')}
+			ON ${sessionsTable}(${ownerIdColumn}, ${lastMessageAtColumn} DESC, ${idColumn})`,
+		);
+
+		// Index intended for faster sessionRepository.getOneByIdAndUserId queries and joins
```
ivov, line 29: Stale comment?

ivov (commented):
Thanks for doing this 🙏🏻 

Mostly looked at the migration only

packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
Cadiac: Here yes, on the other no, so I chose to not use the DSL on both of them for consistency

packages/cli/src/modules/chat-hub/chat-hub-agent.repository.ts
Cadiac: true.

Mind if I do a pass like this on a followup PR, I know we're using the withTransaction with `false` option pattern over at data tables too

packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
Cadiac: This together with what we had ended up being helpful at `getOneByIdAndUserId`, avoiding table scan so I didn't add a more specialized one

packages/@n8n/db/src/migrations/common/1766068346315-AddChatMessageIndices.ts
Cadiac: comment is accurate, this was targeting `getOneByIdAndUserId ` and made it faster

ivov (approved):

PR #23400, opened 2025-12-18
fix(editor): Track changes from iframe and adjust history actions

nikhilkuria:
## Summary
The workflow history actions render changes based on the focus panel. This enables users to view the focus panel and close it if not needed. 

- Created new WorkflowHistoryActions component - Extracted action-related UI (version info display, action toggle button) from WorkflowHistoryContent.vue into a dedicated component for better separation of concerns
- Uses cross-iframe communication for FocusPanel width - Added postMessage-based communication between the iframe (containing FocusPanel) and the parent window to dynamically adjust WorkflowHistoryActions width when FocusPanel is active

https://www.loom.com/share/656d82235546495986ef268da0dfb382

https://linear.app/n8n/issue/ADO-4557/bug-focus-tab-not-showing-up-in-workflow-history

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/WorkflowPreview.vue
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowPreview.vue
@@ -37,6 +37,7 @@ const props = withDefaults(
 
 const emit = defineEmits<{
 	close: [];
+	focusPanelStateChanged: [{ width: number; active: boolean }];
 }>();
 
 const i18n = useI18n();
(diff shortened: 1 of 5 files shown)


packages/frontend/editor-ui/src/app/components/WorkflowPreview.vue
```diff
@@ -178,6 +159,38 @@ const onError = () => {
 	emit('close');
 };
 
+interface IframeMessage {
+	command: string;
+	width?: number;
```
MiloradFilipovic, line 164: The name `IframeMessage` suggests this is a generic `iframe` message but it has focus panel-specific properties. I suggest either renaming to `FocusPanelMessage` or use generic `payload` property to transfer data.

packages/frontend/editor-ui/src/app/components/WorkflowPreview.vue
```diff
@@ -178,6 +159,38 @@ const onError = () => {
 	emit('close');
 };
 
+interface IframeMessage {
```
MiloradFilipovic, line 162: Nit: Any reason why this is an interface and `CommandHandler` is a type? Also, we could move these to the top of the file, they are hard to spot in this place

MiloradFilipovic (commented):
Nice, kudos for extracting it to a separate component. I left a few comments and I generally agree with AI review (using typeguards)

PR #23405, opened 2025-12-18
fix(core): Validate missing `code` param in Code node

ivov:
Noticed this issue in passing - the current typing of node params is best-effort so it's not guaranteed that a Code node task always has a populated `code` param.

https://n8nio.sentry.io/issues/6913452586/?project=4508342481780736


mfsiega (approved):
💯

PR #23413, opened 2025-12-18
fix: Limit access to files based on regex pattern

dlavrenuek:
## Summary

Limit read/write access based on (configurable) file pattern

https://linear.app/n8n/issue/NODE-4106

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/__tests__/file-system-helper-functions.test.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/__tests__/file-system-helper-functions.test.ts
@@ -201,7 +201,7 @@ describe('isFilePathBlocked', () => {
 
 	it('should allow access when pattern matching is disabled', async () => {
 		securityConfig.blockFilePatterns = '';
-		expect(isFilePathBlocked(await resolvePath("'/tmp/.git'"))).toBe(false);
+		expect(isFilePathBlocked(await resolvePath('/tmp/.git'))).toBe(false);
 	});
 
 	describe('when multiple file patterns are configured', () => {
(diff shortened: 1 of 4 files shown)


packages/core/src/execution-engine/node-execution-context/utils/__tests__/file-system-helper-functions.test.ts
```diff
@@ -185,6 +192,48 @@ describe('isFilePathBlocked', () => {
...
+		[
+			'hello',
+			'xhellox',
+			'subpath/hello/',
+			'/there',
+			'/subpath/there',
+			'where',
+			'where-is/it',
+		].forEach((path) => {
+			it(`should block access to ${path}`, async () => {
```
tomi: nit:

```ts
test.each([
  ...
])('should block access to %s', () => {
})
```

would be a bit more readable and idiomatic jest

dlavrenuek: Nice, I was not aware of this syntax, fixed 👍

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -47,6 +47,16 @@ async function resolvePath(path: PathLike): Promise<ResolvedFilePath> {
...
 }
 
+function isFilePatternBlocked(resolvedFilePath: ResolvedFilePath): boolean {
+	const { blockFilePatterns } = Container.get(SecurityConfig);
+
+	return blockFilePatterns
+		.split(';')
+		.map((pattern) => pattern.trim())
+		.filter((pattern) => pattern)
+		.some((pattern) => new RegExp(pattern, 'mi').test(resolvedFilePath));
```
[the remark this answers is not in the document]
dlavrenuek: @tomi I did apply this suggestion with the logic that if the file pattern is broken, it will block all inputs.

On the other hand it might be beneficial not to add error handling here - in this case any api request containing access to read/write file would result in a 500 and I assume a logged error. This means that the a broken regex syntax will be much easier to discover by admins

tomi (approved):
🚀

PR #23417, opened 2025-12-18
feat(ai-builder): Multi-agent workflow builder improvements, best practices and better template usage

mike12345567:
## Summary

This PR brings some benefits found throughout the research carried out for linear ticket: https://linear.app/n8n/issue/AI-1784/tweaks-to-templates-mvp-to-improve-outcomes

There are some key areas of improvement this PR brings.

1. This is comparing with the validation fix in both the current evals, and the template improvements. The validation fix doesn't seem to improve evals that much on current master, although they are a little more consistent
2. Adding template knowledge to the builder/configurator does seem to help, builder makes a bigger difference as it seems to understand better how nodes should be connected, using switch, merge, loop nodes quite a bit better
3. Mermaid improvements, I took all of David's suggestions and implemented them into the Mermaid generation which does seem to have helped, core ideas that helped are the use of conditional nodes in the mermaid to signify if/switch/filter, wrapping agents and their tools/sub-nodes in a subgraph to show them all go together, and adding stickies as comments and subgraphs (previously stickies were ignored)
4. A lot of best practices added which the multi-agent architecture respects a lot more easily than the monolith, especially with the validation fixes
5. This does also include some improvements to the evaluators, specifically around RAG - there was a bit of a fundamental mis-understanding of how AI connections should be made and I've updated the evaluator prompts to improve their understanding of these, so part of the score improvement comes from this

This does significantly improve evaluation results in our broad evaluations, but its important to remember that I've also updated the evaluators a bit in this PR as well. Specifically there was a few key areas that I updated as it was making a lot of mistakes:

False Positive Fixes:                              
- Document Loader is a capability-only node with no main inputs - this is correct, not an error       
- AI sub-nodes are the SOURCE of ai_* connections (Document Loader → Vector Store), not targets      
- Text Splitter → Document Loader [ai_textSplitter] is always valid                                   
- Using embedding models when user requests chat models for embeddings is correct (technically impossible otherwise)                              
  
Scope Clarification:                               
- Node Configuration Evaluator now explicitly ignores connection issues (handled elsewhere)      

### Mermaid changes
I've Also updated the mermaid generation in the following ways:
1. Use subgraphs to group all the sub-nodes of an agent together
2. Use subgraphs to wrap nodes in their stickies
3. Add stickies in for singular nodes as comments
4. Added operations alongside the node type for each node definition
5. Changed how nodes are defined to reduce token usage a bit
6. Conditional nodes (if, filter, switch) use conditional mermaid node definition

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/connections-evaluator.ts
```diff
@@ -52,13 +52,67 @@ Your task is to verify that every connection follows n8n's sourcing rules, suppo
...
+4. Embeddings connects TO Vector Store via **ai_embedding** - provides vectorization capability
+
+**THE FOLLOWING ARE ALL CORRECT - NEVER FLAG AS VIOLATIONS:**
+- Document Loader has NO main connections (no main input, no main output) - THIS IS CORRECT BY DESIGN
+- Document Loader connects TO Vector Store via ai_document - THIS IS THE ONLY WAY TO USE DOCUMENT LOADER
+- Extract From File/PDF/CSV connects directly to Vector Store via main - THIS IS CORRECT
+- Data flows: Extract → Vector Store (main) while Document Loader → Vector Store (ai_document) - BOTH ARE NEEDED, BOTH ARE CORRECT
+- Document Loader appears "disconnected" from the main workflow path - THIS IS CORRECT, IT ONLY USES ai_document
+
+**INVALID VIOLATION EXAMPLES - DO NOT OUTPUT THESE:**
```
mutdmour: The double negative is confusing and might be confusing for the agent as the context grows. Could we frame it in more positive clearer terms. 
i.e. Here are valid examples
Here are violations

packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/functionality-evaluator.ts
```diff
@@ -64,7 +64,98 @@ Evaluate ONLY the functional aspects - whether the workflow achieves the intende
...
+### FORBIDDEN Violations - NEVER Output These
+You must NEVER output violations like:
+- ❌ "User requested gpt-4o-mini but workflow uses text-embedding-3-small"
+- ❌ "User explicitly requested embeddings using 'gpt-4.1-mini' but workflow uses 'text-embedding-3-small'"
+- ❌ "User's explicit specification was not followed" (when user specified a chat model for embeddings)
+- ❌ "The workflow does not use the model specified by the user" (for embedding nodes)
+- ❌ Any violation mentioning that an embedding node should use a chat model
+
+### Why This Rule Exists
+This is like a user asking to "cut wood with a hammer" - using a saw instead is correct, not a violation. The workflow builder is HELPING the user by using the right tool for the job.
```
mutdmour, line 134: 💡

packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/connections-evaluator.ts
```diff
@@ -52,13 +52,67 @@ Your task is to verify that every connection follows n8n's sourcing rules, suppo
 ## Patterns to Validate
 
 ### Retrieval-Augmented Generation (RAG)
-- Data source or metadata prep (HTTP Request, Set, etc.) → Vector Store **via main** when mode is insert/upsert
-- Token Splitter → Document Loader [ai_textSplitter]
-- Document Loader → Vector Store [ai_document]
-- Embeddings → Vector Store [ai_embedding]
+
+**CRITICAL: Document Loader is a CAPABILITY-ONLY sub-node. It NEVER receives main data flow.**
```
mutdmour: I put this whole section out of curiosity into anthropic's prompt optimizer for sonnet 4.5, thinking not enabled

```

## Understanding n8n RAG Architecture

Before validating, you must understand the critical architectural principle: some nodes in n8n are "capability providers" rather than data processors in the main flow.

### Document Loader: A Capability-Only Node

The Document Loader is the most important example:
- Document Loader has NO main input and NO main output
- It NEVER receives data via main connections
- It connects TO Vector Store via the `ai_document` connection type
- It reads data from workflow context (binary files, JSON) based on its configuration
- It is a capability provider, not a data processor

**This means Document Loader will appear "disconnected" from the main data flow - THIS IS CORRECT BY DESIGN.**

### Correct RAG Pipeline Pattern

```
Data Source (Extract From File, HTTP Request, etc.)
       │
       │ [main]
       ▼
Vector Store (insert mode) ◄──[ai_document]── Document Loader ◄──[ai_textSplitter]── Text Splitter
       ▲
       │
       └──[ai_embedding]── Embeddings
```

**How it works:**
1. Data source connects to Vector Store via `main` connection - this triggers the insert operation
2. Document Loader connects TO Vector Store via `ai_document` - provides document processing capability
3. Text Splitter connects TO Document Loader via `ai_textSplitter` - provides chunking capability
4. Embeddings connects TO Vector Store via `ai_embedding` - provides vectorization capability

### Patterns That Are ALWAYS CORRECT (Never Flag These)

- Document Loader has NO main connections (no main input, no main output)
- Document Loader → Vector Store via `ai_document` connection
- Text Splitter → Document Loader via `ai_textSplitter` connection
- Extract From File/PDF/CSV → Vector Store via `main` connection
- Data flows: Extract → Vector Store (main) while Document Loader → Vector Store (ai_document)
- Document Loader appears "disconnected" from the main workflow path
- Vector Store (tool mode) → AI Agent via `ai_tool` connection

### Connection Direction Rules

Memorize these correct directions:
- ✅ Text Splitter → Document Loader [ai_textSplitter] (Text Splitter is SOURCE, Document Loader is TARGET)
- ✅ Document Loader → Vector Store [ai_document] (Document Loader is SOURCE, Vector Store is TARGET)
- ✅ Embeddings → Vector Store [ai_embedding] (Embeddings is SOURCE, Vector Store is TARGET)
- ❌ Document Loader → Text Splitter [any] (NEVER valid)
- ❌ Vector Store → Document Loader [any] (NEVER valid)

### Invalid Violations (DO NOT Output These)

These are examples of INCORRECT analysis - never output violations like these:
- "Document Loader is disconnected from main data flow" - WRONG, this is correct behavior
- "Document Loader is completely disconnected" - WRONG, it connects via ai_document
- "Extract From File bypasses Document Loader" - WRONG, main data SHOULD go directly to Vector Store
- "Document Loader should receive the extracted data" - WRONG, Document Loader reads from workflow context
- "Document Loader should receive extracted data via main connection" - COMPLETELY WRONG
- "Text Splitter → Document Loader is reversed" - WRONG, this is the correct direction
- Any violation about Document Loader needing main connections - ALWAYS WRONG

## Validation Instructions

Before providing your validation report, conduct your analysis in <analysis> tags where you systematically work through the following steps. It's OK for this section to be quite long.

1. **Enumerate all nodes and connections**:
   - List each node in the workflow with format: "Node Name (Node Type)"
   - For each node, list ALL its connections in format: "Source Node → Target Node [connection_type]"
   - Be exhaustive - write down every single connection you can identify

2. **Identify capability-only nodes**:
   - List which nodes are capability-only (Document Loader, Text Splitter, Embeddings)
   - For EACH capability-only node, explicitly state: "Does [Node Name] have any main connections? [Yes/No]"
   - If Yes, this is a potential violation (unless it's a false positive pattern)

3. **Validate connection patterns systematically**:
   For each of these expected patterns, write it out and check if it exists:
   - Expected: "Text Splitter → Document Loader [ai_textSplitter]"
     Found in workflow: [Yes/No - describe what you see]
   - Expected: "Document Loader → Vector Store [ai_document]"
     Found in workflow: [Yes/No - describe what you see]
   - Expected: "Embeddings → Vector Store [ai_embedding]"
     Found in workflow: [Yes/No - describe what you see]
   - Expected: "Data Source → Vector Store [main]" (if Vector Store is in insert/upsert mode)
     Found in workflow: [Yes/No - describe what you see]

4. **Check against false positive patterns**:
   - List the "Patterns That Are ALWAYS CORRECT" from above
   - For each pattern you observe in the workflow, explicitly check: "Is this one of the always-correct patterns? [Yes/No]"
   - If Yes, do NOT flag it as a violation

5. **Compile violations**:
   - Based on your systematic checks above, list any actual violations
   - For each potential violation, verify it's not in the false positive list
   - Provide clear reasoning for why each violation is a problem
```

mutdmour: I think what's interesting here is
1. the use of <analysis> tag as a replacement for thinking tags. Not sure if we have thinking enabled for these evaluators but might be a good idea.
2. It changed the language a bit, write full sentences. 
3. Moved some sections around seemingly because they are more relevant to each other.
4. Removed a lot of the duplicated emojis. 
5. Created clear step-by-step validation. 

We could maybe benefit from some of these changes. Especially the step-by-step validation.

packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/discovery.prompt.ts
```diff
@@ -233,7 +233,8 @@ const CRITICAL_RULES = `- NEVER ask clarifying questions
 - NEVER guess node versions - always use search_nodes to find exact versions
 - ONLY flag connectionChangingParameters if they appear in <input> or <output> expressions
 - If no parameters appear in connection expressions, return empty array []
-- Output ONLY: nodesFound with {{ nodeName, version, reasoning, connectionChangingParameters }}`;
+- Output ONLY: nodesFound with {{ nodeName, version, reasoning, connectionChangingParameters }}
+- When user specifies a model name (e.g., 'gpt-4.1-mini'), use it EXACTLY as written - no corrections or substitutions`;
```
mutdmour: should it really ignore obvious typos? `gtp-4.1` for example?

packages/@n8n/ai-workflow-builder.ee/src/tools/best-practices/data-extraction.ts
```diff
@@ -45,6 +45,21 @@ Leverage AI for unstructured data using nodes like Information Extractor or Summ
...
+Purpose: Looping over a set of items extracted from a data set, for example if pulling a lot of data
+from a Google Sheet or database then looping over the items is required. This node MUST be used
+if the user mentions a large amount of data, it is necessary to batch the data to process all of it.
+
+Pitfalls:
+
+- Misconfigured connections: This node has TWO outputs that work differently from most nodes.
+  CRITICAL: Output 0 (first array element) = "Done" branch - connects to nodes that run AFTER all looping completes.
+  CRITICAL: Output 1 (second array element) = "Loop" branch - connects to nodes that process each batch during the loop.
+  This is COUNTERINTUITIVE - the loop processing is on output 1, NOT output 0.
```
mutdmour: 😅 We need to rethink some stuff

mutdmour (commented):
half way through, will finish tomorrow

packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/connections-evaluator.ts
mike12345567: I actually tried giving positive examples - sadly this doesn't work! Giving it examples of what it shouldn't do produced more accurate results.

packages/@n8n/ai-workflow-builder.ee/src/utils/state-reducers.ts
```diff
@@ -9,33 +9,26 @@ export function appendArrayReducer<T>(current: T[], update: T[] | undefined | nu
...
-	update: NodeConfigurationsMap | undefined | null,
-): NodeConfigurationsMap {
-	if (!update || Object.keys(update).length === 0) {
-		return current;
+	// Build a map of existing templates by name for fast lookup
+	const existingByName = new Map(current.map((wf) => [wf.name, wf]));
+
+	// Add new templates that don't already exist
+	for (const workflow of update) {
+		if (!existingByName.has(workflow.name)) {
```
mutdmour: should we be using `id` as its more unique?

packages/@n8n/ai-workflow-builder.ee/src/tools/utils/mermaid.utils.ts
```diff
@@ -30,210 +34,849 @@ const DEFAULT_MERMAID_OPTIONS: Required<MermaidOptions> = {
...
+			const id = this.nodeIdMap.get(nodeName);
+			if (node && id) {
+				this.lines.push(...this.buildSingleNodeLines(node, id));
+			}
+		}
+
+		this.lines.push('end');
+	}
+
+	private buildAgentNodeConnections(
```
mutdmour, line 684: this new functionality could use better coverage here

packages/@n8n/ai-workflow-builder.ee/src/tools/utils/mermaid.utils.ts
mutdmour, line 1: I guess this file should be renamed to mermaid utils

packages/@n8n/ai-workflow-builder.ee/src/tools/utils/node-configuration.utils.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export function collectNodeConfigurationsFromWorkflows(
+	workflows: WorkflowMetadata[],
+): NodeConfigurationsMap {
+	const configurations: NodeConfigurationsMap = {};
+
+	for (const workflow of workflows) {
+		for (const node of workflow.workflow.nodes) {
+			// Skip sticky notes
+			if (node.type === 'n8n-nodes-base.stickyNote') continue;
```
mutdmour, line 61: could also filter out custom node types or unknown node type to an instance. Maybe that's not important if we never request those nodes

packages/@n8n/ai-workflow-builder.ee/src/tools/utils/node-configuration.utils.ts
```diff
@@ -0,0 +1,146 @@
...
+ * Returns null if the node has no parameters or exceeds size limits
+ */
+export function collectSingleNodeConfiguration(
+	node: NodeForConfiguration,
+): NodeConfigurationEntry | null {
+	const hasParams = Object.keys(node.parameters).length > 0;
+	if (!hasParams) return null;
+
+	const parametersStr = JSON.stringify(node.parameters);
+	if (parametersStr.length > MAX_NODE_EXAMPLE_CHARS) return null;
```
mutdmour, line 26: I wonder if we will lose a lot of agents with long system prompts this way. Could we trim the parameter itself instead?

packages/@n8n/ai-workflow-builder.ee/src/tools/get-node-examples.tool.ts
```diff
@@ -0,0 +1,286 @@
...
+const TOOL_CONFIG: Record<NodeExampleType, { meta: BuilderToolBase; description: string }> = {
+	configuration: {
+		meta: {
+			toolName: 'get_node_configuration_examples',
+			displayTitle: 'Getting node configuration examples',
+		},
+		description: `Get real-world parameter configuration examples for a specific node type from community templates.
+
+Use this tool when you need reference examples for configuring node parameters:
+- Before configuring nodes with complex parameters
```
mutdmour: what's the agent's definition for complex? what nodes does this end up getting used for? would this be helpful for nodes like the set node where there might be no real pattern?

packages/@n8n/ai-workflow-builder.ee/src/tools/best-practices/data-extraction.ts
mike12345567: Moved this into the split in batches information for the builder subgraph - it already had info about this but wasn't so clear on the outputs!

packages/@n8n/ai-workflow-builder.ee/src/tools/get-node-examples.tool.ts
mike12345567: True, I've removed this line as it might be causing this tool to be used less.

packages/@n8n/ai-workflow-builder.ee/src/tools/utils/node-configuration.utils.ts
mike12345567: My fear with this is that we will be returning invalid node configurations - theres a few ways I considered trying to achieve this:
1. Trimming any parameters that are over a certain limit - issue with this is we don't know which parameters we should preserve, and which we shouldn't, trimming leaves the parameters in an invalid state (say stringified JSON, or a prompt, it'll be incomplete). I'd say invalid parameters are more risky than no parameters at all.
2. Remove parameters that are over a certain limit - better because it avoids invalid parameters, but with this approach it could leave the node configuration invalid, its now missing quite a bit of information.

In practice this limit is actually not that low, its 5000 tokens (avg) which very few nodes hit in my testing. This is simply to remove the silly node configurations, like putting base64 images into a set node, we just don't want examples like this included at all.

packages/@n8n/ai-workflow-builder.ee/src/tools/utils/node-configuration.utils.ts
mike12345567: Yeah, I don't believe its possible for the builder/configuration subgraph to be aware of these since they won't be in the node parameter list. Something to test for sure though, bigger question would be what happens in the discovery subgraph if it learns about nodes which don't exist in this instance - but will do that as a separate issue (as that work is already merged).

packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/connections-evaluator.ts
mike12345567: Added some of this - changed it a little bit/moved it around but yeah this is some good improvements!

packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/connections-evaluator.ts
```diff
@@ -22,6 +22,37 @@ export type ConnectionsResult = z.infer<typeof connectionsResultSchema>;
 const systemPrompt = `You are an expert n8n workflow evaluator focusing specifically on NODE CONNECTIONS and DATA FLOW.
 Your task is to verify that every connection follows n8n's sourcing rules, supports the requested behaviour, and respects hybrid AI patterns.
 
+## Validation Instructions
```
mutdmour, line 25: nice, we should have a separate task where we put every prompt through the optimizer

mutdmour (approved):
LGTM

packages/@n8n/ai-workflow-builder.ee/evaluations/chains/evaluators/connections-evaluator.ts
mike12345567: Yeah, we should perhaps script this - maybe we should have a CI job to recommend prompt optimisations on PR after we do a pass across everything as well?

PR #23428, opened 2025-12-18
fix(core): Fix verified community packages reinstall

ivov:
With verified community packages enabled, during reinstall of a community package, even a previously successfully installed verified package gets blocked because the reinstall feature doesn't pass the checksum. Sets serialize as `{}` and the actual error gets suppressed:

```
2025-12-18T19:42:52.544Z | info  | Attempting to reinstall missing packages {"missingPackages":{}}
2025-12-18T19:42:52.544Z | error | n8n was unable to install the missing packages.
```

This PR: 

- verifies integrity when reinstalling missing verified packages
- surfaces the install error
- fixes the missing packages log
- fixes a side issue where if any individual community package install fails, the remaining missing package installs were not attempted


RomanDavydchuk (approved):
LGTM and works as expected

PR #23433, opened 2025-12-18
feat(Form Node): Dynamic attributes in form editor

DawidMyslak:
## Summary

Introducing dynamic attributes in the Form Node.

New config options were added to `fixedCollection` type:
```ts
{
  requiredOnly?: boolean;
  requiredOnlyButtonText?: string;
  requiredOnlyAutoShow?: boolean;
}
```
When `requiredOnly` is set to `true`, only required items will be initially displayed.

Recording: **[Loom](https://www.loom.com/share/8896843af45748aaa6388ca1397c5f1f)**

https://linear.app/n8n/issue/NODE-4059/iteration-on-form-label-split

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Form/common.descriptions.ts
```diff
@@ -1,7 +1,59 @@
-import type { INodeProperties } from 'n8n-workflow';
+import type { INodeProperties, INodePropertyCollection, INodePropertyOptions } from 'n8n-workflow';
```
DawidMyslak, line 1: Diff here seems to be very complex, in fact it's just extracting a few parts to own variables. Way more readable here: https://github.dev/n8n-io/n8n/pull/23433/

packages/frontend/editor-ui/src/features/ndv/parameters/components/FixedCollectionParameter.vue
```diff
@@ -101,6 +105,195 @@ const sortable = computed(() => {
...
+
+const initializeAddedOptionalValues = () => {
+	if (!requiredOnly.value) return;
+	if (!isINodePropertyCollectionList(props.parameter.options)) return;
+
+	addedOptionalValues.value.clear();
+
+	for (const property of props.parameter.options) {
+		const propertyPath = `${props.path}.${property.name}`;
+		const propertyValues = get(props.nodeValues, propertyPath) as
```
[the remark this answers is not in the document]
DawidMyslak: Since `get` can return anything, I can't avoid casting.

packages/workflow/src/interfaces.ts
```diff
@@ -1480,6 +1480,9 @@ export interface INodePropertyTypeOptions {
 	assignment?: AssignmentTypeOptions;
 	minRequiredFields?: number; // Supported by: fixedCollection
 	maxAllowedFields?: number; // Supported by: fixedCollection
+	requiredOnly?: boolean; // Supported by: fixedCollection - hide optional values by default
+	requiredOnlyButtonText?: string; // Supported by: fixedCollection with requiredOnly set to true
+	requiredOnlyAutoShow?: boolean; // Supported by: fixedCollection with requiredOnly set to true
```
elsmr: Not sure about the naming here, `requiredOnly` is pretty clear, but the other 2 I'm not sure I would understand without context.

Maybe something like:

```ts
hideOptionalFields?: boolean;
addOptionalFieldLabel?: string;
showEvenWhenOptional?: boolean;
```

packages/frontend/editor-ui/src/features/ndv/parameters/components/FixedCollectionParameter.vue
```diff
@@ -101,6 +105,192 @@ const sortable = computed(() => {
...
+		return locale.baseText('fixedCollectionParameter.addProperty');
+	}
+	return locale.nodeText(activeNode.value?.type).requiredOnlyButtonText(props.parameter);
+});
+
+const getOptionalValuesKey = (propertyName: string, index?: number): string => {
+	return index !== undefined ? `${propertyName}-${index}` : propertyName;
+};
+
+const hasNonDefaultValue = (
```
elsmr, line 124: In node-helpers.ts:657 `getNodeParameters` has some similar logic to detect default parameters, maybe we can reuse it?

elsmr (commented):
Nice one, works great and a clean API! 🎉  just 2 small comments/questions

packages/workflow/src/interfaces.ts
DawidMyslak: Yeah, took me some time to pick any names 😆. I wanted to start with `requiredOnly*` to clearly state they are all linked to a shared logic, but maybe it was unnecessary. I like your proposal.

One thing I've seen that we already have is a pattern around `multipleValues` and `multipleValueButtonText`:
```
multipleValues?: boolean; // Supported by: <All>
multipleValueButtonText?: string; // Supported when "multipleValues" set to true
```

So maybe:
```
hideOptionalFields?: boolean;
optionalFieldButtonText?: string;
showEvenWhenOptional?: boolean;
```

But then, I'm not sure if "Fields" is actually a good term 🤔. They make totally sense for the Form node, but here we try to build a commonly used API, and maybe we need to go for something more generic? Maybe "hideOptionalProperties"?

This is for sure confusing 🤣. All those "fields" are listed inside `values: [...]` array. So not sure anymore what it is. A field, property, attribute or a value 😆 ?

packages/frontend/editor-ui/src/features/ndv/parameters/components/FixedCollectionParameter.vue
DawidMyslak: Checked it, `getNodeParameters` function alone has 350 lines of code. Honestly it's hard to tell which part I could extract to match what `hasNonDefaultValue` does exactly.

Thanks to your suggestion I've spotted one missing bit around `hasNonDefaultValue` - I never added proper default check against objects which we also support.

**Addressed in: https://github.com/n8n-io/n8n/pull/23433/commits/9b71f0d693c6db07141ea837e0994579766bbc64**

packages/workflow/src/interfaces.ts
elsmr: Yeah the naming is not very consistent, I also found these in typeOptions:

```ts
minRequiredFields?: number; // Supported by: fixedCollection
maxAllowedFields?: number; // Supported by: fixedCollection
```

I'm also fine with `properties` as its consistent with the naming in `INodeType`. But it seems like we can't be consistent with everything 😅 I think I like this:

```ts
hideOptionalFields?: boolean;
optionalFieldButtonText?: string; // maybe this one could have "add" somewhere to convey what the button does
showEvenWhenOptional?: boolean;
```

DawidMyslak: I like it too!

Addressed in https://github.com/n8n-io/n8n/pull/23433/commits/54d4fdf61547f05969981c9fc9dccdbcaac12c3f

elsmr (approved):

PR #23437, opened 2025-12-19
fix(editor): Remove running tool messages when AI builder task fails

OlegIvaniv:
## Summary
Fixes a bug in the AI Workflow Builder where the thinking container continued showing loading spinners after aborting a task or when an error occurred. Running tool messages are now filtered out for all error scenarios (abort and service errors), so only completed tools remain visible.

## Changes
- Filter out tool messages with `status: 'running'` in `handleServiceError` before adding error/abort messages
- The filtering now applies to both AbortError (user-initiated stop) and other service errors (network failures, etc.)
- Added two test cases to verify running tools are removed for both abort and service error scenarios

## Review / Merge checklist

- [ ] Tests included. 


burivuhster (approved):
Looks good!

PR #23438, opened 2025-12-19
chore(core): Use interfaces instead of session / agent entities to solve partial deep type issues

Cadiac:
## Summary

Our use of `Partial<ChetHubSession>` is perhaps causing these issues - get rid of use of Partial's on chat hub entities to hopefully solve the remaining

> Type instantiation is excessively deep and possibly infinite.

issues that randomly occur.

Made both ChatHubAgent and ChatHubSession provide interfaces that we should just be able to use in the creates / updates. `QueryDeepPartialEntity` lead to other build issues and couldn't be used here.

Also suppressed some info level logs that probably shouldn't be info.

https://n8nio.slack.com/archives/C03MZF137FV/p1766124829982929?thread_ts=1763993338.525809&cid=C03MZF137FV

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/api-types/src/chat-hub.ts
+++ b/packages/@n8n/api-types/src/chat-hub.ts
@@ -459,7 +459,7 @@ export class ChatHubUpdateAgentRequest extends Z.class({
 	icon: agentIconOrEmojiSchema.optional(),
 	systemPrompt: z.string().min(1).optional(),
 	credentialId: z.string().optional(),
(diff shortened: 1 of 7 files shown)


autologie (approved):
I encountered this error yesterday again, looks like it happens whenever we touch chat hub entities. Thank you and hope this solves the issue finally.

PR #23448, opened 2025-12-19
fix(core): Ensure runner does not terminate pipe reader prematurely

ivov:
## Summary

We have a subprocess running Python code and writing to a pipe, and a parent process with a pipe reader background thread for pulling data from the pipe. We are currently force-closing the pipe reader's read connection after ~3s, which is based on a napkin estimation about typical payload size and throughput. But [Sentry shows](https://n8nio.sentry.io/issues/6997406518/?project=4508342481780736) that large results are blowing past that, so we're prematurely closing even though the subprocess exited 0 and the data is sitting in the pipe buffer, i.e the overly tight timeout is discarding good results.

Hence this PR gives the pipe reader a fair chance to finish reading by reusing `task_timeout` (if the subprocess finished within it, reading the result shouldn't take longer) and if that's not enough only then raise an appropriate error. 

Technically this means we're redefining "task timeout" to mean task processing as a whole (execute + read). We could also track elapsed time and give the read only the time that was not consumed by execution, but looking at the number of Sentry reports I lean towards redefining timeout instead. Let me know if you disagree. I also considered calculating a dynamic timeout based on each result size, but it struck me as over-engineering.

https://linear.app/n8n/issue/CAT-1710

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/task-runner-python/src/config/task_runner_config.py
+++ b/packages/@n8n/task-runner-python/src/config/task_runner_config.py
@@ -22,9 +22,6 @@ from src.constants import (
     ENV_AUTO_SHUTDOWN_TIMEOUT,
     ENV_GRACEFUL_SHUTDOWN_TIMEOUT,
     PIPE_MSG_MAX_SIZE,
-    TYPICAL_PAYLOAD_RATIO,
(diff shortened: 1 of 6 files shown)


tomi (approved):
🚀

PR #23449, opened 2025-12-19
refactor(core): Remove 'beginTransaction' parameter from withTransaction

Cadiac:
## Summary

Followup to https://github.com/n8n-io/n8n/pull/23397#discussion_r2634258652

The optional `beginTransaction` argument on withTransaction brought unnecessary noise to code that didn't want automatic transactions to be started. Prefer using nullish coalescing operators instead, as that's all the function did in `beginTransaction=false`mode.

The helper is still somewhat handy for actually wrapping blocks of code with "start a transaction if there isn't already one" logic, so not getting rid of it fully.

## Review / Merge checklist

- [ ] Tests included. 


ivov (approved):
Thanks for doing this 🙏🏻

PR #23451, opened 2025-12-19
fix(form): prevent CORS error on form execution status polling

### LINKED ISSUE #23262: Form trigger always looping
### Bug Description

When I click on the next form, an infinite loop appears, and it does not proceed to the next form.

before n8n version 2.0, it can load properly.

### To Reproduce

Just make node "On form submission" and make switch node and node "next from page"

Image

Image

### Expected behavior

Can load to the next form like before version 2.0

### Debug Info

# Debug info

## core

- n8nVersion: 2.1.0
- nodeJsVersion: 22.21.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: b40e5767-9497-41ce-b2ba-9283c4a530c1

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/143.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2025-12-16T07:08:23.471Z

### Operating System

Windows 11

### n8n Version

2.1.0

### Node.js Version

22.21.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

barathraj048:
## Summary

Form pages are rendered inside a sandboxed iframe. The current CSP includes
`sandbox` but omits `allow-same-origin`, which forces the iframe origin to
become `null`.

As a result, polling requests to
`/form-waiting/:id/n8n-execution-status` are sent with `Origin: null` and
are blocked by the browser (CORS), even though the endpoint returns `200 OK`.

This PR adds `allow-same-origin` to the sandbox CSP used for form pages,
restoring same-origin behavior and preventing the CORS error.

### How to test
1. Start n8n locally
2. Create a workflow with a Form Trigger
3. Open and submit the form
4. Verify no CORS error appears in DevTools
5. Verify form progresses correctly after submission

## Related issues

Fixes #23262

https://github.com/user-attachments/assets/6ca667eb-1ced-4bea-900d-92aaeb5f4146

## Review / Merge checklist

- [ ] Docs updated (not required – internal CSP fix)
- [ ] Tests included (manual verification; existing E2E coverage applies)
- [ ] PR labeled with `release/backport` (not required)

### DIFF
--- a/packages/core/src/html-sandbox.ts
+++ b/packages/core/src/html-sandbox.ts
@@ -9,7 +9,7 @@ export const isWebhookHtmlSandboxingDisabled = () => {
  * Returns the CSP header value that sandboxes the HTML page into a separate origin.
  */
 export const getWebhookSandboxCSP = (): string => {
-	return 'sandbox allow-downloads allow-forms allow-modals allow-orientation-lock allow-pointer-lock allow-popups allow-presentation allow-scripts allow-top-navigation allow-top-navigation-by-user-activation allow-top-navigation-to-custom-protocols';
+	return 'sandbox allow-same-origin allow-downloads allow-forms allow-modals allow-orientation-lock allow-pointer-lock allow-popups allow-presentation allow-scripts allow-top-navigation allow-top-navigation-by-user-activation allow-top-navigation-to-custom-protocols';
 };
 
 /**
--- a/package.json
+++ b/package.json
@@ -54,12 +54,12 @@
     "worker": "./packages/cli/bin/n8n worker"
   },
   "devDependencies": {
+    "@babel/preset-env": "^7.26.0",
     "@biomejs/biome": "^1.9.0",
     "@n8n/eslint-config": "workspace:*",
     "@types/jest": "^29.5.3",
     "@types/node": "*",
     "@types/supertest": "^6.0.3",
-    "@babel/preset-env": "^7.26.0",
     "babel-plugin-transform-import-meta": "^2.3.2",
     "bundlemon": "^3.1.0",
     "cross-env": "^7.0.3",
@@ -83,7 +83,7 @@
     "tsc-alias": "^1.8.10",
     "tsc-watch": "^6.2.0",
     "turbo": "2.5.4",
-    "typescript": "*",
+    "typescript": "catalog:",
     "zx": "^8.1.4"
   },
   "pnpm": {
--- a/packages/nodes-base/nodes/Form/utils/utils.ts
+++ b/packages/nodes-base/nodes/Form/utils/utils.ts
@@ -567,7 +567,6 @@ export async function formWebhook(
 	};
 	const res = context.getResponseObject();
 	const req = context.getRequestObject();
-
 	try {
 		if (options.ignoreBots && isbot(req.headers['user-agent'])) {
 			throw new WebhookAuthorizationError(403);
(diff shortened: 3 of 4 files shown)


barathraj048:
Hi team, thanks for the response and for adding this to the tracker.

This change addresses a CSP configuration issue that caused form polling
requests to be blocked due to the iframe origin becoming `null`.
The fix is limited to adjusting the sandbox CSP and does not introduce
new logic paths.

At the moment, I verified the fix manually by confirming that the
`/form-waiting/:id/n8n-execution-status` polling no longer triggers
a CORS error and the form progresses correctly.

If you’d prefer a specific test to be added for this case, I’d be happy
to work on one — please let me know the recommended approach.

Thanks again!

packages/core/src/html-sandbox.ts
```diff
@@ -9,7 +9,7 @@ export const isWebhookHtmlSandboxingDisabled = () => {
  * Returns the CSP header value that sandboxes the HTML page into a separate origin.
  */
 export const getWebhookSandboxCSP = (): string => {
-	return 'sandbox allow-downloads allow-forms allow-modals allow-orientation-lock allow-pointer-lock allow-popups allow-presentation allow-scripts allow-top-navigation allow-top-navigation-by-user-activation allow-top-navigation-to-custom-protocols';
+	return 'sandbox allow-same-origin allow-downloads allow-forms allow-modals allow-orientation-lock allow-pointer-lock allow-popups allow-presentation allow-scripts allow-top-navigation allow-top-navigation-by-user-activation allow-top-navigation-to-custom-protocols';
```
dlavrenuek, line 12: Adding `allow-same-origin` renders the sandbox mode useless because it allows access to cookies, session and local storage. Specifically this is prevented by [sandbox mode](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/sandbox), so even if there is a stored XSS attack - an attacker can not steal credentials. Therefore we must not add `allow-same-origin` here

dlavrenuek (commented):
Hi @barathraj048, thank you for your contribution. Unfortunately we can not do the fix this way, which I will comment below. With the sandbox mode for forms introduced for hardened security the implementation for basic auth and form submission is a bit more complex and will he handled in https://github.com/n8n-io/n8n/pull/23795. I will close this PR as not planned

packages/core/src/html-sandbox.ts
barathraj048: Thanks for the clarification @dlavrenuek — that makes sense.

Adding allow-same-origin indeed breaks the isolation guarantee, and I understand why the forms sandbox can’t relax that flag without exposing cookies/localStorage to a potential XSS vector.

The original intention of this PR was strictly to address the polling issue for
/form-waiting/:id/n8n-execution-status, not to weaken the sandbox.
Since the fix needs deeper changes to how form execution and polling work inside a sandboxed iframe, I’ll defer to the structured approach planned in #23795.

If there’s any specific part of that work where I can contribute (testing, patching the polling mechanism, adjusting CSP for controlled subresources, etc.), I’m happy to help.

Thanks for the guidance and the detailed security context.

dlavrenuek: @barathraj048 it's always helpful if you can test the provided fix (#23795) for your scenario, to confirm that it works correctly 🙌

dlavrenuek: Also no worries regarding your proposal. The webhook logic and checks are spread across a big part of the code base and it is pretty challenging to follow the red line especially with continuous changes in the code.

PR #23468, opened 2025-12-19
feat: Support partial executions for HITL tool nodes

yehorkardash:
## Summary

This PR implements partial execution of HITL tools. To achieve that the following changes were done:
- Backend
  - Updated ToolExecutor node to support toolkits and support request-response format for HITL
  - Updated execution engine logic to work with ToolExecutor. ToolExecutor is added dynamically to the workflow and it isn't persisted anywhere, so after send-and-wait webhook executes and tries to rerun the workflow it can't access the node parameters. This issue was fixed by adding null checks in several places.
- Frontend
 - Added metadata to form input fields that is used to identify to which tool each parameter belongs. Previously  `FromAiParametersModal` used input names to identify fields, but that doesn't work well for nested fields.
 - Updated AgentRequest interface to be a map of "Node name" -> "tool parameters"

https://github.com/user-attachments/assets/c7027385-c0c2-4b15-b52e-2fb5add6d30b

https://linear.app/n8n/issue/NODE-3886/implement-partial-execution-of-hitl-tool-node

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/nodes-langchain/nodes/ToolExecutor/ToolExecutor.node.ts
```diff
@@ -66,17 +107,56 @@ export class ToolExecutor implements INodeType {
...
+
+								// prepare request for execution engine to execute the HITL node
+								const engineRequest: EngineRequest<RequestResponseMetadata>['actions'] = [
+									{
+										actionType: 'ExecutionNodeAction' as const,
+										nodeName: node,
+										input: {
+											tool: toolName,
+											toolParameters: JSON.stringify(toolInput),
+											...hitlInput,
```
elsmr, line 158: I noticed this in this UI, do we need to stringify this? also notice the empty hitlParameters object

Image

elsmr (commented):
I noticed a bug with some of the Notion MCP tools, maybe a case in their tool schemas we don't handle:

https://github.com/user-attachments/assets/1e7e4e39-4448-4d15-82c0-d01c7f1f893b

packages/@n8n/nodes-langchain/nodes/ToolExecutor/ToolExecutor.node.ts
yehorkardash: @elsmr, tested Notion MCP and verified it. It's using complex properties like `array` or `oneOf` which current implementation does not support. However, I'd say this is out of scope of this task.

I also found a few additional bugs that I fixed in the latest commit:
- bug, when an error from mcp was not shown in the modal.
- bug, when code tool with schema would fail in HITL partial execution

elsmr (approved):

PR #23491, opened 2025-12-20
fix(n8n Form Trigger Node): Add support for fieldName in JSON config

### LINKED ISSUE #23462: Form field propety `fieldName` doesn't exist but is described in the docs
### Bug Description

I'm trying to define a schema using JSON.

In the docs (https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.form/#defining-the-form-using-json) there is `fieldName` property description: _"fieldName: The field name that is used in the output of the Form node and to reference the field in expressions."_ that doesn't seem to be a thing.

### To Reproduce

1. Create form node
2. Switch to json configuration
3. Try adding `fieldName` property into field object

### Expected behavior

Setting `fieldName` allows to rename form output property from for example "Выберите модели" into "selectedModels" (so i don't have "Выберите модели" as key in the form output)

### Debug Info

# Debug info

## core

- n8nVersion: 2.0.3
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

### Operating System

Windows 11

### n8n Version

2.0.3

### Node.js Version

v25.2.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

saurabhssonkar:
This PR adds support for the `fieldName` property in the Form node when defining fields using JSON configuration.

The documentation already describes `fieldName` as a way to control the output key of a form field, but the current implementation does not accept this property and marks the schema as invalid.

This change updates the code to properly recognize and handle `fieldName`, allowing users to define a stable output key independent of the visible field label.

## What was changed

- Added support for the `fieldName` property in the Form node JSON configuration
- Ensures form output uses `fieldName` when provided

## Related Issue

Fixes #23462

## Testing

- Local tests pass
- `pnpm test --filter @n8n/codemirror-lang-sql`

## Notes

- No breaking changes introduced
- Existing form configurations continue to work as before

image

image

image


ShireenMissi (approved):
I tested locally and the change works as expected, I also created an issue to update docs with the other new keys added

Thanks for your contribution!

PR #23500, opened 2025-12-20
ci: Codecov bundle analysis

shortstacked:
## Review / Merge checklist

- [ ] Tests included. 


codecov.yml
```diff
@@ -2,6 +2,16 @@ codecov:
   max_report_age: off
   require_ci_to_pass: true
 
+# Bundle Analysis - tracks frontend bundle size changes
+# Docs: https://docs.codecov.com/docs/javascript-bundle-analysis
+bundle_analysis:
+  status: warning # Will show warning status on PR if threshold exceeded
+  warning_threshold: 1% # ~330KB for our 33MB bundle
```
ivov: This comment will go stale as bundle size changes

codecov.yml
```diff
@@ -2,6 +2,16 @@ codecov:
   max_report_age: off
   require_ci_to_pass: true
 
+# Bundle Analysis - tracks frontend bundle size changes
+# Docs: https://docs.codecov.com/docs/javascript-bundle-analysis
+bundle_analysis:
+  status: warning # Will show warning status on PR if threshold exceeded
```
ivov:
```suggestion
  status: warning
```

codecov.yml
```diff
@@ -2,6 +2,16 @@ codecov:
...
+  warning_threshold: 1% # ~330KB for our 33MB bundle
+
+comment:
+  require_bundle_changes: bundle_increase # Only comment when bundle size increases
+  bundle_change_threshold: 100Kb # Report any increase over 100KB
+
```
ivov, line 12: These comments don't add value

packages/frontend/editor-ui/vite.config.mts
```diff
@@ -183,6 +184,17 @@ const plugins: UserConfig['plugins'] = [
 				}),
 			]
 		: []),
+	// Codecov bundle analysis - runs in CI when CODECOV_TOKEN is set
```
ivov:
```suggestion

```

turbo.json
```diff
@@ -1,5 +1,15 @@
...
+	"globalEnv": [
+		"CI",
+		"COVERAGE_ENABLED",
+		"BUILD_WITH_COVERAGE",
+		"CODECOV_TOKEN",
+		"GITHUB_ACTIONS",
+		"GITHUB_REPOSITORY",
+		"GITHUB_SHA",
+		"GITHUB_HEAD_REF",
+		"GITHUB_REF"
```
ivov: Should we be using [passthrough envs](https://turborepo.com/docs/crafting-your-repository/using-environment-variables#passthrough-variables) for these? Else how do we make sure that e.g. every SHA doesn't bust the cache?

.bundlemonrc.json
ivov, line 1: > So this is the full bundle size. Bundlemon shows the estimated gzip size of just the css/js files. This is the size of the unzipped bundle including images etc.

It'd be nice if the comment reported compressed size instead as that's the standard for bundle monitoring. Nothing major though.

turbo.json
shortstacked: Thanks for that!

.bundlemonrc.json
shortstacked: It's not available out of the box unfortunately. Although something we could look at getting in the future with a quick script etc. So we have to weight up the gzip size view vs the historical view trends we get.

ivov (approved):

PR #23524, opened 2025-12-22
fix(core): Fix access control checks for displaying workflow and execution links on Chat hub

Cadiac:
## Summary

Fix the logic for displaying workflow and execution links on Chat hub to display those for member users instead of just admins on workflow agents. In order to do this pass workflow scopes from backend to the FE on /models request, and use that data instead of global `workflow:read` permission checks that only admins pass.

https://linear.app/n8n/issue/CHA-92/correct-check-on-open-workflow-button-visibility

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/api-types/src/chat-hub.ts
+++ b/packages/@n8n/api-types/src/chat-hub.ts
@@ -1,3 +1,4 @@
+import type { Scope } from '@n8n/permissions';
 import {
 	type StructuredChunk,
 	type JINA_AI_TOOL_NODE_TYPE,
@@ -231,6 +232,7 @@ export interface ChatModelMetadataDto {
 		functionCalling: boolean;
 	};
 	available: boolean;
+	scopes?: Scope[];
 }
 
 export interface ChatModelDto {
--- a/packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatAgentCard.vue
+++ b/packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatAgentCard.vue
@@ -7,7 +7,6 @@ import { N8nActionDropdown, N8nIconButton, N8nText } from '@n8n/design-system';
 import type { ActionDropdownItem } from '@n8n/design-system/types';
 import { useI18n } from '@n8n/i18n';
 import { RouterLink } from 'vue-router';
-import { hasPermission } from '@/app/utils/rbac/permissions';
 
 const { agent } = defineProps<{
 	agent: ChatModelDto;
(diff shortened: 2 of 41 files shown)


packages/@n8n/api-types/src/chat-hub.ts
```diff
@@ -231,6 +232,7 @@ export interface ChatModelMetadataDto {
 		functionCalling: boolean;
 	};
 	available: boolean;
+	scopes?: Scope[];
```
Cadiac, line 235: we could also put this under `ChatHubN8nModel`, but felt like having them here optionally is nicer since we don't need to pass scopes around at every place where we use the `ChatHubN8nModel` type

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessageActions.vue
```diff
@@ -31,7 +32,7 @@ const currentAlternativeIndex = computed(() => {
 });
 
 const showExecutionUrl = computed(() => {
-	return hasPermission(['rbac'], { rbac: { scope: 'workflow:read' } });
+	return model?.model.provider === 'n8n' && model.metadata.scopes?.includes('workflow:read');
```
autologie: Considering switching models in conversation, model should be taken from message (I think `unflattenModel` helper function can be used)

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatAgentCard.vue
```diff
@@ -31,7 +30,7 @@ const menuItems = computed<Array<ActionDropdownItem<MenuAction>>>(() => {
 const canEdit = computed(
 	() =>
 		agent.model.provider === 'custom-agent' ||
-		hasPermission(['rbac'], { rbac: { scope: ['workflow:read'] } }),
+		(agent.model.provider === 'n8n' && agent.metadata.scopes?.includes('workflow:update')),
```
autologie: The icon on the button is edit icon, but users can read access may also open the workflow

Cadiac: oh, I'll make this check for read access then

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessageActions.vue
Cadiac: it won't have the scopes data available, I think..

autologie: Hm right, maybe just include `message.workflowId === model.model.workflowId &&` in the condition?

autologie (approved):

PR #23525, opened 2025-12-22
feat(core): Do not pull jobs if the worker is unhealthy

mfsiega:
## Summary

Stop pulling jobs if the worker is unhealthy.

The worker checks its health on an interval, and if it detects that it's unhealthy, it pauses pulling from the queue until it's healthy again.

Disabled by default.

As a next step, we could consider trying to automatically recover as well.

https://linear.app/n8n/issue/CAT-2002/stop-pulling-jobs-if-the-worker-is-unhealthy

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+		} catch (error) {
+			if (error instanceof CircuitBreakerOpen) {
+				this.logger.warn('Health check circuit breaker is open, worker is unhealthy');
+			} else {
+				this.logger.error('Health check failed', { error });
+			}
+
+			// Pause queue if not already paused
+			if (!this.isQueuePaused) {
+				await this.pauseQueue();
```
mfsiega, line 113: Note: if we want to do something like auto-restart the container, we could rather put a `process.exit(1)` either immediately or on a timer.

packages/@n8n/config/src/configs/scaling-mode.config.ts
```diff
@@ -17,6 +17,15 @@ class HealthConfig {
 	/** IP address for worker server to listen on. */
 	@Env('N8N_WORKER_SERVER_ADDRESS')
 	address: string = '::';
+
+	/**
+	 * How often (in milliseconds) to check worker health and pause/resume queue accordingly.
+	 * Worker will automatically pause queue processing if database or Redis connectivity is lost.
+	 * Set to 0 to disable automatic health monitoring.
+	 * @default 0 (disabled)
```
ivov, line 25: If we'll use this for docs, a user might misinterpret "pause/resume queue".

Maybe: "How often (in milliseconds) to check worker readiness, i.e. connectivity to Postgres and Redis. On three failed readiness checks, the worker will stop pulling jobs from the queue until it becomes ready again."

packages/@n8n/config/src/configs/scaling-mode.config.ts
```diff
@@ -17,6 +17,15 @@ class HealthConfig {
...
+	 */
+	@Env('QUEUE_HEALTH_CHECK_INTERVAL')
+	checkInterval: number = 0;
```
ivov, line 28:
```suggestion
	@Env('WORKER_HEALTH_CHECK_INTERVAL')
	workerHealthCheckInterval: number = 0;
```

packages/cli/src/commands/worker.ts
```diff
@@ -165,6 +165,10 @@ export class Worker extends BaseCommand<z.infer<typeof flagsSchema>> {
 		await this.scalingService.setupQueue();
 
 		this.scalingService.setupWorker(this.concurrency);
+
+		// Start health monitoring after queue and worker are set up
```
ivov, line 169: Nit - Here and elsewhere, can we remove superfluous comments 🙏🏻 

```suggestion

```

packages/cli/src/scaling/scaling.service.ts
```diff
@@ -140,20 +140,55 @@ export class ScalingService {
 		else if (instanceType === 'worker') await this.stopWorker();
 	}
 
-	private async pauseQueue() {
-		await this.queue.pause(true, true); // no more jobs will be enqueued or picked up
-		this.logger.debug('Paused queue');
+	/**
+	 * Pause queue processing locally (this worker only).
+	 * Other workers continue pulling jobs normally.
+	 * Already running jobs will continue to completion.
```
ivov, line 146: Bull docs say:

> If `doNotWaitActive` is true, `pause` will not wait for any active jobs to finish before resolving. Otherwise, pause will wait for active jobs to finish. See `Queue#whenCurrentJobsFinished` for more information.

So we aren't actually waiting for completion?

packages/cli/src/scaling/scaling.service.ts
```diff
@@ -140,20 +140,55 @@ export class ScalingService {
...
+	 */
+	async pauseQueueLocal() {
+		this.assertQueue();
+		await this.queue.pause(true, true); // local=true, doNotWaitActive=true
+		this.logger.debug('Paused queue locally');
```
ivov, line 151:
```suggestion
		this.logger.warn('Stopped pulling jobs from the queue');
```

packages/cli/src/scaling/scaling.service.ts
```diff
@@ -140,20 +140,55 @@ export class ScalingService {
...
+		this.logger.debug('Paused queue globally');
+	}
+
+	/**
+	 * Resume queue processing locally (this worker only).
+	 */
+	async resumeQueueLocal() {
+		this.assertQueue();
+		await this.queue.resume(true);
+		this.logger.debug('Resumed queue locally');
```
ivov, line 171:
```suggestion
		this.logger.info('Resumed pulling jobs from queue');
```

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+ * This prevents unhealthy workers from pulling jobs they cannot successfully process.
+ *
+ * QUEUE_HEALTH_CHECK_ACTIVE must be true and QUEUE_HEALTH_CHECK_INTERVAL must be >0 to
+ * enable this.
+ */
+@Service()
+export class WorkerHealthMonitor {
+	private healthCheckInterval: NodeJS.Timeout | undefined;
+
+	private isQueuePaused = false;
```
ivov, line 29: References to "pausing the queue" make me do a double take. Poor naming in Bull seeping in to us. Can we flip to `isPullingJobs`? 🙏🏻

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+		private readonly instanceSettings: InstanceSettings,
+	) {
+		assert(this.instanceSettings.instanceType === 'worker');
+
+		this.logger = this.logger.scoped('scaling');
+
+		this.checkIntervalMs = this.globalConfig.queue.health.checkInterval;
+
+		// Circuit breaker prevents flapping between healthy/unhealthy states
+		this.circuitBreaker = new CircuitBreaker({
```
ivov, line 51: Nice! Didn't know we had this

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+
+	private scalingService: ScalingService | undefined;
+
+	private circuitBreaker: CircuitBreaker;
+
+	private readonly checkIntervalMs: number;
+
+	constructor(
+		private readonly logger: Logger,
+		private readonly globalConfig: GlobalConfig,
```
ivov, line 39: Do we need the entire global config or can we pass in only the relevant subset?

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+
+		this.scalingService = scalingService;
+
+		await this.checkHealthAndUpdateQueue();
+
+		this.healthCheckInterval = setInterval(async () => {
+			await this.checkHealthAndUpdateQueue();
+		}, this.checkIntervalMs);
+
+		this.logger.info('Worker health monitoring started', {
```
ivov, line 82:
```suggestion
		this.logger.debug('Worker health monitoring started', {
```

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+	/**
+	 * Check worker health and pause/resume queue accordingly.
+	 */
+	private async checkHealthAndUpdateQueue() {
+		try {
+			// Use circuit breaker to check health
+			// If circuit is open, checkHealth throws CircuitBreakerOpen
+			await this.circuitBreaker.execute(async () => {
+				if (!(await this.checkHealth())) {
+					throw new Error('Worker is unhealthy');
```
ivov, line 96: 1. `OperationalError`
2. Should we report _what_ exactly failed? PG or Redis?

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+				}
+			});
+
+			// Health check passed, so we can resume pulling from the queue if it has been paused.
+			if (this.isQueuePaused) {
+				await this.resumeQueue();
+			}
+		} catch (error) {
+			if (error instanceof CircuitBreakerOpen) {
+				this.logger.warn('Health check circuit breaker is open, worker is unhealthy');
```
ivov, line 106: Circuit breaker is an implementation detail that the user is unaware of and will likely be confused by during an outage. Let's think about what they'd want to know when seeing this warning.

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+			// Pause queue if not already paused
+			if (!this.isQueuePaused) {
+				await this.pauseQueue();
+			}
+		}
+	}
+
+	private async checkHealth(): Promise<boolean> {
+		return await Promise.resolve(checkWorkerReadiness(this.dbConnection, this.redisClientService));
+	}
```
ivov, line 120: Can be a bit confusing re: liveness vs. readiness state? Ideally we should make it clear everywhere that this is strictly about readiness.

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+
+	/**
+	 * Pause queue processing to prevent pulling new jobs.
+	 */
+	private async pauseQueue() {
+		if (!this.scalingService) {
+			throw new Error('ScalingService not initialized');
```
ivov, line 127: `UnexpectedError`

packages/cli/src/scaling/worker-readiness.ts
```diff
@@ -0,0 +1,16 @@
+import type { DbConnection } from '@n8n/db';
+
+import type { RedisClientService } from '@/services/redis-client.service';
+
+/**
+ * Check if worker is ready (DB connected and migrated, Redis connected).
+ * Shared by health check endpoints and health monitor.
+ */
+export function checkWorkerReadiness(
```
ivov, line 9:
```suggestion
export function isWorkerReady(
```

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+			return;
+		}
+
+		this.scalingService = scalingService;
+
+		await this.checkHealthAndUpdateQueue();
+
+		this.healthCheckInterval = setInterval(async () => {
+			await this.checkHealthAndUpdateQueue();
+		}, this.checkIntervalMs);
```
ivov, line 80: Is this a race condition? `setInterval` with an async callback doesn't wait for the previous invocation to complete. Double pause is harmless, but interleaved pause/resume could cause issues?

packages/cli/src/scaling/worker-health-monitor.ts
ivov: Agree, I thought this'd be part of this PR!

packages/cli/src/commands/worker.ts
```diff
@@ -165,6 +165,10 @@ export class Worker extends BaseCommand<z.infer<typeof flagsSchema>> {
...
+		const { WorkerHealthMonitor } = await import('@/scaling/worker-health-monitor');
+		await Container.get(WorkerHealthMonitor).start(this.scalingService);
```
ivov, line 171: Passing a `ScalingService` like this is a bit awkward no? DI doesn't work? Wonder if there's a better alternative.

packages/cli/src/scaling/scaling.service.ts
```diff
@@ -140,20 +140,55 @@ export class ScalingService {
...
+	}
+
+	/**
+	 * Pause queue processing globally (all workers).
+	 * All workers stop pulling new jobs.
+	 * Already running jobs will continue to completion.
+	 */
+	async pauseQueueGlobal() {
+		this.assertQueue();
+		await this.queue.pause(false, true); // local=false, doNotWaitActive=true
```
ivov, line 161: > true, true -> false, true

A bit confused, is this change intended? Old behavior was main stopping enqueueing on shutdown, but workers could independently continue processing jobs until _they_ started shutdown.

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+
+			// Pause queue if not already paused
+			if (!this.isQueuePaused) {
+				await this.pauseQueue();
+			}
+		}
+	}
+
+	private async checkHealth(): Promise<boolean> {
+		return await Promise.resolve(checkWorkerReadiness(this.dbConnection, this.redisClientService));
```
ivov, line 119: Isn't this a promise already?

packages/cli/src/commands/worker.ts
```diff
@@ -165,6 +165,10 @@ export class Worker extends BaseCommand<z.infer<typeof flagsSchema>> {
 		await this.scalingService.setupQueue();
 
 		this.scalingService.setupWorker(this.concurrency);
+
+		// Start health monitoring after queue and worker are set up
+		const { WorkerHealthMonitor } = await import('@/scaling/worker-health-monitor');
```
ivov, line 170: This always runs so lazyloading makesno difference no?

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+		} catch (error) {
+			this.logger.error('Failed to resume queue', { error });
+		}
+	}
+
+	/**
+	 * Stop health monitoring and cleanup.
+	 */
+	@OnShutdown()
+	stop() {
```
ivov, line 164: Maybe nothing to do, but it struck me that this assumes this only ever gets called during shutdown, as stopping the monitor does not resume the queue, i.e. it doesn't clean up what it did.

packages/cli/src/scaling/worker-health-monitor.ts
```diff
@@ -0,0 +1,181 @@
...
+
+		this.logger.info('Worker health monitoring started', {
+			checkIntervalMs: this.checkIntervalMs,
+		});
+	}
+
+	/**
+	 * Check worker health and pause/resume queue accordingly.
+	 */
+	private async checkHealthAndUpdateQueue() {
```
ivov, line 90: The way I read this we're actually stopping pulling jobs on the first failure rather than on the third because we pause on any exception? So a single blip immediately triggers a pause?

ivov (commented):
Thanks for tackling this! 🙏🏻

mfsiega:
I'm going to close this for now, it's unlikely I'll get time to pick this up in the near future. It can live as prior art if we look back into it in the future.

PR #23541, opened 2025-12-22
feat(editor): Add/remove files in editing ChatHub chat message

autologie:
## Summary

This PR implements the support of adding and removing files in editing chat messages. This makes it possible for the user to continue a conversation when they encounter a file related error, by deleting or replacing the file that the model could not handle.

https://linear.app/n8n/issue/CHA-87/feature-add-the-support-for-editing-attachment

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -260,56 +260,91 @@ export class ChatHubService {
 		const { sessionId, editId, messageId, message, model, credentials, timeZone } = payload;
 		const tz = timeZone ?? this.globalConfig.generic.timezone;
 
-		const workflow = await this.messageRepository.manager.transaction(async (trx) => {
-			const session = await this.getChatSession(user, sessionId, trx);
-			if (!session) {
-				throw new NotFoundError('Chat session not found');
-			}
+		// Store new attachments early to populate 'id' field via BinaryDataService
+		const newStoredAttachments =
```
Cadiac: maybe we do this a bit too early now - we haven't even verified that the session belongs to this user, but we're storing files for that session?

Cadiac: We do have a similar issue on the main send message endpoint too, and that also has been bothering me a little 🤔

autologie: I'll check if I can defer storing files for both endpoints.

Cadiac: it shouldn't be an issue since we roll it back if the session check fails, but it bothers me a little if there's some timing issue there where the files are saved for a brief moment before the session check happens 🤔

Cadiac (commented):
Seems to work well, if we can defer the storing of binary files until session check is done I would be happier (perhaps initialize `let newStoredAttachments = []` first for easy rollback?) , but other than that nice feature addition

packages/cli/src/modules/chat-hub/chat-hub.service.ts
autologie: Addressed in [95195c9](https://github.com/n8n-io/n8n/pull/23541/commits/95195c9138e4c5d0b8f2c737c81eb9873661cffb)

Cadiac (approved):

PR #23542, opened 2025-12-22
fix(core): Stronger allowed path enforcement for read/write Node

mfsiega:
## Summary

Fix allowed path enforcement when reading/writing files.

This change ensures that the path we check for safety _actually_ matches the file we read/write.

https://linear.app/n8n/issue/CAT-1852

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
@@ -130,21 +130,27 @@ export const getFileSystemHelperFunctions = (node: INode): FileSystemHelperFunct
 			}
 		}
 
-		// Verify that the handle we've opened is the same as the path we checked earlier.
-		// This ensures nothing has changed between checking and reading.
-		const fileHandleIdentity = await fileHandle.stat();
-		if (
-			fileHandleIdentity.dev !== pathIdentity.dev ||
-			fileHandleIdentity.ino !== pathIdentity.ino
-		) {
-			throw new NodeOperationError(node, 'The file has changed and cannot be accessed.', {
-				level: 'warning',
-			});
-		}
+		try {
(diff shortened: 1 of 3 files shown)


packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -108,34 +112,59 @@ export const getFileSystemHelperFunctions = (node: INode): FileSystemHelperFunct
...
-							description: 'Symlinks are not allowed.',
-						}),
-					);
-				} else {
-					reject(error);
-				}
+		// Verify that the handle we've opened is the same as the path we checked earlier.
+		// This ensures nothing has changed between checking and reading.
+		const fileHandleIdentity = await fileHandle.stat();
+		if (
```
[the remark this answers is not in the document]
mfsiega: Good catch, I'll fix this.

packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
```diff
@@ -90,6 +91,9 @@ function isFilePathBlocked(resolvedFilePath: ResolvedFilePath): boolean {
 
 export const getFileSystemHelperFunctions = (node: INode): FileSystemHelperFunctions => ({
```
Cadiac, line 92: Are we sure this new logic also works on Windows (for instance)?

mfsiega: Great question. I did actually research briefly when I started this PR and it suggested the answer was 'yes' - `dev` and `ino` apparently mean something slightly different on Windows but can still be relied upon to be stable like this; but I'll test this on Windows to make sure.

Cadiac (approved):
Seems pretty solid to me - didn't test on windows, but I trust your earlier research on that.

PR #23572, opened 2025-12-23
refactor(core, editor): Remove Pyodide

ivov:


packages/cli/src/task-runners/task-runner-module.ts
```diff
@@ -114,6 +114,13 @@ export class TaskRunnerModule {
 	private async startInternalTaskRunners() {
 		a.ok(this.taskBrokerWsServer, 'Task Runner WS Server not loaded');
 
+		const { InternalTaskRunnerDisconnectAnalyzer } = await import(
+			'@/task-runners/internal-task-runner-disconnect-analyzer'
+		);
+		this.taskBrokerWsServer.setDisconnectAnalyzer(
+			Container.get(InternalTaskRunnerDisconnectAnalyzer),
+		);
+
```
ivov, line 123: Repositioned this to fix a bug. The disconnect analyzer was outside the `if (isNativePythonRunnerEnabled)` block, but with a `return` statement inside it, so if Python was enabled but requirements failed, the disconnect analyzer wasn't set up for the internal JS runner.

tomi (approved):
🚀

PR #23575, opened 2025-12-23
fix(core): Fix race condition in executions pruning on leadership change

ivov:
## Summary

On leader stepdown, `stopPruning` was not actually stopping the pruning timers. This method checks `isEnabled` which depends on `isLeader`, but `markAsFollower` is called _before_ the `@OnLeaderStepdown` event fires, so `isLeader` was already `false` leading `stopPruning` to bail early, resulting in redundant work.

This bug manifests only in executions pruning. Checked all other decorated services and cleanup there is either unconditional or relies on whether a timer or resource exists, not leadership state.

Longer term, we should:

- Ensure cleanup methods don't guard on state, i.e. keep them idempotent.
- Ensure consumers only rely on the takeover or stepdown event, not on leadership state.

I'll add a follow-up for this.

https://n8nio.slack.com/archives/C069HS026UF/p1766493370062609

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/services/pruning/executions-pruning.service.ts
+++ b/packages/cli/src/services/pruning/executions-pruning.service.ts
@@ -77,8 +77,6 @@ export class ExecutionsPruningService {
 
 	@OnLeaderStepdown()
 	stopPruning() {
-		if (!this.isEnabled) return;
-
 		clearInterval(this.softDeletionInterval);
 		clearTimeout(this.hardDeletionTimeout);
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/services/pruning/executions-pruning.service.ts
```diff
@@ -77,8 +77,6 @@ export class ExecutionsPruningService {
 
 	@OnLeaderStepdown()
 	stopPruning() {
-		if (!this.isEnabled) return;
-
 		clearInterval(this.softDeletionInterval);
 		clearTimeout(this.hardDeletionTimeout);
```
CharlieKolb, line 83: Should we only post the debug line below if the interval or timeout were `!== undefined`?

CharlieKolb (dismissed):
Thanks for the quick fix 🙏

CharlieKolb (approved):

PR #7307, opened 2023-09-29
feat(core): Switch binary filesystem mode to nested path structure

ivov:
Depends on #7253 | Story: [PAY-863](https://linear.app/n8n/issue/PAY-863/switch-binary-filesystem-mode-to-nested-path-structure)

This PR introduces `filesystem-v2` to store binary data in the filesystem in the same format as `s3`.

PR #23580, opened 2025-12-23
refactor(core): Remove flat format for binary data

ivov:
## Summary

Remove the legacy flat format for binary data, i.e. storing files with no nesting at `/home/node/.n8n/binaryData/` which used to require scanning the dir and pattern matching to identify binary data files to delete.

We moved away from this format [over two years ago](https://github.com/n8n-io/n8n/pull/7307), so the only way binary data in this format may survive today is (a) if an instance never prunes at all and started storing binary data over two years ago, or (b) if a user's pruning cycle is longer than two years. As discussed, both are unlikely enough that we find it reasonable to remove this format after v2 is released.

https://linear.app/n8n/issue/CAT-1794

## Review / Merge checklist

- [ ] Tests included. 


tomi (approved):
🚀

PR #23606, opened 2025-12-24
fix(ai-builder): Improve handling of max recursion errors

schrothbn:
## Summary
 Improves error handling in the AI workflow builder's multi-agent system when internal recursion limits are reached. Previously, the system would tell users to "build the workflow manually" even when a workflow was successfully created, causing confusion and harming trust.

 ### The Problem:
  - When a sub-agent hits its recursion limit, the error propagated to the supervisor as a hard failure
  - The responder would apologize and claim it couldn't build the workflow
  - In some cases, the workflow WAS already built or nearly complete, creating confusion
  - Users couldn't continue or retry - they were stuck with "build it manually" messages

### The Solution:
  This PR implements context-aware error handling that:

  1. Differentiates error scenarios - Provides different guidance based on:
    - Recursion error WITH workflow created → Acknowledge partial success, offer to continue
    - Recursion error WITHOUT workflow → Suggest simplifying the request
    - Other errors → General retry guidance
  2. Enables error recovery - Adds a "clear error state" mechanism that:
    - Allows users to continue after hitting recursion limits
    - Preserves partial workflow metadata for recovery
    - Prevents infinite error loops
  3. Improves configurator recovery - When configurator receives a partially built workflow from a failed builder run, it enters "recovery mode" with specific instructions to complete the configuration

https://linear.app/n8n/issue/AI-1812/supervisor-needs-to-handle-internal-errors-better-instead-of-telling

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/subgraphs/test/integration/multi-agent-error-handling.integration.test.ts
```diff
@@ -0,0 +1,326 @@
...
+			const messageContent =
+				typeof lastMessage.content === 'string'
+					? lastMessage.content.toLowerCase()
+					: JSON.stringify(lastMessage.content).toLowerCase();
+
+			console.log('\n📨 Responder message:', lastMessage.content);
+			console.log('\n📊 Workflow state:', {
+				nodeCount: result.workflowJSON.nodes.length,
+				hasError: errorLog.length > 0,
+			});
```
burivuhster: Do we need these console logs or were they just for debugging the test?

packages/@n8n/ai-workflow-builder.ee/src/utils/state-modifier.ts
```diff
@@ -44,6 +46,33 @@ export function determineStateAction(
...
+				errorMessage.includes('iteration limit')
+			);
+		});
+
+		const hasAlreadyCleared = coordinationLog.some(
+			(entry) =>
+				entry.phase === 'state_management' &&
+				entry.summary.includes('Cleared') &&
+				entry.summary.includes('recursion'),
+		);
```
burivuhster: Should we look only for entries happened after max recursion errors?
If we look through the whole log – then it will apply to the whole session, not to the most recent error

burivuhster (approved):
Thanks for addressing my comments! LGTM 🚀

Issue #23602, opened 2025-12-24
basic auth, n8n form. Problem submitting response


AN1301:
### Bug Description

N8n Form has basic auth enabled and a Next Form Page. When I go to it, I get "Problem submitting response. Please try again or contact support if the problem persists."

### To Reproduce

basic auth enabled
1)Trigger n8n form
2)Next Form Page
Error Problem submitting response Please try again or contact support if the problem persists"
 

### Expected behavior

Go to Next Form Page after Trigger n8n form

### Debug Info

# Debug info

## core

- n8nVersion: 2.1.4
- platform: docker (self-hosted)
- nodeJsVersion: 22.21.1
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- license: enterprise (production)
- consumerId: 8d8a6ebe-5108-4de5-9092-7812ec4b393a

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/143.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2025-12-24T05:49:57.291Z

### Operating System

Ubuntu 22.04

### n8n Version

2.1.4

### Node.js Version

22.21.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

KazuSh1geru:
I can reproduce the same behavior on my side. 

Environment:
- Basic Auth enabled
- n8n Forms with Next Form Page

Result:
- Submission fails with

“Problem submitting response. Please try again or contact support if the problem persists.”

<img width="504" height="191" alt="Image" src="https://github.com/user-attachments/assets/59def20b-9c9c-477b-8715-cb2fe1dba130" />

```
n8n version: 1.123.6
Plan: Pro (Cloud)
```

KazuSh1geru:
## Root Cause Analysis

The issue occurs when Basic Authentication is enabled on the Form Trigger node and the form has a "Next Form Page" configured. 

**Problem**: The JavaScript `fetch()` call on line 996 of `form-trigger.handlebars` does not include authentication credentials, causing the POST request to fail with a 401 Unauthorized error when Basic Auth is enabled.

[...]

PR #23608, opened 2025-12-24
fix(Form Trigger Node): Skip authentication check for form submissions

KazuSh1geru:
## Summary
Fixes an issue where multi-page forms with Basic Authentication fail to submit because authentication is unnecessarily checked on POST requests (form submissions).

## Root Cause
The Form Trigger node validates authentication on both GET (form display) and POST (form submission) requests. However, if a user can already see the form (passed GET authentication), they should be able to submit it without re-authenticating on POST.

## Changes
1. **Core Fix** (`packages/nodes-base/nodes/Form/utils/utils.ts`)
   - Modified `formWebhook()` to only validate authentication on GET requests
   - Added comment explaining the rationale
   - Moved `method` variable declaration earlier for clarity

2. **Template Fix** (`packages/cli/templates/form-trigger.handlebars`)
   - Fixed quote syntax for `useResponseData` value attribute (minor cleanup)

3. **E2E Test** (`packages/testing/playwright/tests/e2e/nodes/form-trigger-node.spec.ts`)
   - Added comprehensive test for multi-page forms with Basic Auth
   - Uses Playwright's `httpCredentials` to simulate authenticated browser context
   - Validates that form submission works after authentication

## Test Plan
- [x] Added E2E test for Basic Auth multi-page forms
- [x] Test passes: form displays with auth and submits successfully
- [x] Existing tests pass (verified no regression)
- [x] Manual testing: multi-page forms with Basic Auth work end-to-end

## Breaking Changes
None. This is a bug fix that makes the authentication behavior more intuitive.

## Related Issues

- #23602

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/cli/templates/form-trigger.handlebars
+++ b/packages/cli/templates/form-trigger.handlebars
@@ -689,7 +689,6 @@
 					<a id='redirectUrl' href='{{redirectUrl}}' style='display: none;'></a>
 				{{/if}}
 				<input id="useResponseData" style="display: none;" value="{{useResponseData}}" />
-				<input id="requiresAuth" style="display: none;" value="{{requiresAuth}}" />
 			</section>
 		</div>
 		<script>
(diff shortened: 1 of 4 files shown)


KazuSh1geru:
close #23602

PR #23609, opened 2025-12-24
feat(ai-builder): Improve generation across data_transformation category

burivuhster:
## Summary

This PR adds instructions to the prompt to:
- Prefer basic nodes like Set, Filter, etc. over Code node
- Prefer Data Table node for data storage when it's not specified otherwise

It also adds a command line argument for evals to filter examples by category/technique, e.g.:
```
pnpm eval:pairwise --technique data_transformation --name "data_transformation_2025_12_23"
```

PR also includes one small but important fix in spec evals:
There were cases when LLM-judge incorrectly treated pass as violation. Prepending each check from "don'ts" section with "DO NOT" seems to solve it.

- https://linear.app/n8n/issue/AI-1822/feature-improve-eval-scores-of-1-technique-data-transformation

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/builder.prompt.ts
```diff
@@ -299,6 +299,18 @@ COMMON MISTAKES TO AVOID:
 ✅ Window Buffer Memory → AI Agent (CORRECT)
 </connection_type_reference>`;
 
+const BASIC_NODES_PREFERENCE = `**Built-in Nodes vs. Code Node**
```
harshalnpatil: can we move these edits to the data-transformation file? our goal for this ticket was to focus our energy on 1 technique best practice at a time. this reduces context bloat in the main prompts. we want to also 
- part 1 we can mention code nodes non-usage on L70 of data-transformation. #### Code Node (n8n-nodes-base.code) 
- part 2 (using sheets or tables doesn't seem to be a focus in data-transformation file yet, which is confusing.)

packages/@n8n/ai-workflow-builder.ee/src/tools/best-practices/scraping-and-research.ts
```diff
@@ -104,6 +104,10 @@ Pitfalls:
 
 Purpose: Introduces delays to respect rate limits and avoid overloading servers
```
harshalnpatil, line 106: not sure where to add comments on the data-transformation file as that was focus for this ticket but no edits to it.
- should remove from it as is generic: ### General Workflow Issues 
- this seems generic as well, especially about using sub-workflows as that isn't what builder will do rightaway and is not technique-specific. ### Performance Pitfalls
- this is not specific on what to do what not to do so should remove (or make specific) ### Data Validation Pitfalls
- this issue doesn't have a clear fix because "validate" doesn't make sense for the agent. "Missing Keys" secrtion in "merge node"
- "Workflow Orchestration" section is generic and should be removed too. these generic sections are diluting our prompt effectiveness.

burivuhster: * Removed `### General Workflow issues`
* Agree about sub-workflows, but other seem relevant to data transformation
* Changed validation section a bit
* Fixed
* Agree, removed

packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/builder.prompt.ts
burivuhster: re data tables – the proper way would be to add a new category/technique "data_persistence" and move it there. I would do this in a separate PR

harshalnpatil (approved):
thank you for the updates. lets go ahead with these changes. the eval script changes will help and these prompt improvements will help.

PR #23610, opened 2025-12-24
feat(core): Add data table resources to the public API

RicardoE105:
## Summary

This PR adds a **Data Table Row resource** to the n8n public API, enabling programmatic CRUD operations on data table rows via API keys.

### New Endpoints

**GET `/data-tables/:dataTableId/rows`** - Retrieve rows
- Supports filtering, sorting, searching, and pagination
- Filter conditions: `eq`, `neq`, `gt`, `gte`, `lt`, `lte`, `like`, `ilike`
- Filter types: `AND`, `OR` combinations
- Example:
  ```bash
  curl -H "X-N8N-API-KEY: your-key" \
    "http://localhost:5678/api/v1/data-tables/{id}/rows?filter={\"type\":\"and\",\"filters\":[{\"columnName\":\"status\",\"condition\":\"eq\",\"value\":\"active\"}]}&skip=0&take=100"
  ```

**POST `/data-tables/:dataTableId/rows`** - Insert rows
- Supports different return types: `count` (default), `id`, or `row`
- Example:
  ```bash
  curl -X POST -H "X-N8N-API-KEY: your-key" \
    -H "Content-Type: application/json" \
    -d '{"data":[{"name":"John","email":"john@example.com"}],"returnType":"row"}' \
    "http://localhost:5678/api/v1/data-tables/{id}/rows"
  ```

**PATCH `/data-tables/:dataTableId/rows/update`** - Update rows
- Filter-based bulk updates
- Optional `returnData` and `dryRun` parameters
- Example:
  ```bash
  curl -X PATCH -H "X-N8N-API-KEY: your-key" \
    -H "Content-Type: application/json" \
    -d '{"filter":{"type":"and","filters":[{"columnName":"status","condition":"eq","value":"pending"}]},"data":{"status":"completed"},"returnData":true}' \
    "http://localhost:5678/api/v1/data-tables/{id}/rows/update"
  ```

**POST `/data-tables/:dataTableId/rows/upsert`** - Upsert row
- Insert or update a single row based on filter
- Example:
  ```bash
  curl -X POST -H "X-N8N-API-KEY: your-key" \
    -H "Content-Type: application/json" \
    -d '{"filter":{"type":"and","filters":[{"columnName":"email","condition":"eq","value":"john@example.com"}]},"data":{"name":"John Doe","status":"active"},"returnData":true}' \
    "http://localhost:5678/api/v1/data-tables/{id}/rows/upsert"
  ```

**DELETE `/data-tables/:dataTableId/rows/delete`** - Delete rows
- Requires filter to prevent accidental deletion of all data
- Example:
  ```bash
  curl -X DELETE -H "X-N8N-API-KEY: your-key" \
    "http://localhost:5678/api/v1/data-tables/{id}/rows/delete?filter={\"type\":\"and\",\"filters\":[{\"columnName\":\"status\",\"condition\":\"eq\",\"value\":\"archived\"}]}"
  ```

### Authentication & Authorization

**New API Key Scopes:**
- `dataTableRow:create` - Insert rows
- `dataTableRow:read` - Retrieve rows
- `dataTableRow:update` - Update rows
- `dataTableRow:upsert` - Upsert rows
- `dataTableRow:delete` - Delete rows

The new scopes are available at `GET /rest/api-keys/scopes` for users with `apiKey:manage` permission.

### How to Test

1. **Test GET (retrieve rows):**
   ```bash
   curl -H "X-N8N-API-KEY: your-key" \
     "http://localhost:5678/api/v1/data-tables/{dataTableId}/rows"
   ```

2. **Test POST (insert rows):**
   ```bash
   curl -X POST -H "X-N8N-API-KEY: your-key" \
     -H "Content-Type: application/json" \
     -d '{"data":[{"name":"Test","email":"test@example.com","status":"active"}],"returnType":"row"}' \
     "http://localhost:5678/api/v1/data-tables/{dataTableId}/rows"
   ```

3. **Test filtering:**
   ```bash
   curl -H "X-N8N-API-KEY: your-key" \
     "http://localhost:5678/api/v1/data-tables/{dataTableId}/rows?filter={\"type\":\"and\",\"filters\":[{\"columnName\":\"status\",\"condition\":\"eq\",\"value\":\"active\"}]}"
   ```

4. **Test UPDATE:**
   ```bash
   curl -X PATCH -H "X-N8N-API-KEY: your-key" \
     -H "Content-Type: application/json" \
     -d '{"filter":{"type":"and","filters":[{"columnName":"status","condition":"eq","value":"active"}]},"data":{"status":"completed"},"returnData":true}' \
     "http://localhost:5678/api/v1/data-tables/{dataTableId}/rows/update"
   ```

5. **Test DELETE with dry run:**
   ```bash
   curl -X DELETE -H "X-N8N-API-KEY: your-key" \
     "http://localhost:5678/api/v1/data-tables/{dataTableId}/rows/delete?filter={\"type\":\"and\",\"filters\":[{\"columnName\":\"status\",\"condition\":\"eq\",\"value\":\"completed\"}]}&dryRun=true"
   ```

**Testing:**
- ✅ 36 integration tests covering all CRUD operations, filtering, security
- ✅ 23 unit tests for handler logic, mocking, error cases
- All tests passing

https://linear.app/n8n/issue/ADO-4066

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/permissions/src/public-api-permissions.ee.ts
```diff
@@ -112,8 +133,7 @@ export const getApiKeyScopesForRole = (user: AuthPrincipal) => {
 
 	return [
 		...new Set(
-			user.role.scopes
-				.map((scope) => scope.slug)
+			(user.role.scopes.map((scope) => scope.slug) as Array<Scope | ApiKeyScope>)
```
CharlieKolb, line 151: is this `as` needed? Imo we should prioritize keeping ApiKeyScope a subset of Scope by reusing the same key string, even if this one sounds better.

packages/cli/build.log
```diff
@@ -1,4 +1,8 @@
 
-> n8n@1.121.0 build /Users/mutasem/repos/n8n-worktree/global-creds/packages/cli
+> n8n@2.2.0 build /Users/ricardo/Documents/n8n-io/n8n/packages/cli
```
CharlieKolb: does this file do anything? looks like an accidental addition and now change? :D

packages/cli/src/permissions.ee/check-access.ts
```diff
@@ -24,7 +24,13 @@ export async function userHasScopes(
 		credentialId,
 		workflowId,
 		projectId,
-	}: { credentialId?: string; workflowId?: string; projectId?: string } /* only one */,
+		dataTableId,
+	}: {
```
CharlieKolb, line 29: [nit If you're feeling fancy...]

```
type ExactlyOne<T, Keys extends keyof T = keyof T> = 
  Keys extends keyof T
    ? { [K in Keys]: T[K] } & { [K in Exclude<keyof T, Keys>]?: never }
    : never;

Partial<ExactlyOne<{
  credentialId: string;
  workflowId: string;
  dataTableId: string;
  projectId: string;
}>>;
```

packages/cli/src/permissions.ee/check-access.ts
```diff
@@ -99,9 +105,24 @@ export async function userHasScopes(
 		);
 	}
 
+	if (dataTableId) {
+		const { DataTableRepository } = await import('@/modules/data-table/data-table.repository');
```
CharlieKolb, line 115: What happens if the module is disabled? Would pinging the API enable the module? 🤔

CharlieKolb (commented):
Some initial thoughts before heading out for the day

packages/@n8n/permissions/src/public-api-permissions.ee.ts
RicardoE105: The problem is that the scopes in the public API are more granular than the internal API. The reason being that the user can control what operations the API key can perform. In the internal we have only read and write, and in the pubic api dataTableRow:create, dataTableRow:read, dataTableRow:update, dataTableRow:delete, dataTableRow:upsert.

packages/cli/src/permissions.ee/check-access.ts
RicardoE105: Why would the module be disabled? 🤔

RicardoE105: there is not env variable to disable data tables 🤔

CharlieKolb: Well users can disable the module if they choose to 😅 Via `N8N_DISABLED_MODULES`

I guess technically all of these APIs should return an error in that case

RicardoE105: ah good point. will add a check for that.

CharlieKolb (dismissed):

packages/nodes-base/nodes/DataTable/actions/table/list.operation.ts
```diff
@@ -67,9 +67,8 @@ export const description: INodeProperties[] = [
 				options: [
 					{ name: 'Created', value: 'createdAt' },
 					{ name: 'Name', value: 'name' },
-					{ name: 'Size', value: 'sizeBytes' },
 					{ name: 'Updated', value: 'updatedAt' },
-				] satisfies Array<{ name: string; value: ListDataTableOptionsSortByKey }>,
+				],
```
RicardoE105, line 70: remove sizeBytes sorting as that never worked

packages/@n8n/permissions/src/public-api-permissions.ee.ts
CharlieKolb: Ah alright, thanks for explaining 👍  We can do `.map<ApiKeyScope | Scope>((scope) => scope.slug)` to get rid of the `as` if you want to please cubic

CharlieKolb (approved):
Great work 🙌

PR #23674, opened 2025-12-28
Fix SplitInBatches behavior across v1, v2, v3

aarav-12:
## Problem
Nested "Loop Over Items" nodes skip processing items from the second outer loop iteration.

**Example:**
- Outer loop: 2 departments
- Inner loop: 2 employees per department
- **Expected:** All 4 employees processed
- **Actual:** Only first 2 employees processed

## Root Cause
When the inner loop finished processing the first batch, it didn't detect that new data from the outer loop was a new batch. It treated it as a continuation of the previous loop.

## Solution
Added detection to reset the inner loop when it receives new items after completion. This ensures each outer loop iteration properly triggers a new inner loop cycle.

## Files Changed
- `SplitInBatches/v1/SplitInBatchesV1.node.ts`
- `SplitInBatches/v2/SplitInBatchesV2.node.ts`
- `SplitInBatches/v3/SplitInBatchesV3.node.ts`

### DIFF
--- a/packages/nodes-base/nodes/SplitInBatches/v1/SplitInBatchesV1.node.ts
+++ b/packages/nodes-base/nodes/SplitInBatches/v1/SplitInBatchesV1.node.ts
@@ -72,8 +72,19 @@ export class SplitInBatchesV1 implements INodeType {
 
 		const options = this.getNodeParameter('options', 0, {});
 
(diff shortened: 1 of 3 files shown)


v-karbovnichy:
wow this is such a great fix! I've been stumbling this for years and said to others as a mantra "do not use nested loops, they do not work"

PR #23677, opened 2025-12-28
feat(Data Table Node): Add Order By feature

jason-schell:
## Summary

Add Order By-feature to the Data Table node. Users can now order by a column in data tables in an ascending or descending order.

https://linear.app/n8n/issue/ADO-4574/feature-add-order-by-support-for-data-tables-node

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/DataTable/common/utils.ts
```diff
@@ -47,10 +46,8 @@ async function resolveDataTableId(
 
 		return response.data[0].id;
 	} else {
-		// For 'list' and 'id' modes, extract the value directly
-		return ctx.getNodeParameter(DATA_TABLE_ID_FIELD, index, undefined, {
-			extractValue: true,
-		}) as string;
+		// For 'list' and 'id' modes, the value is already the data table ID
+		return resourceLocator.value;
```
CharlieKolb: We usually need to use the index here to support expressions that consume the id via expression from a previous node, where the id could change for each input row.

CharlieKolb (commented):
Thanks for the contribution! Code looks clean and well-tested 👌 

We should probably fill one of our system columns (id, createdAt, updatedAt) for the default column instead of `''`.

Note that just as with `Limit` this applies per input row. Not sure how unexpected or worth changing this is
image

I'm happy to proceed personally - though unlike filter (where we need the db to do it for us) we could also sort our results manually if we prefer this.

I'd also like to wait for Rob to be back from holiday Jan 2nd to review the UX if you don't mind.

CharlieKolb (approved):
🚀

PR #23684, opened 2025-12-29
ci: Skip Docker base image builds for forks

mawazawa:
## Summary

Forks don't have the `DOCKER_USERNAME` secret configured, causing the Docker base image builds to fail with invalid tag format errors. This PR adds a condition to skip the build job for non-upstream repositories while preserving `workflow_dispatch` functionality for manual triggers.

## Changes

- Added `if` condition to `jobs.build` to only run for `n8n-io/n8n` or manual triggers
- Prevents fork builds from failing due to missing Docker credentials

## Test plan

- [x] Fork builds will skip the Docker job (no more failures)
- [x] Upstream `n8n-io/n8n` continues to build on push to master
- [x] `workflow_dispatch` manual triggers still work for testing

---

### DIFF
--- a/.github/workflows/docker-base-image.yml
+++ b/.github/workflows/docker-base-image.yml
@@ -20,6 +20,13 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    # Skip Docker builds for forks (they don't have DOCKER_USERNAME secret)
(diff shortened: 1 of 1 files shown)


shortstacked:
Hey @mawazawa ,
Thank you for the PR. This PR won't be required though as changes to the base image come from inside the team so there shouldn't be a case where a community PR triggers this workflow.

PR #23694, opened 2025-12-29
fix: Add data insight metadata  migration

ireneea:
## Summary

Changed the `insights_metadata` `workflowId` property to varchar(36) instead of `varchar(16)` 

fixes PAY-4373

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/migrations/common/1766500000000-ExpandInsightsWorkflowIdLength.ts
```diff
@@ -0,0 +1,34 @@
+import type { IrreversibleMigration, MigrationContext } from '../migration-types';
+
+const INSIGHTS_METADATA_TABLE_NAME = 'insights_metadata';
```
guillaumejacquart: nit: i think we do not use that uppercase style for const anymore. Also maybe the the const here is not necessary, we can put the string in the const tableName (same for columnName)

packages/@n8n/db/src/migrations/common/1766500000000-ExpandInsightsWorkflowIdLength.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Note: SQLite does not enforce VARCHAR length limits, so no migration is needed for SQLite.
+ */
+export class ExpandInsightsWorkflowIdLength1766500000000 implements IrreversibleMigration {
+	async up({ isMysql, isPostgres, escape, queryRunner }: MigrationContext) {
+		const tableName = escape.tableName(INSIGHTS_METADATA_TABLE_NAME);
+		const columnName = escape.columnName(WORKFLOW_ID_COLUMN_NAME);
+
+		if (isMysql) {
+			await queryRunner.query(`ALTER TABLE ${tableName} MODIFY COLUMN ${columnName} VARCHAR(36);`);
```
guillaumejacquart: I think we do not add mysql migrations anymore (as mysql is not supported anymore). If the fix is for version > 2.0.0, I think we should remove that

guillaumejacquart (commented):
thanks ! some comments

packages/@n8n/db/src/migrations/common/1766500000000-ExpandInsightsWorkflowIdLength.ts
```diff
@@ -0,0 +1,31 @@
...
+ * The fix aligns with the codebase standard of varchar(36) for workflowId columns.
+ *
+ * Note: SQLite does not enforce VARCHAR length limits, so no migration is needed for SQLite.
+ */
+export class ExpandInsightsWorkflowIdLength1766500000000 implements IrreversibleMigration {
+	async up({ isPostgres, escape, queryRunner }: MigrationContext) {
+		// Only PostgreSQL needs this migration:
+		// - SQLite does not enforce VARCHAR length constraints
+		// - MySQL is no longer supported in n8n 2.0+
+		if (isPostgres) {
```
guillaumejacquart: If we only have postgres issue, I guess we can just put this file on the postgres migration folder, no ? and remove the if

guillaumejacquart (commented):
one more comment :)

packages/@n8n/db/src/migrations/common/1766500000000-ExpandInsightsWorkflowIdLength.ts
ireneea: sorted ✅

guillaumejacquart (approved):

PR #23717, opened 2025-12-30
fix(editor): Fix NDV state getting stale when navigating using the floating buttons

scdekov:
## Summary

There was a similar issue in the FilterCondition component used in the If node where local component state was not being updated when switching to an empty node.

https://linear.app/n8n/issue/ADO-4573/set-node-detail-panel-doesnt-change-when-navigating-using-buttons-on

https://www.loom.com/share/0a6a2c1c0a8249f98a89a57e0872412e

## Review / Merge checklist

- [ ] Tests included. 


CharlieKolb (approved):
Please update the title to be more user facing for the release notes

PR #23724, opened 2025-12-30
chore(editor): Address Vue warnings

autologie:
## Summary

This PR cleans up following warnings on the browser's console.

When navigating
```
[Vue warn]: onMounted is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup(). If you are using async setup(), make sure to register lifecycle hooks before the first await statement.
[Vue warn]: onBeforeUnmount is called when there is no active component instance to be associated with. Lifecycle injection APIs can only be used during execution of setup(). If you are using async setup(), make sure to register lifecycle hooks before the first await statement.
```

When opening a dynamic modal
```
[Vue warn]: Vue received a Component that was made a reactive object. This can lead to unnecessary performance overhead and should be avoided by marking the component with `markRaw` or using `shallowRef` instead of `ref`. 
```

When opening NDV (with specific parameter)
```
[Vue warn]: Invalid prop: type check failed for prop "values". Expected Object, got Undefined
```
```
[Vue warn]: Failed setting prop "nodeType" on <div>: value [object Object] is invalid. TypeError: Cannot set property nodeType of #<Node> which has only a getter
```

When building `@n8n/chat` package
```
../design-system/src/v2/components/DropdownMenu/DropdownMenu.vue:216:30 - error TS2774: This condition will always return true since this function is always defined. Did you mean to call it instead?

216   <DropdownMenuTrigger v-if="slots.trigger" as-child :disabled="disabled">
                                 ~~~~~~~~~~~~~
../design-system/src/v2/components/DropdownMenu/DropdownMenu.vue:241:17 - error TS2774: This condition will always return true since this function is always defined. Did you mean to call it instead?

241     <slot v-if="$slots.content" name="content" />
```

When ChatHub settings page is opened
```
ElementPlusError: [ElOnlyChild] no valid child node found
```

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
@@ -112,7 +112,6 @@ import {
 	isCommunityPackageName,
 } from 'n8n-workflow';
 import { computed, nextTick, ref } from 'vue';
-import { useClipboard } from '@/app/composables/useClipboard';
 import { useUniqueNodeName } from '@/app/composables/useUniqueNodeName';
 import { injectWorkflowState } from '@/app/composables/useWorkflowState';
 import { isPresent, tryToParseNumber } from '@/app/utils/typesUtils';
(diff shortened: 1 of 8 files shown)


packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
autologie, line 1: This change *happens to* fix warnings regarding onMount/onBeforeUnmount because `useClipboard` from `@vueuse/core` doesn't use it. A proper fix would involve more work (to stop using composables outside of setup())

CharlieKolb (approved):
Lovely!

PR #23732, opened 2025-12-30
fix(core): ChatHub: respond error details from execution

autologie:
## Summary

This PR implements the logic in ChatHub streaming response to respond error details when it happens, to help us troubleshoot error reports better and let the user know what to do (if the error is actionable)

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -813,6 +813,12 @@ export class ChatHubService {
 				return text;
 			}
 
+			if (chunk.type === 'error' && !chunk.content) {
+				chunk.content = executionId
+					? await this.waitForErrorDetails(executionId, workflowData.id)
```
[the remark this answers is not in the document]
autologie: In this case FE will show "Something went wrong" as it does today, I think the current behavior is fine

packages/cli/src/modules/chat-hub/stream-capturer.ts
```diff
@@ -8,15 +8,18 @@ import type { ChatHubMessage } from './chat-hub-message.entity';
...
 
-export type ChunkTransformer = (chunk: string) => void;
+export type ChunkTransformer = (chunk: string) => Promise<string>;
 
 export function interceptResponseWrites<T extends Response>(
 	res: T,
 	transform: ChunkTransformer,
 ): T {
 	const originalWrite = res.write.bind(res) as Write;
+	const originalEnd = res.end.bind(res);
```
Cadiac: Image

can we do a cast to `ServerResponse['end']` here, otherwise there's quite a few warnings on this file (that my editor displays :D)

Cadiac (dismissed):
Seems to work, small suggestion on preventing warnings with a typecast. 👍 Thanks for improving this

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -1546,4 +1552,50 @@ export class ChatHubService {
...
+		const maxRetries = 5;
+		let retries = 0;
+		let errorText: string | undefined;
+
+		while (!errorText) {
+			try {
+				const execution = await this.executionRepository.findWithUnflattenedData(executionId, [
+					workflowId,
+				]);
+				if (execution) {
```
Cadiac: Should we check here that the execution has also stopped, or can we trust that's the case?

On my branch / non multimain poller similar code did also check for

`EXECUTION_FINISHED_STATUSES.includes(execution.status)`, 

```
export const EXECUTION_FINISHED_STATUSES: ExecutionStatus[] = [
	'canceled',
	'crashed',
	'error',
	'success',
] as const;
```

That's... probably unnecessary here?

autologie: Added the check, and changed to break loop regardless of error text is found (data won't change once finished status is reached) [8b8d581](https://github.com/n8n-io/n8n/pull/23732/commits/8b8d58157cc98c93b2e03301e4c850735a5cc1dc)

Cadiac (approved):

PR #23736, opened 2025-12-30
feat(ai-builder): LangSmith evaluation improvements

OlegIvaniv:
## Summary
- Add `--name` option to LLM-as-judge evaluations for custom experiment naming
- Add `--verbose` flag to LLM-as-judge evaluations for detailed logging
- Add LangSmith trace filtering to reduce payload sizes and avoid 403 errors

### --name option
Adds support for specifying a custom experiment name when running LangSmith evaluations via `--name <name>` flag.

### --verbose flag
Adds `--verbose` / `-v` flag to `eval:langsmith` for consistent logging output using EvalLogger. This ensures trace filtering statistics are properly formatted.

### Trace filtering
Adds trace payload filtering to reduce the size of data sent to LangSmith during evaluations:
- Summarizes large fields like `cachedTemplates`, `parsedNodeTypes`, and `workflowContext`
- Uses closure-scoped state to avoid issues with parallel evaluations
- Adds `LANGSMITH_MINIMAL_TRACING` env var (default: `true`) to control filtering
- Logs filtering statistics at the end of evaluation runs

This prevents errors from oversized payloads during high-concurrency evaluations.

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/evaluations/core/trace-filters.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/evaluations/core/trace-filters.ts
@@ -15,6 +15,16 @@ const LARGE_STATE_FIELDS = ['cachedTemplates', 'parsedNodeTypes'] as const;
 const LANGCHAIN_SERIALIZABLE_KEYS = ['lc_serializable', 'lc_kwargs', 'lc_namespace'] as const;
 
 /**
+ * Large context fields within workflowContext that should be filtered.
+ */
+const LARGE_CONTEXT_FIELDS = ['executionData', 'executionSchema', 'expressionValues'] as const;
+
+/**
+ * Threshold for summarizing workflows instead of including full definition.
+ */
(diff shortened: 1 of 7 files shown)


packages/@n8n/ai-workflow-builder.ee/evaluations/core/trace-filters.ts
```diff
@@ -0,0 +1,346 @@
...
+	}
+}
+
+/**
+ * Summarize a large context field to a placeholder string.
+ */
+function summarizeContextField(key: string, value: unknown): string {
+	switch (key) {
+		case 'executionData':
+			return '[execution data omitted]';
```
burivuhster, line 125: Can it create a confusion later on between execution data actually omitted VS trimmed in logger?

burivuhster (approved):
Looks good! My only concern is that replacing data with [omitted] values may create a confusion later on. Would it make sense to use some more explicit placeholders here, like `trimmed-by-logger`?

PR #1509, opened 2021-03-04
Fixed an issue with queue mode for executions that should not be saved

krynble:
We had an issue with workflows that were set to not save success or error when working with queue.

In case of success or error, the worker process would finish the execution and remove it from database. Only then it would signal the main process about the conclusion.

Main process would be unable to properly return data (to a webhook for instance) and would fail, even though the execution actually succeeded.

We are now delegating this responsibility back to main process.

Also had to change the executions list update on frontend as executions might need to be removed. Consider this example: you have a main workflow that executes a child workflow via `Execute workflow` node. This second execution is saved to the database but neither `activeExecutions` instance nor Redis know about it. So it is an execution that until terminates, is something we cannot know what will happen. It may get deleted once finished, it may be saved or it may simply crash and be stuck forever. We are displaying it as `Unknown` since #1475 . If it was deleted, the frontend would contained a cached version with this execution. Unless the user reopens the Executions window, it would stay there forever. Now we are comparing results from backend with information in `store` to avoid these inconsistencies.

Also two other issues were fixed:
- For regular executions (no queue), the deletion of successful and errored executions was not being `await`-ed. This means that a race condition was created, making most of the execution not get deleted.
- Executions (with our without queue) that were running on `own` process as `EXECUTIONS_MODE` would never finish correctly when attempting to terminate n8n gracefully. The main process would receive `SIGINT` OR `SIGTERM` and immediately close child processes, causing executions to abruptly interrupt and making the main process wait forever.

PR #23738, opened 2025-12-30
fix: Stop saving execs when flag is set to not save

shortstacked:
## Summary

In queue mode, executions were being saved to the database even when workflow settings said not to save them.

### Root Cause

The bug manifested when main was under pressure but workers weren't (e.g., 1 main with 3 workers):

1. Worker **always** saved executions to DB via `hookFunctionsSaveWorker`
2. Main was supposed to delete based on settings via `getLifecycleHooksForScalingMain`
3. Main's delete depended on Bull's `job.finished()` resolving
4. Under load, `job.finished()` fails while `global:completed` events still work (different Redis mechanisms)
5. Workers complete jobs faster than main can track, causing executions to leak

### Fix

Move the save decision to the worker level via a new `handleExecutionSaveDecision()` function. The worker now decides whether to save or delete at execution time, eliminating the distributed coordination problem.

### Testing

- Added regression test at `tests/infrastructure/regression/CAT-1437-execution-save-race.spec.ts`
- Fires 500 concurrent requests against a workflow configured not to save
- Verifies 0 executions leak into the database
- Uses 1 main + 3 workers to reproduce the original failure scenario

**Reference:** https://internal.users.n8n.cloud/workflow/284/executions/39834796

https://linear.app/n8n/issue/CAT-1437

## Review / Merge checklist

- [x] Tests included.


shortstacked:
@claude

ivov (changes requested):
Quick pass. Sorry for the delay!

Open to improving this, but I'm not confident about this change.

1. What is the root cause here? The explanation "workers completing jobs faster than main could track via `job.finished()`" does not make sense to me because `job.finished()` is pubsub-based and if Redis were degraded, we'd see much bigger problems across the board. Maybe notifications arriving while main is down? Or notifications arriving but main's DB deletes failing?
2. If the worker deletes the execution, then when we [check for execution errors](https://github.com/n8n-io/n8n/blob/bfb11c0e2a3d8c4942712cf92d451c16971197b5/packages/cli/src/scaling/job-processor.ts#L252) we won't find any for an erroring execution that we deleted, because we bail early if there's no execution to be found, so we'll misreport it as successful.
3. If the worker deletes the execution, when main [queries for run data](https://github.com/n8n-io/n8n/blob/bfb11c0e2a3d8c4942712cf92d451c16971197b5/packages/cli/src/workflow-runner.ts#L462) there'll be nothing to find for the execution, so we'll error out before we get a chance to [finalize the execution and run main's hooks on it](https://github.com/n8n-io/n8n/blob/bfb11c0e2a3d8c4942712cf92d451c16971197b5/packages/cli/src/workflow-runner.ts#L480-L484) and the promise rejection from that will be [silently swallowed](https://github.com/n8n-io/n8n/blob/bfb11c0e2a3d8c4942712cf92d451c16971197b5/packages/cli/src/workflow-runner.ts#L490-L494).
4. Assuming the explanation is accurate and pubsub is overwhelmed, what's the rationale for keeping the saving decision on both main and workers? 
5. This old PR #1509 originally moved the saving decision from workers to main, the inverse of this PR. Let's double check we aren't undoing what that PR was fixing back then?

PR #23763, opened 2025-12-31
feat(core): Adjust Workflow Compaction History algorithm (no-changelog)

CharlieKolb:
## Summary

Adjust to some more sensible defaults and improve the rules a bit

## Review / Merge checklist

- [ ] Tests included. 


packages/workflow/src/workflow-diff.ts
```diff
@@ -145,8 +146,10 @@ function nodeIsAdditive<T extends DiffableNode>(prevNode: T, nextNode: T) {
 	if (!compareNodes({ ...prev, parameters: {} }, { ...next, parameters: {} })) return false;
 
 	const params = Object.keys(prevParams);
-	// abort if prev has some field next does not have
+	// abort if fields differ
 	if (params.some((x) => !Object.prototype.hasOwnProperty.call(nextParams, x))) return false;
+	if (Object.keys(nextParams).some((x) => !Object.prototype.hasOwnProperty.call(prevParams, x)))
```
scdekov, line 159: Seems like this function is no more checking for "isAdditive", right?

packages/workflow/src/workflow-diff.ts
```diff
@@ -179,10 +182,23 @@ function mergeAdditiveChanges<N extends DiffableNode = DiffableNode>(
 	return true;
 }
 
+function skipTimeDifference<N extends DiffableNode = DiffableNode>(
+	prev: GroupedWorkflowHistory<DiffableWorkflow<N>>,
+	next: GroupedWorkflowHistory<DiffableWorkflow<N>>,
+) {
+	const timeDifference = next.to.createdAt.getTime() - prev.from.createdAt.getTime();
+
+	return Math.abs(timeDifference) > 30 * 60 * 1000;
```
scdekov: Should this be using `compactingMinimumAgeHours`? Can we add a comment on top of this function explaining why are we doing this?

packages/workflow/src/workflow-diff.ts
CharlieKolb: It does - without this change we treated a change away from a default value as additive, whereas it's actually a modification as it'll be a different value - especially relevant for operations and resources.

packages/workflow/src/workflow-diff.ts
CharlieKolb: No, this is an unrelated value. I could expose it as a config value if I define the rule in the cli code if you prefer that. This value here is how much time needs to pass between two versions before we won't merge them anymore, because we're treating them as different "sessions".

Admittedly if the value is larger than `compactingMinimumAgeHours` we will never invoke this rule in the current setup.

I'll add some more comments tho!

Also this reminds me I wanted to add a rule to never merge versions from different users! Will do so now

packages/workflow/src/workflow-diff.ts
scdekov: But also adding new props will now say it's not additive where it could be, no?

packages/workflow/src/workflow-diff.ts
scdekov: Sounds good. I don't think we have to expose that as config but having that comment in the code would definitely be useful.

packages/workflow/src/workflow-diff.ts
CharlieKolb: I guess the name is a bit misleading, more accurately we're checking for whether we're losing any information at all if we merge `prev` into `next`. So `additive` as in information, not nodes/parameters

scdekov: Got it. Maybe again a comment either above that condition or updating the comment above the function itself may be useful

scdekov (dismissed):

scdekov (approved):

PR #23768, opened 2025-12-31
ci: Adjusts playwright test shards and command

shortstacked:
## Summary

Problem: The isolated tests were running at the same time, causing DB resets to trigger while other tests were still running. It always happened in the same test as it was the last in line.

Modifies the Playwright test CI workflow to use an e2e-only test command.
Increases the number of shards for local e2e tests to improve test execution time. Also, increases shards for isolated tests.

## Review / Merge checklist

- [ ] Tests included. 


ShireenMissi (approved):
LGTM ✅

PR #23850, opened 2026-01-05
fix(AI Agent Node): Normalize toolInput to object for LangChain ToolCall compatibility (no-changelog)

OlegIvaniv:
## Summary
Fixes TypeScript error in `memoryManagement.ts` where `toolInput` type was incompatible with LangChain's `ToolCall` which requires `args` to be `Record<string, any>`.

The fix ensures `toolInput` is always an object by removing the string unwrapping logic in `buildSteps.ts` that was extracting string values from `{ input: "string" }` format.

## Changes
- Simplified `buildSteps.ts` to always keep `toolInput` as object
- Updated `ToolCallData.toolInput` type from `Record<string, unknown> | string` to `Record<string, unknown>`
- Updated tests to expect object format

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/utils/agent-execution/buildSteps.ts
+++ b/packages/@n8n/nodes-langchain/utils/agent-execution/buildSteps.ts
@@ -231,14 +231,8 @@ export function buildSteps(
 			});
 
 			// Extract tool input arguments for the result
-			let toolInputForResult: IDataObject | string;
-			if (toolInput.input && typeof toolInput.input === 'string') {
-				toolInputForResult = toolInput.input;
-			} else {
-				// Exclude metadata fields: id, log, type
-				const { id, log, type, ...actualInput } = toolInput;
-				toolInputForResult = actualInput;
-			}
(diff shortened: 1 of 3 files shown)


aalises:
That's great! Narrowing types sounds always good, not having to worry if "is it a string or an object..."? Did this create some of the models to crash expecting an object?

OlegIvaniv:
> That's great! Narrowing types sounds always good, not having to worry if "is it a string or an object..."? Did this create some of the models to crash expecting an object?

No crashes - the change is backwards compatible. Previously, when input was { input: "2+2" }, we were unwrapping it to just "2+2". Now we keep it as { input: "2+2" }. LangChain tools that accept a single string argument define their schema with an input property, so { input: "2+2" } is the correct format they expect. The string unwrapping was actually a deviation that could cause issues downstream (like the type error in memoryManagement.ts where LangChain's ToolCall.args requires Record<string, any>)

mutdmour (approved):

PR #23867, opened 2026-01-05
feat(core): Merge HITL and tool schemas

yehorkardash:
## Summary

This PR adds support of `fromAI` expression inside HITL tools by merging tool schemas when wrapping tools into toolkit.
The PR also removes `toolName` property and leaves just `tool`

https://github.com/user-attachments/assets/e9f31c8b-ccaa-47a1-a7a1-fea511b9c86a

https://linear.app/n8n/issue/NODE-4100/fromai-is-not-functioning-inside-hitl-tool

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/nodes-langchain/utils/agent-execution/createEngineRequests.ts
```diff
@@ -104,7 +104,8 @@ export async function createEngineRequests(
 				input = { ...input, tool: toolCall.tool };
 			}
 			if (hitlMetadata) {
-				input = { ...input, toolName: toolCall.tool };
+				// This input will be used as HITL node input
+				input = { ...input, ...(input.hitlParameters as IDataObject) };
```
dlavrenuek, line 108: What happens if HITL and a tool use the same parameters? It looks like the tool parameters will be overridden by HITL params.

yehorkardash: HITL parameters exist inside `hiltParameters` and tool parameters exist inside `toolParameters`. For a HITL node the input will be like that
```
{
  // ...someFields,
  hitlParameters: { parameter1: 'hitl' }
  toolParameters: { parameter1: 'tool' }
  parameter1: 'hitl'
}
```
We unnest hitl parameters, because the HITL node uses `fromAI`, which can't access nested parameters like `hitlParameters.parameter1` selectively

A data that will be passed to a tool, is defined in `extractHitlMetadata`. It only uses `toolParameters`

dlavrenuek (approved):
Works as expected and code looks good 🚀 

Not related to changes here but in general to HITL implementation: while testing with azure open ai I've hit the error every time:

> Bad request - please check your parameters
Invalid 'messages[2].tool_calls[0].id': string too long. Expected a string with maximum length 40, but got a string with length 43 instead.


This is due the tool call id being extended [here](https://github.com/n8n-io/n8n/blob/node-3883-generate-hitl-tool-nodes-for-sendandwait-operations/packages/%40n8n/nodes-langchain/utils/agent-execution/processHitlResponses.ts#L113), which exceeds the size limit of 40 characters for azure open ai models. I suggest generating a new id instead of extending the existing one, so (I assume) langchain takes maximum tool call id size into account.

Google says:

> The toolCallId length limit is provider-specific and varies significantly between different Language Model (LLM) providers. 
OpenAI/Azure: The limit is typically 40 characters. IDs generated by these services are often UUIDs (36 characters) which fit this limit.
Mistral: Tool call IDs must be alphanumeric and exactly 9 characters long.
Other Providers: The length can range from around 40 characters to potentially longer limits, such as a 1024-character limit mentioned in an OpenAI community forum for an API upgrade (likely for other properties, not the ID itself). 
If you are generating your own toolCallId in an application (e.g., when using frameworks like LangChain, LlamaIndex, or [AI SDK](https://ai-sdk.dev/)), it is crucial to ensure the IDs you create or truncate comply with the specific length and format requirements of the target LLM API to avoid "Bad Request" errors. 
The limit for the tool_outputs (the results returned using the ID) can be much larger, with discussions mentioning limits around 1024 characters or more for the description/output content.

packages/@n8n/nodes-langchain/utils/agent-execution/createEngineRequests.ts
dlavrenuek: Tested and works as expected

PR #23889, opened 2026-01-05
fix(DataTable Node): Use extractValue to properly resolve data table ID

sohamingle:
## Summary
Fixes the `resolveDataTableId` function in the DataTable node to properly extract the data table ID when using 'list' or 'id' modes in the resource locator.

The function now uses `ctx.getNodeParameter` with the `extractValue: true` option instead of directly returning `resourceLocator.value`. This ensures the ID is correctly extracted from the resource locator regardless of the internal structure.

### Changes
- Added `index` parameter to `resolveDataTableId` function
- For 'list' and 'id' modes, use `ctx.getNodeParameter` with `extractValue: true` to properly resolve the data table ID
- Updated `getDataTableProxyExecute` to pass the index parameter

#### Before
image

#### After
image

https://linear.app/n8n/issue/GHC-6167
Fixes #23884

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/DataTable/common/utils.ts
+++ b/packages/nodes-base/nodes/DataTable/common/utils.ts
@@ -28,7 +28,6 @@ function isDateLike(v: unknown): v is DateLike {
 async function resolveDataTableId(
 	ctx: IExecuteFunctions | ILoadOptionsFunctions,
 	resourceLocator: { mode: 'list' | 'id' | 'name'; value: string },
-	index?: number,
 ): Promise<string> {
 	if (resourceLocator.mode === 'name') {
 		// Look up table by name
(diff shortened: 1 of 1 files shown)


salzig:
Looks like this got "reimplemented" and merged with https://github.com/n8n-io/n8n/pull/23936

Joffcom:
Thanks for the fix! Before this can be reviewed, please add a test that exercises `resolveDataTableId` with the 'list' and 'id' resource locator modes (e.g. an entry in `packages/nodes-base/nodes/DataTable/**/__tests__/` or a sibling `*.test.ts`) so we can confirm the regression stays fixed. Also worth double-checking the PR description against the diff — the description says the function now uses `ctx.getNodeParameter(..., { extractValue: true })`, but the diff actually removes that branch and returns `resourceLocator.value` directly. Could you update the summary so it matches what the code does?

Issue #23782, opened 2026-01-01
Bug Report: Slack Trigger "Usernames to Ignore" fails on message_changed events


dor-justt:
### Bug Description

The Slack Trigger node (v2.2 and v2.4) fails to ignore messages sent by specific User IDs when the event is of subtype message_changed. This occurs because Slack asynchronously "enriches" messages (e.g., adding locale metadata for non-English languages), which moves the sender's identity into a nested message object. The trigger node's internal filter appears to only check the root-level user field, which is often empty in these events, causing an infinite loop.

### To Reproduce

Create a Slack Trigger node and add a Bot User ID (e.g., U0A5AGQU6LC) to the Usernames or IDs to Ignore field.

Create a workflow that responds to a message in Slack.

Send a message in a language that triggers Slack's asynchronous language detection (e.g., Hebrew).

Observe that the bot sends the initial message (which is ignored correctly), but Slack immediately sends a follow-up message_changed event to update the locale.

The Slack Trigger node fires on this "change" event despite the ID being in the ignore list, resulting in an infinite loop.

### Expected behavior

The node should inspect both the root user field and the nested message.user field (if it exists) before deciding whether to trigger the workflow.
Note: Root-level user field is missing, while message.user contains the ID to be ignored.

### Debug Info

# Debug info

## core

- n8nVersion: 2.0.3
- platform: docker (self-hosted)
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- license: enterprise (production)
- consumerId: 4683eba3-67ce-4873-bcde-8a0658745d85

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/143.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-01-01T21:46:40.971Z

### Operating System

Linux

### n8n Version

2.0.3

### Node.js Version

22.16

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

PR #23894, opened 2026-01-06
fix(Slack Trigger Node): Fix user ignore filter for message_changed events

mag123c:
## Summary
Fix `Usernames or IDs to Ignore` filter not working for message_changed events.

When Slack asynchronously enriches messages (e.g., adding locale metadata for non-English languages), it sends a message_changed event where the user ID is in event.message.user instead of event.user. The ignore filter was only checking event.user, causing the filter to fail and resulting in infinite loops.

Now checks both event.user and event.message.user (as fallback).

Related Linear tickets, Github issues, and Community forum posts

Issue: #23782
## Review / Merge checklist

- [x] Tests included.


packages/nodes-base/nodes/Slack/SlackTrigger.node.ts
```diff
@@ -388,9 +388,11 @@ export class SlackTrigger implements INodeType {
 		}
 
 		// Check if user should be ignored
+		// For message_changed events, user ID is in event.message.user instead of event.user
 		if (options.userIds) {
 			const userIds = options.userIds as string[];
-			if (userIds.includes(req.body.event.user ?? req.body.event.message?.user)) {
+			const userId = req.body.event.user ?? req.body.event.message?.user;
```
garritfra, line 394: Nit: I think a cleaner solution would be to map the field based on the event type, but since it's not done anywhere else in the node, this is outside the scope of this PR 😄

garritfra (approved):
LGTM, thanks!

mag123c:
@garritfra 
The same fix landed in #26414 which also closed #23782. The remaining delta here is a no-op truthiness guard and a test set that's a subset of what's already on master. Thanks for the review.

PR #23898, opened 2026-01-06
fix(editor): Make AI assistant tool steps visible during streaming

aalises:
## Summary

- Show AI assistant steps by default: `ThinkingMessage` is now expanded on render, making tool call steps visible immediately rather than hidden behind a collapsed header
- Auto-collapse on completion: Steps automatically collapse when streaming finishes (or is aborted), keeping the final view clean while allowing users to see progress during generation
- Remove placeholder: Replaced `AskAssistantLoadingMessage` with a `ThinkingMessage` with an initial message. This makes it more consistent and also allows for optimistic updates
- Removed ellipsis: Changed "Thinking..." to "Thinking" for cleaner copy, and "Processing..." to Processing. Moved all copy to i18n.
- Fixed some race conditions where "Workflow generated" appeared for a second when the next tool wasn't yet processed

## Technical Details

  - Added `defaultExpanded` prop to `ThinkingMessage` with watcher to collapse when `isStreaming` transitions from true to false (default not expanded for it to be backwards compatible)
  - Removed the unused `AskAssistantLoadingMessage` component and its associated stories/tests, not used anymore
  - Created `ThinkingMessage` with an initial fake message as a minimal component for pre-tool-call state

## Test Plan

  - Verify thinking steps are visible during AI streaming
  - Confirm steps auto-collapse when generation completes
  - Check placeholder appears before first tool call arrives
  
  ## Demo
  
Kapture 2026-01-06 at 09 43 46

## Possible improvements

The thinking message does not automatically collapse on intermediate steps when finished (e.g. a thinking block, followed by a message, then another thinking block, the first one won't collapse). On finishing the workflow, all of them will collapse.

  ---
  Related issue: https://linear.app/n8n/issue/AI-1806


burivuhster (approved):
Looks good! 🚀
Congrats with the first PR 🎉

mike12345567 (approved):

PR #5235, opened 2023-01-23
feat(Salesforce Node): Add HasOptedOutOfEmail field to lead resource

RicardoE105:
https://community.n8n.io/t/there-is-no-email-opt-out-field-in-the-saleforce-create-lead/18616/9

PR #23905, opened 2026-01-06
feat(Salesforce Node): Add middle name, suffix, and identity fields to contacts

Gulianrdgd:
## Summary
In the Netherlands middle names are very common and I noticed this field was missing in the contact field from the Salesforce node. I basically added the other fields which are in the [Salesforce API](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contact.htm) but were missing in the node.

I only changed the Contact field, I did check some others but I am not so experienced in Sales to know what these fields do so I left them.

The docs did not need to be updated

None

## Review / Merge checklist

- [x] Tests included. 


colons-litzmann:
Hello @Gulianrdgd !

I too stumbled upon one missing field in the Contact Object (https://github.com/n8n-io/n8n/pull/28259) and think your effort to include ALL of the fields that are present in the Salesforce API is better than my approach of just adding the one field I was missing.

I noticed however that you do not touch the `packages/nodes-base/nodes/Salesforce/Salesforce.node.ts` Node itself, just the Interface and the Description. I am new to the Codebase myself but I feel like the missing fields need to be added there too if you want to set them during create / update / upsert workflows. Maybe its worth taking a look! I literally copied the changes I made from https://github.com/n8n-io/n8n/pull/5235 which was made by a n8n team member and then just added the fields to tests and schemas to fit how they looked for the "Lead" Object.

Gulianrdgd:
@colons-litzmann Good catch. I could have sworn I tested it, but I have added it just to be sure. Sorry I wrote this a long time ago.

Gulianrdgd:
@yehorkardash Are you supposed to do a review? Or is github bugging out? 😃

yehorkardash:
Hey @Gulianrdgd, thanks for your contribution, sorry for dragging out the review 🙏
Will get to review it this week

packages/nodes-base/nodes/Salesforce/ContactDescription.ts
```diff
@@ -398,6 +412,159 @@ export const contactFields: INodeProperties[] = [
...
+				default: '',
+				description:
+					'ID of the data privacy record associated with this contact (for Data Protection and Privacy)',
+			},
+			{
+				displayName: 'Pronouns',
+				name: 'pronouns',
+				type: 'options',
+				default: '',
+				options: [
```
yehorkardash: I believe this field can be customized, so better to keep it as string parameter

packages/nodes-base/nodes/Salesforce/ContactDescription.ts
```diff
@@ -398,6 +412,159 @@ export const contactFields: INodeProperties[] = [
...
+				name: 'genderIdentity',
+				type: 'string',
+				default: '',
+				description: "The contact's gender identity",
+			},
+			{
+				displayName: 'Mailing Latitude',
+				name: 'mailingLatitude',
+				type: 'number',
+				default: undefined,
```
yehorkardash: Number fields using `undefined` cannot be added as a parameter in the options UI. Use `''` or `0` as default value

packages/nodes-base/nodes/Salesforce/ContactDescription.ts
```diff
@@ -705,6 +886,159 @@ export const contactFields: INodeProperties[] = [
...
+				default: '',
+				description:
+					'ID of the data privacy record associated with this contact (for Data Protection and Privacy)',
+			},
+			{
+				displayName: 'Pronouns',
+				name: 'pronouns',
+				type: 'options',
+				default: '',
+				options: [
```
yehorkardash: this option is duplicated

yehorkardash (commented):
Hey @Gulianrdgd! Thanks for the changes
Changes to MiddleName, Suffix, and identity look good, but I left some minor comments
As for other fields I suggest that you remove them to keep the scope of the PR small. These fields can also be added using "Custom Fields" option


EDIT: Oh, I see there has also been a mention of "Has Opted Out of Email" in  https://github.com/n8n-io/n8n/pull/28259, let's keep it too

colons-litzmann:
BTW @Gulianrdgd - unrelated to the MR but maybe worth noting as a workaround: You can bypass this Restriction when using "Custom Object":

image

Just hardcode the Object name as an expression as seen above. Now it shows me all the fields:

image

I noticed this because it seems n8n is completely missing the Order Object in the list of "default" list of Resources (!) but it also does not appear in the custom object list (!!) - so when using the SF node as intended, updating an Order Object is just not there. (Maybe I am holding it wrong, might be!) I then tried hardcoding the name of the Object and that works just fine, as n8n is evidently capable of programmatically getting all the Fields that any given Object has.

colons-litzmann:
> These fields can also be added using "Custom Fields" option

Actually thats only partially true. It seems that when dealing with Resources that the Salesforce node treats as default, because they exist by default in any Salesforce environment (such as Accounts, Contacts, Leads, Cases etc.) only the fields that are created by the Salesforce User themselves get shown in the custom fields dropdown. All the fields in this PR are part of the default set of fields (see [here](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contact.htm)) and therefore are not pickable in the custom fields menu, but they also dont appear in the list of non-custom fields, because these are not yet hardcoded into n8n. 

I found out today you can just hardcode the keys of the fields in a custom field (only when saving the field as an expression!). N8N will then tell you that this value is not supported, but it will work. I consider this more of a workaround than a feature though.

See 
image

In other words: Fields that are Salesforce defaults, but are not hardcoded in n8n are not pickable via UI, because fields that are not actually custom dont  show up in the 'custom fields' picker

The same seems to be true for the Order Object as a whole (see my other comment). It is **NOT** a *custom* object and therefore does not appear in the list of custom objects. But it is also not hardcoded into n8n, so therefore it just does not appear in the UI at all. If you just write its name it into the 'custom object name' field it will warn you but then work like a charm; even grabbing all the fields for you! It seems as if this is a systemic problem with the salesforce node's *UI*. The actual functionality is working just fine!

packages/nodes-base/nodes/Salesforce/ContactDescription.ts
colons-litzmann: Can confirm. Salesforce Docs mention default values, but I can easily update the List for the Object

yehorkardash:
@colons-litzmann, thanks for noticing the issue 🙌

The parameter name validation can be fixed by adding `allowArbitraryValues: true` to parameter or by updating the field to list both custom and default parameters, something we've missed at the time of creation of this node. I can file a ticket or if you implement a fix, I'll be happy to review it!

For this PR specifically, I'd keep only the 3-4 new fields and then come up with a better solution for listing parameters, because I didn't find fields like `Other Longitude` to be listed/enabled by default in salesforce

colons-litzmann:
@yehorkardash Thanks for the additional insight! I didnt know about `allowArbitraryValues` yet, seems like its rarely used. Always showing all objects sounds a bit more complex though.

> I can file a ticket or if you implement a fix, I'll be happy to review it!

I will have a look next week and let you know if I can do it. I am new to n8n but its a learning opprtunity. Regarding the other fields I personally agree with your points now, but in the end its @Gulianrdgd 's PR :D

Gulianrdgd:
> BTW @Gulianrdgd - unrelated to the MR but maybe worth noting as a workaround: You can bypass this Restriction when using "Custom Object":
> image
> 
> Just hardcode the Object name as an expression as seen above. Now it shows me all the fields:
> image
> 
> I noticed this because it seems n8n is completely missing the Order Object in the list of "default" list of Resources (!) but it also does not appear in the custom object list (!!) - so when using the SF node as intended, updating an Order Object is just not there. (Maybe I am holding it wrong, might be!) I then tried hardcoding the name of the Object and that works just fine, as n8n is evidently capable of programmatically getting all the Fields that any given Object has.

Cool I tried the custom object but for some reason it did not really work. Thanks for the info!

Gulianrdgd:
> @colons-litzmann, thanks for noticing the issue 🙌
> 
> The parameter name validation can be fixed by adding `allowArbitraryValues: true` to parameter or by updating the field to list both custom and default parameters, something we've missed at the time of creation of this node. I can file a ticket or if you implement a fix, I'll be happy to review it!
> 
> For this PR specifically, I'd keep only the 3-4 new fields and then come up with a better solution for listing parameters, because I didn't find fields like `Other Longitude` to be listed/enabled by default in salesforce

Yeah, I was unsure what to do with all the fields. Because the custom field did not work for me, I simply copied all the fields from the Salesforce docs that I could find. Some fields were very weird, but I can imagine that they are useful to some. 

I have reduced them now to: MiddleName, Suffix, HasOptedOutOfEmail, Pronouns, GenderIdentity. I left in the email one because that one seemed reasonable to have in the default list. If you disagree, I can still remove it.

yehorkardash (commented):
@Gulianrdgd 
Just two final comments and the PR is good to go 🚀

Could you extend `should handle contact create with all additional fields` and `should handle contact update with all fields` to assert the new field mappings?

Could you reorder the new fields next to thematically related ones (group Email Opt Out with the bounce/email block; group Pronouns / Gender Identity near Salutation)?

Gulianrdgd:
@yehorkardash Got it! All good now?

yehorkardash (approved):

PR #23914, opened 2026-01-06
chore(ai-builder): Support filtering by do/don't criteria in pairwise evals

mike12345567:
## Summary

Adding a command line option for filtering on do/dont criteria of pairwise evals which allows filtering for a specific statement within the criteria. For ticket https://linear.app/n8n/issue/AI-1801/feature-make-builder-use-structured-output-parser I wanted to run any spec-wise evals which involved checking for a structured output parser.

I have refactored how we supply filters, previously we could specify `--technique` and `--notion-id`, but I think we'll need to add more filters as running the entire suite is slow and expensive, better ways to filter down to what you really are testing is necessary. The pairwise evals now have a `--filter` argument which takes a key value pair, like `--filter "do:structured output"` or `--filter "technique:data_transformation"`.

I've kept the old `--technique` and `--notion-id` to avoid needing to update these, but they are easy to remove once everyone is aware of the new way of specifying filters.

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/evaluations/index.ts
```diff
@@ -22,6 +22,8 @@ const VALID_FLAGS = [
 	'--repetitions',
 	'--notion-id',
 	'--technique',
+	'--do',
+	'--dont',
```
OlegIvaniv: Could we be more specific with these? We already have:
```
	'--dos',
	'--donts',
```

For when running it locally with `--prompt`

mike12345567: Ah - I didn't realise there was an option for providing a prompt/individual do/donts - I'll have a think about how to update the filtering parameters to not overlap.

mike12345567: @OlegIvaniv I've addressed this, would you mind perhaps reviewing this PR since you've taken a look at it already?

OlegIvaniv (approved):
LGTM!

PR #23928, opened 2026-01-06
chore: Move codemirror-lang-html-n8n into monorepo

elsmr:
## Summary

- Move [`codemirror-lang-html-n8n`](https://github.com/ivov/codemirror-lang-html-n8n) into the n8n monorepo as `@n8n/codemirror-lang-html`
- Updates `@n8n/codemirror-lang` and `editor-ui` to use the new internal package
- Adds pnpm catalog entries for CodeMirror dependencies

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/codemirror-lang/package.json
```diff
@@ -2,8 +2,18 @@
   "name": "@n8n/codemirror-lang",
   "version": "0.3.0",
   "description": "Language support package for CodeMirror 6 in n8n",
-  "private": true,
```
elsmr, line 5: I think this was invalid because [`@n8n/codemirror-lang-sql`](https://www.npmjs.com/package/@n8n/codemirror-lang-sql?activeTab=code) is published with `@n8n/codemirror-lang` as a dependency, but `@n8n/codemirror-lang` itself is not on npm.

.github/workflows/release-standalone-package.yml
```diff
@@ -8,10 +8,13 @@ on:
         required: true
         type: choice
         options:
-          - '@n8n/node-cli'
+          - '@n8n/codemirror-lang'
+          - '@n8n/codemirror-lang-html'
+          - '@n8n/codemirror-lang-sql'
           - '@n8n/create-node'
```
ivov, line 14: For my understanding, why do we want to externally release CM packages?

ivov (approved):
Thank you for doing this 🙏🏻 

Gave it a spin and it works, trusting the content was moved but otherwise untouched

.github/workflows/release-standalone-package.yml
elsmr: We don't have to, but I saw [`@n8n/codemirror-lang-sql`](https://www.npmjs.com/package/@n8n/codemirror-lang-sql?activeTab=versions) was published already. I figured if a package could be useful outside of n8n, why not publish it 😄 

This workflow just allows publishing these packages on demand, but `@n8n/codemirror-lang-sql` is already being published by the normal weekly release workflows.

PR #23929, opened 2026-01-06
feat: Add new api for generating invite links

BGZStephen:
## Summary

This PR creates a new POST API endpoint `/rest/users/:id/invite-link` to support generating signed invite links with expirations to provide a tamper proof mechanism for inviting and accepting invites 

PAY-4390

## Review / Merge checklist

- [x] Tests included. 


geemanjs (approved):
This a clever solution to that bug. I quite like it! 

1. Please tag me on the other side of the PR where we validate and check the token 😄...

From a security point of view I have some more questions:
1. Should it be possible to revoke an invitation? This feels like something we should be able to support
2. Is this safe from a replay attack? i.e. someone reusing the same token multiple times? JWT's are stateless after-all.

If we wanted some added protection (this prevents both above) an option might be to create a unique `inviteId`, drop that into the db with a state `pending` and put it on the JWT also.

```
| id         | state    | createdAt | lastUpdatedAt | 
| <inviteId> | pending  |           |               |
| <inviteId> | complete |           |               |
| <inviteId> | revoked  |           |               |
```

When they signup we extract the `inviteId` from the JWT and check it's state. If we want to revoke the invite we can manage that on the `inviteId` table. Alternatively, we can delete the `inviteId` from the table once it's complete/revoked - that probably scales better (and we can extend to emit events if someone wants the extra information in a SIEM)

3. Should we maybe pass the `expiresIn` through the body of the controller to let the user configure how long the link is valid for when they create the link. Default it to a sensible 7 days in an input box. Otherwise, I fear in 6 months time we'll get a request to reduce from 90d to something shorter and/or be able to configure it via an env var or something 👀

packages/@n8n/permissions/src/roles/scopes/global-scopes.ee.ts
```diff
@@ -56,6 +56,7 @@ export const GLOBAL_OWNER_SCOPES: Scope[] = [
 	'user:delete',
 	'user:list',
 	'user:resetPassword',
+	'user:generateInviteLink',
```
guillaumejacquart, line 59: Should we also add this scope to all custom roles with user:create scope (the current role used for invitation) in a migration file ?

packages/cli/src/controllers/users.controller.ts
```diff
@@ -167,6 +172,34 @@ export class UsersController {
 		return { link };
 	}
 
+	@Post('/:id/invite-link')
+	@GlobalScope('user:generateInviteLink')
+	async generateInviteLink(req: AuthenticatedRequest<{ id: string }, {}, {}, {}>, _res: Response) {
+		const inviterId = req.user.id;
+		const inviteeId = req.params.id;
+
+		const targetUser = await this.userRepository.findOne({ where: { id: inviteeId } });
```
guillaumejacquart, line 181: Why not do all that in the users service ? It already has urlService dependency, so would only need the new jwtService dependency.

guillaumejacquart (commented):
Looks good ! 2 small comments. Also I'm wondering, we will replace the invitation controller route with this one eventually ? sending the invite by email as well (eventually) ?

BGZStephen:
> This a clever solution to that bug. I quite like it!
> 
> 1. Please tag me on the other side of the PR where we validate and check the token 😄...
> 
> From a security point of view I have some more questions:
> 
> 1. Should it be possible to revoke an invitation? This feels like something we should be able to support
> 2. Is this safe from a replay attack? i.e. someone reusing the same token multiple times? JWT's are stateless after-all.
> 
> If we wanted some added protection (this prevents both above) an option might be to create a unique `inviteId`, drop that into the db with a state `pending` and put it on the JWT also.
> 
> ```
> | id         | state    | createdAt | lastUpdatedAt | 
> | <inviteId> | pending  |           |               |
> | <inviteId> | complete |           |               |
> | <inviteId> | revoked  |           |               |
> ```
> 
> When they signup we extract the `inviteId` from the JWT and check it's state. If we want to revoke the invite we can manage that on the `inviteId` table. Alternatively, we can delete the `inviteId` from the table once it's complete/revoked - that probably scales better (and we can extend to emit events if someone wants the extra information in a SIEM)
> 
> 3. Should we maybe pass the `expiresIn` through the body of the controller to let the user configure how long the link is valid for when they create the link. Default it to a sensible 7 days in an input box. Otherwise, I fear in 6 months time we'll get a request to reduce from 90d to something shorter and/or be able to configure it via an env var or something 👀

> Should it be possible to revoke an invitation

Currently no, it's not functionality that exists at the moment, but might be nice to add in the future.

> 2. Is this safe from a replay attack? i.e. someone reusing the same token multiple times? JWT's are stateless after-all.

Yes, as once the user has used this token and provided a password, the signup flow will no longer be accessible, we currently check if the user has completed the sign up flow to avoid this.

> If we wanted some added protection (this prevents both above) an option might be to create a unique `inviteId`, drop that into the db with a state `pending` and put it on the JWT also.

We had preferred not adding or modifying the DB mostly due to the hassle of migrations, in an ideal world we would have a table for invites, but it's not strictly necessary at the moment

> 3. Should we maybe pass the `expiresIn` through the body of the controller to let the user configure how long the link is valid for when they create the link. Default it to a sensible 7 days in an input box. Otherwise, I fear in 6 months time we'll get a request to reduce from 90d to something shorter and/or be able to configure it via an env var or something 👀

I think it's OK to just have a default (at the moment there is no expiry, so this moves us in the right direction) if there is appetite from a customer perspective, we can always add this later

packages/@n8n/permissions/src/roles/scopes/global-scopes.ee.ts
BGZStephen: Let me add this to the list of tickets, will do this in a follow up once the rest of the code is in place

packages/cli/src/controllers/users.controller.ts
BGZStephen: Makes sense, I'll move this as part of the other tickets in this feature

PR #23930, opened 2026-01-06
fix(editor): Keep structured output parser error icon in correct position when running

mike12345567:
## Summary

This PR fixes an issue with the structured output parser (and likely other similarly styled nodes) where the error icon would move around while the running animation was active. I've made the location of the icon static consistently now, rather than excluding it while it is running.

structured-output-animation

Addresses: https://linear.app/n8n/issue/AI-1807/bug-structured-output-parser-animation-causes-state-icon-to-move

## Review / Merge checklist

- [ ] Tests included. 


aalises (approved):
Great! Maybe that margin was added back then because there was some other different element appearing when running/not running that has since been deleted but the margin kept...

OlegIvaniv (approved):

PR #23952, opened 2026-01-07
feat: Add banner enhancemenets

seemewalkin:
## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/executions/execution-recovery.service.ts
```diff
@@ -68,6 +68,17 @@ export class ExecutionRecoveryService {
 
 				if (workflow.activeVersionId !== null) {
 					await this.workflowRepository.updateActiveState(workflowId, false);
+
+					// Store auto-deactivation metadata
+					const autoDeactivatedAt = new Date().toISOString();
```
endor, line 73: I remember we explicitly left this out of our last PR due to concerns over database activity. @ivov is this still a concern?

packages/frontend/editor-ui/src/app/components/WorkflowCard.vue
```diff
@@ -270,6 +270,10 @@ const isWorkflowPublished = computed(() => {
 	return props.data.activeVersionId !== null;
 });
 
+const isAutoDeactivated = computed(() => {
+	return !!props.data.meta?.autoDeactivatedAt && !props.data.active;
```
endor, line 274: Should we use `isWorkflowPublished` instead of checking `props.data.active` here?

packages/cli/src/executions/execution-recovery.service.ts
```diff
@@ -68,6 +68,17 @@ export class ExecutionRecoveryService {
...
+					await this.workflowRepository.update(workflowId, {
```
endor, line 74: I also wonder, should we have any sort of telemetry about how (often) this feature is used?

PR #23962, opened 2026-01-07
feat: Use new generated links within the UI and emails if feature is enabled

BGZStephen:
## Summary

Adds a feature flag gated implementation of tamper proof invite links for the UI > Settings > Users page where users are invited. Also updates the backend for email configured instances when sending out emails.

PAY-4394

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/services/user.service.ts
```diff
@@ -153,9 +159,39 @@ export class UserService {
...
+							inviteeId: id,
+						},
+						{
+							expiresIn: '90d',
+						},
+					);
+					inviteAcceptUrl = `${domain}/signup?token=${token}`;
+				} else {
+					// Use legacy invite links when feature flag is disabled
+					inviteAcceptUrl = `${domain}/signup?inviterId=${owner.id}&inviteeId=${id}`;
```
phyllis-noester, line 193: so the issue was that previously with brute force you could find a working invite link and add yourself to a random org but now because we sign it with JWT we can validate that the link was actually signed correctly and is therefore safe?

BGZStephen: Exactly, you can no longer just provide an arbitrary inviteeId as you'd need a method of signing the token to spoof that it came from the back-end

phyllis-noester: noice! thanks :)

phyllis-noester: PR LGTM but let's rather wait for someone with more context to approve ;)

packages/cli/src/services/user.service.ts
```diff
@@ -153,9 +159,39 @@ export class UserService {
 
 		const inviteLinksEmailOnly = this.globalConfig.userManagement.inviteLinksEmailOnly;
 
+		// Check if tamper-proof invite links feature flag is enabled for the owner
+		let useTamperProofLinks = false;
+		try {
+			const featureFlags = await this.postHog.getFeatureFlags({
+				id: owner.id,
```
guillaumejacquart, line 166: It feels a bit weird that the owner id is used for targeting the flag. Especially since here I think owner is actually the user making the invite request, meaning any user with user:create scope. So if we apply some targeting on posthog on the user id, different admin users could trigger different experiences.
Why not using the instance id instead ?

packages/frontend/editor-ui/src/features/settings/users/components/InviteUsersModal.vue
```diff
@@ -245,10 +263,24 @@ function onSubmitClick() {
...
-function onCopyInviteLink(user: IUser) {
-	if (user.inviteAcceptUrl && showInviteUrls.value) {
-		void clipboard.copy(user.inviteAcceptUrl);
-		showCopyInviteLinkToast([]);
+async function onCopyInviteLink(user: IUser) {
+	if (!showInviteUrls.value) {
+		return;
+	}
+
+	if (isTamperProofInviteLinksEnabled.value) {
```
guillaumejacquart, line 271: Is it worth moving that into it's own function also used in line ~180 ?

guillaumejacquart (approved):
Approved with minor comments. signup handler will come next, is that right ?

packages/cli/src/services/user.service.ts
BGZStephen: I didn't spend too much time on this as it's temporary really, the ID will be useful if we do targeted enabling of FF, but this is very much a boolean flag.

The owner ID is weirdly used when emails are sent containing invite links, as opposed to inviterId too, which I also found quite odd!

packages/frontend/editor-ui/src/features/settings/users/components/InviteUsersModal.vue
BGZStephen: Yeah, if this wasn't purely temporary during development of the feature I'd look for a more permanent solution, but this should all go away fairly shortly

PR #23963, opened 2026-01-07
fix: Fix ollama node url path and thinking tokens

Gulianrdgd:
## Summary
 This PR fixes two issues with the Ollama nodes:

  1. **Thinking tokens issue**: Ollama models with reasoning capabilities were dumping thinking tokens directly into the model output. Updated the `@langchain/ollama` package to the latest version and added an option to disable thinking output. I have tested this extensively and it works really well. However currently in the AI Agent node with a chat model, we cannot see the thinking output. I have written some code for that but it is rather ugly so I did not want to include it into this pull request. I noticed that the other AI models also do not show the thinking tokens, but I do think there is some use for it. If there is maybe we can think about changing the AI agent node.

We only send the think field if it is explicitly set to false. The default on Ollama is set to true. When we send `think: true` and the model does not support it, we get an error. 

The think field should also be there on the Ollama node instead of only the OllamaChat node but sadly this is not yet implemented in LangChain. I will create a pull request there and when they have a new release I will create a new pull reuqest here.
  
  2. **URL composition bug**: The URL constructor was incorrectly removing path segments after the domain. For example, `https://openwebui/ollama` + `/api/tags` resulted in `https://openwebui/api/tags` instead of `https://openwebui/ollama/api/tags`. Simplified the URL composition logic to handle paths correctly. I know that technically this should be another pull request but the change is so small that I included it here.
  
**Thinking on:**
Screenshot 2026-01-07 at 08 48 40

 **Thinking off:** 
Screenshot 2026-01-07 at 10 51 19

None

## Review / Merge checklist

- [x] Tests included. 


Gulianrdgd:
Small bump?

packages/@n8n/nodes-langchain/nodes/llms/LMChatOllama/LmChatOllama.node.ts
```diff
@@ -60,15 +58,17 @@ export class LmChatOllama implements INodeType {
 
 		const modelName = this.getNodeParameter('model', itemIndex) as string;
 		const options = this.getNodeParameter('options', itemIndex, {}) as ChatOllamaInput;
+
+		if (options.think) {
+			delete options.think;
+		}
+
```
dlavrenuek: I would suggest dropping the `think` key instead of deleting it afterwards

```suggestion
		const { think: _, ...options } = this.getNodeParameter(
			'options',
			itemIndex,
			{},
		) as ChatOllamaInput;
```

packages/@n8n/nodes-langchain/nodes/vendors/Ollama/actions/image/analyze.operation.ts
```diff
@@ -427,13 +436,26 @@ export async function execute(this: IExecuteFunctions, i: number): Promise<INode
...
 		model,
 		messages,
 		stream: false,
 		options: processedOptions,
 	};
 
+	if (thinkOption !== undefined && !thinkOption) {
+		body.think = false;
+	}
+
```
dlavrenuek: That's a lot of code just to pass the think option. What if we do following on line 386:

```ts
const { think, ...options } = this.getNodeParameter('options', i, {}) as MessageOptions;
```

and here simply:

```ts
	const body = {
		model,
		messages,
		stream: false,
		options: processedOptions,
		think,
	};
```

No need to modify the `processedOptions` or cast think, since it is already either `undefined` or `true/false`

packages/@n8n/nodes-langchain/nodes/vendors/Ollama/actions/text/message.operation.ts
```diff
@@ -392,14 +401,28 @@ export async function execute(this: IExecuteFunctions, i: number): Promise<INode
 			.filter(Boolean);
 	}
 
-	const body = {
+	const thinkOption = processedOptions.think;
+	delete processedOptions.think;
+
+	const body: {
```
dlavrenuek: Same approach as I suggested for analyze operation can be applied here

packages/@n8n/nodes-langchain/nodes/llms/LMChatOllama/LmChatOllama.node.ts
```diff
@@ -60,15 +58,17 @@ export class LmChatOllama implements INodeType {
 
 		const modelName = this.getNodeParameter('model', itemIndex) as string;
 		const options = this.getNodeParameter('options', itemIndex, {}) as ChatOllamaInput;
+
+		if (options.think) {
```
dlavrenuek: The `think` option is missing in this node's description and can not be configured. It should be in 
packages/@n8n/nodes-langchain/nodes/llms/LMOllama/description.ts
to be available in both the Ollama model and Ollama Chat model (these can be connected to the AI agent)

packages/@n8n/nodes-langchain/nodes/llms/LMChatOllama/LmChatOllama.node.ts
dlavrenuek: Why would you actually remove the think option? If the goal of these changes is to make `thinking` configurable, then it should be passed to the model.

packages/@n8n/nodes-langchain/nodes/vendors/Ollama/helpers/interfaces.ts
```diff
@@ -1,6 +1,7 @@
 export interface OllamaMessage {
 	role: 'system' | 'user' | 'assistant' | 'tool';
 	content: string;
+	thinking?: string;
```
dlavrenuek: This new parameter does not seem to be used anywhere, or? If it's not used, then we can omit it

dlavrenuek (changes requested):
Hi @Gulianrdgd, thank you for the contribution! I did test the ollama node and it works as expected 🎉. The LmChatOllama and LmOllama nodes are missing the `think` options, it can not be configured. Other suggestions are below.

packages/@n8n/nodes-langchain/nodes/vendors/Ollama/helpers/interfaces.ts
dlavrenuek: It is not available in the node output and also not used in the code otherwise, so let's remove it

packages/@n8n/nodes-langchain/nodes/llms/LMChatOllama/LmChatOllama.node.ts
Gulianrdgd: Yes this was because of an annoying bug. Thinking is enabled by default in ollama, but when we explicitly set it to true and the model does not support thinking the requests fails. So we now essentially only set it to false when we explictly set the option to false in n8n.

Another option would be to check whether the models supports thinking and then selectively adding the think option. But this seemed simpler.

packages/@n8n/nodes-langchain/nodes/vendors/Ollama/actions/image/analyze.operation.ts
Gulianrdgd: Yeah so we do this becuase of the reason mentioned earlier. Setting think to true when we do not have a thinking model the request fails. So we should filter it out.

packages/@n8n/nodes-langchain/nodes/vendors/Ollama/helpers/interfaces.ts
Gulianrdgd: Good catch. I put this there to eventually put the thinking tokens. However, this is much more of a headache than I thought. For now, I will remove it.

packages/@n8n/nodes-langchain/nodes/llms/LMChatOllama/LmChatOllama.node.ts
Gulianrdgd: Wow.... I deleted that by mistake in one commit. I have added them in the newest commit.

packages/@n8n/nodes-langchain/nodes/llms/LMChatOllama/LmChatOllama.node.ts
dlavrenuek: But when the parameter `think` is not added via options - it will be undefined, so there should be no problem with non-thinking models

Gulianrdgd: Yes this is for the case that when you have set the option think to true, but the model you have selected does not support thinking. I would say that blocking the entire request because we enabled thinking on a non thinking model is not what we want. But if you think we should then I'll change it.

dlavrenuek: Having an error thrown (failed request to ollama) if you set think to true but the model does not support thinking is exactly what I would expect. As a user I would just not add the think option when using a modal that does not support thinking

Gulianrdgd: Screenshot 2026-04-14 at 12 24 35

Gulianrdgd: Agreed, I fixed all the comments. The thinking option is now passed allong

Gulianrdgd:
@dlavrenuek Is it okay like this? Or do I need to make more changes?

dlavrenuek:
Hi @Gulianrdgd, thank you for the changes, the code looks good now. There is just a dependency conflict in the lockfile to resolve. Please use the "Rerequest review" button when ready, otherwise the PR does not show up in PRs to review on github and is easy to miss

Bildschirmfoto 2026-04-21 um 15 59 56

dlavrenuek (dismissed):
Thank you for the improvements to the Ollama nodes @Gulianrdgd 🎉 Everything looks good and works as expected 🏅

dlavrenuek:
@Gulianrdgd we have added a new checkbox `I have seen this code, I have run this code, and I take responsibility for this code.` in the PR description, the CI fails if it's not checked. I have added it to PR description but as the author you should check it yourself :)

Gulianrdgd:
@dlavrenuek Thanks! I have checked the box

dlavrenuek:
Hi @Gulianrdgd, sorry for the back and forth. Some of our e2e tests were broken so I could not merge the PR, this should be resolved now. Could you resolve the new conflict, so I can rerun the build/tests and hopefully merge it?

Gulianrdgd:
@dlavrenuek Got it! Sorry, I tried to merge the origin instead of upstream. So it took some attempts to fix it correctly

dlavrenuek (approved):

PR #23974, opened 2026-01-07
docs: Update Playwright test command in AGENTS.md

DawidMyslak:
## Summary

The command to start the tests was outdated.

https://linear.app/n8n/issue/NODE-4182/improve-agentsmd-for-playwright

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/testing/playwright/AGENTS.md
+++ b/packages/testing/playwright/AGENTS.md
@@ -2,7 +2,8 @@
 
 ## Commands
 
-- Use `pnpm --filter=n8n-playwright test:local <test-file>` to execute tests.
+- Use `pnpm --filter=n8n-playwright test:local <file-path>` to execute tests.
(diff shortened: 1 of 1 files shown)


packages/testing/playwright/AGENTS.md
```diff
@@ -2,8 +2,8 @@
 
 ## Commands
 
-- Use `pnpm test:local -- --reporter=line --grep="..."` to execute tests.
-
+- Use `pnpm --filter=n8n-playwright test:local <file-path>` to execute tests.
```
dlavrenuek, line 5: There should no need to use `--filter` because this command is supposed to be executed from `packages/testing/playwright` anyway, or?

DawidMyslak: Then it would require `cd`, I think we want to be able to run it without it, directly from the root dir.

dlavrenuek (approved):
Thank you for the improvement!

PR #23978, opened 2026-01-07
chore(ai-builder): Consolidate structured output parser recommendation and guidance prompts

mike12345567:
## Summary
The structured output parser (s.o.p) is a relatively complex sub-node that requires some specific configuration and recommendations for when to use. I've reduced the recommendations we make for this in the best practices (as it shouldn't be necessary) and centralised all the explanations from the multi-agent prompts into one location.

I've re-ran the pairwise evals and haven't seen a difference in performance when it comes to usage of the structured output parser.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/builder.prompt.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/builder.prompt.ts
@@ -6,11 +6,7 @@
  */
 
 import { prompt } from '../builder';
-import {
-	AI_AGENT_PARSER_CONFIG,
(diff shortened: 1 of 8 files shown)


aalises (dismissed):
LGTM! something that we could maybe consider in the future if we're to continue modularising prompts is to setup some kind of convention for these "modules" for the node prompts, e.g.:

- Discovery: prompts are suffixed by `criteria` (when to use the node)
- Builder: prompts are suffixed by `config` (how to configure the node)
- Connector: prompts are suffixed by `steps` (how to setup the node)

so it makes it easier to identify what each prompt does what

aalises (approved):

PR #23989, opened 2026-01-07
fix(ai-builder): Disable autosave functionality while workflow builder is generating (no-changelog)

mike12345567:
## Summary

Current behaviour has workflow builder generation saving multiple times throughout the generation when the workflow is an unfinished state. This PR disables saving while generation is streaming.

I've also removed all of the logic related to saving after the workflow builder has finished as auto-save can now look after this instead - currently two saves occur at the end as auto-save detects the changes and the workflow builder triggers a save.

Addresses: https://linear.app/n8n/issue/AI-1878/feature-stop-auto-saving-when-builder-is-in-progress

## Review / Merge checklist

- [ ] Tests included. 


mutdmour (commented):
Implementation here looks okay. Some new tests would be nice. 

But I think there's a missing piece. 

the new collaboration feature adds a write lock while the user is actively working on a workflow. The workflow is locked for 20 seconds if user is inactive.

The builder does not seem to have any effect on this. When sending a message, a write lock is requested. 20 seconds later the lock is released, even while the builder is working away.

That could lead to conflicts if another user acquires the lock while the builder is working for another user. 
Ideally the UI should keep the write lock while the builder is toiling away. 

Screenshot 2026-01-08 at 12 11 36

mutdmour (approved):

PR #23995, opened 2026-01-07
feat: Change resolveSignupToken to use new JWT for invite urls

BGZStephen:
## Summary

This PR updates resolveSignupToken to use the new JWT for validating invite urls IF the feature is enabled

PAY-4392

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/services/user.service.ts
```diff
@@ -399,4 +406,78 @@ export class UserService {
...
+		token?: string;
+		inviterId?: string;
+		inviteeId?: string;
+	}): Promise<{ inviterId: string; inviteeId: string }> {
+		if (payload.token && (payload.inviteeId || payload.inviterId)) {
+			this.logger.error('Invalid invite url containing both token and inviterId / inviteeId');
+			throw new BadRequestError('Invalid invite URL');
+		}
+
+		const instanceOwner = await this.userRepository.findOne({
```
BGZStephen, line 454: there's a weird mix of instance owner usage (for sending emails) and inviter id logic. As this feature flag is purely boolean, using the instance owner ID is perfectly fine for this

packages/cli/src/services/user.service.ts
```diff
@@ -399,4 +406,78 @@ export class UserService {
...
+			isTamperProofLinksEnabled = featureFlags[TAMPER_PROOF_INVITE_LINKS_EXPERIMENT] === true;
+		} catch (error) {
+			this.logger.debug('Failed to check feature flags for tamper-proof invite links', { error });
+		}
+
+		if (isTamperProofLinksEnabled && payload.token) {
+			return await this.processTokenBasedInvite(payload.token);
+		}
+
+		if (!isTamperProofLinksEnabled && payload.inviterId && payload.inviteeId) {
```
guillaumejacquart: What happens if the flag is enabled, but then an user clicks on an old invite link, sent before the flag was turned on ? it would get a 400 I think. I think we need to keep the old format in place without flagging for a while

guillaumejacquart (commented):
1 comment, otherwise LGTM

packages/cli/src/services/user.service.ts
BGZStephen: If this wasn't a security issue, I would agree, but I think given the account takeover capability, I think it's ok to show that an invite link is invalid and prompt users to get a new link, I think trying to support both is wasted effort, a small note in release notes to say invite links for prior versions will no longer work is fine

guillaumejacquart: I see ! maybe we could suggest asking for a new link in the error message ?

BGZStephen: Actually, while this feature is in development, we can support both, and just turn on enforcement once we remove the feature flag code, let me rejig it

BGZStephen: done

PR #24026, opened 2026-01-08
fix: Bump validator dependency

shortstacked:
## Summary

Bumps validator dependency

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -263,7 +263,7 @@ overrides:
   form-data: 4.0.4
   tmp: 0.2.4
   nodemailer: 7.0.11
-  validator: 13.15.22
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
mfsiega, line 1: Just to double check, this is intentional? (Can never remember when the lockfile should be checked in.)

mfsiega (approved):
LGTM, just double checking on the lockfile.

pnpm-lock.yaml
shortstacked: Yeah we should always update the lockfile since we do our builds/deploys based on the frozen version. (I forgot to commit the lockfile in the 1.x PR)

PR #24033, opened 2026-01-08
fix(lmChatOpenAi Node): Add validation for models requiring Responses API

aalises:
related ticket: https://linear.app/n8n/issue/AI-1863/bug-gpt-52-pro-requires-user-responses-api-enabled
## Summary

Adds validation to ensure users enable the "Use Responses API" toggle when using models that require OpenAI's Responses API endpoint (/v1/responses), while correctly allowing legacy models to use the chat/completions endpoint (/v1/chat/completions).

## Changes:

- Map the error message coming from the OpenAI API with clear error message

**Before:** Users would get a confusing "The resource you are requesting could not be found" error from OpenAI.
**After:** Users get a helpful error: "Model 'gpt-5.2-pro' requires 'Use Responses API' to be enabled"

Kapture 2026-01-08 at 14 03 14


packages/@n8n/nodes-langchain/nodes/vendors/OpenAi/helpers/constants.ts
```diff
@@ -1,3 +1,15 @@
+/**
+ * Regex pattern for models that require the Responses API.
+ * These models are not supported in the /v1/chat/completions endpoint.
+ * Matches:
+ * - gpt-5.2-pro (only pro variant of 5.2)
+ * - gpt-5.3+ (all variants of 5.3 and higher)
+ * - codex-*-max models
+ * - computer-use models
+ */
+export const MODELS_REQUIRING_RESPONSES_API_PATTERN =
```
mike12345567: Is this a safe approach, versus parsing the error and converting it to a warning about the need for the Responses API? As new models are added (say gpt-6 or something like that) this will require constant updates which is concerning.

aalises: Good point. see https://linear.app/n8n/issue/AI-1863/bug-gpt-52-pro-requires-user-responses-api-enabled#comment-b12cb2c6 this comment. The core issue is that the OpenAI API should return a semantic error in this case, but it returns instead the `resource not found` error, which is misleading.

Ideally the OpenAI API tells us exactly what went wrong and we just use that error back instead of our own validation but they're not there yet...

My concern is that I'm not sure if the "resource not found" error is legit in some other cases, or it will always be related to the `responses API` missing...

packages/@n8n/nodes-langchain/nodes/vendors/OpenAi/helpers/error-handling.ts
```diff
@@ -16,7 +16,28 @@ export function isOpenAiError(error: any): error is OpenAIError {
 	return error instanceof OpenAIError;
 }
 
-export const openAiFailedAttemptHandler = (error: any) => {
+function isNonChatModelError(error: unknown): boolean {
+	if (typeof error !== 'object' || error === null) {
+		return false;
+	}
+	const errorObj = error as Record<string, unknown>;
```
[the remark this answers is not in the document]
aalises: Fixed

mike12345567 (approved):
LGTM!

PR #24038, opened 2026-01-08
fix: Apply source control configuration changes to all multi main instances

ireneea:
## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/source-control.ee/source-control.service.ee.ts
```diff
@@ -79,6 +87,33 @@ export class SourceControlService {
...
+
+		this.isReloading = true;
+		try {
+			this.logger.debug('Source control configuration changed, reloading from database');
+
+			const wasConnected = this.sourceControlPreferencesService.isSourceControlConnected();
+			await this.refreshServiceState();
+			const isNowConnected = this.sourceControlPreferencesService.isSourceControlConnected();
+
+			if (wasConnected && !isNowConnected) {
```
BGZStephen, line 108: A couple of things:

* This feels a little like a side effect for this reload function, I wonder if there should be a different event for disconnecting git repo folders?
* If there are multiple main instances, is there a potential race condition here where the repository may be attempted to be deleted multiple times?

ireneea: Yeah I did hesitate as well creating two separate events. It does bother me that when there is a disconnect we still have to call `refreshServiceState` 

Eventually the reason why I chose not to is the added complexity when publishing and handling a disconnection event. Form the settings/preferences point of view there is no concept of the "disconnect" action. So whenever the preferences changes, we'll have to add some extra logic in there to check if the changes cause a disconnect in order to publish the event. Then there is the questions of do we publish two events when that happens, a reload + a disconnect or just one the disconnect. 

If we publish one event, then there is an added potential of race condition. Because the reload event handling would potentially create folders. And the disconnect event handler should delete those folders.

In then end I think that it's much simpler to combine both in the same event logic

ireneea: In multi-main, instance node has it's own local folder so there is no risk of race condition involving cross-instance logic. 

But in terms of handling concurrent reload events that's a bit more tricky though. I've just re-used the same mitigation pattern as the other handlers, i.e. `isReLoading` perfect but good enough in this context. 

But in the case of trying to delete an already deleted folder, It would cause an error to be logged but the error will not be propagated. So that's quite low impact

BGZStephen (dismissed):

konstantintieber (approved):

PR #24040, opened 2026-01-08
fix(core): Ensure all streaming chunks are processed before deleting ChatHub workflow

autologie:
## Summary

This PR is a follow-up of #23732 to fix FK constraint violation error caused by the changed event ordering in the PR. This PR fixes it by
- Waiting for stream to close before workflow cleanup happens
- Make streaming event aggregator callback functions async to ensure the correct event ordering

https://linear.app/n8n/issue/CHA-97/bug-foreign-key-constraint-failure-on-failed-chat-message

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/chat-hub/chat-hub.service.ts
+++ b/packages/cli/src/modules/chat-hub/chat-hub.service.ts
@@ -63,6 +63,7 @@ import {
 	JSONL_STREAM_HEADERS,
 	NODE_NAMES,
 	PROVIDER_NODE_TYPE_MAP,
+	STREAM_CLOSE_TIMEOUT,
 	TOOLS_AGENT_NODE_MIN_VERSION,
 } from './chat-hub.constants';
 import { ChatHubModelsService } from './chat-hub.models.service';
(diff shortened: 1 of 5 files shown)


packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -101,7 +103,9 @@ export class ChatHubService {
 	) {}
 
 	private async deleteChatWorkflow(workflowId: string): Promise<void> {
+		this.logger.debug('[deleteChatWorkflow] start');
```
Cadiac: These are on debug level, but do we need these?

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -757,8 +761,9 @@ export class ChatHubService {
 					model,
 					retryOfMessageId,
 				});
+				this.logger.debug('[onBegin] end');
```
Cadiac: This also seems like forgotten debug print

packages/cli/src/modules/chat-hub/chat-hub.service.ts
autologie: Ah sorry, these are leftover of investigation, will remove

Cadiac (approved):

PR #24052, opened 2026-01-08
fix(editor): Only shift downstream nodes to the right of insertion point

sandra0503:
## Summary

Refines the node shifting logic to prevent unnecessary layout changes when inserting a new node into a workflow.

Changes:
The set of downstream candidate nodes is now filtered based on their X position. Only nodes that visually overlap with the insertion point (`insertX`) or are entirely to the right of it are added to the list of nodes that must be shifted.

Previously, when a node was inserted, all nodes found in the 'downstream' path were shifted right, regardless of their current X position relative to the insertion point. In complex or looped workflows where downstream nodes might wrap back to the left, this caused:
1. Unnecessary shifting of all nodes to the right.
2. Incorrect final positioning for the newly inserted node.

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/testing/playwright/tests/e2e/regression/PAY-4367-node-shifting-cyclic.spec.ts
+++ b/packages/testing/playwright/tests/e2e/regression/PAY-4367-node-shifting-cyclic.spec.ts
@@ -1,0 +1,58 @@
+import { test, expect } from '../../../fixtures/base';
+
+/**
+ * PAY-4367: Node shifting in cyclic workflows
+ *
+ * Bug: When inserting a node, ALL downstream nodes were shifted right,
+ * including nodes to the LEFT of the insertion point (reachable via cycle).
+ *
+ * Workflow structure:
+ *   Trigger(-220) → Start(0) → Middle(220) → End(440)
+ *                      ↑___________________________|  (cycle)
+ *
+ * When inserting between Start and Middle:
+ * - insertX ≈ 158 (midpoint between Start's right edge at 96 and Middle at 220)
+ * - Start is "downstream" via cycle: Middle → End → Start
+ * - But Start (x=0, rightEdge=96) is LEFT of insertX
+ *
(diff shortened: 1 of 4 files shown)


alexgrozav (dismissed):
🙌 Thanks for the fix!

sandra0503:
It can be tested here

```
URL: https://test-node-shifting.stage-app.n8n.cloud
Email: test@n8n.io
Password: helloWorld7
```

packages/testing/playwright/tests/e2e/regression/PAY-4367-node-shifting-cyclic.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+		await n8n.canvas.addNodeBetweenNodes('Start', 'Middle', 'HTTP Request');
+
+		// Record positions AFTER insertion
+		const posStartAfter = await n8n.canvas.getNodePosition('Start');
+		const posMiddleAfter = await n8n.canvas.getNodePosition('Middle');
+		const posEndAfter = await n8n.canvas.getNodePosition('End');
+
+		// ASSERT: Start should NOT have shifted (it's to the left of insertion)
+		// This was the bug - Start would shift because it's "downstream" via the cycle
+		expect(posStartAfter.x).toBeCloseTo(posStartBefore.x, 0);
```
shortstacked: Is toBeCloseTo an exact match or just close to? Do we need to be exact?

seemewalkin: toBeCloseTo(expected, 0) allows a difference up to 0.5. Changed to toBe() for exact match since the node position shouldn't change at all for the fix! If we see flakiness, we can add a small tolerance later.

alexgrozav (approved):

PR #24058, opened 2026-01-08
feat(core): Add execution source to workflow-executed event

MarcL:
## Summary

Adds remaining non-user execution source values to the `workflow-executed` event for log streaming. This continues the work from #23372 which introduced the `source` field with initial user values.

https://linear.app/n8n/issue/PAY-2936

## Review / Merge checklist

- [x] Tests included. 
- ~[ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)~

### DIFF
--- a/packages/cli/src/events/relays/log-streaming.event-relay.ts
+++ b/packages/cli/src/events/relays/log-streaming.event-relay.ts
@@ -5,8 +5,15 @@ import type { IWorkflowBase } from 'n8n-workflow';
 
 import { MessageEventBus } from '@/eventbus/message-event-bus/message-event-bus';
 import { EventService } from '@/events/event.service';
-import type { RelayEventMap } from '@/events/maps/relay.event-map';
+import type { RelayEventMap, UserLike } from '@/events/maps/relay.event-map';
 import { EventRelay } from '@/events/relays/event-relay';
+
+type WorkflowExecutedEvent = RelayEventMap['workflow-executed'];
+type WorkflowExecutedEventWithUser = WorkflowExecutedEvent & { user: UserLike };
+
+function hasUser(event: WorkflowExecutedEvent): event is WorkflowExecutedEventWithUser {
+	return event.user !== undefined;
+}
 
 @Service()
 export class LogStreamingEventRelay extends EventRelay {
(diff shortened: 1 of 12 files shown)


packages/cli/src/events/relays/log-streaming.event-relay.ts
```diff
@@ -258,7 +257,13 @@ export class LogStreamingEventRelay extends EventRelay {
 		void this.eventBus.sendAuditEvent({
 			eventName: 'n8n.audit.workflow.executed',
 			payload: {
-				...(user && { ...user }),
+				...(user && {
+					userId: user.id,
+					_email: user.email,
```
konstantintieber: What's the intention behind this change? Looking at the `EventPayloadAudit` interface, there are properties for `firstName`, `lastName` and `userEmail`.
Why the underscores here?

MarcL: The `@Redactable()` decorator throws a `RedactableError` if it can't find the user property in the method arguments and the user is now optional (e.g. doesn't exist for the webhook etc.). The redactable transforms the user object with the properties with underscores so that's why I'm now doing it manually.

I don't like this either. 🙈 

Would it be clearer to have 2 functions for user and non-user events but still keep the audit event naming consistent? Then we can keep the redactable decorator if the user is present.

MarcL: Something like this?

```typescript
@Redactable()
private workflowExecutedWithUser({ user, workflowId, workflowName, executionId, source }) {
    void this.eventBus.sendAuditEvent({
        eventName: 'n8n.audit.workflow.executed',
        payload: { user, workflowId, workflowName, executionId, source },
    });
}

private workflowExecutedWithoutUser({ workflowId, workflowName, executionId, source }) {
    void this.eventBus.sendAuditEvent({
        eventName: 'n8n.audit.workflow.executed',
        payload: { workflowId, workflowName, executionId, source },
    });
}

private workflowExecuted(event: RelayEventMap['workflow-executed']) {
    if (event.user) {
        this.workflowExecutedWithUser(event);
    } else {
        this.workflowExecutedWithoutUser(event);
    }
}
```

konstantintieber: Oh yes I like that 👍

MarcL: Should have done this in the first place! 🙈 Fixing.

MarcL: Fixy fix: 475468a
Better?

konstantintieber (approved):

PR #24084, opened 2026-01-09
feat: Add new acceptance API, make UI use correct API for tokens

BGZStephen:
## Summary

This PR adds a new API endpoint for accepting invites based on tokens, as the previous endpoint used a clear test inviteeId no longer available for tokens.

PAY-4424

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/core/auth/views/SignupView.vue
```diff
@@ -81,17 +82,21 @@ const inviteMessage = computed(() => {
 onMounted(async () => {
 	const inviterIdParam = getQueryParameter('inviterId');
 	const inviteeIdParam = getQueryParameter('inviteeId');
+	const tokenParam = getQueryParameter('token');
+
 	try {
-		if (!inviterIdParam || !inviteeIdParam) {
+		if (!tokenParam && !inviterIdParam && !inviteeIdParam) {
```
geemanjs, line 88: I think this is impossible to fall into now? as we either use the token method or the inviterId/inviteeId param method

packages/frontend/@n8n/rest-api-client/src/api/users.ts
```diff
@@ -115,7 +115,7 @@ export async function setupOwner(
 
 export async function validateSignupToken(
 	context: IRestApiContext,
-	params: { inviterId: string; inviteeId: string },
+	params: { token?: string; inviterId?: string; inviteeId?: string },
```
geemanjs: Would a better type here be:
```javascript
{ token: string } | { inviterId: string; inviteeId: string }
```

geemanjs (commented):
Given this is "the first experience of n8n" for some users. Is it worth adding a E2E test / more complex integration test for this? Both the "happy day flow" with a valid token and the "unhappy day flow" where the token has expired?

packages/frontend/editor-ui/src/features/core/auth/views/SignupView.vue
BGZStephen: I think it's still possible to end up here, if you visit the sign up page without a token or inviteeId or ioinviterId, but this would generally happen if someone tries to manually go there, so it's more of a catch for that scenario

geemanjs (approved):

packages/frontend/editor-ui/src/features/settings/users/components/InviteUsersModal.vue
```diff
@@ -377,7 +377,7 @@ onMounted(() => {
 			<div v-if="showInviteUrls">
 				<N8nUsersList :users="invitedUsers">
 					<template #actions="{ user }">
-						<N8nTooltip v-if="isTamperProofInviteLinksEnabled">
+						<N8nTooltip v-if="isTamperProofInviteLinksEnabled && !user.firstName">
```
guillaumejacquart, line 380: Could you explain the reason behind that additional condition ?

guillaumejacquart (approved):
Just one question, but looks good !

packages/frontend/editor-ui/src/features/settings/users/components/InviteUsersModal.vue
BGZStephen: I don't think it's strictly reqwuired, I'll rmove it when we flip the switch on enabling the feature

PR #24092, opened 2026-01-09
feat(core): Add Chat hub support for 'When Last Node Finishes' and 'Using Response Nodes' response modes

Cadiac:
## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
```diff
@@ -67,8 +67,9 @@ import { useTelemetry } from '@/app/composables/useTelemetry';
 import { deepCopy, type INode } from 'n8n-workflow';
 import { convertFileToBinaryData } from '@/app/utils/fileUtils';
 import { ResponseError } from '@n8n/rest-api-client';
+import { STORES } from '@n8n/stores/constants';
 
-export const useChatStore = defineStore(CHAT_STORE, () => {
```
Cadiac, line 72: Changed the store name to match naming of others

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -163,12 +175,21 @@ export class ChatHubService {
...
-		let executionData: IRunExecutionData;
-		let workflowData: IWorkflowBase;
-		let responseMode: ChatTriggerResponseMode;
-
+		let workflow: PreparedChatWorkflow | undefined;
 		try {
-			const result = await this.messageRepository.manager.transaction(async (trx) => {
+			if (model.provider === 'n8n') {
+				// Validate and prepare n8n workflow outside the transaction to avoid deadlocks on Postgres
+				// with pool size 1 (findWorkflowForUser calls role service which may hit DB without using the transaction)
```
Cadiac: I'm not happy about this - I'm considering fixing the underlying issue at `findWorkflowForUser`'s role lookups not using the transaction, but I wasn't comfortable with the changes I had to make there so I gave up on it on friday...

packages/@n8n/api-types/src/chat-hub.ts
```diff
@@ -468,8 +468,11 @@ export class ChatHubUpdateAgentRequest extends Z.class({
 	tools: z.array(INodeSchema).optional(),
 }) {}
 
-export interface EnrichedStructuredChunk extends StructuredChunk {
```
autologie, line 473: Nice to remove workflow execution derived fields 👍

packages/cli/src/modules/chat-hub/__tests__/chat-hub.service.integration.test.ts
```diff
@@ -1344,4 +1346,697 @@ describe('chatHub', () => {
...
+				flushHeaders: jest.fn(),
+				flush: flushMock,
+				headersSent: false,
+			}) as unknown as Response;
+
+			sessionId = crypto.randomUUID();
+			messageId = crypto.randomUUID();
+		});
+
+		it('should respond with lastNode response mode and extract output from json', async () => {
```
autologie: Thank you for adding these tests. Maybe grouping tests into separate describe() by response mode improves readability

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -101,6 +101,7 @@ export class ChatHubWorkflowService {
 			newWorkflow.connections = connections;
 			newWorkflow.settings = {
 				executionOrder: 'v1',
+				saveDataSuccessExecution: 'all',
```
autologie: What was this about? Maybe useful to leave a comment for why

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessageActions.vue
```diff
@@ -80,7 +80,11 @@ function handleReadAloud() {
 					: i18n.baseText('chatHub.message.actions.readAloud')
 			}}</template>
 		</N8nTooltip>
-		<N8nTooltip v-if="message.status === 'success'" placement="bottom" :show-after="300">
+		<N8nTooltip
+			v-if="message.status === 'success' && !(message.provider === 'n8n' && message.type === 'ai')"
```
autologie: Can we create `isEditable(message: ChatMessage)` in utils?

packages/cli/src/modules/chat-hub/chat-hub.constants.ts
```diff
@@ -1,14 +1,16 @@
...
+
 export const EXECUTION_POLL_INTERVAL = 1000;
 export const STREAM_CLOSE_TIMEOUT = 5 * 60 * 1000; // 5 minutes
 export const EXECUTION_FINISHED_STATUSES: ExecutionStatus[] = [
 	'canceled',
 	'crashed',
 	'error',
 	'success',
-];
+] as const;
```
autologie: [nits] I think `: ExecutionStatus[]` makes `as const` ineffective, you might want to use `as const satisfies ExecutionStatus[]` for preserving concrete values

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: I should've written a comment when I added this before christmas 😓 
I think I wanted to have title generation work on instances that didn't save executions, and figured that forcing this on shouldn't hurt anyways as any chat executions are anyways removed after completition.

I'll check this and either add a comment or remove.

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -231,39 +236,80 @@ export class ChatHubService {
...
+		}
+
+		const previousMessage = await this.ensurePreviousMessage(previousMessageId, sessionId);
+		if (
+			model.provider === 'n8n' &&
+			workflow.responseMode === 'responseNodes' &&
+			previousMessage?.status === 'waiting' &&
+			previousMessage?.executionId
+		) {
+			const execution = await this.executionRepository.findSingleExecution(
```
autologie, line 250: Is there a way to ensure this execution belongs to the user?

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -742,10 +796,388 @@ export class ChatHubService {
...
+				executionId,
+				messageId,
+				previousMessageId,
+				retryOfMessageId,
+			);
+		}
+	}
+
+	private async resumeChatExecution(
+		execution: IExecutionResponse | undefined,
```
autologie: [nits] Looks like undefined is eliminated at call sites.

packages/cli/src/modules/chat-hub/chat-hub.service.ts
autologie, line 1: n8n workflow agent requires a lot of special handling, I feel we should probably extract them into a separate service. Could be in a separate PR.

packages/cli/src/modules/chat-hub/chat-hub.service.ts
Cadiac: not really, but the only way they could've gotten the executionId saved to their chat message is through previously valid chat messaging 🤔

Cadiac: Executions just belong to workflows, and we're ensuring they have access to the workflow on prepareReplyWorkflow -> prepareWorkflowAgentWorkflow

autologie: Yes I don't think `executionId` can be compromised but thought it's better to check if possible (was surprised that execution doesn't have ownerId). I think we can keep it like this

Cadiac: executions aren't usually related to users, as they start from things like triggers (unless they're manual executions)

packages/cli/src/modules/chat-hub/__tests__/chat-hub.service.integration.test.ts
Cadiac: Reorganized the tests

```
    sendHumanMessage
      Base LLM chats
        ✓ should respond and persist generated response chunks sent from workflow execution (266 ms)
        ✓ should respond and persist an error chunk sent from workflow execution (281 ms)
        ✓ should respond and persist an error set in the workflow execution (1297 ms)
        ✓ should clear stream timeout when stream closes normally (136 ms)
        ✓ should log warning when stream does not close within timeout (200 ms)
      n8n workflow agents
        "When Last Node Finishes" response mode
          ✓ should respond with "lastNode" response mode and extract output from json (276 ms)
          ✓ should respond with "lastNode" response mode and handle errors (278 ms)
          ✓ should extract text field when output is not present in "lastNode" mode (284 ms)
        "Using Response Nodes" response mode
          ✓ should respond with "responseNodes" response mode and resume immediately when waitUserReply is false (282 ms)
          ✓ should respond with "responseNodes" response mode and wait for user when waitUserReply is true (285 ms)
        "Using 'Respond to Webhook' Node" response mode
          ✓ should throw error when "responseNode" mode is used (74 ms)
```

autologie (approved):
Great, this is an epic! And thank you for addressing comments

PR #24093, opened 2026-01-09
fix(editor): Correct undo state after adding node inbetween existing nodes

CharlieKolb:
## Summary

When adding a node inbetween existing nodes, we'd previously end up in a state where undo-ing didn't remove the node. After investigating this turned out due to a missing `trackBulk: false` in the `deleteConnection()` call, which would override the move + add combination already on the stack.

The addConnection calls then need to be amended to support correctly redo-ing this operation.

https://linear.app/n8n/issue/ADO-4599

## Review / Merge checklist

- [ ] Tests included. 


alexgrozav (approved):
🚀

PR #24099, opened 2026-01-09
fix(ai-builder): Reducing confusion in spec evals from double negative don't criteria

mike12345567:
## Summary
There was a pattern in the spec evals which was potentially causing some confusion for the judge, with double negatives in the don't categories.

An example of this is that the don't criteria could map into the prompt like:
```
[DON'T]
DO NOT use a code node
DO NOT use a http request node
```
This is sort of clear, but the fact it is wrapped in a don't section can lead to confusion/negating the DO NOT, causing it to think it should check for it using a code node. Running the same evals which were failing previously, they now pass with this change.

I've also clarified the properties of the output parser to hopefully reduce the chance of it getting ignored/not being filled in and throwing an error.

Finally, updated the judge chain to use the new prompt builder for consistency.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/evaluations/pairwise/judge-chain.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/evaluations/pairwise/judge-chain.ts
@@ -2,6 +2,7 @@ import type { BaseChatModel } from '@langchain/core/language_models/chat_models'
 import type { RunnableConfig } from '@langchain/core/runnables';
 import { z } from 'zod';
 
+import { prompt } from '../../src/prompts/builder';
 import type { SimpleWorkflow } from '../../src/types/workflow';
 import { createEvaluatorChain, invokeEvaluatorChain } from '../chains/evaluators/base';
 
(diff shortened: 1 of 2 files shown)


packages/@n8n/ai-workflow-builder.ee/evaluations/pairwise/judge-chain.ts
```diff
@@ -17,19 +18,27 @@ const pairwiseEvaluationLLMResultSchema = z.object({
...
-				justification: z.string(),
+				rule: z.string().describe('The criterion that was violated'),
+				justification: z
+					.string()
+					.describe('Evidence from the workflow explaining why this is a violation'),
 			}),
 		)
-		.describe('List of criteria that were violated'),
+		.describe(
+			'List of criteria that were violated. If no criteria passed, include an entry explaining why.',
```
burivuhster: Is the goal of this addition here to make sure it mentions all failed checks?

burivuhster (dismissed):

burivuhster (approved):
That's a good addition!

PR #24100, opened 2026-01-09
Change default read status to 'both' in GmailTrigger

FazlOmar9:
Seems more intuitive to leave it on both by default.

## Summary

Instead of reading unread emails only, the trigger should check all emails by default. Then if someone specifically wants it to read only unread/read emails they can change it using the filter.


neelpatel19 (commented):
s

PR #24111, opened 2026-01-09
feat(core): Support disabling publishing workflows in personal space (no-changelog)

geemanjs:
## Summary

Adds two new endpoints:
- GET /rest/settings/security
- POST /reset/settings/security

These endpoints allow to fetch and change the value of the new `personalSpacePublishing: boolean` setting.

Adds a new route under `/settings/security` to interact with the new endpoint for controlling whether users can publish workflows from their personal space.

Set envvar variable `N8N_ENV_FEAT_PERSONAL_SECURITY_SETTINGS` to true to enable the UI part of this feature.

image

When this setting is disabled, this member user can no longer publish personal workflows:

image

PAY-4331
closes https://linear.app/n8n/issue/IAM-117/complete-existing-pocs-and-create-follow-up-tickets-if-needed

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/services/__tests__/auth.roles.service.test.ts
```diff
@@ -492,112 +366,185 @@ describe('AuthRolesService', () => {
...
-			//
 			Object.keys(ALL_ROLES).forEach((roleNamespace) => {
 				expect(logger.debug).toHaveBeenCalledWith(
 					expect.stringContaining(`${roleNamespace} roles`),
 				);
 			});
 		});
+
+		describe('personal space publishing', () => {
+			test('should add workflow:publish to personalOwner role when setting is null (backward compatibility)', async () => {
```
konstantintieber, line 429: @geemanjs love these tests 👍

packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -96,6 +103,35 @@ export class AuthRolesService {
...
+		return setting?.value === 'true' || setting === null;
+	}
+
+	/**
+	 * Modifies the expected scopes for a role based on settings.
+	 * Currently only applies to project:personalOwner role.
+	 * Uses a "closed first" approach: workflow:publish is not in the base definition
+	 * and is added when the setting is enabled.
+	 */
+	private async updateScopesBasedOnSettings(
```
phyllis-noester, line 118: imo this should not be called updateScopesBasedOnSettings but rather getScopesBasedOnSettings since it doesn't update the db

packages/cli/src/services/security-settings.service.ts
```diff
@@ -0,0 +1,37 @@
...
+	async setPersonalSpacePublishing(enabled: boolean): Promise<void> {
+		await this.settingsRepository.upsert(
+			{
+				key: PERSONAL_SPACE_PUBLISHING_SETTING_KEY,
+				value: enabled.toString(),
+				loadOnStartup: true,
+			},
+			['key'],
+		);
+
```
phyllis-noester, line 25: just to understand: afaik we store all auth config (roles, permissions) in the db right? So there is probably a middleware that ensures `this.settingsRepository.upsert` and `this.roleService.addScopeToRole` are executed in a transaction such that we can never get into an inconsistent state where e.g. the repo setting is updated but the scope is not added/removed from the role?

phyllis-noester: CC @afitzek do you know?

afitzek: No these are not running in a transaction, I'll give you more details in slack.

cstuncsik (approved):

guillaumejacquart (approved):

PR #24112, opened 2026-01-09
feat: Add security rules for Cubic PR reviews (no-changelog)

MarcL:
## Summary

Add initial security rules for Cubic PR reviews.

N/A

## Review / Merge checklist

- ~[ ] Tests included. ~
- ~[ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)~

### DIFF
--- a/cubic.yaml
+++ b/cubic.yaml
@@ -107,6 +107,78 @@ reviews:
         - Exports, types, configs
         - Metadata files
         - Version files
+    - name: Security Review
+      description: |-
+        Proactively review PRs and flag security vulnerabilities specific to n8n's architecture:
+
+        **Code Execution & Sandboxing:**
+        - Changes that weaken expression evaluation sandbox protections
+        - Modifications to Code node sandboxes (JavaScript/Python) that reduce isolation
+        - New access to Node.js builtins or external modules in sandboxed contexts
+        - Prototype pollution or constructor access bypasses
+        - Weakening of prototype sanitizers or function context validators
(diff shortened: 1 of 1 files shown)


cubic.yaml
```diff
@@ -107,6 +107,78 @@ reviews:
         - Exports, types, configs
         - Metadata files
         - Version files
+    - name: Security Review
+      description: |-
+        Proactively review PRs and flag security vulnerabilities specific to n8n's architecture:
+
```
phyllis-noester, line 113: Super nice 🥳 

a few ideas but non blocking:

maybe some more stuff around the db? e.g.

- Missing db indexes in important queries and other very resource intensive queries
- missing data encryption for sensitive fields

also, given that we store a lot of credentials, we could highlight that credentials access can be scoped to instance,project or person (or whatever all the available option are) and it is essential that we only expose them or data fetched via the credentials to the allowed user group. 

we could also add something about modularization. afaik newer features are implemented into appropriate modules whereas some olders are not which can lead to issues with licensing if enterprise code is made available to non enterprise customers

phyllis-noester (dismissed):

cubic.yaml
MarcL: Added some changes in c4bc854.
Thoughts?

phyllis-noester: niceeeee 🥳

phyllis-noester (approved):

PR #24113, opened 2026-01-09
fix(core): Put source control preferences behind auth

phyllis-noester:
## Summary

Tested that source control preferences still work manual, however, related e2e tests are currently disabled. should be enabled again after https://linear.app/n8n/issue/PAY-4365/bug-source-control-operations-fail-in-multi-main-deployment

Given that this endpoint is for internal use of the n8n instance, I assume the changelog would be enough & we don't need to call this out to customers.

https://linear.app/n8n/issue/PAY-3328/restsource-controlpreferences-should-be-behind-auth

## Review / Merge checklist

- [ ] Tests included. 


afitzek (commented):
Was there a reason why the authentication was skipped before or was it just an oversight? Can we please add test cases, to ensure that the endpoint is authenticated? :)

afitzek (approved):
Looks good to me, I thought it was important to have a negative test, basically trying to call the endpoint without authentication to make sure this test would fail if this was changed at any point in the future, the test you added with checking the registry, should be good and work fine. 👍

PR #24114, opened 2026-01-09
fix(ai-builder): Long running prompts throw internal 401 error rather than expected complexity error

mike12345567:
## Summary

Converts "401 Expired token" errors from LangChain during long-running workflow generations to the existing "too complex" error message, providing a better user experience instead of showing raw LangChain troubleshooting URLs.

Addresses: https://linear.app/n8n/issue/AI-1882/bug-multi-agent-seems-to-have-timed-out-and-shows-internal-error

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/workflow-builder-agent.ts
```diff
@@ -631,15 +634,46 @@ export class WorkflowBuilderAgent {
...
 
 		// Re-throw any other errors
 		throw error;
 	}
 
+	/**
+	 * Checks if the error is a 401 expired token error from the LLM provider.
+	 * This typically occurs during very long-running workflow generations.
+	 */
+	private isTokenExpiredError(error: unknown): boolean {
```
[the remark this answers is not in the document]
mike12345567: Added test cases.

packages/@n8n/ai-workflow-builder.ee/src/workflow-builder-agent.ts
```diff
@@ -631,15 +634,46 @@ export class WorkflowBuilderAgent {
...
+			return false;
+		}
+
+		// Check for status/statusCode property being 401
+		if (
+			('status' in error && error.status === 401) ||
+			('statusCode' in error && error.statusCode === 401)
+		) {
+			return true;
+		}
```
burivuhster: I'm a bit concerned we'll mask ALL 401 errors with the "workflow is too complex" error. Should we be more specific here and look for some unique attributes of the error originating in our ai-assistant-service rather than just http code?

mike12345567: My thinking with this is that if there is any kind of token issue, we'd rather mask the error to the user but perhaps it would be better to handle it in the frontend in some way for that reason?

burivuhster: It makes sense to mask token-related issues, my point was to add some condition that this error originates from ai-assistant service and not from somewhere else. There is only only one place (which is ai assistant service) now, but we may add e.g. internet browsing tool in the future and ALL http requests failing with 401 error will end up with a "workflow is too complex error"

mike12345567: Ah yes - totally understand, good recommendation!

mike12345567: After some investigation I can't confirm the failing request came from the ai-assistant-service, but I can confirm it was from within langchain and specific to the request being denied - so pretty much the same benefit!

aalises (dismissed):

burivuhster (approved):
Thanks for addressing my comments 🙌

PR #24134, opened 2026-01-11
Add working example for N8N_CONTENT_SECURITY_POLICY trusted domains

gkp1:
The example that is currently in there does not work in prod and might prevent n8n frontend from loading and obviously webhooks too
- New example includes 'self' in the frame ancestors, should cover both n8n host == localhost or domain 
- Tested on ubuntu with nginx + https, embedding in the respective sites was not allowed -> is now allowed

## Review / Merge checklist

- [ x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] Tests included. 


Joffcom:
Hey @gkp1 

This seems like it might be better to have in the docs, we don't generally expect users to look through the code to see how something works

PR #24151, opened 2026-01-12
feat(HTTP Request Node): Add support for PROPFIND, MKCOL, MOVE, COPY and REPORT methods

axodouble:
## Summary

Adds support for the WebDAV HTTP methods `PROPFIND`, `MKCOL`, `MOVE`, `COPY`, and `REPORT` to the HTTP Request node and the HTTP Request Tool node.

Previously the HTTP Request node only sent a request body for `PATCH`, `POST`,`PUT`, and `GET` methods (see `HttpRequestV3.node.ts`), so requests using other methods silently dropped their body. This made it impossible to use n8n with WebDAV servers and similar services that rely on these methods. This PR adds the new methods to the Method dropdown and includes them in the body-sending logic, so a body is sent when `sendBody` is enabled.

The `IHttpRequestMethods` type is extended with the new methods so they are recognised across the codebase.

## How to test

1. Add an **HTTP Request** node to a workflow.
2. Open the **Method** dropdown — the new options `PROPFIND`, `MKCOL`, `MOVE`, `COPY`, and `REPORT` are now available.
3. Point the node at a WebDAV-capable endpoint, select e.g. `PROPFIND`, enable **Send Body**, and provide a body.
4. Execute the node and confirm the request is sent with the chosen method and the body is included.

The same options are also available on the **HTTP Request Tool** node.

Automated tests in `HttpRequestV3.test.ts` cover that the new methods appear in the Method dropdown and that each method issues a request with the expected method and URL.

Community forum: https://community.n8n.io/t/add-support-in-the-http-request-node-for-the-propfind-http-method/27697

## Review / Merge checklist

- [x] Tests included.


Axodouble:
If there are issues with the code or tests, I have been struggling setting up the environment locally due to a pnpm issue that is occuring on my development environment, once I am home I can review this on my own setup if this is not sufficient.
Thanks for your patience!

mfernandezfidesol:
i need!

muuky:
Is there any update for this?
I need PORPFIND for my caldav sync automation.

DevEddy:
Any news on this?

axodouble:
Unfortunately, no changes merged yet, still waiting for approval from n8n.

Bazoogle:
Hopefully this is still in the review queue, as this would be a very useful addition. No good way to do CalDAV and CardDAV without the added request methods. This seems much easier than adding dedicated DAV nodes.

flipper:
Kinda wild how this simple PR has been unnoticed by n8n team for half a year. This is a very easy win for the community of n8n.

The automations to automate caldav is endless.
I can even see this useful for those AI agent lovers too.

Joffcom:
Let me run this through the readiness checks, as this is one of the most popular nodes we do need to be careful with changes to it to avoid any unexpected breaking changes.

Not that I expect this to introduce any breaking changes.

packages/nodes-base/nodes/HttpRequest/V3/HttpRequestV3.node.ts
```diff
@@ -446,7 +446,7 @@ export class HttpRequestV3 implements INodeType {
 				}
 
 				// Change the way data get send in case a different content-type than JSON got selected
-				if (sendBody && ['PATCH', 'POST', 'PUT', 'GET'].includes(requestMethod)) {
+				if (sendBody && ['PATCH', 'POST', 'PUT', 'GET', 'PROPFIND', 'MKCOL', 'MOVE', 'COPY', 'REPORT'].includes(requestMethod)) {
```
dlavrenuek, line 449: When we flip this condition it will be more targeted


```suggestion
				if (sendBody && !['GET', 'HEAD', 'OPTIONS', 'TRACE'].includes(requestMethod)) {
```

dlavrenuek (changes requested):
Hi @Axodouble, this addition will be definitely useful for people using WebDAV. With the very popular built in nodes, like the http request node, we have to be very careful about changes, tho. Considering that most people using this node will never need the WebDAV methods, having them in the list will actually worsen the user experience for most. To still provide support for new methods while keeping the same user experience I suggest adding a new optional option `Enable WebDAV methods` and `displayOptions` for the webdav options. Together with a flipped check for sending a body (separate comment) we will support new use cases while keeping a clean ui

packages/nodes-base/nodes/HttpRequest/V3/HttpRequestV3.node.ts
rourke: Excellent suggestion. Can this go into the live version? I really need these methods.

axodouble:
Sorry for the hiatus on making the changes, I have not had the time to look into this properly, I will be applying this as soon as possible.

packages/@n8n/nodes-langchain/nodes/tools/ToolHttpRequest/ToolHttpRequest.node.ts
```diff
@@ -109,6 +109,26 @@ export class ToolHttpRequest implements INodeType {
 						name: 'PUT',
 						value: 'PUT',
 					},
+					{
```
dlavrenuek: This node is deprecated and hidden, let's remove the changes here. The HttpRequestV3 node can be used as tool for AI nodes

packages/nodes-base/nodes/HttpRequest/V3/Description.ts
```diff
@@ -26,6 +54,7 @@ export const mainProperties: INodeProperties[] = [
 				name: 'HEAD',
 				value: 'HEAD',
 			},
+			...webdavMethodOptions,
```
dlavrenuek: Spreading the methods in the middle of the existing methods leads to breaking the currently correct sorting. Since the WebDAV methods have their own logical scope we can spread them at the end of the current array

packages/nodes-base/nodes/HttpRequest/V3/Description.ts
```diff
@@ -1,7 +1,35 @@
...
+import type { INodeProperties, INodePropertyOptions } from 'n8n-workflow';
 
 import { optimizeResponseProperties } from '../shared/optimizeResponse';
 
+const webdavMethodOptions: INodePropertyOptions[] = [
+	'PROPFIND',
+	'MKCOL',
+	'MOVE',
+	'COPY',
+	'REPORT',
```
dlavrenuek: One rule we have for the options is that these must be sorted alphabetically, let's rearrange the methods here, so these are sorged

packages/nodes-base/nodes/HttpRequest/V3/Description.ts
```diff
@@ -1,7 +1,35 @@
...
+				'options.webdavMethods': [true],
+			},
+		},
+	},
+	{
+		name: method,
+		value: method,
+		displayOptions: {
+			hide: {
+				method: [{ _cnd: { not: method } }],
```
dlavrenuek, line 26: Nice solution to keep the option visible while the flag is turned off afterwards

dlavrenuek (changes requested):
Hi @axodouble, thanks for following up and great that you did catch that `GET` method must be kept in the allow body methods 🏅 Few minor changes we need to address due to structural rules we have for nodes and parameters and we are good to go!

rourke:
Not sure what happened here? It looks like this is closed by n8n-assistant because of the protected nodes. But this seems to have gotten the approval from the n8n team already.

dlavrenuek:
@axodouble we now have a policy to not allow external contributions for the nodes like http request node, but since this change was opened before and is already reviewed, let's try to get the changes in (if it's not autoclosed again)

packages/@n8n/nodes-langchain/nodes/tools/ToolHttpRequest/ToolHttpRequest.node.ts
axodouble: Ah, sorry about that, I was quite confused as to why I had found multiple references to the node. Will resolve.

dlavrenuek (approved):
Thank you for the contribution and for all requested improvements @axodouble 🎸

PR #24159, opened 2026-01-12
feat(core): Add signature validation for waiting webhooks and forms

elsmr:
Adds signature validation for waiting webhooks and forms to improve request handling security.

- Add signature validation for incoming requests to waiting endpoints
- Add test coverage for waiting endpoint scenarios

More info and testing notes in Linear

  https://linear.app/n8n/issue/NODE-3829

  ## Review / Merge checklist

  - [x] Tests included.


packages/cli/src/webhooks/waiting-forms.ts
```diff
@@ -85,6 +85,20 @@ export class WaitingForms extends WaitingWebhooks {
 
 		const execution = await this.getExecution(executionId);
 
+		// Validate token for forms if required
+		if (execution?.data.resumeToken) {
```
elsmr, line 75: Existing executions won't have `resumeToken` in the execution data, so they can still resume without a signature

packages/workflow/src/utils.ts
```diff
@@ -13,6 +13,17 @@ import * as LoggerProxy from './logger-proxy';
 
 const readStreamClasses = new Set(['ReadStream', 'Readable', 'ReadableStream']);
 
+/**
+ * Generates a cryptographically secure random token.
+ * Uses the Web Crypto API which is available in both browsers and Node.js 15+
```
tomi: Do we need to have this work in both browser and node? Don't we generate these only on the BE? Would be much simpler if it's only nodejs:

```
return crypto.randomBytes(32).toString('hex');
```

packages/cli/src/webhooks/waiting-webhooks.ts
```diff
@@ -142,15 +169,24 @@ export class WaitingWebhooks implements IWebhookManager {
 
 		const execution = await this.getExecution(executionId);
 
+		// Only validate signature for executions that have validateSignature flag set
+		// This provides backwards compatibility for old executions created before signature validation
 		if (execution?.data.validateSignature) {
```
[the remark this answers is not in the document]
elsmr: `resumeToken` was on this branch before, but was never released, now back to using HMAC

michael-radency (dismissed):
code looks good, working as expected, approved 👍

packages/cli/src/webhooks/webhook-helpers.ts
```diff
@@ -661,7 +667,14 @@ export async function executeWebhook(
 		});
 
 		if (responseMode === 'formPage' && !didSendResponse) {
-			res.send({ formWaitingUrl: `${additionalData.formWaitingBaseUrl}/${executionId}` });
+			const formUrl = new URL(`${additionalData.formWaitingBaseUrl}/${executionId}`);
+			// Sign the form URL using HMAC and include signature in URL
+			const instanceSettings = Container.get(InstanceSettings);
+			const urlForSigning = prepareUrlForSigning(formUrl);
+			const signature = generateUrlSignature(urlForSigning, instanceSettings.hmacSignatureSecret);
+			formUrl.searchParams.set('signature', signature);
```
tomi: `WAITING_TOKEN_QUERY_PARAM`?

packages/cli/src/webhooks/waiting-forms.ts
```diff
@@ -85,6 +86,23 @@ export class WaitingForms extends WaitingWebhooks {
 
 		const execution = await this.getExecution(executionId);
 
+		// Validate signature for forms if required (backwards compat: skip for old executions)
```
tomi: This method is already super long. Can we clean it up a bit before adding new things? (e.g. extract somer parts to separate methods)

packages/workflow/src/run-execution-data-factory.ts
```diff
@@ -57,28 +85,10 @@ export function createRunExecutionData(
...
-						metadata: options.executionData?.metadata ?? {},
-						waitingExecution: options.executionData?.waitingExecution ?? {},
-						waitingExecutionSource: options.executionData?.waitingExecutionSource ?? {},
-						runtimeData: options.executionData?.runtimeData,
-					},
+		resultData: buildResultData(options.resultData),
+		executionData: buildExecutionData(options.executionData),
 		parentExecution: options.parentExecution,
-		validateSignature: options.validateSignature,
+		resumeToken: options.resumeToken ?? generateSecureToken(),
```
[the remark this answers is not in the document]
elsmr: Supporting executions without signature would not fix the issue, also this is only a problem if users manually construct the resume URL, if they use `$execution.resumeUrl`, it will have a `resumeToken` included.

packages/core/src/execution-engine/node-execution-context/utils/get-additional-keys.ts
```diff
@@ -16,6 +16,16 @@ import {
 } from './execution-metadata';
 import { getSecretsProxy } from './get-secrets-proxy';
 
+function appendResumeToken(url: string, token: string): string {
+	try {
+		const urlObj = new URL(url);
+		urlObj.searchParams.set(WAITING_TOKEN_QUERY_PARAM, token);
+		return urlObj.toString();
+	} catch {
+		return url;
```
tomi: Are there any valid cases when this might happen? It makes me a bit uneasy to fallback to an insecure mode in case of an error

tomi (commented):
Looks good 👏 Couple questions/comments

packages/core/src/execution-engine/node-execution-context/utils/get-additional-keys.ts
elsmr: Not really, I think this was a shortcut to fix a bunch of tests where the base URLs were not set correctly. Removed the catch and updated the tests!

packages/cli/src/webhooks/waiting-forms.ts
```diff
@@ -171,4 +134,29 @@ export class WaitingForms extends WaitingWebhooks {
 			suffix,
 		});
 	}
+
+	private handleStatusRequest(
```
tomi, line 144: Could we have JSdoc for this? It's not immediately clear from the name what this does

tomi (dismissed):
LGTM 🚀 Just one minor nit

tomi (approved):
🚀

PR #24165, opened 2026-01-12
fix(ai-builder): Fix dirty state for nodes when importing workflows

mutdmour:
## Summary
- fix issue where nodes shown as dirty after workflow changes
- refactor, only apply workflow diffs instead of deleting nodes and recreating them
- fix undo/redo for builder changes

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
```diff
@@ -0,0 +1,330 @@
...
+			trackBulk: false,
+		});
+	}
+
+	/**
+	 * Updates workflow with new data, preserving existing node positions
+	 * and handling renames, additions, and removals efficiently.
+	 * Note: Undo recording should be managed by the caller (e.g., start when streaming begins, stop when it ends).
+	 */
+	async function updateWorkflow(
```
aalises, line 295: If any step fails (e.g. addNodes throws), the workflow will be left potentially in a partially updated state.

Consider adding some error handling here (`try/catch`, returning `{success: false, error }` on failure, implementing rollback logic (perhaps not needed).

packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	/**
+	 * Restore pinned data by node ID to current node names
+	 */
+	function restorePinnedData(pinnedDataById: Map<string, INodeExecutionData[]>): void {
+		workflowsStore.allNodes.forEach((node) => {
+			const savedPinData = pinnedDataById.get(node.id);
+			if (savedPinData) {
+				workflowsStore.workflow.pinData = {
```
aalises: Storing to store directly, consider using a store action/method like `setWorkflowPinData` from `workflowState.`

packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
```diff
@@ -0,0 +1,330 @@
...
+			const nodeType = nodeTypesStore.getNodeType(node.type);
+			if (!nodeType?.credentials) return;
+
+			// Try to find and set the first available credential
+			for (const credentialConfig of nodeType.credentials) {
+				const credentials = credentialsStore.getCredentialsByType(credentialConfig.name);
+				if (!credentials || credentials.length === 0) continue;
+
+				const credential = credentials[0];
+				const existingNode = workflowsStore.getNodeByName(node.name);
```
aalises: [Not a biggie] perhaps instead of iterating over `nodes` and then doing

```ts
const existingNode = workflowsStore.getNodeByName(node.name);
if (!existingNode) continue;
```

We can directly get all the nodes from the store from `nodes` (`nodes -> store nodes`) in batch, and then iterate directly from all the found nodes in the store. 1 batch operation vs. multiple one-of fetches.

packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
mutdmour: That's a good point. Will add some try/catch. Not sure what the best UX is. 
I am thinking it might be best to show an error, stop generating, and allow the user to manually revert using the restore feature.

packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
mutdmour: Good point. Thanks for catching. I just copied this logic over from the store without review.

packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
aalises: That'd work 💪

packages/frontend/editor-ui/src/app/composables/useErrorHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+	context?: string;
+	/** Override the default toast title */
+	title?: string;
+}
+
+/**
+ * Composable for unified error handling.
+ * Shows error toast, logs to console, and captures exception to Sentry.
+ */
+export function useErrorHandler(options: ErrorHandlerOptions) {
```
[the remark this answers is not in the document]
mutdmour: tests are there

aalises (dismissed):
Approving for merge, only minor comments that have been addressed.

packages/frontend/editor-ui/src/app/composables/useWorkflowUpdate.ts
mutdmour: I avoided `setWorkflowPinData` because it has side effects. We are not pinning the data again, we are just ensuring it does not get lost.

That being said, maybe we don't need this anymore. We needed this before when we deleted nodes. Will test without

mutdmour: Yup not needed, will remove

aalises (approved):

PR #24169, opened 2026-01-12
feat: Add feature for clearing credentials on resolver update

BGZStephen:
## Summary

This PR forces a user to decide whether to clear credentials connected to a resolver when updating fields other than the name field.

PAY-4414

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver.service.ts
```diff
@@ -132,6 +135,14 @@ export class DynamicCredentialResolverService {
 			existing.name = params.name;
 		}
 
+		if (params.clearCredentials === true) {
```
afitzek, line 136: This needs to be delegated to the resolver instance, because not all resolvers will use the database entity, for example a resolver implementation decides where the entities are stored. An http resolver would need to call an endpoint on a third party service to trigger this action.

BGZStephen: Yep, makes sense, I've fetched the resolver from registry and added a delete all secrets, which is optional in the same way as deleteSecret is, and added the function on to the entry storage, I think this is what you were meaning?

afitzek: I think we would ideally not only pass in the id, but the whole resolver handle, because the handle has the configuration for the resolver to operate on. For example for the http resolver, it might hold the base url etc. The id is also part of the handle and this would follow the rest of the interface.

BGZStephen: Sure, that makes sense, tweaked it

afitzek (approved):

PR #24177, opened 2026-01-12
refactor(core): Improve expressions handling

ivov:
https://linear.app/n8n/issue/CAT-2086


mfsiega (approved):
I'm not an expert on AST stuff, but the intent and testing looks clear.

PR #24185, opened 2026-01-12
fix(core): Use fsRealpath instead of resolve to get the real path

mfsiega:
## Summary

Today, if we're resolving a filepath and the file doesn't exist, `fsRealpath` throws an error. In this case we fall back to `resolve`. This lets us handle the case where we want to create the file eventually e.g., if we're going to `write`. However, `resolve` doesn't actually resolve symlinks. This defeats part of the purpose of our `resolvePath` helper, which is to ensure that we're working with fully resolved, canonical path. (Since this is what we'll check against allow/block lists.)

This change introduces a strategy where:
- If the file doesn't exist, get the realpath of the containing directory. If *that* doesn't exist, then we return "file not found" (since the eventual write operation would fail anyway). If it _does_ exist, we append the basename and use that as our path. Note that we don't have to worry about symlinks in the "last hop" because we read/write with NO_FOLLOW.

Tested manually to verify that we can still write files that didn't previously exist. Also verified that symlinks end up getting resolved correctly.

https://linear.app/n8n/issue/CAT-1852

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/file-system-helper-functions.ts
@@ -12,7 +12,7 @@ import {
 	open as fsOpen,
 } from 'node:fs/promises';
 import { homedir } from 'node:os';
-import { resolve, posix } from 'node:path';
+import { resolve, posix, dirname, basename, join } from 'node:path';
 
 import {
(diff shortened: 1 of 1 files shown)


Cadiac (approved):
Seems to work as described on the PR 👍

PR #24192, opened 2026-01-12
fix: Regenerate form webhook ids when pasting workflow data

dlavrenuek:
## Summary

When form nodes containing webhooks are copy&pasted into a workflow the webhook url (form url) will always be regenerated.

- https://linear.app/n8n/issue/NODE-4054

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
@@ -2309,7 +2309,7 @@ export function useCanvasOperations() {
 						nodeNames.add(newName);
 					}
 
-					// Generate new webhookId for all webhook nodes
+					// Generate new webhookId
 					if (node.webhookId && UPDATE_WEBHOOK_ID_NODE_TYPES.includes(node.type)) {
 						if (node.webhookId) {
 							nodeHelpers.assignWebhookId(node);
--- a/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
@@ -2314,7 +2314,7 @@ export function useCanvasOperations() {
 
 						if (node.parameters.path) {
 							node.parameters.path = node.webhookId;
-						} else if ((node.parameters.options as IDataObject).path) {
+						} else if ((node.parameters.options as IDataObject)?.path) {
 							(node.parameters.options as IDataObject).path = node.webhookId;
 						}
 					}
--- a/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
@@ -31,7 +31,6 @@ import {
 	FORM_TRIGGER_NODE_TYPE,
 	MCP_TRIGGER_NODE_TYPE,
 	STICKY_NODE_TYPE,
-	UPDATE_WEBHOOK_ID_NODE_TYPES,
 	VIEWS,
 	WEBHOOK_NODE_TYPE,
 } from '@/app/constants';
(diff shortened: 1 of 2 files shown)


packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
```diff
@@ -2309,19 +2308,14 @@ export function useCanvasOperations() {
...
 
-							if (node.parameters.path) {
-								node.parameters.path = node.webhookId;
-							} else if ((node.parameters.options as IDataObject).path) {
-								(node.parameters.options as IDataObject).path = node.webhookId;
-							}
+						if (node.parameters.path) {
+							node.parameters.path = node.webhookId;
+						} else if ((node.parameters.options as IDataObject)?.path) {
+							(node.parameters.options as IDataObject).path = node.webhookId;
```
DawidMyslak: A few questions around this approach (I'm potentially missing some context), so let's clarify:

---

### 1. Removed node type check

This code applies to all nodes with a `webhookId`, but `resolveNodeWebhook` only updates paths for:
- `WEBHOOK_NODE_TYPE`  
- `FORM_TRIGGER_NODE_TYPE`
- `MCP_TRIGGER_NODE_TYPE`

Many trigger nodes have `webhookId` but use fixed paths (e.g., Slack Trigger uses `path: 'webhook'` hardcoded in its definition). This change could unintentionally modify nodes that shouldn't have their path parameter touched.

---

### 2. Path location handling

This code handles `node.parameters.options.path` as a fallback (I think it was like that before your change, let's double check), but `resolveNodeWebhook` doesn't. Looking at how webhook URLs are resolved in `useWorkflowHelpers.ts`:

```994:1000:packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
	function getWebhookPath(trigger: INode) {
		if (trigger.type === WEBHOOK_NODE_TYPE) {
			return (trigger.parameters.path as string) || (trigger.webhookId as string);
		}
		if (trigger.type === FORM_TRIGGER_NODE_TYPE) {
			return ((trigger.parameters.options as { path: string }) || {}).path ?? trigger.webhookId;
		}
```

Form Trigger v2+ uses `options.path`, so handling it makes sense, but this creates some inconsistency between `importWorkflowData` and `resolveNodeWebhook`.

DawidMyslak (changes requested):
Just wanted to ask a few clarification questions to this implementation 🙏.

packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
dlavrenuek: Thank you for the feedback @DawidMyslak. I had another round of thought and decided to add back the node type check when regenerating webhook url and only remove the duplicate check, other than that the logic of updating `path` either in parameters or in parameters.options is unchanged - other than adding optional check `?` in `(node.parameters.options as IDataObject)?.path` to prevent undefined access exceptions when `UPDATE_WEBHOOK_ID_NODE_TYPES` is extended with node types that do not have options.

If I see it correctly `resolveNodeWebhook` and the changed functionality serve different purposes. in `resolveNodeWebhook` there is a check `if (nodeTypeDescription.webhooks?.length && !node.webhookId) {` - so a new webhook id is only assigned when it is empty. If I see it correctly this is used for node creation, also you have already mentioned that it is not consistent because it does not check `options.path` that is used in forms. Do you think I should add `options.path` handling for `resolveNodeWebhook` in this PR?

dlavrenuek: Also to add some more context: if `path` is set it is used instead of the workflow id in the url. Therefore the path parameter must also be updated if we want to generate new webhook urls

DawidMyslak: > If I see it correctly this is used for node creation, also you have already mentioned that it is not consistent because it does not check options.path that is used in forms. Do you think I should add options.path handling for resolveNodeWebhook in this PR?

I think in this PR we should rather focus purely on the pasting fix and limit the scope of changes to minimum, so I'm happy you brought back the type check.

IMHO adding `options.path` to `resolveNodeWebhook` is a good idea but should be a separate follow-up PR to keep scope focused. Not an urgent action, looks like it's been like this for a long time.

DawidMyslak (approved):

packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
dlavrenuek: @dlavrenuek great, I fully agree with that

PR #24194, opened 2026-01-12
feat: Add public API testing to e2e tests

shortstacked:
## Summary

- Allows multiple user creation via public API
- Adds jti to jwt claim for uniqueness
- Added e2e tests to cover adding users/unique browser sessions

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/services/public-api-key.service.ts
```diff
@@ -171,7 +172,7 @@ export class PublicApiKeyService {
 		const nowInSeconds = Math.floor(Date.now() / 1000);
 
 		return this.jwtService.sign(
-			{ sub: user.id, iss: API_KEY_ISSUER, aud: API_KEY_AUDIENCE },
+			{ sub: user.id, iss: API_KEY_ISSUER, aud: API_KEY_AUDIENCE, jti: randomUUID() },
```
konstantintieber, line 175: What is the motivation for adding the [jti](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7) here?

shortstacked: When running the tests in parallel there was collision due to the tokens being created in the same second. So this adds more uniqueness.

konstantintieber (dismissed):
Code looks good to me 👍

konstantintieber (approved):

PR #24197, opened 2026-01-12
chore(core): Extract log streaming log streaming standalone handling of destinations

guillaumejacquart:
## Summary

This PR migrates the destination handling business logic from the MessageEventBus to the LogStreaming module (LogStreamingDestinationService)
The Log streaming service now listens to the event bus messages in and calls relevant destinations to forward the message.
I also removed the @OnPubSubEvent('restart-event-bus') handler on the Message Event Bus, as all the relevant logic - listening to pubsub event to reload destinations - is moved to the LogStreamingDestinationService

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/modules/log-streaming.ee/log-streaming-destination.service.ts
```diff
@@ -1,27 +1,51 @@
-import { Logger } from '@n8n/backend-common';
+import { LicenseState, Logger } from '@n8n/backend-common';
+import { OnPubSubEvent } from '@n8n/decorators';
 import { Service } from '@n8n/di';
 import type { DeleteResult } from '@n8n/typeorm';
+import type { MessageEventBusDestinationOptions } from 'n8n-workflow';
 
+import type { EventMessageTypes } from '@/eventbus';
```
phyllis-noester, line 7: hey hey quick question for my understanding: log streaming uses the general event bus logic meaning it consumes the events from the event bus and then forwards them to the log streaming destination, correct?

why does packages/cli/src/modules/log-streaming.ee/destinations/message-event-bus-destination.ee.ts live inside the log streaming module? isn't that more generic?

guillaumejacquart: Hey !
You're correct in your understanding of the flow (new flow since extracting the module)
message-event-bus-destination.ee.ts might seem generic or a contract interface, but it's "just" an abstract class to centralize some common behaviors for all destinations, such as circuit breaker mechanism. It allows concrete destination to implement a finer contract. Also, it's used only by the destination module, so it makes sense I think that it's there.
I think what can be confusing is that `MessageEventBusDestinationType` lives in the eventbus core folder (in message-event-bus.ts), but it's actually confusing, I'll change that, since it's also used only by the log streaming module

guillaumejacquart: @phyllis-noester I pushed a new commit to remove the `MessageEventBusDestinationType`, as the abstract is simple enough for concrete destinations to implement

packages/cli/test/integration/eventbus.ee.test.ts
```diff
@@ -306,7 +311,10 @@ test('should send message to sentry ', async () => {
 });
 
 test('DELETE /eventbus/destination delete all destinations by id', async () => {
-	const existingDestinationIds = [...Object.keys(eventBus.destinations)];
+	const existingDestinations = await destinationService.findDestination();
+	const existingDestinationIds = existingDestinations
```
BGZStephen: You could skip a loop (and some type funkiness) with a reduce:

```
const existingDestinationIds = existingDestinations.reduce<string[]>(
  (acc, d) => {
    if (d.id) {
      acc.push(d.id);
    }
    return acc;
  },
  []
);
```

packages/cli/src/controllers/e2e.controller.ts
```diff
@@ -282,9 +280,11 @@ export class E2EController {
 	}
 
 	private async resetLogStreaming() {
```
BGZStephen, line 282: Are partial failures a concern here? An error from one: `await this.logStreamingDestinationsService.removeDestination` would cause the remaining removals to fail

guillaumejacquart: It's only used by the e2e controller, so i guess it's ok (maybe even better for diagnostic) to have it fail after the first one

BGZStephen (approved):

PR #24215, opened 2026-01-13
fix: Fix CLI import

dariacodes:
## Summary

We forgot to add records to workflow history during workflow import. This led to having problems with adding workflow publish history records, but even if we remove these, the problem will still be with not being able to activate imported workflows because of missing workflow history records.

This PR always adds workflow history records and adds records to workflow publish history only when an existing workflow has an active version.

I'm creating new uuids for versions during import so that if this version already exists, we wouldn't update it, but rather create a new version with "import" as author.

I also fixed issues with trying to deactivate non-existent non-active workflows which led to console errors:
```
Importing 1 workflows...                                                                                                             
  Could not find workflow                                                                                                              
  Error: Could not find workflow                                                                                                       
  at ActiveWorkflowManager.clearWebhooks (/n8n/packages/cli/src/active-workflow-manager.ts:246:10)                    
  at ActiveWorkflowManager.remove (/n8n/packages/cli/src/active-workflow-manager.ts:883:4)          
```

Closes [ADO-4632](https://linear.app/n8n/issue/ADO-4632/bug-cli-import-still-broken)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/services/__tests__/import.service.test.ts
+++ b/packages/cli/src/services/__tests__/import.service.test.ts
@@ -1,10 +1,6 @@
 import { safeJoinPath, type Logger } from '@n8n/backend-common';
 import type { DatabaseConfig } from '@n8n/config';
-import type {
-	CredentialsRepository,
-	TagRepository,
-	WorkflowPublishHistoryRepository,
-} from '@n8n/db';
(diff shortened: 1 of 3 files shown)


CharlieKolb (dismissed):
🚀

CharlieKolb (approved):

PR #24216, opened 2026-01-13
refactor(core): Normalize exception attribute access in Python task runner

ivov:
https://linear.app/n8n/issue/CAT-2087


tomi (approved):
🚀

PR #24222, opened 2026-01-13
chore(core): Update turbo and use new config to exclude storybook from build

konstantintieber:
## Summary

Since turbo version 2.7.0 there is a new project-specific configuration availble, which allows us to configure on each project whether it should be included in a command that has a `dependsOn: ^command-name` config in the root turbo.json.

Link to docs: https://turborepo.dev/docs/reference/package-configurations#excluding-a-task-entirely

This way we also no longer need to provide `--filter=@n8n/storybook` when running the root level `pnpm run build`.

Until now, we haven't been using package-level turbo.json files in the n8n repo. I see leveraging these more as a big opportunity to improve developer experience.

Closes [LIGO-40](https://linear.app/n8n/issue/LIGO-40/fix-broken-build-command-on-latest-master-and-speed-it-up-by-excluding)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.github/workflows/test-visual-chromatic.yml
+++ b/.github/workflows/test-visual-chromatic.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Setup Node.js
         uses: ./.github/actions/setup-nodejs
         with:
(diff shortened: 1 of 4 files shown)


alexgrozav (approved):
🙌 Awesome work! Thanks.

PR #24232, opened 2026-01-13
refactor(core): Move N8N_BLOCK_ENV_ACCESS_IN_NODE to @n8n/config

seemewalkin:
## Summary

Moves the `N8N_BLOCK_ENV_ACCESS_IN_NODE` environment variable to `@n8n/config` package so it follows the same semantics as other env configs.

### Changes
- Add `blockEnvAccessInNode` property to `SecurityConfig`
- Update `createEnvProviderState()` to accept config value as parameter
- Inject `SecurityConfig` in `BaseExecuteContext` and pass to env provider
- Update breaking-changes rule to check if env var is explicitly set

https://linear.app/n8n/issue/CAT-1863

## Review / Merge checklist

- [ ] Tests included. 


packages/workflow/src/workflow-data-proxy-env-provider.ts
```diff
@@ -9,18 +9,22 @@ export type EnvProviderState = {
...
  */
-export function createEnvProviderState(): EnvProviderState {
+export function createEnvProviderState(isEnvAccessBlocked?: boolean): EnvProviderState {
 	const isProcessAvailable = typeof process !== 'undefined';
-	const isEnvAccessBlocked = isProcessAvailable
-		? process.env.N8N_BLOCK_ENV_ACCESS_IN_NODE !== 'false'
-		: false;
+	const envBlocked =
+		isEnvAccessBlocked ??
+		(isProcessAvailable ? process.env.N8N_BLOCK_ENV_ACCESS_IN_NODE !== 'false' : false);
```
tomi, line 21: Now that this config has been moved to the config package, we shouldn't read it from the env directly anymore. Also this function is a bit weird now as it takes the value as a param, but then it default to reading it from env. We should do one or the other, not both

PR #24262, opened 2026-01-13
feat(core): Add rate limiting to password reset endpoints (no-changelog)

frstavares:
## Summary

- Adds a new RateLimitService that supports both in-memory (single instance) and Redis (scaling mode) backends
- Implements per-email rate limiting for the forgot-password endpoint (3 requests/hour per email) to prevent abuse
- Adds rate limits to password-related endpoints: resolve-password-token and change-password (5 requests/hour)
- The service uses a "fail open" strategy on Redis errors to avoid blocking legitimate requests

Test plan
- Manual testing: Verify rate limiting works in single-instance mode
- Manual testing: Verify rate limiting works with Redis in scaling mode

https://linear.app/n8n/issue/IAM-84/add-rate-limiting-for-account-registration

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/controllers/password-reset.controller.ts
```diff
@@ -104,6 +106,25 @@ export class PasswordResetController {
...
+		});
+
+		if (!rateLimitStatus.allowed) {
+			this.logger.warn('Password reset rate limit exceeded for email', {
+				email,
+				remaining: rateLimitStatus.remaining,
+				resetAt: rateLimitStatus.resetAt,
+			});
+
+			return;
```
konstantintieber, line 125: Should we throw a `TooManyRequestsError` here instead of just returning?

packages/cli/src/services/rate-limit.service.ts
```diff
@@ -0,0 +1,321 @@
...
+}
+
+export interface RateLimitStatus {
+	allowed: boolean;
+	remaining: number;
+	resetAt: Date;
+}
+
+@Service()
+export class RateLimitService {
```
konstantintieber, line 26: Given that we're already using https://www.npmjs.com/package/express-rate-limit for decorator-based rate limiting, did you also explore using that one in combination with https://www.npmjs.com/package/rate-limit-redis?

Looking at the implementation on `/forgot-password` on this PR, it seems `express-rate-limit` and the `RateLimitService.tryConsume` now do rate limiting twice.

Where the express-rate-limit is using the IP address as key and the `RateLimitService` the provided string which enables rate limiting per e-mail address rather than client IP.

Out of curiosity: what does the extra rate limiting per user email add over already strict "3 requests per hour" per IP address?

konstantintieber: Update: now I know it's about abusing the endpoint to spam password reset emails to a user just by knowing their email.

packages/cli/src/services/rate-limit.service.ts
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	private async tryConsumeRedis({
+		key,
+		limit,
+		windowMs,
+	}: RateLimitOptions): Promise<RateLimitStatus> {
+		try {
+			// Note: We intentionally avoid using a Lua script here to keep the implementation simple.
+			// The trade-off is a small race window where if the process crashes between INCR and
```
afitzek, line 131: I think we are not implementing a sliding window algorithm here, but just start the window from the first instance. The whole key expires after the X amount of time after the initial entry was created. This can be fine, just wanted to highlight it. To implement sliding window, we would have to store the timestamps at which each event occurs and remove instances from that list on each check. Same goes for memory implementation ofc.

packages/cli/src/services/rate-limit.service.ts
afitzek: Does it make sense to try to reuse either the express-rate-limit middleware here, or vice versa start using this service as an implementation of the express-rate-limit middleware somehow? I would like to avoid having different implementations of the same thing as @konstantintieber mentions.

What would be awesome, is if we could register a function that receives the request object for example in the decorator to extract the key for this specific endpoint. This would allow us to have a single place where we handle rate limiting and we can also combine different factors in rate limiting. For example IP + email.

packages/cli/src/controllers/password-reset.controller.ts
```diff
@@ -42,12 +43,13 @@ export class PasswordResetController {
...
 		private readonly userRepository: UserRepository,
 		private readonly eventService: EventService,
+		private readonly rateLimitService: RateLimitService,
 	) {}
 
 	/**
 	 * Send a password reset email.
 	 */
-	@Post('/forgot-password', { skipAuth: true, rateLimit: { limit: 3 } })
+	@Post('/forgot-password', { skipAuth: true, rateLimit: { limit: 3, windowMs: 60 * 60 * 1000 } })
```
BGZStephen, line 52: For enterprise customers who use a VPN or sit within the same office / WAN that would show multiple users originating from the same IP, this rate limiting could cause some enterprise customers to become blocked resetting passwords.

I'd be somewhat inclined to move away from rate limiting for the endpoint by IP, and more in favour of tracking password reset generation in the DB, and adding a sliding window check for number of password reset generations between a given timeframe to determine if another one can be generated.

I would suggest not adding rate limiting in its current setup to the `resolve-password-token` and `change-password` mainly based on the comment above around shared IPs for VPN users and enterprise office based users on the same public IP

From a slightly higher level perspective, I have some reservations on adding rate limiting within the n8n repository in general, it's a very opinionated implementation with little confugrability for customers, which, while it would work for cloud, raises some questions certainly for self hosted.

If the only offering we had was for cloud, and we were purely a SaaS offering, having our own opinionated rate limiting would make sense, but it feels too rigid for self hosting.

To offer an alternative solution, I would encourage looking in to a proxy based approach for rate limiting, which doesn't sit within the n8n distributions, but can be used to add this type of rate limiting for cloud. For self hosted, I think it would be useful to provide this as an advisory step, and allow them control over their rate limiting strategy.

packages/cli/src/services/rate-limit.service.ts
konstantintieber: Claude proposed this implementation to achieve email-specific rate limiting with `express-rate-limit`:
<details>
<summary>Click to show code to extend decorator</summary>

```ts
Extend the decorator metadata to support custom key functions:

@Post('/forgot-password', {
  skipAuth: true,
  rateLimit: {
    limit: 3,
    windowMs: 60 * 60 * 1000,
    keyType: 'email',  // New option
    keyPath: 'body.email'  // JSON path to extract
  }
})
Update createRateLimitMiddleware to handle custom keys:

private createRateLimitMiddleware(rateLimit: RateLimit): RequestHandler {
  return expressRateLimit({
    windowMs: rateLimit.windowMs,
    limit: rateLimit.limit,
    keyGenerator: (req) => {
      if (rateLimit.keyType === 'email') {
        return req.body?.email?.toLowerCase() || req.ip;
      }
      return req.ip;  // Default
    },
    message: { message: 'Too many requests' },
  });
}
```

</details>

But I feel as long as we don't have more than just this email use case for a "special rate limit key", we could also just call `expressRateLimit` with the custom logic directly in the controller method, and remove the `rateLimit` config from the decorator used on the endpoint.

PR #24266, opened 2026-01-13
fix(core): Publish indicator not shown when lacking publish scope (no-changelog)

konstantintieber:
## Summary

This was discovered while working on https://github.com/n8n-io/n8n/pull/24111

With the most recent changes for shipping auto-save, the published indicator is not longer visible on published workflows when the active user does not have the workflow:publish permission.
This is not consistent, because in the workflow history and workflow list view, these users can see the published indicator.
And they should be able to see it to get the right picture anyway.

### Before

Indicator is visible on version history, but not on disabled publish button.

image

### After

Indicator is visible on disabled publish button too.

image

Closes https://linear.app/n8n/issue/LIGO-88/fe-fix-publish-button-doesnt-show-published-indicator-when-user-has-no

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
@@ -158,10 +158,8 @@ const publishButtonConfig = computed(() => {
 			tooltip: i18n.baseText('workflows.publish.permissionDenied'),
 			showVersionInfo: false,
 		};
-		const isWorkflowPublished = workflowsStore.workflow.activeVersion;
+		const isWorkflowPublished = !!workflowsStore.workflow.activeVersion;
 		if (isWorkflowPublished) {
-			defaultConfigForNoPermission.showIndicator = true;
-			defaultConfigForNoPermission.indicatorClass = 'published';
 			return {
 				...defaultConfigForNoPermission,
 				showIndicator: true,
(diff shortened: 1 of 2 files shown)


packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -150,14 +150,27 @@ const onPublishButtonClick = async () => {
...
-		return {
+		const defaultConfigForNoPermission = {
 			text: i18n.baseText('workflows.publish'),
 			enabled: false,
 			showIndicator: false,
 			indicatorClass: '',
 			tooltip: i18n.baseText('workflows.publish.permissionDenied'),
 			showVersionInfo: false,
 		};
+		const isWorkflowPublished = workflowsStore.workflow.activeVersion;
```
dariacodes: Could you make it `!== null` so that it is a true boolean?

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -150,14 +150,27 @@ const onPublishButtonClick = async () => {
...
+		if (isWorkflowPublished) {
+			defaultConfigForNoPermission.showIndicator = true;
+			defaultConfigForNoPermission.indicatorClass = 'published';
```
dariacodes: Why do we need to change these in the default config if we're setting them below in the return statement?

dariacodes (commented):
Thank you for tackling this! 🚀  It was definitely an oversight

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
konstantintieber: Yup, definitely duplicated logic. Will remove it.

dariacodes (approved):
🚢

PR #24267, opened 2026-01-13
feat(ai-builder): Add node recommendations for AI content generation

mike12345567:
## Summary

This PR updates the get best practices tool to allow collecting more than just best practices, it can now also provide node recommendations.

We do recommend some nodes as part of the best practices, but the goal of this work is to provide explicit individual nodes to use for the category being requested. Right now the categories are:
Text manipulation - we recommend using an AI Agent with an OpenAI chat model connected
For Video, Image and Audio generation we recommend using the OpenAI node, we list the particular operations that it should use as well.

I've created a specific structure for node recommendations, including recommendations for connected nodes (like the agent using the openAI chat model) to simplify the addition of other node recommendations and keeping the format very consistent.

Expanding/replacing the get best practices tool seemed like the most efficient method to do this and opens the way for further documentation to be added through it as well.

Addresses: https://linear.app/n8n/issue/AI-1674/feature-add-best-practices-section-for-ai-nodes-in-workflows-add-node

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/tools/test/get-best-practices.tool.test.ts
```diff
@@ -1,133 +0,0 @@
...
-
-// Mock LangGraph dependencies
-jest.mock('@langchain/langgraph', () => ({
-	getCurrentTaskInput: jest.fn(),
-	Command: jest.fn().mockImplementation((params: Record<string, unknown>) => ({
-		content: JSON.stringify(params),
-	})),
-}));
-
-describe('GetBestPracticesTool', () => {
```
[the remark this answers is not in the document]
mike12345567: The tool has been replaced by the get-documentation tool, which the tests have been moved into. The tests have not been removed, they've just been moved to cover the new tool instead.

packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/discovery.prompt.ts
```diff
@@ -289,6 +290,14 @@ const CRITICAL_RULES = `- NEVER ask clarifying questions
 - When user specifies a model name (e.g., 'gpt-4.1-mini') try to use this if it is a valid option
 - PREFER native n8n nodes (especially Edit Fields) over Code node`;
 
+const NODE_RECOMMENDATIONS_GUIDANCE = `When to include node_recommendations in get_documentation requests:
+- User mentions generic tasks like "generate image", "transcribe audio", "analyze text"
+- The user's request falls within one of the node recommendation categories: ${Object.values(RecommendationCategory).join(', ')}
+
+Do NOT request node_recommendations when:
```
harshalnpatil, line 297: Isn't it possible that the user is very specific about generating an image with openai, but not specific about how to analyze the text? In that case, the same prompt has one thing that needs a node recommendation and another that does not.

packages/@n8n/ai-workflow-builder.ee/src/prompts/shared/node-recommendations/audio-generation.ts
```diff
@@ -0,0 +1,19 @@
+import {
+	type NodeRecommendationDocument,
+	RecommendationCategory,
+} from '@/types/node-recommendations';
+
+export const audioGenerationRecommendation: NodeRecommendationDocument = {
+	category: RecommendationCategory.AUDIO_GENERATION,
+	version: '1.0.0',
+	recommendation: {
+		defaultNode: '@n8n/n8n-nodes-langchain.openAi',
```
harshalnpatil, line 10: Earlier, we had these node recommendations in the techniques themselves. Why did we need to move it out now? Does this result in less input tokens or higher eval scores than adding this info to the technique docs?

mike12345567: Adding it to the techniques means adding it to every best practice that its applicable to - while this is content generation, I think it makes sense to break out these node recommendations into their own state. Text manipulation is the example, its not specific to one of the best practices, it could be applicable to a few.

I've found this is respected quite a bit more as we can state this explicit recommendation should be followed, rather than it being part of a larger document, with other node recommendations in it as well. This creates a nice explicit structure we can use, no risk of the prompt differing between recommendations.

packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/discovery.prompt.ts
mike12345567: True, I'll add a note that it should consider it for each of the categories it detects in the user prompt.

packages/@n8n/ai-workflow-builder.ee/src/prompts/shared/node-recommendations/audio-generation.ts
harshalnpatil: got you. could you measure input tokens or eval score diff?

mike12345567: Added the eval scores to the ticket, there is no difference in the eval scores for spec evals this would impact.

packages/@n8n/ai-workflow-builder.ee/src/types/recommendation-category.ts
aalises, line 1: Empty file, can be deleted?

packages/@n8n/ai-workflow-builder.ee/src/tools/get-documentation.tool.ts
```diff
@@ -0,0 +1,230 @@
...
+	if (foundRecs.length === 0) {
+		return '';
+	}
+
+	return prompt().section('node_recommendations', foundRecs.join('\n')).build();
+}
+
+export const GET_DOCUMENTATION_TOOL: BuilderToolBase = {
+	toolName: 'get_documentation',
+	displayTitle: 'Getting best practices',
```
aalises: Perhaps we should update the display title with the updated functionality

packages/@n8n/ai-workflow-builder.ee/src/tools/test/get-documentation.tool.test.ts
```diff
@@ -0,0 +1,196 @@
...
+				},
+				mockConfig,
+			);
+			const result = getMessageContent(command);
+
+			expect(result).toContain('<connected_nodes>');
+			expect(result).toContain('lmChatOpenAi');
+			expect(result).toContain('ai_languageModel');
+		});
+	});
```
aalises, line 169: We can add a test to account for the case when no documentation is found (e.g. empty results scenario)

aalises (commented):
Love the refactor of the tool! it allows us for more extensibility on different types of documentation that we want to pass to the discovery prompt to inform the decision to include or not the node.

Added some minor comments, not a blocker at all

aalises (dismissed):

aalises (approved):

PR #24279, opened 2026-01-13
feat(editor): Add user_message_id to AI assistant feedback telemetry

aalises:
## Summary                                                                
  - Exposes `lastUserMessageId` from the builder store                      
  - Includes `user_message_id` in the feedback telemetry event to track which user message the feedback is associated with
  ---
  Related ticket: https://linear.app/n8n/issue/AI-1897/telemetry-add-user-message-id-to-feedback-table

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/AskAssistantBuild.vue
+++ b/packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/AskAssistantBuild.vue
@@ -157,6 +157,7 @@ function onFeedback(feedback: RatingFeedback) {
 			feedback: feedback.feedback,
 			workflow_id: workflowsStore.workflowId,
 			session_id: builderStore.trackingSessionId,
+			user_message_id: builderStore.lastUserMessageId,
 		});
 	}
 }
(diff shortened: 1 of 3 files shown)


packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/AskAssistantBuild.vue
```diff
@@ -157,6 +157,7 @@ function onFeedback(feedback: RatingFeedback) {
 			feedback: feedback.feedback,
 			workflow_id: workflowsStore.workflowId,
 			session_id: builderStore.trackingSessionId,
+			user_message_id: builderStore.lastUserMessageId,
```
mutdmour, line 160: I think I called it `last_user_message_id` in another event but maybe it does not matter

mutdmour (approved):

packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/AskAssistantBuild.vue
aalises: Just checked and we use both names in different events. Do you have any preference @mutdmour ? I think in that case `user_message_id` might be alright as it's feedback about a specific message (does not need to necessarily be the "last")?

mutdmour: feedback usually only relates to the last message so I think it's fine

PR #24290, opened 2026-01-14
feat(core): Add drain timeout for runners

ivov:
Allow graceful shutdown of active tasks across runners.


packages/@n8n/task-runner-python/src/task_runner.py
```diff
@@ -239,13 +240,20 @@ async def _handle_message(self, message: BrokerMessage) -> None:
...
                 self.logger.warning(f"Unhandled message type: {type(message)}")
 
     async def _handle_info_request(self) -> None:
         response = RunnerInfo(name=self.name, types=[TASK_TYPE_PYTHON])
         await self._send_message(response)
 
+    async def _handle_drain(self) -> None:
+        self.can_send_offers = False
+        await self._cancel_coroutine(self.offers_coroutine)
+        self.logger.info("Received drain signal, stopped accepting new tasks")
```
johnniee, line 255: `drainTimeout` variable could be benifital here

packages/@n8n/config/src/configs/runners.config.ts
```diff
@@ -70,6 +70,13 @@ export class TaskRunnersConfig {
 	@Env('N8N_RUNNERS_HEARTBEAT_INTERVAL')
 	heartbeatInterval: number = 30;
 
+	/**
+	 * How long (in seconds) to wait for running tasks to complete during shutdown
+	 * before force-closing connections. Must be greater than or equal to 0.
+	 */
+	@Env('N8N_RUNNERS_DRAIN_TIMEOUT')
```
tomi: I don't think we need another timeout for this. We can use e.g. 80-90% of what graful sthudown timeout is

packages/cli/src/task-runners/task-broker/task-broker-ws-server.ts
```diff
@@ -178,8 +178,30 @@ export class TaskBrokerWsServer {
...
+		const start = Date.now();
+		while (this.taskBroker.hasActiveTasks() && Date.now() - start < drainTimeoutMs) {
+			await sleep(100);
+		}
+
+		if (this.taskBroker.hasActiveTasks()) {
+			this.logger.warn(
+				`Drain timeout reached after ${drainTimeout}s, will force-shutdown with active tasks...`,
+			);
+		}
```
tomi, line 216: nit: The draining part could be it's own function

tomi (approved):
🚀

PR #24299, opened 2026-01-14
fix(core): Allow resetting Error Workflow settings to default

dariacodes:
## Summary

There was a bug happening when trying to reset error workflow settings back to the default value - the UI showed that the request was successful, but after page reload you could see that the error workflow settings was kept intact.

- The main change is in `WorkflowSettings.vue`. 
- I also moved cleaning up settings to a util and removed object mutation. 
- Moved `settingsChanged` assignment to a better place, after the clean-up (it didn't change the behavior for now, but could have caused issues in the future).
- Added tests.

Closes [ADO-4640](https://linear.app/n8n/issue/ADO-4640/workflow-publish-option-isnt-being-triggered-for-workflow-settings)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/__tests__/workflow-helpers.test.ts
+++ b/packages/cli/src/__tests__/workflow-helpers.test.ts
@@ -1,5 +1,6 @@
 import { mockInstance } from '@n8n/backend-test-utils';
 import type { Project, Variables } from '@n8n/db';
+import type { IWorkflowSettings } from 'n8n-workflow';
 
 import { VariablesService } from '@/environments.ee/variables/variables.service.ee';
 import { OwnershipService } from '@/services/ownership.service';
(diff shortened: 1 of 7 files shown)


packages/cli/src/__tests__/workflow-helpers.test.ts
```diff
@@ -97,3 +100,82 @@ describe('shouldRestartParentExecution', () => {
 		expect(shouldRestartParentExecution(parentExecution)).toBe(true);
 	});
 });
+
+describe('removeDefaultValues', () => {
+	const DEFAULT_EXECUTION_TIMEOUT = 3600;
+
+	it('should remove errorWorkflow when set to DEFAULT', () => {
+		const settings = {
+			errorWorkflow: 'DEFAULT' as const,
```
CharlieKolb: [nit] `const DEFAULT = 'DEFAULT' as const` probably makes this section a bit prettier :D

CharlieKolb (dismissed):

CharlieKolb (approved):

PR #24322, opened 2026-01-14
fix(ai-builder): Filter LLM internal reasoning tags from chat output

mutdmour:
## Summary
- Enhanced `cleanContextTags()` to filter LLM internal reasoning XML tags from user-facing messages
- Applied filtering in `processAgentNodeUpdate()` before messages are streamed to users
- Added unit tests for the new tag filtering

## Issue
Fixes https://linear.app/n8n/issue/AI-1894/bug-multi-agent-exposes-intermediate-steps-as-xml-tags

## Root Cause
Claude/Anthropic models sometimes output internal reasoning in XML format during complex tasks (e.g., `<attempt_search>`, `<search_query>`, `<attempt_completion>`). These tags are emergent LLM behavior, not n8n-defined, and were leaking through to users.

## Test plan
- [x] Unit tests added for all filtered tag patterns
- [x] Typecheck passes
- [ ] Manual test with prompts that trigger web search reasoning

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/src/utils/stream-processor.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/src/utils/stream-processor.ts
@@ -144,11 +144,38 @@ function extractMessageContent(messages: MessageContent[]): string | null {
  * Remove context tags from message content that are used for AI context
  * but shouldn't be displayed to users.
  *
- * This removes the entire context block from <current_workflow_json> through
- * </current_execution_nodes_schemas>
+ * This removes:
+ * 1. The workflow context block from <current_workflow_json> through </current_execution_nodes_schemas>
+ * 2. LLM internal reasoning tags (AI-1894) - Claude sometimes outputs these XML tags
+ *    as part of its internal planning/reasoning process
  */
 export function cleanContextTags(text: string): string {
-	return text.replace(/\n*<current_workflow_json>[\s\S]*?<\/current_execution_nodes_schemas>/, '');
+	// Remove workflow context block
+	let cleaned = text.replace(
+		/\n*<current_workflow_json>[\s\S]*?<\/current_execution_nodes_schemas>/,
+		'',
+	);
+
+	// Remove LLM internal reasoning XML tags (AI-1894)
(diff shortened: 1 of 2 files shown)


packages/@n8n/ai-workflow-builder.ee/src/utils/stream-processor.ts
```diff
@@ -144,11 +144,38 @@ function extractMessageContent(messages: MessageContent[]): string | null {
...
+	// Remove LLM internal reasoning XML tags (AI-1894)
+	// These are specific tags mentioned in the issue that Claude outputs during complex reasoning
+	const reasoningTagPatterns = [
+		/<attempt_completion>[\s\S]*?<\/attempt_completion>/g,
+		/<result>[\s\S]*?<\/result>/g,
+		/<attempt_search>[\s\S]*?<\/attempt_search>/g,
+		/<query>[\s\S]*?<\/query>/g,
+		/<search_type>[\s\S]*?<\/search_type>/g,
+		/<search_query>[\s\S]*?<\/search_query>/g,
+		/<search_focus>[\s\S]*?<\/search_focus>/g,
```
OlegIvaniv, line 168: Wouldn't this also remove these strings from user message? We're using same `cleanContextTags` in `formatHumanMessage`. Imo this should be a different function, the original was just about stripping system-injected workflow context (applies to HumanMessage), here we're stripping  LLM reasoning artifacts (should ONLY apply to AIMessage). But also,  how do we know <result>some example here</result> is Claude's reasoning and not legitimate response to user? `<result>` and `<query>` seem way too generic

OlegIvaniv (changes requested):

packages/@n8n/ai-workflow-builder.ee/src/utils/stream-processor.ts
mutdmour: Thanks for the catch. 

I guess we don't have a way to know. 
Only other solution I could think of is render any XML tag Claude reuturns  on the FE but collapsed?

mutdmour:
closing in favor of https://github.com/n8n-io/n8n/pull/24457

PR #24324, opened 2026-01-14
chore(core): Introduce CORS settings for dynamic credential interface

afitzek:
## Summary

This adds CORS support for dynamic credentials interface, in case customers need to directly interface with n8n from the frontend.

closes https://linear.app/n8n/issue/IAM-16/add-cors-settings-for-new-api

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
@@ -1,0 +1,20 @@
+import { Config, Env } from '@n8n/config';
+
+@Config
+export class DynamicCredentialsConfig {
+	/**
+	 * Comma-separated list of allowed CORS origins for dynamic credentials endpoints.
+	 * Example: 'https://app.example.com,https://admin.example.com'
+	 * When null, CORS is disabled and endpoints return 404 for preflight requests.
+	 */
+	@Env('N8N_DYNAMIC_CREDENTIALS_CORS_ORIGIN')
+	corsOrigin: string | null = null;
+
+	/**
+	 * Whether to allow credentials (cookies, authorization headers) in CORS requests.
+	 * Must be false when using wildcard (*) in N8N_DYNAMIC_CREDENTIALS_CORS_ORIGIN.
+	 * Default: false
+	 */
+	@Env('N8N_DYNAMIC_CREDENTIALS_CORS_ALLOW_CREDENTIALS')
+	corsAllowCredentials: boolean = false;
+}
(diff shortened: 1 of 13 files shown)


packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
```diff
@@ -0,0 +1,20 @@
+import { Config, Env } from '@n8n/config';
+
+@Config
+export class DynamicCredentialsConfig {
+	/**
+	 * Comma-separated list of allowed CORS origins for dynamic credentials endpoints.
+	 * Example: 'https://app.example.com,https://admin.example.com'
+	 * When empty CORS is disabled and endpoints return 404 for preflight requests.
+	 */
+	@Env('N8N_DYNAMIC_CREDENTIALS_CORS_ORIGIN')
```
BGZStephen, line 10: nit: `N8N_DYNAMIC_CREDENTIALS_CORS_ORIGINS` if it is expected to be multiple.

It would also be really nice not to have to redeploy n8n to update these, I wonder if it's worth the pain of moving these to an overwriteable config setting?

BGZStephen (approved):
This seems generally like a reasonable approach. My only real concern is providing domains via ENV but not the UI could make it cumbersome for workflow builders who maybe are not the deployers of the system to get set up quickly / make changes easily.

packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
```diff
@@ -0,0 +1,20 @@
...
+	corsOrigin: string = '';
+
+	/**
+	 * Whether to allow credentials (cookies, authorization headers) in CORS requests.
+	 * Must be false when using wildcard (*) in N8N_DYNAMIC_CREDENTIALS_CORS_ORIGIN.
+	 * Default: false
+	 */
+	@Env('N8N_DYNAMIC_CREDENTIALS_CORS_ALLOW_CREDENTIALS')
+	corsAllowCredentials: boolean = false;
```
phyllis-noester, line 19: does it make sense to have this as a separate env var? We could just set this to true if a static auth token is defined (https://linear.app/n8n/issue/IAM-14/add-static-authentication-for-new-api-endpoints)

afitzek: The static credentials should probably never be used in a browser, where CORS is relevant, since this static credential should protect from M2M communication. Controlling this setting allows for controlling if the browser is allowed to include credentials like cookies, client tls authentication or basic authentication headers (username and password) for the target domain. So If someone sends their access token via a custom header in the fetch request, I think it makes sense to allow the operators to control this to lock down the system.

But if we opt for moving these settings into a over writeable setting, than we might not want to keep this setting as an envvar.

packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
afitzek: I think you are right, controlling this dynamically would be nicer, but I would argue that we can keep this simple as an envvar for the alpha. If the need arises we can still migrate this.

BGZStephen: Yep, it was more an observation as a nice to have for GA

PR #24334, opened 2026-01-14
fix(editor): Hold autosave when new template is imported until the user makes a change/run the workflow

scdekov:
## Summary

Update the import workflow functionalities to not set the workflow state as dirty which fixes autosave being triggered incorrectly.

https://linear.app/n8n/issue/ADO-4636/inserted-template-is-being-saved-immediately

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/testing/playwright/tests/e2e/workflows/templates/templates.spec.ts
+++ b/packages/testing/playwright/tests/e2e/workflows/templates/templates.spec.ts
@@ -228,9 +228,10 @@ test.describe('Workflow templates', () => {
 		test('should save template id with the workflow', async ({ n8n }) => {
 			await n8n.templatesComposer.importFirstTemplate();
 
-			// Trigger manual save since imported templates don't auto-save immediately
+			// Execute workflow to trigger autosave (imported templates don't auto-save immediately)
+			await n8n.canvas.hitExecuteWorkflow();
+
 			const saveResponsePromise = n8n.canvas.waitForSaveWorkflowCompleted();
-			await n8n.canvas.hitSaveWorkflow();
 			const saveResponse = await saveResponsePromise;
 
 			const requestBody = saveResponse.request().postDataJSON();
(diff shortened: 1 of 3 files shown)


packages/testing/playwright/tests/e2e/workflows/templates/templates.spec.ts
```diff
@@ -228,7 +228,11 @@ test.describe('Workflow templates', () => {
 		test('should save template id with the workflow', async ({ n8n }) => {
 			await n8n.templatesComposer.importFirstTemplate();
 
-			const saveResponse = await n8n.canvas.waitForSaveWorkflowCompleted();
+			// Execute workflow to trigger autosave (imported templates don't auto-save immediately)
+			await n8n.canvas.hitExecuteWorkflow();
```
dariacodes, line 232: I feel like putting here an execution call will make this test more prone to flakiness. Could you just rename the workflow and wait for the save to complete after it?

dariacodes (approved):
Code looks good! Just left a small comment about the e2e

packages/testing/playwright/tests/e2e/workflows/templates/templates.spec.ts
scdekov: I've tried renaming as well as moving a node around but it doesn't work as the loading overlay is still there and I was not able to find a way to wait for this one to be hidden. Have you dealt with that in the past?

dariacodes: have you tried doing this?
```js
      await this.n8n.canvas.setWorkflowName('newName');
      await this.n8n.page.keyboard.press('Enter');
      await this.n8n.canvas.waitForSaveWorkflowCompleted();
```

dariacodes: by the way, could you also remove `hitSaveWorkflow` from the codebase? 🙏

scdekov: I've used `hitSaveWorkflow` in the previous commit but it didn't work for the same reason as the renaming doesn't which is that the loading overlay is shown.

PR #24338, opened 2026-01-14
ci: GH actions for private fork

ivov:
Add GH actions to manage the private security fork:

- On PR merge to private's master, cherry-picks the security fix onto public's master, then syncs back to private
- Regularly keep private in sync with public, can be triggered manually for conflict recovery

### DIFF
--- a/.github/workflows/sec-publish-fix.yml
+++ b/.github/workflows/sec-publish-fix.yml
@@ -54,6 +54,8 @@ jobs:
           git remote add public https://x-access-token:${{ steps.generate_token.outputs.token }}@github.com/n8n-io/n8n.git
           git fetch public master
           git checkout public/master
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git cherry-pick "$MERGE_COMMIT"
           git push public HEAD:master
 
--- a/.github/workflows/sec-publish-fix.yml
+++ b/.github/workflows/sec-publish-fix.yml
@@ -51,18 +51,18 @@ jobs:
       - name: Cherry-pick to public repo
         run: |
           COMMIT_TO_PUBLISH=$(git rev-parse HEAD)
-          git remote add public https://x-access-token:${{ steps.generate_token.outputs.token }}@github.com/n8n-io/n8n.git
-          git fetch public master
-          git checkout public/master
+          git remote add public-repo https://x-access-token:${{ steps.generate_token.outputs.token }}@github.com/n8n-io/n8n.git
+          git fetch public-repo master
+          git checkout public-repo/master
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git cherry-pick "$COMMIT_TO_PUBLISH"
-          git push public HEAD:master
+          git push public-repo HEAD:master
 
       - name: Sync public back to private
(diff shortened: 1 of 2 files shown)


.github/workflows/sec-publish-fix.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          app-id: ${{ secrets.N8N_ASSISTANT_APP_ID }}
+          private-key: ${{ secrets.N8N_ASSISTANT_PRIVATE_KEY }}
+          owner: n8n-io
+          repositories: n8n
+
+      - name: Cherry-pick to public repo
+        run: |
+          MERGE_COMMIT=$(git rev-parse HEAD)
+          git remote add public https://x-access-token:${{ steps.generate_token.outputs.token }}@github.com/n8n-io/n8n.git
```
tomi: Is this the syntax you define a remote with the access token?

.github/workflows/sec-publish-fix.yml
```diff
@@ -0,0 +1,70 @@
...
+          git fetch public master
+          git checkout public/master
+          git cherry-pick "$MERGE_COMMIT"
+          git push public HEAD:master
+
+      - name: Sync public back to private
+        run: git push origin HEAD:master --force
```
tomi: Do we need to force push? If everything went OK, it should be fine. Otherwise if there has been a merge/commit in public repo in-between, we will overwrite. I think a better strategy would be to retry the sync+cherrypick+push

.github/workflows/sec-sync-public-to-private.yml
```diff
@@ -0,0 +1,54 @@
...
+            elif [ "${{ inputs.force }}" != "true" ]; then
+              echo "Private is $AHEAD_COUNT commit(s) ahead of public, skipping (force not enabled)"
+              exit 0
+            else
+              echo "Private is $AHEAD_COUNT commit(s) ahead of public, force syncing anyway"
+            fi
+          fi
+
+          git reset --hard public-master
+          git push origin master --force
```
tomi: Maybe `--force-with-lease` to be a little safer

.github/workflows/sec-publish-fix.yml
ivov: Based on the [docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation), yes

.github/workflows/sec-publish-fix.yml
tomi: `COMMIT_TO_PUBLISH` could be a more meaningful name for this

.github/workflows/sec-publish-fix.yml
```diff
@@ -54,11 +54,15 @@ jobs:
...
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git cherry-pick "$MERGE_COMMIT"
           git push public HEAD:master
 
       - name: Sync public back to private
-        run: git push origin HEAD:master --force
+        run: |
+          git fetch public master
+          git push origin public/master:master --force
```
tomi: I found it very difficult to follow what's happening here, what is a branch and what is an origin. Maybe we could rename the origin to `public-repo` to make it clearer?

shortstacked (approved):
🥇

PR #24348, opened 2026-01-14
 feat(ai-builder): Add rename_node tool to AI workflow builder (no-changelog)

aalises:
## Summary                                                                
                                                                            
Add a `rename_node` tool to the AI workflow builder that allows the  assistant to rename nodes in a workflow. This follows the same  implementation pattern as the `remove_connection` tool (PR #20439).       
                                                                            
### The tool:                                                                 
- Validates the node exists and the new name doesn't conflict with  existing nodes                                                            
- Creates a `renameNode` operation that updates both the node's name, all connection references, and expressions (since connections use node names as identifiers and expressions mention the old name)                                                              
- Reports progress during execution                                       
                                                                            
### Changes

- **New tool**: `rename-node.tool.ts` with Zod schema validation and progress reporting.
- **New operation**: `renameNode` type in workflow operations.
- **Operation handler**: Updates node name and all connection references atomically.
- **Tests**: 10 unit tests for the tool and 8 tests for the operation processor.
- **Expression Logic**: Added expression update logic to `applyRenameNodeOperation()` in `operations-processor.ts`, following the same pattern used in the core `n8n-workflow` package (`Workflow.renameNode()`).

The fix handles all node reference syntaxes:
- `$('Node Name')` / `$("Node Name")`
- `$node['Node Name']` / `$node["Node Name"]`
- `$node.NodeName`
- `$items('Node Name', n)`

### Special Node Types

Some nodes contain code or HTML that can reference other nodes without the `=` expression prefix. These require special handling:

| Node Type | Parameter | Content |
| :--- | :--- | :--- |
| **Code, Function, AI Transform** | `jsCode` | JavaScript code |
| **HTML, Mailgun** | `html` | HTML content |
| **Form** | `formFields.values[].html` | HTML in form fields |

### Implementation Details

- **Imports**: Imported `applyAccessPatterns` and node type constants from `n8n-workflow`.
- **Helpers**: Added `renameNodeInParameterValue()` helper for recursive parameter traversal.
- **Updates**: Updated `applyRenameNodeOperation()` to process expressions in all nodes.
- **Coverage**: Added tests covering all expression patterns and special node types.
  
  ---
  _related ticket: https://linear.app/n8n/issue/AI-1517/feature-rename-node-tool_


packages/@n8n/ai-workflow-builder.ee/src/utils/operations-processor.ts
```diff
@@ -271,6 +326,151 @@ function applySetNameOperation(
 	};
 }
 
+/**
+ * Recursively update node name references in parameter values
+ */
+function renameNodeInParameterValue(
```
OlegIvaniv: We already have this method in `n8n-workflow` package, why can't we reuse it?

packages/workflow/src/index.ts
```diff
@@ -84,6 +84,7 @@ export * from './node-parameters/filter-parameter';
 export * from './node-parameters/parameter-type-validation';
 export * from './node-parameters/node-parameter-value-type-guard';
 export * from './node-parameters/path-utils';
+export { renameFormFields } from './node-parameters/rename-node-utils';
```
burivuhster: Do we still need this?

burivuhster (dismissed):
Solid work! LGTM

packages/workflow/src/index.ts
aalises: true, deleted!

burivuhster (dismissed):

OlegIvaniv (approved):

PR #24359, opened 2026-01-14
test: Refactor sharing spec tests to break up isolation 

shortstacked:
## Summary

Refactors the monolithic `sharing.spec.ts` E2E test file into 4 focused, parallel-safe test files:

- **credential-sharing.spec.ts** - Tests for sharing credentials between users
- **workflow-sharing.spec.ts** - Tests for sharing workflows between users  
- **credential-visibility.spec.ts** - Credential dropdown visibility rules in different contexts
- **access-control.spec.ts** - Permission boundaries and access control enforcement

### Key improvements

- **Removed `@isolated` dependency** - Tests now run in parallel using modern isolation patterns
- **Uses `withUser()`** - Isolated browser contexts per user (no session bleeding)
- **API-based setup** - Fast, reliable test data creation with thin UI verification
- **`nanoid` naming** - Unique resource names for parallel safety
- **Reduced isolated shards** - From 2 → 1 in CI (credential-visibility moved out)
- **Updated test metrics** - Realistic durations for proper shard balancing

### CI Impact

- Tests run ~10x faster (was ~5-9 min serial, now ~30-130s parallel)
- Better shard distribution across 15 shards
- Reduced container overhead

https://linear.app/n8n/issue/CAT-2089

## Review / Merge checklist

- [x] Tests included.


mfsiega (commented):
In general this seems like a good change, so I'm basically on board.

I'm struggling with the size of the PR, and the fact that it doesn't seem like a mechanical refactor that just moves things around. So for example: in line 175 of the deleted sharing.spec.ts, there's an assertion that verifies that when you open a credential that's been shared with you, it should automatically be tested. I can't find that assertion in the new tests; is it there somewhere, or was it not needed?

In general, I'm having a hard time figuring out whether this PR loses any test coverage. Is there some strategy I can apply beyond just going test-by-test?

mfsiega (commented):
In the specific case of the auto-test, it's not clear to me if it is related to sharing - I guess the auto-test is triggered when you have a credential shared with you, which seems sharing related?

More generally - my concern is just that we might be losing test coverage here, which would be a negative change. Is there a way we can get to our end goal such that it's clear that we're not losing coverage? Or is there something I'm missing that gives you confidence we're not losing coverage here?

mfsiega (approved):
Approved based on the overall strategy, and I don't want to hold this work up.

I haven't reviewed line-by-line.

PR #24380, opened 2026-01-15
feat: Add new scopes + migration for splitting view and use credentials

BGZStephen:
## Summary

This feature splits out read and use for credential scopes to allow for more fine grained usage in custom roles.

IAM-77

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/evaluations/__tests__/cli.test.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/evaluations/__tests__/cli.test.ts
@@ -14,6 +14,7 @@ import type { SimpleWorkflow } from '@/types/workflow';
 
 // Store mocks for dependencies
 const mockParseEvaluationArgs = jest.fn();
+const mockArgsToStageModels = jest.fn();
 const mockSetupTestEnvironment = jest.fn();
 const mockCreateAgent = jest.fn();
 const mockGenerateRunId = jest.fn();
@@ -30,6 +31,7 @@ const mockCreatePairwiseEvaluator = jest.fn();
 // Mock all external modules
 jest.mock('../cli/argument-parser', () => ({
 	parseEvaluationArgs: (): unknown => mockParseEvaluationArgs(),
+	argsToStageModels: (...args: unknown[]): unknown => mockArgsToStageModels(...args),
 	getDefaultDatasetName: (suite: unknown): unknown =>
 		suite === 'pairwise' ? 'notion-pairwise-workflows' : 'workflow-builder-canvas-prompts',
 	getDefaultExperimentName: (suite: unknown): unknown =>
@@ -98,9 +100,19 @@ function createMockArgs(overrides: Record<string, unknown> = {}) {
 
 /** Helper to create mock environment */
 function createMockEnvironment() {
+	const mockLlm = mock<BaseChatModel>();
 	return {
 		parsedNodeTypes: [] as INodeTypeDescription[],
-		llm: mock<BaseChatModel>(),
+		llms: {
+			default: mockLlm,
+			supervisor: mockLlm,
+			responder: mockLlm,
+			discovery: mockLlm,
+			builder: mockLlm,
+			configurator: mockLlm,
+			parameterUpdater: mockLlm,
+			judge: mockLlm,
+		},
 		lsClient: mock<Client>(),
 	};
 }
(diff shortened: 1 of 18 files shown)


packages/@n8n/db/src/migrations/common/1768475341537-AddCredentialUseScopeToRoles.ts
```diff
@@ -0,0 +1,102 @@
...
+		const descriptionColumn = escape.columnName('description');
+
+		const roleTableName = escape.tableName('role');
+		const roleScopeTableName = escape.tableName('role_scope');
+		const roleSlugColumn = escape.columnName('slug');
+		const roleScopeRoleSlugColumn = escape.columnName('roleSlug');
+		const roleScopeScopeSlugColumn = escape.columnName('scopeSlug');
+
+		// Step 1: Ensure credential:use scope exists
+		const isPostgresOrSqlite = dbType === 'postgresdb' || dbType === 'sqlite';
```
guillaumejacquart: Do we want to backport to v1? otherwise I think we're supposed to ignore mysql entirely

packages/@n8n/db/src/migrations/common/1768475341537-AddCredentialUseScopeToRoles.ts
```diff
@@ -0,0 +1,102 @@
...
+	async down({ escape, runQuery, logger }: MigrationContext) {
+		const roleScopeTableName = escape.tableName('role_scope');
+		const roleScopeRoleSlugColumn = escape.columnName('roleSlug');
+		const roleScopeScopeSlugColumn = escape.columnName('scopeSlug');
+
+		// Remove credential:use scopes only from roles that also have credential:read
+		// This is the exact inverse of what up() does, ensuring we don't remove
+		// manually granted credential:use scopes from roles without credential:read.
+		// Using EXISTS for cross-database compatibility.
+		const deleteQuery = `
```
guillaumejacquart: What happens with the role_scope with scope = 'credential:use' without a matching 'credential:read' ? 
also we keep 'credential:use' in the scope table ?

packages/cli/src/credentials/credentials-finder.service.ts
```diff
@@ -31,10 +31,12 @@ export class CredentialsFinderService {
 
 	/**
 	 * Checks if the scopes allow read-only access to global credentials.
-	 * Global credentials can be accessed with credential:read scope only.
+	 * Global credentials can be accessed with credential:read or credential:use scope.
 	 */
 	hasGlobalReadOnlyAccess(scopes: Scope[]): boolean {
-		return scopes.length === 1 && scopes[0] === 'credential:read';
+		return (
+			scopes.length === 1 && (scopes[0] === 'credential:read' || scopes[0] === 'credential:use')
```
guillaumejacquart: Not sure I fully get it, but are we saying that global credentials are usable by anyone ? like if we are querying a credential id with either only 'credential:read' or 'credential:use' scope, then this will return true and allow using the global credential ?
Also, what if the 2 scopes are there (I could not find a scenario where it would be the case but just in case)

packages/cli/src/credentials/credentials-finder.service.ts
```diff
@@ -75,6 +77,72 @@ export class CredentialsFinderService {
...
+					? this.roleService.rolesWithScope('credential', ['credential:read'])
+					: Promise.resolve([]),
+				shouldCheckUse && (shouldCheckRead || shouldCheckReadForOtherScopes)
+					? this.roleService.rolesWithScope('credential', ['credential:use'])
+					: Promise.resolve([]),
+				// For other credential scopes (like update, delete, share), get roles with those scopes
+				hasOtherCredentialScopes
+					? this.roleService.rolesWithScope(
+							'credential',
+							scopes.filter(
```
guillaumejacquart: Might be worth extracting this into a variable so that it's used for the hasOtherCredentialScopes computation

packages/@n8n/db/src/migrations/common/1768475341537-AddCredentialUseScopeToRoles.ts
BGZStephen: technically yes, as it is the *opposite* of the action we take, which is to only add it where read exists

PR #24391, opened 2026-01-15
fix: Apply source control configuration changes to all multi main instances

ireneea:
## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/environments.ee/source-control/source-control.controller.ee.ts
```diff
@@ -3,7 +3,7 @@ import { PullWorkFolderRequestDto, PushWorkFolderRequestDto } from '@n8n/api-typ
 import type { SourceControlledFile } from '@n8n/api-types';
 import { AuthenticatedRequest } from '@n8n/db';
 import { Get, Post, Patch, RestController, GlobalScope, Body } from '@n8n/decorators';
-import express from 'express';
+import * as express from 'express';
```
konstantintieber, line 6: Just noticed that this is the only place where we import express like this. What's the reasoning?

ireneea: It's used for the types `express.Response` would make more sense to import the types directly. From memory I changed this because there was a linting error. 

Either way doesn't really make a big difference here

konstantintieber (approved):

PR #24393, opened 2026-01-15
test(editor): Fix flaky chat session ID reset E2E test

mutdmour:
## Summary

Fixes the flaky E2E test `Chat session ID reset > should update session ID in node output when session is reset` which had an 8-25% failure rate.

**Root Cause:** The test used a tooltip locator `.n8n-tooltip` with `hasText: 'click to copy'` which matched multiple elements during tooltip transitions when both old and new tooltips were visible simultaneously.

**Fix:** Use clipboard to extract session ID instead of parsing tooltip text. The session ID button already copies the full session ID to clipboard when clicked, so we can read it directly - avoiding any tooltip matching issues.

**Verification:** Ran 20 repetitions of the test with 0 failures (previously 8-25% flaky rate).

https://linear.app/n8n/issue/AI-1935

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/testing/playwright/pages/components/LogsPanel.ts
+++ b/packages/testing/playwright/pages/components/LogsPanel.ts
@@ -1,7 +1,7 @@
 import type { Locator } from '@playwright/test';
 
-import type { ClipboardHelper } from '../../helpers/ClipboardHelper';
 import { RunDataPanel } from './RunDataPanel';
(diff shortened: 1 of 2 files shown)


shortstacked (dismissed):
Thanks!

shortstacked (approved):

PR #24394, opened 2026-01-15
fix(core): Check license when updating scopes (no-changelog)

RicardoE105:
## Summary

Fixes missing license check where users without the `feat:apiKeyScopes` license could bypass frontend restrictions and modify API key scopes via direct HTTP requests.

  ## Problem

  Users on plans without API Key Scopes feature could send crafted requests to `POST /rest/api-keys` and `PATCH /rest/api-keys/:id` to customize scopes, bypassing subscription enforcement. Frontend blocked this, but backend did not validate the license.

https://linear.app/n8n/issue/ADO-4589/bypass-free-plan-limitation-allows-modification-of-api-key-scopes-via

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
```diff
@@ -64,9 +64,20 @@ export class ActivateExecuteWorkflowTriggerWorkflows1763048000000 implements Irr
 
 		const inactiveWorkflows = `SELECT ${idColumn}, ${nodesColumn}, ${connectionsColumn}, ${versionIdColumn}, ${activeVersionIdColumn} FROM ${tableName} WHERE ${activeColumn} = false AND (${nodesColumnForLike} LIKE '%n8n-nodes-base.executeWorkflowTrigger%' OR ${nodesColumnForLike} LIKE '%n8n-nodes-base.errorTrigger%')`;
 
+		const migrationName = 'ActivateExecuteWorkflowTriggerWorkflows';
+
 		await runInBatches<Workflow>(inactiveWorkflows, async (workflows) => {
 			for (const workflow of workflows) {
-				const nodes = parseJson(workflow.nodes);
+				let nodes: Node[];
```
CharlieKolb: This change seems unrelated

RicardoE105: good catch

packages/cli/src/controllers/api-keys.controller.ts
```diff
@@ -108,4 +121,15 @@ export class ApiKeysController {
...
 		return scopes;
 	}
+
+	/**
+	 * Resolves the scopes to be used for an API key based on license status.
+	 * If the API Key Scopes feature is not licensed, returns all available scopes for the user's role.
+	 * Otherwise, returns the requested scopes.
+	 */
+	private resolveScopesForUser(user: AuthenticatedRequest['user'], requestedScopes: ApiKeyScope[]) {
+		return this.license.isLicensed(LICENSE_FEATURES.API_KEY_SCOPES)
```
Cadiac, line 131: I was expecting we'd use the `@Licensed('feat:apiKeyScopes)` decorator we have on API routes here on this controller?

Cadiac: nvm I guess it isn't convenient here if these endpoints can still be called anyways even without that.. I thought they weren't actually supposed to be callable at all

RicardoE105: if we use the decorator the requests will fail when the instance does not have it the api keys scopes feature and it should not. It should still work but do not allow to modify the scopes which is the part the license blocks

Cadiac: makes sense, when I looked into this briefly before I thought these endpoints weren't supposed to be callable at all without that license permission, but this solution does handle that

packages/cli/src/controllers/api-keys.controller.ts
Cadiac, line 97: but if they can't modify the scopes why can they call this endpoint at all without the license?

RicardoE105: without the license you can still update the API key label, and the expiration for example

RicardoE105: the license only applies to the scopes

Cadiac (approved):

PR #24396, opened 2026-01-15
fix(core): Fix concurrency control not releasing capacity if execution continuation fails

tomi:
## Summary

If the setting execution running or resuming it fails, we are not releasing capacity from concurrency control, which can leak capacity and grind the instance to halt.

https://linear.app/n8n/issue/CAT-2064

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/active-executions.ts
```diff
@@ -52,61 +53,74 @@ export class ActiveExecutions {
...
+					throw new ExecutionAlreadyResumingError(maybeExecutionId);
+				}
 			}
+		} catch (error) {
+			capacityReservation.release();
+			throw error;
 		}
 
+		// At this point executionId is guaranteed to be defined - capture it for use in closures
+		const executionId = maybeExecutionId;
```
tomi, line 120: In theory the code here might throw as well, but opted for code clarity and left it outside the try..catch. It's all error free

packages/cli/src/concurrency/__tests__/concurrency-capacity-reservation.test.ts
ivov, line 1: These tests are great! But we aren't testing for the actual bug, i.e. ensuring capacity is released when `updateExistingExecution` returns `false`. Or skipped because `ActiveExecutions` is not easily testable?

packages/cli/src/active-executions.ts
```diff
@@ -52,61 +53,74 @@ export class ActiveExecutions {
...
-			fullExecutionData.retryOf = executionData.retryOf ?? undefined;
-
-			const workflowId = executionData.workflowData.id;
-			if (workflowId !== undefined && isWorkflowIdValid(workflowId)) {
-				fullExecutionData.workflowId = workflowId;
-			}
+		const capacityReservation = new ConcurrencyCapacityReservation(this.concurrencyControl);
+
+		try {
+			if (maybeExecutionId === undefined) {
```
ivov, line 65: Since we're touching this, we could check if status is `new` to show intent more clearly.

packages/cli/src/concurrency/concurrency-capacity-reservation.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	async reserve(capacityFor: CapacityTarget) {
+		await this.concurrencyControlService.throttle(capacityFor);
+		this.acquiredReservation = capacityFor;
+	}
+
+	release() {
+		if (!this.acquiredReservation) return;
+
+		this.concurrencyControlService.remove(this.acquiredReservation);
```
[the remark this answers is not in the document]
ivov: Agree with the bot!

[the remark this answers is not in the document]
tomi: Ahhh good catch! I actually misunderstood how the queue behaves. My intention was to make sure we only release capacity if we have it reserved for that execution, but we don't seem to track that. The reservation handles that now so we don't need it on the queue side

packages/cli/src/active-executions.ts
tomi: That causes a TS error in the else branch, since we are narrowing the type here. This code could definitely use some clean up in general but didn't want to start refactoring it too much

packages/cli/src/concurrency/__tests__/concurrency-capacity-reservation.test.ts
tomi: Good point. I added tests for this now 👍

ivov (approved):

PR #24406, opened 2026-01-15
feat(ai-builder): Ask mode callout explaining its usage

mike12345567:
## Summary

Adds onboarding improvements for Ask mode to help users understand its purpose:

- Coachmark tooltip: Appears when users enter Ask mode for the first time, explaining what the mode does and how to switch to Build mode
- Empty state: New placeholder UI with guidance when no messages exist in Ask mode
- Updated input placeholder: Changed to "What do you need help with?"

The coachmark only shows once (persisted via user settings) and won't appear when opening via error/credential helpers since those provide their own context.

Addresses ticket: https://linear.app/n8n/issue/AI-1677

## Demo

ask-coachmark

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/assistant/composables/useAskModeCoachmark.ts
```diff
@@ -0,0 +1,58 @@
...
+import { useCalloutHelpers } from '@/app/composables/useCalloutHelpers';
+import { BUILDER_ENABLED_VIEWS } from '../constants';
+import type { VIEWS } from '@/app/constants';
+
+export const ASK_MODE_COACHMARK_KEY = 'ask-mode-coachmark';
+
+const isBuilderEnabledView = (name: unknown): name is VIEWS =>
+	BUILDER_ENABLED_VIEWS.includes(name as VIEWS);
+
+export function useAskModeCoachmark() {
```
burivuhster, line 15: Is the term "coachmark" the same as "callout", or they describe different concepts? If they are the same – would it make sense to name this composable `useAskModeCallout`?

burivuhster (dismissed):
Looks good! Nitpick naming comment, nothing blocking the merge

packages/frontend/editor-ui/src/features/ai/assistant/composables/useAskModeCoachmark.ts
mike12345567: I think the separation of naming here is somewhat important, a coachmark is a one off, once the user dismisses it they shouldn't see it again. Callouts are something that are more static/appear every time, although perhaps we use the term callout for these dismissable components already?

burivuhster: I see, that makes sense, thanks for explanation! We didn't use the term coachmark before, not sure about the existing one off behavior for callouts

burivuhster (dismissed):

burivuhster (approved):

PR #24412, opened 2026-01-15
feat: Add migration to split use and read scopes

BGZStephen:
## Summary

This PR provides a migration to add an additional `credential:use` scope for any roles that currently have the `credential:read` scope.

There are also code changes to support using the scope for custom roles:

image

Which has the following behaviour:

* If the user selects `use`, they are auto granted `use`
* If the user deselects `view` and they have `use` it removes `use`
* The user can have `view` without having `use`

Code has been updated so that when a workflow is executed, where we previously checked for `credential:read` to determine if a credential can be used, we now instead check for `credential:use`

IAM-77

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/migrations/common/1768475341537-AddCredentialUseScopeToRoles.ts
```diff
@@ -0,0 +1,114 @@
...
+		const descriptionColumn = escape.columnName('description');
+
+		const roleTableName = escape.tableName('role');
+		const roleScopeTableName = escape.tableName('role_scope');
+		const roleSlugColumn = escape.columnName('slug');
+		const roleScopeRoleSlugColumn = escape.columnName('roleSlug');
+		const roleScopeScopeSlugColumn = escape.columnName('scopeSlug');
+
+		// Step 1: Ensure credential:use scope exists
+		const isPostgresOrSqlite = dbType === 'postgresdb' || dbType === 'sqlite';
```
guillaumejacquart, line 32: You only add the migration to postgres and sqlite, so it this check really necessary ?

guillaumejacquart (commented):
1 comment 
Also, if we merge this now, will we need another migration once the new scope is implemented, to make sure to run the 		// Step 2: Add credential:use to all roles that have credential:read (all role types) again for all new roles ?

packages/@n8n/db/src/migrations/common/1768475341537-AddCredentialUseScopeToRoles.ts
BGZStephen: The automated code reviewer complains if we don't add support for mysql :D so It felt fine to add it

PR #24435, opened 2026-01-16
feat(editor): Add a clarifying callout to Merge Node schema view

dariacodes:
## Summary

- Add a possibility to add dismissive callouts to the input and output panes (reuses the existing `useCalloutHelpers`)
- Add a callout to the input panel of nodes receiving data from Merge Nodes that appears if there is more than 1 item [changes in `useDataSchema.ts`] 
image

- Add a callout to the output panel of the Merge Node that appears if there is more than 1 input item [changes in `VirtualSchema.vue`]
image

Closes [ADO-4633](https://linear.app/n8n/issue/ADO-4633/add-a-clarifying-note-in-the-merge-node-output-in-schema-view)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useDataSchema.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useDataSchema.ts
@@ -318,6 +318,14 @@ export type RenderNotice = {
 	message: string;
 };
 
+export type RenderCallout = {
+	id: string;
+	type: 'callout';
+	level: number;
+	message: string;
+	theme?: 'info' | 'success' | 'warning' | 'danger' | 'secondary';
+};
+
 export type RenderEmpty = {
 	id: string;
 	type: 'empty';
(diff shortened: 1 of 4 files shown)


packages/frontend/editor-ui/src/app/composables/useDataSchema.ts
```diff
@@ -507,6 +521,17 @@ export const useFlattenSchema = () => {
 				return acc;
 			}
 
+			if (item.node.type === 'n8n-nodes-base.merge' && item.itemsCount > 1) {
```
CharlieKolb, line 524: Doesn't the callout apply to all nodes conceptually? 🤔

Don't mind keeping it for merge where it's most common, just so I understand correctly

dariacodes: I think the point was that with merge specifically, you can combine inputs and this is why some keys could be absent. With other nodes, just conceptually, it's quite clear that all the items should have these keys, no?

CharlieKolb (approved):

PR #24440, opened 2026-01-16
refactor(core): Add `storedAt` column to `execution_entity` table

ivov:
https://linear.app/n8n/issue/CAT-2090


guillaumejacquart (approved):
LGTM
nit questions: 
- should we maybe allow for more that 2 chars to be future proof ?
- is s3 maybe too format specific, like 'objectStorage' would be more generic ? or is it what we really mean (only aws s3 or compatible engine) ?

PR #24446, opened 2026-01-16
feat: Optimize community nodes types fetching with local data and incremental updates

michael-radency:
## Summary
This PR adds a persistence layer for community node types with pre-bundled data files and incremental updates to improve reliability and performance.

### Key Changes

**1. Pre-bundled Community Node Types Data**
- Added `staging-node-types.json` and `production-node-types.json` files containing pre-fetched community node types
- Files are bundled with the build and copied to `dist/` during compilation
- Provides immediate data availability on instance startup without requiring API calls

**2. File-Based Persistence Layer**
- Service now hydrates from local JSON files on initialization
- Automatically persists updates to disk when new data is fetched from API
- Uses atomic write operations (write to temp file, then rename) to prevent corruption
- Acts as a fallback when API is unreachable

**3. Incremental Update Strategy**
- Fetches only metadata (name, version, updatedAt) to detect changes
- Requests full data only for nodes that have been updated
- Significantly reduces API load and bandwidth usage

**4. CLI Command for Data Updates**
- New `update-node-types-data` command to refresh bundled JSON files
- Fetches latest data from both staging and production APIs
- Allows maintainers to update pre-bundled data before releases

**5. Optimized Package Re-installation**
- Changed from fetching all community node types to fetching only the specific package needed
- Uses Strapi filters to request single package data during re-install
- Reduces unnecessary API calls and improves performance

**6. Build Process Integration**
- Updated `build.mjs` to copy community packages data files to dist
- Ensures JSON files are available in production builds

### Benefits

- **Faster cold starts**: Local cache provides immediate data availability, eliminating initial large fetch
- **Improved reliability**: Local files act as fallback when API is unavailable or slow
- **Reduced API load**: Incremental updates retrieve only changed nodes instead of re-fetching all types
- **Better performance**: Nodes available immediately on instance start with lightweight subsequent updates
- **Optimized re-installs**: Fetches only required package data instead of entire dataset

### Trade-offs

- **Potentially stale data on first load**: Local JSON data needs periodic updates (added CLI command)
- **Larger bundle size**: JSON data files are included in build output

https://linear.app/n8n/issue/NODE-4169/community-nodes-failing-to-load-on-cloud

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/community-packages/update-node-types.command.ts
```diff
@@ -0,0 +1,58 @@
...
+			const filePath = join(
+				process.cwd(),
+				'packages',
+				'cli',
+				'src',
+				'modules',
+				'community-packages',
+				'data',
+				`${environment}-node-types.json`,
+			);
```
tomi, line 47: This path will not be correct in our docker image where the cwd is `/home/node` and the packages are in `/usr`. We should use `__dirname` as the starting point instead

packages/cli/src/modules/community-packages/update-node-types.command.ts
```diff
@@ -0,0 +1,58 @@
...
+			const jsonContent = JSON.stringify(nodeTypes, null, 2);
```
tomi, line 48: Why do we want this pretty printed? It should be only programmatically consumable, and one can format it manually if needed

packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -22,13 +30,70 @@ export class CommunityNodeTypesService {
...
+			}
+		}
+
+		return typesToUpdate;
+	}
+
+	private hydrateNodeTypes(environment: 'staging' | 'production' = 'production') {
+		const filePath = join(__dirname, 'data', `${environment}-node-types.json`);
+		try {
+			const fileContent = readFileSync(filePath, 'utf-8');
```
tomi, line 67: Don't use sync methods as it block the entire event loop

packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -22,13 +30,70 @@ export class CommunityNodeTypesService {
 		private communityPackagesService: CommunityPackagesService,
 	) {}
 
+	async init() {
+		if (this.communityNodeTypes.size > 0) return;
+		this.logger.debug('Initializing community node types...');
+
+		const environment = this.detectEnvironment();
+		this.hydrateNodeTypes(environment);
```
tomi, line 38: Keeping all the community nodes in-memory can be quite a bit of memory overhead. It would be better to do the hydration lazily. I.e. when the node types are requested for the first time.

packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -22,13 +30,70 @@ export class CommunityNodeTypesService {
...
+	}
+
+	private hydrateNodeTypes(environment: 'staging' | 'production' = 'production') {
+		const filePath = join(__dirname, 'data', `${environment}-node-types.json`);
+		try {
+			const fileContent = readFileSync(filePath, 'utf-8');
+			const nodeTypes: StrapiCommunityNodeType[] = JSON.parse(fileContent);
+			this.setCommunityNodeTypes(nodeTypes);
+		} catch (error) {
+			this.logger.error(
```
tomi, line 71: We should errorReporter to get these into Sentry. Only logging leaves them unnoticed

packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -22,13 +30,70 @@ export class CommunityNodeTypesService {
...
+
+		for (const entry of communityNodesMetadata) {
+			const nodeType = this.communityNodeTypes.get(entry.name);
+
+			if (
+				!nodeType ||
+				nodeType.npmVersion !== entry.npmVersion ||
+				nodeType.updatedAt !== entry.updatedAt
+			) {
+				this.logger.info(
```
tomi, line 54: Let's use debug for this. We don't need to be too verbose

packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -22,13 +30,70 @@ export class CommunityNodeTypesService {
...
 		try {
+			// Cloud sets ENVIRONMENT to 'production' or 'staging' depending on the environment
+			const environment = this.detectEnvironment();
+
+			if (this.communityNodeTypes.size === 0) {
+				this.hydrateNodeTypes(environment);
+			}
+
 			let data: StrapiCommunityNodeType[] = [];
 			if (this.config.enabled && this.config.verifiedEnabled) {
```
tomi, line 87: Shouldn't we the `enabled` in the bginnig of this method? Do we need this check as shouldn't this only be loaded if it's enabled?

packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -54,15 +119,35 @@ export class CommunityNodeTypesService {
...
+				});
+			});
+		}
+	}
+
+	private async persistNodeTypesToFile(nodeTypes: StrapiCommunityNodeType[]) {
+		if (nodeTypes.length === 0) return;
+
+		const environment = this.detectEnvironment();
+		const filePath = join(__dirname, 'data', `${environment}-node-types.json`);
```
tomi, line 145: We should not write to paths that are not the working dir, temp dir or other explicitly mounted path. In k8s/Docker, these writes end up in the node's root file system, and can cause disk pressure and performance issues across the cluster. They are also not persisted across restarts.

So if we really want to persist these it should be e.g. in the n8n home folder, and only in regular mode as we can't assume persisted disk on scaling mode deployments. A better place for these in scaling mode would be the cache (redis)

packages/cli/src/modules/community-packages/community-node-types-utils.ts
```diff
@@ -21,16 +21,48 @@ export type StrapiCommunityNodeType = {
...
 
+function getUrl(environment: 'staging' | 'production'): string {
+	return environment === 'production'
+		? N8N_VETTED_NODE_TYPES_PRODUCTION_URL
+		: N8N_VETTED_NODE_TYPES_STAGING_URL;
+}
+
 export async function getCommunityNodeTypes(
 	environment: 'staging' | 'production',
+	qs: { filters?: IDataObject; fields?: string[] } = {},
```
tomi, line 42: Let's please stop using `IDataObject`. It's essentially the same `any` and we don't want to use it

packages/cli/src/modules/community-packages/community-node-types-utils.ts
```diff
@@ -21,16 +21,48 @@ export type StrapiCommunityNodeType = {
...
-	return await paginatedRequest<StrapiCommunityNodeType>(url);
+export async function getCommunityNodesMetadata(
+	environment: 'staging' | 'production',
+): Promise<CommunityNodesMetadata[]> {
+	const url = getUrl(environment);
+	const params = {
+		fields: ['npmVersion', 'name', 'updatedAt'],
+		pagination: {
+			page: 1,
+			pageSize: 500,
```
tomi, line 64: Are we fetching only the first 500 or all of them? It's not very clear just by reading the code

packages/cli/src/modules/community-packages/community-packages.service.ts
```diff
@@ -327,19 +327,27 @@ export class CommunityPackagesService {
...
-				const versionMap = new Map<string, string>();
-				versionMap.set(p.npmVersion, p.checksum);
-				for (const v of p.nodeVersions ?? []) versionMap.set(v.npmVersion, v.checksum);
-				checksums.set(p.packageName, versionMap);
-			}
 
 			for (const missingPackage of missingPackages) {
 				try {
-					const checksum = checksums.get(missingPackage.packageName)?.get(missingPackage.version);
+					const [vettedPackage] = await getCommunityNodeTypes(environment, {
```
tomi, line 333: It would be much better resource consumption wise to fetch these packages in a single fetch instead of firing a separate request for each one

packages/cli/src/modules/community-packages/data/production-node-types.json
tomi, line 1: Can we:
1. Minify instead of prettify them. Would cut down the size by 1MB
2. Gzip it

tomi, line 1: When are these updated? We should probably do the same as with node popularity, that gets updated once a week. See `util-update-node-popularity.yml`. That can be a followup

packages/cli/src/modules/community-packages/update-node-types.command.ts
tomi, line 1: Is this really needed? I don't imagine users actually running this and in docker deployments (what is our only recommended way of running n8n) it's pretty useless as it's not persisted across restarts

tomi (commented):
Few comments

packages/cli/src/modules/community-packages/update-node-types.command.ts
michael-radency: maybe not, but I wonder if those files, since they are included in the source code, should be readable

packages/cli/src/modules/community-packages/community-node-types.service.ts
michael-radency: then we may not need  init method, all types would be requested when user opens workflow

packages/cli/src/modules/community-packages/community-node-types-utils.ts
michael-radency: it's paginated, 500 per page

packages/cli/src/modules/community-packages/community-packages.service.ts
michael-radency: unfortunately so far did not manage to construct such filter with multiple entries, looks like Strapi limitation

packages/cli/src/modules/community-packages/data/production-node-types.json
michael-radency: yes, agree, unless we want them to be readable, since they are included in source

michael-radency: I would imagine we would do update PR  after large community nodes release batch

michael-radency: would be nice to automate it though, maybe integrate into release process

packages/cli/src/modules/community-packages/update-node-types.command.ts
michael-radency: it's intended to be used by developers, to update json data files, for this purpose I think it's useful, I'll update descriptions

PR #24448, opened 2026-01-16
fix: Fix issue preventing some community nodes from installing

Joffcom:
## Summary
When checking Community Nodes we were using `includes` which sometimes resulted in nodes returning the incorrect checksum and failing to verify for installation. This PR changes it so we use the packageName instead of the Keyname and allows us to drop the includes for more accurate matching.

https://linear.app/n8n/issue/NODE-4220/community-nodes-package-matching-sometimes-fails

## Review / Merge checklist

- [x] Tests included. 


ShireenMissi (approved):
I tested locally and although I can't reproduce the reported issue I confirm that original functionality works
LGTM ✅

PR #24452, opened 2026-01-16
refactor(core): Initial setup for execution data management

ivov:
https://linear.app/n8n/issue/CAT-2092 (only scaffolding, unused yet)

### DIFF
--- a/packages/cli/src/executions/failed-run-factory.ts
+++ b/packages/cli/src/executions/failed-run-factory.ts
@@ -8,7 +8,7 @@ import {
 } from 'n8n-workflow';
 
 @Service()
-export class ExecutionDataService {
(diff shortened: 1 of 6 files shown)


packages/cli/src/executions/failed-run-factory.ts
ivov, line 1: This is an `IRun`, not exactly execution data, so I'm freeing up the name to use soon.

mfsiega (approved):
For my understanding, this PR does 2 things:
- Renames the existing `ExecutionDataService` to `FailedRunFactory` because that's a better name for what it actually does.
- Introduces an `ExecutionDataManager` type, which will be used in the future.

If so, LGTM - I would just edit the PR description to reflect this.

PR #24455, opened 2026-01-16
fix(core): Add session token authentication for chat WebSocket connections

DawidMyslak:
## Summary

This PR adds session token authentication to the `/chat` WebSocket endpoint used by the Chat Trigger node.

### Background

The Chat Trigger node uses two protocols:
1. **HTTP** - Initial request to start a chat workflow execution
2. **WebSocket** - Real-time messaging during the chat session

When HTTP authentication (Basic Auth or n8n User Auth) is configured, it only applied to the HTTP request. The WebSocket endpoint was set up independently and did not have its own authentication mechanism - it only required an `executionId`, which is not a secret value.

### Solution

We now generate a cryptographically secure session token when a chat execution starts and require it for WebSocket connections:

1. **Token Generation**: When a chat workflow execution begins (via HTTP), we generate a signed JWT token containing the `executionId` and return it in the response
2. **Token Validation**: The WebSocket connection must include this token as a query parameter. The server validates the JWT signature, checks the algorithm (HS256 only), verifies the `executionId` matches, and confirms the token hasn't expired
3. **Token Lifecycle**: Tokens are stateless (no server-side storage) with a 5-minute expiry. This works across multi-main setups since all instances share the same signing secret derived from the encryption key

This ensures that only the client who initiated the chat execution (and received the token) can establish the WebSocket connection.

https://linear.app/n8n/issue/NODE-4206

## Review / Merge checklist

- [x] Tests included


michael-radency (approved):
LGTM

PR #24456, opened 2026-01-16
feat(editor): Pause autosave on connection failures

dariacodes:
## Summary

Pause autosave when experiencing connection problems. Resume once the tab goes back online.
Reuse the current `PushConnectionTracker` to show network error message (Git unfortunately doesn't show that the file was just renamed, and there are only a few changes)

Closes [ADO-4626](https://linear.app/n8n/issue/ADO-4626/feature-delay-or-pause-autosave-on-request-failures)

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/composables/useNetworkStatus.ts
```diff
@@ -0,0 +1,28 @@
...
+		window.addEventListener('offline', handleOffline);
+	});
+
+	onUnmounted(() => {
+		window.removeEventListener('online', handleOnline);
+		window.removeEventListener('offline', handleOffline);
+	});
+
+	return {
+		isOnline: networkStore.isOnline,
```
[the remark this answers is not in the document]
dariacodes: `networkStore.isOnline` is already a ref from the Pinia store, so it should remain reactive when returned

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -2164,6 +2164,8 @@
 	"pushConnectionTracker.cannotConnectToServer": "You have a connection issue or the server is down. <br />n8n should reconnect automatically once the issue is resolved.",
 	"pushConnectionTracker.connectionLost": "Connection lost",
 	"pushConnectionTracker.connectionLost.message": "Attempting to reconnect...",
+	"autosave.connectionLost": "Connection lost, auto save paused",
```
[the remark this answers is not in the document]
dariacodes: Still waiting for the decision on what text to put here

packages/frontend/editor-ui/src/app/stores/network.store.ts
```diff
@@ -0,0 +1,18 @@
+import { defineStore } from 'pinia';
+import { ref } from 'vue';
+
+/**
+ * Store for tracking app-wide network connectivity status.
+ */
+export const useNetworkStore = defineStore('network', () => {
+	const isOnline = ref(typeof navigator !== 'undefined' ? navigator.onLine : true);
```
MiloradFilipovic: What's the purpose of having two online checks (`navigator.onLine` here and backend reachability test in the `useNetworkStatus` composable)? I was wondering if we can remove this check if we cannot fully rely on it?

packages/frontend/editor-ui/src/app/stores/network.store.ts
```diff
@@ -0,0 +1,18 @@
...
+
+	function setOnline(value: boolean) {
+		isOnline.value = value;
+	}
+
+	return {
+		isOnline,
```
MiloradFilipovic: Nit: With both exported like this, you can update `isOnline` directly, making the setter method redundant. If you would like to keep it and make it the only way of updating the ref, you can export the `isOnline` as readonly:
```
return {
	isOnline: readonly(isOnline),
	setOnline,
};
```

MiloradFilipovic (dismissed):
Works as expected 🚀 I left two comments for the code but looks good from my side.

packages/frontend/editor-ui/src/app/stores/network.store.ts
dariacodes: Just to react to offline changes quicker when possible, because we're sending health checks not thaat often (and I didn't want to have too frequent requests). I researched that usually the combination of the two is a good solution

dariacodes: upd.: When developing on localhost, it's important to not pause autosave, though browser is sending an `offline` event, so we decided to switch to only pinging backend (just more often) and not listening to browser events, as, otherwise, we would need to send a request after each offline event anyway to check whether there is truly no network.

packages/frontend/editor-ui/src/app/components/ConnectionTracker.vue
```diff
@@ -0,0 +1,85 @@
...
+	  }
+	| {
+			hasError: false;
+			message: null;
+			tooltip: null;
+	  };
+
+const connectionStatus = computed<ConnectionStatus>(() => {
+	// Priority 1: Check if browser is offline
+	if (!networkStore.isOnline) {
```
MiloradFilipovic: Is this still needed now that we are only relying on backend connectivity check?

packages/frontend/editor-ui/src/app/stores/network.store.ts
```diff
@@ -0,0 +1,18 @@
+import { defineStore } from 'pinia';
+import { readonly, ref } from 'vue';
+
+/**
+ * Store for tracking app-wide network connectivity status.
+ */
+export const useNetworkStore = defineStore('network', () => {
```
MiloradFilipovic: Can we remove this store now that we are relying only on backend check?

MiloradFilipovic (commented):
Thanks for addressing the feedback 🚀
Left some comments about the cleanup so let me know wdyt 🙏

packages/frontend/editor-ui/src/app/components/ConnectionTracker.vue
dariacodes: It's still needed as I'm setting the online status in the store so that it's available everywhere in the app. But let me fix the comment and the store name to improve clarity

MiloradFilipovic (approved):
🚢

PR #24467, opened 2026-01-16
feat(editor): Add new Button component to design system (no-changleog)

heymynameisrob:
## Summary

Adds a new `N8nButton` component to the v2 design system with a modern, streamlined API. This component is designed to replace the legacy button implementation with:

- **5 variants**: `solid`, `subtle`, `outline`, `ghost`, `destructive`
- **3 sizes**: `xsmall`, `small`, `medium`
- **States**: loading (with spinner), disabled, icon-only
- **Link support**: Renders as `<a>` when `href` is provided
- **Accessibility**: Dev-time warnings for icon-only buttons missing accessible labels
- **Flexible icons**: Icons passed via slots for leading/trailing positioning

### Key changes:
- New `Button.vue` component with TypeScript props interface
- Comprehensive Storybook stories for all variants and states
- Component specification document with API docs and migration guide
- New `cn` utility for className merging
- Focus ring mixin for consistent focus states
- Updated design tokens and primitives
- Unit tests for all props, variants, accessibility, and edge cases

https://linear.app/n8n/issue/DS-302

NB Replacing old component will happen in a separate PR

## Review / Merge checklist

- [x] Tests included.


packages/frontend/@n8n/design-system/src/css/_primitives.scss
heymynameisrob, line 1: FYI: // comments were breaking my LSP so moved to a more traditional CSS commenting style.

CharlieKolb (commented):
Reassigning to Milorad as good design system principles are a bit outside my expertise and this seems quite core :D

pnpm-workspace.yaml
```diff
@@ -80,6 +80,12 @@ catalog:
   zod-to-json-schema: 3.23.3
 
 catalogs:
+  e2e:
+    '@currents/playwright': ^1.15.3
```
konstantintieber: It seems like several files in this PR were changed just for formatting and reordering things.
See also the whitespace changes in the package.json.
Do you know how these came to be @heymynameisrob?

packages/frontend/@n8n/design-system/src/v2/components/Button/component-button.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Migration from Legacy N8nButton
+
+| Legacy             | New                                              |
+| ------------------ | ------------------------------------------------ |
+| `type="primary"`   | `variant="solid"`                                |
+| `type="secondary"` | `variant="subtle"`                               |
+| `type="tertiary"`  | `variant="ghost"`                                |
+| `type="danger"`    | `variant="destructive"`                          |
+| `type="success"`   | `variant="solid"` + custom class (see overrides) |
```
konstantintieber, line 182: What's the motivation behind making `type=success` not a supported variant, but requiring a user of the component to set `variant=solid` plus a CSS class?

pnpm-workspace.yaml
heymynameisrob: I am not :/ I assume it's some prettier formatting gone wrong. Do we not have a CI job that'll put these back? If not I can recheck it.

konstantintieber: If it's merged in this PR, that's what'll be tracked in the repo from then on, until the next engineer edits it with maybe different formatting settings.

packages/frontend/@n8n/design-system/src/v2/components/Button/component-button.md
heymynameisrob: Fair question.

Mostly backwards compatibility. I'm not 100% sure we want to keep a Success variant button but kept this here in case the need arises. Seems to me to be a one-off case that's only present in a few instances (found 3).

Arguably we could just keep a success variant but as a designer I'd be worried if by having it we encourage people to use it alot (when we'd rather not 😂  ) 

Don't feel strongly either way tbh

pnpm-workspace.yaml
heymynameisrob: Same has happened with `package.json` as the only actual change I did was add the `clsx` package

packages/frontend/@n8n/design-system/src/v2/components/Button/component-button.md
konstantintieber: True, searching for `type="success"` in our frontend shows only four usages of it today. That's nothing.

packages/frontend/@n8n/design-system/src/v2/components/Button/Button.test.ts
```diff
@@ -0,0 +1,402 @@
...
+					slots: { default: 'Button' },
+					global: { stubs },
+				});
+				const button = wrapper.container.querySelector('button');
+				expect(button?.className).toContain('small');
+			});
+		});
+
+		describe('loading', () => {
+			it('should show loading spinner when loading', () => {
```
konstantintieber, line 129: Would be nice to have a test case "should be disabled while loading" here.
Since that prevents submitting a form with a pending request twice.

packages/frontend/@n8n/design-system/src/v2/components/Button/Button.test.ts
```diff
@@ -0,0 +1,402 @@
...
+					props: { loading: true },
+					slots: { default: 'Button' },
+					global: { stubs },
+				});
+				const button = wrapper.container.querySelector('button');
+				expect(button?.className).toContain('loading');
+				expect(wrapper.container.querySelector('n8n-icon-stub')).toBeInTheDocument();
+			});
+
+			it('should have pointer-events none when loading', () => {
```
konstantintieber: The test name and the assertion here do not match up imo.
If the class definition of `.loading` no longer changes the pointer-events property, this test will still be green.
Might be worth renaming the test to "adds loading CSS class to wrapper element" to prevent false security from the current test name.

packages/frontend/@n8n/design-system/src/v2/components/Button/Button.vue
```diff
@@ -0,0 +1,365 @@
...
+});
+
+const $style = useCssModule();
+const attrs = useAttrs() as ButtonHTMLAttributes;
+
+onMounted(() => {
+	if (import.meta.env.DEV && props.icon) {
+		const hasAccessibleLabel = attrs['aria-label'] || attrs['aria-labelledby'] || attrs.title;
+		if (!hasAccessibleLabel) {
+			console.warn(
```
konstantintieber, line 46: Love this development feature!

packages/frontend/@n8n/design-system/src/v2/components/Button/Button.test.ts
heymynameisrob: good catch

pnpm-lock.yaml
```diff
@@ -30731,7 +30734,7 @@ snapshots:
       isstream: 0.1.2
       jsonwebtoken: 9.0.3
       mime-types: 2.1.35
-      retry-axios: 2.6.0(axios@1.12.0)
+      retry-axios: 2.6.0(axios@1.12.0(debug@4.4.3))
```
konstantintieber, line 30737: Hmm, this debug part is currently not on master: https://github.com/n8n-io/n8n/blob/master/pnpm-lock.yaml#L30734

I don't expect that installing `clsx` would require this change.

konstantintieber: Aha your pnpm-lock inverted the definition. On master the debug version is set further up above: https://github.com/n8n-io/n8n/blob/master/pnpm-lock.yaml#L30724

konstantintieber: Which pnpm version are you using? I am on 10.22.0

heymynameisrob: Same 10.22.0

heymynameisrob: I'll recheck this out too.

konstantintieber (approved):

pnpm-lock.yaml
konstantintieber: When I checkout your branch and run `pnpm install` I get the same lock file changes.
On master of the n8n repo I don't.

heymynameisrob: Seems to always want to dedupe the axios-retry :shrug:

PR #24470, opened 2026-01-16
fix(editor): Handle large values breaking data tables UI grid

scdekov:
## Summary

Inserting huge values in data table cells (mostly via the node) was causing the UI grid to hang and crash the browser page

image

https://linear.app/n8n/issue/ADO-4256/community-issue-data-table-ui-breaks-if-a-cell-contains-too-much-data

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/core/dataTable/components/dataGrid/OversizedCellRenderer.vue
+++ b/packages/frontend/editor-ui/src/features/core/dataTable/components/dataGrid/OversizedCellRenderer.vue
@@ -1,11 +1,17 @@
 <script setup lang="ts">
 import { useI18n } from '@n8n/i18n';
+import { N8nTooltip } from '@n8n/design-system';
 
 const i18n = useI18n();
 </script>
 
 <template>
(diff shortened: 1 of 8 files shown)


packages/frontend/editor-ui/src/features/core/dataTable/components/dataGrid/OversizedCellRenderer.vue
```diff
@@ -0,0 +1,16 @@
+<script setup lang="ts">
+import { useI18n } from '@n8n/i18n';
+
+const i18n = useI18n();
+</script>
+
+<template>
+	<span class="n8n-oversized-value">{{ i18n.baseText('dataTable.cell.oversized') }}</span>
```
CharlieKolb: If we can easily add a tooltip here lets add one that points out that the user can modify this value via a data table node? Since we make the cell itself non-editable

CharlieKolb (dismissed):
One small note, happy to merge as is tho!

CharlieKolb (approved):

PR #24494, opened 2026-01-17
chore(core): Extract ldap into module, and introduce auth handler for registering it

guillaumejacquart:
## Summary

Extract the LDAP feature into it's own module. 
We still have some direct call from the core to the module (auth.controller.ts), we will tackle that in a followup PR

https://linear.app/n8n/issue/IAM-100/ldap-module-basic-module-extraction

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/auth/auth-handler.registry.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * Registry service that manages authentication handlers.
+ * Modules can register their authentication handlers here,
+ * and the core can query if any handler can handle a specific authentication method.
+ */
+@Service()
+export class AuthHandlerRegistry {
+	private handlers: AuthenticationHandler[] = [];
```
ireneea: Can we have multiple `AuthenticationHandler` with the same `authenticationMethod`.

Asking this because I'm wondering if this should be a `Map` rather than an array.

packages/cli/src/sso.ee/sso-helpers.ts
ireneea: This is a bit confusing 😅, how do we know which should be used? 

- `isSSOCurrentAuthenticationMethod`
- `isSsoCurrentAuthenticationMethod`

guillaumejacquart: oh woa, thanks, good catch ! I think I'll reduce the scope of this PR to extraction only, and see about refactoring afterwards

packages/cli/src/auth/auth-handler.registry.ts
guillaumejacquart: Good point. I removed it for now, as it needs to be part of a bigger conversion for all the handlers. Will keep your remark in mind

ireneea (dismissed):

ireneea (approved):

PR #24497, opened 2026-01-18
fix(core): Handle Azure AD roles array format in OIDC provisioning

tzachish1:
This fix addresses multiple issues with OIDC/SSO role provisioning when using Azure AD (Microsoft Entra ID) as the identity provider:

1. Instance role provisioning now accepts arrays
   - Azure AD sends all App Roles as an array in the 'roles' claim
   - Filters for 'global:*' roles when given an array
   - Prioritizes roles: owner > admin > member

2. Project role provisioning filters out global roles
   - Skips entries starting with 'global:' to prevent parsing 'global:admin' as projectId='global', role='admin'
   - Logs debug message explaining global roles are handled separately

3. FK constraint race condition fixed for new users
   - Moves applySsoProvisioning() call outside the user creation transaction
   - Ensures user is committed before project relations are created

4. Proxy now works for OIDC discovery requests
   - Passes customFetch to client.discovery() call
   - Ensures discovery metadata is fetched through corporate proxy

Includes unit tests for all new behavior.

## Summary

This PR fixes multiple issues with OIDC/SSO role provisioning when using Azure AD (Microsoft Entra ID) as the identity provider.

## Problem

When using Azure AD with App Roles for n8n SSO:
1. **Instance role provisioning fails** - Azure AD sends roles as an array, but n8n expected a string
2. **Project role provisioning fails** - `global:admin` is parsed as `projectId=global`, causing "project does not exist" errors  
3. **FK constraint violation** - New users get a database error because project relations are created before user is committed
4. **Proxy not used for discovery** - OIDC discovery requests bypass corporate proxy

### 1. Instance role accepts arrays
- Filters for `global:*` roles when given an array
- Prioritizes: `global:owner` > `global:admin` > `global:member`

### 2. Project roles filter out global roles
- Skips entries starting with `global:` 
- Logs debug message explaining they're handled by instance role provisioning

### 3. FK constraint fix
- Moves `applySsoProvisioning()` outside the user creation transaction
- Ensures user is committed before project relations are created

### 4. Proxy fix for discovery
- Passes `customFetch` to `client.discovery()` call
- Ensures OIDC metadata is fetched through corporate proxy

## Testing

- Added 7 unit tests covering:
  - Array input handling for instance roles
  - Role priority (admin over member)
  - Filtering global roles from project provisioning
  - Empty array handling

## Review / Merge checklist

- [ ] Tests included. 


tzachish1 (commented):
Fixed TypeScript types for proxy fetch
Fixed discovery() parameter order
Improved error logging
Userinfo fallback to ID token claims

guillaumejacquart:
Hi @tzachish1, thanks a lot for this PR, and for the detailed write-up of the Azure AD / Entra ID issue 🙏 

Before we move on the role-mapping parts of this PR, we’d like to step back and understand the underlying use case a bit better, because since you opened this we’ve shipped expression-based role mapping for SSO, which changes the picture. It lets an admin write rules that map arbitrary claims (including an array like Azure’s roles) to instance and project roles, rather than us hardcoding how a single claim is interpreted. It may already cover what you’re after, or expose a real gap we should fix properly.

A few questions to make sure we solve the right problem:

- Your Azure setup: Are all roles arriving in a single roles claim (global + project mixed together), or do you have them split across separate claims? Roughly what do the values look like?
- The goal: What are you trying to achieve with role mapping today that you can’t? e.g. assigning a specific instance role, granting project access, mapping to custom global roles, etc.
- Have you had a chance to try expression-based mapping? If yes, where does it fall short for your Azure setup? Here is [the doc](https://docs.n8n.io/administer/manage-users-and-access/verify-user-identity/use-oidc/set-up-oidc#map-rules-inside-n8n)

Separately, a couple of the other fixes here (the proxy-for-discovery change, and the FK/transaction ordering for new users) look potentially useful independent of the role-mapping question. If it’s easier could you split those out so they’re not blocked by this discussion. (Also heads-up: the diff currently touches .gitignore and removes scripts/dockerize-n8n.mjs, which look unintentional)

guillaumejacquart:
Hey @tzachish1 we're thinking of closing this PR as the main use case (roles array in the claim) can be handled through expression-based mapping. Feel free to reach out if you think there is still a legitimate case for natively accepted roles array.
We'll close that PR in a week otherwise
Thanks !

tzachish1:
> Hey @tzachish1 we're thinking of closing this PR as the main use case (roles array in the claim) can be handled through expression-based mapping. Feel free to reach out if you think there is still a legitimate case for natively accepted roles array. We'll close that PR in a week otherwise Thanks !

Hi @guillaumejacquart,
thanks for the reply - i have tested and confirm that the feature of role expression based mapping is working as expected - this PR was raised before this feature was there - the PR can be closed without merge.
Thanks.

PR #24510, opened 2026-01-18
feat(editor): Resource center experiment (no-changelog)

romeobalta:
## Summary

Implements the resource center.

https://linear.app/n8n/issue/GRO-185

## Review / Merge checklist

- [ ] Tests included. 


frstavares (approved):
Codewise it looks good!

Some small nit-pick, that might not even be related to the changes:
- We add "no setup needed", but then user has to connect the credentials (we show a "Setup template" on the top left corner of the canvas
- On hover we move the thumbnail, but not the text
- Maybe we should show some description below each card?
- When a user clicks on the ready to run workflows, we insert the workflow directly, which means that if the user clicks and comes back and clicks again, he ends up with two workflows
- We are loading a screenshot of the workflows/templates in the resource center, which do not render quite well. Does it make sense to use the preview somehow to improve this?
- When clicking a tile or "view all" we emit the same event, and maybe we should split this? Or at least have a standard on the naming convention (we use "quick-start", but then we use "Popular templates"
-

PR #24516, opened 2026-01-19
fix: Bump dependencies

shortstacked:
## Summary

Minor bump of dependnecies for
- diff
- undici
- tar

## Review / Merge checklist

- [ ] Tests included. 


package.json
```diff
@@ -129,7 +129,11 @@
       "jws@4": "4.0.1",
       "qs@6": "6.14.1",
       "@smithy/config-resolver": ">=4.4.0",
-      "@rudderstack/rudder-sdk-node@<=3.0.0": "3.0.0"
+      "@rudderstack/rudder-sdk-node@<=3.0.0": "3.0.0",
+      "diff": "8.0.3",
+      "undici@6": "^6.23.0",
+      "undici@7": "^7.18.2",
+      "tar": "^7.5.3"
```
tomi, line 136: Do we need to add these to overrides? Is some package depending on a specific version of these? Or would it be enough to just bump them without adding an override?

shortstacked: They are transitive dependencies so bumping won't work. It seems to be the case that most of the time we get these issues it's transitive so we have to do this. It might be worth looking to see if we can clean up some of hte older overrides though at least.

tomi (dismissed):

tomi (approved):

PR #24522, opened 2026-01-19
fix(editor): Html strings for table names are escaped

nikhilkuria:
## Summary

applied proper HTML escaping using the escapeHtml function from the xss library

Steps to reproduce is in the linked linear issue: https://linear.app/n8n/issue/ADO-4676/stored-html-injection-via-data-table-name-field-on-appn8ncloud-version

https://linear.app/n8n/issue/ADO-4676/stored-html-injection-via-data-table-name-field-on-appn8ncloud-version

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/@n8n/frontend-utils/src/htmlUtils.ts
+++ b/packages/frontend/@n8n/frontend-utils/src/htmlUtils.ts
@@ -1,11 +1,19 @@
 import { toValue, type MaybeRef } from 'vue';
-import xss, { escapeAttrValue } from 'xss';
+import xss, { escapeAttrValue, escapeHtml } from 'xss';
 
 import { ALLOWED_HTML_ATTRIBUTES, ALLOWED_HTML_TAGS } from './constants/sanitization';
 
 /*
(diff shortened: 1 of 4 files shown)


scdekov (commented):
I wonder if we should rather not allow such values for the name similar to what we do here https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/api-types/src/dto/user/user-update-request.dto.ts#L5

We may have to do both actually if we want to make sure that already inserted malicious values are not rendered as html

nikhilkuria:
> I wonder if we should rather not allow such values for the name similar to what we do here 

This will be a bigger change as 
- We need to figure out how to handle names with these characters already
- There could be workflows referring data tables on names.
- It's not limited to data tables

These are all considered in this breaking change for v3 https://app.notion.com/p/n8n/Reject-XSS-value-in-all-resources-names-workflows-data-tables-etc-38b5b6e0c94f8020a319d4b9ed4a80ff
So, I suggest we keep the scope simple and only escape the html now

scdekov (approved):

PR #24523, opened 2026-01-19
fix(editor): Add debounce to WebSocket disconnect to prevent race conditions (no-changelog)

alexgrozav:
## Summary

During route transitions, disconnect/connect may be called in quick succession when the old view unmounts before the new view mounts. This adds a debounce mechanism where connect always wins within a 500ms window, preventing unnecessary disconnections and reconnections.

https://linear.app/n8n/issue/CAT-2147/fix-race-conditions-on-pushconnection-while-redirecting

## Review / Merge checklist

- [x] Tests included. 


Cadiac (approved):
Seems like a pretty solid way to handle this - didn't test locally, but the logic look good to me

PR #24531, opened 2026-01-19
fix(editor): Always connect start node to trigger node

nikhilkuria:
## Summary
Fixes a bug where the Start trigger node in an extracted sub-workflow was disconnected from the first extracted node when multiple nodes were extracted via the right-click context menu. The startNodeConnection was only created when an explicit start node was identified in the extraction selection (startNodeTarget). When extracting multiple nodes without an explicit start node, startNodeTarget was undefined, causing no connection to be created. 

Solution: Always create a connection from the Start trigger node to firstNode (which is either the explicit startNodeTarget or falls back to the topmost node by Y position). This ensures the Start node is properly connected in all extraction scenarios.

https://linear.app/n8n/issue/ADO-4679/bug-start-node-disconnected-after-extract-to-sub-workflow

## Review / Merge checklist

- [ ] Tests included. 


CharlieKolb (changes requested):
The reason for the original behavior was to avoid invalid node connections like connection to a solely extracted tool:
image

We could check whether the first node has a valid `main` input connection.


The tests don't actually test anything, so I don't think they're good additions as is since they'll create a false sense of security.
Note that this code is tested via e2e tests in `packages/testing/playwright/tests/e2e/workflows/editor/subworkflows/extraction.spec.ts` instead.

CharlieKolb (approved):
🚀

PR #24533, opened 2026-01-19
chore(core): Integrate ChatHub Extractor

afitzek:
## Summary

This integrates the chat hub extractor into the chat hub endpoints, to establish an execution context for chat hub triggered workflows.

closes https://linear.app/n8n/issue/IAM-86/integrate-chathub-extractor 

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/auth/auth.service.ts
Cadiac: Should this talso take `getBrowserIdIfApplicable` into use?

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
```diff
@@ -344,4 +363,16 @@ export class ChatHubController {
 
 		res.status(204).send();
 	}
+
+	private extractChatHubTrigger(req: AuthenticatedRequest): ChatHubTriggerItem {
```
Cadiac: Is this a descriptive name for this function, and is the type name `ChatHubTriggerItem` good? Maybe something like `extractAuthMetadata`, at least this not extracting anything from the trigger node wasn't what I expected, as we also extract things from the actual chat trigger's definition at the chat hub 🤔

Or is this terminology of triggerItems something that matches other dynamic credentials code?

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
```diff
@@ -38,26 +48,38 @@ export class ChatHubExtractor implements IContextEstablishmentHook {
...
-					},
-				},
-			};
+					};
+				} else {
+					this.logger.warn('Invalid format for encryptedMetadata in chathub extractor');
+				}
+			} catch (error) {
+				this.logger.error('Failed to decrypt/parse encrypted chat metadata', {
+					error: error instanceof Error ? error.message : String(error),
```
Cadiac: we have `ensureError` helper at n8n-workflow which might be helpful here

Cadiac (commented):
Very nice, I'm excited!

Left some rather minor comments that we might want to improve before merging, but nothing major.

packages/cli/src/auth/auth.service.ts
```diff
@@ -155,6 +155,19 @@ export class AuthService {
 		};
 	}
 
+	getCookieToken(req: AuthenticatedRequest) {
+		return req.cookies[AUTH_COOKIE_NAME];
+	}
+
+	getBrowserIdIfApplicable(req: AuthenticatedRequest) {
```
BGZStephen: I'm not familiar with thise, why would we skip browser id checks?

packages/cli/src/auth/auth.service.ts
```diff
@@ -155,6 +155,19 @@ export class AuthService {
...
+		const endpoint = req.route ? `${req.baseUrl}${req.route.path}` : req.baseUrl;
+		if (req.method === 'GET' && this.skipBrowserIdCheckEndpoints.includes(endpoint)) {
+			this.logger.debug(`Skipped browserId check on ${endpoint}`);
+			return false;
```
BGZStephen: false feels a little odd as a return value here, isn't `null` more accurate?

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
```diff
@@ -6,21 +6,31 @@ import {
...
+});
+
+const ChatHubAuthenticationMetadataSchema = z.object({
 	authToken: z.string(),
 	browserId: z.string().optional(),
 });
 
-export type ChatHubTriggerItem = z.output<typeof ChatHubTriggerItemSchema>;
+export type ChatHubAuthenticationMetadata = z.output<typeof ChatHubAuthenticationMetadataSchema>;
+export const CHATHUB_EXTRACTOR_NAME = 'ChatHubExtractor';
```
BGZStephen, line 28: Worth exporting these to a central constants file?

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
```diff
@@ -38,26 +48,38 @@ export class ChatHubExtractor implements IContextEstablishmentHook {
...
 						},
-					},
-				},
-			};
+					};
+				} else {
+					this.logger.warn('Invalid format for encryptedMetadata in chathub extractor');
+				}
+			} catch (error) {
+				this.logger.error('Failed to decrypt/parse encrypted chat metadata', {
```
BGZStephen, line 106: This isn't strictly accurate, parsing has already occurred and succeeded due to `encryptedMetadataResult.success`

Should we also be adding a debug error message for failed parsing of encryptedMetadataResult?

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
```diff
@@ -344,4 +363,16 @@ export class ChatHubController {
...
+
+	private extractAuthenticationMetadata(req: AuthenticatedRequest): ChatHubAuthenticationMetadata {
+		const authToken = this.authService.getCookieToken(req);
+		if (!authToken) {
+			throw new UnauthenticatedError('No authentication token found');
+		}
+		const browserId = this.authService.getBrowserIdIfApplicable(req);
+		return {
+			authToken,
+			browserId: browserId === false ? undefined : browserId,
```
BGZStephen: Code smell resulting in odd return value from `getBrowserIdIfApplicable`:  `false ? undefined`

BGZStephen (changes requested):

packages/cli/src/auth/auth.service.ts
afitzek: There is a condition, if the method is GET and the endpoint is in a special list where we should skip the browserId check. Why this needs to happen, I'm not familiar with.

packages/cli/src/auth/auth.service.ts
afitzek: I thought about null as well, but I wanted to make this more explicit. Basically the data type needs to transport 3 meanings:
1. The actual browserId (some string)
2. The fact that the browserId is not present on the request
3. The option to skip the browserId check all together

Therefore I thought the type string | undefined | false seems like a valid approach. But we could make the data object more explicit maybe. 🤔

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
afitzek: No there is a call to jsonParse, just after the decryption function, this is meant with the parse part of this message.

packages/cli/src/auth/auth.service.ts
BGZStephen: I think undefined seems right given the code usage later?

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
afitzek: I think this part is wrong and should be changed to also store the false value, since we need to transport this information through the workflow execution and the token validation in the resolver.

packages/cli/src/auth/auth.service.ts
afitzek: But just undefined is not enough, because this could mean, that either the check can be skipped or that there is no id on the request object. 🤔

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
BGZStephen: Sure, but there is no logging in the event this parse fails?

packages/cli/src/auth/auth.service.ts
BGZStephen: Perhaps it needs to be split in two? browserIdCheckPerformed and browserId?

To allow for a null value in browserId and a truthy / falsey value for whether the id check was performed?

BGZStephen: I think the main comment here is that there is a mixed set of concerns for this function, if it were split in to two:

`shouldBrowserIdCheckBePerformed` and `getBrowserId`, which would possibly make it more SOLID

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
afitzek: Not sure what you mean? Isn't this log line the logging in the event this parse fails?

BGZStephen: No, line 79 has a safeParse check for `if (encryptedMetadataResult.success) {`, this silently capture schema validation failures, this catch only real captures errors from `const parsed = jsonParse(decrypted);`

BGZStephen (dismissed):

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
afitzek: Ah now I get it 👍 , good catch

packages/cli/src/auth/auth.service.ts
Cadiac: There are some requests that need to do this, like when an attachment is opened (in a new tab?), those are GET cases where it's just the browser sending the request without JS in the mix and nothing sets the browser-id header I believe

packages/cli/src/modules/chat-hub/chat-hub-extractor.ts
Cadiac: there's `n8n/packages/cli/src/modules/chat-hub/chat-hub.constants.ts` but happy to just keep these here for now, we've been moving things around a lot anyways.

afitzek: I would also leave it here, I think it is a quite narrow constant, leaving it near the relevant class makes sense in my mind.

Cadiac (approved):
👍 tested this a bit with debugger and as far as I understand this seems to work

PR #24548, opened 2026-01-19
fix(core): Worker status in multi-main only accessible to admin

konstantintieber:
## Summary

While the [sidebar navigation item](https://github.com/n8n-io/n8n/blob/master/packages/frontend/editor-ui/src/app/composables/useSettingsItems.ts#L107) for the worker overview checks for workersView:manage scope to be present, the vue router definition does not.

This PR [adds](https://github.com/n8n-io/n8n/pull/24548/commits/f753eca53e6f15b9b6355f13b0e6649a620f30e6) the missing route guard, and [will only publish](https://github.com/n8n-io/n8n/pull/24548/commits/47c45cd928421808713c67a3c23e387bc332c5f3) the worker status updates to the user who originally requested them.

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/scaling/worker-status.service.ee.ts
+++ b/packages/cli/src/scaling/worker-status.service.ee.ts
@@ -20,29 +20,39 @@ export class WorkerStatusService {
 		private readonly push: Push,
 	) {}
 
-	async requestWorkerStatus() {
+	async requestWorkerStatus(requestingUserId: string) {
 		if (this.instanceSettings.instanceType !== 'main') return;
 
-		return await this.publisher.publishCommand({ command: 'get-worker-status' });
-	}
-
-	@OnPubSubEvent('response-to-get-worker-status', { instanceType: 'main' })
-	handleWorkerStatusResponse(payload: WorkerStatus) {
-		this.push.broadcast({
-			type: 'sendWorkerStatusMessage',
-			data: {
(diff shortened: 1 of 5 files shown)


packages/cli/src/scaling/worker-status.service.ee.ts
```diff
@@ -20,29 +20,39 @@ export class WorkerStatusService {
 		private readonly push: Push,
 	) {}
 
-	async requestWorkerStatus() {
+	async requestWorkerStatus(requestingUserId: string) {
```
geemanjs, line 23: Does n8n have a preference (coding style) when we pass a parameter vs object?

For me personally (back in the java world), we'd always pass an object into a `service` to give it fast flexibility for extension. 

But if we have no preference maybe we stick with 1-2 param fine, more than that and we wrap it into an object?

geemanjs (approved):

packages/cli/src/scaling/worker-status.service.ee.ts
konstantintieber: In the n8n codebase I've observed that we don't use object params by default. I think up to three params in a parameter list are fine.
And if it goes beyond that, there's also the additional question if another function can be extracted and those then composed.

PR #24551, opened 2026-01-19
chore: Add ExternalSecretProvider entity

ireneea:
## Summary

- Add `SecretsProviderConnection`  (name, type, settings, state, connection status)
- Add `SharedSecretsProviderConnection` entity to manage project-level sharing with role-based access
- Create `SecretsProviderConnectionRepository` with findAll method
- Define externalSecretsProvider:owner role with read, update, delete, and sync scopes
- Extend RoleCacheService to support the new externalSecretsProvider role namespace

**Notes**
- There will be a follow-up task to the migration that creates the new tables LIGO-131

closes LIGO-110

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/external-secrets.ee/database/entities/external-secrets-provider.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+export type ExternalSecretsProviderState = 'connected' | 'tested' | 'initializing' | 'error';
+
+@Entity()
+@Index(['name', 'projectId'], { unique: true })
+export class ExternalSecretsProvider extends WithTimestampsAndStringId {
+	@Column()
+	name: string;
+
+	@Column()
+	displayName: string;
```
ireneea: Need to confirm how this will actually be used (if it's used)

packages/cli/src/modules/external-secrets.ee/database/entities/external-secrets-provider.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	@Column()
+	type: string;
+
+	@ManyToOne(() => Project, { onDelete: 'CASCADE', nullable: true })
+	@JoinColumn({ name: 'projectId' })
+	project: Project | null;
+
+	@Column({ nullable: true })
+	projectId: string | null;
```
ireneea: Need to confirm if we want to go with a sharing model, and allow a secret to be shard with multiple projects

packages/@n8n/db/src/entities/secrets-provider-connection.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { DateTimeColumn, WithTimestampsAndStringId } from './abstract-entity';
+import type { SharedSecretsProviderConnection } from './shared-secrets-provider-connection';
+
+export type SecretsProviderConnectionState = 'connected' | 'tested' | 'initializing' | 'error';
+
+@Entity()
+export class SecretsProviderConnection extends WithTimestampsAndStringId {
+	@Column({ unique: true })
+	name: string;
```
geemanjs: We're we going to use the `id` field for this instead? Should we put some "length" constraints on it?

packages/@n8n/db/src/entities/secrets-provider-connection.ts
```diff
@@ -0,0 +1,39 @@
...
+	displayName: string | null;
+
+	@Column()
+	type: string;
+
+	/**
+	 * Whether the secrets provider connection is available for use by all projects.
+	 */
+	@Column({ default: false })
+	isGlobal: boolean;
```
geemanjs: Should we change this to an "string/enum" type value? it might enable us to scope them differently in the future?

```
scope: "GLOBAL" | "PROJECT" 
```

packages/@n8n/db/src/entities/secrets-provider-connection.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	@OneToMany('SharedSecretsProviderConnection', 'secretsProviderConnection')
+	shared: SharedSecretsProviderConnection[];
+
+	@Column()
+	settings: string;
+
+	@Column({ default: false })
+	connected: boolean;
```
geemanjs: Should we lock boolean values to `isXYZ`?

packages/@n8n/db/src/entities/secrets-provider-connection.ts
```diff
@@ -0,0 +1,61 @@
...
+	 * Whether the secrets provider connection is enabled.
+	 * When enabled, a connection attempt will be made to the external secrets provider.
+	 * If the connection is successful, secrets will be available to be used in credentials.
+	 *
+	 * When disabled, the secrets provider connection will not be used to connect to the external secrets provider.
+	 *
+	 * This describes an intent rather than a state.
+	 */
+	@Column({ default: false })
+	enabled: boolean;
```
geemanjs: Likewise here should we make it `isEnabled`? I'm unsure if there are pre-existing standards here

packages/@n8n/db/src/entities/secrets-provider-connection.ts
```diff
@@ -0,0 +1,61 @@
...
+	 * Shared secrets provider connections are used to share the secrets provider connection with other projects.
+	 */
+	@OneToMany('SharedSecretsProviderConnection', 'secretsProviderConnection')
+	shared: SharedSecretsProviderConnection[];
+
+	/**
+	 * Encrypted JSON string containing the connection settings for the secrets provider.
+	 */
+	@Column()
+	settings: string;
```
geemanjs: Should we call this `encryptedSettings` to convey the reality better?

packages/@n8n/db/src/entities/secrets-provider-connection.ts
geemanjs: This might actually be a bad idea 🤔 - more wondering your thoughts?

packages/@n8n/db/src/entities/secrets-provider-connection.ts
konstantintieber: Added a related comment on the dto PR here, proposing that we call this `id` instead of `name`: https://github.com/n8n-io/n8n/pull/24718#discussion_r2717520118

packages/@n8n/db/src/entities/secrets-provider-connection.ts
konstantintieber: Or is a provider connection automatically global as long as it is not connected to any project?

packages/@n8n/db/src/entities/secrets-provider-connection.ts
ireneea: Decided to keep an `id` field and a separate `name` field. The `id` comes from the `extends WithTimestampsAndStringId`. 

Having an `id` makes it more consistent with the other "resources".
- consistent API
- consistent owner resolution

packages/@n8n/db/src/entities/secrets-provider-connection.ts
ireneea: Because we now have the "shared" table there will always be a project associated with a connection.

Again this is consistent with the current "sharing" feature where there is always an owner.
The owner is either a team project or a personal project. This means that we explicitly need an `isGlobal` flag to identify secrets that are shared with all projects

packages/@n8n/db/src/entities/secrets-provider-connection.ts
konstantintieber: What do you mean by `consistent owner resolution`?

What do you think about still somehow choosing a value for this property that somehow conveys that it's an id.
For example calling it `providerId`.

packages/@n8n/db/src/entities/secrets-provider-connection.ts
konstantintieber: > The owner is either a team project or a personal project

Associating the personal project of a user with a global provider connection feels weird to me. Feels like something I wouldn't want to copy from a sharing feature that was implemented somewhere else.

What speaks against:
- a provider connection can either be global (no project associated), or shared, by being associated with a project.

packages/@n8n/db/src/entities/secrets-provider-connection.ts
konstantintieber: Or another idea for a "uniqueness associated name": `key`

geemanjs: `providerKey`, `key` I think may be too close to `id`, `providerId`, `slug`  - also I have not overly looked into how we do database tables but we will need an index on whatever field name we choose.

ireneea: I like `providerKey` 👍

ireneea: As discussed, there is no clear argument for not making the `providerKey` the primary column. That has now been changed

packages/@n8n/db/src/entities/secrets-provider-connection.ts
ireneea: Drastically changed this to get away from the current "share" feature. Instead we'll just have a simple join table (that we can extent later if needed). 

No need for an `isGlobal` field anymore.  A `SecretProviderConnection` is global if it has no links to a project. It's project-scoped if it has at least one link to a project. let me know if this makes more sense.

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -0,0 +1,15 @@
+import { Service } from '@n8n/di';
+import { DataSource, Repository } from '@n8n/typeorm';
+
+import { SecretsProviderConnection } from '../entities';
+
+@Service()
+export class SecretsProviderConnectionRepository extends Repository<SecretsProviderConnection> {
```
ireneea, line 7: This is just a placeholders the actual method needed will become apparent when we start implementing the consumers of this class

konstantintieber (approved):
I like that we go for a simpler sharing relationship schema here. Talking about the value of an approach like in https://github.com/n8n-io/n8n/pull/24551/commits/b1bfe50a2486cc04075c005bb97459bedf33ef96 may be a good topic for the next architecture meeting.
Where someone can help us understand the always-project-based ownership concept.

PR #24557, opened 2026-01-19
fix(core): Workflow tag removal syncing

geemanjs:
## Summary
Fixes a bug where removing tags from workflows did not sync the removal to the tags.json file on export and then also failed to import the changes correctly.

This lead to inconsistencies between instances when importing/export tags with source control

BONUS - I also added additional tests for tags as they seemed relatively light beforehand. Now we cover export and import. 

I think we'll cover it in other improvements but there are still other problems in this space. The main one being that tags can end up "orphaned" in the second instance. 

LIGO-62

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
+++ b/packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
@@ -537,10 +537,7 @@ export class SourceControlImportService {
 		return { tags: [], mappings: [] };
 	}
 
-	async getLocalTagsAndMappingsFromDb(context: SourceControlContext): Promise<{
-		tags: TagEntity[];
-		mappings: WorkflowTagMapping[];
-	}> {
+	async getLocalTagsAndMappingsFromDb(context: SourceControlContext): Promise<ExportableTags> {
 		const localTags = await this.tagRepository.find({
 			select: ['id', 'name'],
 		});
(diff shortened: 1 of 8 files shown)


packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
```diff
@@ -861,7 +858,7 @@ export class SourceControlImportService {
 		return importCredentialsResult.filter((e) => e !== undefined);
 	}
 
-	async importTagsFromWorkFolder(candidate: SourceControlledFile) {
+	async importTagsFromWorkFolder(candidate: SourceControlledFile, user: User) {
```
geemanjs, line 861: @ireneea are we ok to bring the user into this context? I wasn't 100% sure but I think the way this works is fine. When we import we only import workflows for the current user so doing the same in tags feels correct 👀

konstantintieber (approved):
Reviewing a PR with good tests is the most fun 👍 

Link to recording of me testing this manually: https://www.loom.com/share/786de5a0326d4a39a9d2d9d628da5d3f

packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
ireneea: Yup that makes sense, even if tags themselves are not restricted, since we are updating workflows it does make sense to only touch workflows that the user has access to 👍

ireneea (approved):

PR #24568, opened 2026-01-20
feat(core): Implement execution data stores

ivov:
https://linear.app/n8n/issue/CAT-2094
https://linear.app/n8n/issue/CAT-2095


packages/cli/src/executions/execution-data/__tests__/mocks.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const workflowData: IWorkflowBase = {
+	id: workflowId,
+	name: 'Test Workflow',
+	nodes: [],
+	connections: {},
+	active: false,
+	isArchived: false,
+	activeVersionId: null,
+	createdAt: '2026-01-01T00:00:00.000Z' as unknown as Date,
```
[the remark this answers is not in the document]
ivov: Pre-existing type mismatch. Needs fixing upstream, out of scope.

packages/cli/src/executions/execution-data/corrupted-execution-data.error.ts
```diff
@@ -0,0 +1,15 @@
+import { UnexpectedError } from 'n8n-workflow';
+
+import type { ExecutionRef } from './types';
+
+export class CorruptedExecutionDataError extends UnexpectedError {
```
tomi, line 9: Can we comment these and put them into their own files for discoverability

packages/cli/src/executions/execution-data/fs-store.ts
```diff
@@ -0,0 +1,85 @@
...
+			this.storageConfig.storagePath,
+			'workflows',
+			workflowId,
+			'executions',
+			executionId,
+		);
+	}
+
+	private resolveBundlePath(ref: ExecutionRef) {
+		return path.join(this.resolveExecutionDir(ref), 'execution_data', 'data.json');
```
tomi: Should we also call the file `bundle.json` if we use that term in the code?

packages/cli/src/executions/execution-data/fs-store.ts
```diff
@@ -0,0 +1,85 @@
...
+		await assertDir(path.dirname(writePath));
+
+		// for atomicity, first write to temp file and then rename
+		const tempPath = `${writePath}.tmp.${Date.now()}`;
+		let success = false;
+
+		try {
+			await fs.writeFile(
+				tempPath,
+				JSON.stringify({ ...payload, version: EXECUTION_DATA_BUNDLE_VERSION }),
```
tomi: We should use `jsonStringify` to be compatible with what we have now

packages/cli/src/executions/execution-data/fs-store.ts
```diff
@@ -0,0 +1,85 @@
...
+				'utf-8',
+			);
+			await fs.rename(tempPath, writePath);
+			success = true;
+		} catch (error) {
+			throw new ExecutionDataWriteError(ref, error);
+		} finally {
+			if (!success) await fs.rm(tempPath, { force: true });
```
tomi: If `rm` throws it can hide the original exception. We could eg


```suggestion
			if (!success) await fs.rm(tempPath, { force: true }).catch(e => this.errorReporter.error(e))
```

packages/cli/src/executions/execution-data/fs-store.ts
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+
+	async read(ref: ExecutionRef) {
+		const bundlePath = this.resolveBundlePath(ref);
+
+		if (!(await exists(bundlePath))) return null;
+
+		const content = await fs.readFile(bundlePath, 'utf-8');
```
tomi: I would imagine a big majority of time we are reading files that exist, right? In that case we could optimize for that path, and read directly and map `ENOENT` to `null`

packages/cli/src/executions/execution-data/__tests__/db-store.integration.test.ts
```diff
@@ -0,0 +1,139 @@
...
+beforeAll(async () => {
+	await testDb.init();
+	repository = Container.get(ExecutionDataRepository);
+	executionRepository = Container.get(ExecutionRepository);
+	dbStore = Container.get(DbStore);
+});
+
+beforeEach(async () => {
+	await testDb.truncate(['ExecutionData', 'ExecutionEntity', 'WorkflowEntity']);
+	await createWorkflow({ id: workflowId });
```
tomi: nit: We could also create a single WF in `beforeAll` and reuse that for all to speed this up a little bit. Or might there be issues that way?

packages/cli/src/executions/execution-data/types.ts
```diff
@@ -1,19 +1,23 @@
...
-type ExecutionDataPayload = {
+export type ExecutionDataPayload = {
 	data: string;
 	workflowData: IWorkflowBase;
 	workflowVersionId: string | null;
 };
 
-export interface ExecutionDataManager {
+export type ExecutionDataBundle = ExecutionDataPayload & {
+	version: number;
```
tomi: Should this be `version: 1`?

packages/cli/src/executions/execution-data/types.ts
```diff
@@ -1,19 +1,23 @@
 import type { IWorkflowBase } from 'n8n-workflow';
 
-type ExecutionRef = {
+export type ExecutionRef = {
 	workflowId: string;
 	executionId: string;
 };
```
tomi, line 6: I think it would be nice to have a factory function for this to ensure all instantiations are centralized

```ts
export function createExecutionRef(workflowId: string, executionId: string) {
  return ({
    workflowId, executionId
  })
}
```

tomi (commented):
Nice stuff 👏 Couple comments

tomi (approved):
Thank you for addressing the feedback 🫶 LGTM 🚀

PR #24591, opened 2026-01-20
feat(editor): Add Dialog component to design system

heymynameisrob:
## Summary

Adds new Dialog and AlertDialog components to the n8n design system (`@n8n/design-system`).

### N8nDialog
A composable, accessible dialog component built on `reka-ui` primitives:

- **N8nDialog** - Pre-composed dialog component with controlled state support
- **N8nDialogClose** - Close button primitive
- **N8nDialogContent** - Main dialog container with size variants and animations
- **N8nDialogOverlay** - Backdrop overlay with fade animation
- **N8nDialogHeader** - Header container for title and description
- **N8nDialogFooter** - Footer container for actions
- **N8nDialogTitle** - Accessible title component
- **N8nDialogDescription** - Accessible description component

Features:
- **Controlled state preferred** - Uses `open` prop with `update:open` emit for predictable state management
- Supports `defaultOpen` for uncontrolled usage when needed
- Shorthand `header` and `description` props for common dialog patterns
- Multiple sizes: `small`, `medium`, `large`, `xlarge`, `2xlarge`, `fit`, `full`, `cover`
- Built-in accessibility with ARIA support and screen reader fallbacks
- Keyboard navigation and focus trapping
- Customizable close button visibility
- Smooth enter/exit animations with reduced motion support

### N8nAlertDialog
A pre-composed alert dialog for confirmations and destructive actions:

- Controlled open state via `v-model:open`
- Configurable action button variants (`solid`, `destructive`)
- Customizable button labels
- Loading state support
- Sizes: `small`, `medium`

### Additional Changes
- Added focus ring mixin (`_focus.scss`) for consistent focus styles
- Added Storybook stories with comprehensive examples
- Added unit tests for both components

## Related Linear tickets, GitHub issues, and Community forum posts
https://linear.app/n8n/issue/DS-354/component-dialog

## Review / Merge checklist

- [x] Tests included.


packages/frontend/@n8n/design-system/src/components/N8nDialog/Dialog.stories.ts
```diff
@@ -0,0 +1,456 @@
...
+		setup() {
+			return { args };
+		},
+		template: `
+		<div style="display: flex; justify-content: center; padding: 40px;">
+			<N8nDialogRoot>
+				<N8nDialogTrigger as-child>
+					<N8nButton label="Open Dialog" />
+				</N8nDialogTrigger>
+				<N8nDialogPortal>
```
konstantintieber: These `N8nDialogPortal` and `N8nDialogOverlay` seem to be here purely for technical reasons. I would even call them boilerplate.

Can we abstract those away for the consumer, so that using our dialog component is as simple as this?

Note: the `header` prop on the N8nDialog element is intended as a "shorthand alternative" to using a `N8nDialogHeader` child.
```html
<N8nButton @click="isOpen = !isOpen" label="Open Dialog" />

<N8nDialog v-model:open="isOpen" header="Edit Profile" >
   <N8nDialogContent>
        <div style="padding: var(--spacing--sm) 0;">
			<p>Dialog content goes here...</p>
		</div>
   </N8nDialogContent>
   <N8nDialogFooter>
	  <N8nDialogClose as-child>
		<N8nButton type="secondary" label="Cancel" />
	  </N8nDialogClose>
	  <N8nButton label="Save changes" />
   </N8nDialogFooter>
</N8nDialog>
```

packages/frontend/@n8n/design-system/src/components/N8nDialog/Dialog.stories.ts
```diff
@@ -0,0 +1,456 @@
...
+			N8nDialogClose,
+			N8nButton,
+		},
+		setup() {
+			return { args };
+		},
+		template: `
+		<div style="display: flex; justify-content: center; padding: 40px;">
+			<N8nDialogRoot>
+				<N8nDialogTrigger as-child>
```
konstantintieber: What do you think about using  the "controlled component" style with `v-model:open="isOpen"` on the `N8nDialogRoot` in our story implementation?
I would imagine that while I'm using dialogs in my views out there, I would rather use the controlled API than having the trigger as a child element of my `N8nDialogRoot`, as that makes for more reusable and readable template code.

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogDescription.vue
```diff
@@ -0,0 +1,27 @@
...
+	 * Merge props onto child element
+	 */
+	asChild?: boolean;
+}
+
+defineProps<DialogDescriptionProps>();
+</script>
+
+<template>
+	<DialogDescription :as-child="asChild" :class="$style.description">
```
konstantintieber, line 15: What's the intention behind the `N8nDialogHeader description`? I can't remember coming across a dialog in our UI that had such a description.

konstantintieber (commented):
The UI of the AlertDialog and Dialog looks really good 👍 

I still have some questions on the component API for consumers.

packages/frontend/@n8n/design-system/src/components/N8nDialog/Dialog.stories.ts
heymynameisrob: Yeah - we could include Overlay and Portal as part of the DialogContent by default so we don't need to define them every time. That would be better.

packages/frontend/@n8n/design-system/src/components/N8nDialog/Dialog.stories.ts
heymynameisrob: You can do that it's just not required. It means you don't need to control the state from outside the component if it's just a simple dialog but it's possible to:
https://reka-ui.com/docs/components/dialog#root

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogDescription.vue
heymynameisrob: Dialogs should have a description for a11y purposes and the DialogDescription adds that whilst making it visible on the UI. It's an optional component as you can just pass `ariaDescription` if you don't want it visually.

It also standardises any subtitle we have in Dialog UI

packages/frontend/@n8n/design-system/src/components/N8nDialog/Dialog.stories.ts
konstantintieber: I saw that it's supported, what I'm trying to say is that storybook template code is a convenient reference for frontend developers at n8n to "see how they should use this component".
And I would argue that the "controlled dialog" approach should be the best practice in the n8n codebase, because it will result in more readable templates.

A button that opens a modal may be part of a group of three buttons on a page. If you then use the "trigger as child" approach, the html markup of that button group is going to be much less readable.
While a controlled approach allows the frontend dev to have a clean button group and the modal markup definition at the bottom

heymynameisrob: > And I would argue that the "controlled dialog" approach should be the best practice in the n8n codebase, because it will result in more readable templates.

Makes sense

packages/frontend/@n8n/design-system/src/components/N8nAlertDialog/AlertDialog.vue
```diff
@@ -0,0 +1,123 @@
...
+export type AlertDialogSize = 'small' | 'medium';
+
+export interface AlertDialogProps {
+	/**
+	 * Controlled open state of the alert dialog
+	 */
+	open?: boolean;
+	/**
+	 * The open state of the alert dialog when it is initially rendered.
+	 * Use when you do not need to control its open state.
```
konstantintieber, line 22: A+ doc blocks

packages/frontend/@n8n/design-system/src/components/N8nAlertDialog/AlertDialog.test.ts
```diff
@@ -0,0 +1,443 @@
...
+			});
+
+			await waitFor(() => {
+				expect(getByRole('dialog')).toBeInTheDocument();
+			});
+		});
+	});
+
+	describe('action button behavior', () => {
+		it('should NOT automatically close on action (allows async operations)', async () => {
```
konstantintieber, line 387: As usual, great tests 👍

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
```diff
@@ -0,0 +1,224 @@
+<script setup lang="ts">
+import { DialogContent, DialogTitle, DialogDescription, VisuallyHidden } from 'reka-ui';
+import { computed, useCssModule } from 'vue';
+
+import N8nDialogClose from './DialogClose.vue';
+
+export type DialogContentSize =
```
konstantintieber, line 7: Non-merge-blocking:
Why do we need this many dialog sizes?
 I wonder if we need that many, given that we also have the separate `AlertDialog` for small ones.
It is however better than what we currently have: ElPlus modals where custom width and height props are passed in templates: [Example](https://github.com/n8n-io/n8n/blob/master/packages/frontend/editor-ui/src/app/components/WhatsNewModal.vue#L85-L87)
Also: Haven't seen a fullscreen dialog in the n8n UI yet. Is this something we want to adopt? From my experience fullscreen dialogs often end up being abused for views that should've been a dedicated page.

konstantintieber (approved):
Looking good

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
heymynameisrob: It's a good point and one we debated in the design team. Usually we can be the gate for this, defining them at the design stage but it's something to keep an eye on.

> From my experience fullscreen dialogs often end up being abused for views that should've been a dedicated page.

I agree with this too but my thoughts were if we end up using them at least we can implement them well and consistently.

konstantintieber: Agreed.

PR #24595, opened 2026-01-20
feat(ai-builder): Providing instructions from workflow builder on creation of data tables

mike12345567:
## Summary

This PR addresses: https://linear.app/n8n/issue/AI-1601/ai-workflow-builder-should-instruct-user-to-create-datatable

I have introduced three changes in this PR:
1. Pushing the builder/configurator sub-agents to use a set node before the data table to map the columns onto the data table. Benefit of this approach is that the data table node can map automatically to the set node columns - the usage of the manual mapping can cause problems as the table doesn't exist, therefore there are no columns to map onto. This avoid mis-configured data table nodes that are directly connected to agents/other processing nodes and don't map onto the columns in the data table correctly.
2. We can use the set node before the data table to correctly describe the list of columns that need to be created in the user table. We extract this information and feed it to the responder agent to make sure it consistently reports the correct columns to create - we do not want users to follow the instructions and then create a column incorrectly/un-necessarily.
3. Uses a placeholder for the data table name, previously it was inserting the name it wanted to use for the data table and this was causing some confusion as it looked like the node was configured, but in reality the data table ID hadn't been selected properly. This also means that it shows up in the todo list.

Also quite a few test cases for some of this new functionality.

Example of response:
image

Example of workflow generated with set nodes:
image

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/validation/checks/connections.ts
```diff
@@ -133,6 +138,47 @@ function checkMergeNodeConnections(
 	return issues;
 }
 
+function checkDataTableHasSetNodePredecessor(
```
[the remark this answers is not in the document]
mike12345567: Test cases have been added.

packages/@n8n/ai-workflow-builder.ee/src/utils/data-table-helpers.ts
```diff
@@ -0,0 +1,194 @@
...
+	/** The name of the Set node that defines the columns (if found) */
+	setNodeName?: string;
+	/** The operation (insert, update, upsert, get, delete) */
+	operation: DataTableRowOperation;
+}
+
+/**
+ * Find nodes that connect to a target node
+ */
+export function findPredecessorNodes(workflow: SimpleWorkflow, targetNodeName: string): string[] {
```
OlegIvaniv: Couldn't we reuse some methods from workflows package instead of reimplementing nodes traversal? 

```
// Before
const predecessors = findPredecessorNodes(workflow, node.name);

// After
const connectionsByDestination = mapConnectionsByDestination(workflow.connections);
const predecessors = getParentNodes(connectionsByDestination, node.name, NodeConnectionTypes.Main, 1);
```

packages/@n8n/ai-workflow-builder.ee/src/validation/checks/connections.ts
```diff
@@ -138,6 +138,48 @@ function checkMergeNodeConnections(
...
+	// Only check for Set node on row column operations (insert, update, upsert)
+	// Read operations (get, getAll) and delete don't need a Set node
+	const operationParam = node.parameters?.operation;
+	const operation = typeof operationParam === 'string' ? operationParam : 'insert';
+	if (!isDataTableRowColumnOperation(operation)) {
+		return [];
+	}
+
+	// Check if any predecessor is a Set node
+	const hasSetNodePredecessor = Object.entries(workflow.connections).some(([sourceName, outputs]) =>
```
OlegIvaniv: If we pass already computed `connectionsByDestination: IConnections` we could simplify this:
```
const predecessors = getParentNodes(connectionsByDestination, node.name, NodeConnectionTypes.Main, 1);
        const hasSetNodePredecessor = predecessors.some((name) =>
                nodesByName.get(name)?.type === SET_NODE_TYPE
        )
```

packages/@n8n/ai-workflow-builder.ee/src/utils/data-table-helpers.ts
```diff
@@ -0,0 +1,194 @@
...
+
+import type { DataTableRowOperation } from 'n8n-workflow';
+
+import type { SimpleWorkflow } from '../types';
+
+export const DATA_TABLE_NODE_TYPE = 'n8n-nodes-base.dataTable';
+export const SET_NODE_TYPE = 'n8n-nodes-base.set';
+
+// Re-export type from n8n-workflow for convenience
+export type { DataTableRowOperation };
```
OlegIvaniv: IMO not worthy, creates ambiguity when importing, n8n-workflow should be source of truth for this type

packages/@n8n/ai-workflow-builder.ee/src/prompts/agents/responder.prompt.ts
```diff
@@ -71,6 +76,72 @@ export function buildGeneralErrorGuidance(): string {
...
+					`(the "${table.nodeName}" node uses the "${table.operation}" operation)`,
+			);
+		}
+
+		parts.push('4. After creating the table, select it in the Data Table node');
+		parts.push('');
+	}
+
+	return parts.join('\n');
+}
```
OlegIvaniv, line 132: This is quite hard to follow :D Also it has a bug step order bug for multiple data-tables:
```
  responder.prompt.ts:102-129:
  parts.push('1. Go to the [Data Tables tab]...');

  for (const table of dataTables) {
      parts.push(`2. Click "Create Data Table"...`);  // Always "2"
      parts.push('3. Add these columns...');           // Always "3"
      parts.push('4. After creating...');              // Always "4"
  }

  With 2 tables, output is: 1, 2, 3, 4, 2, 3, 4
```

```suggestion

/**
 * Build column guidance for a single data table based on operation type.
 */
function buildColumnGuidance(table: DataTableInfo): string {
	if (!isDataTableRowColumnOperation(table.operation)) {
		return (
			'Ensure the table has the columns you want to read/query ' +
			`(the "${table.nodeName}" node uses the "${table.operation}" operation)`
		);
	}

	if (table.columns.length > 0) {
		const columnList = table.columns.map((col) => `   - \`${col.name}\` (${col.type})`).join('\n');
		const sourceNote = table.setNodeName
			? `\n   (These columns are defined in the "${table.setNodeName}" node)`
			: '';
		return `Add these columns with the following types:\n${columnList}${sourceNote}`;
	}

	if (table.setNodeName) {
		return (
			`Add columns matching the fields defined in the "${table.setNodeName}" node. ` +
			'Open that node to see the field names and types.'
		);
	}

	return 'Add columns based on the data you want to store';
}

/**
 * Build setup steps for a single data table.
 */
function buildTableSteps(table: DataTableInfo, startStep: number): string {
	const steps = [
		`${startStep}. Click "Create Data Table" and name it appropriately for the "${table.nodeName}" node`,
		`${startStep + 1}. ${buildColumnGuidance(table)}`,
		`${startStep + 2}. After creating the table, select it in the Data Table node`,
	];
	return steps.join('\n');
}

const DATA_TABLE_INTRO = `The workflow uses Data Table nodes, but data tables must be created manually.
Do NOT tell the user that data tables will be created automatically.

Include the following steps in your response to the user:`;

/**
 * Build guidance for data table creation.
 * Users need to manually create data tables since the AI workflow builder cannot create them automatically.
 *
 * For row column operations, column definitions are inferred from the Set node
 * (n8n-nodes-base.set) that precedes each Data Table node.
 * For read operations (get, getAll, delete), no column guidance is needed.
 */
export function buildDataTableCreationGuidance(dataTables: DataTableInfo[]): string {
	if (dataTables.length === 0) {
		return '';
	}

	const STEPS_PER_TABLE = 3;
	const tableSteps = dataTables
		.map((table, index) => buildTableSteps(table, 2 + index * STEPS_PER_TABLE))
		.join('\n\n');

	return prompt({ format: 'markdown', separator: '\n\n' })
		.section('Data Table Setup Required', DATA_TABLE_INTRO)
		.section('Steps to Create Data Tables', `1. Go to the [Data Tables tab](/home/datatables)\n${tableSteps}`)
		.build();
}

```

OlegIvaniv (commented):
Nice job! Left couple of comments for discussion

OlegIvaniv (approved):

PR #24621, opened 2026-01-21
fix(core): Prevent dual loading of community packages

ivov:
## Summary

When a worker receives a `community-package-install` or `community-package-update` pubsub event, it calls `installOrUpdateNpmPackage`. This loads the package directly without unloading it first, so if the package is already loaded (e.g., the worker loaded it from disk at startup), `loadPackage` throws because the package already exists in `loaders`. 

https://github.com/n8n-io/n8n/blob/78a16b70314f65d9a5dca29d92a43a5d7a7131f3/packages/cli/src/modules/community-packages/community-packages.service.ts#L496-L502

Compare to `installPackage` which unloads first:

https://github.com/n8n-io/n8n/blob/78a16b70314f65d9a5dca29d92a43a5d7a7131f3/packages/cli/src/modules/community-packages/community-packages.service.ts#L442-L443

To reproduce:

1. Start main + worker
2. Install a community package like `n8n-nodes-baserow`
3. Restart only the worker
4. Update the community package → `error: nodes package n8n-nodes-baserow is already loaded. Please delete this second copy at path /Users/ivov/.n8n/nodes/node_modules/n8n-nodes-baserow { "file": "error-reporter.js", "function": "defaultReport" }`

n/a

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
+++ b/packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
@@ -810,4 +810,26 @@ describe('CommunityPackagesService', () => {
 			);
 		});
 	});
+
+	describe('handleInstallEvent', () => {
+		test('should call unloadPackage before loadPackage to handle already-loaded packages', async () => {
+			const callOrder: string[] = [];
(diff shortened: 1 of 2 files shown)


michael-radency (approved):
LGTM

PR #24623, opened 2026-01-21
fix(Azure AI Search Node): Add metadata transformation for document insertion

aalises:
## Summary                                                                
Fix metadata not being stored when inserting documents into Azure AI Search                                                                    
- Add "Metadata Keys to Insert" option to filter which metadata fields are stored                                                                   
- Transform metadata into Azure AI Search's required `attributes` array format with `{key, value}` pairs     

documentation on the type expected: https://v03.api.js.langchain.com/types/_langchain_community.vectorstores_azure_aisearch.AzureAISearchDocumentMetadata.html

---
**related ticket: https://linear.app/n8n/issue/AI-1920/bug-azure-ai-search-vector-store-node-doesnt-insert-metadata**

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreAzureAISearch/VectorStoreAzureAISearch.node.test.ts
+++ b/packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreAzureAISearch/VectorStoreAzureAISearch.node.test.ts
@@ -12,6 +12,7 @@ import {
 	VectorStoreAzureAISearch,
 	getIndexName,
 	clearAzureSearchIndex,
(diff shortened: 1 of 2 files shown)


mike12345567 (commented):
Looks good - worth addressing the cubic comment on typecasting - this will need to be backported to stable/beta likely, so we will need to label it as such.

Will approve once cubic comment addresed!

aalises:
> Looks good - worth addressing the cubic comment on typecasting - this will need to be backported to stable/beta likely, so we will need to label it as such.
> 
> Will approve once cubic comment addresed!

Fixed!

mike12345567 (approved):

PR #16765, opened 2025-06-27
fix(editor): Make sure connectors are clickable when overlapped by node description

MiloradFilipovic:
## Summary

https://github.com/user-attachments/assets/c42f33a0-bbf2-44fd-a4d0-9df9b1f5e078

In order to make sure that node connectors are clickable, this PR disables pointer events on node descriptions. This was the least-complicated way to tackle the problem I could think of but will prevent users from selecting nodes when clicking descriptions, so we'll also need some product feedback about this.

## Related Linear tickets, Github issues, and Community forum posts
Fixes ADO-3658

## Review / Merge checklist

- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #24631, opened 2026-01-21
fix(editor): Prevent connection lines from overlapping plus handle buttons

sandra0503:
## Summary
The issue was that edge delete button was overlapping input plus handles on Ai Agent, making them difficult to click.

Changes: 
1. Event Blocker
A pseudo-element on the plus handle blocks edge hover detection around the handle to prevent visual overlap.

2. Z-Index Management for Edges
Update edge toolbar zIndex logic to consider bringToFront prop

See my Loom for further explanation:

https://www.loom.com/share/83fa3aee54e64ea99b75124a10cee993

Fixes [LIGO-104](https://linear.app/n8n/issue/LIGO-104/bug-tool-addition-and-tool-connection-buttons-overlapping-on-the): Edge delete buttons overlapping non-main input plus handles, making them hard to click.

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/workflows/canvas/components/elements/handles/render-types/parts/CanvasHandlePlus.vue
```diff
@@ -32,8 +32,7 @@ const emit = defineEmits<{
 const style = useCssModule();
 
 const classes = computed(() => [
-	'canvas-handle-plus-wrapper',
-	style.wrapper,
+	style.container,
```
konstantintieber, line 36: It looks like the `canvas-handle-plus-wrapper` class was added for this fix in the past: https://github.com/n8n-io/n8n/pull/16765

I think it's worth verifying that the new implementation also still has the "plus buttons" working when behind a text description.

Will test this after lunch.

konstantintieber: If we do end up leaving this class out, we should remove the CSS from https://github.com/n8n-io/n8n/blob/master/packages/frontend/editor-ui/src/app/css/plugins/_vueflow.scss#L86 like cubic mentioned.

sandra0503: Hi Konsti, I see that the class is still needed! I would add it back in

konstantintieber: Perfect, thx for checking

packages/frontend/editor-ui/src/features/workflows/canvas/components/elements/handles/render-types/parts/CanvasHandlePlus.vue
```diff
@@ -106,50 +105,67 @@ function onClick(event: MouseEvent) {
...
+			</g>
+		</svg>
+	</div>
 </template>
 
 <style lang="scss" module>
+.container {
+	position: relative;
+
+	&.bottom::before {
```
konstantintieber, line 159: As I was doing another check over the "Files changed" rather than by commits, I noticed myself thinking for a while of what this pseudo element's purpose was again.
With the context of the commit message or the PR description, it's easy to know that this is about:
> event blocker pseudo element that prevents edge hover detection when mouse is over plus handle area

Since it has no semantic class name, I think it'd good to include an inline comment explaining the purpose of this CSS for future readers.
Maybe also explaining that this is about the plus button taking priority over the delete button.

sandra0503: I added a comment, hope it's more clear now!

konstantintieber (approved):
Awesome!

PR #24643, opened 2026-01-21
feat(editor): New operations in the Slack node

nikhilkuria:
## Summary

This PR adds two operations on the Slack node
- Get members from a group
- Add members to a group

https://www.loom.com/share/29cabd18f1c947f7b0ac03d98c8b1478

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
+++ b/packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
@@ -1474,7 +1474,7 @@ export class SlackV2 implements INodeType {
 
 						const options = this.getNodeParameter('options', i);
 
-						const resolveData = options.resolveData !== false;
+						const resolveData = options.resolveData ?? true
 
 						// Get the user group with user list
 						const groupData = await slackApiRequest.call(
--- a/packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
+++ b/packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
@@ -1474,7 +1474,7 @@ export class SlackV2 implements INodeType {
 
 						const options = this.getNodeParameter('options', i);
 
-						const resolveData = options.resolveData ?? true
+						const resolveData = options.resolveData ?? true;
 
 						// Get the user group with user list
 						const groupData = await slackApiRequest.call(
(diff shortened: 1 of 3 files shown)


packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
```diff
@@ -1423,6 +1423,101 @@ export class SlackV2 implements INodeType {
...
+
+						responseData = responseData.usergroup;
+					}
+					//https://api.slack.com/methods/usergroups.list
+					if (operation === 'getUsers') {
+						const userGroupId = this.getNodeParameter('userGroupId', i) as string;
+
+						const options = this.getNodeParameter('options', i);
+
+						const resolveData = options.resolveData !== false;
```
ShireenMissi: suggestion for clarity
```suggestion
						const resolveData = options.resolveData ?? true
```

packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
```diff
@@ -1423,6 +1423,101 @@ export class SlackV2 implements INodeType {
...
+							this,
+							'POST',
+							'/usergroups.users.update',
+							body,
+							qs,
+						);
+
+						responseData = responseData.usergroup;
+					}
+					//https://api.slack.com/methods/usergroups.list
```
ShireenMissi, line 1471: this comment is misleading
```suggestion
```

ShireenMissi (commented):
Thank you for adding this @nikhilkuria it looks good, only left some minor comments

packages/nodes-base/nodes/Slack/V2/SlackV2.node.ts
nikhilkuria: I see a similar pattern on other operations of adding the actual API being called. The `usergroups.list` is used to get the list of users.

ShireenMissi: you are right, we usually add the api endpoint that maps to the operation so I thought this meant to be getUsers not groups

ShireenMissi (approved):
LGTM ✅

PR #24644, opened 2026-01-21
feat(FormTrigger Node): Support ip filtering for the FormTrigger node

mfsiega:
## Summary

Bring the same IP filtering that the Webhook node has into the Form trigger node.

Also adds support for CIDR ranges and IPv6; previously we only supported specific IPv4 addresses.

Verified locally that individual addresses as well as ranges are enforced.

https://community.n8n.io/t/allowlist-trusted-ip-ranges-for-forms-submissions/252770

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
+++ b/packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
@@ -78,7 +78,7 @@ const descriptionV1: INodeTypeDescription = {
 					default: 'Your response has been recorded',
 				},
 				{
-					displayName: 'IP(s) Whitelist',
+					displayName: 'IP(s) Allowlist',
 					name: 'ipWhitelist',
 					type: 'string',
 					placeholder: 'e.g. 127.0.0.1, 192.168.1.0/24',
--- a/packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
+++ b/packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
@@ -77,6 +77,15 @@ const descriptionV1: INodeTypeDescription = {
 					type: 'string',
 					default: 'Your response has been recorded',
 				},
+				{
+					displayName: 'IP(s) Whitelist',
+					name: 'ipWhitelist',
+					type: 'string',
+					placeholder: 'e.g. 127.0.0.1, 192.168.1.0/24',
+					default: '',
+					description:
+						'Comma-separated list of allowed IP addresses or CIDR ranges. Leave empty to allow all IPs.',
+				},
 			],
 		},
 	],
--- a/packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
+++ b/packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
@@ -14,6 +14,7 @@ import {
 	formTitle,
 	formTriggerPanel,
 	webhookPath,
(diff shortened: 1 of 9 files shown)


packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
```diff
@@ -77,6 +77,15 @@ const descriptionV1: INodeTypeDescription = {
...
 				},
+				{
+					displayName: 'IP(s) Allowlist',
+					name: 'ipWhitelist',
+					type: 'string',
+					placeholder: 'e.g. 127.0.0.1, 192.168.1.0/24',
+					default: '',
+					description:
+						'Comma-separated list of allowed IP addresses or CIDR ranges. Leave empty to allow all IPs.',
+				},
```
DawidMyslak: Since both `FormTriggerV1` and `FormTriggerV2` define the same IP allowlist option, consider extracting it to `common.descriptions.ts`.

packages/nodes-base/nodes/Form/v2/FormTriggerV2.node.ts
```diff
@@ -165,6 +165,15 @@ const descriptionV2: INodeTypeDescription = {
 					default: false,
 					description: 'Whether to ignore requests from bots like link previewers and web crawlers',
 				},
+				{
+					displayName: 'IP(s) Allowlist',
+					name: 'ipWhitelist',
```
DawidMyslak: I just want to clarify something. I see we use 2 approaches for naming: `Whitelist` and `Allowlist` and then corresponding helper functions and test files are also affected with this mix.

Not a big deal overall, just wanted to double check if it's worth unifying?

DawidMyslak (dismissed):
Approved with minor comments 👏.

packages/nodes-base/nodes/Form/v1/FormTriggerV1.node.ts
mfsiega: Nice, done.

packages/nodes-base/nodes/Form/v2/FormTriggerV2.node.ts
mfsiega: Indeed I changed it everywhere except the name of this option, so I figured that would break existing workflows.

I guess if I want to change it here as well I could maybe add a node version? I don't have much experience with how we do node versioning.

packages/nodes-base/nodes/Form/v2/FormTriggerV2.node.ts
```diff
@@ -17,6 +17,7 @@ import {
 	formRespondMode,
 	formTitle,
 	formTriggerPanel,
+	ipAllowlist,
```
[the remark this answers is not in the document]
mfsiega: Was this addressed in the latest commit?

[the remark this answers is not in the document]
mfsiega: It seems like in line 142 it's part of the options array?

packages/nodes-base/nodes/Form/v2/FormTriggerV2.node.ts
DawidMyslak: Probably not worth versioning, backward compatibility is more valuable here than naming on our end.

DawidMyslak (approved):
👍

PR #24647, opened 2026-01-21
feat(core): Check if dynamic credentials auth token is set, return 400 otherwise

guillaumejacquart:
## Summary

Force `N8N_DYNAMIC_CREDENTIALS_ENDPOINT_AUTH_TOKEN` environment variable to be set in order to call the dynamic credentials external endpoints for marketplace
Also allow x-authorization header to be sent from the marketplace on dev, so that the static auth token can be sent from a frontend marketplace in dev

https://linear.app/n8n/issue/IAM-132/force-static-authentication

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/middlewares/cors.ts
+++ b/packages/cli/src/middlewares/cors.ts
@@ -8,7 +8,7 @@ export const corsMiddleware: RequestHandler = (req, res, next) => {
 		res.header('Access-Control-Allow-Methods', 'GET, POST, OPTIONS, PUT, PATCH, DELETE');
 		res.header(
 			'Access-Control-Allow-Headers',
-			'Origin, X-Requested-With, Content-Type, Accept, push-ref, browser-id, anonymousid, authorization',
+			'Origin, X-Requested-With, Content-Type, Accept, push-ref, browser-id, anonymousid, authorization, x-authorization',
 		);
 	}
 
--- a/packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
@@ -247,9 +247,11 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 		const { endpointAuthToken } = this.dynamicCredentialConfig;
 		if (!endpointAuthToken?.trim()) {
 			return (_req: Request, res: Response, _next: NextFunction) => {
-				res.status(400).json({
-					message:
-						'You must provide an endpoint auth token to access dynamic credentials external endpoints.',
+				this.logger.error(
+					'Dynamic credentials external endpoints require an endpoint auth token. Please set the N8N_DYNAMIC_CREDENTIALS_ENDPOINT_AUTH_TOKEN environment variable to enable access.',
+				);
+				res.status(500).json({
+					message: 'Dynamic credentials configuration is invalid. Check server logs for details.',
 				});
 				return;
 			};
--- a/packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
@@ -1,6 +1,7 @@
 import { Logger } from '@n8n/backend-common';
 import { CredentialResolverDataNotFoundError } from '@n8n/decorators';
 import { Service } from '@n8n/di';
+import { NextFunction, Request, Response } from 'express';
 import { Cipher } from 'n8n-core';
 import type {
 	ICredentialDataDecryptedObject,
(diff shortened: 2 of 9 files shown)


packages/cli/src/middlewares/cors.ts
```diff
@@ -8,7 +8,7 @@ export const corsMiddleware: RequestHandler = (req, res, next) => {
 		res.header('Access-Control-Allow-Methods', 'GET, POST, OPTIONS, PUT, PATCH, DELETE');
 		res.header(
 			'Access-Control-Allow-Headers',
-			'Origin, X-Requested-With, Content-Type, Accept, push-ref, browser-id, anonymousid, authorization',
+			'Origin, X-Requested-With, Content-Type, Accept, push-ref, browser-id, anonymousid, authorization, x-authorization',
```
phyllis-noester, line 11: ah thanks I forgot about that!

phyllis-noester: actually is that needed? because I thought we only use the x-authorization/M2M token from the BE

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -244,7 +245,16 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 	 */
 	getDynamicCredentialsEndpointsMiddleware() {
 		const { endpointAuthToken } = this.dynamicCredentialConfig;
+		if (!endpointAuthToken?.trim()) {
+			return (_req: Request, res: Response, _next: NextFunction) => {
+				res.status(400).json({
+					message:
+						'You must provide an endpoint auth token to access dynamic credentials external endpoints.',
```
phyllis-noester: would it make sense to specify the name of the env var here so the developers can debug it more easily? or if we do not want to expose the inner workings refer them to some docs?

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -244,7 +245,16 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 	 */
 	getDynamicCredentialsEndpointsMiddleware() {
 		const { endpointAuthToken } = this.dynamicCredentialConfig;
+		if (!endpointAuthToken?.trim()) {
+			return (_req: Request, res: Response, _next: NextFunction) => {
+				res.status(400).json({
```
phyllis-noester: thinking about it again now.. shouldn't this be a 500? because it's a misconfiguration on the server and unrelated to the client no?

phyllis-noester (changes requested):

packages/cli/src/middlewares/cors.ts
guillaumejacquart: this middleware is used for dev only (check on NODE_ENV), so I thought it would make sense to add this allowed header, at least for engineers when we're working on the marketplace example

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
guillaumejacquart: I agree !

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
guillaumejacquart: I did not want to expose indeed.
I added a more explicit error log on the server, and pointed the user to check server logs in the json response. What do you think ?

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -244,7 +245,18 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 	 */
 	getDynamicCredentialsEndpointsMiddleware() {
 		const { endpointAuthToken } = this.dynamicCredentialConfig;
+		if (!endpointAuthToken?.trim()) {
+			return (_req: Request, res: Response, _next: NextFunction) => {
+				this.logger.error(
+					'Dynamic credentials external endpoints require an endpoint auth token. Please set the N8N_DYNAMIC_CREDENTIALS_ENDPOINT_AUTH_TOKEN environment variable to enable access.',
+				);
+				res.status(500).json({
```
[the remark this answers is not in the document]
guillaumejacquart: We switch to a 500 as it's rather a server configuration error (does not depend on the client request)

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
phyllis-noester: makes sense :)

packages/cli/src/middlewares/cors.ts
phyllis-noester: thanks for explaining

phyllis-noester (approved):

PR #24655, opened 2026-01-21
chore: Introduce backport github action

dlavrenuek:
## Summary

Adds a github action to automatically create backport PRs when `backport XXX` (where `XXX` is the target branch) label is added to a PR.

Uses https://github.com/korthout/backport-action. It is possible to run the backport action when a PR is merged or when a label with `/backport XXX` is added after the merge.

To demonstrate how automatic backporting works I have created an example:
- PR https://github.com/dlavrenuek/n8n/pull/2 is merged and includes a `backport 1.x` label
- https://github.com/dlavrenuek/n8n/pull/3 was automatically created

## Review / Merge checklist

- [ ] Tests included. 


.github/workflows/backport.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          pull_description: |-
+            # Description
+            Backport of #${pull_number} to `${target_branch}`.
+
+            # Original description
+
+            ${pull_description}
+          pull_title: ${pull_title} (backport ${target_branch})
+          add_author_as_assignee: true
```
shortstacked, line 31: We should add: 'copy_requested_reviewers' as well for consistency.

dlavrenuek: Added, also added the ability to comment `/backport 1.x` on already merged PRs

.github/workflows/backport.yml
```diff
@@ -0,0 +1,43 @@
+name: Backport PRs
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
dlavrenuek: Afaik running it as pull_request should be correct, since only maintainers can merge PRs, which then runs the action with permissions of the person who merged it and not the one who created the PR

.github/workflows/backport.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+
+    if: >
+      (
+        github.event_name == 'pull_request' &&
+        github.event.pull_request.merged
+      ) || (
+        github.event_name == 'issue_comment' &&
+        github.event.issue.pull_request &&
+        startsWith(github.event.comment.body, '/backport')
```
shortstacked: Anyone can comment and trigger this. We should add something like
  contains(fromJSON('["OWNER", "COLLABORATOR"]'), github.event.comment.author_association)      
to properly reduce scope.                                                                                            
We have a a few other comment triggered actions you could check them.

dlavrenuek: The workflows with `issue_comment` all specify permissions to (at least) read PRs on workflow level (other than `Util: Claude` which specify permissions on job level). Without explicit permissions it should use the permissions of the comment author. May be someone with and without maintainer permissions can test it on https://github.com/dlavrenuek/n8n/pull/2 (for me it triggers the backport)?

dlavrenuek: PS: removed the comments option and added `labeled` to pr triggers. The action checks if the backport branch exists and will not create a [new PR](https://github.com/korthout/backport-action/blob/main/src/backport.ts#L348) if that is the case

.github/workflows/backport.yml
```diff
@@ -0,0 +1,38 @@
...
+            Backport of #${pull_number} to `${target_branch}`.
+
+            # Original description
+
+            ${pull_description}
+          pull_title: ${pull_title} (backport ${target_branch})
+          add_author_as_assignee: true
+          copy_assignees: true
+          copy_requested_reviewers: true
+          copy_labels_pattern: '.*'
```
tomi, line 34: I don't think we want to include the backport label on the new PR, right? Would that cause recursion?

dlavrenuek: > Also I have added label copying (will exclude backport labels)

The backport labels (anything that matches the backport label pattern) are excluded from the copied labels [REF](https://github.com/korthout/backport-action?tab=readme-ov-file#copy_labels_pattern)

tomi: Gotcha 👍 We should also ignore "Released" label as it is used to designate whether a PR has been included in a release already or not (to post comments)

tomi: @dlavrenuek did you see my comment here?

tomi (approved):
One more comment, but we can also do that as a followup

.github/workflows/backport.yml
dlavrenuek: Sounds good. Let's do some testing and incorporate any changes when going live with the new approach. I have created a going live checklist, feel free to add topics there: https://linear.app/n8n/issue/NODE-4263/automatic-pr-backport-going-live-checklist

PR #24657, opened 2026-01-21
fix(core): Make Respond to Chat nodes work on Chat hub with multi-main instances

Cadiac:
## Summary

We didn't consider `waiting` state as a finished state, so execution polling never stopped on multi-main instances.
Stopping like this if workflow enters waiting state isn't maybe ideal either, but Chat hub doesn't really support messages after waits of other type than Respond to Chat nodes.

This PR fixes the bug by simply accepting `waiting` (and `unknown`) as finished states in polling, and  adds tests for multi-main polling.

https://linear.app/n8n/issue/CHA-110/bug-respond-to-chat-nodes-dont-work-on-multi-main-on-chat-hub

## Review / Merge checklist

- [ ] Tests included. 


autologie (approved):
Wondering what's the good word for "nothing will occur unless something is triggered from outside". Halt..? Anyway, good fix 👍

PR #24660, opened 2026-01-21
feat: Dynamic creds workflow fails if no dynamic creds exist

phyllis-noester:
## Summary

GIVEN a workflow uses dynamic credentials, 
WHEN an external user triggers the workflow with invalid auth token,
THEN the execution fails with an error that dynamic creds could not be fetched.

GIVEN a workflow uses dynamic credentials, 
WHEN an external user triggers the workflow with auth token
AND the webhook does not have the Bearer Token Extractor set up correctly,
THEN the execution fails with an error about incorrect execution context.

GIVEN a workflow uses dynamic credentials, 
WHEN an a developer triggers the workflow manually,
THEN the execution succeeds using the static credentials.

https://linear.app/n8n/issue/IAM-134/do-not-fall-back-on-static-credentials-when-users-execute-flow-with

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
```diff
@@ -24,3 +24,5 @@ export class DynamicCredentialsConfig {
 	@Env('N8N_DYNAMIC_CREDENTIALS_ENDPOINT_AUTH_TOKEN')
 	endpointAuthToken: string = '';
 }
+
+export const DYNAMIC_CREDENTIALS_ALLOWED_NODE_TYPES = ['n8n-nodes-base.webhook', 'webhook'];
```
guillaumejacquart: I'm not sure it's the best solution to introduce a hard coded list of trigger nodes that imply allowing to resolve a credential (this list will get longer and longer, and we already have it on the identity extractors)
If the goal is to prevent resolving from manual workflow runs, you could use the `mode` parameter in the `getDecrypted` function in the credential-helpers maybe ?

guillaumejacquart (commented):
Nice work ! one comment

packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.config.ts
phyllis-noester: Hey @guillaumejacquart Andi and me discussed that doing this in a clean way would probably be too time consuming for alpha. I've now implemented that we do not fall back on static creds when there is no resolver defined. Meanwhile Andi is working on failing the workflow if no bearer token is passed along (instead of falling back to static creds)

afitzek (approved):
👍

PR #24687, opened 2026-01-22
ci: Fix security cherrypick workflow

ivov:
Prevent checkout from caching credentials so the cherry-pick workflow uses the App token instead of github-actions[bot] when pushing to the public repo.

https://github.com/n8n-io/n8n-private/actions/runs/21242054933/job/61122104076

### DIFF
--- a/.github/workflows/sec-publish-fix.yml
+++ b/.github/workflows/sec-publish-fix.yml
@@ -38,6 +38,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
+          persist-credentials: false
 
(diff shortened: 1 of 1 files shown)


.github/workflows/sec-publish-fix.yml
```diff
@@ -38,6 +38,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
+          persist-credentials: false
```
shortstacked: I think the issue is you need to generate the app token first, and do the checkout with it. So you checkout/commit as the same user.
Same as https://github.com/n8n-io/n8n/blob/master/.github/workflows/release-create-pr.yml

shortstacked (approved):

PR #24693, opened 2026-01-22
perf(core): Lazyload login strategies and external secrets

ivov:
Let's lazyload heavy libs until their modules can be disabled by default, for 7 MB memory gain today


BGZStephen (approved):
makes sense to me, we have dynamic modules for SSO on the horizon so this will be a good way to tide us over for now

PR #24706, opened 2026-01-22
fix: Validate sql

michael-radency:
## Summary
validate sql

https://linear.app/n8n/issue/NODE-4199

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Salesforce/GenericFunctions.ts
```diff
@@ -179,29 +179,229 @@ export function sortOptions(options: INodePropertyOptions[]): void {
...
+ * Custom metadata types: MyMetadata__mdt
+ * @throws Error if the object name contains invalid characters
+ */
+export function validateSoqlObjectName(objectName: string): string {
+	// Salesforce object names: alphanumeric, underscore
+	// Can end with: __c (custom), __x (external), __e (platform event), __b (big object),
+	// __mdt (custom metadata), __Share (sharing), __History (history), __Feed (feed), __ChangeEvent (change events)
+	// Can have namespace prefix like Namespace__ObjectName__c
+	// Standard objects: Account, Contact, Lead, etc.
+	const validObjectPattern =
```
RomanDavydchuk: Should we support more [suffixes](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_concepts_object_types.htm)? Or alternatively, don't validate that the suffix is correct, just make sure it contains only English letters, similar to what's done in `validateSoqlFieldName`

packages/nodes-base/nodes/Salesforce/GenericFunctions.ts
```diff
@@ -179,29 +179,229 @@ export function sortOptions(options: INodePropertyOptions[]): void {
...
+			}
+			if (typeof v === 'boolean') {
+				return v.toString();
+			}
+			throw new Error('Array values must be strings, numbers, or booleans');
+		});
+		return `(${escapedValues.join(',')})`;
+	}
+
+	if (typeof value === 'string') {
```
RomanDavydchuk, line 342: if it's a string, might be a good idea to `.trim()` it here too

packages/nodes-base/nodes/Salesforce/GenericFunctions.ts
```diff
@@ -179,29 +179,229 @@ export function sortOptions(options: INodePropertyOptions[]): void {
...
+
+		// Check for N_DAYS_AGO, N_WEEKS_AGO, etc. patterns
+		if (
+			/^N_(DAYS|WEEKS|MONTHS|QUARTERS|YEARS|FISCAL_QUARTERS|FISCAL_YEARS)_AGO:\d+$/.test(upperValue)
+		) {
+			return upperValue;
+		}
+
+		// Check for Salesforce datetime format: YYYY-MM-DDTHH:mm:ss(.SSS)?(Z|[+-]HH:mm)
+		if (/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{3})?(Z|[+-]\d{2}:\d{2})?$/.test(value)) {
```
RomanDavydchuk, line 367: I don't think Salesforce's datetime includes milliseconds, but I guess we can leave it just in case, since the group is optional anyway
image

packages/nodes-base/nodes/Salesforce/GenericFunctions.ts
```diff
@@ -220,23 +420,26 @@ export function getDefaultFields(sobject: string) {
...
 export function getQuery(options: IDataObject, sobject: string, returnAll: boolean, limit = 0) {
+	const validSobject = validateSoqlObjectName(sobject);
+
 	const fields: string[] = [];
 	if (options.fields) {
 		// options.fields is comma separated in standard Salesforce objects and array in custom Salesforce objects -- handle both cases
 		if (typeof options.fields === 'string') {
-			fields.push.apply(fields, options.fields.split(','));
+			const fieldList = options.fields.split(',').map((f) => f.trim());
+			fields.push(...fieldList.map((f) => validateSoqlFieldName(f)));
```
RomanDavydchuk: nit: I think we prefer `arr.push.apply(arr, items)` over `arr.push(...items)`
image
shouldn't be an issue here since the number of items should be small, but thought I'd bring this up. same in other occurrences below

packages/nodes-base/nodes/Salesforce/Salesforce.node.ts
```diff
@@ -389,6 +390,9 @@ export class Salesforce implements INodeType {
 				if (resource === 'customObject') {
 					resource = this.getNodeParameter('customObject', 0) as string;
 				}
+
+				resource = escapeSoqlString(resource as string);
```
RomanDavydchuk, line 394: Should we also use `validateSoqlObjectName` here?

packages/nodes-base/nodes/Salesforce/GenericFunctions.ts
```diff
@@ -179,29 +179,229 @@ export function sortOptions(options: INodePropertyOptions[]): void {
...
+	}
+
+	if (typeof value === 'boolean') {
 		return value;
-	} else if (typeof value === 'string') {
-		return `'${value}'`;
-	} else {
+	}
+
+	if (typeof value === 'number') {
```
RomanDavydchuk, line 317: I think that the `value` is always `string` (for conditions). At least it does not work with numbers:
image
as you can see `0` is inside of single quotes (`'0'`)

RomanDavydchuk (changes requested):
The core logic looks good and works, left a couple comments with possible improvements. But, seems like the logic for handling different data types when creating conditions is broken, at least for numbers (see the last comment in code)

packages/nodes-base/nodes/Salesforce/Salesforce.node.ts
michael-radency: maybe lets stick to `escapeSoqlString`, `validateSoqlObjectName` is more strict and could be breaking

packages/nodes-base/nodes/Salesforce/GenericFunctions.ts
michael-radency: added new node version to not break old behavior

RomanDavydchuk: Still broken for old version. I think here's what happens:
- the `Value` is always returned as a string from the node parameters
- in he old version, we weren't escaping string, which is why it was interpolated just fine (e.g. `WHERE AnnualRevenue > 0`)
- the automatic conversion from string to number only happens for node version > 1.1, so the value stays as string and gets escaped in quotes

So I think it should be in reverse: for version 1, if string "looks like a number" - convert to a number, for 1.1 - treat as string. Or alternatively, always try to convert to number if it looks like a number, so no need for version 1.1?

RomanDavydchuk (changes requested):
Left a comment regarding number handling, still broken for old node version ☹️ 

Also I think some tests are broken, looking at the CI

RomanDavydchuk (approved):
LGTM

PR #24713, opened 2026-01-22
fix(Git Node): Clean up URLs returned from config

RomanDavydchuk:
## Summary

Clean up URLs returned by the List Config operation, handle the case when there are multiple URLs for the same config key

https://linear.app/n8n/issue/NODE-4105

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Git/GenericFunctions.ts
```diff
@@ -40,3 +41,45 @@ export function validateGitReference(reference: string, node: INode): void {
...
+function sanitizeUrl(url: string): string {
+	const urlObj = new URL(url);
+	urlObj.username = '';
+	urlObj.password = '';
+	return urlObj.toString();
+}
+
+export function mapGitConfigList(config: ConfigListSummary) {
+	const data = [];
+	for (const fileName of Object.keys(config.values)) {
```
yehorkardash, line 58: should we have opt-in env variable that disables this feature?

RomanDavydchuk: I'm not sure I can think of a reason why some users might want to have this. if they want to get some of the URLs from config and use it in other Git nodes, they can just get the URL and use the same Git credentials. let me know if I'm missing something or if you have a different opinion

yehorkardash: Ah, true, forgot about credentials. I was thinking about cloning repositories with provided credentials from git config

yehorkardash (approved):

PR #24718, opened 2026-01-22
chore: Add secrets provider api contracts

ireneea:
## Summary

Add API types and DTOs for secret provider connections in @n8n/api-types

closes LIGO-113

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/api-types/src/dto/secrets-provider/create-secret-provider-connection.dto.ts
```diff
@@ -0,0 +1,14 @@
+import type { IDataObject } from 'n8n-workflow';
+import { z } from 'zod';
+import { Z } from 'zod-class';
+
+import { secretsProviderTypeSchema } from '../../schemas/secrets-provider.schema';
+
+export class CreateSecretProviderConnectionDto extends Z.class({
+	name: z.string().min(1).max(128),
```
konstantintieber: Do we stick to `name` here or should we call it `id` to reflect its uniqueness?

packages/@n8n/api-types/src/dto/secrets-provider/create-secret-provider-connection.dto.ts
```diff
@@ -0,0 +1,14 @@
...
+	type: secretsProviderTypeSchema,
+	displayName: z.string().min(1).max(128).optional(),
+	isGlobal: z.boolean(),
+	projectId: z.string().min(1),
```
konstantintieber: How about we omit `isGlobal` altogether, but have an optional `projectId` parameter.

If `projectId` is set -> creating project scoped provider
if not -> global provider

konstantintieber: With the schema presented here I don't understand the `projectId: z.string().min(1)` in combination with `isGlobal: true`, because `projectId` is in this case not needed at all.

packages/@n8n/api-types/src/dto/secrets-provider/update-secret-provider-connection.dto.ts
```diff
@@ -0,0 +1,12 @@
+import type { IDataObject } from 'n8n-workflow';
+import { z } from 'zod';
+import { Z } from 'zod-class';
+
+import { secretsProviderTypeSchema } from '../../schemas/secrets-provider.schema';
+
+export class UpdateSecretProviderConnectionDto extends Z.class({
+	type: secretsProviderTypeSchema.optional(),
+	displayName: z.string().min(1).max(128).optional(),
+	isGlobal: z.boolean().optional(),
```
konstantintieber: Wondering about `isGlobal`: Do we intend to support changing a global provider to a non-global one? Or what is its intention?

I think from a migration perspective (users need to manually migrate workflows that use this global credential but these workflows are in multiple projects), I expect that we should not allow changing an existing global provider to a project-specific one.
Users would rather need to use the same token of the external provider to create a project-specific one, then they can migrate existing workflows gradually without them breaking right away.

Making a project-specific provider global seems fine, but I'm also not sure if we'd want to support that.

packages/@n8n/api-types/src/dto/secrets-provider/create-secret-provider-connection.dto.ts
```diff
@@ -0,0 +1,14 @@
+import type { IDataObject } from 'n8n-workflow';
+import { z } from 'zod';
+import { Z } from 'zod-class';
+
+import { secretsProviderTypeSchema } from '../../schemas/secrets-provider.schema';
+
+export class CreateSecretProviderConnectionDto extends Z.class({
+	name: z.string().min(1).max(128),
+	type: secretsProviderTypeSchema,
+	displayName: z.string().min(1).max(128).optional(),
```
konstantintieber: Should we omit the `displayName` altogether for now, since we don't know how to incorporate it into our expression editor yet?

packages/@n8n/api-types/src/dto/secrets-provider/create-secret-provider-connection.dto.ts
geemanjs: We should probably make this an array just to keep things simple later on

packages/@n8n/api-types/src/dto/secrets-provider/transfer-secret-provider-connection.dto.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod';
+import { Z } from 'zod-class';
+
+/**
+ * Transfer connection to different project.
+ */
+export class TransferSecretProviderConnectionDto extends Z.class({
```
geemanjs: I'm unsure about "transfer" as a concept here. I believe it is here to match how credentials work - but I don't think we should copy that behaviour 👀. 

I feel like your secret provider connection is shared with one or more projects. and transfer is "sharing" / "unsharing" it with a project

packages/@n8n/api-types/src/dto/secrets-provider/create-secret-provider-connection.dto.ts
konstantintieber: Fun fact: the hardcoded providers we have in classes also already have the concept of `displayName`: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/modules/external-secrets.ee/providers/gcp-secrets-manager/gcp-secrets-manager.ts#L17

Even though it seems right now it's only used in logs. The frontend never renders it.

packages/@n8n/api-types/src/dto/secrets-provider/create-secret-provider-connection.dto.ts
ireneea: See PR #24551, the link between providers and projects has been drastically simplified 👍

ireneea: See PR #24551, the link between providers and projects has been drastically simplified 👍

packages/@n8n/api-types/src/schemas/secrets-provider.schema.ts
```diff
@@ -0,0 +1,112 @@
...
+
+// ==========
+// #endregion
+
+// #region RESPONSE TYPES
+// ======================
+
+/**
+ * Full connection detail
+ * - secrets is optional (not returned in list views)
```
konstantintieber, line 65: Rather than making secrets optional in this type: How about we make two schemas:
`SecretProviderConnectionListItem` and `SecretProviderConnection`, where `SecretProviderConnection` is:

```ts
// secretProviderConnectionListItemSchema defined above here
export type SecretProviderConnectionListItem = z.infer<typeof secretProviderConnectionListItemSchema>;
export const secretProviderConnectionSchema = secretProviderConnectionListItemSchema.extend({
  secrets: z.array(secretSummarySchema)
});
export type SecretProviderConnection = z.infer<typeof secretProviderConnectionSchema>;
```

Or we go for a `secretProviderConnectionCommonSchema` and have `...ConnectionListItem` and `...Connection` each extend that one.

Like this, we'll benefit from accurate types matching the backend responses.

packages/@n8n/api-types/src/dto/secrets-provider/create-secrets-provider-connection.dto.ts
```diff
@@ -0,0 +1,12 @@
+import type { IDataObject } from 'n8n-workflow';
+import { z } from 'zod';
+import { Z } from 'zod-class';
+
+import { secretsProviderTypeSchema } from '../../schemas/secrets-provider.schema';
+
+export class CreateSecretsProviderConnectionDto extends Z.class({
+	providerKey: z.string().min(1).max(128),
```
geemanjs, line 8: Should we lock these to a regex? 
lower case only would be: `^[a-z0-9]+(?:-[a-z0-9]+)*$` 
including upper case would be: `^[A-Za-z0-9]+(?:-[A-Za-z0-9]+)*$`

geemanjs (approved):

packages/@n8n/api-types/src/schemas/secrets-provider.schema.ts
ireneea: Added this as a comment on LIGO-114

PR #24730, opened 2026-01-22
feat(core): Add new landing page on Chat hub for first time users

Cadiac:
## Summary

This PR adds a new landing page for (non-chat only) users landing on Chat hub with empty conversation history.

Once you have some messages in your history this page is no longer shown, and the intent is to better explain what this service can do.

Invite button is shown if users can invite users, and it's enabled if license is such that advanced permissions are allowed.

Partially covers https://linear.app/n8n/issue/CHA-109/feature-ui-for-users-landing-on-chat-hub-for-the-first-time,
continuing to adding automatic credentials next.

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatStarter.vue
```diff
@@ -1,28 +1,176 @@
...
+	void goToUpgrade('chat-hub', 'upgrade-advanced-permissions');
+}
 </script>
 
 <template>
 	<div :class="[$style.starter, { [$style.isMobileDevice]: isMobileDevice }]">
-		<N8nHeading tag="h2" bold size="xlarge">
-			{{ greetings }}
-		</N8nHeading>
+		<Transition name="welcome-fade" mode="out-in">
```
autologie: I like using fade effect here, but transition could be cleaner (ChatPrompt is pushed down while transitioning)

https://github.com/user-attachments/assets/52bcbe96-cd79-480e-8eb8-7492c73c2bdf

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
```diff
@@ -87,6 +95,14 @@ const canSelectTools = computed(
 		!!selectedModel.value?.metadata.capabilities.functionCalling,
 );
 
+const showWelcomeScreen = computed(
```
autologie: It would be nice to avoid showing normal chat UI momentarily while waiting for sessions API response.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatStarter.vue
```diff
@@ -1,28 +1,176 @@
...
+					<div
+						:class="[$style.cardWrapper, $style.cardLast]"
+						data-test-id="welcome-card-base-models"
+					>
+						<N8nCard :class="$style.card">
+							<div :class="$style.cardHeader">
+								<div :class="$style.providerIcons">
+									<CredentialIcon credential-type-name="openAiApi" :size="20" />
+									<CredentialIcon credential-type-name="anthropicApi" :size="20" />
+									<CredentialIcon credential-type-name="googlePalmApi" :size="20" />
```
autologie: Looks like these logo images are slightly larger than icons on other panels

packages/frontend/editor-ui/src/app/composables/useFreeAiCredits.ts
```diff
@@ -0,0 +1,81 @@
...
+
+			telemetry.track('User claimed OpenAI credits', { source });
+
+			return true;
+		} catch (e) {
+			toast.showError(
+				e,
+				i18n.baseText('freeAi.credits.showError.claim.title'),
+				i18n.baseText('freeAi.credits.showError.claim.message'),
+			);
```
autologie, line 67: Personally seeing this without any action feels a bit strange

packages/frontend/editor-ui/src/app/composables/useFreeAiCredits.ts
```diff
@@ -0,0 +1,81 @@
...
+		claimingCredits.value = true;
+
+		try {
+			await credentialsStore.claimFreeAiCredits(projectsStore.currentProject?.id);
+
+			if (usersStore?.currentUser?.settings) {
+				usersStore.currentUser.settings.userClaimedAiCredits = true;
+			}
+
+			telemetry.track('User claimed OpenAI credits', { source });
```
autologie, line 59: This is after await, so if we're getting the telemetry, auto claim is actually working 👍

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
Cadiac: it would be, but I couldn't figure out a way that wouldn't slow down displaying the chat prompt normally - we'd need to get the sessions to know whether user has history or not 🤔 
can you think of a better way?

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatStarter.vue
Cadiac: This was now done according to the figma design, but I could match what we have elsewhere instead?

packages/frontend/editor-ui/src/app/composables/useFreeAiCredits.ts
Cadiac: how do you see this? We shouldn't try claiming if it wasn't possible? The fade was here mainly to make the "briefly seen prompt" -> "starter" -> prompt again slightly less ugly as I couldn't figure out how to fully hide the prompt initially if we don't want to slow down showing it normally

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatStarter.vue
Cadiac: Mind committing that if you have it? that does look better

autologie: The recording is unmodified your branch, and I'm not fan of the transition period where both cards and chat prompt are shown. I pushed an attempt to fix https://github.com/n8n-io/n8n/commit/fd932e59c1097a0cc66aa357edd79bc3f9b74e98

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
autologie: A bit more complex, but tried to fix this too in https://github.com/n8n-io/n8n/pull/24730#discussion_r2722964667

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatStarter.vue
autologie: if it's intentional 👍

packages/frontend/editor-ui/src/app/composables/useFreeAiCredits.ts
autologie: I modified `isAiCreditsEnabled` to `true` as a quick way of testing, so encountering the error itself was expected.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatStarter.vue
Cadiac: Keeping track of `LOCAL_STORAGE_CHAT_HUB_HAD_CONVERSATION_BEFORE` is a nice way to avoid that ugly state, I'm cherry-picking that commit into this PR

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
```diff
@@ -86,6 +96,25 @@ const canSelectTools = computed(
 		selectedModel.value?.model.provider === 'custom-agent' ||
 		!!selectedModel.value?.metadata.capabilities.functionCalling,
 );
+const hadConversationBefore = useLocalStorage(
```
Cadiac, line 99: This now stays true even if user deletes their last conversation, so the welcome screen isn't shown again in that case.

but seems like an empty state and just seeing the prompt in that case doesn't seem that wrong, I don't think this is worth handling?

autologie (dismissed):
👍

autologie (approved):

PR #24733, opened 2026-01-22
fix(editor): Add support for dynamic credentials and credential resolver indicators in UI

cstuncsik:
## Summary

Improve dynamic credentials and credential resolver visibility and discoverability

https://linear.app/n8n/issue/IAM-17/show-workflow-as-dynamic-in-the-ui-listings-canvas-and-ndv

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/build.log
```diff
@@ -1,4 +1,8 @@
 
-> n8n@1.121.0 build /Users/mutasem/repos/n8n-worktree/global-creds/packages/cli
+> n8n@2.5.0 build /Users/csaba/conductor/workspaces/n8n/missoula/packages/cli
 > tsc -p tsconfig.build.json && tsc-alias -p tsconfig.build.json && pnpm run build:data
 
+
```
phyllis-noester: why do we commit this?

packages/frontend/editor-ui/src/app/components/WorkflowCard.vue
```diff
@@ -288,6 +291,22 @@ const isWorkflowPublished = computed(() => {
...
+
+const hasDynamicCredentials = computed(() => {
+	return isDynamicCredentialsEnabled.value && props.data.hasResolvableCredentials;
+});
+
+const isResolverMissing = computed(() => {
+	return (
+		isDynamicCredentialsEnabled.value &&
+		props.data.hasResolvableCredentials &&
+		!props.data.settings?.credentialResolverId
```
phyllis-noester, line 306: is the BE afaik we can either have a resolver on the credential or on the workflow level - does this cover both?

packages/cli/src/workflows/workflow.service.ee.ts
```diff
@@ -271,6 +271,64 @@ export class EnterpriseWorkflowService {
...
+		// Extract all credential IDs from all workflows
+		const credentialIdToWorkflowIds = new Map<string, string[]>();
+		for (const workflow of workflows) {
+			if (!workflow.nodes) continue;
+			for (const node of workflow.nodes) {
+				if (!node.credentials) continue;
+				for (const credentialType of Object.keys(node.credentials)) {
+					const credentialId = node.credentials[credentialType]?.id;
+					if (credentialId) {
+						const workflowIdsList = credentialIdToWorkflowIds.get(credentialId) ?? [];
```
phyllis-noester, line 297: granted i haven't looked at the data structure; just wondering: is this the most efficient way to get all workflows with dynamic creds?

One alternative could be: we start by fetching all dynamic credentials the user has access to and then try to map them back to the workflow

packages/frontend/editor-ui/src/app/components/WorkflowCard.vue
cstuncsik: I thought you can set a resolver only for the workflow in the workflow settings.
You can set a credential to dynamic in the credential config.

The workflow then resolves the credential with the resolver so it's workflow only property.

So a workflow can have dynamic credentials and a credential resolver and having only one of the set is still possible

packages/cli/src/workflows/workflow.service.ee.ts
cstuncsik: This code is AI-generated but I hoped it's doing its best.
I asked it your question and here is the reply:

> The current approach is more efficient for paginated lists because:
> 
>  - It's scoped to only the workflows being displayed (typically 10-20)
>  - Credential IDs are extracted in-memory from the already-fetched workflow nodes
>  - The credential query uses an indexed id column with IN clause
>
> The alternative would require fetching all dynamic credentials globally, then mapping them back to workflows which would need either a full table scan or complex JOINs on the JSON nodes field.

phyllis-noester: ah true it's paginated ofc :) thanks!

packages/cli/build.log
cstuncsik: I think Mutasem also pushed it accidentally
Deleted

packages/frontend/editor-ui/src/features/workflows/canvas/components/elements/nodes/render-types/parts/CanvasNodeSettingsIcons.vue
```diff
@@ -84,6 +115,29 @@ const size = 'medium';
 				<N8nIcon icon="continue-on-error" :size="size" />
 			</div>
 		</N8nTooltip>
+
+		<N8nTooltip v-if="isDynamicCredentialsEnabled && hasDynamicCredentials">
```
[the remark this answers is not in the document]
cstuncsik: Added tests in commit https://github.com/n8n-io/n8n/pull/24733/commits/cdf2e5837fcf9d2a0c27439748b92012b8af9a68

packages/frontend/editor-ui/src/app/components/WorkflowCard.vue
afitzek: Phyllis is correct, but in the UI one cannot set a resolver on a credential currently, we skipped this part for now, since we are not sure that it adds value and it makes the use of the dynamic credential system much more complex for customers.

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -124,6 +130,33 @@ const selected = computed<Record<string, INodeCredentialsDetails>>(
 	() => props.node.credentials ?? {},
 );
 
+const isDynamicCredentialsEnabled = computed(() =>
+	checkEnvFeatureFlag.value('DYNAMIC_CREDENTIALS'),
```
afitzek, line 134: Just in general, it feels like we should have a global file with constants for feature flags. 🤔

cstuncsik: I'm not sure if I understand
I'm not sure if these are worth storing and changing continuously in constants.
These are temporary, not long-lived stuff.

afitzek (approved):

PR #24750, opened 2026-01-23
feat: Add telemetry for dynamic credential changes

BGZStephen:
## Summary

This PR adds telemetry for the following:

* When a workflow changes (create, update), tracks the resolverId, and says whether the identity extractor changed
* When a credential is created / updated, whether it was configured to be dynamic
* When an execution happens, did it use dynamic credentials

IAM-74

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/events/relays/telemetry.event-relay.ts
+++ b/packages/cli/src/events/relays/telemetry.event-relay.ts
@@ -1295,7 +1295,8 @@ export class TelemetryEventRelay extends EventRelay {
 	 */
 	private hasIdentityExtractors(nodes: INode[]): boolean {
 		for (const node of nodes) {
-			if (this.extractIdentityExtractorHooks(node).length > 0) {
+			const hooks = this.extractIdentityExtractorHooks(node);
+			if (hooks.length > 0) {
 				return true;
 			}
 		}
@@ -1303,9 +1304,13 @@ export class TelemetryEventRelay extends EventRelay {
 	}
 
 	/**
-	 * Extracts identity extractor hook names from a node's parameters.
+	 * Extracts identity extractor hook configurations from a node's parameters.
+	 * Returns the full hook objects including all parameters (hookName, isAllowedToFail, etc.)
+	 * to enable detection of any configuration changes.
 	 */
-	private extractIdentityExtractorHooks(node: INode): string[] {
+	private extractIdentityExtractorHooks(
+		node: INode,
+	): Array<{ hookName: string; isAllowedToFail?: boolean; [key: string]: unknown }> {
 		const nodeParams = {
 			...(node.parameters ?? {}),
 		};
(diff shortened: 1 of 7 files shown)


packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -1200,6 +1232,94 @@ export class TelemetryEventRelay extends EventRelay {
...
+			const currentHooks = this.extractIdentityExtractorHooks(currentNode);
+			const previousHooks = this.extractIdentityExtractorHooks(previousNode);
+
+			if (JSON.stringify(currentHooks) !== JSON.stringify(previousHooks)) {
+				return true;
+			}
+		}
+
+		// Check if any nodes were removed that had identity extractors
+		const currentNodesMap = new Map<string, INode>();
```
guillaumejacquart, line 1276: I think a set of current nodes name is enough here, no ?

packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -1200,6 +1232,94 @@ export class TelemetryEventRelay extends EventRelay {
...
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	 * Checks if any nodes have identity extractors configured.
+	 */
+	private hasIdentityExtractors(nodes: INode[]): boolean {
```
guillaumejacquart, line 1296: Maybe a more concise version could be:
````
private hasIdentityExtractors(nodes: INode[]): boolean {
    return nodes.some(node => this.extractIdentityExtractorHooks(node).length > 0);
}

File: packages/cli/src/events/relays/telemetry.event-relay.ts [hunk 3]
```diff
@@ -1200,6 +1232,94 @@ export class TelemetryEventRelay extends EventRelay {
...
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * Extracts identity extractor hook names from a node's parameters.
+	 */
+	private extractIdentityExtractorHooks(node: INode): string[] {
+		const nodeParams = {
```
guillaumejacquart: Why the spread here ? I think toExecutionContextEstablishmentHookParameter is a pure function, no ?

guillaumejacquart (dismissed):
nit comments

packages/cli/src/events/relays/telemetry.event-relay.ts
BGZStephen: Yeah, set or map, either way works ok

packages/cli/src/events/relays/telemetry.event-relay.ts
BGZStephen: You're right, I assumed it was because of the potential nullness of node, but it should be safe without spread with defaults

guillaumejacquart (approved):

PR #24751, opened 2026-01-23
fix(editor): Cleanup data table add column popover error state on open (no-changelog)

scdekov:

https://linear.app/n8n/issue/ADO-4702/losing-connection-whilst-adding-data-table-column-cant-be-cleared

## Review / Merge checklist

- [ ] Tests included. 


MiloradFilipovic (approved):
Good to go with this fix from my end but wondering how hard would it be to detect connection issues like these and don't disable the button in that case?

PR #24761, opened 2026-01-23
fix(editor): Show publish actions on read-only instances

dariacodes:
## Summary

There was a bug with not showing publish actions (publish button & unpublish action) for read-only instances. This PR fixes that and refactors the read-only props to reduce confusion.

Closes [ADO-4703](https://linear.app/n8n/issue/ADO-4703/bug-publish-button-is-hidden-on-read-only-instances)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
@@ -21,9 +21,9 @@ import {
 import { nodeViewEventBus } from '@/app/event-bus';
 import CollaborationPane from '@/features/collaboration/collaboration/components/CollaborationPane.vue';
 import TimeAgo from '../TimeAgo.vue';
+import { useCollaborationStore } from '@/features/collaboration/collaboration/collaboration.store';
 
 const props = defineProps<{
-	readOnly?: boolean;
 	id: IWorkflowDb['id'];
 	tags: IWorkflowDb['tags'];
 	name: IWorkflowDb['name'];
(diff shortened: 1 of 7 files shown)


packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -102,6 +89,20 @@ const workflowPublishState = computed((): WorkflowPublishState => {
 	return hasChanges ? 'published-with-changes' : 'published-no-changes';
 });
 
+const collaborationReadOnly = computed(() => collaborationStore.shouldBeReadOnly);
+const hasUpdatePermission = computed(() => props.workflowPermissions.update);
+const hasPublishPermission = computed(() => props.workflowPermissions.publish);
+
+const shouldHidePublishButton = computed(() => {
+	if (props.isNewWorkflow) return false;
+	return props.isArchived || !hasUpdatePermission.value;
```
scdekov: I'm not sure if it's possible to have publish but not update permissions but if it is we should check that both are not present there when hiding the button

scdekov (dismissed):

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: Yeah, good point. I have no idea what the initial thinking was here, but I guess it might make sense to give some users publish permissions without update permissions, so let me fix that

scdekov: Yep, I just checked and for enterprise we allow custom roles that have publish but no edit permissions.
image

scdekov (approved):

PR #24776, opened 2026-01-23
feat(Kafka Trigger Node): Improvements

michael-radency:
## Summary

**Kafka Trigger node improvements (v1.2):**

- **Automatic timeout configuration**: Session Timeout now automatically matches Workflow Timeout (+5s buffer); Heartbeat Interval is set to 1/3 of Session Timeout
- **Execution timeout handling**: Added timeout promise that races against workflow execution to properly handle long-running workflows and prevent hanging consumers
- **New "Resolve Offset" option**: Control when offsets are committed:
  - On Execution Completion - resolve after execution regardless of status
  - On Execution Success - resolve only on success
  - On Allowed Execution Statuses - resolve only on selected execution statuses
  - Immediately - resolve right after message received
- **New "Disable Auto Commit" option**: When enabled (available for non-immediate resolve modes), disables automatic offset commits and hides autoCommitThreshold/autoCommitInterval options
- **Rebalance protection**: Added `isRunning()` and `isStale()` checks to stop batch processing when partition is revoked or consumer stops
- **Notice in the node's UI** that explains Session Timeout and Heartbeat Interval handling and prompts user to set Workflow Timeout option
- **Test(manual) executions**: Fixed issue when in manual mode and if execution waits for done promise it will run indefinitely. In manual mode execution will always run not waiting for done promise, updated parameters description explaining this behaviour
- **Code refactoring**: Extracted utilities to utils.ts for better maintainability

https://linear.app/n8n/issue/NODE-4238/kafka-enable-more-control-over-commit-timing-improve-ui-timeout

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Kafka/KafkaTrigger.node.ts
```diff
@@ -486,6 +350,12 @@ export class KafkaTrigger implements INodeType {
 			await startConsumer();
 			return { closeFunction };
 		} else {
+			if (nodeVersion >= 1.2) {
+				throw new NodeOperationError(this.getNode(), 'Manual testing mode is not supported', {
```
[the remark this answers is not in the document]
michael-radency: added

packages/cli/src/workflow-runner.ts
```diff
@@ -338,7 +338,7 @@ export class WorkflowRunner {
 					if (workflowExecution.isCanceled) {
 						fullRunData.finished = false;
 					}
-					fullRunData.status = this.activeExecutions.getStatus(executionId);
```
mfsiega, line 343: It might be good to put this in its own PR? Bit of a hassle, but IMO it's a bit weird to lump everything in one place.

michael-radency: thanks @mfsiega , created https://github.com/n8n-io/n8n/pull/24897, added you as a reviewer, since you already have some context

yehorkardash:
@claude review

packages/nodes-base/nodes/Kafka/test/KafkaTrigger.node.test.ts
```diff
@@ -139,6 +162,7 @@ describe('KafkaTrigger Node', () => {
 		const { close, emit } = await testTriggerNode(KafkaTrigger, {
 			mode: 'trigger',
 			node: {
+				typeVersion: 1,
```
yehorkardash, line 165: should we run unit tests for each version we have?

mfsiega (commented):
I've started going through this, it will take some more time.

High-level: in general, I would prefer to go with smaller PRs that each have a single focused change, if it's possible. There are a few unrelated things happening in this PR (different offset resolution, plus the session timeout stuff, maybe other stuff?) and it's quite difficult to meaningfully review.

So if it's possible to break this down into a few smaller PRs, that would make things go faster 🙏 and less likely we'll break something.

michael-radency:
> Manual execution doesn't seem to be working in 1.2, but works in 1.1
> 2026-01-28.16-58-20.mp4

Thanks @yehorkardash 🙏

I finally managed to reproduce this issue. It can happen in two cases:

If a workflow is published and you try to run a manual (test) execution of the same workflow.

If a Kafka-triggered workflow is published and you try to publish another Kafka-triggered workflow with the same group ID, or manually run a workflow with the same group ID.

The underlying problem is the same as what we had with conflicting webhook paths where one workflow ends up shadowing the other.

I’m working on a fix.

PR #24779, opened 2026-01-23
chore(core): Adopt UI/UX for user identifier and fix parsing for hook definitions

afitzek:
## Summary

Small adoptions to UI/UX for user identifier and fix parsing for establishment hooks.

image
image

closes https://linear.app/n8n/issue/IAM-110/update-ui-and-copy-for-configuring-identity-extract-in-webhook
closes https://linear.app/n8n/issue/IAM-141/fix-hook-options-parsing-failure-when-hooks-are-not-configured

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/load-nodes-and-credentials.ts
+++ b/packages/cli/src/load-nodes-and-credentials.ts
@@ -353,9 +353,10 @@ export class LoadNodesAndCredentials {
 		// Each hook becomes a separate item that can be added multiple times
 		const allHookValues: INodeProperties[] = [
 			{
-				displayName: 'Hook',
+				displayName: 'User Identifier',
 				name: 'hookName',
 				type: 'options',
+				noDataExpression: true,
 				options: hooks.map((hook) => {
 					const displayName = hook.hookDescription.displayName ?? hook.hookDescription.name;
 					return {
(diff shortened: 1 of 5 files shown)


guillaumejacquart (approved):
LGTM !
Quick question: why no data expression for the header extraction options ?

packages/cli/src/load-nodes-and-credentials.ts
```diff
@@ -367,7 +368,8 @@ export class LoadNodesAndCredentials {
 				// No default - force user to explicitly select a hook
 				// This ensures hookName is always serialized in the workflow JSON
 				default: '',
-				description: 'Select which context establishment hook to use',
+				description:
+					'Configure how n8n extracts the identity token of the user triggering a webhook. It is used to run each execution with the correct user.',
```
phyllis-noester, line 372: one question: do we expect ppl to provide an access token or id token here? Given that we extract it from authorization  header i'd assume access token .... but we keep calling it identity token which i find misleading @afitzek

afitzek: The exact nature of the token depends on the resolver and what the used resolver can handle.

phyllis-noester: makes sense; as user I would just find it confusing that we use language that is very close to id token

PR #24783, opened 2026-01-23
feat(core): Add $tool.name and $tool.parameters expressions

yehorkardash:
## Summary

This PR adds a new shortcut for `fromAI('toolParameters')`
It also removes stringification of `toolParameters` when it's passed to tool. Previously it was done for simpler access, because by default object get converted to `[object Object]` and requires use of JSON.stringify
image

https://linear.app/n8n/issue/NODE-4235/add-dollartoolname-and-dollartoolparameters-expressions-and-update-the

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
@@ -22,7 +22,7 @@ import {
 	CHAT_TRIGGER_NODE_TYPE,
 	createEmptyRunExecutionData,
 	FORM_TRIGGER_NODE_TYPE,
-	isToolType,
+	isHitlToolType,
 	NodeConnectionTypes,
 	NodeHelpers,
 	WEBHOOK_NODE_TYPE,
@@ -125,7 +125,7 @@ function resolveParameterImpl<T = IDataObject>(
 			resumeFormUrl: PLACEHOLDER_FILLED_AT_EXECUTION_TIME,
 		},
 		$vars: envVars,
-		$tool: isToolType(activeNode?.type)
+		$tool: isHitlToolType(activeNode?.type)
 			? {
 					name: PLACEHOLDER_FILLED_AT_EXECUTION_TIME,
 					parameters: PLACEHOLDER_FILLED_AT_EXECUTION_TIME,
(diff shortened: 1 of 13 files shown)


packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
```diff
@@ -125,7 +125,12 @@ function resolveParameterImpl<T = IDataObject>(
 			resumeFormUrl: PLACEHOLDER_FILLED_AT_EXECUTION_TIME,
 		},
 		$vars: envVars,
-		$tool: isToolType(activeNode?.type) ? PLACEHOLDER_FILLED_AT_EXECUTION_TIME : undefined,
+		$tool: isToolType(activeNode?.type)
```
RomanDavydchuk: Since this is available only in HITL tools, shouldn't we check for that specifically, instead of all tools? Currently `$tool` is available in regular tools too:
image

packages/workflow/src/workflow-data-proxy.ts
```diff
@@ -1079,7 +1079,12 @@ export class WorkflowDataProxy {
 		const handleTool = (throwOnError = true) => {
 			const fallbackValue = that.additionalKeys?.['$tool'];
 			try {
-				return handleFromAi('tool', '', 'string', fallbackValue);
+				const toolName = handleFromAi('tool', '', 'string');
+				const toolParameters = handleFromAi('toolParameters', '', 'string');
```
RomanDavydchuk: should this still be of type `string` (third arg) since it's not stringified anymore? not sure how critical this is though

RomanDavydchuk (changes requested):
Code looks good mostly and works as expected. Left a couple questions in code before I can approve this

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
yehorkardash: Good catch, changed the condition

packages/workflow/src/workflow-data-proxy.ts
yehorkardash: The third parameter isn't used in handleFromAI and is 'string' by default. Removed this parameter in these function calls

RomanDavydchuk (approved):
LGTM

PR #24786, opened 2026-01-23
ci: Allow runner fallback from global variable

shortstacked:
## Summary

Allows a fallback of our current CI jobs to Github runners using a repository variable:
image

If the value is set to 'github' we fallback to ubuntu-latest runners. This is useful for times there could be an outage, but also allows better control over the runners we use in jobs.

## Review / Merge checklist

- [ ] Tests included. 


packages/core/src/execution-engine/node-execution-context/utils/__tests__/request-helper-functions.test.ts
ivov, line 1: Looks unrelated?

.github/workflows/test-linting-reusable.yml
ivov, line 1: Do I understand correctly that Blacksmith runners are 4 vCPU but ubuntu-latest are 2 vCPU. Shouldn't these match or we'll be at risk of timeouts, hitting memory limits, slowness, etc.?

.github/actions/setup-nodejs/action.yml
ivov, line 1: Should we document this new env at `WORKFLOWS.md`?

.github/workflows/ci-pull-requests.yml
```diff
@@ -64,7 +64,7 @@ jobs:
   typecheck:
     name: Typecheck
     if: needs.install-and-build.outputs.non_python_changed == 'true'
-    runs-on: blacksmith-4vcpu-ubuntu-2204
+    runs-on: ${{ vars.RUNNER_PROVIDER == 'github' && 'ubuntu-latest' || 'blacksmith-4vcpu-ubuntu-2204' }}
```
ivov, line 67: This shows up in 9 spots. Anyway to abstract this to a single spot? A typo would fall us back to GH with no warning

.github/workflows/test-e2e-reusable.yml
```diff
@@ -107,7 +107,7 @@ jobs:
 
   test:
     needs: matrix
-    runs-on: ${{ inputs.runner }}
+    runs-on: ${{ vars.RUNNER_PROVIDER == 'github' && 'ubuntu-latest' || inputs.runner }}
```
ivov, line 110: Looks inconsistent? Runner is ignored in fallback mode

packages/core/src/execution-engine/node-execution-context/utils/__tests__/request-helper-functions.test.ts
shortstacked: Github runs on ubuntu 24, blacksmith runs on ubuntu 22! The headers mismatched so we had to relax them. Absolute madness though.

.github/workflows/ci-pull-requests.yml
shortstacked: Unfortunately not since runners are sorted at run time. I could look at adding a rule to actionlint to check for typos.

.github/workflows/test-e2e-reusable.yml
shortstacked: Yeah that's by design, it's basically in fallback mode we only use ubuntu-latest since it's all we have access to. So inputs/defaults are skipped. I updated the docs for it.

.github/workflows/test-linting-reusable.yml
shortstacked: Yes but that would be the "fallback" scenario, things would be slower but things could keep going.

ivov (approved):

PR #24788, opened 2026-01-23
fix(ai-builder): Ask/Build coachmark appearing as chat panel slides out (no-changelog)

mike12345567:
## Summary
Addressing an issue found by @harshalnpatil with work carried out in https://github.com/n8n-io/n8n/pull/24406

The Coachmark was appearing as the chat panel was sliding out which would occasionally leave the coachmark in the wrong location. Waiting until after animation has completed.

I've also switched this over to use a Popover rather than a tooltip, the new tooltip version doesn't have a controlled mode as its not how a tooltip should be used, the controlled mode also had some issues like appearing on hover despite being controlled.

## Review / Merge checklist

- [ ] Tests included. 


burivuhster (approved):
Looks good!

PR #24795, opened 2026-01-23
fix: Make label optional in TOTP credential

Joffcom:
## Summary
This PR makes the label optional in the TOTP node credential and updates the node to handle it, It also updates the tests so we actually call the TOTP package instead of just mocking a response so we can make sure the package and node is correctly working.

https://linear.app/n8n/issue/NODE-4251/nit-totp-api-credentials-have-optional-field-label-marked-as-required

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Totp/test/Totp.node.test.ts
```diff
@@ -1,28 +1,40 @@
...
+
+	afterAll(() => {
+		jest.restoreAllMocks();
+	});
+
+	// Pre-calculate expected token using the real OTPAuth library
+	// Note: The token value is the same whether label/issuer is present or not,
+	// since those fields are metadata and don't affect the TOTP algorithm
+	const expectedToken = new OTPAuth.TOTP({
+		secret: 'BVDRSBXQB2ZEL5HE',
```
yehorkardash: Move this string into a variable, as it is used later on in the test

yehorkardash (changes requested):
Looks good, just one small comment

packages/nodes-base/nodes/Totp/test/Totp.node.test.ts
Joffcom: Good shout, Done.

yehorkardash (approved):

PR #24798, opened 2026-01-23
feat(Crypto Node): Add credentials for Hmac and Sign operations

krisn0x:
## Summary

For the Hmac and Sign operations, we used to keep the Secret and Private Key in the node parameters, in plain text. In v2 of the Crypto node, they are now moved into credentials for increased security.

https://linear.app/n8n/issue/NODE-4253/community-issue-crypto-node-does-not-support-secret-as-a-credential
closes #16035

## Review / Merge checklist

- [X] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [X] Tests included. 


packages/nodes-base/credentials/Crypto.credentials.ts
```diff
@@ -0,0 +1,37 @@
...
+			name: 'hmacSecret',
+			type: 'string',
+			description: 'Secret used in the Hmac action',
+			typeOptions: {
+				password: true,
+			},
+			default: '',
+		},
+		{
+			displayName: 'Sign Private Key',
```
Joffcom: This should probably just be private key

packages/nodes-base/nodes/Crypto/v2/CryptoV2.node.ts
```diff
@@ -0,0 +1,576 @@
...
+				{
+					name: 'SHA384',
+					value: 'SHA384',
+				},
+				{
+					name: 'SHA512',
+					value: 'SHA512',
+				},
+			],
+			default: 'MD5',
```
Joffcom: We should maybe make sha256 the default as md5 is generally considered to be broken.

packages/nodes-base/nodes/Crypto/v2/CryptoV2.node.ts
```diff
@@ -0,0 +1,576 @@
...
```
Joffcom: Same as above

packages/nodes-base/credentials/Crypto.credentials.ts
```diff
@@ -0,0 +1,39 @@
+import type { ICredentialType, INodeProperties, Icon, ThemeIconColor } from 'n8n-workflow';
+
+// eslint-disable-next-line n8n-nodes-base/cred-class-name-unsuffixed
```
dlavrenuek: I was wondering why there is no linting error for some other credential types like `Aws` or `Ftp` and found that there is an exclude list in the linting rule: https://github.com/ivov/eslint-plugin-n8n-nodes-base/blob/master/lib/constants.ts#L202

It should not prevent merging the changes here, instead I would follow up on updating the linting rule, release a new version and then update the version + remove the `// eslint-disable...` comments here

dlavrenuek (dismissed):
One follow up suggestion from my side. Works as expected, nice work @krisn0x 🚀

packages/nodes-base/credentials/Crypto.credentials.ts
krisn0x: Good point, will do that.

krisn0x: https://github.com/ivov/eslint-plugin-n8n-nodes-base/pull/205

krisn0x: updated linter version and placed in in the pnpm exclusions, to remove the 3 day wait after a new release for that package

Joffcom (approved):

PR #24800, opened 2026-01-23
fix(core): Handle null workflowId's on workflow_statistics table at migration

Cadiac:
## Summary

On instances with old data `ChangeWorkflowStatisticsFKToNoAction1767018516000` migration at 2.5.2 might fail:

```
20:22:49.422   info    Starting migration ChangeWorkflowStatisticsFKToNoAction1767018516000 { "file": "migration-helpers.js", "function": "logMigrationStart" }
Migration "ChangeWorkflowStatisticsFKToNoAction1767018516000" failed, error: SQLITE_CONSTRAINT: NOT NULL constraint failed: TMP_workflow_statistics.workflowId
20:22:49.430   error   There was an error running database migrations { "file": "error-reporter.js", "function": "defaultReport" }
20:22:49.430   error   SQLITE_CONSTRAINT: NOT NULL constraint failed: TMP_workflow_statistics.workflowId { "file": "error-reporter.js", "function": "defaultReport" }
```

Apparently `workflow_statistics` table had previously a nullable `workflowId` column, which the new migration didn't take into account. It being nullable before seems like an older mistake, as the entity had been typed as non nullable anyways.

Some old instances from early days might have data with null `workflowId` in them, and we ran into this on couple of instances with sqlite databases.

This PR changes the migration to dorp such rows, I don't think they should exist in the DB anyways. Keeping the down migration to recreate the table with non-nullable column also seems right, as it being nullable before seems like a mistake.

On postgres the column already wasn't nullable, so this doesn't affect postgres.

https://n8nio.slack.com/archives/C034S4SV6LX/p1769192966742429

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
+++ b/packages/@n8n/db/src/migrations/sqlite/1767018516000-ChangeWorkflowStatisticsFKToNoAction.ts
@@ -12,7 +12,7 @@ export class ChangeWorkflowStatisticsFKToNoAction1767018516000 implements Revers
 	transaction = false as const; // Disable FK checks for table recreation
 
 	async up({ queryRunner, tablePrefix }: MigrationContext) {
-		// Create new table with id primary key, workflowName column, and no foreign key constraint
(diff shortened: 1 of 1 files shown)


guillaumejacquart (approved):
Good catch ! thanks

PR #24849, opened 2026-01-26
fix(editor): Allow '-' and '_' in subworkflow caller IDs

ibex088:
## Problem

When setting **Allowed Caller Workflows** for a subworkflow, IDs containing `-` or `_` were getting stripped out. So an ID like `luPpn77f_KhU1e-F9bQXu` would turn into `luPpn77fKhU1eF9bQXu`, making it impossible to authorize the right workflow.

## Solution

This updates the input validation to allow `-` and `_` in workflow IDs so they’re no longer removed. I also added a test case covering this to make sure it doesn’t break again in the future.

Fixes #24841
Fixes GHC-6532

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/WorkflowSettings.test.ts
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowSettings.test.ts
@@ -223,8 +223,9 @@ describe('WorkflowSettingsVue', () => {
 		});
 	});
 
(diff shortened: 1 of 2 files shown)


mfsiega (approved):
Thanks a lot for the contribution!

PR #24862, opened 2026-01-26
feat(core): Fill out external secret provider endpoints for CRUD operations

geemanjs:
## Summary

Allow Create, Read, Update and Delete operations for external secret provider connections.  

closes LIGO-115
closes LIGO-114

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.controller.ee.ts
```diff
@@ -32,72 +41,122 @@ export class SecretProvidersConnectionsController {
 		next();
 	}
 
-	@Put('/:providerKey')
+	@Post('/')
+	@Licensed('feat:externalSecrets')
```
ireneea: The whole module is guarded with the `feat:externalSecrets` license, so this extra decorator `@Licensed` decorator is not really needed.

Same for the other handlers

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -0,0 +1,142 @@
...
+		private readonly projectAccessRepository: ProjectSecretsProviderAccessRepository,
+		private readonly cipher: Cipher,
+	) {}
+
+	async createConnection(
+		providerKey: string,
+		type: string,
+		projectIds: string[],
+		settings: IDataObject,
+	): Promise<SecretsProviderConnection> {
```
ireneea, line 26: Would it be simpler to take a `CreateSecretsProviderConnectionDto` object as parameter here?

Same for the `updateConnection` method

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -0,0 +1,142 @@
...
+		},
+	): Promise<SecretsProviderConnection> {
+		const connection = await this.repository.findOne({ where: { providerKey } });
+
+		if (!connection) {
+			throw new NotFoundError(`Connection with key "${providerKey}" not found`);
+		}
+		if (updates.type !== undefined) {
+			connection.type = updates.type;
+		}
```
ireneea, line 82: Should we consider clearing the existing settings when changing the connection type?

packages/cli/test/integration/secret-providers-connections.api.test.ts
```diff
@@ -0,0 +1,553 @@
...
+		test('should fail for non-existent connection', async () => {
+			const response = await ownerAgent
+				.get('/secret-providers/connections/does-not-exist')
+				.expect(404);
+
+			expect(response.body.message).toBe('Connection with key "does-not-exist" not found');
+		});
+	});
+
+	describe('Security', () => {
```
ireneea, line 426: Would some tests for the access control to make sure non-instance owner or instance admins can access those endpoints

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ts
```diff
@@ -0,0 +1,142 @@
...
+			id: connection.providerKey,
+			name: connection.providerKey,
+			type: connection.type as SecretsProviderType,
+
+			isEnabled: connection.isEnabled,
+			projects: connection.projectAccess.map((access) => ({
+				id: access.project.id,
+				name: access.project.name,
+			})),
+			settings: {}, // We should never return these back to the FE so force it to an empty object
```
konstantintieber: I could see this "always empty settings" object confusing myself as an engineer debugging issues via the network tab.
What do you think about omitting it altogether from the public connection response? We'd introduce a dedicated type for the response of these endpoints then instead of using the `SecretProviderConnection` entity.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
geemanjs: I don't think it will ever be used in that way - but right now that situation is possible.

I will add some validation to check that if you change the type you must also specify new settings.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ts
geemanjs: This is a good idea - do you want a "specific type" here or can I just use an `Omit<SecretProviderConnection, "settings">` to remove them from the creation of the public dto?

konstantintieber: Is `SecretProviderConnection` considered the public dto? To me it's the entity. I'm leaning towards a dedicated type. E.g. like we have for workflows: `IWorkflowResponse`

konstantintieber: I can see that for many APIs under `/rest` the response types live inside https://github.com/n8n-io/n8n/blob/master/packages/cli/src/requests.ts

I don't think the types of this API need to live in that file, but the structure of these types seems best practice.

geemanjs: It probably makes sense to add a `requests.ts` in a module 👍 

I'll do that and update 👍

packages/cli/src/modules/external-secrets.ee/secrets-providers.responses.ee.ts
```diff
@@ -0,0 +1,8 @@
+import type { SecretProviderConnection } from '@n8n/api-types';
+
+export declare namespace SecretsProvidersResponses {
+	type StrippedConnection = Omit<SecretProviderConnection, 'settings' | 'secretsCount' | 'state'>;
```
konstantintieber, line 4: Nit: while `Omit` is shorter to write, `Pick` would be more strict by default by always forcing a dev to expose newly added properties to the public response consciously.

konstantintieber (approved):
Awesome!

ireneea (approved):

PR #24867, opened 2026-01-26
feat(editor): Prioritize newer models on Chat hub model picker

Cadiac:
## Summary

Add optional priority field on model metadata, and using that priotize newer models on Chat hub model picker so that new models get shown first on the list & automatically selected when user comes to chat hub without a selected credential.

image

Previously we picked the first on the list, which in openai's case what chatgpt-4o-latest..

https://linear.app/n8n/issue/CHA-118/feature-prioritize-new-models-when-automatically-selecting-a-model

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
+++ b/packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
@@ -345,18 +345,25 @@ export function findOneFromModelsResponse(
 	response: ChatModelsResponse,
 	providerSettings: Partial<Record<ChatHubLLMProvider, ChatProviderSettingsDto>>,
 ): ChatModelDto | undefined {
+	let bestModel: ChatModelDto | undefined;
+	let bestPriority = -Infinity;
+
 	for (const provider of chatHubProviderSchema.options) {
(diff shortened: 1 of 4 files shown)


autologie (dismissed):
nice, thanks!

packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
```diff
@@ -345,18 +345,25 @@ export function findOneFromModelsResponse(
...
+			const priority = model.metadata.priority ?? 0;
+			if (priority > bestPriority) {
+				bestPriority = priority;
+				bestModel = model;
+			}
 		}
 	}
 
-	return undefined;
+	return bestModel;
```
Cadiac: Hm, this now slightly changes the behaviour - if the first provider finds one with priority "50", and second has one with priority "100" we pick the second one, instead of stopping after first provider that was available - should I just make this work like before?

Cadiac: Restored this previous behaviour of just considering providers one by one, stopping after first provider that had an available model (and picking the highest priority one from among them)

autologie (approved):

PR #24873, opened 2026-01-26
chore: Add secrets provider connection cache service

ireneea:
## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/external-secrets.ee/secrets-provider-connection-store-cache.service.ts
```diff
@@ -0,0 +1,117 @@
...
+		return savedConnection;
+	}
+
+	async update(connection: SecretsProviderConnection): Promise<SecretsProviderConnection> {
+		const result = await this.connectionRepository.updateByProviderKey(
+			connection.providerKey,
+			connection,
+		);
+
+		if (result.affected) {
```
konstantintieber, line 84: What's the motivation for handling the case of calling the cache.update() method with arguments that have no matching provider in the DB gracefully?

I would've expected something like `if (!result.affected) throw new SecretsProviderNotFound(providerKey)`

Same for delete.

PR #24876, opened 2026-01-26
fix(core): Ensure execution ID is available at runtime in Webhook response

ivov:
https://linear.app/n8n/issue/CAT-2213

### DIFF
--- a/packages/cli/src/webhooks/webhook-helpers.ts
+++ b/packages/cli/src/webhooks/webhook-helpers.ts
@@ -567,14 +567,9 @@ export async function executeWebhook(
 			return;
 		}
 
-		// Now that we know that the workflow should run we can return the default response
-		// directly if responseMode it set to "onReceived" and a response should be sent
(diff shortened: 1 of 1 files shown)


mfsiega (commented):
I would lightly push back on the idea that we should make this change without attempting to clean the function up first.

I agree the behaviour is broken, but as far as I can tell it's been this way for a long time (years?) so in a sense this is de facto the "correct" behaviour. I see some customers are asking for this, and I agree it's an improvement, but I don't totally see the urgency that justifies making the code health worse.

Is there some aspect I'm missing?

mfsiega (approved):
On second thought, ship it. 🚀 

This behaviour just seems way more consistent. If `WorkflowRunner.run` throws an error and the workflow never gets executed at all, we shouldn't be returning a `200` to the webhook. I think this problem probably exists with other triggers (we ack the event without a strong guarantee that the workflow execution happened), but we may as well fix it here.

I do still want to clean this function up, at least a bit 😄 might try to get to that this week.

PR #24877, opened 2026-01-26
chore: Add secrets provider connection tables migration

ireneea:
## Summary

Add secrets provider connection tables migration

Resolves LOGO-131

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
+++ b/packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
@@ -23,12 +23,12 @@ export class CreateSecretsProviderConnectionTables1769433700000 implements Rever
 		// Create project_secrets_provider_access table (join table with FKs)
 		await createTable(projectSecretsProviderAccessTable)
 			.withColumns(
-				column('providerKey').varchar().primary.notNull,
+				column('secretsProviderConnectionId').varchar().primary.notNull,
 				column('projectId').varchar(36).primary.notNull,
 			)
-			.withTimestamps.withForeignKey('providerKey', {
+			.withTimestamps.withForeignKey('secretsProviderConnectionId', {
 				tableName: secretsProviderConnectionTable,
-				columnName: 'providerKey',
+				columnName: 'id',
 				onDelete: 'CASCADE',
 			})
 			.withForeignKey('projectId', {
--- a/packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
+++ b/packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
@@ -4,10 +4,11 @@ const secretsProviderConnectionTable = 'secrets_provider_connection';
 const projectSecretsProviderAccessTable = 'project_secrets_provider_access';
 
 export class CreateSecretsProviderConnectionTables1769433700000 implements ReversibleMigration {
-	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+	async up({ schemaBuilder: { createTable, column, createIndex } }: MigrationContext) {
 		// Create secrets_provider_connection table first (parent table)
 		await createTable(secretsProviderConnectionTable).withColumns(
-			column('providerKey').varchar().primary.notNull,
+			column('id').varchar().primary.notNull,
+			column('providerKey').varchar(128).notNull,
 			column('type')
 				.varchar(36)
 				.notNull.comment(
@@ -16,6 +17,8 @@ export class CreateSecretsProviderConnectionTables1769433700000 implements Rever
 			column('encryptedSettings').text.notNull,
 			column('isEnabled').bool.default(false).notNull,
 		).withTimestamps;
+
+		await createIndex(secretsProviderConnectionTable, ['providerKey'], true);
 
 		// Create project_secrets_provider_access table (join table with FKs)
 		await createTable(projectSecretsProviderAccessTable)
--- a/packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
+++ b/packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
@@ -28,9 +28,7 @@ export class CreateSecretsProviderConnectionTables1769433700000 implements Rever
 				tableName: 'project',
 				columnName: 'id',
 				onDelete: 'CASCADE',
-			})
-			.withIndexOn('providerKey')
-			.withIndexOn('projectId');
+			});
 	}
 
 	async down({ schemaBuilder: { dropTable } }: MigrationContext) {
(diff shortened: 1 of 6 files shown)


packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,39 @@
...
+			.withTimestamps.withForeignKey('providerKey', {
+				tableName: secretsProviderConnectionTable,
+				columnName: 'providerKey',
+				onDelete: 'CASCADE',
+			})
+			.withForeignKey('projectId', {
+				tableName: 'project',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			});
```
geemanjs, line 38: Should we add some indexes here? 
```suggestion
			})
			.withIndexOn(['providerKey'])
			.withIndexOn(['projectId']);
```

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const secretsProviderConnectionTable = 'secrets_provider_connection';
+const projectSecretsProviderAccessTable = 'project_secrets_provider_access';
+
+export class CreateSecretsProviderConnectionTables1769433700000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		// Create secrets_provider_connection table first (parent table)
+		await createTable(secretsProviderConnectionTable).withColumns(
+			column('providerKey').varchar().primary.notNull,
+			column('type').varchar().notNull,
```
geemanjs: Do we want to apply any length constraints here? Or do we normally just leave them at the defaults?

ireneea: There seems to be a a mix in the code. I don't have a strong opinions on this because it doesn't seem like it makes a big difference.

Any preferences?

geemanjs: I have no preference - the only "reason" I can think for adding them is that each db we support probably has a different default value for the length which could lead to something unexpected popping up in one vs another. However we mostly do validation before the DB layer so I think this is fine 😄

geemanjs (dismissed):

packages/@n8n/db/src/migrations/mysqldb/index.ts
```diff
@@ -263,4 +264,5 @@ export const mysqlMigrations: Migration[] = [
 	AddAgentIdForeignKeys1765886667897,
 	AddWorkflowVersionIdToExecutionData1765892199653,
 	AddPublishedVersionIdToWorkflowDependency1769000000000,
+	CreateSecretsProviderConnectionTables1769433700000,
```
tomi: No need to add to mysql anymore, it's no longer supported

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,41 @@
...
```
tomi: Let's add some sensible max sizes to these

tomi: Do we already know what values this can take? We could add a `.comment()` on it or a check constraint on the table level `.withEnumCheck()`. It's also fine to keep it flexible so we don't need a migration every time we introduce a new secrets provider type

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,41 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const secretsProviderConnectionTable = 'secrets_provider_connection';
+const projectSecretsProviderAccessTable = 'project_secrets_provider_access';
+
+export class CreateSecretsProviderConnectionTables1769433700000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		// Create secrets_provider_connection table first (parent table)
+		await createTable(secretsProviderConnectionTable).withColumns(
+			column('providerKey').varchar().primary.notNull,
```
tomi: Is this something we generate or something users give? If it's coming from users, would make more sense to have a surrogate PK (e.g. `id`) and something like `name` that is coming from users

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,41 @@
...
+				tableName: secretsProviderConnectionTable,
+				columnName: 'providerKey',
+				onDelete: 'CASCADE',
+			})
+			.withForeignKey('projectId', {
+				tableName: 'project',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withIndexOn('providerKey')
```
tomi: We don't need this index. PK already creates a unique index (providerKey, projectId), which can serve these queries. The projectId index can still be added if we need to query all secrets providers for a specific project

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
ireneea: I've added a .comment(), we've had a recent case of a provider being deprecated. Might add a more strict approach if we have a confidence that the provider list is stable

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,43 @@
...
+			.withTimestamps.withForeignKey('providerKey', {
+				tableName: secretsProviderConnectionTable,
+				columnName: 'providerKey',
+				onDelete: 'CASCADE',
+			})
+			.withForeignKey('projectId', {
+				tableName: 'project',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			});
```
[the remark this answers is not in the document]
tomi: ☝️ IF we are querying all the providers per project

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
```diff
@@ -0,0 +1,46 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const secretsProviderConnectionTable = 'secrets_provider_connection';
+const projectSecretsProviderAccessTable = 'project_secrets_provider_access';
+
+export class CreateSecretsProviderConnectionTables1769433700000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column, createIndex } }: MigrationContext) {
+		// Create secrets_provider_connection table first (parent table)
+		await createTable(secretsProviderConnectionTable).withColumns(
+			column('id').varchar().primary.notNull,
```
tomi: Let's use `autoGenerate2` so it gets autogenerated and is a simple and fast numeric type

tomi (approved):
Thank you for addressing the feedback 🫶 🚀

packages/@n8n/db/src/migrations/common/1769433700000-CreateSecretsProvidersConnectionTables.ts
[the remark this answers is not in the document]
ireneea: added back the `projectId` index because we will have requests filtering by projectId (and secretsProviderConnectionId but that's already covered by the compound index)

PR #24879, opened 2026-01-26
fix(core): Ensure workflows with response promises respect graceful shutdown

ivov:
## Summary

- Before: Workflows using "Respond to Webhook" were immediately cancelled on `SIGTERM`, even though the HTTP connection remains open until the process actually dies
- After: These workflows are now given the same graceful shutdown treatment as regular workflows, allowing them to complete and respond if they manage to finish before `N8N_GRACEFUL_SHUTDOWN_TIMEOUT`

The original logic assumed the response couldn't be sent after `SIGTERM`, but that's only true after the process terminates, not at the moment the signal is received

https://linear.app/n8n/issue/CAT-2212/

## Review / Merge checklist

- [ ] Tests included. 


mfsiega (approved):
LGTM.

PR #24884, opened 2026-01-26
ci: Fix release tag merge failing with unrelated histories

seemewalkin:
## Summary

  Fixes the [release tag merge job](https://github.com/n8n-io/n8n/actions/runs/21363602394/job/61491568277) failing with "unrelated histories" error by:
  - Using full clone (fetch-depth: 0) instead of shallow clone to provide git with complete history
  - Changing --ff-only to --ff to allow merge commits when master has diverged (PRs merged during release publishing)

  https://linear.app/n8n/issue/CAT-2224

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -202,7 +202,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           ref: master
-          fetch-depth: 0
+          fetch-depth: 500
           token: ${{ steps.generate_token.outputs.token }}
 
       - name: Verify release tag exists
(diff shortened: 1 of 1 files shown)


shortstacked (dismissed):

.github/workflows/release-publish.yml
```diff
@@ -202,7 +202,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           ref: master
-          fetch-depth: 1
+          fetch-depth: 0
```
tomi: We have a really long history already and there's no reason for us needing to fetch it all (it already takes 12s on CI). We can make it something more reasonable like 500 which should cover the history up to the point where the release branch got forked.

seemewalkin: was able to reproduce locally without both changes 👍  fetch-depth: 500 should work!

tomi (approved):
🚀

PR #7138, opened 2023-09-08
refactor(core): Move event and telemetry handling into workers in queue mode

flipswitchingmonkey:
# Motivation

In Queue mode, finished executions would cause the main instance to always pull all execution data from the database, unflatten it and then use it to send out event log events and telemetry events, as well as required returns to Respond to Webhook nodes etc.

This could cause OOM errors when the data was large, since it had to be fully unpacked and transformed on the main instance’s side, using up a lot of memory (and time).

This PR attempts to limit this behaviour to only happen in those required cases where the data has to be forwarded to some waiting webhook, for example.

# Changes

Execution data is only required in cases, where the active execution has a `postExecutePromise` attached to it. These usually forward the data to some other endpoint (e.g. a listening webhook connection).

By adding a helper `getPostExecutePromiseCount()`, we can decide that in cases where there is nothing listening at all, there is no reason to pull the data on the main instance.

Previously, there would always be postExecutePromises because the telemetry events were called. Now, these have been moved into the workers, which have been given the various InternalHooks calls to their hook function arrays, so they themselves issue these telemetry and event calls.

This results in all event log messages to now be logged on the worker’s event log, as well as the worker’s eventbus being the one to send out the events to destinations. The main event log does…pretty much nothing.

We are not logging executions on the main event log any more, because this would require all events to be replicated 1:1 from the workers to the main instance(s) (this IS possible and implemented, see the worker’s `replicateToRedisEventLogFunction` - but it is not enabled to reduce the amount of traffic over redis).

Partial events in the main log could confuse the recovery process and would result in, ironically, the recovery corrupting the execution data by considering them crashed.

# Refactor

I have also used the opportunity to reduce duplicate code and move some of the hook functionality into `packages/cli/src/executionLifecycleHooks/shared/sharedHookFunctions.ts` in preparation for a future full refactor of the hooks

PR #5112, opened 2023-01-09
feat(core): Add execution runData recovery and status field

flipswitchingmonkey:
This PR will add a `status` field to the execution table and sets the status at appropriate points during the executions lifecycle.

It also leverages the event logging journal to attempt to recover execution run data in the case of a crashed execution (e.g. through out of memory issues), to pinpoint the node that caused the crash.

Also updates the frontend accordingly to read the status field.

https://github.com/n8n-io/n8n/issues/2195

PR #24897, opened 2026-01-27
fix: Fix status overwrite for donePromise

michael-radency:
## Summary

overwriting status here resolves that returned IRun data have status 'running'

https://linear.app/n8n/issue/CAT-2228/fix-status-overwrite-for-runpromise

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/workflow-runner.ts
+++ b/packages/cli/src/workflow-runner.ts
@@ -338,7 +338,6 @@ export class WorkflowRunner {
 					if (workflowExecution.isCanceled) {
 						fullRunData.finished = false;
 					}
-					fullRunData.status = this.activeExecutions.getStatus(executionId);
 					this.activeExecutions.resolveExecutionResponsePromise(executionId);
 					this.activeExecutions.finalizeExecution(executionId, fullRunData);
 				})


mfsiega (commented):
Looked a bit more into this - it seems like we need this line at least to handle cancellations and maybe some other similar cases, where the status gets updated "from the outside".

Looking to see if it's possible to end up in a case where `activeExecutions` is not up-to-date, and we should prefer the run data.

ivov (approved):
As far I can tell this change is correct. 

The in-memory `activeExecutions` map is not the source of truth for final status, and the overwrite does cause a stale status. When cancelling, the engine's `onCancel` sets status to `canceled` and uses that status when building `fullRunData` that it returns, later `determineFinalExecutionStatus` preserves this `canceled` status.

Digging back this was added at #5112. Back then this was not a bug because the engine returned run data without status so `activeExecutions` was the source of truth, but #7138 started setting status internally with hooks determining final status.

Hard to be 100% certain as this flow is an ancient tangle, but in principle this looks safe. Ideally we could consider if we should also update `activeExecutions` so no future consumers receive stale state, but that may be out of scope. Up to you.

PR #24902, opened 2026-01-27
perf(core): Lazyload Swagger, PSL and MCP libs

ivov:
Lazy load rarely used heavy libs for 11 MB baseline memory gain.

- Before: https://share.cleanshot.com/m4VDDhr5
- After: https://share.cleanshot.com/pzcBnR43

Manually tested Public API docs page and `psl` telemetry change. Not familiar with MCP module to test it.

Follow up to #24693 and #24107 for a total of 51 MB baseline memory reduction


Cadiac (dismissed):
I'm not familiar with all of this code, but changing these to lazy-loaded seems fine and the gain to baseline memory usage is nice 👍

Cadiac (approved):

PR #24904, opened 2026-01-27
fix(editor): Disable WF history button when autosave is not complete

dariacodes:
## Summary

Disable workflow history button until autosave is complete.
We have 3 autosave states: `Idle`, `Scheduled` and `InProgress`.

This PR makes sure that, as before, we show the loading indicator when autosave is running (`InProgress`), and if it is either scheduled or running, we're disabling the workflow history button and showing a tooltip depending on the autosave state:

image
image

I had to use conditional logic for displaying the router component to prevent the button clicks going through.

Closes [ADO-4681](https://linear.app/n8n/issue/ADO-4681/bug-workflow-history-button-is-enabled-when-autosave-is-running)

## Review / Merge checklist

- [ ] Tests included. 


MiloradFilipovic (dismissed):
Nice one, looks good on my end. Was wondering if we should disable more things (publish button, rename field, tags, ...) while it's saving (couldn't make get it to break using these but just wondering.

MiloradFilipovic (approved):
🚢

PR #24910, opened 2026-01-27
refactor(core): Use task runner for sort operations

tomi:
## Summary

Migrates sort operations to the task runner infrastructure:
- Adds 'runCode' execution mode to JS task runner for lightweight code execution without RPC/builtin access
- Updates Sort node and ItemLists (V1/V2/V3) nodes to use task runner for custom sort code
- Simplifies JsTaskRunnerSandbox API by removing nodeMode from constructor

https://linear.app/n8n/issue/CAT-2146

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/task-runner/src/js-task-runner/js-task-runner.ts
ivov, line 1: An updated main sending `nodeMode: 'runCode'` to a runner that hasn't been updated yet would cause the runner to fall through to each item mode and run the task in that context.

packages/@n8n/task-runner/src/js-task-runner/js-task-runner.ts
```diff
@@ -252,6 +270,50 @@ export class JsTaskRunner extends TaskRunner {
...
+			const result = await new Promise<unknown>((resolve, reject) => {
+				const abortHandler = () => {
+					reject(new TimeoutError(this.taskTimeout));
+				};
+
+				abortSignal.addEventListener('abort', abortHandler, { once: true });
+
+				let taskResult: Promise<unknown>;
+
+				if (this.mode === 'secure') {
```
ivov: Do we expect `runCode` to allow access to third-party libs where the secure/insecure distinction matters? There's no `require` in the lightweight context.

ivov (commented):
I'll test in a bit

packages/@n8n/task-runner/src/js-task-runner/js-task-runner.ts
tomi: Uff that's unfortunate. Should fix it to fail in that case. However I see the chances of this happening being slim, since runners are usually deployed as sidecar containers, and they get deployed together with n8n. I created https://linear.app/n8n/issue/CAT-2257/validate-task-runner-task-settings for this

packages/@n8n/task-runner/src/js-task-runner/js-task-runner.ts
tomi: No, but I wanted to keep it unified with the rest of the runner.

ivov (approved):
🚀

PR #24912, opened 2026-01-27
refactor(core): Route execution creation through `ExecutionPersistence`

ivov:
## Summary

- Refactor `ExecutionRepository.createNewExecution` into `ExecutionRepository.create` with identical signature, using execution data stores
- Migrate callsites to `ExecutionRepository.create`

https://linear.app/n8n/issue/CAT-2097

## Review / Merge checklist

- [x] Tests included. 


packages/cli/test/integration/database/repositories/execution.repository.test.ts
```diff
@@ -17,72 +17,6 @@ describe('ExecutionRepository', () => {
 		await testDb.terminate();
 	});
 
-	describe('createNewExecution', () => {
```
ivov, line 20: Moved but untouched (except naming ofc)

packages/cli/src/executions/execution-persistence.ts
```diff
@@ -18,8 +23,39 @@ export class ExecutionPersistence {
 		private readonly executionRepository: ExecutionRepository,
 		private readonly binaryDataService: BinaryDataService,
 		private readonly fsStore: FsStore,
+		private readonly storageConfig: StorageConfig,
 	) {}
 
+	async create(payload: CreateExecutionPayload) {
```
tomi, line 34: Can we have a comment what this method does

packages/cli/src/executions/execution-persistence.ts
```diff
@@ -18,8 +23,39 @@ export class ExecutionPersistence {
...
+		const executionEntity = { ...rest, createdAt: new Date(), storedAt };
+		const data = stringify(rawData);
+		const workflowVersionId = workflowData.versionId ?? null;
+
+		return await this.executionRepository.manager.transaction(async (tx) => {
+			const { identifiers } = await tx.insert(ExecutionEntity, executionEntity);
+			const executionId = String(identifiers[0].id);
+
+			if (storedAt === 'db') {
+				await tx.insert(ExecutionData, {
```
tomi, line 48: I would imagine in the long run each "driver" have the exact same interface. That way we wouldn't have to create conditional code like this, but instead would say something like `persistence.writeExecution(..)` where `persistence` refers to the currently configured driver. So the DB driver would implement the same interface as all the other drivers

packages/cli/src/executions/execution-persistence.ts
```diff
@@ -18,8 +23,39 @@ export class ExecutionPersistence {
...
+					executionId,
+					workflowData: { connections, nodes, name, settings, id },
```
tomi: Here we pass in the `workflowData` as a newly constructed object, below as is. Shouldn't it be same for both?

tomi (commented):
Couple comments

packages/cli/src/executions/execution-persistence.ts
ivov: > each "driver" have the exact same interface

I anticipated this in the plan :) The DB store is a special case because we need a transaction to maintain consistency between an execution and its execution data, but we don't want to leak a DB-specific concern to the method signature in the common interface. Open to ideas if you know of a way to avoid special-casing this, but I'm good also with accepting the conditional and refactoring later if need be.

tomi: The options are to either to
1. Add it to the interface
2. Keep it conditional

I guess I would add it to the interface even tho it doesn't necessarily belong there. At least it then avoids this special casing in the caller side

tomi: But open to refactoring it later on as well

ivov: Let's bring this in and refactor later

tomi (approved):
🚀

PR #24922, opened 2026-01-27
fix: Validate OpenAI urls

yehorkardash:
## Summary

This PR adds OpenAI base url verification.
It also reverts this [change](https://github.com/n8n-io/n8n/pull/24702), as it introduces a regression in how 'none' is being handled by default

https://linear.app/n8n/issue/NODE-4259

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOpenAI/EmbeddingsOpenAi.node.ts
+++ b/packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOpenAI/EmbeddingsOpenAi.node.ts
@@ -1,20 +1,20 @@
 import { OpenAIEmbeddings } from '@langchain/openai';
+import { AiConfig } from '@n8n/config';
+import { Container } from '@n8n/di';
 import {
 	NodeConnectionTypes,
+	type INodeProperties,
 	type INodeType,
 	type INodeTypeDescription,
-	type SupplyData,
 	type ISupplyDataFunctions,
-	type INodeProperties,
+	type SupplyData,
 } from 'n8n-workflow';
 import type { ClientOptions } from 'openai';
 
-import { logWrapper } from '@utils/logWrapper';
-
+import { checkDomainRestrictions } from '@utils/checkDomainRestrictions';
 import { getProxyAgent } from '@utils/httpProxyAgent';
+import { logWrapper } from '@utils/logWrapper';
 import { getConnectionHintNoticeField } from '@utils/sharedFields';
-import { Container } from '@n8n/di';
-import { AiConfig } from '@n8n/config';
 
 const modelParameter: INodeProperties = {
 	displayName: 'Model',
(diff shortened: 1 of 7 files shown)


packages/cli/src/load-nodes-and-credentials.ts
```diff
@@ -725,7 +717,7 @@ export class LoadNodesAndCredentials {
 						description: 'Block all requests when used in the HTTP Request node',
 					},
 				],
-				default: 'none',
+				default: 'all',
```
DawidMyslak, line 720: 👏

packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOpenAI/EmbeddingsOpenAi.node.ts
```diff
@@ -253,6 +253,7 @@ export class EmbeddingsOpenAi implements INodeType {
 			defaultHeaders,
 		};
 		if (options.baseURL) {
+			checkDomainRestrictions(this, credentials, options.baseURL);
```
DawidMyslak, line 256: When `options.baseURL` is not provided, the code falls back to using `credentials.url` without validating it against domain restrictions. This allows users to bypass security policies they configured on their credentials. Is this expected? Maybe I'm missing some context.

packages/@n8n/nodes-langchain/nodes/vendors/OpenAi/v1/actions/assistant/message.operation.ts
```diff
@@ -180,6 +181,10 @@ export async function execute(this: IExecuteFunctions, i: number): Promise<INode
 		preserveOriginalTools?: boolean;
 	};
 
+	if (options.baseURL) {
+		checkDomainRestrictions(this, credentials, options.baseURL);
```
DawidMyslak, line 185: So the goal of this function to return `void` when it's all good, but throw an error when something is wrong? No `true/false` returns?

DawidMyslak (changes requested):
Hey @yehorkardash - I've left clarification questions, nothing blocking.

I will set it as "Request changes" temporarily and unblock once we're happy.

packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOpenAI/EmbeddingsOpenAi.node.ts
yehorkardash: We don't validate url in credentials.url, because it's the url a user puts into credentials, and they supposed to know what url they are using. Some people use OpenAI nodes with other compatible services that don't use 'openai.com' url, so we don't restrict it

packages/@n8n/nodes-langchain/nodes/vendors/OpenAi/v1/actions/assistant/message.operation.ts
yehorkardash: yes, a similar approach is used in http request node, I chose it to be somewhat consistent in the implementation

DawidMyslak (approved):

PR #24926, opened 2026-01-27
fix(core): Update computeIsPending logic to account for external auth identities

cstuncsik:
## Summary

Fix `pending` user state for other auth identities like SSO, LDAP

https://linear.app/n8n/issue/IAM-72/user-management-active-user-showing-up-as-pending-after-rolling-out

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/entities/user.ts
+++ b/packages/@n8n/db/src/entities/user.ts
@@ -113,7 +113,12 @@ export class User extends WithTimestamps implements IUser, AuthPrincipal {
 	@AfterLoad()
 	@AfterUpdate()
 	computeIsPending(): void {
-		this.isPending = this.password === null && this.role?.slug !== GLOBAL_OWNER_ROLE.slug;
+		const hasExternalAuthIdentity =
+			this.authIdentities?.some((identity) => identity.providerType !== 'email') ?? false;
+		this.isPending =
+			this.password === null &&
+			!hasExternalAuthIdentity &&
+			this.role?.slug !== GLOBAL_OWNER_ROLE.slug;
 	}
 
 	toJSON() {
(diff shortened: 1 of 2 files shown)


packages/@n8n/db/src/entities/user.ts
```diff
@@ -113,7 +113,12 @@ export class User extends WithTimestamps implements IUser, AuthPrincipal {
 	@AfterLoad()
 	@AfterUpdate()
 	computeIsPending(): void {
-		this.isPending = this.password === null && this.role?.slug !== GLOBAL_OWNER_ROLE.slug;
+		const hasExternalAuthIdentity =
+			this.authIdentities?.some((identity) => identity.providerType !== 'email') ?? false;
+		this.isPending =
+			this.password === null &&
+			!hasExternalAuthIdentity &&
```
guillaumejacquart, line 120: is there a risk of authIdentities not being loaded from the db here, leading to wrongfully thinking it has none ?

guillaumejacquart (commented):
one question

packages/@n8n/db/src/entities/user.ts
cstuncsik: Yes, but eagerly loading `authIdentities` might incur some performance overhead.
The defensive fallback (`?? false`) preserves backward compatibility when authIdentities aren’t loaded, which I think is acceptable (in the worst case, the same behavior occurs as in the current production).
The affected queries (users list) already load authIdentities, other queries don’t need `isPending`

guillaumejacquart: makes sense, thanks !

guillaumejacquart (approved):

PR #24934, opened 2026-01-27
feat(editor): Add redesigned Switch component

DawidMyslak:
## Summary

Adds a new redesigned Switch component using Reka UI primitives, integrated into the Node Details View (NDV) parameter panel behind the `COLLECTION_OVERHAUL_EXPERIMENT` feature flag.

### Changes

**Design System (`@n8n/design-system`):**
- New `N8nSwitch2` component built with Reka UI (`SwitchRoot`, `SwitchThumb`)
- Horizontal layout with switch on left, label on right
- Two sizes: `small` (32×16px for NDV) and `large` (40×20px for settings)
- Proper accessibility support (keyboard navigation, ARIA roles)
- Unit tests and Storybook stories

**Editor UI:**
- `ParameterInput.vue`: Conditionally renders `N8nSwitch2` for boolean parameters when feature flag enabled
- `ParameterInputFull.vue`: New inline layout for switch parameters with options (Fixed/Expression) on the same row

### How to test

1. Enable the `048_collection_overhaul` feature flag
2. Open any node with boolean parameters (e.g., HTTP Request node → "Required Field")
3. Verify the switch appears with horizontal layout (switch left, label right)
4. Verify hover shows Fixed/Expression toggle inline on the right

https://linear.app/n8n/issue/NODE-4266

## Review / Merge checklist

- [x] Tests included.


packages/frontend/@n8n/design-system/src/v2/components/Switch/component-switch.md
```diff
@@ -0,0 +1,82 @@
+# Component specification
```
DawidMyslak, line 1: Heavily inspired by the existing checkbox component and its API.

packages/frontend/@n8n/design-system/src/v2/components/Switch/Switch.vue
```diff
@@ -0,0 +1,164 @@
...
+const props = withDefaults(defineProps<SwitchProps>(), {
+	size: 'small',
+});
+const slots = defineSlots<SwitchSlots>();
+const emit = defineEmits<SwitchEmits>();
+const uuid = computed(() => props.id ?? useId());
+
+const rootProps = useForwardProps(reactivePick(props, 'required', 'value', 'defaultValue'));
+
+const modelValue = defineModel<boolean>({ default: undefined });
```
elsmr, line 18: Doesn't this add an `update:model-value` emit and a `modelValue` prop? Does that mean the `change` emit and `value` prop are redundant?

elsmr (commented):
Looks really good, kudos! 🎉  one question on the component props/emits

packages/frontend/@n8n/design-system/src/v2/components/Switch/Switch.vue
DawidMyslak: Good catch @elsmr. I copy/pasted this line from existing `Checkbox.vue`, assuming we need this.

You're right about it being redundant. With `defineModel`, component consumers can already do:
- `v-model="checked"` - two-way binding
- `@update:modelValue="handleChange"` - one-way listening

The `change` emit adds a third option:
- `@change="handleChange"` - with a synthetic Event, a bit awkward (note the `@ts-expect-error` comment), I also copied that assuming we need to support it

I would vote to drop `change` event support, what do you think?

elsmr: Agree to drop the `change` and `value` if there's no other need for it 👍

DawidMyslak: Done, pushed 👍. Also tested locally, no issues as I haven't used `change` event in my implementation.

elsmr:
I think there's a leftover test for the `change` event, other than that it looks good to go!

elsmr (approved):

PR #24937, opened 2026-01-27
refactor(core): Rename `~/.n8n/binaryData` to `~/.n8n/storage`

ivov:
https://linear.app/n8n/issue/CAT-2101


tomi (commented):
I wonder if we should make this migration opt-in at this point and actually do it in v3 in case someone has mounted a volume on that path 🤔

packages/core/src/storage.config.ts
```diff
@@ -23,16 +25,85 @@ export class StorageConfig {
...
+	 * Migration skips if...
+	 * - already migrated,
+	 * - `N8N_STORAGE_PATH` is set (user wants custom new path),
+	 * - `N8N_BINARY_DATA_STORAGE_PATH` is set (user wants custom old path), or
+	 * - `~/.n8n/binaryData` does not exist (nothing to migrate)
+	 *
+	 * Migration throws if...
+	 * - `~/.n8n/storage` happens to be in use already (user must rename).
+	 */
+	private migrateStorageDir() {
```
mfsiega, line 66: Feels like it might be worth an info log when this migration happens?

mfsiega (approved):
LGTM, one tiny comment.

packages/core/src/storage.config.ts
ivov: Ah good call, saw this after adding to the merge queue, will add as a followup

PR #24978, opened 2026-01-28
fix(Sharepoint Node): Ensure tokens are refreshed

yehorkardash:
## Summary

Previously some operations in Sharepoint Node had property `ignoreHttpStatusErrors: true` which prevented access tokens from being refreshed, because errors were ignored. 

This PR introduces a new more granular approach that allow handling 401 errors.

Testing:
1. Setup a schedule trigger that runs every 2 hours
2. Connect Sharepoint node with getItem operation
3. Second scheduled execution should not fail

https://linear.app/n8n/issue/NODE-4081/bcg-sharepoint-get-an-item-node-fails-after-authorization-code-expired

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Microsoft/SharePoint/descriptions/common.descriptions.ts
```diff
@@ -154,3 +154,8 @@ export const siteRLC: INodeProperties = {
 	required: true,
 	type: 'resourceLocator',
 };
+
+export const ignoreHttpStatusErrorsConfig = {
+	ignore: true as const,
+	except: [401],
+};
```
michael-radency, line 161: descriptions is usually contain node properties, lets move this into utils

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
```diff
@@ -843,7 +843,14 @@ export function convertN8nRequestToAxios(n8nRequest: IHttpRequestOptions): Axios
 	}
 
 	if (n8nRequest.ignoreHttpStatusErrors) {
-		axiosRequest.validateStatus = () => true;
+		const ignoreHttpStatusErrors = n8nRequest.ignoreHttpStatusErrors;
+		if (typeof ignoreHttpStatusErrors === 'object' && ignoreHttpStatusErrors.ignore) {
+			axiosRequest.validateStatus = (status) => {
+				return !ignoreHttpStatusErrors.except.includes(status);
+			};
+		} else {
```
michael-radency: lets maybe use guard here instead

michael-radency (dismissed):
looks good 👍, added some suggestions, but not blocking

packages/nodes-base/nodes/Microsoft/SharePoint/descriptions/common.descriptions.ts
yehorkardash: This part is used only in descriptions, so I think it's fitting there

michael-radency (approved):
LGTM

PR #24983, opened 2026-01-28
fix(core): Use stricter flags when starting python runner

tomi:
## Summary

Start python runner with -I and -B flags:
- [`-I`](https://docs.python.org/3/using/cmdline.html#cmdoption-I) restrict imports to the site packages dir (/opt/runners/task-runner-python/.venv/bin/python)
- [`-B`](https://docs.python.org/3/using/cmdline.html#cmdoption-B) don't write .pyc files (the dir is readonly)

https://linear.app/n8n/issue/CAT-2223

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/task-runner-python/pyproject.toml
```diff
@@ -21,10 +21,19 @@ dev = [
...
 pythonpath = ["."]
 testpaths = ["tests"]
 addopts = "-v"
+
+[build-system]
+requires = ["setuptools>=68", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools]
+packages = ["src", "src.config", "src.errors", "src.message_types"]
```
ivov: If in future we add a new dir and forget to update here (very likely) this might fail silently or confusingly.

What about e.g. [autodiscovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html)?

packages/@n8n/task-runner-python/uv.lock
ivov, line 1: Why is there so much churn in the lockfile?

docker/images/runners/Dockerfile
```diff
@@ -74,6 +74,12 @@ RUN uv sync \
 			--all-extras \
       --no-editable
 
+# Install the python runner package itself into site packages
+RUN uv pip install .
+
+# The app is now installed into site packages, so we can remove the src directory
+RUN rm -rf /app/task-runner-python/src
+
```
ivov: Combine into a single layer? So the source is not an intermediate one

docker/images/runners/n8n-task-runners.json
```diff
@@ -47,10 +52,10 @@
 				"DEPLOYMENT_NAME"
 			],
 			"env-overrides": {
-				"PYTHONPATH": "/opt/runners/task-runner-python",
+				"PYTHON_DISABLE_REMOTE_DEBUG": "true",
```
ivov: Docs say `-I` implies `-E` which says:

> Ignore all `PYTHON*` environment variables, e.g. [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH) and [`PYTHONHOME`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHOME), that might be set.

ivov (commented):
I'll test in a bit

packages/@n8n/task-runner-python/uv.lock
tomi: Looks like I had a very old `uv` version. Should be fixed

packages/@n8n/task-runner-python/pyproject.toml
tomi: Changed to use autodiscovery

docker/images/runners/n8n-task-runners.json
tomi: I moved it to an arg to python

docker/images/runners/Dockerfile
tomi: Fixed

ivov (approved):
Changeset looks clean, haven't retested

PR #24987, opened 2026-01-28
fix(Typeform Node): Refactor signature validation

yehorkardash:
## Summary

This PR makes signature verification more generic, adds signatures to typeform webhooks, enables timestamp validation for slack node even if signature secret is not set

https://linear.app/n8n/issue/NODE-4318

## Review / Merge checklist

- [x] Tests included. 


yehorkardash:
@claude review this please

packages/nodes-base/nodes/Slack/SlackTriggerHelpers.ts
```diff
@@ -83,53 +83,43 @@ export async function downloadFile(this: IWebhookFunctions, url: string): Promis
 
 export async function verifySignature(this: IWebhookFunctions): Promise<boolean> {
 	const credential = await this.getCredentials('slackApi');
-	if (!credential?.signatureSecret) {
-		return true; // No signature secret provided, skip verification
-	}
```
DawidMyslak, line 88: Was this here to guarantee backwards compatibility?

packages/nodes-base/nodes/Slack/test/SlackTriggerHelpers.test.ts
```diff
@@ -158,5 +158,39 @@ describe('SlackTriggerHelpers', () => {
 			// Verify that update was called with the expected string (using the timestamp from the request)
 			expect(mockHmac.update).toHaveBeenCalledWith(`v0:${testTimestamp}:${testBody}`);
 		});
+
+		it('should verify timestamp even if signature secret is not set', async () => {
```
DawidMyslak, line 162: Not a problem introduced in this PR, but can we please rename the file to match the original implementation, so `*Helpers.test.ts` (plural).

DawidMyslak (commented):
2 minor questions, looks good!

packages/nodes-base/nodes/Slack/SlackTriggerHelpers.ts
yehorkardash: This is to handle the case when users doesn't fill that field.
Replaced it with `skipIfNoExpectedSignature: !signatureSecret || typeof signatureSecret !== 'string'`

packages/nodes-base/nodes/Slack/test/SlackTriggerHelpers.test.ts
yehorkardash: done 👍

packages/nodes-base/nodes/Slack/SlackTriggerHelpers.ts
DawidMyslak: True, but notice the first check before it gets to `skipIfNoExpectedSignature`:
```
const timestamp = req.header('x-slack-request-timestamp');
if (!timestamp) {
  return false;
}
```

If timestamp is not provided, it will exit early anyway. Unless every request has it, regardless of the secret being set?

yehorkardash: every request has it

DawidMyslak (approved):

packages/nodes-base/nodes/Slack/SlackTriggerHelpers.ts
DawidMyslak: Approved 👏

PR #24988, opened 2026-01-28
fix(Form Node): Render `\n` properly in the Form Ending page

### LINKED ISSUE #22277: [Form] Completion message ignores `\n` for linebreaks
### Bug Description

Hi,

i wanted to add line breaks to my completion message parameter of the Node "Form Ending". I tried it with `\n`, but that did not work, then i read the docs:

Image

Luckily `</br>` does the job: 

Image

The Form Description in the "On Form" node does render the `\n` correctly as a linebreak though.

Here a few more screenshots:

Image

Image

### To Reproduce

1. Create a new workflow
2. Select the trigger "On Form" and paste this into the form description: `Testing with backslash n: \n
Testing with html br: </br>
End`
3. Add the "Form Ending" node and paste this into the completion message: `Testing with backslash n: \n
Testing with html br: </br>
End`
4. Test the workflow

### Expected behavior

I would have expected to see both field behave the same and render both `</br>` and `\n`.

### Debug Info

# Debug info

## core

- n8nVersion: 1.120.4
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- consumerId: d1329a82-b442-4287-b0f9-3ba008304730

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2025-11-25T10:56:20.888Z

### Operating System

Ubuntu 24.04.3

### n8n Version

1.120.4

### Node.js Version

22.21.0

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

RomanDavydchuk:
## Summary

Form Trigger's "Description" field supports HTML and  `\n` newline characters. So does the "Completion Message" in Form Ending page, but `\n` didn't actually work. This PR fixes it by:
- replacing `\n` string with actual newline character
- making sure newlines are properly rendered in the handlebars template

https://linear.app/n8n/issue/NODE-4070/community-issue-form-completion-message-ignores-n-for-linebreaks
Closes #22277

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Form/utils/utils.ts
```diff
@@ -518,7 +518,7 @@ export function renderForm({
 	customCss?: string;
 	authToken?: string;
 }) {
-	formDescription = (formDescription || '').replace(/\\n/g, '\n').replace(/<br>/g, '\n');
+	formDescription = (formDescription || '').replace(/\\n/g, '\n');
```
RomanDavydchuk: `.replace(/<br>/g, '\n')` is not necessary, since `formDescription` goes through `sanitizeHtml` before getting to this line and that "normalizes" all `<br>` to `<br />`. Therefore, this `.replace` call does nothing, and the `<br />` would be still rendered as HTML anyway

yehorkardash (commented):
Is there a way to unescape and print `\n` as text?
image
image

RomanDavydchuk:
> Is there a way to unescape and print `\n` as text? image image

I don't think so, since the regex will always replace `\n` with a newline, good catch. The solution would be to update the regex to not replace `\n` with newlines if there is a `\` before it and instead replace `\\n` with `\n`

yehorkardash (dismissed):

yehorkardash (approved):

PR #24995, opened 2026-01-28
refactor(core): Use websockets on Chat hub message streaming

Cadiac:
## Summary

To be able to support long (>100s) executions, server initiated chat exchanges and reconnect / streaming to multiple clients this PR changes the transport of chat hub endpoints from chunked JSONL responses to a websockets based on using the existing push connection functionality we use also on canvas. Motivation comes also partially from wanting to support interactions where server resumes execution after a wait (HITL interactions) without user sending a message on chat hub.

Chat hub endpoints now just start human message / retry / edit with the REST request, and the data those executions produce gets streamed to all active clients of the correct user via the websocket push mechanism, `this.push.sendToUsers(message, [params.userId]);`.

Messages can be edited or retried and that should get reflected on all active browser windows belonging to this user, and execution can be stopped on any of the clients.

There's also support for reconnecting mid stream and receiving chunks of the current message that have been sent already, so user can refresh their page / navigate away from n8n and back and still rejoin the ongoing conversation if AI is still responding. `chat-session-store.service.ts` holds the buffer of chunks that's used to make reconnecting possible.

On multi-main this was quite a bit more complex than I initially thought, as a situation where user's websocket connects to main `A` and the rest request gets handled by main `B`. In that case the main `B` would be coordinating the chat execution (and a worker would likely be what actually runs the execution), but main `A` needs to be the one that pushes the data to the client.

To support this I copied the approach from `relay-execution-lifecycle-event` pubsub mechanism, and now we publish `relay-chat-stream-event` events from chat hub for other mains to consume, each main messaging their own connected clients belonging to the user (if any). As the AI message chunks are pretty small (~10 characters) this is quite spammy on the pubsub channel, but I'll have a chat with Cats team on whether we want to do some bundling of those chunks into larger chunks - this would add a bit of delay and make the streaming experience worse. At least locally this doesn't seem to be an issue, and local multi-main setup works pretty well.

On multi-main we persist the ongoing chat session's streaming state to Redis on `chat-session-store.service.ts`, so that any main can handle the reconnection regardless which one of them the user reconnecting connects to.

https://linear.app/n8n/issue/CHA-102/use-websockets-to-support-ai-replies-after-=100s-on-cloud

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
Cadiac, line 1: Changes to this file were almost fully from claude, so take them with a grain of salt - what we have here now seems to work though

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
```diff
@@ -903,6 +850,321 @@ export const useChatStore = defineStore(STORES.CHAT_HUB, () => {
 		return saved;
 	}
 
+	/**
+	 * Handle the beginning of a WebSocket stream
+	 */
+	function handleWebSocketStreamBegin(data: {
```
autologie: Is it possible to use interface in api-types?

Cadiac: probably, I can take a look to reduce the repetition

packages/@n8n/api-types/src/chat-hub.ts
```diff
@@ -509,3 +509,36 @@ export interface ChatHubModuleSettings {
...
+ * Request query parameters for reconnecting to a chat stream
+ */
+export class ChatReconnectRequest extends Z.class({
+	lastSequence: z.coerce.number().int().min(0).optional(),
+}) {}
+
+/**
+ * Response containing pending chunks for reconnection replay
+ */
+export interface ChatReconnectResponse {
```
autologie, line 532: How does this work when multiple messages are sent (e.g. web search) while the browser is disconnected?

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
```diff
@@ -56,18 +65,69 @@ import { useSettingsStore } from '@/app/stores/settings.store';
...
 const toast = useToast();
 const isMobileDevice = useMediaQuery(MOBILE_MEDIA_QUERY);
 const documentTitle = useDocumentTitle();
 const uiStore = useUIStore();
 const i18n = useI18n();
 
+// Initialize WebSocket push handler for chat streaming
+const chatPushHandler = useChatPushHandler();
+
+onBeforeMount(() => {
```
autologie, line 83: I feel these logic for managing lifecycle and events of websocket can be in useChatPushHandler

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
```diff
@@ -557,24 +481,38 @@ export const useChatStore = defineStore(STORES.CHAT_HUB, () => {
...
+			previousMessageId,
+			tools,
+			attachments,
+			agentName: agent.name,
+			timeZone: Intl.DateTimeFormat().resolvedOptions().timeZone,
+		};
+
+		try {
+			// Add human message to the conversation immediately
+			addMessage(sessionId, createHumanMessageFromStreamingState(streaming.value));
```
autologie: Adding message before OK response causes "The previous message does not exist in the session" that we've already fixed, can we do this on "begin" message?

packages/@n8n/api-types/src/chat-hub.ts
Cadiac: `ChatStreamStateService` only holds the current message being generated - as soon as a message becomes complete it gets dropped from the buffer on `endExecution`. As such, the reconnect only ever returns chunks from the current message being generated.

Every time a message is fully complete we write it to database, so when user reloads the session their get conversation by ID request returns all complete messages up to that point, and the reconnect message is the one that finds the ongoing message, if there's one.

so if AI has responded with multiple messages while the user has been away the full messages are loaded via the `GET /conversations/:sessionId` request and if there's still something being streamed that gets restored via the reconnect, and new data after that is pushed through the pushConnect socket.

There could be some race conditions here, maybe they might miss a chunk or two, but this seems to be working pretty well at least locally.

autologie: True, thanks for the explanation

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
Cadiac: I assume you're worried of the case where starting the message then fails, but we did add the human message? I'll see if I could cleanly change that.

autologie: Yes. Haven't confirmed, but maybe as simple as moving sendMessageApi before addMessage

packages/frontend/editor-ui/src/features/ai/chatHub/ChatView.vue
Cadiac: Refactored this logic to be contained within useChatPushHandler. Now this just calls initialize and terminate when mounted, and the composable takes care of the rest

Cadiac: Or well, this still has the watch logic to load the initial state after browser refresh, as ChatView holds the `isNewSession` logic etc - moving that to within the composable seemed harder to do.

autologie (dismissed):

autologie (dismissed):

autologie (dismissed):

autologie (approved):

PR #24998, opened 2026-01-28
feat: Add option to select eval suite on manual workflow dispatch

aalises:
## Summary
- Adds `eval_type` input parameter to the AI evals workflow for manual triggers   
- Allows selecting which evaluation suite to run: `both` (default), `spec` only, or `matrix` only                                                                  
- Push events to master continue to run both suites automatically


burivuhster (approved):
Thanks!

PR #25010, opened 2026-01-28
feat(editor): Add base for the setup panel feature (no-changelog)

MiloradFilipovic:
## Summary
- Add new feature flag that we'll temporarily use to toggle the feature while in development
- Introduce new component, `FocusSidebar` with tab support
- Move existing Focus panel content to a new tab
- Add placeholder tab for the Setup panel

Closes ADO-4715
Closes ADO-4716

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/config/src/configs/setup-panel.config.ts
```diff
@@ -0,0 +1,7 @@
+import { Config, Env } from '../decorators';
+
+@Config
+export class SetupPanelConfig {
+	@Env('N8N_SETUP_PANEL_ENABLED')
```
scdekov: Isn't a feature flag more appropriate here instead of env variable? I assume once we're ready with the feature we'd gradually enable this/ab test it which would require ff.

packages/frontend/editor-ui/src/app/components/FocusPanel.vue
```diff
@@ -396,6 +409,15 @@ watch([() => focusPanelStore.lastFocusTimestamp, () => expressionModeEnabled.val
 	focusWithDelay(),
 );
 
+watch(
+	() => resolvedParameter.value,
+	(newValue, oldValue) => {
+		if (newValue && newValue !== oldValue) {
```
scdekov: I'm not sure why is this needed. Should we add a comment explaining?

packages/frontend/editor-ui/src/features/setupPanel/components/SetupPanelTabs.vue
```diff
@@ -0,0 +1,60 @@
...
+			size="small"
+			@update:model-value="onTabSelected"
+		/>
+	</div>
+</template>
+
+<style module lang="scss">
+.container {
+	display: flex;
+	padding: var(--spacing--3xs);
```
scdekov: The padding of the tabs(may not be this line only) seems a little off. Maybe add a little more on top and reduce on the left?

Image

design:

Image

packages/frontend/editor-ui/src/features/setupPanel/components/FocusSidebarTabs.vue
```diff
@@ -0,0 +1,60 @@
...
+import type { SetupPanelTabs } from '@/features/setupPanel/types';
+import type { TabOptions } from '@n8n/design-system';
+import { N8nTabs } from '@n8n/design-system';
+
+const i18n = useI18n();
+
+const props = withDefaults(
+	defineProps<{
+		modelValue?: SetupPanelTabs;
+		tabLabels?: {
```
scdekov, line 13: nit: maybe `labelOverrides` is a better name for this?

packages/frontend/editor-ui/src/app/components/FocusPanel.vue
```diff
@@ -49,6 +49,9 @@ import { useTelemetryContext } from '@/app/composables/useTelemetryContext';
 import { type ContextMenuAction } from '@/features/shared/contextMenu/composables/useContextMenuItems';
 import { type CanvasNode, CanvasNodeRenderType } from '@/features/workflows/canvas/canvas.types';
 import { useCanvasOperations } from '@/app/composables/useCanvasOperations';
+import SetupPanelTabs from '@/features/setupPanel/components/SetupPanelTabs.vue';
```
scdekov: I guess with adding the Setup panel it makes more sense to rather have a parent component eg ParentPannel(def not a good name) that is using the tabs and based on the selected tab renders either SetupPanel or Focus panel. It may be an overkill to already do this though especially since for a few more weeks or so for the users there will be only the SetupPanel so the current change is may be more appropriate for the transition  so up to you.

packages/frontend/editor-ui/src/app/components/FocusPanel.vue
```diff
@@ -480,199 +507,210 @@ function onRenameNode(value: string) {
...
-									{{ locale.baseText('nodeView.focusPanel.noExecutionData') }}
-								</N8nInfoTip>
+				<div v-if="isSetupPanelEnabled">
+					<SetupPanelTabs v-model="selectedTab" :tab-labels="tabLabels" />
+				</div>
+				<div v-if="showSetupPanel" :class="$style.content">
+					<SetupPanel />
+				</div>
+				<div v-else data-test-id="focus-panels" :class="$style.content">
+					<ExperimentalFocusPanelHeader
```
scdekov: This is just the old component with same params moved inside the new container, right? The diff is a little weird

scdekov (dismissed):

packages/@n8n/config/src/configs/setup-panel.config.ts
MiloradFilipovic: Good shout about the A/B test. I think we'll do that eventually so feature flag makes more sense 🙌
Will update

packages/frontend/editor-ui/src/app/components/FocusPanel.vue
MiloradFilipovic: OK, will document it. This will switch to 'focus' tab when a parameter is focused in NDV

packages/frontend/editor-ui/src/app/components/FocusPanel.vue
MiloradFilipovic: Yeah, that was my initial thought as well but since the `FocusPanel` has some logic to determine what is currently focused (single parameter/node) and that info is also needed for the tabs, I went with just adding tabs and `SetupPanel` inside it. TBH it did feel weird so glad to hear that I'm not wrong.
I guess it makes sense to re-organize the components and extract this logic to the store.

packages/frontend/editor-ui/src/app/components/FocusPanel.vue
MiloradFilipovic: Yeah, just moved to a conditionally rendered `div`

scdekov (dismissed):

scdekov (approved):

PR #25028, opened 2026-01-29
chore(core): Improve access controlled credentials and workflows fetching

guillaumejacquart:
## Summary

This PR changes the way we fetch access controlled workflows.
Before:
- get all allowed resource ids (from shared entity and project relations)
- fetch the resources according to those ids

After:
- build the query builder to fetch them all at once, introducing an access control object to help build the query

### Benchmark results
| Database | Scale | Old Approach | New Approach | Improvement | Status |
|----------|-------|--------------|--------------|-------------|---------|
| SQLite | 500 workflows | ~28ms | ~19ms | ~32% | ✅ Both work |
| SQLite | 100 workflows (paginated) | 3ms | 2ms | 33% | ✅ Both work |
| SQLite | 20,000 workflows | 137ms | 50ms | **63.5%** | ✅ Both work |
| SQLite | 130,000 workflows | ❌ FAILED* | 808ms | N/A | ⚠️ Only new works |
| PostgreSQL | 500 workflows | 26ms | 20ms | 23.1% | ✅ Both work |
| PostgreSQL | 20,000 workflows | 150ms | 106ms | 29.3% | ✅ Both work |
| PostgreSQL | 130,000 workflows | ❌ FAILED** | 342ms | N/A | ⚠️ Only new works |

https://linear.app/n8n/issue/IAM-159/unusable-workflow-and-credentials-listing-at-large-scale#comment-8f150397

## Review / Merge checklist

- [X] Tests included. 


afitzek (commented):
I do like the approach a lot, what would make this even nicer, would be if we also don't prefetch the roles, but join from the roles to the necessary scopes in the DB as well. These db queries are quite complex, I would really like some integration tests, that run these queries against both db systems.

BGZStephen (dismissed):
Some merge conflicts but otherwise looks good

BGZStephen (approved):

PR #25029, opened 2026-01-29
feat: Split cache manager between lifecyle and secrets access (no-change-log)

ireneea:
## Summary

- Extracts cache refresh logic into a dedicated `SecretsCacheRefreshService`
- Removes unused exports and proxy methods

## TODO

- Improve automated tests
- Remove deprecated `ExternalSecretsManager` code

Resolves LIGO-121

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
ireneea, line 1: The idea here is to make reduce this class to implement the IExternalSecretsManager and not deal with the whole lifecycle management and orchestration.

We've also mention that controllers will use dedicated services to manage providers connections. So that part will also be removed, but probably not in this PR

packages/core/src/execution-engine/external-secrets-proxy.ts
```diff
@@ -17,10 +16,6 @@ export class ExternalSecretsProxy {
 		this.manager = manager;
 	}
 
-	async update() {
```
konstantintieber, line 20: I just checked in my local copy and it seems this method was actually unused until now. Is that correct?

ireneea: Yes that's correct that's why I removed it 👍

packages/cli/src/modules/external-secrets.ee/secrets-cache-refresh.service.ee.ts
```diff
@@ -0,0 +1,182 @@
...
+import { ExternalSecretsConfig } from './external-secrets.config';
+import { ExternalSecretsProviderLifecycle } from './provider-lifecycle.service';
+import { ExternalSecretsProviderRegistry } from './provider-registry.service';
+import { ExternalSecretsRetryManager } from './retry-manager.service';
+import { ExternalSecretsSecretsCache } from './secrets-cache.service';
+import { ExternalSecretsSettingsStore } from './settings-store.service';
+import { SecretsProviderSettings } from './types';
+
+@Service()
+export class SecretsCacheRefresh {
```
geemanjs, line 18: I feel like this is a bit of a "god class". 

It's mixing quite a few responsibilities together - that in the initial implementation were separate:
1. Feature branching - based on the flag
2. Scheduling
3. Retrying failures
4. Settings decryption
5. Caching

I wonder if we should split it out into separate things. The "original" implementation did it quite well I always thought:
  - ExternalSecretsService - API layer with redaction logic            
  - ExternalSecretsManager - Orchestration and public API              
  - ExternalSecretsProviderLifecycle - Individual provider             
  init/connect/disconnect                                              
  - ExternalSecretsProviderRegistry - Simple collection management     
  - ExternalSecretsRetryManager - Backoff logic isolated  
  - ExternalSecretsSecretsCache - Refresh / read / write operations only

PR #25030, opened 2026-01-29
feat(editor): Add dynamic templates experiment (no-changelog)

scdekov:
## Summary

As part of https://www.notion.so/n8n/Dynamic-templates-API-2f55b6e0c94f80bd9b97c01ed33e21fa?source=copy_link

Fetch the templates that we show on the empty screen from a new service where they're more dynamically configured. As a next step we'll be updating this request to include user parameters which can be used to pick specific templates based on them.

TODO: make sure that the templates service url is finalized before merging this

https://linear.app/n8n/issue/ADO-4709/spike-api-to-return-a-list-of-templates

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/controllers/dynamic-templates.controller.ts
+++ b/packages/cli/src/controllers/dynamic-templates.controller.ts
@@ -1,0 +1,20 @@
+import type { AuthenticatedRequest } from '@n8n/db';
+import { Get, RestController } from '@n8n/decorators';
+
+import { InternalServerError } from '@/errors/response-errors/internal-server.error';
+import { DynamicTemplatesService } from '@/services/dynamic-templates.service';
+
+@RestController('/dynamic-templates')
+export class DynamicTemplatesController {
+	constructor(private readonly dynamicTemplatesService: DynamicTemplatesService) {}
+
+	@Get('/')
+	async getRecommendedTemplates(_req: AuthenticatedRequest) {
+		try {
+			const templates = await this.dynamicTemplatesService.fetchDynamicTemplates();
+			return { templates };
+		} catch {
+			throw new InternalServerError('Failed to fetch dynamic templates');
+		}
+	}
+}
--- a/packages/cli/src/controllers/dynamic-templates.controller.ts
+++ b/packages/cli/src/controllers/dynamic-templates.controller.ts
@@ -9,7 +9,7 @@ export class DynamicTemplatesController {
 	constructor(private readonly dynamicTemplatesService: DynamicTemplatesService) {}
 
 	@Get('/')
-	async getRecommendedTemplates(_req: AuthenticatedRequest) {
+	async get(_req: AuthenticatedRequest) {
 		try {
 			const templates = await this.dynamicTemplatesService.fetchDynamicTemplates();
 			return { templates };
--- a/packages/cli/src/services/dynamic-templates.service.ts
+++ b/packages/cli/src/services/dynamic-templates.service.ts
@@ -1,0 +1,26 @@
+import { Logger } from '@n8n/backend-common';
+import { Service } from '@n8n/di';
+import axios from 'axios';
+
+export const DYNAMIC_TEMPLATES_URL = 'https://dynamic-templates.n8n.io/templates';
+export const REQUEST_TIMEOUT_MS = 5000;
+
+type DynamicTemplate = Record<string, unknown>;
+
+@Service()
+export class DynamicTemplatesService {
+	constructor(private readonly logger: Logger) {}
+
+	async fetchDynamicTemplates(): Promise<DynamicTemplate[]> {
+		try {
+			const response = await axios.get<{ templates: DynamicTemplate[] }>(DYNAMIC_TEMPLATES_URL, {
+				headers: { 'Content-Type': 'application/json' },
+				timeout: REQUEST_TIMEOUT_MS,
+			});
+			return response.data.templates;
+		} catch (error) {
+			this.logger.error('Error fetching dynamic templates', { error });
+			throw error;
+		}
+	}
+}
--- a/packages/cli/src/services/dynamic-templates.service.ts
+++ b/packages/cli/src/services/dynamic-templates.service.ts
@@ -15,6 +15,9 @@ export class DynamicTemplatesService {
 	) {}
 
 	async fetchDynamicTemplates(): Promise<DynamicTemplate[]> {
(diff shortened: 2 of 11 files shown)


packages/cli/src/services/dynamic-templates.service.ts
```diff
@@ -0,0 +1,26 @@
+import { Logger } from '@n8n/backend-common';
+import { Service } from '@n8n/di';
+import axios from 'axios';
+
+export const DYNAMIC_TEMPLATES_URL = 'https://dynamic-templates.n8n.io/templates';
```
RicardoE105: We should set this via env variables.

packages/cli/src/services/dynamic-templates.service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type DynamicTemplate = Record<string, unknown>;
+
+@Service()
+export class DynamicTemplatesService {
+	constructor(private readonly logger: Logger) {}
+
+	async fetchDynamicTemplates(): Promise<DynamicTemplate[]> {
+		try {
+			const response = await axios.get<{ templates: DynamicTemplate[] }>(DYNAMIC_TEMPLATES_URL, {
```
RicardoE105: Maybe we could also set the default retry mechanism

packages/frontend/editor-ui/src/features/workflows/templates/recommendations/recommendedTemplates.store.ts
```diff
@@ -1,12 +1,15 @@
...
 import templateIds from './data/recommendedTemplateIds.json';
 import { useSettingsStore } from '@/app/stores/settings.store';
 import { useNodeTypesStore } from '@/app/stores/nodeTypes.store';
+import { usePostHog } from '@/app/stores/posthog.store';
+import { useRootStore } from '@n8n/stores/useRootStore';
 import type { ITemplatesWorkflowFull } from '@n8n/rest-api-client';
 import sampleSize from 'lodash/sampleSize';
+import { getDynamicRecommendedTemplates } from './dynamicTemplates.api';
 
 export const NUMBER_OF_TEMPLATES = 6;
```
RicardoE105, line 14: Do we need to hard coded this? should not the service determine how many templates it returns?

packages/frontend/editor-ui/src/features/workflows/templates/recommendations/recommendedTemplates.store.ts
```diff
@@ -49,6 +61,19 @@ export const useRecommendedTemplatesStore = defineStore('recommendedTemplates',
...
 
+		if (isDynamicTemplatesEnabled()) {
+			try {
+				const response = await getDynamicRecommendedTemplates(rootStore.restApiContext);
+				return response.templates
+					.map((template) => template.workflow)
+					.slice(0, NUMBER_OF_TEMPLATES);
+			} catch (error) {
+				// Fallback to existing behavior on error
+				console.warn('Dynamic templates failed, falling back to static IDs', error);
```
RicardoE105, line 72: logger instead of console

packages/frontend/editor-ui/src/features/workflows/templates/recommendations/recommendedTemplates.store.ts
```diff
@@ -49,6 +61,19 @@ export const useRecommendedTemplatesStore = defineStore('recommendedTemplates',
 	async function loadRecommendedTemplates(): Promise<ITemplatesWorkflowFull[]> {
 		await nodeTypesStore.loadNodeTypesIfNotLoaded();
```
RicardoE105, line 62: Are we loading every go to the home screen only? what if the during start up we do it, then going to the workflows screen will show the templates right way.

packages/frontend/editor-ui/src/features/workflows/templates/recommendations/recommendedTemplates.store.ts
scdekov: No I think we should keep this here to make sure that the selected number looks good with the component we're using. The service doesn't have to necessarily know how we're presenting the selected templates.

packages/cli/src/services/dynamic-templates.service.ts
scdekov: Just to better understand. I don't see this having to be changed for different use cases so the point of having it as a variable is that in the future if needed it will be possible to change the value for older n8n versions, right?

packages/frontend/editor-ui/src/features/workflows/templates/recommendations/recommendedTemplates.store.ts
scdekov: I don't think we have defined logger for the front end code

packages/frontend/editor-ui/src/features/workflows/templates/recommendations/recommendedTemplates.store.ts
scdekov: What do you mean by "during start up"? Rn we're loading this when mounting the first page that the user sees.

packages/cli/src/services/dynamic-templates.service.ts
scdekov: Axios doesn't have built in retry mechanism and from what I see we don't retry requests to the creators site neither. I think since that's not a critical functionality it's fine to keep it simple.

packages/cli/src/controllers/dynamic-templates.controller.ts
```diff
@@ -0,0 +1,20 @@
...
+@RestController('/dynamic-templates')
+export class DynamicTemplatesController {
+	constructor(private readonly dynamicTemplatesService: DynamicTemplatesService) {}
+
+	@Get('/')
+	async get(_req: AuthenticatedRequest) {
+		try {
+			const templates = await this.dynamicTemplatesService.fetchDynamicTemplates();
+			return { templates };
+		} catch {
```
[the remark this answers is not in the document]
scdekov: We're already logging the original error

packages/cli/src/services/dynamic-templates.service.ts
RicardoE105: It's more like a patter that we have in n8n. For example the templates url has never changed and we still set it as and env variables. For example
Embed patterns could server their own dynamic server that returns the templates that they want.

RicardoE105 (approved):

PR #25034, opened 2026-01-29
fix: Update stack cleanup command message to work from any directory

aalises:
## Summary
While spinning up the repo as my first-time doing it I got hit with this when copying the command that appears after Starting the n8n stack. 
Screenshot 2026-01-29 at 13 30 01

Making the message more explainable and adding a command that works in all the monorepo, for those who come after me:

- Updates the cleanup message in `n8n-start-stack.ts` to show the correct command that works from anywhere in the monorepo                            
- Changed from `pnpm stack:clean:all` to `pnpm --filter n8n-containers stack:clean:all`

### DIFF
--- a/packages/testing/containers/n8n-start-stack.ts
+++ b/packages/testing/containers/n8n-start-stack.ts
@@ -314,7 +314,9 @@ async function main() {
 
 		console.log('');
 		log.info('Containers are running in the background');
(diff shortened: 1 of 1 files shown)


shortstacked (approved):
Solid PR!

PR #25036, opened 2026-01-29
chore(core): Support cookie for auth token extraction in dynamic credential endpoints

afitzek:
## Summary

This PR refactors the dynamic credentials authentication system to support both bearer token and cookie-based authentication. Previously, the system only supported bearer token authentication. This change enables the system to:
1. Accept authentication via bearer tokens OR session cookies
2. Support an explicit authSource query parameter to specify which authentication method to use
3. Maintain backward compatibility with existing bearer token authentication

closes https://linear.app/n8n/issue/IAM-90/support-cookie-for-auth-token-extraction

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/auth/auth.service.ts
+++ b/packages/cli/src/auth/auth.service.ts
@@ -6,7 +6,7 @@ import type { AuthenticatedRequest, User } from '@n8n/db';
 import { GLOBAL_OWNER_ROLE, InvalidAuthTokenRepository, UserRepository } from '@n8n/db';
 import { Service } from '@n8n/di';
 import { createHash } from 'crypto';
-import type { NextFunction, Response } from 'express';
+import type { NextFunction, Request, Response } from 'express';
 import { JsonWebTokenError, TokenExpiredError } from 'jsonwebtoken';
 import type { StringValue as TimeUnitValue } from 'ms';
 
@@ -156,19 +156,28 @@ export class AuthService {
 		};
 	}
 
-	getCookieToken(req: AuthenticatedRequest) {
-		return req.cookies[AUTH_COOKIE_NAME];
+	getCookieToken(req: Request) {
+		// This models the behavior of
+		if (typeof req.cookies === 'object' && req.cookies !== null) {
+			const cookies = req.cookies as Record<string, string | undefined>;
+			return cookies[AUTH_COOKIE_NAME];
+		}
+		return undefined;
 	}
 
-	getBrowserId(req: AuthenticatedRequest) {
-		return req.browserId;
+	getBrowserId(req: Request) {
+		// This models the behavior of APIRequest type having an optional browserId property of type string
+		if ('browserId' in req && typeof req.browserId === 'string') {
+			return req.browserId;
+		}
+		return undefined;
 	}
 
-	getMethod(req: AuthenticatedRequest) {
+	getMethod(req: Request) {
 		return req.method;
 	}
 
-	getEndpoint(req: AuthenticatedRequest) {
+	getEndpoint(req: Request) {
 		return req.route ? `${req.baseUrl}${req.route.path}` : req.baseUrl;
 	}
 
--- a/packages/cli/src/auth/auth.service.ts
+++ b/packages/cli/src/auth/auth.service.ts
@@ -157,7 +157,7 @@ export class AuthService {
 	}
 
 	getCookieToken(req: Request) {
-		// This models the behavior of
+		// This models the behavior of an AuthenticatedRequest type having an optional cookies property of type Record<string, string>
 		if (typeof req.cookies === 'object' && req.cookies !== null) {
 			const cookies = req.cookies as Record<string, string | undefined>;
 			return cookies[AUTH_COOKIE_NAME];
--- a/packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/__tests__/n8n-credential-resolver.test.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/__tests__/n8n-credential-resolver.test.ts
@@ -226,26 +226,5 @@ describe('N8NCredentialResolver', () => {
 				expect(mockStorage.deleteAllCredentialData).toHaveBeenCalledWith(handle);
 			});
 		});
-
-		describe('validateIdentity', () => {
-			it('should validate JWT with browserId set to false', async () => {
-				const identity = 'jwt-token-to-validate';
-				const handle = testHelpers.createHandle({});
-
-				await resolver.validateIdentity(identity, handle);
-
-				// Verify N8NIdentifier was called with browserId=false
-				expect(mockIdentifier.resolve).toHaveBeenCalledWith(
-					{
-						identity: 'jwt-token-to-validate',
-						version: 1,
-						metadata: {
-							browserId: false,
-						},
-					},
-					{},
-				);
-			});
-		});
 	});
 });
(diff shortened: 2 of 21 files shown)


packages/cli/src/auth/auth.service.ts
```diff
@@ -156,19 +156,28 @@ export class AuthService {
 		};
 	}
 
-	getCookieToken(req: AuthenticatedRequest) {
-		return req.cookies[AUTH_COOKIE_NAME];
+	getCookieToken(req: Request) {
+		// This models the behavior of
```
cstuncsik: of what? ;)

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential-web.service.ts
```diff
@@ -0,0 +1,86 @@
...
+import { z } from 'zod';
+import { ICredentialContext } from 'n8n-workflow';
+import { Request } from 'express';
+import { UnauthenticatedError } from '@/errors/response-errors/unauthenticated.error';
+
+class AuthSourceQuerySchema extends Z.class({
+	authSource: z.enum(['bearer', 'cookie']).optional(),
+}) {}
+
+const BEARER_TOKEN_REGEX = /^[Bb][Ee][Aa][Rr][Ee][Rr]\s+(.+)$/;
```
cstuncsik, line 13: Just a nitpick on the regex, I think this would suffice as well
```suggestion
const BEARER_TOKEN_REGEX = /^bearer\s+(.+)$/i;
```

cstuncsik (dismissed):
Left comments about very minor things if you want to fix

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential-web.service.ts
afitzek: I think this would not match 'Bearer' for example

packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.controller.ts
```diff
@@ -113,10 +111,13 @@ export class DynamicCredentialsController {
 		this.dynamicCredentialCorsService.preflightHandler(req, res, ['post', 'options']);
 	}
 
-	@Post('/:id/authorize', { skipAuth: true, middlewares: getDynamicCredentialMiddlewares() })
+	@Post('/:id/authorize', {
+		allowUnauthenticated: true,
```
phyllis-noester, line 115: why do we introduce `allowUnauthenticated` instead of continuing to use `skipAuth`?

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -250,7 +251,11 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
 	getDynamicCredentialsEndpointsMiddleware() {
 		const { endpointAuthToken } = this.dynamicCredentialConfig;
 		if (!endpointAuthToken?.trim()) {
-			return (_req: Request, res: Response, _next: NextFunction) => {
+			return (req: AuthenticatedRequest, res: Response, next: NextFunction) => {
+				// If a user was authenticated for this request, we allow access irrelevant of the static authentication
```
phyllis-noester, line 255: It makes sense that we only need one of the other but IMO it's not great to do it implicitly

packages/cli/src/modules/dynamic-credentials.ee/dynamic-credentials.controller.ts
phyllis-noester: ah ok didn't know this was already implemented in `createAuthMiddleware`. So to summarize for my understanding: We do not skip auth, because we want to check whether the user is authenticated (as would be the case if we're using n8n resolver), but we don't throw an error if they aren't.
if they are authenticated via n8n we do not check static credential but if they aren't (e.g. in the bearer token scenario) we do evaluate whether the static auth token is set

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
phyllis-noester: I'm guessing once the feature goes to GA we would split these endpoints into internal API and public API for external use and use different auth on them. In that case, I think it's fine to keep this as is right now

afitzek: Ideally we would have these endpoints in the public API yes, but since the public API is still lacking the option to inject handlers from modules, we don't really have a nice way forward on this. Maybe we can improve the public API with something like this till we reach GA, otherwise we should move these handlers to the public API and proxy the calls into the module if it is loaded and licensed. 👍

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential.service.ts
```diff
@@ -250,7 +251,11 @@ export class DynamicCredentialService implements ICredentialResolutionProvider {
...
+				if (req.user) {
```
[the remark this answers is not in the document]
afitzek: This bypasses the static authentication for dynamic credential endpoints only, this is necessary, because we will use these endpoints from the chathub in n8n. In this case we don't want to enforce the static authentication for these endpoints.

packages/cli/src/modules/dynamic-credentials.ee/services/dynamic-credential-web.service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function getBearerToken(req: Request): string | null {
+	const headerValue = req.headers['authorization']?.toString();
```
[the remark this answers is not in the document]
afitzek: This is for the enduser token, you are mixing up things.

packages/cli/src/auth/auth.service.ts
```diff
@@ -156,19 +156,28 @@ export class AuthService {
...
+		if (typeof req.cookies === 'object' && req.cookies !== null) {
+			const cookies = req.cookies as Record<string, string | undefined>;
```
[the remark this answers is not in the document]
afitzek: The type check one line above is good enough for this case.

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/__tests__/n8n-credential-resolver.test.ts
```diff
@@ -226,26 +226,5 @@ describe('N8NCredentialResolver', () => {
...
-
-		describe('validateIdentity', () => {
-			it('should validate JWT with browserId set to false', async () => {
-				const identity = 'jwt-token-to-validate';
-				const handle = testHelpers.createHandle({});
-
-				await resolver.validateIdentity(identity, handle);
-
-				// Verify N8NIdentifier was called with browserId=false
-				expect(mockIdentifier.resolve).toHaveBeenCalledWith(
```
phyllis-noester, line 238: don't we still want to test that the funciotn is called correctly?

phyllis-noester (approved):

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/__tests__/n8n-credential-resolver.test.ts
afitzek: This test become not really relevant anymore, because we refactor the validateIdentity method, it just sends through the context anymore, and we checked here the we set the browser id in the method.

PR #25037, opened 2026-01-29
feat: Trigger validation service

michael-radency:
## Summary

This PR introduces a centralized trigger validation system that enforces constraints on specific trigger types during workflow execution and activation.

### New Trigger Configuration Constants

Added two constants in [`trigger-validation.service.ts`](packages/cli/src/workflows/trigger-validation.service.ts:12):

- **`TRIGGERS_PREVENT_TEST_WHILE_ACTIVE`** - A `Set<string>` defining trigger types that cannot run test/manual executions while the workflow is active. This addresses limitations in certain triggers (e.g., Kafka) where simultaneous production and test listeners cause conflicts.

- **`TRIGGERS_UNIQUE_PARAMETERS`** - A `Record<string, string[]>` mapping trigger types to parameter names that must be unique across all active workflows. This prevents resource conflicts (e.g., two active workflows cannot share the same Kafka consumer group ID).

### Validation Service

The `TriggerValidationService` provides two main validation methods:

1. **`validateManualExecution()`** - Called before test/manual runs to:
   - Block test executions for restricted triggers when the workflow is active
   - Check for parameter conflicts with other active workflows

2. **`validateWorkflowActivation()`** - Called before workflow activation to:
   - Ensure unique parameters don't conflict with already-active workflows

### Error Handling

- `SingleTriggerError` - Renamed `SingleWebhookTriggerError`, thrown when attempting to test a restricted trigger while the workflow is active
- `TriggerParameterConflictError` - Thrown when a parameter value conflicts with an active workflow

## Example Configuration

```typescript
// Triggers that cannot be tested while active
export const TRIGGERS_PREVENT_TEST_WHILE_ACTIVE: Set<string> = new Set([
  'n8n-nodes-base.kafkaTrigger',
]);

// Parameters that must be unique across active workflows
export const TRIGGERS_UNIQUE_PARAMETERS: Record<string, string[]> = {
  'n8n-nodes-base.kafkaTrigger': ['groupId'],
};
```

---

https://linear.app/n8n/issue/NODE-4323/trigger-validation-service

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/workflows/trigger-validation.service.ts
```diff
@@ -0,0 +1,213 @@
...
+ */
+export const TRIGGERS_PREVENT_TEST_WHILE_ACTIVE: Set<string> = new Set([
+	'n8n-nodes-base.kafkaTrigger',
+]);
+
+/**
+ * Parameters that must be unique across all active workflows, keyed by trigger type.
+ */
+export const TRIGGERS_UNIQUE_PARAMETERS: Record<string, string[]> = {
+	'n8n-nodes-base.kafkaTrigger': ['groupId'],
```
mfsiega: Strictly speaking we still can't prevent this, because we support expressions, right? But of course it helps if somebody does this accidentally.

packages/cli/src/workflows/workflow-execution.service.ts
```diff
@@ -106,6 +108,15 @@ export class WorkflowExecutionService {
 		// Check whether this workflow is active.
 		const workflowIsActive = await this.workflowRepository.isActive(payload.workflowData.id);
 
+		const triggerToStartFrom =
```
mfsiega: I think we probably want to move this lower, down around line 221? Since in between here and there we do some work to potentially choose the relevant trigger (e.g., if it's a partial execution to a destination node, we pick a trigger on the backend).

There's also the question of pinned data - if we have pinned data for the trigger we won't actually end up activating it. So in that case, I think we would want to allow it.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -967,6 +969,8 @@ export class ActiveWorkflowManager {
 			return false;
 		}
 
+		await this.triggerValidation.validateWorkflowActivation(workflow);
```
mfsiega: We have some existing validation in the `WorkflowService.activateWorkflow`, which I think is a slightly better place for this. Should it live there instead? (And maybe we can unify some of the logic a bit?)

mfsiega (commented):
I really like the general approach!

I do worry that there are a lot of tricky corner cases.

Haven't taken a full look line-by-line yet.

packages/cli/src/workflows/trigger-validation.service.ts
michael-radency: true, but I do think that this is unlikely, as it isfor triggers

packages/cli/src/workflows/trigger-validation.service.ts
```diff
@@ -0,0 +1,210 @@
...
+
+		if (!hasUniqueParams) return;
+
+		const activeWorkflowIds = this.activeWorkflows
+			.allActiveWorkflows()
+			.filter((id) => id !== workflowId);
+
+		if (!activeWorkflowIds.length) return;
+
+		const activeWorkflows = await this.workflowRepository.find({
```
mfsiega, line 69: Pulling back so many workflows can still be very expensive, especially without pagination - it all needs to get stored here in memory, and deserialized.

Now that https://github.com/n8n-io/n8n/pull/25431 is merged we build an index on the workflows by default. So we can join on that table to only retrieve the workflows that have the matching node type for our trigger. (I imagine we'd add a method in workflow-index.service.ts.)

michael-radency: @mfsiega 
but this will prevent as for backporting to v1, right?

mfsiega: We could backport all the necessary changes to v1, I think there were only 3-4 PRs that happened after v2.

Though I'm not sure a backport is worth it. 1.x is only supported for 3 months after the release of 2.0, which is about a month from now: https://blog.n8n.io/introducing-n8n-2-0/#faq-what-you-need-to-know.

michael-radency:
@mfsiega 
I'm closing this PR for now, as implementation might need to be revisited

PR #25038, opened 2026-01-29
fix(core): Support reconnecting on Redis failover

mfsiega:
## Summary

If the Redis server fails over, it will start to throw `READONLY` errors when you try to write because you're trying to write to a readonly server. In this case, the client should reconnect.

By default this doesn't happen, so we leave this disabled by default as well, but we expose an option to enable it. 

https://linear.app/n8n/issue/CAT-2240/n8n-fails-to-reconnect-after-redisvalkey-failover-in-multi-az-setup

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/config/src/configs/scaling-mode.config.ts
```diff
@@ -88,6 +88,10 @@ class RedisConfig {
 	/** TCP keep-alive interval in milliseconds. */
 	@Env('QUEUE_BULL_REDIS_KEEP_ALIVE_INTERVAL')
 	keepAliveInterval: number = 5000;
+
+	/** Whether to reconnect to Redis on READONLY errors i.e., failover events. */
+	@Env('QUEUE_BULL_REDIS_RECONNECT_ON_FAILOVER')
+	reconnectOnFailover: boolean = false;
```
ivov: Why are we not enabling this by default?

packages/cli/src/services/redis-client.service.ts
```diff
@@ -160,10 +169,21 @@ export class RedisClientService extends TypedEmitter<RedisEventMap> {
...
+			options.reconnectOnError = (redisErr: Error) => {
+				const targetError = 'READONLY';
+				if (redisErr.message.includes(targetError)) {
+					this.logger.warn('Reconnecting to Redis due to READONLY error (possible failover event)');
+					return true;
+				}
+				return false;
+			};
+		}
+
```
ivov, line 189: So if a write e.g. to enqueue a Bull job fails and we fail over that job is lost forever? Should we resend the failed write as well?

> This feature is useful when using Amazon ElastiCache instances with Auto-failover disabled. On these instances, test your reconnectOnError handler by manually promoting the replica node to the primary role using the AWS console. The following writes fail with the error READONLY. Using reconnectOnError, we can force the connection to reconnect on this error in order to connect to the new master. Furthermore, if the reconnectOnError returns 2, ioredis will resend the failed command after reconnecting.

https://github.com/redis/ioredis

packages/@n8n/config/src/configs/scaling-mode.config.ts
mfsiega: Generally just to follow the underlying ioredis defaults; but in this case it seems very harmless to enable this by default, so 👍

packages/cli/src/services/redis-client.service.ts
mfsiega: Nice! Can't hurt.

ivov (approved):

PR #25041, opened 2026-01-29
feat(core): Add default Postgres query timeout

ivov:
https://linear.app/n8n/issue/CAT-2241

### DIFF
--- a/packages/@n8n/db/src/connection/db-connection-options.ts
+++ b/packages/@n8n/db/src/connection/db-connection-options.ts
@@ -106,6 +106,8 @@ export class DbConnectionOptions {
 			};
 		}
 
+		const { statementTimeoutMs } = postgresConfig;
+
 		return {
 			type: 'postgres',
 			...this.getCommonOptions(),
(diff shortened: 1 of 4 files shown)


packages/@n8n/db/src/connection/db-connection-options.ts
```diff
@@ -117,6 +119,7 @@ export class DbConnectionOptions {
 			ssl,
 			extra: {
 				idleTimeoutMillis: postgresConfig.idleTimeoutMs,
+				...(statementTimeoutMs > 0 && { options: `-c statement_timeout=${statementTimeoutMs}` }),
```
tomi, line 122: `node-pg` client also has [`statement_timeout`](https://node-postgres.com/apis/client#new-client). I think that is the same thing? Pool [seems to pass](https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/index.js#L233) the options directly to the client so this could


```suggestion
				...(statementTimeoutMs > 0 && { statement_timeout: statementTimeoutMs }),
```

if I'm not mistaken

tomi (approved):

packages/@n8n/db/src/connection/db-connection-options.ts
ivov: Will keep in mind! It was in the merge queue so it automerged

PR #25049, opened 2026-01-29
feat(core): Add profiling support via sentry

tomi:
## Summary

Adds Sentry profiling support:
- Added tracesSampleRate and profilesSampleRate configuration options in @n8n/config
- Updated error reporter in core to conditionally enable profiling via @sentry/profiling-node
- Extended Sentry configuration in both Python and JavaScript task runners
- Created a new sentry catalog in pnpm-workspace for managing Sentry package versions
- Added tests for new profiling-related functionality

Example data:
https://n8nio.sentry.io/explore/traces/?project=4503924908883968&statsPeriod=6h

https://linear.app/n8n/issue/CAT-2242/add-sentry-profiling-support

## Review / Merge checklist

- [ ] Tests included. 


packages/core/src/errors/error-reporter.ts
```diff
@@ -123,13 +131,24 @@ export class ErrorReporter {
...
 			'LinkedErrors',
 			'OnUnhandledRejection',
 			'ContextLines',
-		];
+			'LinkedErrors',
+		]);
+
+		const isTracingEnabled = tracesSampleRate > 0;
+		if (isTracingEnabled) {
+			enabledIntegrations.add('Http').add('Postgres').add('Redis').add('Express');
```
ivov: Should adding Postgres and Redis be conditional on queue mode being enabled?

packages/core/src/errors/error-reporter.ts
```diff
@@ -123,13 +131,24 @@ export class ErrorReporter {
...
 
-		const enabledIntegrations = [
+		const enabledIntegrations = new Set([
 			'InboundFilters',
 			'FunctionToString',
 			'LinkedErrors',
 			'OnUnhandledRejection',
 			'ContextLines',
-		];
+			'LinkedErrors',
```
ivov: Dupe

packages/@n8n/task-runner-python/src/sentry.py
ivov, line 1: The JS runner warns if profiling is enabled but tracing is disabled. Should we do the same in the Py runner?

packages/@n8n/task-runner/src/config/sentry-config.ts
```diff
@@ -1,11 +1,29 @@
...
+	 * This determines what percentage of transactions are profiled.
+	 *
+	 * @default 0 (disabled)
+	 */
+	@Env('N8N_SENTRY_PROFILES_SAMPLE_RATE', sampleRateSchema)
+	profilesSampleRate: number = 0;
+
+	/**
+	 * Sample rate for Sentry traces (0.0 to 1.0).
+	 * This determines what percentage of transactions are profiled.
```
ivov, line 20:
```suggestion
	 * This determines what percentage of transactions are traced.
```

packages/core/src/errors/error-reporter.ts
```diff
@@ -140,16 +159,18 @@ export class ErrorReporter {
...
 
+		const profilingIntegration = isProfilingEnabled ? await this.getProfilingIntegration() : [];
+
 		init({
 			dsn,
 			release,
 			environment,
-			tracesSampleRate: inProduction ? 0.01 : 0,
 			serverName,
-			beforeBreadcrumb: () => null,
```
ivov, line 149: It's intended that we're enabling breadcrumbs now?

packages/core/src/errors/error-reporter.ts
```diff
@@ -140,16 +159,18 @@ export class ErrorReporter {
...
+			...(isTracingEnabled ? { tracesSampleRate } : {}),
```
ivov, line 182: Not the biggest thing, but this changes behavior because the default is 0

packages/core/src/errors/error-reporter.ts
tomi: I'll make it configurable

packages/core/src/errors/error-reporter.ts
tomi: Docs say
> If you want to disable tracing, you should not set tracesSampleRate at all. Setting it to 0 will not disable tracing

So we should omit it entirely if it's not enabled

packages/core/src/errors/error-reporter.ts
tomi: Yes. The tracing wouldn't work if we dropped all the breadcrumbs

ivov (approved):
🥇

PR #25051, opened 2026-01-29
feat(core): Introduce table to track actually published workflow versions (no-changelog)

mfsiega:
## Summary

Introduce a new table to track what version of a workflow is *actually* published.

This is part of the workflow publication project. A subsequent PR will update the various webhooks/triggers/etc to use this table.

Today: webhooks read from the `workflow_entity` table (with a caching layer). Triggers/pollers/timers hold the workflow definition in memory. This means that when a new version of a workflow is published, there's no synchronization between the various triggers - they switch from the old version to the new version gradually.

After this PR and follow-ups: every trigger will read from this table, with a caching layer in front of it. When a new version is published and we're ready to start executing it, we'll invalidate the cache and update this table.

https://linear.app/n8n/issue/CAT-2190

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/migrations/mysqldb/index.ts
```diff
@@ -263,4 +264,5 @@ export const mysqlMigrations: Migration[] = [
 	AddAgentIdForeignKeys1765886667897,
 	AddWorkflowVersionIdToExecutionData1765892199653,
 	AddPublishedVersionIdToWorkflowDependency1769000000000,
+	CreateWorkflowPublishedVersionTable1769698710000,
```
tomi: No need to support mysql anymore

tomi (dismissed):

tomi (approved):
🚀

PR #25070, opened 2026-01-30
feat(ai-builder): Add Workflow Context Tools for On-Demand Data Fetching

OlegIvaniv:
## Summary

This PR introduces a set of workflow context tools that allow the AI workflow builder agents to fetch workflow and execution data on-demand rather than including everything in the initial context.

### New Tools

Six new tools have been added to the AI workflow builder:

- **get_workflow_overview**: Returns a visual Mermaid diagram of the workflow structure or a text summary, with optional parameter inclusion
- **get_node_context**: Returns detailed information about a specific node including its configuration, parent/child connections, and execution data
- **get_execution_logs**: Returns execution run data and error information, optionally filtered to a specific node
- **get_execution_schema**: Returns the schema of execution data for nodes that have been executed

### Context Injection Changes

- Added `buildWorkflowIndicator()` function that provides a lightweight summary pointing agents to use the tools for details
- Added `buildConversationContext()` for including relevant previous conversation turns
- Added `buildExecutionStatus()` for summarizing execution state (success/error, data flow between nodes)
- The responder agent prompt was updated to explain the available tools to users

### Supporting Changes

- Added `truncateJson()` utility for safely truncating large JSON payloads in tool responses
- Enhanced mermaid utilities to support node classification (trigger, AI parent, AI subnode, regular)
- Added constants for execution data truncation limits and workflow indicator node limits
- Added shared test utilities in `test-utils.ts` for creating mock workflows, run data, and state

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/tools/get-node-context.tool.ts
```diff
@@ -0,0 +1,365 @@
...
+ * Classifies a node based on its type and connections
+ */
+function classifyNode(
+	nodeType: string,
+	hasAiInputs: boolean,
+	hasAiOutputs: boolean,
+): NodeClassification {
+	// Check if it's a trigger node
+	if (
+		nodeType.toLowerCase().includes('trigger') ||
```
aalises: This trigger node detection logic gets repeated a bit with slight variations, to consider extracting to a `isTriggerNode(nodeType: string): boolean` utility in a common helpers file or here

packages/@n8n/ai-workflow-builder.ee/src/utils/context-builders.ts
```diff
@@ -127,11 +299,187 @@ export function buildExecutionSchemaBlock(
...
+		totalItems += branchOutput?.length ?? 0;
+	}
+	return totalItems;
+}
+
+/**
+ * Build data flow string showing item counts per node
+ * Format: "Node1 (5 items) → Node2 (3 items) → Node3 (0 items)"
+ */
+function buildDataFlowString(runData: RunData, executedNodeNames: string[]): string {
```
aalises, line 333: For large workflows, maybe the string would be too long. It's alright but maybe for looong workflows we can summarise them or truncate them as we do with `MAX_WORKFLOW_INDICATOR_NODES`

aalises (dismissed):

packages/@n8n/ai-workflow-builder.ee/src/utils/context-builders.ts
OlegIvaniv: Even for 80 executed nodes this would still be under 1k tokens, IMO not worth the benefit to trim it.

packages/@n8n/ai-workflow-builder.ee/src/tools/get-node-context.tool.ts
OlegIvaniv: Good point, extracted into common utility.

aalises (approved):

PR #1340, opened 2021-01-14
Merge unify execution id + queue system branch + fixes

krynble:
This branch comprises of 2 previous branches + fixes to work properly.

More information was added to JIRA ticket 670.

PR #25073, opened 2026-01-30
perf(core): Minimize execution data fetching

ivov:
## Summary

In scaling mode, a main instance can experience event loop blocking for seconds when sync running `flatted.parse` on execution data at job completion. `flatted.parse` is ~50x slower than `JSON.parse` and execution data can be hundreds of MB.

In the current flow:

1. Worker executes workflow, writes execution data to DB
2. Worker signals job completion via Bull
3. Main receives `job.finished()` notification
4. Main fetches full execution data (blocking parse)
5. Main runs `workflowExecuteAfter` hooks
6. Main resolves `postExecutePromise` for listeners

But, it is rare for main to actually need the entire execution data for post-execution handling. And a similar issue happens in the worker, where we fetch execution data immediatelyafter execution completion, despite already having the result in memory. Been this way ever since queue mode was introduced at #1340.

This PR adds a flag so the worker skips fetching and sends a handful of props via a Bull job progress message, so main can in the majority of cases read those props instead of fetching the full execution data. There is still a minority of cases when we can't skip fetching, to optimize in future.

n/a

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/scaling/scaling.service.ts
+++ b/packages/cli/src/scaling/scaling.service.ts
@@ -350,7 +350,7 @@ export class ScalingService {
 					 * because `removeOnComplete: true` prevents `job.finished()`
 					 * from returning a value that is no longer in Redis.
 					 */
-					if ('version' in msg && msg.version === 2) {
+					if (msg.version === 2) {
 						this.jobResults.set(msg.executionId, {
 							success: msg.success,
 							error: msg.error,
--- a/packages/cli/src/scaling/scaling.service.ts
+++ b/packages/cli/src/scaling/scaling.service.ts
@@ -30,6 +30,7 @@ import type {
 	JobOptions,
 	JobStatus,
 	JobId,
+	JobFinishedProps,
 	QueueRecoveryContext,
 	JobMessage,
 	JobFailedMessage,
@@ -38,6 +39,8 @@ import type {
 @Service()
 export class ScalingService {
 	private queue: JobQueue;
+
+	private jobResults = new Map<string, JobFinishedProps>();
 
 	constructor(
 		private readonly logger: Logger,
@@ -175,6 +178,13 @@ export class ScalingService {
 	// #endregion
 
 	// #region Jobs
+
+	/** Get and remove the result for a completed job. */
+	popJobResult(executionId: string): JobFinishedProps | undefined {
+		const result = this.jobResults.get(executionId);
+		this.jobResults.delete(executionId);
+		return result;
+	}
 
 	async getPendingJobCounts() {
 		const { active, waiting } = await this.queue.getJobCounts();
@@ -334,6 +344,22 @@ export class ScalingService {
 							statusCode: 500,
 						});
 					}
+
+					/**
+					 * We track the result received via `job-finished` message,
+					 * because `removeOnComplete: true` prevents `job.finished()`
+					 * from returning a value that is no longer in Redis.
+					 */
+					this.jobResults.set(msg.executionId, {
+						success: msg.success,
+						error: msg.error,
+						status: msg.status,
+						lastNodeExecuted: msg.lastNodeExecuted,
+						usedDynamicCredentials: msg.usedDynamicCredentials,
+						metadata: msg.metadata,
+						startedAt: new Date(msg.startedAt),
+						stoppedAt: new Date(msg.stoppedAt),
+					});
 
 					this.logger.info(`Execution ${msg.executionId} (job ${jobId}) finished`, {
 						workerId: msg.workerId,
--- a/packages/cli/src/scaling/scaling.service.ts
+++ b/packages/cli/src/scaling/scaling.service.ts
@@ -350,16 +350,18 @@ export class ScalingService {
 					 * because `removeOnComplete: true` prevents `job.finished()`
 					 * from returning a value that is no longer in Redis.
 					 */
-					this.jobResults.set(msg.executionId, {
-						success: msg.success,
-						error: msg.error,
-						status: msg.status,
-						lastNodeExecuted: msg.lastNodeExecuted,
-						usedDynamicCredentials: msg.usedDynamicCredentials,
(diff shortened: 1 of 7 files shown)


packages/cli/src/webhooks/webhook-helpers.ts
```diff
@@ -631,6 +631,8 @@ export async function executeWebhook(
 			responsePromise,
 		);
 
+		Container.get(ActiveExecutions).setResponseMode(executionId, responseMode);
```
tomi, line 638: What's this for? 🤔 Can we document it

packages/cli/src/workflow-runner.ts
```diff
@@ -498,4 +533,17 @@ export class WorkflowRunner {
...
+	/**
+	 * Whether main must retrieve full execution data from the DB on job completion.
+	 */
+	private needsFullExecutionData(executionMode: WorkflowExecuteMode, executionId: string): boolean {
+		if (!process.env.N8N_MINIMIZE_EXECUTION_DATA_FETCHING) return true;
+
+		return (
+			executionMode === 'integrated' ||
+			this.activeExecutions.getResponseMode(executionId) === 'lastNode' ||
+			this.externalHooks.hasHook('workflow.postExecute')
```
tomi, line 548: Why these? Can we document them

packages/cli/src/scaling/scaling.service.ts
```diff
@@ -176,6 +179,13 @@ export class ScalingService {
 
 	// #region Jobs
 
+	/** Get and remove the result for a completed job. */
+	popJobResult(executionId: string): JobFinishedProps | undefined {
```
tomi, line 183: Do we ensure somewhere that this always gets popped? So we don't start accummulating data here

packages/cli/src/workflow-runner.ts
```diff
@@ -461,24 +463,57 @@ export class WorkflowRunner {
...
+						mode: fullExecutionData.mode,
+						startedAt: fullExecutionData.startedAt,
+						stoppedAt: fullExecutionData.stoppedAt,
+						status: fullExecutionData.status,
+						data: fullExecutionData.data,
+						jobId: job.id.toString(),
+						storedAt: fullExecutionData.storedAt,
+					};
+				} else {
+					runData = {
```
tomi, line 493: Could we have separate types for these so its easier to ensure down the line that we are not missing anything?

tomi (commented):
Couple questions & comments. Very difficult to assess the full impact of this, so nice that it's behind a flag 👍

packages/cli/src/scaling/scaling.types.ts
```diff
@@ -52,13 +52,23 @@ export type RespondToWebhookMessage = {
...
+	status: ExecutionStatus;
+	lastNodeExecuted?: string;
+	usedDynamicCredentials?: boolean;
+	metadata?: Record<string, string>;
+	startedAt: Date;
+	stoppedAt: Date;
+};
+
+/** Message sent by worker to main to report a job has finished. */
 export type JobFinishedMessage = {
```
tomi: I think we should also version these messages so the mains can handle either version coming in. That should make the transition easier, as workers might be running a different version than mains momentarily

packages/cli/src/workflow-runner.ts
ivov: I don't follow, why separate types? It's one type and `tsc` already checks both branches against `IRun`.

packages/cli/src/scaling/scaling.types.ts
ivov: The new `JobFinishedMessage` is a superset of the old one so backwards compatible.

packages/cli/src/scaling/scaling.service.ts
ivov: The catch block after `job.finished()` doesn't return after `reject`, so execution always falls through to `popJobResult()` on both success and error paths. I have my [doubts](https://n8nio.slack.com/archives/C0AA207KJBH/p1769767938287129) that this fallthrough is incorrect, but I'd rather not change control flow in this PR.

packages/cli/src/scaling/scaling.types.ts
tomi: Yes, but if main is running with this flag enabled, and workers are running on an older version or this flag disabled, main will just blindly assume that the message contains all the new fields, when it doesn't. Hence, it would be better to version the message to avoid issues like this.

packages/cli/src/workflow-runner.ts
tomi: They have the same type, but they are not actually the same one because all the data has been stripped away from the other one. Basically everything downstream assumes that you have the full data available, where as that might not be the case anymore. It would be more safe to distinguish these two different states/types, and ensure all downstream functions behave correctly when passed either one of these

packages/cli/src/scaling/scaling.types.ts
```diff
@@ -64,8 +64,19 @@ export type JobFinishedProps = {
...
-export type JobFinishedMessage = {
+export type JobFinishedMessage = JobFinishedMessageV1 | JobFinishedMessageV2;
+
+/** @deprecated Old format without execution result details. */
+type JobFinishedMessageV1 = {
+	kind: 'job-finished';
+	executionId: string;
+	workerId: string;
+	success: boolean;
+};
```
tomi, line 76: Could we get rid of the `'version' in message` check if we added `version?: undefined` here?

tomi (dismissed):
One more comment but otherwise LGTM 🚀

tomi (approved):
🚀

packages/cli/src/workflow-runner.ts
ivov: Discussed internally, given current `IRun` usage this would mean a massive refactor of downstream consumers, so we'll test it out via flag instead.

PR #25078, opened 2026-01-30
chore(core): Add auth handler registry to reverse control of the ldap call

guillaumejacquart:
## Summary

Introduce an auth handler registry to avoid manually calling the ldap module. The ldap service instead registers as an auth handler to be called by the auth controller

https://linear.app/n8n/issue/IAM-99/improve-auth-providers-core-coupling

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/auth/auth-handler.registry.ts
+++ b/packages/cli/src/auth/auth-handler.registry.ts
@@ -4,6 +4,10 @@ import type { AuthType, IAuthHandler, IPasswordAuthHandler } from '@n8n/decorato
 import { AuthHandlerEntryMetadata } from '@n8n/decorators';
 import { Container, Service } from '@n8n/di';
 
+function isUserHandler(handler: IAuthHandler): handler is IAuthHandler<User> {
+	return handler.userClass === User;
+}
+
 /**
  * Registry service for discovering, instantiating, and managing auth handler implementations.
  * Automatically discovers all classes decorated with @AuthHandler() and makes them available by name.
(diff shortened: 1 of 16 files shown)


packages/cli/src/auth/auth-handler.registry.ts
```diff
@@ -0,0 +1,27 @@
+import type { User } from '@n8n/db';
+import { Service } from '@n8n/di';
+
+export interface AuthHandler {
+	handleLogin(loginId: string, password: string): Promise<User | undefined>;
```
afitzek: This is very specific to a password based authentication. I would suggest to either rename the AuthHandler interface to a more password specific auth handler interface, or we think about different mechanisms of credential transportation, that these handlers can perform, or maybe we hand over the whole request object, so that the auth handler can extract the credentials themself? I'm thinking in terms of login through email, or other password less authentication mechanism.

I'm not sure if this is necessary, just something that I want to bring up.

packages/cli/src/auth/auth-handler.registry.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/**
+ * Registry for authentication handlers. Allows modules to register
+ * themselves dynamically without creating direct dependencies.
+ */
+@Service()
+export class AuthHandlerRegistry {
```
afitzek, line 16: Would it make sense to have this registry in the decorator package and use decorators to register handlers? Similar to extractors, or resolver types, etc.

This would also allow to "build" auth handlers outside of the cli package.

packages/cli/src/auth/auth-handler.registry.ts
guillaumejacquart: Good point, I adjusted by adding a type to the auth handler metadata, for now only 'password'

packages/cli/src/auth/auth-handler.registry.ts
guillaumejacquart: Good point, i changed it

afitzek (approved):
Thank you for addressing the comments, looks great. 🚀

PR #25086, opened 2026-01-30
fix(core): Add tokensValidAfter to invalidate all tokens emitted before mfa toggling

guillaumejacquart:
## Summary

Add a new `tokensValidAfter` field on user, and set it to now when toggle MFA.
This timestamp is used to invalidate all previously emitted auth cookie by comparing with iat of the token.

https://linear.app/n8n/issue/IAM-164/disabling-2fa-does-not-invalidate-existing-sessions-across-browsers

## Review / Merge checklist

- [X] Tests included. 


afitzek (approved):
🚀

BGZStephen:
I don't think it needs to be this complicated,

I think we simply need to add a new check to `packages/cli/src/auth/auth.service.ts` > `createAuthMiddleware`, there is currently code that checks the cookie:

* If MFA was used and the instances is configured for MFA, lets through
* If didn't use MFA, and the instances is configured for MFA, throws an error

I think we just need to add a check here for:

* If MFA was used, but the instance isn't configured, throw an error
* If MFA was not used, and the instance isn't MFA, let through (I assume this is already fine)

PR #25088, opened 2026-01-30
feat(Kafka Trigger Node): Refactoring and fixes

michael-radency:
## Summary

Refactoring and fixes for Kafka trigger node, base for v1.3
- fix: Check for `isRunning` and `isStale` to prevent issues during processing in a rebalance.
- fix: `runWithHeartbeat` added to periodically send a heartbeat during long-running executions to prevent session timeouts.
- fix: Manual/test executions will no longer wait for `donePromise`.
- fix: improvements for parameters descriptions

https://linear.app/n8n/issue/NODE-4238/kafka-enable-more-control-over-commit-timing-improve-ui-timeout

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Kafka/KafkaTrigger.node.ts
```diff
@@ -207,13 +188,13 @@ export class KafkaTrigger implements INodeType {
 						name: 'parallelProcessing',
 						type: 'boolean',
 						default: true,
+						description:
+							'Whether to process messages in parallel resolving offsets independently or in order resolving offsets after execution completion. In the manual mode, when execution started by clicking on Execute Workflow or Execute Step button, messages are processed in parallel resolving offsets immediately.',
 						displayOptions: {
-							hide: {
-								'@version': [1],
+							show: {
```
yehorkardash, line 307: This will stop the property showing in 1.2, however it was shown previously. Is it expected?

michael-radency: no, it should be included in v1.2
there was community PR that did add v1.2, so later changes had to updated to v1.3
thanks 🙏

packages/nodes-base/nodes/Kafka/utils.ts
```diff
@@ -0,0 +1,442 @@
...
+			ctx.logger.error(error.message, { error });
+			return { success: false };
+		} finally {
+			if (timeoutId) clearTimeout(timeoutId);
+		}
+	};
+}
+
+/**
+ * Determines auto-commit settings based on node version and offset resolution configuration
```
yehorkardash: outdated comment

packages/nodes-base/nodes/Kafka/utils.ts
```diff
@@ -0,0 +1,442 @@
...
+ * @returns Object with auto-commit configuration
+ */
+export function getAutoCommitSettings(options: KafkaTriggerOptions) {
+	const eachBatchAutoResolve = options.eachBatchAutoResolve ?? false;
+
+	const autoCommitInterval = options.autoCommitInterval ?? undefined;
+	const autoCommitThreshold = options.autoCommitThreshold ?? undefined;
+
+	return {
+		autoCommit: true,
```
yehorkardash, line 409: Small question, without autoCommitInterval or autoCommitThreshold autoCommit doesn't work even if set to true?

yehorkardash (commented):
Looks good, just a few small comments

packages/nodes-base/nodes/Kafka/utils.ts
michael-radency: `autoCommit: true` is the default setting in `kafkajs`.
`autoCommitInterval` and `autoCommitThreshold` can be used to configure related behavior, both default to `null`

Before the update, we used `eachMessage`, which automatically commits offsets at the end of each call.
After switching to `eachBatch`, offset committing is handled explicitly via `resolveOffset` and `commitOffsetsIfNecessary`. 

Option `eachBatchAutoResolve` when `true` will make `kafka` automatically resolve and commit

yehorkardash (approved):

PR #25089, opened 2026-01-30
perf(core): Optimize saving execution progress

ivov:
## Summary

When saving execution progress, after every node runs we fetch the full execution fromthe DB including an expensive `flatted.parse` then append the current node's data, then write it back. This is unnecessary because the in-memory `executionData` passed to the hook already contains all accumulated `runData`, since `WorkflowExecute` pushes `taskData` before calling `nodeExecuteAfter` in every code path.

On `master` we already trust in-memory state for `executionData.executionData` (queue stack, wait state), but inconsistently fetch `runData` from the DB. This inconsistency apparently dates back to #1340, when queue mode and execution progress saving were introduced together.

This PR removes the DB fetch and uses the in-memory state directly, replacing race condition guards with a conditional `UPDATE` which is atomic so eliminates the race window between fetch and update. Using in-memory state arguably is also more reliable in failure scenarios, because if a previous progress save fails, the next save recovers the lost data because in-memory state is cumulative, whereas the fetch-append approach would only see what the DB has.

n/a

## Review / Merge checklist

- [ ] Tests included. 


tomi (approved):
🚀

PR #25118, opened 2026-01-31
chore(core): Add tracing abstraction layer

tomi:
## Summary

Implements a tracing interface with no-op and Sentry implementations to enable tracing of different operations. Integrates tracing into polling triggers to capture performance metrics.

https://linear.app/n8n/issue/CAT-2256

## Review / Merge checklist

- [ ] Tests included. 


packages/core/src/errors/__tests__/sentry-tracing.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import { SentryTracing } from '../sentry-tracing';
+
+describe('SentryTracing', () => {
+	let mockSentry: jest.Mocked<typeof Sentry>;
+	let sentryTracing: SentryTracing;
+
+	beforeEach(() => {
+		mockSentry = {
+			startSpan: jest.fn(),
+		} as unknown as jest.Mocked<typeof Sentry>;
```
ivov: Might be cleaner with `mock` from `jest-mock-extended`

packages/core/src/observability/tracing/tracing.ts
```diff
@@ -0,0 +1,62 @@
...
+		id: 'n8n.node.id',
+		name: 'n8n.node.name',
+		type: 'n8n.node.type',
+		typeVersion: 'n8n.node.type_version',
+	},
+} as const;
+
+/**
+ * Class to instrument the application with tracing. This class is a
+ * singleton that delegates to the actual tracing implementation. The
```
ivov, line 46: Why not DI `@Service` instead of a singleton?

packages/core/src/observability/tracing/tracing.ts
ivov, line 1: Nit, tracing is observability, doesn't quite fit in `errors/` dir

packages/core/src/errors/tracing.ts
```diff
@@ -0,0 +1,62 @@
...
+import type Sentry from '@sentry/node';
+
+import { NoopTracing } from './noop-tracing';
+
+/**
+ * Interface for concrete tracing implementations
+ */
+export interface TracingInterface {
+	startSpan<T>(options: StartSpanOptions, spanCb: (span?: Sentry.Span) => Promise<T>): Promise<T>;
+}
```
ivov: This isn't really an abstraction over a tracing mechanism if we're coupled to Sentry types...

Also the span isn't really optional?

```suggestion
export interface Tracer {
	startSpan<T>(options: StartSpanOptions, spanCb: (span?: Sentry.Span) => Promise<T>): Promise<T>;
}
```

packages/core/src/execution-engine/active-workflows.ts
```diff
@@ -251,4 +232,57 @@ export class ActiveWorkflows {
...
+	 * Creates a function that executes the poll function for a given workflow
+	 * and node and triggers a workflow execution based on the output.
+	 */
+	private createPollExecuteFn(
+		workflow: Workflow,
+		node: INode,
+		pollFunctions: IPollFunctions,
+	): (testingTrigger?: boolean) => Promise<void> {
+		return async (testingTrigger = false) => {
+			return await Tracing.startSpan(
```
ivov: Should we mark the span as errored if the poll throws?

packages/core/src/execution-engine/active-workflows.ts
```diff
@@ -251,4 +232,57 @@ export class ActiveWorkflows {
...
+					name: 'Workflow Trigger Poll',
+					op: 'trigger.poll',
+					attributes: {
+						[Tracing.commonAttrs.workflow.id]: workflow.id,
+						[Tracing.commonAttrs.workflow.name]: workflow.name,
+						[Tracing.commonAttrs.node.id]: node.id,
+						[Tracing.commonAttrs.node.name]: node.name,
+						[Tracing.commonAttrs.node.type]: node.type,
+						[Tracing.commonAttrs.node.typeVersion]: node.typeVersion,
+					},
```
ivov, line 254: This will show up in lots of places as we add it. Maybe a method to encapsulate populating tracing metadata from workflow and node?

packages/core/src/errors/tracing.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const commonN8nAttributes = {
```
ivov:
```suggestion
const COMMON_TRACE_ATTRIBUTES = {
```

packages/core/src/observability/tracing/tracing.ts
tomi: I wanted to keep it similar to how Sentry exposes it but using a `@Service` would make more sense in our context. I'll change it

packages/core/src/errors/tracing.ts
tomi: Good point. I wanted to hide that it's based on Sentry while still not redefining everything. I'll define the types here as well so it's a better abstraction.

Span is optional because the noop implementation doesn't provide a span. We could implement "an empty Span" that implements the interface but is noop. That would keep the interface nicer but might cause unnecessary noop function calls. I changed it to use that model

packages/core/src/errors/error-reporter.ts
```diff
@@ -199,6 +205,8 @@ export class ErrorReporter {
 			],
 		});
 
+		this.tracing.setTracingImplementation(new SentryTracing(sentry));
```
ivov: Should this be conditional on traces sample rate > 0? Else Sentry will create the traces but not send them

ivov (commented):
One last one, I should've spotted this earlier

packages/core/src/errors/error-reporter.ts
tomi: Good point! Should be

ivov (approved):

PR #25146, opened 2026-02-02
fix(core): Revert ensuring execution ID in webhook response

ivov:
## Summary

Revert #24876 due to performance regression. Arguably for correctness we should return 200 only after the execution embryo has been persisted, but given that users are now used to the existing throughput, given the current state of the webhook handling logic making this hard to refactor, and given that users have a workaround with Respond to Webhook node for `$execution.id`, the value of the fix for an edge case is not worth the loss in throughput for everyone.

## Review / Merge checklist

- [ ] Tests included. 


mfsiega (approved):
LGTM as a pure rollback.

PR #25149, opened 2026-02-02
perf(core): Optimize access for webhooks and workflows

ivov:
## Summary

Two query paths eagerly load `projectRelations` (and in one case `user`) through the `shared` relation, but no consumer ever reads those fields, so we can trim relation depth to:

- remove 1 unneeded JOIN from the webhook handling hot path, and 
- remove 2 unneeded JOINs from 27 callsites across workflow CRUD, workflow history, workflow statistics, MCP, ChatHub, Public API, and bootup.

`live-webhooks.ts` → The onlyaccess to `shared` after the query is `workflowData.shared.find((s) => s.role === 'workflow:owner')?.project.id`. Reads `role` and `project.id`. Later `workflowData` is spread into `activeWorkflowData` and passed to `WebhookHelpers.executeWebhook`, which does not access `shared`.

`findWorkflowWithOptions` → Called only by `findWorkflowForUser`
- 23 don't access `.shared` at all. They use the workflow for `null` checks, reading `.id`, `.activeVersionId`, `.settings`, `.nodes`, etc.
- 2 pass it to `addOwnerAndSharings` (`workflows.controller.ts:241, 378`) which delegates to `addOwnedByAndSharedWith`. That function iterates `shared[]` reading only `role` and `project.{id, type, name, icon}`. The `.shared` field is then deleted from the response before returning to the client.
- 1 reads `.shared[].role` and `.shared[].projectId` (`workflows.controller.ts:687`) to find editor project IDs.
- 1 reads `.shared[].role` and `.shared[].project` (`workflow.service.ee.ts:354`) to find the owner project during transfer. Accesses `project` directly, never `project.projectRelations`.
0 of these 27 call sites access `projectRelations` or `user`.

n/a

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/repositories/shared-workflow.repository.ts
+++ b/packages/@n8n/db/src/repositories/shared-workflow.repository.ts
@@ -168,7 +168,7 @@ export class SharedWorkflowRepository extends Repository<SharedWorkflow> {
 			},
 			relations: {
 				workflow: {
-					shared: { project: { projectRelations: { user: true } } },
+					shared: { project: true },
 					tags: includeTags,
 					parentFolder: includeParentFolder,
 					activeVersion: includeActiveVersion ? { workflowPublishHistory: true } : false,
--- a/packages/cli/src/webhooks/live-webhooks.ts
+++ b/packages/cli/src/webhooks/live-webhooks.ts
@@ -98,7 +98,7 @@ export class LiveWebhooks implements IWebhookManager {
 			where: { id: webhook.workflowId },
 			relations: {
 				activeVersion: true,
-				shared: { project: true },
+				shared: true,
 			},
 		});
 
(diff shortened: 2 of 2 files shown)


packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -98,7 +98,7 @@ export class LiveWebhooks implements IWebhookManager {
 			where: { id: webhook.workflowId },
 			relations: {
 				activeVersion: true,
-				shared: { project: { projectRelations: true } },
+				shared: { project: true },
```
tomi: I don't think we even need the project since we need only the project ID, which is available on `shared`

packages/@n8n/db/src/repositories/shared-workflow.repository.ts
```diff
@@ -168,7 +168,7 @@ export class SharedWorkflowRepository extends Repository<SharedWorkflow> {
 			},
 			relations: {
 				workflow: {
-					shared: { project: { projectRelations: { user: true } } },
+					shared: { project: true },
```
tomi, line 171: It's hard  to reason the implications of this, as it's used it massive amount of places. I trust that you have assessed the implications of this.

tomi (commented):
👍

tomi (approved):
🚀

PR #25152, opened 2026-02-02
ci: Fix flaky folders test

ivov:
Fix flaky `FolderRepository` test by using explicit `updatedAt` timestamps instead of relying on sequential insert timing for sort order.
https://app.blacksmith.sh/n8n-io/runs/21582970475/jobs/62184350817?attempt=1&autoLogin=true&tab=tests&status=failed&test_id=c444ad50-5a6e-4ac4-ac79-759872be07c6


MiloradFilipovic (commented):
This looks good but I am still experiencing some flakiness in the `should select only id and name when specified` test. Looks like we are getting folders in reversed order to what's expected.

Should we maybe updated it also in the `beforeEach` for the `describe('select')` [section](https://github.com/n8n-io/n8n/blob/50706d82e1b097d9b4bcd3c8656db865f3be2c0e/packages/cli/src/databases/repositories/__tests__/folder.repository.integration.test.ts#L322) so test folders have expected `updatedAt`:

```ts
beforeEach(async () => {
	const parentFolder = await createFolder(project, {
		name: 'Parent Folder',
		updatedAt: DateTime.now().minus({ minutes: 1 }).toJSDate(),
	});
	const tag = await createTag({ name: 'test-tag' });
	testFolder = await createFolder(project, {
		name: 'Test Folder',
		parentFolder,
		tags: [tag],
		updatedAt: DateTime.now().toJSDate(),
	});
	await createWorkflow({ parentFolder: testFolder });
});
```

MiloradFilipovic (approved):
🛳️

PR #25170, opened 2026-02-02
feat: Disallow usage of unsupported protocols in oauth fields

BGZStephen:
## Summary

This PR blocks the usage of protocols other than http / https for the oauth credentials type to avoid js injection into urls.

image

IAM-210

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/credentials/credentials.service.ts
```diff
@@ -471,6 +473,11 @@ export class CredentialsService {
 			// @ts-ignore
 			updateData.data.oauthTokenData = decryptedData.oauthTokenData;
 		}
+
+		this.checkCredentialData(
+			updateData.type,
+			updateData.data as unknown as ICredentialDataDecryptedObject,
```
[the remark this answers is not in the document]
BGZStephen: This is the current accepted approach for interacting with this function

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -887,7 +894,35 @@ export class CredentialsService {
...
+
+	/**
+	 * Validates that OAuth credential URL fields (authUrl, accessTokenUrl, etc.) use http/https only.
+	 * No-op if the credential type is not OAuth1 or OAuth2 (including extended types).
+	 */
+	private validateOAuthCredentialUrls(type: string, data: ICredentialDataDecryptedObject) {
+		const parentTypes = this.credentialTypes.getParentTypes(type) ?? [];
+		const isOAuth2 = type === 'oAuth2Api' || parentTypes.includes('oAuth2Api');
+		const isOAuth1 = type === 'oAuth1Api' || parentTypes.includes('oAuth1Api');
+		if (isOAuth2) {
```
phyllis-noester, line 908: do we need to evaluate oauth1 and oauth2 separately?

afaik the function would also pass if we just iterated over const 
`oauthUrlFields = ['authUrl', 'accessTokenUrl',  'requestTokenUrl', 'serverUrl'] as const;`

regardless of oauth type since either requestTokenUrl or serverUrl would be undefined

BGZStephen: I see what you're saying, but I quite like the specificity of only targeting expected fields per auth type, it feels less prone to oddities and side effects.

packages/cli/src/credentials/credentials.service.ts
phyllis-noester, line 455: so basically, in case someone already has a credential that doesn't use http/https they could continue to use it, but they could not update it unless they change the url, right?

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -1057,6 +1057,33 @@ async function oAuthCredentialAuthorize() {
 		return;
 	}
 
+	if (url === undefined || url === '') {
```
phyllis-noester, line 1060: why do we check that separately rather than relying on the try/catch below?

phyllis-noester (approved):

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
BGZStephen: One of the unit tests was complaining about a type error, which was somewhat valid so I just added this in

PR #25183, opened 2026-02-02
feat(core): Add localstack for testing aws external secret provider

geemanjs:
- **feat(core): E2E test for configuring external secret**
- **feat(core): E2E test for configuring external secret**

## Summary

Adds localstack to E2E tests for testing with aws as an external secrets provider

LIGO-109

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
+++ b/packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
@@ -115,11 +115,6 @@ export class AwsSecretsManager extends SecretsProvider {
 			clientConfig.credentials = { accessKeyId, secretAccessKey };
 		}
 
-		// Support endpoint override for testing with LocalStack
-		if (process.env.AWS_ENDPOINT_URL) {
-			clientConfig.endpoint = process.env.AWS_ENDPOINT_URL;
-		}
-
 		const { SecretsManager } = await import('@aws-sdk/client-secrets-manager');
 		this.client = new SecretsManager(clientConfig);
 
--- a/packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
+++ b/packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
@@ -115,6 +115,11 @@ export class AwsSecretsManager extends SecretsProvider {
 			clientConfig.credentials = { accessKeyId, secretAccessKey };
 		}
 
+		// Support endpoint override for testing with LocalStack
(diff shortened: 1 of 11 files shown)


packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
```diff
@@ -115,6 +115,11 @@ export class AwsSecretsManager extends SecretsProvider {
 			clientConfig.credentials = { accessKeyId, secretAccessKey };
 		}
 
+		// Support endpoint override for testing with LocalStack
```
geemanjs: @shortstacked do we do anything special when we "only" want this block of code to execute in tests?

packages/testing/containers/services/localstack.ts
```diff
@@ -0,0 +1,266 @@
...
+
+	/**
+	 * Create a secret.
+	 * @param name - Secret name
+	 * @param value - Secret value (string or object that will be JSON-serialized)
+	 */
+	async createSecret(name: string, value: string | Record<string, unknown>): Promise<void> {
+		const secretString = typeof value === 'string' ? value : JSON.stringify(value);
+
+		const response = await this.makeRequest('CreateSecret', {
```
geemanjs: If we find ourselves using lots of AWS things - it may make sense to use the official aws client for this stuff

packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
shortstacked: Actually according the AWS SDK docs we probably don't need this call at all since the SDK respects the value if it's set, and it will only be set in test environments anyway: https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html

packages/testing/containers/services/localstack.ts
shortstacked: That's a fair point it would probably be easier to do it that way. Less custom code, and it matches expectations exactly.

geemanjs: Cool - I'll just do it right away.

packages/cli/src/modules/external-secrets.ee/providers/aws-secrets-manager.ts
geemanjs: Even better - removed and it seemed to work locally

konstantintieber (approved):
Here's to new kinds of integration tests for aws 🚀

PR #25184, opened 2026-02-02
feat: Selectively skip auth only for dynamic creds

BGZStephen:
## Summary

This PR fixes a loophole where static credentials would have their oauth callback checks disabled inadvertently for dynamic credentials. We now always skip auth checks for dynamic (as they have no user to check against) and allow the previous feature flag to work only for static credentials as expected

IAM-214

## Review / Merge checklist

- [ ] Tests included. 


guillaumejacquart (approved):
good catch !

PR #25203, opened 2026-02-03
fix(core): Auto set pairedItem when N input items create 1 output item

elsmr:
## Summary

When a node produces 1 item from many input items, the following nodes would not have paired item information, causing errors. The user would have to use `first()`, `last()` or an index to get data from previous nodes.

This PR automatically sets the pairedItem to the first item in this case, causing `$('Node').item` to work without errors.

It is a compromise because we cannot be sure which input item(s) produced the output item. When the node itself or the user in a Code node sets the pairedItem, that takes priority over this logic.

See test workflow in Linear

https://linear.app/n8n/issue/NODE-4157/bug-code-node-breaks-pair-item-information

## Review / Merge checklist

- [ ] Tests included. 


michael-radency (approved):
LGTM, working as expected 👍 

this is not related, but it's strange that I could not use `itemMatching` in this case by  providing index, it's always `[Execute previous nodes for preview]` or `[ERROR: Can’t determine which item to use]`

this is test on cloud:

Image

PR #25205, opened 2026-02-03
feat(core): Add secrets completions API endpoints for global and project scopes

ireneea:
## Summary

Adds new API endpoints for retrieving secret completions (secret names) for the external secrets feature:

- `GET /secret-providers/completions/secrets/global` - Returns secret names from global connections (connections not scoped to any project)
- `GET /secret-providers/completions/secrets/project/:projectId` - Returns secret names from connections scoped to a specific project

### Changes
- New `SecretProvidersCompletionsController` with middleware to check the `externalSecretsForProjects` feature flag
- Service methods `getGlobalCompletions()` and `getProjectCompletions(projectId)` in `SecretsProvidersConnectionsService`
- Repository methods `findGlobalConnections()` and `findByProjectId()` for filtering connections by scope
- Renamed API type from `AutocompletionSecretsResponse` to `SecretCompletionsResponse` for consistency
- Comprehensive integration tests covering authorization (owner/admin/member), enabled/disabled connections, and edge cases

### Response Format
```json
{
  "providerKey1": ["secret1", "secret2"],
  "providerKey2": ["secret3"]
}
```

https://linear.app/n8n/issue/LIGO-119

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -12,4 +12,60 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
...
+			return await this.find({
+				where: { isEnabled: true },
+				relations: ['projectAccess'],
+			});
+		}
+
+		// Return enabled connections that are NOT in the list of connections with access
+		return await this.find({
+			where: {
+				id: Not(In(connectionIds)),
```
konstantintieber: While the Not(In()) approach works, I think we could get by with just a single query to the DB to get the global connections like this:
```ts
async findGlobalConnections(): Promise<SecretsProviderConnection[]> {
    return await this.manager
        .createQueryBuilder(SecretsProviderConnection, 'connection')
        .leftJoin('connection.projectAccess', 'access')
        .where('access.secretsProviderConnectionId IS NULL')
        .andWhere('connection.isEnabled = :isEnabled', { isEnabled: true })
        .getMany();
}
```

Or is there a specific reason we should use the Not(in(ids)) approach here?

packages/@n8n/api-types/src/schemas/secrets-provider.schema.ts
```diff
@@ -91,10 +91,11 @@ export const secretProviderTypeResponseSchema = z.object({
 export type SecretProviderTypeResponse = z.infer<typeof secretProviderTypeResponseSchema>;
 
 /**
- * Autocompletion secrets - keyed by connectionName
+ * Secret completions response - maps providerKey to a list of secret names
+ * Example: { providerA: ["secret1", "secret2"], providerB: ["secret3"] }
```
geemanjs, line 95: I added a file for this
https://github.com/n8n-io/n8n/blob/master/packages/cli/src/modules/external-secrets.ee/secrets-providers.responses.ee.ts

Based on the comments on here https://github.com/n8n-io/n8n/pull/24862#discussion_r2736832482

Again, we probably should decide upon a consistent way of doing this and update the other side to do it. I am relatively unopinonated @konstantintieber do you have any preference?

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
geemanjs: I was thinking this exact same thing! Thanks for actually writing the query!

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -12,4 +12,60 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
...
+
+		// Return enabled connections with the projectAccess relation loaded
+		return await this.find({
+			where: {
+				id: In(connectionIds),
+				isEnabled: true,
+			},
+			relations: ['projectAccess'],
+		});
+	}
```
geemanjs: I think we can probably do this in one query too...

packages/cli/src/modules/external-secrets.ee/__tests__/secrets-providers-connections.service.ee.test.ts
```diff
@@ -87,4 +99,78 @@ describe('Secret Providers Connections Service', () => {
 			});
 		});
 	});
+
+	describe('getGlobalCompletions', () => {
```
geemanjs: This test feels like it crosses into different spaces. 

I think we should test at the unit level:
1. The mapping of the entity from database object to client side object
2. `groupByProviderKey`

The tests here are overly complex for what they actually validate

packages/cli/test/shared/external-secrets/utils.ts
```diff
@@ -262,3 +261,109 @@ export class TestFailProvider extends SecretsProvider {
 		return Object.keys(this.secrets);
 	}
 }
+
+/**
```
geemanjs: NIT - this comment feels relatively useless.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -120,6 +122,24 @@ export class SecretsProvidersConnectionsService {
...
+	async getProjectCompletions(projectId: string): Promise<SecretCompletionsResponse> {
+		const connections = await this.repository.findByProjectId(projectId);
+		return this.groupByProviderKey(connections);
+	}
+
+	private groupByProviderKey(connections: SecretsProviderConnection[]): SecretCompletionsResponse {
+		return connections.reduce((acc, connection) => {
+			const secretNames = this.externalSecretsManager.getSecretNames(connection.providerKey);
+			acc[connection.providerKey] = secretNames;
+			return acc;
```
konstantintieber: eslint is complaining about unsafe return on this `acc` variable here.

We could replace the type casting by passing a generic param to reduce like `reduce<SecretCompletionsResource>` (would still result in unsafe return)

With the help of claude I found this implementation which also fix the unsafe return:
```
return Object.fromEntries(
        connections.map((connection) => [
            connection.providerKey,
            this.externalSecretsManager.getSecretNames(connection.providerKey),
        ]),
    );
```

packages/@n8n/api-types/src/schemas/secrets-provider.schema.ts
konstantintieber: Hmm, I would need to understand into the intended purpose of `@n8n/api-types` package first.
Just judging by its name it sounds like a good place for these types...

packages/cli/test/shared/external-secrets/utils.ts
ireneea: removed 👍

packages/cli/src/modules/external-secrets.ee/__tests__/secrets-providers-connections.service.ee.test.ts
ireneea: - renamed `groupProviderKey` to `toSecretCompletionsResponse` and made it public
- removed unit tests for `getGlobalCompletions` and `getProjectCompletions`
- added unit tests for `toSecretCompletionsResponse`
- updated the controller to call  `toSecretCompletionsResponse`

Let me know what you think

packages/cli/src/modules/external-secrets.ee/secrets-providers-completions.controller.ee.ts
```diff
@@ -0,0 +1,49 @@
...
+	@Get('/secrets/global')
+	@GlobalScope('externalSecret:list')
+	async listGlobalSecrets(): Promise<SecretCompletionsResponse> {
+		this.logger.debug('Listing global secrets');
+		const connections = await this.connectionsService.getGlobalCompletions();
+		return this.connectionsService.toSecretCompletionsResponse(connections);
+	}
+
+	@Get('/secrets/project/:projectId')
+	@ProjectScope('externalSecret:list')
```
konstantintieber: As far as I know `externalSecret:list` only exists as global scope.
While we should aim to be able to give access to this endpoint via  a projectScope as well, I think we can rely on `GlobalScope` here while the project scope is not implemented yet.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
ireneea: I didn't see the eslint error, but changed it anyway 👍

konstantintieber: I must be using some extra picky eslint rules then :D

konstantintieber (dismissed):
Great stuff!

geemanjs (dismissed):

konstantintieber (approved):

PR #25210, opened 2026-02-03
perf(core): Use Map for O(n) execution deduplication

ivov:
Replace `acc.find()` with a Map in `reduceExecutionsWithAnnotations`, to bring down execution list deduplication from O(n^2) to O(n). This endpoint is hit on every visit to the executions page and on each poll interval.

### DIFF
--- a/packages/@n8n/db/src/repositories/execution.repository.ts
+++ b/packages/@n8n/db/src/repositories/execution.repository.ts
@@ -843,40 +843,32 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
 			}
 		>,
 	) {
-		return rawExecutionsWithTags.reduce(
-			(
-				acc,
-				{
-					annotation_id: _,
-					annotation_vote: vote,
-					annotation_tags_id: tagId,
-					annotation_tags_name: tagName,
-					...row
-				},
-			) => {
-				const existingExecution = acc.find((e) => e.id === row.id);
+		const map = new Map<string, ExecutionSummary>();
 
-				if (existingExecution) {
-					if (tagId) {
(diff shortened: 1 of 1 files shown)


packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -843,40 +843,32 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
-							vote,
-							tags: tagId ? [{ id: tagId, name: tagName }] : [],
-						},
-					});
-				}
-				return acc;
-			},
-			[] as ExecutionSummary[],
-		);
+		const map = new Map<string, ExecutionSummary>();
```
tomi: nit: Would be more descriptive
```suggestion
		const summariesById = new Map<string, ExecutionSummary>();
```

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -843,40 +843,32 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+			annotation_vote: vote,
+			annotation_tags_id: tagId,
+			annotation_tags_name: tagName,
+			...row
+		} of rawExecutionsWithTags) {
+			let execution = map.get(row.id);
+			if (!execution) {
+				execution = {
+					...row,
+					annotation: { vote, tags: tagId ? [{ id: tagId, name: tagName }] : [] },
```
tomi: nit: can we split this into multiple lines this is hard to follow

tomi (commented):
👍 couple minor comments

tomi (approved):
🚀

PR #25212, opened 2026-02-03
perf(core): Cache invalid auth tokens

ivov:
Eliminates DB round-trip on every authenticated request by caching invalidated tokens.

### DIFF
--- a/packages/cli/src/auth/auth.service.ts
+++ b/packages/cli/src/auth/auth.service.ts
@@ -99,13 +99,21 @@ export class AuthService {
 		return `invalid-auth-token:${token}`;
 	}
 
+	private async cacheInvalidatedToken(token: string, exp: number) {
+		const ttlMs = Math.max(0, exp * 1000 - Date.now());
+		if (ttlMs > 0) {
+			const cacheKey = this.cacheKeyForInvalidToken(token);
+			await this.cacheService.set(cacheKey, '1', ttlMs);
+		}
+	}
+
 	private async isTokenInvalid(token: string): Promise<boolean> {
 		const cacheKey = this.cacheKeyForInvalidToken(token);
 
 		try {
 			if ((await this.cacheService.get(cacheKey)) !== undefined) return true;
-		} catch {
-			// fall through
+		} catch (error) {
+			this.logger.warn('Failed to check invalid token cache', { error });
 		}
 
 		const isInvalidPerDb = await this.invalidAuthTokenRepository.existsBy({ token });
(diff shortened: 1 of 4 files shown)


packages/cli/src/auth/auth.service.ts
```diff
@@ -195,6 +227,12 @@ export class AuthService {
 					token,
 					expiresAt: new Date(exp * 1000),
 				});
+
+				const ttlMs = Math.max(0, exp * 1000 - Date.now());
+				if (ttlMs > 0) {
+					const cacheKey = this.cacheKeyForInvalidToken(token);
+					await this.cacheService.set(cacheKey, '1', ttlMs);
+				}
```
tomi: This could be its own method, e.g. `cacheInvalidatedToken`

packages/cli/src/auth/auth.service.ts
```diff
@@ -93,6 +95,36 @@ export class AuthService {
...
+		return `invalid-auth-token:${token}`;
+	}
+
+	private async isTokenInvalid(token: string): Promise<boolean> {
+		const cacheKey = this.cacheKeyForInvalidToken(token);
+
+		try {
+			if ((await this.cacheService.get(cacheKey)) !== undefined) return true;
+		} catch {
+			// fall through
```
tomi: Shouldn't we at least report this? How would we know if this fails for whatever reason?

packages/cli/src/auth/auth.service.ts
```diff
@@ -93,6 +95,36 @@ export class AuthService {
...
+
+		if (isInvalidPerDb) {
+			try {
+				const { exp } = this.jwtService.decode(token);
+				if (exp) {
+					const ttlMs = Math.max(0, exp * 1000 - Date.now());
+					if (ttlMs > 0) await this.cacheService.set(cacheKey, '1', ttlMs);
+				}
+			} catch {
+				// non-critical
```
tomi: Same here. Can hide bugs

tomi (commented):
Couple comments

packages/cli/src/auth/auth.service.ts
ivov: Realistically this can only fail on Redis down or network blip, so I decided to warn and move on.

packages/cli/src/auth/auth.service.ts
ivov: Realistically this can only fail on Redis down or network blip, so I decided to warn and move on.

tomi (commented):
I actually started thinking whether this change makes sense. Most of the time the token is valid. Meaning it hasn't expired and there is no row for it in the `invalid_auth_token` table. Before this PR we had to do single check from DB. Now we have to do two checks:
1. Cache which is a miss
2. DB which is also a miss

For the base scenario we are actually doing more work. If we wanted to get rid of that DB check we would have to cache also in the scenario where the token is valid, which brings its own issue of keeping the cache in sync

PR #25214, opened 2026-02-03
refactor(editor): Move `initState` from `useWorkflowHelpers` to `useWorkflowState` (no-changelog)

alexgrozav:
## Summary

Consolidate workflow initialization logic into the workflow state composable for better separation of concerns.

https://linear.app/n8n/issue/CAT-2262/move-initstate-from-useworkflowhelpers-to-useworkflowstate

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
@@ -132,7 +132,6 @@ import { useRoute, useRouter } from 'vue-router';
 import { useTemplatesStore } from '@/features/workflows/templates/templates.store';
 import { isValidNodeConnectionType } from '@/app/utils/typeGuards';
 import { useParentFolder } from '@/features/core/folders/composables/useParentFolder';
-import { useWorkflowState } from '@/app/composables/useWorkflowState';
 import { useClipboard } from '@vueuse/core';
 
 type AddNodeData = Partial<INodeUi> & {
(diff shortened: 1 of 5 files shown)


packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
```diff
@@ -837,43 +829,6 @@ export function useWorkflowHelpers() {
 		return obj;
 	}
 
-	async function updateWorkflow(
```
alexgrozav, line 863: This isn't used anywhere right now.

packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
```diff
@@ -2168,7 +2167,7 @@ export function useCanvasOperations() {
 	}
 
 	async function initializeWorkspace(data: IWorkflowDb) {
-		await workflowHelpers.initState(data, useWorkflowState());
+		workflowState.initState(data);
```
CharlieKolb, line 2170: I think this might be harmful currently, as `initializeWorkspace` is called from e.g. https://github.com/n8n-io/n8n/blob/31a11c07feec6810223f8c9ae0f1d1f1666dab77/packages/frontend/editor-ui/src/app/composables/usePushConnection/handlers/workflowActivated.ts#L26, which initializes `useCanvasOperations` outside of a top-level vue context, which will fail the inject call to create the `workflowState`

I do acknowledge we already have this issue further below in this function, but I assume those were just broken without affecting the rest of the function 🙃 

Also the events calling this are not fired on my local instance, I assume those are queue-mode exclusive?

Maybe we should really just make `injectWorkflowState()` return a fresh `useWorkflowState()` for now, these issues are so subtle

CharlieKolb: Alternatively doing `useWorkflowStae().initState(data);` in the meantime should avoid this, but I'm in favor of not actually injecting for now

PR #25216, opened 2026-02-03
fix(editor): Fix node suggestion in node creator to always be request node

scdekov:
## Summary

Decided to add both nodes as suggestions as both are likely to be useful on that step

https://linear.app/n8n/issue/ADO-4738/no-result-state-in-node-panel-suggesting-webhook-node

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/shared/nodeCreator/components/Modes/NodesMode.vue
+++ b/packages/frontend/editor-ui/src/features/shared/nodeCreator/components/Modes/NodesMode.vue
@@ -11,7 +11,6 @@ import type {
 import {
 	TRIGGER_NODE_CREATOR_VIEW,
 	HTTP_REQUEST_NODE_TYPE,
-	WEBHOOK_NODE_TYPE,
 	REGULAR_NODE_CREATOR_VIEW,
 	AI_NODE_CREATOR_VIEW,
 	AI_OTHERS_NODE_CREATOR_VIEW,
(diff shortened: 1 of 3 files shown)


packages/frontend/editor-ui/src/features/shared/nodeCreator/components/Modes/NodesMode.vue
```diff
@@ -11,8 +11,8 @@ import type {
 import {
 	TRIGGER_NODE_CREATOR_VIEW,
 	HTTP_REQUEST_NODE_TYPE,
-	WEBHOOK_NODE_TYPE,
 	REGULAR_NODE_CREATOR_VIEW,
+	WEBHOOK_NODE_TYPE,
```
RicardoE105, line 14: did we need the change in this file? 👁️

RicardoE105 (commented):
should we add a test for this?

RicardoE105 (commented):
should we add a tests for the changs? 🤔

packages/frontend/editor-ui/src/features/shared/nodeCreator/components/Modes/NodesMode.vue
scdekov: Good point, I did a change and then reverted something and missed to cleanup. Will do

RicardoE105 (approved):

PR #25228, opened 2026-02-03
feat: Add new scope for workflow:unpublish, assignable to personal projects and regular project roles

BGZStephen:
## Summary

This PR adds a workflow:unpublish scope, this scope is to support unpublishing of workflows from within personal projects which will no longer support workflow:publish, hence the split.

A migration has been added to ensure custom roles created previously with workflow:publish have workflow:unpublish, except for personal project roles.

IAM-127

image

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/migrations/common/1769900001000-AddWorkflowUnpublishScopeToCustomRoles.ts
```diff
@@ -0,0 +1,101 @@
...
+		const descriptionColumn = escape.columnName('description');
+
+		const roleTableName = escape.tableName('role');
+		const roleScopeTableName = escape.tableName('role_scope');
+		const roleSlugColumn = escape.columnName('slug');
+		const roleScopeRoleSlugColumn = escape.columnName('roleSlug');
+		const roleScopeScopeSlugColumn = escape.columnName('scopeSlug');
+
+		const dbTypeStr = dbType as string;
+		const useInsertIgnore = isMySQLOrMariaDB(dbTypeStr);
```
guillaumejacquart, line 36: Do we still need that ? I don't remember what we said about it

packages/@n8n/db/src/migrations/common/1769900001000-AddWorkflowUnpublishScopeToCustomRoles.ts
```diff
@@ -0,0 +1,101 @@
...
+         VALUES (:slug, :displayName, :description)
+         ON CONFLICT (${scopeSlugColumn}) DO NOTHING`;
+
+		await runQuery(insertScopeQuery, {
+			slug: 'workflow:unpublish',
+			displayName: 'Unpublish Workflow',
+			description: 'Allows unpublishing workflows.',
+		});
+
+		logger.debug('Ensured workflow:unpublish scope exists');
```
guillaumejacquart: should we keep that ?

packages/@n8n/db/src/migrations/common/1769900001000-AddWorkflowUnpublishScopeToCustomRoles.ts
```diff
@@ -0,0 +1,101 @@
...
+			WHERE ${roleScopeScopeSlugColumn} = :unpublishScope
+				AND ${roleScopeRoleSlugColumn} != :personalOwnerSlug
+		`;
+
+		await runQuery(deleteQuery, {
+			unpublishScope: 'workflow:unpublish',
+			personalOwnerSlug: PERSONAL_OWNER_ROLE_SLUG,
+		});
+
+		logger.info('Removed workflow:unpublish scope from roles except project:personalOwner');
```
guillaumejacquart: not sure I understand why we keep the scope for personal owners. 
shouldn't we even remove the scope from the scope table ?

packages/cli/src/services/__tests__/auth.roles.service.test.ts
```diff
@@ -50,7 +50,12 @@ describe('AuthRolesService', () => {
 	}
 
 	function createAllScopes(): Scope[] {
-		return ALL_SCOPES.map((slug) => createScope(slug));
+		const scopes = ALL_SCOPES.map((slug) => createScope(slug));
+		// Ensure workflow:unpublish exists so personal-owner publishing tests pass (scope is in RESOURCES; add if missing from ALL_SCOPES)
+		if (!scopes.some((s) => s.slug === 'workflow:unpublish')) {
```
guillaumejacquart: Shouldn't it exists on ALL_SCOPES ?

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -716,9 +716,12 @@ export class WorkflowService {
...
 		publicApi: boolean = false,
 	): Promise<WorkflowEntity> {
-		const workflow = await this.workflowFinderService.findWorkflowForUser(workflowId, user, [
-			'workflow:publish',
-		]);
+		const workflow = await this.workflowFinderService.findWorkflowForUser(
+			workflowId,
+			user,
+			['workflow:unpublish'],
+			{ includeActiveVersion: true },
```
guillaumejacquart: Could you explain why you added this includeActiveVersion param ?

guillaumejacquart (commented):
Looks good ! some comments and questions

packages/@n8n/db/src/migrations/common/1769900001000-AddWorkflowUnpublishScopeToCustomRoles.ts
BGZStephen: I'd rather add it and not need it, than not add it and need it :D

packages/cli/src/workflows/workflow.service.ts
BGZStephen: From what I gather, we should be adding includeActiveVersion to ensure the response includes publishing history which may be required for the front-end? 

Not sure if it's strictly required though as it wasn't before, let me remove it

packages/@n8n/db/src/migrations/common/1769900001000-AddWorkflowUnpublishScopeToCustomRoles.ts
BGZStephen: removed

packages/@n8n/db/src/migrations/common/1769900001000-AddWorkflowUnpublishScopeToCustomRoles.ts
BGZStephen: Also forced removal from personal scopes 👍

packages/cli/src/services/__tests__/auth.roles.service.test.ts
BGZStephen: It should, removed

guillaumejacquart (approved):
Thanks for addressing the comments

PR #25230, opened 2026-02-03
feat(core): Introduce named versions as a licensed feature

dariacodes:
## Summary

Put update version endpoint under license and update pruning logic.

Closes [ADO-4731](https://linear.app/n8n/issue/ADO-4731/introduce-named-versions-as-licensed-feature-and-update-pruning-logic)

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/workflows/workflow-history/workflow-history.controller.ts
```diff
@@ -75,6 +75,7 @@ export class WorkflowHistoryController {
 		}
 	}
 
+	@Licensed('feat:namedVersions')
```
CharlieKolb, line 78: This is a bit unfortunate if we add other fields to versions in the future without license, right?
Does this make it so that I can't add/change descriptions for published versions if I don't have named versions? I guess that's might already be the case tho?

CharlieKolb (approved):
Happy to merge as is if we don't do descriptions on active versions without named license

packages/cli/src/workflows/workflow-history/workflow-history.controller.ts
dariacodes: Yeah, I guess it's what it means. Users are not allowed to do PATCH requests on history versions. They can only update names and descriptions via publishing

CharlieKolb: Actually yeah, where do I see the description in general? I added a Description to this newer version
image

dariacodes: left upper corner of the canvas
and I also had a hard time finding it some time ago 😆 
image

CharlieKolb: Ohh dear :D Do we want to be able to edit these eventually? Seems weird not to... So this license check would get in the way? But I don't mind tackling that when we implement that, so feel free to merge 👍

dariacodes: Yeah, it'll become tricky if we want to let users edit this 😆 But now we don't and there is a walkaround with republishing

PR #25234, opened 2026-02-03
fix(core): Make restartExecutionId get passed on queue mode

Cadiac:
## Summary

Turns out we're not passing `restartExecutionId` to workers in queue mode, which causes workers to always emit `workflowExecuteBefore` at https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/workflow-execute.ts#L1439-L1441 even when workflow is resumed, unlike in single main mode.

Passing this in `JobData` and handling it in job-processor fixes this.

This is a prerequisite for new lifecycle hook `workflowExecuteResume`, that I need in Chat hub related code to monitor executions resuming.

https://n8nio.slack.com/archives/C03MZF137FV/p1770105386939879

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/__tests__/workflow-runner.test.ts
+++ b/packages/cli/src/__tests__/workflow-runner.test.ts
@@ -316,6 +316,39 @@ describe('enqueueExecution', () => {
 
 		expect(setupQueue).toHaveBeenCalledTimes(1);
 	});
+
+	it('should include restartExecutionId in job data when provided', async () => {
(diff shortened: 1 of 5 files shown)


ivov (approved):
Over time it'd be good to reduce rather than keep enlarging the job API, but I take it there is no reliable way to derive `restartExecutionId` from DB state.

PR #25242, opened 2026-02-03
feat(core): Populate workflow published version mapping

mfsiega:
## Summary

Implement the workflow published version mapping by adding a new `WorkflowPublishedVersion` entity and repository.

This allows tracking the currently published workflow version in production, distinct from `workflow_entity.activeVersionId` which tracks the intended published version. When the `useWorkflowPublicationService` feature flag is enabled, workflow activation and deactivation now transactionally update both tables to maintain consistency.

NOTE:
- This PR shouldn't affect anything running when the feature flag is disabled.
- The point of this PR is mostly to add the repository/etc. We populate the field here so that we can start reading from it in the next PRs and add e2e tests to verify it, but ultimately we'll write this field from the workflow publication outbox consumer.

https://linear.app/n8n/issue/CAT-2192/populate-the-new-workflow-id-to-published-version-mapping

## Review / Merge checklist

- [x] Tests included (integration tests for feature flag behavior)

### DIFF
--- a/packages/@n8n/db/src/entities/workflow-published-version.ts
+++ b/packages/@n8n/db/src/entities/workflow-published-version.ts
@@ -13,13 +13,13 @@ export class WorkflowPublishedVersion extends WithTimestamps {
 	publishedVersionId: string;
 
 	@ManyToOne('WorkflowEntity', {
-		onDelete: 'CASCADE',
+		onDelete: 'RESTRICT',
 	})
 	@JoinColumn({ name: 'workflowId' })
 	workflow: Relation<WorkflowEntity>;
 
 	@ManyToOne('WorkflowHistory', {
-		onDelete: 'CASCADE',
+		onDelete: 'RESTRICT',
 	})
 	@JoinColumn({ name: 'publishedVersionId', referencedColumnName: 'versionId' })
 	publishedVersion: Relation<WorkflowHistory>;
--- a/packages/@n8n/db/src/entities/workflow-published-version.ts
+++ b/packages/@n8n/db/src/entities/workflow-published-version.ts
@@ -1,0 +1,26 @@
+import { Column, Entity, JoinColumn, ManyToOne, PrimaryColumn, Relation } from '@n8n/typeorm';
+
+import { WithTimestamps } from './abstract-entity';
+import type { WorkflowEntity } from './workflow-entity';
+import type { WorkflowHistory } from './workflow-history';
+
+@Entity({ name: 'workflow_published_version' })
+export class WorkflowPublishedVersion extends WithTimestamps {
+	@PrimaryColumn({ type: 'varchar', length: 36 })
+	workflowId: string;
+
+	@Column({ type: 'varchar', length: 36 })
+	publishedVersionId: string;
+
+	@ManyToOne('WorkflowEntity', {
+		onDelete: 'CASCADE',
+	})
+	@JoinColumn({ name: 'workflowId' })
+	workflow: Relation<WorkflowEntity>;
+
+	@ManyToOne('WorkflowHistory', {
+		onDelete: 'CASCADE',
+	})
+	@JoinColumn({ name: 'publishedVersionId', referencedColumnName: 'versionId' })
+	publishedVersion: Relation<WorkflowHistory>;
+}
--- a/packages/@n8n/db/src/migrations/common/1769698710000-CreateWorkflowPublishedVersionTable.ts
+++ b/packages/@n8n/db/src/migrations/common/1769698710000-CreateWorkflowPublishedVersionTable.ts
@@ -17,12 +17,12 @@ export class CreateWorkflowPublishedVersionTable1769698710000 implements Reversi
 			.withForeignKey('workflowId', {
 				tableName: 'workflow_entity',
 				columnName: 'id',
-				onDelete: 'CASCADE',
+				onDelete: 'RESTRICT',
 			})
 			.withForeignKey('publishedVersionId', {
 				tableName: 'workflow_history',
 				columnName: 'versionId',
-				onDelete: 'CASCADE',
+				onDelete: 'RESTRICT',
 			}).withTimestamps;
 	}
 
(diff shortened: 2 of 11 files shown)


packages/cli/src/workflows/workflow.service.ts
```diff
@@ -15,6 +15,7 @@ import {
 	WorkflowTagMappingRepository,
 	SharedWorkflowRepository,
 	WorkflowRepository,
+	WorkflowPublishedVersionRepository,
```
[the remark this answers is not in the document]
mfsiega: In the future, we'll populate the mapping from a different service. That service doesn't exist yet - so we populate it here for now to support integration testing and similar.

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -542,6 +544,14 @@ export class WorkflowService {
 		} finally {
 			if (didPublish) {
 				assert(workflow.activeVersionId !== null);
+
```
mfsiega, line 547: Let's add a comment here explaining that this is temporary. In the future this mapping will be set by the workflow publication service. We set it here for now so we can write tests and work incrementally.

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -782,6 +792,10 @@ export class WorkflowService {
 
 		const deactivatedVersionId = workflow.activeVersionId;
 
+		if (this.globalConfig.workflows.useWorkflowPublicationService) {
```
mfsiega, line 799: Add a comment - this is temporary, we'll remove it when we populate this via the workflow publication service.

packages/cli/src/workflows/workflow.service.ts
mfsiega, line 547: Add a comment - this is temporary, we'll remove it when we add the workflow publication service.

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -877,6 +891,11 @@ export class WorkflowService {
 
 		if (workflow.activeVersionId !== null) {
 			await this.activeWorkflowManager.remove(workflowId);
+
+			if (this.globalConfig.workflows.useWorkflowPublicationService) {
```
mfsiega, line 900: Add a comment - we'll remove it when we introduce the workflow publication service.

packages/cli/test/integration/workflows/workflow.service.test.ts
```diff
@@ -483,6 +488,108 @@ describe('deactivateWorkflow()', () => {
...
+		test('should write to workflow_published_version on activation', async () => {
+			const owner = await createOwner();
+			const workflow = await createWorkflowWithHistory({}, owner);
+
+			await workflowService.activateWorkflow(owner, workflow.id);
+
+			const publishedVersion = await workflowPublishedVersionRepository.findOne({
+				where: { workflowId: workflow.id },
+			});
+			expect(publishedVersion).not.toBeNull();
```
mfsiega: This is probably unnecessary, since if it's null the assertion in the next line will fail anyway?

packages/cli/test/integration/workflows/workflow.service.test.ts
```diff
@@ -483,6 +488,108 @@ describe('deactivateWorkflow()', () => {
...
+
+			await workflowService.activateWorkflow(owner, workflow.id);
+
+			const publishedVersion = await workflowPublishedVersionRepository.findOne({
+				where: { workflowId: workflow.id },
+			});
+			expect(publishedVersion).toBeNull();
+		});
+
+		test('should not write to workflow_published_version on deactivation', async () => {
```
mfsiega, line 579: Is this test really verifying anything? Since we know from the previous test that we don't write on activation, then it seems like this is a bit redundant.

mfsiega: Note: it _does_ provide coverage for the deactivation path.

packages/@n8n/db/src/entities/workflow-published-version.ts
```diff
@@ -0,0 +1,26 @@
...
+	publishedVersionId: string;
+
+	@ManyToOne('WorkflowEntity', {
+		onDelete: 'CASCADE',
+	})
+	@JoinColumn({ name: 'workflowId' })
+	workflow: Relation<WorkflowEntity>;
+
+	@ManyToOne('WorkflowHistory', {
+		onDelete: 'CASCADE',
```
tomi: If this cascades on delete, it could lead to a situation where the published information is lost but the triggers are still kept running. And potentially leave webhooks registered in 3rd party services. Is this intentional or should we block deletes and require that the version or workflow is unpublished first?

packages/@n8n/db/src/repositories/workflow-published-version.repository.ts
```diff
@@ -0,0 +1,24 @@
...
+	async setPublishedVersion(workflowId: string, publishedVersionId: string): Promise<void> {
+		await this.upsert({ workflowId, publishedVersionId }, ['workflowId']);
+	}
+
+	async removePublishedVersion(workflowId: string): Promise<void> {
+		await this.delete({ workflowId });
+	}
+
+	async getPublishedVersionId(workflowId: string): Promise<string | null> {
+		const record = await this.findOne({ where: { workflowId } });
```
tomi: nit: Since we are only interested in the `publishedVersionId`, we should select only that and not entire row

tomi (dismissed):
Couple comments, but mostly LGTM 👍

packages/@n8n/db/src/migrations/common/1769698710000-CreateWorkflowPublishedVersionTable.ts
```diff
@@ -17,12 +17,12 @@ export class CreateWorkflowPublishedVersionTable1769698710000 implements Reversi
 			.withForeignKey('workflowId', {
 				tableName: 'workflow_entity',
 				columnName: 'id',
-				onDelete: 'CASCADE',
+				onDelete: 'RESTRICT',
```
tomi: I don't think we need to change this migration if we add a new one that does this?

mfsiega: Yeah I wasn't sure about this - seemed weird to first migrate it wrong, and then fix it - but also seems weird to migrate it right and then have a redundant migration.

Probably leaving the existing migration as-is is better, easier to understand what happened if you read the code.

packages/@n8n/db/src/entities/workflow-published-version.ts
mfsiega: Seems like this applies to the workflow table as well, right?

tomi (approved):
🚀

PR #25245, opened 2026-02-03
fix(core): Skip AuthRolesService.init() for workers to prevent scope deletion

afitzek:
## Summary

Workers should not modify auth roles/scopes as they may have different code versions. When a worker's ALL_SCOPES doesn't include a scope that exists in the database (e.g., workflow:publish), the syncScopes() method would incorrectly identify it as "obsolete" and delete it, stripping it from all roles.

This fixes a regression where workspace owners with global permissions could not publish workflows after upgrading to n8n 2.3.5+ when running in queue mode with workers.

Fixes: IAM-220

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/commands/base-command.ts
+++ b/packages/cli/src/commands/base-command.ts
@@ -136,7 +136,11 @@ export abstract class BaseCommand<F = never> {
 			);
 
 		// Initialize the auth roles service to make sure that roles are correctly setup for the instance.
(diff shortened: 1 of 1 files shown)


phyllis-noester (approved):
AAAAAAAH makes sense! thanks for fixing :)

BGZStephen (approved):

PR #25253, opened 2026-02-03
feat(core): Add log streaming for personal publishing restriction changes

guillaumejacquart:
## Summary

Add log streaming/audit events when admins enable or disable personal space publishing and sharing restrictions. The events capture admin changes and include user metadata for audit purposes.

- `n8n.audit.personal-publishing-restricted.enabled` - when publishing is restricted
- `n8n.audit.personal-publishing-restricted.disabled` - when publishing is allowed
- `n8n.audit.personal-sharing-restricted.enabled` - when sharing is restricted
- `n8n.audit.personal-sharing-restricted.disabled` - when sharing is allowed

## Related Linear tickets

https://linear.app/n8n/issue/IAM-123
https://linear.app/n8n/issue/IAM-119

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/events/relays/log-streaming.event-relay.ts
+++ b/packages/cli/src/events/relays/log-streaming.event-relay.ts
@@ -7,6 +7,7 @@ import { MessageEventBus } from '@/eventbus/message-event-bus/message-event-bus'
 import { EventService } from '@/events/event.service';
 import type { RelayEventMap, UserLike } from '@/events/maps/relay.event-map';
 import { EventRelay } from '@/events/relays/event-relay';
+import { assertNever } from '@/utils';
 
 type WorkflowExecutedEvent = RelayEventMap['workflow-executed'];
 type WorkflowExecutedEventWithUser = WorkflowExecutedEvent & { user: UserLike };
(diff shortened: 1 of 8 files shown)


packages/cli/src/events/relays/log-streaming.event-relay.ts
```diff
@@ -808,4 +809,31 @@ export class LogStreamingEventRelay extends EventRelay {
...
+
+	// #region Instance Policies
+
+	@Redactable()
+	private instancePoliciesUpdated({
+		user,
+		settingName,
+		value,
+	}: RelayEventMap['instance-policies-updated']) {
+		if (settingName === 'workflow_publishing') {
```
afitzek: This does not handle the settingName `2fa_enforcement`. What would be nice if we could enforce this to handle every possible type using typescript for example something like this:

```
switch(settingName) {
    case "workflow_publishing":
      return ...;
    case "workflow_sharing":
      return ...;
    case "2fa_enforcement":
      return ...;
    default:
      // This ensures exhaustiveness
      const _exhaustive: never = status;
      throw new Error(`Unhandled settingName: ${_exhaustive}`);
  }
}
```

This should be a compile time error if we extend the type of settingName to more options.

afitzek (commented):
I think we are missing to map one event type to an audit event, other than that it looks good.

packages/cli/src/events/relays/log-streaming.event-relay.ts
guillaumejacquart: Thanks, good point ! I updated it and used the existing `assertNever` utils function

afitzek (approved):
🚀

PR #25259, opened 2026-02-03
feat(core): Support disabling sharing from personal space

phyllis-noester:
## Summary

Allows admins & workspace owners to disable sharing of credentials & workflows

Screenshot 2026-02-03 at 19 16 18
Screenshot 2026-02-03 at 19 16 07

https://linear.app/n8n/issue/IAM-124/beandfe-support-disabling-sharing-anything-from-personal-space

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -103,8 +104,8 @@ export class AuthRolesService {
 		}
 	}
 
-	private async shouldPersonalOwnerHavePublishScope(): Promise<boolean> {
-		const setting = await this.settingsRepository.findByKey(PERSONAL_SPACE_PUBLISHING_SETTING_KEY);
+	private async shouldPersonalOwnerHaveSettingScope(settingKey: string): Promise<boolean> {
```
guillaumejacquart: Maybe we could type the setting key ?
like `settingKey: keyof typeof PERSONAL_SPACE_PUBLISHING_SETTING | PERSONAL_SPACE_SHARING_SETTING` (not sure about the ts syntax, but you get the idea)

packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -119,17 +120,20 @@ export class AuthRolesService {
...
 		// Special handling for project:personalOwner role
 		if (roleSlug === PROJECT_OWNER_ROLE_SLUG) {
-			const shouldHavePublish = await this.shouldPersonalOwnerHavePublishScope();
-			if (shouldHavePublish) {
-				this.logger.debug(
-					`Personal space publishing is enabled - adding workflow:publish scope to ${PROJECT_OWNER_ROLE_SLUG} role`,
-				);
-				return [...defaultScopes, 'workflow:publish'];
+			for (const setting of [PERSONAL_SPACE_PUBLISHING_SETTING, PERSONAL_SPACE_SHARING_SETTING]) {
+				const shouldHave = await this.shouldPersonalOwnerHaveSettingScope(setting.key);
```
guillaumejacquart: for the sake of performance, might be worth introducing a 
`await this.shouldPersonalOwnerhaveSettingsScope([PERSONAL_SPACE_PUBLISHING_SETTING, PERSONAL_SPACE_SHARING_SETTING])` (plural)

packages/cli/src/controllers/security-settings.controller.ts
```diff
@@ -13,9 +17,15 @@ export class SecuritySettingsController {
 	@Get('/')
 	async getSecuritySettings(_req: AuthenticatedRequest, _res: Response) {
 		const personalSpacePublishing =
-			await this.securitySettingsService.isPersonalSpacePublishingEnabled();
+			await this.securitySettingsService.isPersonalSpaceSettingEnabled(
+				PERSONAL_SPACE_PUBLISHING_SETTING,
+			);
 
-		return { personalSpacePublishing };
+		const personalSpaceSharing = await this.securitySettingsService.isPersonalSpaceSettingEnabled(
```
guillaumejacquart: same, might be worth adding something like
`await this.securitySettingsService.arePersonalSpaceSettingsEnabled()` (with the service knowing the specific keys) 
or 
`await this.securitySettingsService.arePersonalSpaceSettingsEnabled([PERSONAL_SPACE_PUBLISHING_SETTING, PERSONAL_SPACE_SHARING_SETTING])`
What do you think ?

packages/cli/src/controllers/security-settings.controller.ts
```diff
@@ -25,8 +35,22 @@ export class SecuritySettingsController {
 		_res: Response,
 		@Body dto: UpdateSecuritySettingsDto,
 	) {
-		await this.securitySettingsService.setPersonalSpacePublishing(dto.personalSpacePublishing);
+		const updatedSettings: Record<string, boolean> = {};
+		if (dto.personalSpacePublishing !== undefined) {
+			await this.securitySettingsService.setPersonalSpaceSetting(
```
guillaumejacquart, line 31: same here, introduce plural method ? 
also maybe as a single db query for atomic changes ?

packages/cli/src/dto/security-settings.dto.ts
```diff
@@ -2,5 +2,6 @@ import { z } from 'zod';
 import { Z } from 'zod-class';
 
 export class UpdateSecuritySettingsDto extends Z.class({
```
guillaumejacquart: I might have missed something, but I don't see the difference between this DTO and the one inside api-types package 🤔

packages/cli/src/services/security-settings.service.ts
```diff
@@ -13,25 +16,30 @@ export class SecuritySettingsService {
 		private readonly roleService: RoleService,
 	) {}
 
-	async setPersonalSpacePublishing(enabled: boolean): Promise<void> {
+	async setPersonalSpaceSetting(
+		setting: typeof PERSONAL_SPACE_PUBLISHING_SETTING | typeof PERSONAL_SPACE_SHARING_SETTING,
```
guillaumejacquart, line 20: ah, nice ! might not be worth doing so on the repo as I suggested before

packages/frontend/editor-ui/src/features/settings/security/SecuritySettings.vue
```diff
@@ -65,11 +37,78 @@ async function updatePublishingSetting(value: boolean) {
...
-		// Revert optimistic update on error
-		state.value = !value;
+		if (state.value) {
+			state.value = { ...state.value, personalSpacePublishing: !value };
+		}
 		showError(error, i18n.baseText('settings.security.personalSpace.publishing.error'));
 	}
 }
+
+async function updateSharingSetting(value: boolean) {
```
guillaumejacquart: might be worth extracting a common updateSharingSetting(key, value, toastTitle) or something like that.
Might not though 🤔

guillaumejacquart (commented):
Nice work ! 
I have some minor comments

packages/@n8n/permissions/src/roles/scopes/project-scopes.ee.ts
```diff
@@ -56,15 +56,12 @@ export const PERSONAL_PROJECT_OWNER_SCOPES: Scope[] = [
...
 	'workflow:execute-chat',
-	'workflow:share',
 	'workflow:move',
 	'credential:create',
 	'credential:read',
 	'credential:update',
 	'credential:delete',
 	'credential:list',
-	'credential:share',
-	'credential:move',
```
BGZStephen, line 67: How does this impact moving credentials from personal spaces to projects? Does it mean they can never move now even if the workflow is moved?

packages/@n8n/db/src/repositories/scope.repository.ts
```diff
@@ -1,15 +1,30 @@
...
+	) {
 		super(Scope, dataSource.manager);
 	}
 
 	async findByList(slugs: string[]) {
 		return await this.findBy({ slug: In(slugs) });
 	}
+
+	async findByListOrFail(slugs: string[]) {
+		const uniqueSlugs = [...new Set(slugs)];
```
BGZStephen: Nit: You could have kept a set here and just converted it for the `In()` usage, and had a faster lookup than a filter / some, but I doubt it'll make much difference given there likely won't be thousands of scopes being checked here

packages/frontend/editor-ui/src/features/settings/security/SecuritySettings.vue
```diff
@@ -65,11 +37,78 @@ async function updatePublishingSetting(value: boolean) {
...
+	try {
+		await securitySettingsApi.updateSecuritySettings(rootStore.restApiContext, {
+			personalSpaceSharing: value,
+		});
+		showToast({
+			type: 'success',
+			title: value
+				? i18n.baseText('settings.security.personalSpace.sharing.success.enabled' as BaseTextKey)
```
BGZStephen: There `as` should be removed, they normally exist because i18n hasn't been rebuilt in the IDE and AI just casts them

packages/@n8n/permissions/src/roles/scopes/project-scopes.ee.ts
phyllis-noester: yes exactly, there is a follow-up ticket around UX for this but the idea is that we don't move secrets when we move workflow

packages/frontend/editor-ui/src/features/settings/security/SecuritySettings.vue
phyllis-noester: found guilty for vibing a bit too hard haha, thank youuu

packages/cli/src/dto/security-settings.dto.ts
phyllis-noester: You are right, weird! removing it

packages/cli/src/controllers/security-settings.controller.ts
phyllis-noester: I was also going back and forth about that... at the moment the two settings are always updated separately so I kept it to ensure that we don't touch more on the db than necessary. what do you think? I'm fine either way

guillaumejacquart: it's probably fine for now, we can always change that later if we introduce more fields

packages/frontend/editor-ui/src/features/settings/security/SecuritySettings.vue
phyllis-noester: I played around with that as well but I think it's better to keep it:
- we only allow updating one or the other atm, if we pass along both it would allow for the scenario where they are both updated at the same timeor we have to check that there is never more than one being updated which would not be straightforward
- we need to make sure the one that is not updated is not overwritten in the state 

therefore I thought this is more transparent. but open to change if you have a suggestion. IMO: once we introduce more settings, we should generally rethink how settings are updated. atm they are both toggled and live updated upon toggle. if we have more settings, we might want to allow the user to change several at once and then click on a save button

guillaumejacquart: Make sense, thanks !

guillaumejacquart (dismissed):
Looks good ! thanks for fixing the feedbacks

guillaumejacquart (approved):

PR #25263, opened 2026-02-03
ci: Add Kent for local testing of Sentry events and traces

shortstacked:
## Summary

Adds **Kent** (Sentry's official mock server) to the testcontainers infrastructure for testing Sentry SDK integrations. This enables E2E tests to verify that errors and traces are actually captured and sent to Sentry - critical for the Shift-Right Observability initiative.

### Why this approach

Currently, we have no way to verify that Sentry is working correctly in E2E tests. We configure DSNs and sample rates, but never validate that:
- Frontend errors actually reach Sentry
- Backend transactions are traced
- Task runner events have correct tags
- Filters (like user code error suppression) work as intended

Kent is Sentry's own testing tool - it's a mock Sentry server that captures events for inspection. By running it in our testcontainers setup, we can assert on actual Sentry payloads.

### What's included

1. **Kent service** (`packages/testing/containers/services/kent.ts`)
   - Docker container running Kent with CORS support (required for browser → container communication)
   - `KentHelper` thin API client exposed via `n8nContainer.services.kent`
   - Methods: `clear()`, `getEvents()`, `getSource()`, `getType()`, `getErrorMessage()`

2. **CORS wrapper** (`packages/testing/containers/dockerfiles/kent/`)
   - Flask-CORS wrapper for Kent to allow browser-based testing
   - Without CORS, the frontend SDK can't send events to Kent in the container

3. **Capability integration** (`kent` capability in `capabilities.ts`)
   - Use with `test.use({ capability: 'kent' })` in tests
   - Automatically sets `N8N_SENTRY_DSN`, `N8N_FRONTEND_SENTRY_DSN`, and related env vars

4. **Baseline tests** (`tests/e2e/sentry/sentry-baseline.spec.ts`)
   - Verifies frontend errors are captured
   - Verifies backend transactions are captured
   - Demonstrates event filtering by source/type

### How to test locally

```bash
# From packages/testing/playwright
pnpm test:e2e --grep "Sentry baseline"
```

Requirements:
- Docker running (testcontainers will build Kent image on first run)
- n8n container mode (not local dev - tests need the container environment)

The tests will:
1. Start Kent container with CORS support
2. Start n8n with Sentry DSNs pointing to Kent
3. Trigger frontend errors / navigate to capture transactions
4. Poll Kent API to verify events were received

### Use cases this enables

- **Baseline validation**: Verify current Sentry behavior before making changes
- **Filter testing**: Add tests for specific Sentry filters (e.g., user code error suppression)
- **Regression prevention**: Ensure Sentry config changes don't break event capture
- **Future @n8n/sentry-config**: Validate central Sentry config produces same events

https://linear.app/n8n/issue/CAT-429/add-sentry-monitoring-in-staging

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/testing/containers/dockerfiles/kent/Dockerfile
+++ b/packages/testing/containers/dockerfiles/kent/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.12-slim
+
+# Install Kent and flask-cors for browser CORS support
+RUN pip install --no-cache-dir kent flask-cors
+
+# Kent default port
+EXPOSE 8000
+
+# Copy and use our wrapper script that enables CORS
+COPY kent_cors.py /app/kent_cors.py
+
+CMD ["python", "/app/kent_cors.py"]
--- a/packages/testing/containers/dockerfiles/kent/Dockerfile
+++ b/packages/testing/containers/dockerfiles/kent/Dockerfile
@@ -1,12 +1,9 @@
 FROM python:3.12-slim
 
-# Install Kent and flask-cors for browser CORS support
 RUN pip install --no-cache-dir kent flask-cors
 
-# Kent default port
 EXPOSE 8000
 
-# Copy and use our wrapper script that enables CORS
 COPY kent_cors.py /app/kent_cors.py
 
 CMD ["python", "/app/kent_cors.py"]
--- a/packages/testing/containers/dockerfiles/kent/kent_cors.py
+++ b/packages/testing/containers/dockerfiles/kent/kent_cors.py
@@ -1,0 +1,9 @@
+"""Kent server wrapper with CORS support for browser-based testing."""
+from flask_cors import CORS
+from kent.app import create_app
+
+app = create_app()
+CORS(app)  # Enable CORS for all origins
+
+if __name__ == "__main__":
+    app.run(host="0.0.0.0", port=8000)
--- a/packages/testing/containers/services/kent.ts
+++ b/packages/testing/containers/services/kent.ts
@@ -65,7 +65,7 @@ export const kent: Service<KentResult> = {
 			N8N_FRONTEND_SENTRY_DSN: result.meta.frontendDsn,
 			N8N_SENTRY_TRACES_SAMPLE_RATE: '1.0',
 			ENVIRONMENT: 'test',
-			N8N_SENTRY_DISABLE_FILTERING: 'true',
+			DEPLOYMENT_NAME: 'e2e-test-deployment',
 		};
 	},
 };
@@ -74,6 +74,12 @@ export const kent: Service<KentResult> = {
 
 export type EventSource = 'backend' | 'frontend' | 'task_runner' | 'unknown';
 export type EventType = 'error' | 'transaction' | 'session' | 'unknown';
+
+export interface KentEventFilter {
+	source?: EventSource;
+	type?: EventType;
+	messageContains?: string;
+}
 
 export interface KentEvent {
 	event_id: string;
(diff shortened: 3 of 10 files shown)


packages/testing/containers/dockerfiles/kent/kent_cors.py
```diff
@@ -0,0 +1,9 @@
+"""Kent server wrapper with CORS support for browser-based testing."""
+from flask_cors import CORS
+from kent.app import create_app
+
+app = create_app()
+CORS(app)  # Enable CORS for all origins
```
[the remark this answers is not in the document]
shortstacked: This is a local container for testing so CORS is allowed to be permissive.

shortstacked:
@claude Review please

packages/testing/containers/dockerfiles/kent/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.12-slim
+
+# Install Kent and flask-cors for browser CORS support
```
tomi: There's no need to repeat in comments what each step does

packages/testing/containers/services/kent.ts
```diff
@@ -0,0 +1,156 @@
...
+			exception?: { values: Array<{ type: string; value: string }> };
+			spans?: unknown[];
+			[key: string]: unknown;
+		};
+	};
+}
+
+// ==================== Helper (thin API client) ====================
+
+export class KentHelper {
```
tomi, line 104: Can we move this to a separate file?

packages/testing/playwright/tests/e2e/sentry/sentry-baseline.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+			setTimeout(() => {
+				throw new Error('Test frontend error');
+			}, 0);
+		});
+
+		// Poll until Kent captures the frontend error
+		await expect
+			.poll(
+				async () => {
+					const events = await kent.getEvents();
```
tomi: Would be nice if the `getEvents` supported these filters directly, or have a separate method like `getEventsByFilter`

packages/testing/playwright/tests/e2e/sentry/sentry-baseline.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+					return events.find(
+						(e) => kent.getSource(e) === 'backend' && kent.getType(e) === 'transaction',
+					);
+				},
+				{ timeout: 10000 },
+			)
+			.toBeTruthy();
+	});
+
+	test('can count events by source and type', async ({ n8n, n8nContainer }) => {
```
tomi: Is this any different from the previous test?

tomi (commented):
Couple comments

packages/testing/containers/services/kent.ts
tomi: Did you see this comment?

shortstacked: Ah yes sorry I forgot to reply. The normal standard so far is to keep all the container/types/helpers self contained. It's the same pattern for the others, the reason was to keep the directory as flat as possible and easy to find. Since the container creation logic itself is usually tiny it didn't cause the files to be overlarge anyway. It reduces the amount of files to change when adding a new service to:
- service, registry. If we separate them it would be - service, helper, types, registry. 

Which would be fine for now but I'd like to see us get to 40/50 different containers eventually and then it would get messy.

tomi (approved):

PR #25282, opened 2026-02-04
chore: Remove dependency on run-script-os

Matsuuu:
## Summary

While setting up my environment, I noticed we were sparsely using the dependency [run-script-os](https://github.com/charlesguse/run-script-os). It's been in the codebase since release and all it does is make the `npm start` command choose between running `n8n` vs `n8n.cmd`.

Relying on an unmaintained npm package comes with it's own risks. As this can be replaced with [a small helper script](scripts/os-normalize.mjs), I think we should eliminate the dependency from n8n.

As I don't have a windows machine at hand, making sure there's no hiccups on Windows side needs to be done by someone (or the CI ?). However the check for windows `process.platform === "win32"` is used across the software so it should work without any hassle.

## Review / Merge checklist

- [ ] Tests included. 


shortstacked:
There is a build windows CI script, we could add a simple run check in to it there for sanity?

Matsuuu:
@shortstacked That would be a good way to ensure it works yeah.

The library was only used to run built binaries so if it already includes a build step to produce those, we could just add on top of that a step that checks it can run the `npm start` command without a hitch.

Do we have a job already for running that kind of a smoke test? Or should we write one into e.g. https://github.com/n8n-io/n8n/blob/master/.github/workflows/build-windows.yml. ?

.github/workflows/build-windows.yml
```diff
@@ -39,6 +39,30 @@ jobs:
         with:
           build-command: pnpm build
 
+      - name: Smoke test pnpm start (1s)
+        shell: pwsh
+        run: |
+          # Start pnpm start in background
```
shortstacked: Is there a need for it to be a background process, would a simple "pnpm start" not work. If it doesn't start it will exit with a failure anyway so the CI will treat it as a failure.

scripts/os-normalize.mjs
```diff
@@ -0,0 +1,57 @@
...
+
+const { dir = '.' } = argv;
+const [run, ...args] = argv._;
+
+if (!dir || !run) {
+	printUsage();
+	process.exit(2);
+}
+
+cd(dir ?? '.');
```
ivov: `??` will never run as we're setting a default

scripts/os-normalize.mjs
```diff
@@ -0,0 +1,57 @@
...
+const cmd = normalizeCommand(run);
+
+if (run) {
+	echo(chalk.cyan(`$ Running (dir: ${dir}) ${cmd} ${args.join(' ')}`));
+	await $({ stdio: 'inherit' })`${cmd} ${args}`;
+} else {
+	echo(chalk.red('No runnable command provided.'));
+	echo('\n');
+	printUsage();
+}
```
ivov: `run` is guaranteed by now as we bail early

scripts/os-normalize.mjs
```diff
@@ -0,0 +1,57 @@
...
+ * */
+function normalizeCommand(baseName) {
+	const isWindows = process.platform === 'win32';
+	if (!isWindows) {
+		// On mac/linux, run local executable
+		return `./${baseName}`;
+	}
+
+	// Force Powershell
+	usePowerShell();
```
ivov: Let's not add superfluous comments please

ivov (commented):
Thanks for this! Minor stuff

shortstacked (approved):

PR #25284, opened 2026-02-04
fix(core): Prevent SQL syntax error when filtering users with empty ids array

tomi:
## Summary

The `buildUserQuery` method would generate invalid SQL `IN ()` when `filter.ids` was an empty array. Added a length check to skip the IN clause when the array is empty.

https://linear.app/n8n/issue/CAT-2268/error-cannot-read-properties-of-null-reading-map-line-3
https://n8nio.sentry.io/organizations/n8nio/issues/6734869174/events/87e6fb5ea375488f8f728f8f2cacbbef/?project=4503924908883968

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/repositories/user.repository.ts
+++ b/packages/@n8n/db/src/repositories/user.repository.ts
@@ -242,7 +242,7 @@ export class UserRepository extends Repository<User> {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


guillaumejacquart (approved):
Good catch, thanks !

PR #25294, opened 2026-02-04
feat(core): Test connection & refresh secrets

geemanjs:
## Summary
* Add `test` endpoint
* Add `reload` endpoint
* Remove `isEnabled` from the frontend
* Update responses to use zod from `@n8n/api-types` instead of typing only

Closes #ligo-116

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
+++ b/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
@@ -20,12 +20,7 @@ import { ExternalSecretsProviderRegistry } from './provider-registry.service';
 import { ExternalSecretsRetryManager } from './retry-manager.service';
 import { ExternalSecretsSecretsCache } from './secrets-cache.service';
 import { ExternalSecretsSettingsStore } from './settings-store.service';
-import type {
-	ExternalSecretsSettings,
-	SecretsProvider,
-	SecretsProviderSettings,
-	TestProviderSettingsResult,
-} from './types';
+import type { ExternalSecretsSettings, SecretsProvider, SecretsProviderSettings } from './types';
 
 /**
  * Orchestrates external secrets management
@@ -209,10 +204,7 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
 		this.broadcastReload();
 	}
 
-	async testProviderSettings(
-		provider: string,
-		data: IDataObject,
-	): Promise<TestProviderSettingsResult> {
+	async testProviderSettings(provider: string, data: IDataObject) {
 		const testSettings: SecretsProviderSettings = {
 			connected: true,
 			connectedAt: new Date(),
@@ -242,7 +234,7 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
 
 			const [success, error] = await result.provider.test();
 
-			// Explicitly handle the discriminated union
+			// This mostly forces the "typing" to work correctly
 			if (!success) {
 				return { success: false, testState: 'error', error };
 			}
--- a/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
+++ b/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
@@ -20,7 +20,12 @@ import { ExternalSecretsProviderRegistry } from './provider-registry.service';
 import { ExternalSecretsRetryManager } from './retry-manager.service';
 import { ExternalSecretsSecretsCache } from './secrets-cache.service';
 import { ExternalSecretsSettingsStore } from './settings-store.service';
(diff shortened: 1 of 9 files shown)


packages/cli/src/modules/external-secrets.ee/secrets-providers.responses.ee.ts
```diff
@@ -1,8 +1,12 @@
 import type { SecretProviderConnection } from '@n8n/api-types';
 
 export declare namespace SecretsProvidersResponses {
-	type StrippedConnection = Omit<SecretProviderConnection, 'settings' | 'secretsCount' | 'state'>;
+	type StrippedConnection = Omit<
```
geemanjs, line 4: @konstantintieber still got this nit - I will update it to use the Zod types instead on a separate PR as it affects other items 👍

packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
```diff
@@ -226,14 +226,23 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
 			// Connect the provider to authenticate before testing
 			const connectResult = await this.providerLifecycle.connect(result.provider);
 			if (!connectResult.success) {
-				return { ...errorState, error: connectResult.error?.message };
+				return {
+					...errorState,
+					error: connectResult.error?.message,
```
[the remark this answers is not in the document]
geemanjs: This was just a format of pre-existing code 👀

packages/cli/test/integration/external-secrets/secret-providers-connections.api.test.ts
```diff
@@ -709,4 +702,203 @@ describe('Secret Providers Connections API', () => {
...
+					settings: { region: 'us-east-1' },
+				})
+				.expect(200);
+
+			const testResponse = await ownerAgent
+				.post('/secret-providers/connections/test-connection-fail/test')
+				.expect(200);
+
+			expect(testResponse.body.data.success).toBe(false);
+			expect(testResponse.body.data.error).toBeDefined();
```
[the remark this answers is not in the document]
geemanjs: This will need correcting when we do better error handling.. We need to make a low level change in how the `provider.connect` function works.

packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
```diff
@@ -226,14 +226,23 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
...
 			const [success, error] = await result.provider.test();
+
+			// This mostly forces the "typing" to work correctly
+			if (!success) {
+				return { success: false, testState: 'error', error };
+			}
+
 			const currentSettings = await this.settingsStore.getProvider(provider);
-			const testState = this.determineTestState(success, currentSettings?.connected ?? false);
+			const testState: 'connected' | 'tested' = currentSettings?.connected ? 'connected' : 'tested';
```
konstantintieber, line 243: I see that you didn't change this logic from before, just removed the private method it was in before, but I still need some help understanding:
Why would a connected vault have the state `tested` while `success=true`?
Doesn't a successful test go hand-in-hand with the provider being connected at the time of the test?

My thinking: Only if we would support passing in the settings to the test endpoint before saving a connection, could we support a case where a connection is tested successfully but is not connected yet.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.controller.ee.ts
```diff
@@ -103,48 +105,24 @@ export class SecretProvidersConnectionsController {
...
 		_res: Response,
-		@Param('providerKey') _providerKey: string,
-	) {
-		this.logger.debug('Reloading secrets for secret provider connection');
-		return;
-	}
-
-	@Post('/:providerKey/share')
-	@GlobalScope('externalSecretsProvider:update')
-	shareConnection(
```
konstantintieber, line 142: Sharing was removed here, because ...?

geemanjs: No longer used - this is that hangover you mentioned of creating "templated" endpoints before the full design. 

Sharing is done via adding a `projectId` to the `Patch` update request.

packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
geemanjs: ~You're right - this is impossible in the new model, but possible in the "old" one (which I believe we will need to still support for now?)~

konstantintieber: Oh ok, `tested` existed before, so let's not remove it now 👍

konstantintieber (approved):

PR #25295, opened 2026-02-04
feat(editor): Extend secrets provider connection modal with project s…

sandra0503:
## Summary

- Updated the connection modal to handle project scope updates and share settings.
- Allows to share with exactly one project or globally. Shared globally by default.
- Adjusted the useConnectionModal composable to include permission checks.
- Modified tests to include new properties related to project sharing and scope management.
- Refactored the ProjectSharing component to utilize props for the global sharing label.

Closes [LIGO-99](https://linear.app/n8n/issue/LIGO-99/fe-external-secrets-sharing-and-permissions) 

## Review / Merge checklist

- [x] Tests included. 

## Preview

Screenshot 2026-02-04 at 11 08 01


konstantintieber (commented):
Recorded a video testing this: https://www.loom.com/share/155ac5b3bd884d8ab2594426f75c2b33

Is it by design that I cannot save the change e.g. because the backend is not ready yet, or?

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionCard.ee.vue
```diff
@@ -40,13 +41,20 @@ const actionDropdownOptions = computed(() => {
...
+			value: 'share',
+		},
 	];
 });
 
 function onAction(action: string) {
 	if (action === 'edit') {
 		emit('edit', provider.value.name);
 	}
+	if (action === 'share') {
```
konstantintieber: Looks like this code can be simplified to `emit(action, provider.value.name);`

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
```diff
@@ -23,56 +24,123 @@ import {
...
+	() => modal.connectionProjects.value,
+	async (projects: ConnectionProjectSummary[]) => {
+		const project = projects?.[0];
+		if (!project) {
+			sharedWithProjects.value = [];
+			return;
+		}
+
+		// Fetch project if not in store
+		if (!projectsStore.projects.find((p: ProjectSharingData) => p.id === project.id)) {
```
konstantintieber: What needs to happen for a project to be in the store already when this modal opens? Is it that the same modal has been opened before?
What are we caching here?

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
```diff
@@ -23,56 +24,123 @@ import {
...
+			await projectsStore.fetchProject(project.id);
+			await projectsStore.getAvailableProjects();
```
konstantintieber: Did you take inspiration from other dialogs that allow selecting projects in this implementation?
When searching for references of `projectsStore.getAvailableProjects`, I don't see the usual sharing dialogs we have in our app today in there.
Is it that since we are in the admin area and the "normal" sidebar, which usually depends on fetching these already, is not visible here, that we need to call `getAvailableProjects` here, in case someone visits the page directly via URL?

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
```diff
@@ -332,6 +400,38 @@ const { width } = useElementSize(nameRef);
 								</form>
 							</div>
 						</div>
+
+						<!-- Scope Tab Content (edit mode only) -->
```
konstantintieber, line 402: I had to think a bit longer what is meant by "scope" here. What do yo think about calling the tab id `sharing`?

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
sandra0503: Projects should be already loaded in the global store before accessing the modal. However this makes sure that the project gets loaded in case it has not been loaded already.

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
sandra0503: Yes, that's better!

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
sandra0503: Actually this is not needed, in case projects have not been loaded globally it should be enough to load them one by one. As I mentioned above, they should be available anyway.

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionCard.ee.vue
```diff
@@ -49,12 +49,8 @@ const actionDropdownOptions = computed(() => {
 });
 
 function onAction(action: string) {
-	if (action === 'edit') {
-		emit('edit', provider.value.name);
-	}
-	if (action === 'share') {
-		emit('share', provider.value.name);
-	}
+	if (action !== 'edit' && action !== 'share') return;
```
konstantintieber, line 52: I wonder, since we control all calls of the action string emitted by the modal, can't we make the `action` parameter typed here so that we don't need to do any string value matching?

sandra0503: It's tricky because N8nActionToggle component emits string

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectSharing.vue
```diff
@@ -34,9 +24,21 @@ type Props = {
 	clearable?: boolean;
 	canShareGlobally?: boolean;
 	isSharedGlobally?: boolean;
+	allUsersLabel?: string;
```
konstantintieber, line 27: Where does the addition of this `allUsersLabel` come from? It seems it was added in this PR but I don't see it provided as a prop anywhere.

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
```diff
@@ -23,56 +24,122 @@ import {
...
+		: [],
+);
+
+const scopeReadonly = computed(() => !modal.canUpdate.value && !modal.canRemoveProjectScope.value);
+
+// Sync scope from composable when connection loads (max 1 project)
+watch(
+	() => modal.connectionProjects.value,
+	async (projects: ConnectionProjectSummary[]) => {
+		const project = projects?.[0];
```
konstantintieber: Nit code style: personally I think this check would read better as `if (projects.length === 0)`

konstantintieber (dismissed):
LGTM

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectSharing.vue
sandra0503: It's about the 'Global' option - it was called 'All users and projects' in other places
https://github.com/n8n-io/n8n/pull/25295/changes#diff-998350112058161ca89d51e1deff496faeee998a78cbc584b0cd17fb1553bdafR419

konstantintieber: Aah yes, I didn't consider that I'd need to search for the kebap case name :D 
thx

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
```diff
@@ -108,19 +108,18 @@ const scopeReadonly = computed(() => !modal.canUpdate.value && !modal.canRemoveP
...
-		if (!project) {
+		if (projects.length === 0) {
 			sharedWithProjects.value = [];
 			return;
 		}
 
 		// Fetch project if not in store
-		if (!projectsStore.projects.find((p: ProjectSharingData) => p.id === project.id)) {
-			await projectsStore.fetchProject(project.id);
+		if (!projectsStore.projects.find((p: ProjectSharingData) => p.id === projects[0].id)) {
```
konstantintieber, line 117: This code works today, only because the backend has the limitation that max one project can have access to a vault, right?

konstantintieber (approved):
Just one more clarification question 👍

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/components/SecretsProviderConnectionModal.ee.vue
sandra0503: The limitation is coming more from the Frontend, only one project can be selected whereas the Backend would already work with multiple projects (maybe there is some validation - not sure).

PR #25296, opened 2026-02-04
feat(core): Handle resuming executions on Chat hub

Cadiac:
## Summary

Change how we handle executions in `lastNode` and `responseNodes` response modes on Chat hub (streaming to be changed later), to instead of waiting for the execution to finish on chat hub to rely on lifecycle hooks.

Now the execution is just started, registered as one Chat hub monitors, runs and when it finishes the new lifecycle hooks on `ChatHubExecutionWatcherService` catch that & save the response and sends it to frontend.

This makes it possible to support executions in `waiting` state resuming later for chat hub external reasons - for instance when a Wait node finishes or some HITL node triggers resume. With this mechanism we still catch them resuming & finishing, and can push the responses to UI.

https://linear.app/n8n/issue/CHA-108/support-new-chat-node-approval-buttons-hitl

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/config/src/configs/chat-hub.config.ts
```diff
@@ -0,0 +1,24 @@
...
+@Config
+export class ChatHubConfig {
+	/**
+	 * TTL in seconds for storing execution context in Chat Hub.
+	 * This is the max duration for a single non-streaming Workflow Agent execution on Chat hub,
+	 * including any waiting time. Once this TTL expires, the responses produced by these
+	 * executions won't be captured or sent to the user on frontend.
+	 * */
+	@Env('N8N_CHAT_HUB_EXECUTION_CONTEXT_TTL')
+	executionContextTtl: number = 3600;
```
Cadiac, line 12: I'm not expecting us to really touch these and I hope 1h is enough, but added these envs just in case we need to override these for bigger enterprise customers or something in future

packages/cli/src/modules/chat-hub/chat-stream-state.service.ts
```diff
@@ -348,10 +343,11 @@ export class ChatStreamStateService {
 		if (!this.redisClient) return;
 
 		try {
-			await this.redisClient.setex(
+			await this.redisClient.set(
```
Cadiac, line 346: found out that SETEX is marked as deprecated on redis logs, and the preferred way appears to be SET with EX param

packages/@n8n/config/src/configs/chat-hub.config.ts
autologie: What happens for the user if they responds to a button after this period?

autologie: (It would be nice if we show something like "the action is expired")

Cadiac: > What happens for the user if they responds to a button after this period?

In the case of Chat trigger and it's Approve button if the execution is still in waiting state and the webhook is around the execution is continued when the button is clicked and execution resumes, but nothing gets saved from that execution into chat hub.

In the case of say a Wait node with 10h wait same thing, the execution would continue after 10h and nothing would get stored to chat hub / sent to UI after that.

I'm more worried of leaking resources if we made these infinitely long than these edge cases, but one option could be to handle the expiration at code, and create a chat hub message "Execution expired" in that case... or something.

I was also considering of changing this to just hold those executions infinitely long on the redis store and having a periodic check to see if the execution is still running/waiting, and if it's complete to clear it, but decided against it at least for now.

On in-memory mode after restart we also lose this state, it's not persisted anywhere. This is similar to how the regular Chat trigger works, and I considered fixing that out of scope for now at least.

Cadiac: I suppose this could also cancel the execution when this context expires, but I'm not sure if that's a good idea.

packages/cli/src/modules/chat-hub/chat-hub-execution-store.service.ts
```diff
@@ -0,0 +1,225 @@
...
+	/** Current AI message ID being generated */
+	messageId: ChatMessageId;
+	/** Previous message ID (human message or prior AI message) */
+	previousMessageId: ChatMessageId | null;
+	/** The conversation model configuration */
+	model: ChatHubConversationModel;
+	/** Response mode: 'lastNode' or 'responseNodes' */
+	responseMode: NonStreamingResponseMode;
+	/** Set true if this execution is resuming now */
+	isResuming: boolean;
```
autologie: IMHO this is a bit confusing, from what I understood this is more like `isWaiting` or `isWaitingToBeResumed`?

Cadiac: agreed, it's more like `wasWaitingButSinceYouAreLookingAtThisNowItMustBeResuming`.

I'll try to name it better

Cadiac: Renamed to `awaitingResume` for now, which might reflect better what it means?

packages/cli/src/modules/chat-hub/chat-hub-execution-watcher.service.ts
```diff
@@ -0,0 +1,283 @@
...
+
+			// Update context: link new message to waiting message
+			await this.executionStore.update(executionId, {
+				previousMessageId: context.messageId,
+				messageId: newMessageId,
+				createMessageOnResume: false,
+			});
+
+			context.previousMessageId = context.messageId;
+			context.messageId = newMessageId;
```
autologie: is it not necessary to update context through executionStore here?

Cadiac: I think you're right and this is broken on redis mode, checking

Cadiac: Actually it works because on line 76 just before this we update the context through executionsStore to have this data, and we're writing on `context` here to be able to use that on `this.chatStreamService.startStream` call later on this function.

but that is a bit messy, I'll clean it up

packages/cli/src/modules/chat-hub/chat-hub-execution-watcher.service.ts
```diff
@@ -0,0 +1,283 @@
...
+	@OnLifecycleEvent('workflowExecuteAfter')
+	async handleWorkflowExecuteAfter(ctx: WorkflowExecuteAfterContext): Promise<void> {
+		const { runData, executionId } = ctx;
+
+		const context = await this.executionStore.get(executionId);
+		if (!context) return; // Not a tracked chat hub execution
+
+		this.logger.debug('Handling workflow execution completion', { executionId });
+
+		if (!['success', 'waiting', 'canceled'].includes(runData.status)) {
```
autologie, line 124: [nits] conditions for waiting and canceled status are checked below too. I'd bring this branch after waiting and just check `if (runData.status !== 'success')`

packages/cli/src/modules/chat-hub/chat-hub-execution-watcher.service.ts
autologie: True, my eyes slipped

packages/cli/src/modules/chat-hub/chat-hub-execution-watcher.service.ts
Cadiac: there's also possibility for unexpected statuses like `'crashed'`, `'unknown'` or `'running'`, I wanted this to error if the status is one of those and basically just list the expected statuses here 🤔

Cadiac: (i've not seen evidence of those unexpected statuses in practice entering `workflowExecuteAfter`)

packages/cli/src/modules/chat-hub/chat-hub-execution-watcher.service.ts
Cadiac: I changed this anyways to do an early return on the if branch instead (and to avoid double execution context update).

packages/cli/src/modules/chat-hub/chat-hub-execution-watcher.service.ts
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		const message = this.chatHubExecutionService.extractMessage(runData, context.responseMode);
+
+		if (runData.status === 'waiting') {
+			await this.handleWaitingExecution(context, executionId, message);
+			return;
+		}
+
+		if (runData.finished) {
```
Cadiac, line 146: I just now remembered this solution still uses the `finished` field, which while it works is marked as deprecated.

I needed this on some edge cases on multi-main, where I found out that this hook gets called with status `success` even when the execution isn't complete.

I'll investigate tomorrow if using this could be avoided.

autologie (approved):
🥇

packages/@n8n/config/src/configs/chat-hub.config.ts
autologie: We could maybe send some message on expiration but probably a edge case anyways

PR #25298, opened 2026-02-04
refactor: Improve performance of source control methods

alielkhateeb:
## Summary

This PR should only be an improvement to the performance of pull/push, as well as some readability improvements.

> [!NOTE]  
> The diff looks bigger than it is, most changes are simply wrapping existing logic in `if (verbose)` blocks.

### Performance benchmarks

I ran local benchmarks comparing `master` vs this branch on the `get-status` endpoint with the following setup:

- **100 workflows**
- **1 credential**
- **17 projects**

Since these benchmarks were run locally, the numbers aren’t 100% precise. As usual, I ran multiple iterations (≈3 per branch) and compared averages.

#### Peak memory deltas during execution
```
peakRssDelta: 4.13 MB (branch) vs 5.16 MB (master) → −1.03 MB (~20%)
peakHeapUsedDelta: 2.96 MB (branch) vs 6.87 MB (master) → −3.91 MB (~57%)
```

This shows a consistent reduction in peak memory usage during `get-status`, especially on the JS heap.

---

### Key behavioural improvement

Across all runs, one consistent and clear improvement was the number of items produced:

- **236 items** on `master`
- **118 items** on this branch

This reduction comes from no longer storing verbose items when the `verbose` flag is `false`. Previously, verbose items were collected regardless of the flag value, even when they were not needed.

Given the consistently lower number of items produced across all runs (118 vs 236), I would estimate the **effective memory performance improvement** to be in the order of **~50%**, which aligns with the observed reduction in peak heap usage.

### TODO
Need to modify tests to ensure coverage don't drop

[Ligo-128](https://linear.app/n8n/issue/LIGO-128)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/eslint.config.mjs
+++ b/packages/cli/eslint.config.mjs
@@ -2,7 +2,7 @@ import { defineConfig, globalIgnores } from 'eslint/config';
 import { nodeConfig } from '@n8n/eslint-config/node';
 
 export default defineConfig(
-	globalIgnores(['scripts/**/*.mjs', 'jest.config*.js', 'test/*-testcontainers.js']),
+	globalIgnores(['scripts/**/*.mjs', 'jest.config*.js', 'test/*-testcontainers.js', 'coverage/**']),
 	nodeConfig,
 	{
 		rules: {
--- a/packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
+++ b/packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
@@ -523,13 +523,6 @@ describe('areSameCredentials', () => {
 		expect(areSameCredentials(creds1, creds2)).toBe(true);
 	});
 
-	it('should return true when only ids are different', () => {
-		const creds1 = mockCredential();
-		const creds2 = mockCredential({ id: 'cred2' });
-
-		expect(areSameCredentials(creds1, creds2)).toBe(true);
-	});
-
 	it('should return true when only data is different', () => {
 		const creds1 = mockCredential({ data: { accessToken: 'access token' } });
 		const creds2 = mockCredential({ data: { accessToken: 'different access token' } });
--- a/packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
+++ b/packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
@@ -530,6 +530,13 @@ describe('areSameCredentials', () => {
 		expect(areSameCredentials(creds1, creds2)).toBe(true);
 	});
 
+	it('should return true when only data is different', () => {
+		const creds1 = mockCredential({ data: { accessToken: 'access token' } });
+		const creds2 = mockCredential({ data: { accessToken: 'different access token' } });
+
+		expect(areSameCredentials(creds1, creds2)).toBe(true);
+	});
+
 	it('should return false when names are different', () => {
 		const creds1 = mockCredential();
 		const creds2 = mockCredential({ name: 'Different Name' });
--- a/packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
+++ b/packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
@@ -18,11 +18,13 @@ import {
 	getTrackingInformationFromPullResult,
 	isWorkflowModified,
 	sourceControlFoldersExistCheck,
+	areSameCredentials,
 } from '../source-control-helper.ee';
 import type { SourceControlPreferencesService } from '@/modules/source-control.ee/source-control-preferences.service.ee';
 import type { License } from '@/license';
 
 import type { SourceControlWorkflowVersionId } from '../types/source-control-workflow-version-id';
+import type { StatusExportableCredential } from '../types/exportable-credential';
 
 function createWorkflowVersion(
 	overrides: Partial<SourceControlWorkflowVersionId> = {},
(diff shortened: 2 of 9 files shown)


packages/cli/src/modules/source-control.ee/source-control-status.service.ee.ts
```diff
@@ -217,30 +220,24 @@ export class SourceControlStatusService {
 			// we need to query for all wf in the DB to hide possible deletions,
 			// when a wf went out of scope locally
 			outOfScopeWF = await this.sourceControlImportService.getAllLocalVersionIdsFromDb();
-			outOfScopeWF = outOfScopeWF.filter(
-				(wf) => !wfLocalVersionIds.some((local) => local.id === wf.id),
-			);
+			outOfScopeWF = outOfScopeWF.filter((wf) => !wfLocalIds.has(wf.id));
```
alielkhateeb, line 215: The idea here is that `Set#has()` is O(1) compared to `some` which is O(n)

packages/cli/src/modules/source-control.ee/source-control-helper.ee.ts
```diff
@@ -275,3 +276,15 @@ export function isWorkflowModified(
 
 	return hasVersionIdChanged || hasParentFolderIdChanged || ownerChanged;
 }
+
+export function areSameCredentials(
```
ireneea, line 280: Nice one 👍

packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
```diff
@@ -491,3 +493,75 @@ describe('readFoldersFromSourceControlFile', () => {
...
+	};
+
+	it('should return true when credentials are the same', () => {
+		const creds1 = mockCredential();
+		const creds2 = mockCredential();
+
+		expect(areSameCredentials(creds1, creds2)).toBe(true);
+	});
+
+	it('should return true when only ids are different', () => {
```
geemanjs: Is this test correct?

geemanjs (commented):
I'm not totally confident I understand how this worked vs how it works now so I'll delegate approval to @ireneea 👀

packages/cli/src/modules/source-control.ee/__tests__/source-control-helper.ee.test.ts
alielkhateeb: Yes, `ids` are not part of the comparison, as far as I have seen from the previous code at least

ireneea: Good point, I think the real use case here is that it should return `true` when only data is different

```
	it('should return true when only ids are different', () => {
		const creds1 = mockCredential({ data: { port: 5678 } });
		const creds2 = mockCredential({ data: { port: 1234 } });

		expect(areSameCredentials(creds1, creds2)).toBe(true);
	});
```

alielkhateeb: @ireneea ah good catch this is not covered in tests, adding it now!

alielkhateeb: 4f24287d4e172f9318516d633618e547f75dd745 CC @ireneea

alielkhateeb: @geemanjs had a call with @ireneea and explained why only ids is not useful here, now I get it 😄 removed here 283b69fd86465b83567b87fb432fb37798ab256b

packages/cli/src/modules/source-control.ee/source-control-export.service.ee.ts
```diff
@@ -110,25 +112,28 @@ export class SourceControlExportService {
...
-					versionId: e.versionId,
-					owner: owners[e.id],
-					parentFolderId: e.parentFolder?.id ?? null,
-					isArchived: e.isArchived,
-				};
-				this.logger.debug(`Writing workflow ${e.id} to ${fileName}`);
-				return await fsWriteFile(fileName, JSON.stringify(sanitizedWorkflow, null, 2));
-			}),
-		);
+		const workflowChunks = chunk(workflowsToBeExported, SOURCE_CONTROL_WRITE_FILE_BATCH_SIZE);
```
alielkhateeb, line 115: For visibility: This will help with retained memory while writing workflow files to the git folder by batching the writes instead of trying to dump all 1000+ workflows at the same time

ireneea (dismissed):

packages/cli/eslint.config.mjs
```diff
@@ -2,7 +2,7 @@ import { defineConfig, globalIgnores } from 'eslint/config';
 import { nodeConfig } from '@n8n/eslint-config/node';
 
 export default defineConfig(
-	globalIgnores(['scripts/**/*.mjs', 'jest.config*.js', 'test/*-testcontainers.js']),
+	globalIgnores(['scripts/**/*.mjs', 'jest.config*.js', 'test/*-testcontainers.js', 'coverage/**']),
```
alielkhateeb, line 5: This change is only for locally generated coverage to not be linted (Screenshot for demonstrating the pain 😄), and this also follows the pattern used in [@n8n/ai-workflow-builder.ee](https://github.com/n8n-io/n8n/blob/2716fa5de9fde31d683d49498159fe93a798741e/packages/%40n8n/ai-workflow-builder.ee/eslint.config.mjs#L5)

Image

ireneea (approved):

PR #25044, opened 2026-01-29
chore(core): Change db column type to TEXT for subject_id

afitzek:
## Summary

This PR addresses a database limitation where the `subject_id` column in the `dynamic_credential_entry` table was restricted to `varchar(16)`, which is insufficient for certain dynamic credential resolvers that generate longer subject identifiers.

### Changes Made

- **Migration**: Created `ExpandSubjectIDColumnLength1769690595000` migration that:
- Changes `subject_id` column type from `varchar(16)` to `VARCHAR(2048)`
- Recreates the table for both PostgreSQL and SQLite (required because `subject_id` is part of the composite primary key)
- Preserves all existing data using the `copyTable` utility (batch copying in chunks of 10 rows)
- Maintains all constraints (primary key, foreign keys, indexes)
- Postgres does not support the short syntax for LIMIT and OFFSET (this was changed in the copyTable migration helper)

### Why Table Recreation?

The `subject_id` column is part of a **composite primary key** (`credential_id`, `subject_id`, `resolver_id`). Both PostgreSQL and SQLite have restrictions on altering columns that are part of primary key constraints:

- **PostgreSQL**: Cannot alter column type when it's part of a primary key without dropping the constraint first
- **SQLite**: Cannot drop columns that are part of a primary key

The cleanest and safest approach is to recreate the table with the new schema, which the migration does using n8n's built-in `copyTable` utility that handles data preservation, foreign keys, and indexes.

## Related Linear tickets, Github issues, and Community forum posts

closes https://linear.app/n8n/issue/IAM-204/increase-varchar-limit-for-subjectid-column-in-postgrest-for-dynamic

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [X] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #25301, opened 2026-02-04
fix(core): Wrong pagination argument order for sqlite in copyTables helper

konstantintieber:
## Summary

Today we came across a failing migration on a crash-looping cloud instance:
```
Starting migration ChangeValueTypesForInsights1759399811000                      
Migration "ChangeValueTypesForInsights1759399811000" failed, error: SQLITE_ERROR: near "10": syntax error
```

In sqlite, OFFSET needs to come after LIMIT, otherwise the statement will fail.

This only affects 2.7.x. Regression from https://github.com/n8n-io/n8n/pull/25044

The CI for our migrations runs in multi-main mode with postgres, which is likely why this wasn't caught before merging.

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/migration-helpers.test.ts
+++ b/packages/@n8n/db/src/migrations/migration-helpers.test.ts
@@ -1,0 +1,102 @@
+import { Container } from '@n8n/di';
+import { DataSource } from '@n8n/typeorm';
+
+import { DbConnection } from '../index';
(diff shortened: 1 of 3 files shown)


afitzek (approved):

guillaumejacquart (approved):
Good catch, thanks !

PR #25309, opened 2026-02-04
feat: Add credentials list endpoint + new scopes

BGZStephen:
## Summary

This PR adds a new `credentials:list` scope, and a new GET /credentials endpoint available on the public API. As a follow up, we will extend this endpoint to allow expansion in to sharing details 

image

IAM-172

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/repositories/credentials.repository.ts
+++ b/packages/@n8n/db/src/repositories/credentials.repository.ts
@@ -32,6 +32,19 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
 		return await this.find(findManyOptions);
 	}
 
+	async findManyAndCount(
+		listQueryOptions?: ListQuery.Options & { includeData?: boolean; user?: User },
+		credentialIds?: string[],
+	): Promise<[CredentialsEntity[], number]> {
+		const findManyOptions = this.toFindManyOptions(listQueryOptions);
+
+		if (credentialIds) {
+			findManyOptions.where = { ...findManyOptions.where, id: In(credentialIds) };
+		}
+
+		return await this.findAndCount(findManyOptions);
+	}
+
 	private toFindManyOptions(listQueryOptions?: ListQuery.Options & { includeData?: boolean }) {
 		const findManyOptions: FindManyOptions<CredentialsEntity> = {};
 
--- a/packages/@n8n/db/src/repositories/credentials.repository.ts
+++ b/packages/@n8n/db/src/repositories/credentials.repository.ts
@@ -64,7 +64,14 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
 			'resolvableAllowFallback',
 		];
 
-		if (!listQueryOptions) return { select: defaultSelect, relations: defaultRelations };
+		const defaultOrder = { createdAt: 'DESC' as const };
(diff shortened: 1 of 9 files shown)


packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
```diff
@@ -29,9 +29,45 @@ import {
...
+		async (
+			req: CredentialRequest.GetAll,
+			res: express.Response,
+		): Promise<
+			express.Response<{ data: Partial<CredentialsEntity>[]; nextCursor: string | null }>
+		> => {
+			const offset = Number(req.query.offset) || 0;
+			const limit = Math.min(Number(req.query.limit) || 100, 250);
+
+			const [credentials, count] = await Container.get(CredentialsRepository).findManyAndCount({
```
guillaumejacquart, line 54: question: if we do not setup an order, how can we be sure the cursor (last id) will be deterministic

guillaumejacquart (commented):
looks good ! one comment though

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
BGZStephen: Good catch, let me use createdAt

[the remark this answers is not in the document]
BGZStephen: This is correct, API keys are created with scopes that are used for this endpoint and interaction.

packages/@n8n/db/src/repositories/credentials.repository.ts
```diff
@@ -79,6 +99,10 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
 			findManyOptions.relations = defaultRelations;
 		}
 
+		if (!findManyOptions.order) {
+			findManyOptions.order = { createdAt: 'DESC' as const };
```
guillaumejacquart: we could use defaultOrder here no ?

guillaumejacquart: also, is there a risk this changes how credentials are displayed in the UI ?

guillaumejacquart (commented):
just a small comment

packages/@n8n/db/src/repositories/credentials.repository.ts
BGZStephen: Yeah, I was on the fence about the ordering, on the one hand it feels like we should add default ordering, but I can see your point, let me only order if the property is provided

packages/@n8n/db/src/repositories/credentials.repository.ts
```diff
@@ -79,6 +112,10 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
 			findManyOptions.relations = defaultRelations;
 		}
 
+		if (order !== undefined) {
+			findManyOptions.order = order;
+		}
```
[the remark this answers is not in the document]
BGZStephen: Yes, this is the current behaviour, we don't want to override current behaviour as it may break thing

guillaumejacquart (approved):

PR #25316, opened 2026-02-04
feat(editor): Add `Name version` action to workflow history

dariacodes:
## Summary

- Introduce "Name version" actions on workflow and workflow history pages (as a licensed feature)
- Remove "Unpublish" action from the main actions dropdown menu and move it to the new dropdown near the publish button
- Prepopulate version id, name and description when fetching workflow and show the name version modal with prepopulated data where applicable
- Introduce `WorkflowHistoryVersionFormModal` to avoid duplicated logic for Publish and Name version modals (this is mostly a copy of `WorkflowHistoryPublishModal.vue`)

image
image
image

Closes [ADO-4732](https://linear.app/n8n/issue/ADO-4732/add-name-version-actions-for-workflows)

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -3481,6 +3482,10 @@
 	"workflowHistory.action.clone.success.title": "Successfully cloned workflow version",
 	"workflowHistory.action.clone.success.message": "Open cloned workflow in a new tab",
 	"workflowHistory.action.unpublish.success.title": "Workflow unpublished successfully",
+	"workflowHistory.action.nameVersion.success.title": "Version name updated successfully",
+	"workflowHistory.action.nameVersion.error.title": "Failed to update version name",
+	"workflowHistory.nameVersionModal.title": "Name version",
+	"workflowHistory.nameVersionModal.nameVersion": "Name version",
```
scdekov: This and above are duplicates

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -3496,6 +3501,8 @@
 	"workflows.publish": "Publish",
 	"workflows.published": "Published",
 	"workflows.publish.permissionDenied": "You don't have permission to publish this workflow",
+	"workflows.nameVersion": "Name version",
```
scdekov: Move this in a generic string?

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -3496,6 +3501,8 @@
...
+	"workflows.unpublish": "Unpublish",
```
scdekov, line 3515: We have this 2 times already. Let's move in a generic string

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -271,12 +312,175 @@ const latestPublishDate = computed(() => {
...
+
+	const currentVersion = workflowsStore.workflow.versionId;
+	const currentWorkflow = workflowsStore.workflow;
+	const currentVersionData = workflowsStore.currentVersion;
+
+	const nameVersionEventBus = createEventBus<WorkflowHistoryNameVersionModalEventBusEvents>();
+
+	nameVersionEventBus.once(
+		'save',
+		async (saveData: { versionId: string; name: string; description: string }) => {
```
scdekov: Isn't the type of `saveData` inferred here?

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -271,12 +312,175 @@ const latestPublishDate = computed(() => {
 	return latestPublish?.createdAt;
 });
 
+const enum PUBLISH_ACTIONS {
+	PUBLISH = 'publish',
+	NAME_VERSION = 'name-version',
+	UNPUBLISH = 'unpublish',
+}
+
+const dropdownMenuActions = computed<Array<ActionDropdownItem<PUBLISH_ACTIONS>>>(() => {
```
scdekov: Should we use different name here? We have another `ActionsDropdownMenu` in this component so it gets confusing

packages/frontend/editor-ui/src/app/components/WorkflowVersionForm.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	it('should render version name and description inputs', () => {
+		const { container } = renderComponent({
+			props: {
+				versionName: '',
+				description: '',
+			},
+		});
+
+		expect(container.querySelector('#workflow-version-name')).toBeInTheDocument();
```
scdekov: It's a little weird that we're using `id` here and not `data-test-id`. I think that I've initially used `id` in this component and don't remember if there was a good reason. Maybe check quickly if using data-test-id also works?

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
```diff
@@ -367,7 +590,33 @@ defineExpose({
...
+}
+
+.buttonGroup {
+	display: inline-flex;
+}
+
+.groupButtonLeft,
+.groupButtonLeft:disabled,
+.groupButtonLeft:hover:disabled {
+	border-top-right-radius: 0;
```
scdekov, line 604: todo: check if this can be reusabel

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
```diff
@@ -970,6 +971,24 @@ export function useWorkflowHelpers() {
 			workflowsStore.setWorkflowActiveVersion(workflowData.activeVersion ?? null);
 		}
 
+		// Fetch current version details if versionId exists
+		if (workflowData.versionId) {
```
scdekov, line 986: Should we rather update the get workflow by id endpoint to return this version data? I guess all existing wfs will have versionIds which means that we're adding an extra(sync) request on each wf open which is already not very fast 🤔

packages/frontend/editor-ui/src/app/stores/workflows.store.ts
```diff
@@ -132,6 +137,7 @@ export const useWorkflowsStore = defineStore(STORES.WORKFLOWS, () => {
 		createWorkflowObject(workflow.value.nodes, workflow.value.connections),
 	);
 
+	const currentVersion = ref<WorkflowVersionMetadata | null>(null);
```
scdekov: Maybe call this versionData/metadata? We already have `workflowVersonId` so I guess the `current` part is implicit here.

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryNameVersionModal.vue
```diff
@@ -0,0 +1,52 @@
+<script setup lang="ts">
+import WorkflowHistoryVersionFormModal from './WorkflowHistoryVersionFormModal.vue';
```
scdekov: The naming sounds weird here. Should we go with `WorkflowHistoryFormModal` or `WorkflowVersionFormModal`

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryPublishModal.vue
```diff
@@ -29,129 +26,43 @@ const props = defineProps<{
...
-onBeforeUnmount(() => {
-	modalEventBus.off('opened', onModalOpened);
-});
-
-const closeModal = () => {
-	uiStore.closeModal(props.modalName);
-};
-
-const onCancel = () => {
+const formEventBus = createEventBus<WorkflowHistoryVersionFormModalEventBusEvents>();
```
scdekov: Do we still need this component and `packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryNameVersionModal.vue`? Seems like we can just pass the props to `WorkflowHistoryVersionFormModal` and simplify?

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowVersionFormModal.vue
```diff
@@ -0,0 +1,141 @@
...
+import { generateVersionName } from '@/features/workflows/workflowHistory/utils';
+import type { EventBus } from '@n8n/utils/event-bus';
+
+export type WorkflowHistoryVersionFormModalEventBusEvents = {
+	submit: { versionId: string; name: string; description: string };
+	cancel: undefined;
+};
+
+const props = defineProps<{
+	modalName: string;
```
scdekov: Let's limit this to only the values that are expected here

packages/frontend/@n8n/i18n/src/locales/en.json
dariacodes: One is the title of the modal and another is "name version" action

dariacodes: But if we move it to a generic string, like you suggested below, maybe we can get rid of a separate key for the button for now

scdekov:
> Should we change the main CTA to "Save"? Feels like we have too many Name version/Version name there image

Also I'm thinking if action should be "Rename" either always or only if the version has a name.

scdekov:
I'm stuck in this state when trying to publish a version without a trigger
image

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
dariacodes: I thought about this, but it looked like a big change, and I thought that just sending one extra request on a page reload should be fine

scdekov: That's every time you open a workflow, right?

dariacodes: Yeah, but it's a very fast query and you can open only one workflow at a time. I propose to optimize this later

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: I did it because of these warnings
image

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: Yeah, I also don't like this confusion... Now we have 2 dropdowns and this one could be called "Publish dropdown" unless it didn't have "Name version" inside 🙈

dariacodes: I might choose `versionMenuActions`

packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.vue
dariacodes: Rob said we would have a design system component at some point, so we'll just swap all of these with it later

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.ts
dariacodes: Ah, we can't actually add `currentVersion` relation through TypeORM, because it's not a foreign key. And we can't make it a foreign key, because it's non-nullable and we already have a foreign key "workflow.id => workflow_history.workflowId", so if we're going to create a workflow first, it'll complaint "FK violation" (because versionId does not exist yet), and if we're going to create a workflow version first, it'll also complain because workflow doesn't exist yet... All this means that sending an additional query would be as expensive as adding it on backend as it'll be anyway a separate query.

dariacodes:
@scdekov I addressed all your comments (for some replied in the thread), except for the "Rename" one. I propose not to overcomplicate this with additional conditional logic. Rob initially wanted to also have actions named differently, but we agreed to go with the same one in the end. If it comes up during a product review though, I'll change that.

CharlieKolb (approved):

PR #25320, opened 2026-02-04
fix(Data Table Node): Fallback to workflow-derived projectId if undefined

CharlieKolb:
## Summary

Fallback to the workflow's projectId if we have the workflow but not the project id. Ideally we'd figure this out in the front end, but there are rare edge cases where it's difficult to infer the project id, e.g. when an admin is browsing a workflow located in another users' project.

I'm favoring this approach as it may also fix other rare issues and provides general resilience, rather than covering one more edge case we happened to catch.

https://linear.app/n8n/issue/ADO-4350_wfid

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/controllers/dynamic-node-parameters.controller.ts
+++ b/packages/cli/src/controllers/dynamic-node-parameters.controller.ts
@@ -13,7 +13,7 @@ import { getBase } from '@/workflow-execute-additional-data';
 
 @RestController('/dynamic-node-parameters')
 export class DynamicNodeParametersController {
-	constructor(private readonly service: DynamicNodeParametersService) {}
+	constructor(private readonly dynamicNodeParametersService: DynamicNodeParametersService) {}
 
 	@Post('/options')
 	async getOptions(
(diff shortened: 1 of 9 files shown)


packages/cli/src/controllers/dynamic-node-parameters.controller.ts
RicardoE105: nit: can we rename the var name to dynamicNodeParametersService as we do everywhere else?

packages/cli/src/services/dynamic-node-parameters.service.ts
```diff
@@ -75,6 +76,26 @@ export class DynamicNodeParametersService {
...
+			} else if (payload.projectId === undefined) {
+				const project = await this.sharedWorkflowRepository.getWorkflowOwningProject(
+					payload.workflowId,
+				);
+				if (project) {
+					payload.projectId = project.id;
+				}
+			}
+		}
 	}
```
RicardoE105: nit: maybe

```
if (!payload.workflowId) return;

const hasAccess = await userHasScopes(user, ['workflow:read'], false, {
  workflowId: payload.workflowId,
});

if (!hasAccess) {
  this.logger.warn(
    `Scrubbed inaccessible workflowId ${payload.workflowId} from DynamicNodeParameters request`,
  );
  payload.workflowId = undefined;
  return;
}

if (payload.projectId === undefined) {
  const project = await this.sharedWorkflowRepository.getWorkflowOwningProject(
    payload.workflowId,
  );
  payload.projectId = project?.id;
}
```

RicardoE105 (dismissed):
🚀 

Are we going to create linear tickets to figure the edge cases in the FE? 🤔

RicardoE105 (approved):
🚀

PR #25321, opened 2026-02-04
fix(core): Syslog message id RFC5424 compliance

afitzek:
## Summary

The msg.id property in RFC5424 is limited to 32 characters, this ensures that we only send 32 characters to a syslog destination. The approach is, if the id is too long, we strip hyphens (for UUID normalization), in case this is not enough, we truncated the it to 32 characters.

closes https://linear.app/n8n/issue/IAM-60/syslog-logstream-not-in-valid-rfc5424-format

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/log-streaming.ee/destinations/message-event-bus-destination-syslog.ee.ts
+++ b/packages/cli/src/modules/log-streaming.ee/destinations/message-event-bus-destination-syslog.ee.ts
@@ -95,7 +95,7 @@ export class MessageEventBusDestinationSyslog
 				JSON.stringify(serializedMessage),
 				{
 					severity: msg.eventName.toLowerCase().endsWith('error') ? Severity.Error : Severity.Debug,
-					msgid: msg.id,
+					msgid: msg.id.length > 32 ? msg.id.replace(/-/g, '').substring(0, 32) : msg.id,
 					timestamp: msg.ts.toJSDate(),
 				},
 				async (error) => {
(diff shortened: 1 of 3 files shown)


BGZStephen (approved):

packages/cli/src/modules/log-streaming.ee/destinations/message-event-bus-destination-syslog.ee.ts
```diff
@@ -95,7 +95,7 @@ export class MessageEventBusDestinationSyslog
 				JSON.stringify(serializedMessage),
 				{
 					severity: msg.eventName.toLowerCase().endsWith('error') ? Severity.Error : Severity.Debug,
-					msgid: msg.id,
+					msgid: msg.id.length > 32 ? msg.id.replace(/-/g, '').substring(0, 32) : msg.id,
```
guillaumejacquart, line 98: are we sure that we only send uuid 36 chars ?

guillaumejacquart (approved):
Looks good, one question

packages/cli/src/modules/log-streaming.ee/destinations/message-event-bus-destination-syslog.ee.ts
afitzek: Currently we only use uuids, but in theory this could change in the future. But it also handles arbitrary other strings if they are too long.

guillaumejacquart: Yes, the substring was worrying me, because of collisions.
On top of that, according to the RFC, the msg id should be the "type" of message 🤔 (https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.7)

afitzek: It should be `MSGID           = NILVALUE / 1*32PRINTUSASCII` (https://datatracker.ietf.org/doc/html/rfc5424#section-6)

PR #25331, opened 2026-02-04
chore: Update cubic rules

ivov:
## Summary

Update cubic rules as discussed with the team:
- Remove unreliable type guards rule: https://n8nio.slack.com/archives/C03MZF137FV/p1769506347648479
- Add DTO rule: https://www.notion.so/n8n/Backend-monthly-board-28b5b6e0c94f80418963fadae027e484?p=28b5b6e0c94f806286dfd982731df01b&pm=s

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/cubic.yaml
+++ b/cubic.yaml
@@ -41,7 +41,7 @@ reviews:
 
     Approve if reasonably tested. Let humans handle edge cases.
   custom_rules:
-- name: Tests
+    - name: Tests
       description: |-
         BE REASONABLE when evaluating test coverage:
 
--- a/cubic.yaml
+++ b/cubic.yaml
@@ -41,58 +41,7 @@ reviews:
 
     Approve if reasonably tested. Let humans handle edge cases.
   custom_rules:
-    - name: Prefer Typeguards over Type casting
-      description: |-
-        - Rule Statement:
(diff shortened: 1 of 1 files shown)


cubic.yaml
```diff
@@ -154,6 +103,55 @@ reviews:
         - Decorator order should be: route decorator → `@Licensed` → scope decorators
         - Enterprise features not properly isolated in modules (newer features should use the module system)
         - Enterprise code accessible outside of `*.ee.ts` files or licensed modules
+    - name: Use DTOs for Request Body Validation
```
tomi, line 106: We also have `@Param` decorator. Could we address that within the same rule or create a separate rule?

ivov: `@Param` has no DTO and typically there's no issue there. I'd skip it only for `string`.

tomi: Yeah I meant more about using decorators in general for input validation in endpoints. You could always use `req.params.key` instead of `@Param` and that's not what we prefer, right?

ivov: Trying not to go overboard with all of them as I find cubic trigger happy with false positives, but lmk if you disagree and I'll expand

tomi: Well it's an improvement already. Let's get it in

tomi (approved):

PR #25338, opened 2026-02-04
feat(editor): Support Chat node's 'Send and Wait for Response' mode approval buttons on Chat hub

Cadiac:
## Summary

image
image

Support rendering approval buttons sent by Chat nodes in `Send and Wait for Response` mode, and resuming execution when they are clicked.

Also added bunch of tests.

https://linear.app/n8n/issue/CHA-108/support-new-chat-node-approval-buttons-hitl

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
```diff
@@ -567,3 +567,17 @@ export function splitMarkdownIntoChunks(content: string): string[] {
...
+ * This occurs when the message has 'waiting' status and contains
+ * a with-buttons chunk that blocks user input.
+ */
+export function isWaitingForApproval(message: ChatMessage | null | undefined): boolean {
+	if (!message || message.status !== 'waiting') {
+		return false;
+	}
+
+	const buttonChunk = message.content.find((c) => c.type === 'with-buttons');
+	return buttonChunk?.type === 'with-buttons' && buttonChunk.blockUserInput;
```
autologie: [nits]
```suggestion
	return message.content.some((c) => c.type === 'with-buttons' && c.blockUserInput);
```

packages/@n8n/chat-hub/src/parser.ts
```diff
@@ -19,14 +23,24 @@ export function appendChunkToParsedMessageItems(
...
+		) {
 			// Incomplete command - append chunk and re-parse
 			// Don't mutate the original item, create new content string
 			remaining = lastItem.content + chunk;
 			result.pop(); // Remove it so we can re-parse
 		}
 	}
 
+	// Check if the chunk is button JSON (arrives as complete JSON in one chunk)
+	const buttonChunk = tryParseButtonsJson(remaining);
```
autologie, line 38: Here, we assume the buttons JSON is always in once piece in the chunk and doesn't contain extraneous content before/after. I guess it won't be otherwise, so probably safe

packages/@n8n/chat-hub/src/parser.ts
```diff
@@ -233,6 +247,26 @@ function extractTagContent(xml: string, tagName: string): string | null {
...
+function tryParseButtonsJson(content: string): ChatMessageContentChunk | null {
+	if (!content.startsWith('{')) return null;
+
+	try {
+		const parsed: unknown = JSON.parse(content);
+		const result = chatHubMessageWithButtonsSchema.safeParse(parsed);
+		if (result.success) {
+			return {
+				type: 'with-buttons',
+				content: result.data.text,
```
autologie, line 259: I think I should have named this field something like `rawText`, which is used in parser to construct parts from incomplete last chunk (we don't show `<comm` in UI but it has to be tracked to recognize following `and:create-artifact>` chunk). So, storing original stringified JSON is more consistent, but then we need a different field to store text data. This also doesn't cause real problems as long as the JSON is entire message body

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessageWithButtons.vue
```diff
@@ -0,0 +1,64 @@
...
+
+<template>
+	<div :class="$style.container">
+		<ChatMarkdownChunk :source="{ type: 'text', content: text }" />
+		<div :class="$style.buttons">
+			<template v-for="(button, index) in buttons" :key="button.text">
+				<N8nButton
+					v-if="clickedButtonIndex === null || index === clickedButtonIndex"
+					:type="button.type"
+					:disabled="index === clickedButtonIndex || !isWaiting || isLoading"
```
autologie: shouldn't we disable all buttons when one of buttons is already clicked?

Cadiac: hm, we kind of do through the `isWaiting || isLoading` path, the clickedButtonIndex is a bit useless now I think - maybe I'll just change it to disable any buttons if `clickedButtonIndex` state is set

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessage.vue
```diff
@@ -384,6 +393,12 @@ onBeforeMount(() => {
 					<div v-if="message.type === 'human'">
 						{{ text }}
 					</div>
+					<ChatMessageWithButtons
```
autologie: I feel it's simpler to let ChatMarkdownChunk render `with-buttons` too with optional buttons section after `<VueMarkdown />`

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessageWithButtons.vue
autologie: yes, that would make the intention clearer

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessage.vue
Cadiac: fair, I'll take a look at changing the implementation to be like that

packages/@n8n/chat-hub/src/parser.ts
Cadiac: this JSON is always in one piece, it's the Respond to Chat node that returns with this special JSON instead of text, and that node cannot be used in streaming mode - in nonstreaming mdoes we still send "chunks", but they are always full messages.

packages/@n8n/chat-hub/src/parser.ts
Cadiac: I think this is fine here, given that these buttons JSON replies are never actually chunked

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessage.vue
Cadiac: I changed the implementation, ChatMarkdownChunk renders buttons if the source it receives has type `with-buttons`, but it now also needs to receive `isButtonsDisabled` prop (based on the message's status) to disable the buttons if the message isn't in waiting state.. I think the end result is still better?

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatMessageWithButtons.vue
Cadiac: Made this component (now called ChatButtons.vue) receive explicit `isDisabled` prop & cleaned up this logic

autologie (approved):

PR #25353, opened 2026-02-05
fix(core): Implement evaluation run cancellation for multi-main

schrothbn:
## Summary
Adds multi-main mode support for test run cancellation in the evaluation system. Previously, cancelling a test run in multi-main mode threw a NotImplementedError.

The implementation uses a dual approach:

- Pub/sub messaging (cancel-test-run command) for immediate cancellation across all main instances
- Database flag (cancelRequested column) as a fallback mechanism, polled during the test case loop in multi-main mode

A runningInstanceId column is also added to track which instance is running a given test run.

### How it works
- When cancellation is requested, a cancelRequested flag is set in the database
- Local cancellation is attempted via the abort controller
- In multi-main/queue mode, a cancel-test-run command is broadcast via pub/sub
- If the test run isn’t running locally, it’s marked as cancelled in the database as a fallback
- The running instance also polls the database flag between test cases (multi-main only)

https://linear.app/n8n/issue/AI-1845/tech-debt-scaling-mode-support-for-evaluation-nodes

## Review / Merge checklist

- [x] Tests included. 


OlegIvaniv (dismissed):

despairblue (dismissed):
added a test for the migration, updated the migration to match the schema, renamed and re-ordered the migration to be at the end of the list

only approving the migration. haven't looked at the rest of the code

despairblue (dismissed):

despairblue (dismissed):

despairblue (approved):

PR #25355, opened 2026-02-05
chore(API): Add DTOs for credential endpoints

Jubke:
## Summary

Add Data Transfer Objects (DTOs) for credential-related API endpoints to improve code organization and type safety. Replaces inline Zod validation in the controller with dedicated DTO classes in `@n8n/api-types`, providing better reusability across the codebase.

**Changes include:**
- 5 new DTOs: `CredentialsForWorkflowQueryDto`, `ShareCredentialsBodyDto`, `TestCredentialsBodyDto`, `TransferCredentialBodyDto`, `UpdateCredentialDto`
- 572 lines of comprehensive test coverage for all DTOs
- Improved type safety by using `z.unknown()` instead of `z.any()` where possible (there are some exceptions).
- Controller methods updated to use `@Body` and `@Query` decorators
- All existing tests pass (96/96)

https://linear.app/n8n/issue/CAT-614

## Review / Merge checklist

- [x] Tests included (572 lines of test coverage across 5 DTOs)
- [x] Type safety validated (`pnpm typecheck` passes)
- [x] All tests passing (96 tests)

### DIFF
--- a/packages/@n8n/api-types/src/dto/credentials/__tests__/credentials-for-workflow-query.dto.test.ts
+++ b/packages/@n8n/api-types/src/dto/credentials/__tests__/credentials-for-workflow-query.dto.test.ts
@@ -1,0 +1,65 @@
+import { CredentialsForWorkflowQueryDto } from '../credentials-for-workflow-query.dto';
+
+describe('CredentialsForWorkflowQueryDto', () => {
+	describe('Valid requests', () => {
+		test.each([
+			{
+				name: 'with workflowId',
+				request: {
+					workflowId: 'workflow-123',
+				},
+			},
+			{
+				name: 'with projectId',
+				request: {
+					projectId: 'project-123',
+				},
+			},
+			{
+				name: 'with both workflowId and projectId',
+				request: {
+					workflowId: 'workflow-123',
+					projectId: 'project-123',
+				},
+			},
+		])('should validate $name', ({ request }) => {
+			const result = CredentialsForWorkflowQueryDto.safeParse(request);
+			expect(result.success).toBe(true);
+		});
+	});
+
+	describe('Invalid requests', () => {
+		test.each([
+			{
+				name: 'missing both workflowId and projectId',
+				request: {},
+			},
+			{
+				name: 'empty object with undefined values',
+				request: {
+					workflowId: undefined,
+					projectId: undefined,
+				},
+			},
+		])('should fail validation for $name', ({ request }) => {
+			const result = CredentialsForWorkflowQueryDto.safeParse(request);
+			expect(result.success).toBe(false);
+		});
+
+		test('should fail validation when workflowId is not a string', () => {
+			const result = CredentialsForWorkflowQueryDto.safeParse({
+				workflowId: 123,
+			});
+			expect(result.success).toBe(false);
+			expect(result.error?.issues[0].path).toEqual(['workflowId']);
+		});
+
+		test('should fail validation when projectId is not a string', () => {
+			const result = CredentialsForWorkflowQueryDto.safeParse({
+				projectId: 123,
+			});
+			expect(result.success).toBe(false);
+			expect(result.error?.issues[0].path).toEqual(['projectId']);
+		});
+	});
+});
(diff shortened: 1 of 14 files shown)


packages/@n8n/api-types/src/dto/credentials/test-credentials-body.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+	type: z.enum(['personal', 'team', 'public']),
+	createdAt: z.string(),
+	updatedAt: z.string(),
+});
+
+const credentialsDecryptedSchema = z.object({
+	id: z.string(),
+	name: z.string().optional(),
+	type: z.string(),
+	data: z.record(z.string(), z.any()).optional(),
```
ivov, line 25: PR description says we're switching to `z.unknown`?

packages/@n8n/api-types/src/dto/credentials/update-credential.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod';
+import { Z } from 'zod-class';
+
+export class UpdateCredentialDto extends Z.class({
+	name: z.string().min(1).max(128).optional(),
+	type: z.string().min(1).max(128).optional(),
+	data: z.record(z.string(), z.any()).optional(),
```
ivov, line 8: `z.unknown`?

packages/cli/src/credentials/credentials.controller.ts
```diff
@@ -158,7 +181,7 @@ export class CredentialsController {
 			req.user,
 			storedCredential,
 			decryptedData,
-			mergedCredentials,
+			mergedCredentials as ICredentialsDecrypted,
```
ivov: TS here is hinting the new DTO makes `name` optional, but the original requires it. Curious why the change? From a very quick check it looks like the client always sends the name.

packages/cli/src/credentials/credentials.controller.ts
```diff
@@ -85,11 +89,24 @@ export class CredentialsController {
...
+				workflowId: query.workflowId,
+			});
+		}
+		if (query.projectId !== undefined) {
+			return await this.credentialsService.getCredentialsAUserCanUseInAWorkflow(req.user, {
+				projectId: query.projectId,
+			});
+		}
+		// Should never reach here due to DTO validation
+		throw new BadRequestError('Either workflowId or projectId must be provided');
```
ivov: Is this an improvement? I see why we can't use a DTO here, but the inline `z.union` is more compact and readable no?

packages/@n8n/api-types/src/dto/credentials/__tests__/credentials-for-workflow-query.dto.test.ts
ivov, line 1: This PR is only following the existing pattern, but I wonder if we should start getting stricter about what we add tests for, the bulk of these are just verifying that Zod does what Zod does.

ivov (commented):
Thanks for tackling this

packages/@n8n/api-types/src/dto/credentials/test-credentials-body.dto.ts
Jubke: Initially switched to `z.unknown()` but reverted because `{ [x: string]: unknown }` is not assignable to `ICredentialDataDecryptedObject`. Updated the description.

packages/@n8n/api-types/src/dto/credentials/update-credential.dto.ts
Jubke: see comment above

packages/cli/src/credentials/credentials.controller.ts
Jubke: 👍 Integration tests were not sending `name`. But the right fix was to adjust tests not expectations.

packages/cli/src/credentials/credentials.controller.ts
Jubke: CC accommodating for an unwanted type conversion using `as`. Reverted to keep it cleaner as DTO ensures typing is correct.

packages/@n8n/api-types/src/dto/credentials/__tests__/credentials-for-workflow-query.dto.test.ts
Jubke: Generally agree, since this was more of an exercise for me and I lack context of the broader DTO migration scope, I'll leave that call to you.

PR #25366, opened 2026-02-05
feat(core): Redact/Unredact settings on GET / PATCH of secret providers

geemanjs:
## Summary
Mimic redaction behaviour of settings to the original implementation

- Extracted a common RedactionService shared between the original / new implementation
- Add specific tests for the RedactionService

Closed LIGO-192

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/external-secrets.ee/secrets-providers.responses.ee.ts
```diff
@@ -1,12 +1,19 @@
...
 		'settings' | 'secretsCount' | 'state' | 'isEnabled'
 	>;
 
-	type PublicConnection = Promise<StrippedConnection>;
-	type PublicConnectionList = Promise<StrippedConnection[]>;
+	// Full type for detail responses (includes redacted settings)
+	type Connection = Omit<SecretProviderConnection, 'secretsCount' | 'state' | 'isEnabled'>;
+
+	// Deprecated: Alias to ConnectionListItem for backward compatibility
+	type StrippedConnection = ConnectionListItem;
```
ireneea: [nit] do yo uknow if tsdoc deprecation tag works on `type`?

```suggestion
	/** @deprecated: use ConnectionListItem instead **/
	type StrippedConnection = ConnectionListItem;
```

geemanjs: No idea - I'll accept the suggestion and we can find out - I will add a fastfollow to change these. They've been on my nit list for a while to move into the api types.

packages/cli/src/modules/external-secrets.ee/redaction.service.ee.ts
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	/**
+	 * Determines if a value should be redacted
+	 * @param value - Value to check
+	 * @returns True if value should be redacted
+	 */
+	private shouldRedactValue(value: unknown): boolean {
+		// Only redact string values that are not expressions (starting with '=')
+		return typeof value === 'string' && !value.startsWith('=');
```
ireneea, line 55: Should we be more precise in checking if a value is an expression? Could we not have a value that genuinely start with `=` but that is not an expression?

konstantintieber: Also, as far as I know, values that contain expressions are also sent in the data payload starting with `={{ $vars... }}`.
I would assume that that's also how they're stored in the backend. Didn't verify this.

konstantintieber: But as I understand this logic wasn't invented by @geemanjs, but originates from https://github.com/n8n-io/n8n/blob/master/packages/cli/src/modules/external-secrets.ee/external-secrets.service.ee.ts#L70

So I guess it should work 🤷

geemanjs: Correct - I just copied what was already there - Should we maybe lock down to a regex? that matches `={{<anything>}}`

```
^={{ ?(.*?) ?}}$
```

ireneea: Looks good to me but I'm in no way a regex wizzard but found two example in the source code if this can help

```typescript
// packages/cli/src/modules/source-control.ee/source-control-helper.ee.ts
export function stringContainsExpression(testString: string): boolean {
	return /^=.*\{\{.*\}\}/.test(testString);
}
```

```typescript
// packages/@n8n/ai-workflow-builder.ee/src/validation/utils/expressions.ts

import { isExpression } from 'n8n-workflow';

export function containsExpression(value: unknown): boolean {
	if (!isExpression(value)) {
		return false;
	}

	return /\{\{.*(\$\(.*?\))|(\$\w+).*}}/.test(value);
}
```

geemanjs: I am surprised we don't have a "set standard" for this 😋  The first one looks "looser" and probably suitable for this use-case. I don't really want cross module import though so do I duplicate or do we create a function somewhere else?

ireneea: Yeah I would just duplicate and ship it 🚀

konstantintieber (dismissed):
Tested manually, works as expected.
Nice job on organizing the existing redaction code in a cleaner way.,

konstantintieber (approved):

PR #25380, opened 2026-02-05
test: Add E2E tests for internal MCP service

aalises:
## Summary

Adds comprehensive E2E tests for the internal MCP service (`/mcp-server/http`).

## How the tests work

- Tests communicate with the MCP service using JSON-RPC messages over HTTP
- Authentication uses Bearer tokens (MCP API key)
- Tests run **serially** because n8n only supports one MCP API key at a time - rotation invalidates the previous key, so parallel execution would cause race conditions
- Each test uses `rotateMcpApiKey()` to get a fresh, usable key (since `getMcpApiKey()` returns REDACTED after the first call)
- Test workflows are imported from JSON fixtures in `workflows/mcp-service/`

### Authentication (4 tests)
- Rejects requests without bearer token (401)
- Rejects requests with invalid bearer token (401)
- Accepts valid API key
- Rejects old key after rotation, accepts new key

### MCP Settings (1 test)
- Rejects requests when MCP access is disabled (403)

### tools/list (2 tests)
- Returns all 3 built-in tools (`search_workflows`, `get_workflow_details`, `execute_workflow`)
- Includes proper tool descriptions and input schemas

### search_workflows (5 tests)
- Returns workflows marked as available in MCP
- Does NOT return workflows not marked as available in MCP
- Supports `limit` parameter
- Supports `query` filter for name search
- Returns workflow metadata (id, name, nodes, scopes, canExecute)

### get_workflow_details (4 tests)
- Returns detailed info for accessible workflow (nodes, connections, settings, scopes)
- Returns error for non-existent workflow
- Returns error for workflow not available in MCP
- Includes trigger info in response

### execute_workflow (4 tests)
- Executes workflow successfully and returns executionId
- Returns error for non-existent workflow
- Returns error for workflow not available in MCP
- Executes webhook workflow with input data

### Error Handling (3 tests)
- Handles malformed JSON-RPC messages (missing jsonrpc version)
- Handles unknown methods
- Handles invalid tool parameters (wrong types)

--
https://linear.app/n8n/issue/AI-2002/create-e2e-tests-for-the-mcp-service


packages/testing/playwright/tests/e2e/mcp/mcp-service.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+
+			const malformedMessage = {
+				id: nanoid(),
+				method: 'tools/list',
+			};
+
+			const response = await api.mcp.internalMcpSendMessage(apiKey, malformedMessage);
+
+			const body = await response.text();
+			expect(body).toBeTruthy();
```
shortstacked: This doesn't check the status code, so it could have errored out here right. We should use a more solid assert to make sure it contains what we expect.

packages/testing/playwright/tests/e2e/mcp/mcp-service.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+		});
+
+		test('should handle unknown methods', async ({ api }) => {
+			const { apiKey } = await api.rotateMcpApiKey();
+
+			const message = api.mcp.createMessage('unknown/method');
+			const response = await api.mcp.internalMcpSendMessage(apiKey, message);
+
+			const body = await response.text();
```
shortstacked: Same as the above here.

packages/testing/playwright/tests/e2e/mcp/mcp-service.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+
+			const message = api.mcp.createMessage('tools/call', {
+				name: 'search_workflows',
+				arguments: {
+					limit: 'not-a-number',
+				},
+			});
+
+			const response = await api.mcp.internalMcpSendMessage(apiKey, message);
+			const body = await response.text();
```
shortstacked: And same again

shortstacked (changes requested):
The negative asserts need to be stronger because they just check if the message returns but it could return anything. Also it looks like we've added a lot of unused methods. Overall approach is solid though!

shortstacked (approved):

PR #25384, opened 2026-02-05
feat: Add projects to credentials list response

BGZStephen:
## Summary

This PR adds a concept of projects to credentials list endpoint, allowing people to see which projects a credential is available to

image

IAM-172

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/public-api/v1/handlers/credentials/__tests__/credentials.service.test.ts
```diff
@@ -9,13 +9,89 @@ import type { GenericValue, IDataObject, INodeProperties } from 'n8n-workflow';
...
+
+		it('returns multiple shared entries and skips shared entries without project', () => {
+			const createdAt1 = new Date('2024-01-01T00:00:00.000Z');
+			const updatedAt1 = new Date('2024-01-02T00:00:00.000Z');
+			const createdAt2 = new Date('2024-02-01T00:00:00.000Z');
+			const updatedAt2 = new Date('2024-02-02T00:00:00.000Z');
+			const credential = {
+				shared: [
+					{
+						role: 'credential:owner',
```
phyllis-noester, line 52: does 'credential:owner' also show up as shared? how would it then be possible for a credential not to be shared at all? I assumed it would at least have to be shared with owner

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
```diff
@@ -45,17 +46,41 @@ export = {
...
 			const offset = Number(req.query.offset) || 0;
 			const limit = Math.min(Number(req.query.limit) || 100, 250);
 
-			const [credentials, count] = await Container.get(CredentialsRepository).findManyAndCount({
+			const repo = Container.get(CredentialsRepository);
+			const [credentials, count] = await repo.findAndCount({
 				take: limit,
 				skip: offset,
+				select: ['id', 'name', 'type', 'createdAt', 'updatedAt'],
+				relations: ['shared', 'shared.project'],
```
phyllis-noester, line 69: what happens in the scenario that I'm allowed to list the credential but not allowed to access `shared.project`?

BGZStephen: `credentials:list` doesn't provide access to the credential and it's data, it only allows you to list ID and name of a credential (and project it is shared with) so I don't think this relevant

packages/cli/src/public-api/v1/handlers/credentials/__tests__/credentials.service.test.ts
BGZStephen: Every credenttial will have at least one entry for the shared which shows the `credential:owner`, as credentials are not shared with `user`, they are shared with `projects`, it will show up as `credential:ownerr` for the users personal project, which that user is an admin of and can use that credential.

phyllis-noester: that's what I assumed. I was asking because of the test `'returns empty array when credential has no shared'`

BGZStephen: Yeah, so I guess technically it could happen, the unit tests lack the context of the business logic that means this shouldn't happen, let me remove it as it won't really affect the code coverage

phyllis-noester (approved):

PR #25388, opened 2026-02-05
feat: Add example node in ai-utilities, improve response handling

yehorkardash:
## Summary

This PR adds a pair of nodes that can be used for testing in n8n.
It also updates types that generate() and stream() return to support a full range of responses instead of just text

изображение

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-utilities/examples/models/openai.ts
```diff
@@ -110,68 +113,30 @@ export interface OpenAIErrorResponse {
...
+		'Content-Type': 'application/json',
 	});
-
-	if (!response.ok) {
-		const errorData = (await response.json().catch(() => null)) as OpenAIErrorResponse | null;
-		const errorMessage = errorData?.error?.message ?? (await response.text());
-		throw new Error(
-			`OpenAI API request failed: ${response.status} ${response.statusText}\n${errorMessage}`,
-		);
-	}
```
DawidMyslak, line 138: Just to double check, we removed the error handling logic here. How this will impact the logic flow in case of any error?

packages/@n8n/ai-utilities/src/adapters/langchain-chat-model.ts
```diff
@@ -207,7 +170,51 @@ export class LangchainAdapter<
...
+					}),
+					text: '',
+					generationInfo: {
+						finish_reason: 'error',
+						error: chunk.error,
+					},
+				});
+			} else if (chunk.type === 'content') {
+				const lcMessage = toLcMessage({
+					role: 'ai',
```
DawidMyslak: Will it always be `ai`?

packages/@n8n/ai-utilities/src/converters/message.ts
```diff
@@ -309,7 +309,7 @@ function toLcContent(block: N8nMessages.MessageContent): LangchainMessages.Conte
 			value: block.value,
 		} as LangchainMessages.ContentBlock.NonStandard;
 	}
-	return null;
+	throw new Error(`Failed to convert to Langchain content block: ${JSON.stringify(block)}`);
 }
```
DawidMyslak, line 313: Since now this can throw, how will happen with this part of the code inside `toLcMessage`:
```
	const lcContent = message.content
		.map(toLcContent)
		.filter((c): c is LangchainMessages.ContentBlock => c !== null);
```

packages/@n8n/ai-utilities/src/examples/models/openai.ts
```diff
@@ -519,15 +520,15 @@ export class OpenAIChatModel extends BaseChatModel<OpenAIChatModelConfig> {
 						.filter(Boolean)
 						.join('');
 					if (reasoningText) {
-						yield { type: 'text-delta', textDelta: reasoningText };
+						yield { type: 'text-delta', delta: reasoningText };
```
DawidMyslak: Should this be `reasoning-delta`?

packages/@n8n/ai-utilities/src/types/output.ts
```diff
@@ -18,33 +17,49 @@ export type TokenUsage<T extends Record<string, unknown> = Record<string, unknow
...
-}
+export type StreamChunk = ContentMetadata &
+	(
+		| {
+				type: 'text-delta';
+				id?: string;
+				delta: string;
+		  }
+		| {
+				type: 'reasoning-delta';
```
DawidMyslak, line 41: I think this is not used anywhere yet, added another comment for it.

packages/@n8n/ai-utilities/examples/run.ts
```diff
@@ -1,9 +1,14 @@
+/**
```
DawidMyslak, line 1: Nit: Why `src` should handle examples?

packages/@n8n/ai-utilities/tsconfig.examples.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": ["./tsconfig.json", "@n8n/typescript-config/tsconfig.build.json"],
+	"compilerOptions": {
+		"composite": true,
+		"rootDir": "src",
+		"outDir": "dist",
+		"tsBuildInfoFile": "dist/build.tsbuildinfo"
+	},
+	"include": ["src/**/*.ts", "src/examples/**/*.ts"],
```
DawidMyslak: `src/**/*.ts` already covers `src/examples/**/*.ts`

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiCustom/LmChatOpenAiCustom.node.ts
```diff
@@ -0,0 +1,126 @@
...
+	type IDataObject,
+	type INodeType,
+	type INodeTypeDescription,
+	type ISupplyDataFunctions,
+	type SupplyData,
+} from 'n8n-workflow';
+import type Stream from 'node:stream';
+import { Readable } from 'node:stream';
+
+import { supplyModel } from 'src/suppliers/supplyModel';
```
DawidMyslak, line 12: Should we add some TODO/ticket to update those paths to point to SDK at some point?

DawidMyslak (changes requested):
Minor questions 🙏 . Looking good overall!

DawidMyslak (changes requested):
Minor questions 🙏 . Looking good overall!

packages/@n8n/ai-utilities/examples/models/openai.ts
yehorkardash: Throwing is handled inside httpRequest() function and then error is handled inside of langchain. Forgot to change it inside run.ts

packages/@n8n/ai-utilities/src/adapters/langchain-chat-model.ts
yehorkardash: Yes, in this method we are already returning only AIMessageChunk. I use `toLcMessage` here to mostly format the chunk content

packages/@n8n/ai-utilities/src/converters/message.ts
yehorkardash: Removed it 👍
What do you think is better to do here. I went with the errors because it's more explicit behavior that will let us and developer know if something is wrong with our/their logic

packages/@n8n/ai-utilities/src/examples/models/openai.ts
yehorkardash: Good catch

packages/@n8n/ai-utilities/src/types/output.ts
yehorkardash: Updated example to use it

packages/@n8n/ai-utilities/examples/run.ts
yehorkardash: In another PR I'm adding an integration test that uses openai model from examples, so I thought it made sense to include it in source folder. Do you think it would be better to move it and the test outside of src folder?
https://github.com/n8n-io/n8n/pull/25443

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiCustom/LmChatOpenAiCustom.node.ts
yehorkardash: This example is inside the package, so I don't think it could point to itself. I'm planning to move those examples to node-cli at some point

packages/@n8n/ai-utilities/examples/run.ts
DawidMyslak: Yes, my personal take is that `src` should be reserved for source files. Having examples on the root also helps with discovery.

packages/@n8n/ai-utilities/src/converters/message.ts
DawidMyslak: Indeed we need to think about SDK users here. Previous code would silently continue, this approach makes more sense, good call on your end 👍.

DawidMyslak (dismissed):
Approving, but this comment still stands:
https://github.com/n8n-io/n8n/pull/25388#discussion_r2787597356

DawidMyslak (approved):

PR #25392, opened 2026-02-05
feat(core): Add memory abstractions to AI utilities SDK

DawidMyslak:
## Summary

Adds Memory SDK to `@n8n/ai-utilities` so community node developers can build AI memory nodes without direct LangChain dependencies.

**Key additions:**
- **Two-layer architecture**: `ChatStorage` + `ChatMemory` interfaces with `BaseChatStorage` / `BaseChatMemory` base classes
- **`WindowedChatMemory`**: built-in implementation keeping last N message pairs in context
- **`supplyMemory()`**: supplier function matching the `supplyModel()` pattern, wrapping any `ChatMemory` via `LangchainMemoryAdapter` + `logWrapper`
- **`toLcMessage()`**: new n8n-to-LangChain message converter covering all content types (text, reasoning, file, tool-call, tool-result, citation, provider)
- **README**: Memory SDK docs with architecture overview, core patterns, and full community node example

**Tests:** message round-trip, windowed memory logic, and LangChain adapter integration.

https://linear.app/n8n/issue/NODE-4371

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/ai-utilities/src/adapters/langchain-memory.ts
```diff
@@ -0,0 +1,50 @@
...
+	override chatHistory: LangchainBaseChatMessageHistory;
+
+	constructor(memory: ChatMemory) {
+		super({
+			returnMessages: true,
+			inputKey: 'input',
+			outputKey: 'output',
+		});
+		this.n8nMemory = memory;
+		this.chatHistory = new LangchainChatMessageHistory();
```
yehorkardash: Message history should be supplied by developers. Also, we should avoid using in-memory stores because of multi-main.


The way Langchain handles memory and chat history is that chat history is supplied as parameter to memory and then it's used by memory to save and load messages. I would suggest using the similar architecture. What do you think?

Here's how langchain handles that:
```typescript
interface BaseChatMemoryInput {
  chatHistory?: BaseChatMessageHistory;
  returnMessages?: boolean;
  inputKey?: string;
  outputKey?: string;
}
interface BufferWindowMemoryInput extends BaseChatMemoryInput {
  humanPrefix?: string;
  aiPrefix?: string;
  memoryKey?: string;
  k?: number;
}
// https://github.com/langchain-ai/langchainjs/blob/999497accfc9f5b9b001e66ab5941e1b43a829ad/libs/langchain-classic/src/memory/buffer_window_memory.ts
// https://github.com/langchain-ai/langchainjs/blob/999497accfc9f5b9b001e66ab5941e1b43a829ad/libs/langchain-classic/src/memory/chat_memory.ts
// saving messages is handled in BaseChatMemory
class BufferWindowMemory extends BaseChatMemory {
...
  loadMemoryVariables() {
    const messages = await this.chatHistory.getMessages();
    const result = {
        [this.memoryKey]: messages.slice(-this.k * 2),
      };
    return result;
  }
}
```

I saw that we are using this approach in postgres.ts and redis.ts, but not in base implementation and langchain-memory.ts

packages/@n8n/ai-utilities/src/converters/message.ts
```diff
@@ -219,3 +220,33 @@ export function fromLcMessage(msg: LangchainMessages.BaseMessage): N8nMessages.M
 	}
 	throw new Error(`Provided message is not a valid Langchain message: ${JSON.stringify(msg)}`);
 }
+
+export function toLcMessage(message: Message): LangchainMessages.BaseMessage {
```
yehorkardash, line 315: I think this method should support all types of content blocks and try to map n8n messages to langchain messages as close as possible, using the reverse of `fromLcContent`

packages/@n8n/ai-utilities/src/types/memory.ts
```diff
@@ -0,0 +1,20 @@
+import type { Message } from './message';
+
+export interface ChatHistory {
+	getMessages(): Promise<Message[]>;
+	addMessage(message: Message): Promise<void>;
+	addMessages(messages: Message[]): Promise<void>;
+	clear(): Promise<void>;
+}
+
+export interface MemoryConfig {
```
yehorkardash: Langchain has `humanPrefix` and `aiPrefix`. Should we have them too?

packages/@n8n/ai-utilities/src/memory/in-memory-chat-history.ts
```diff
@@ -0,0 +1,101 @@
+import { BaseChatHistory } from './base-chat-history';
+import type { Message } from '../types/message';
+
+class InMemoryChatHistoryManager {
```
yehorkardash: Do we have a use for in memory chat history somewhere? It's generally discouraged because of multi-main quirks

packages/@n8n/ai-utilities/src/memory/windowed-chat-memory.ts
```diff
@@ -0,0 +1,58 @@
+import { BaseChatMemory } from './base-chat-memory';
+import type { ChatHistory, MemoryConfig } from '../types/memory';
+import type { Message } from '../types/message';
+
+export interface WindowedChatMemoryConfig extends MemoryConfig {
```
yehorkardash: In nodes-langchain we use BufferMemory and WindowBufferMemory, we should add BufferMemory later too

packages/@n8n/ai-utilities/src/adapters/langchain-memory.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	get memoryKeys(): string[] {
+		return ['chat_history'];
+	}
+
+	async loadMemoryVariables(_values: InputValues): Promise<MemoryVariables> {
+		const messages = await this.n8nMemory.loadMessages();
+		const lcMessages = messages.map(toLcMessage);
+
+		await this.chatHistory.clear();
```
yehorkardash: why do we call clear()?

packages/@n8n/ai-utilities/src/converters/message.ts
```diff
@@ -219,3 +220,33 @@ export function fromLcMessage(msg: LangchainMessages.BaseMessage): N8nMessages.M
...
+	switch (message.role) {
+		case 'system':
+			return new LangchainMessages.SystemMessage(text);
+		case 'human':
+			return new LangchainMessages.HumanMessage(text);
+		case 'ai':
+			return new LangchainMessages.AIMessage(text);
+		case 'tool': {
+			const toolResult = message.content.find((c) => c.type === 'tool-result');
+			const toolCallId =
```
yehorkardash: We know the type here, do we need to do  `'toolCallId' in toolResult` and assign default values?
Our message type has `toolCallId` and `result` as required. We can also map tool status from `isError`

packages/@n8n/ai-utilities/src/adapters/langchain-memory.ts
DawidMyslak: Good catch on multi-main! Let me refactor this.

packages/@n8n/ai-utilities/src/adapters/langchain-memory.ts
DawidMyslak: The `clear()` is needed because `this.chatHistory` is a separate in-memory store created in the constructor (line 21). Without clearing, messages would accumulate across multiple `loadMemoryVariables()` calls.

Should be gone after refactoring.

packages/@n8n/ai-utilities/src/types/memory.ts
DawidMyslak: Correct me if I'm wrong, but looks like `humanPrefix/aiPrefix` properties are only used with `returnMessages: false`. Our SDK always works with structured `Message[]` (`role` included) objects. We set `returnMessages: true` in LangchainMemoryAdapter constructor - so we never hit the string formatting path. We could skip for now.

yehorkardash: True, didn't notice it. Seems to be a legacy option

packages/@n8n/ai-utilities/src/memory/in-memory-chat-history.ts
DawidMyslak: Will remove.

packages/@n8n/ai-utilities/src/converters/message.ts
DawidMyslak: Let ma also add a "reversed" test, so calling `from/to` should give us the same initial object.

packages/@n8n/ai-utilities/src/__tests__/converters/message-roundtrip.test.ts
```diff
@@ -0,0 +1,453 @@
...
+import { fromLcMessage, toLcMessage } from '../../converters/message';
+import type { Message } from '../../types/message';
+
+/**
+ * Round-trip tests: n8n -> LC -> n8n (toLcMessage then fromLcMessage)
+ *
+ * Verifies that converting an n8n Message to LangChain and back
+ * produces the same message.
+ */
+describe('message round-trip: n8n -> LC -> n8n', () => {
```
yehorkardash, line 18: Great test 👏

packages/@n8n/ai-utilities/src/memory/windowed-chat-memory.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/** Keeps only the last N message pairs in context. */
+export class WindowedChatMemory extends BaseChatMemory {
+	private readonly _chatHistory: ChatHistory;
+	private readonly windowSize: number;
+
+	constructor(chatHistory: ChatHistory, config?: WindowedChatMemoryConfig) {
+		super();
+		this._chatHistory = chatHistory;
```
yehorkardash: nitpick: I don't think we use `_` for private members across the codebase

pnpm-lock.yaml
```diff
@@ -242,6 +242,9 @@ catalogs:
     unplugin-icons:
       specifier: ^0.19.0
       version: 0.19.0
+    vite-svg-loader:
```
yehorkardash, line 245: I don't see changes in package.json, probably be safer to revert the lock file

DawidMyslak: Really weird it's still in GH diff, I reverted this already. Can you double check?

packages/@n8n/ai-utilities/src/memory/windowed-chat-memory.ts
DawidMyslak: Ah, that was initially coming from the `README.md` example for the community 👍. Let me clean this up.

yehorkardash (approved):
LGTM ✅

pnpm-lock.yaml
yehorkardash: Yes, seems to be gone now

PR #25393, opened 2026-02-05
refactor(core): Replace `zod-class` with minimal implementation

ivov:
## Summary

This PR replaces `zod-class` with a minimal implementation, to unblock migrating to Zod 4. v4 brings significant [performance improvements](https://zod.dev/v4#benchmarks), but our dep `zod-class` is incompatible with v4 as it relies on v3 internals, and we only need minimal functionality from this dep for our DTO setup. 

Side note - Two spots used DTO classes directly as Zod types. `zod-class` made DTO classes behave as full Zod schemas via proxy magic. This minimal implementation does not replicate this, instead these two usages now access the `.schema` property explicitly:

- `z.array(CreateDataTableColumnDto)` → `z.array(CreateDataTableColumnDto.schema)`
- `SamlPreferencesAttributeMapping.optional()` →
  `SamlPreferencesAttributeMapping.schema.optional()`

https://linear.app/n8n/issue/CAT-857

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/chat-hub/chat-hub.models.service.ts
+++ b/packages/cli/src/modules/chat-hub/chat-hub.models.service.ts
@@ -40,7 +40,7 @@ export class ChatHubModelsService {
 
 	async getModels(
 		user: User,
-		credentialIds: Record<ChatHubLLMProvider, string | null>,
+		credentialIds: Partial<Record<ChatHubProvider, string | null>>,
 	): Promise<ChatModelsResponse> {
 		const additionalData = await getBase({ userId: user.id });
 		const providers = chatHubProviderSchema.options;
--- a/packages/frontend/editor-ui/src/features/execution/insights/insights.store.test.ts
+++ b/packages/frontend/editor-ui/src/features/execution/insights/insights.store.test.ts
@@ -16,7 +16,7 @@ vi.mock('vue-router', async (importOriginal) => ({
 }));
 vi.mock('@/features/execution/insights/insights.api');
 
-const mockFilter = { dateRange: 'week' as const };
+const mockFilter = { projectId: 'test-project' };
 const mockData = [
 	{
 		date: '2023-01-01',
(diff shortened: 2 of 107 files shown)


packages/cli/src/modules/chat-hub/chat-hub.models.service.ts
```diff
@@ -40,7 +40,7 @@ export class ChatHubModelsService {
 
 	async getModels(
 		user: User,
-		credentialIds: Record<ChatHubLLMProvider, string | null>,
+		credentialIds: Partial<Record<ChatHubProvider, string | null>>,
```
ivov, line 43: `zod-class` was masking two type mismatches:
- The schema uses `ChatHubProvider` (all providers) but the service expected `ChatHubLLMProvider` (LLM providers only)
- Standard Zod produces `Partial<Record<K, V>>` for `z.record()` but the custom type setup in `zod-class` produced `Record<K, V>`.

packages/frontend/editor-ui/src/features/execution/insights/insights.store.test.ts
```diff
@@ -16,7 +16,7 @@ vi.mock('vue-router', async (importOriginal) => ({
 }));
 vi.mock('@/features/execution/insights/insights.api');
 
-const mockFilter = { dateRange: 'week' as const };
+const mockFilter = { projectId: 'test-project' };
```
shortstacked, line 19: This seems to change the filter completely, is that intentional?

shortstacked (approved):
One small comment on the filter change, other than that looks good.

packages/frontend/editor-ui/src/features/execution/insights/insights.store.test.ts
ivov: Yup, the old `zod-class` was buggy. With the new setup, the type is properly derived from `z.object(shape)` so TS now rejects `{ dateRange: 'week' }` as that property doesn't exist on `InsightsDateFilterDto`. `projectId` does exist and the test itself doesn't care about the filter, it just forwards it unchanged, so any valid filter value works

PR #25398, opened 2026-02-05
feat: Add endpoint to create Quick Connect credentials from the backend

RomanDavydchuk:
## Summary

Add an endpoint that receives a credential type, optional project ID and creates a credential of that type using Quick Connect (if it's supported). Add corresponding controller, service, abstractions, errors. Also update the config shape and add tests. This PR also includes a `SampleHandler` that can be used for local testing. This will be removed once actual handlers are implemented

https://linear.app/n8n/issue/NODE-4380/implement-infrastructure-for-backend-based-flows

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/quick-connect/quick-connect.module.ts
```diff
@@ -4,14 +4,35 @@ import { Container } from '@n8n/di';
...
 	 */
 	async settings() {
 		const { QuickConnectConfig } = await import('./quick-connect.config');
 		const { options } = Container.get(QuickConnectConfig);
-		return { options };
+		// Strip secret from backendFlowConfig before sending to frontend
+		return {
+			options: options.map(({ backendFlowConfig, ...rest }) => ({
+				...rest,
+				backendFlowConfig: backendFlowConfig
```
dlavrenuek: We could also move the `consentText` out of `backendFlowConfig` and simply not send `backendFlowConfig` to the frontend. Will make the separation of backend only configuration a bit clearer

elsmr (approved):

PR #25403, opened 2026-02-05
test: Improve memory profiling in performance tests

shortstacked:
## Summary

Adds a garbage collection trigger endpoint to the CLI, enabling manual GC during memory profiling.

Configures n8n testing container to expose the V8 garbage collector, essential for controlled memory analysis.

Updates memory consumption tests to use the GC endpoint and a more reliable memory stabilization method, ensuring accurate heap measurements.

Introduces a new memory retention test to detect potential memory leaks by measuring heap usage before and after specific actions.

These improvements facilitate more accurate and reliable memory profiling, enabling identification and resolution of memory-related issues.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ packages/cli/src/commands/export/outputs
 .data
 .data/
 .claude/settings.local.json
+lefthook-local.yml
--- a/packages/cli/src/controllers/e2e.controller.ts
+++ b/packages/cli/src/controllers/e2e.controller.ts
@@ -264,6 +264,22 @@ export class E2EController {
 		};
 	}
 
+	/**
+	 * Trigger garbage collection for memory profiling in performance tests.
+	 * Requires Node.js to be started with --expose-gc flag.
+	 */
+	@Post('/gc', { skipAuth: true })
+	triggerGarbageCollection() {
+		if (typeof global.gc === 'function') {
+			global.gc();
+			return { success: true, message: 'Garbage collection triggered' };
+		}
+		return {
+			success: false,
+			message: 'Garbage collection not available. Ensure Node.js is started with --expose-gc flag.',
+		};
+	}
+
 	private resetFeatures() {
 		for (const feature of Object.keys(this.enabledFeatures)) {
 			this.enabledFeatures[feature as BooleanLicenseFeature] = false;
--- a/packages/cli/src/controllers/e2e.controller.ts
+++ b/packages/cli/src/controllers/e2e.controller.ts
@@ -271,6 +271,8 @@ export class E2EController {
 	@Post('/gc', { skipAuth: true })
 	triggerGarbageCollection() {
 		if (typeof global.gc === 'function') {
+			// Call GC twice to allow for more reclaimation
+			global.gc();
 			global.gc();
 			return { success: true, message: 'Garbage collection triggered' };
 		}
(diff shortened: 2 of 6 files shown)


packages/testing/containers/services/n8n.ts
```diff
@@ -12,7 +12,7 @@ const N8N_IMAGE = TEST_CONTAINER_IMAGES.n8n;
 const BASE_ENV: Record<string, string> = {
 	N8N_LOG_LEVEL: 'debug',
 	N8N_ENCRYPTION_KEY: process.env.N8N_ENCRYPTION_KEY ?? 'test-encryption-key',
-	E2E_TESTS: 'false',
+	E2E_TESTS: 'true',
```
ivov: This will enable the entire e2e controller for perf tests, which only need the GC endpoint. Consider a separate `PERF_TESTS`

packages/cli/src/controllers/e2e.controller.ts
```diff
@@ -264,6 +264,22 @@ export class E2EController {
...
 	}
 
+	/**
+	 * Trigger garbage collection for memory profiling in performance tests.
+	 * Requires Node.js to be started with --expose-gc flag.
+	 */
+	@Post('/gc', { skipAuth: true })
+	triggerGarbageCollection() {
+		if (typeof global.gc === 'function') {
+			global.gc();
```
ivov, line 276: V8 sometimes needs multiple GC cycles to fully reclaim memory. Let's do 2 or accept `passes` make it configurable?

.gitignore
```diff
@@ -49,3 +49,4 @@ packages/cli/src/commands/export/outputs
 .data
 .data/
 .claude/settings.local.json
+lefthook-local.yml
```
ivov, line 52: Looks unrelated

packages/testing/containers/services/n8n.ts
shortstacked: Reverted

packages/cli/src/controllers/e2e.controller.ts
shortstacked: Added in the extra pass, so we will call it twice. Apparently 3 times is too much and is diminishing returns territory.

.gitignore
shortstacked: Sort of unrelated but allows us to use local hooks (I set one up to do lint/typecheck on a push)

ivov (commented):
If we disable `E2E_TESTS` controller then `/gc` will need to live elsewhere, i.e. we'd need `PERF_TESTS` to enable a dedicated controller.

Else this won't work as is right now or?

shortstacked:
> If we disable `E2E_TESTS` controller then `/gc` will need to live elsewhere, i.e. we'd need `PERF_TESTS` to enable a dedicated controller.
> 
> Else this won't work as is right now or?

It's enabled at the project level anyway for the tests, we need the controller to actually do things like login/setup other tests anyway.

ivov (approved):
> > If we disable `E2E_TESTS` controller then `/gc` will need to live elsewhere, i.e. we'd need `PERF_TESTS` to enable a dedicated controller.
> > Else this won't work as is right now or?
> 
> It's enabled at the project level anyway for the tests, we need the controller to actually do things like login/setup other tests anyway.

Not familiar with this setup, trust you here

PR #25406, opened 2026-02-05
feat(core): Ensure credentials only use secret providers that are available in their project (no-changelog)

konstantintieber:
## Summary

Project-specific external secret providers have not been released yet.
This is why the added logic in this PR only runs if the env var `N8N_ENV_FEAT_EXTERNAL_SECRETS_FOR_PROJECTS` is true.

This PR adds validation to the internal credentials API that a `$secrets.` expression only references vaults that are accessible in the project the credential belongs to.

Covers these endpoints:
- POST /rest/credentials
- PATCH /rest/credentials/:credentialId
- PATCH /api/v1/credentials/:credentialId

Credential transfer endpoints are covered in a sub-PR: https://github.com/n8n-io/n8n/pull/25625

Does not affect `POST /api/v1/credentials` since that one is limited to only creating credentials in the personal project. Instance owners will be able to use project secrets in there but nobody else will.

https://linear.app/n8n/issue/LIGO-170/be-credential-savingmoving-prevent-saving-a-secret-expression-for-a

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -36,4 +48,27 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
...
 			.getMany();
 	}
+
+	/**
+	 * Checks if a provider is accessible from a project.
+	 * A provider is accessible if it's either:
+	 * - A global provider (no project access restrictions), OR
+	 * - Explicitly granted access to the specified project
+	 */
+	async hasAccessToProvider(providerKey: string, projectId: string): Promise<boolean> {
```
konstantintieber: @geemanjs @ireneea what do you think of using this kind of query on every update/create of a credential that's referencing a `{{ $secrets.<providerKey> }}` expression.
I think it's fine and don't see an alternative. Curious to hear your thoughts.

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -13,6 +13,18 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
 		return await this.find();
 	}
 
+	async hasGlobalProvider(providerKey: string): Promise<boolean> {
+		const count = await this.manager
+			.createQueryBuilder(SecretsProviderConnection, 'connection')
+			.leftJoin('connection.projectAccess', 'access')
+			.where('access.secretsProviderConnectionId IS NULL')
+			.andWhere('connection.providerKey = :providerKey', { providerKey })
+			.andWhere('connection.isEnabled = :isEnabled', { isEnabled: true })
```
ireneea: We no longer need this filter since we're removing the field

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -36,4 +48,27 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
...
+	 * A provider is accessible if it's either:
+	 * - A global provider (no project access restrictions), OR
+	 * - Explicitly granted access to the specified project
+	 */
+	async hasAccessToProvider(providerKey: string, projectId: string): Promise<boolean> {
+		const count = await this.manager
+			.createQueryBuilder(SecretsProviderConnection, 'connection')
+			.leftJoin('connection.projectAccess', 'access')
+			.where('connection.providerKey = :providerKey', { providerKey })
+			.andWhere('connection.isEnabled = :isEnabled', { isEnabled: true })
```
ireneea: Same here this is not needed anymore

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
ireneea: An alternative, would be to actually keep track of the provider keys used as a separate column on the credential. 

@alielkhateeb, this is what you suggested to help with finding the affected credentials when deleting a secret provider connection.

I was initially not really keen because it added complexity. But if this is something we'll need anyway for validation this could be something to consider.

konstantintieber: I also discussed this with @alielkhateeb just now. I was out on Friday when you talked about going for the "decrypt all credentials in the backend when `externalSecretStore=providerKey` param is provided".
I explained that I liked @mfsiega approach with the `workflow_dependency` table outlined in https://www.notion.so/n8n/Workflow-Dependency-Indexing-2865b6e0c94f80c28a6fec455cc5367e and whether we could reuse that for a new `credential_dependency` table.

My suggestion was to prepare an RFC outlining two approaches
- Decrypt and parse all `$secret` expression strings on `GET /rest/credentials?externalSecretStore=<providerKey>` and filter out credentials that don't contain `providerKey`
- Store all `providerKey`s referenced in a credential in the DB, e.g. a `credential_dependency` table on create/update of any credential -> no need to decrypt and parse expression strings anymore

This RFC should be discussed in tomorrow's architecture meeting: https://www.notion.so/n8n/Performance-Querying-dependencies-encoded-in-expression-strings-of-credentials-e-g-external-secre-3025b6e0c94f8076aaf1ee9681e12a40?source=copy_link

Nontheless, I see such a column/table not directly related to this pull request right here, because this one is about validating on update/create, whether the current project has access to the providers referenced in `$secret` expressions (even before the credential is created).
And it's a question that only needs to be answered on update/create of an individual credential that is using a `${{ secrets... }}`, whereas the `externalSecretStore=providerKey` query param needs to check all existing credentials to give an answer.

packages/cli/src/modules/external-secrets.ee/secret-provider-access-check.service.ee.ts
```diff
@@ -0,0 +1,11 @@
+import { SecretsProviderConnectionRepository } from '@n8n/db';
+import { Service } from '@n8n/di';
+
+@Service()
+export class SecretsProviderAccessCheckService {
+	constructor(private readonly connectionRepository: SecretsProviderConnectionRepository) {}
+
+	async canAccessProviderFromProject(providerKey: string, projectId: string): Promise<boolean> {
```
ireneea: What do you think of naming this method `isProviderAvailableInProject` (same for the repository method)?

packages/cli/src/credentials/validation.ts
```diff
@@ -12,10 +13,37 @@ import { getAllKeyPaths } from '@/utils';
...
+ * Supports both dot notation ($secrets.vault.key) and bracket notation ($secrets['vault']['key']).
+ *
+ * @param expression - The expression string containing $secrets reference
+ * @returns The provider key, or null if extraction fails
+ *
+ * @example
+ * extractProviderKey("={{ $secrets.vault.myKey }}") // returns "vault"
+ * extractProviderKey("={{ $secrets['aws']['secret'] }}") // returns "aws"
+ */
+export function extractProviderKey(expression: string): string | null {
```
ireneea: Just thought of the following edge case where there are multiple secrets used in the same expression 

```
={{ $secrets.vault.apiKey + ':' + $secrets.aws.secret }}
```

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -496,6 +503,13 @@ export class CredentialsService {
 		decryptedData: ICredentialDataDecryptedObject,
 	): Promise<CredentialsEntity> {
 		validateExternalSecretsPermissions(user, data.data, decryptedData);
+		if (this.externalSecretsConfig.externalSecretsForProjects && data.projectId && data.data) {
```
ireneea: Is the `data.projectId` always present in updates? Trying to understand if it's optional because of global credentials or if updates behave as patch and only properties that are modified are sent by the FE.

packages/cli/src/modules/external-secrets.ee/secret-provider-access-check.service.ee.ts
konstantintieber: That's better. Implemented in https://github.com/n8n-io/n8n/pull/25406/commits/0fb28fdf21a45c22a8db6037374f90c4c643b743

packages/cli/src/credentials/validation.ts
konstantintieber: Good catch.
Added support for multiple secrets in https://github.com/n8n-io/n8n/pull/25406/commits/0ba3ca4f24c3ff5708b5bdc7e4c15a626c248355

packages/cli/src/credentials/credentials.service.ts
konstantintieber: Just tested this for POST:
- internal API `/rest/credentials` will fail if no projectId provided
- public API `/api/v1/credentials` will create a personal credential if no projectId provided

PATCH:
- internal API `/rest/credentials/:credentialId`: this one I don't understand yet. The payload sent by the frontend sends the `projectId` data inside `req.body.homeProject.id`, but somehow on the backend controller it ends up being inside `req.body.projectId`
- public API `/api/v1/credentials/:credentialId` update succeeds without providing `homeProject` inside the payload, however the [updateCredential](https://github.com/n8n-io/n8n/blob/ligo-170-be-credential-savingmoving-prevent-saving-a-secret/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts#L205) implementation here does not use `prepareUpdateData`, but is a separate one, that will always fetch the owning project from the DB to perform the validation

konstantintieber: But either way we shouldn't use the projectId provided in the payload as source of truth for validation in `prepareUpdateData`.
We should use the `projectId` stored in our backend. Getting on that.

konstantintieber: Addressed in https://github.com/n8n-io/n8n/pull/25406/commits/3ac9565546538602a44c393fe50b6ae9cb09b7ed

ireneea (approved):

PR #25414, opened 2026-02-06
feat(core): Add versions support to `n8n:export` command

RicardoE105:
## Summary                                                        
  
Adds version-awareness to the CLI export:workflow command, enabling users to export specific historical versions or the published/active version of a workflow. Includes support for preserving version metadata during export/import.

  ### Changes

  New CLI Flags

  - --version <versionId>: Export a specific historical version
  by its versionId
  - --published: Export the published/active version of the
  workflow
  - --all --published: Export all workflows in their published
  versions (skips unpublished workflows)

  ### Version Metadata Preservation

  - Exported workflows now include a workflowHistory property
  containing historical name and description
  - Import command reads and preserves workflowHistory metadata
  in the WorkflowHistory table
  - Workflow name/description remain from the current entity
  (not overridden)

  Examples

  # Export current draft (unchanged behavior)
  n8n export:workflow --id=abc123 --output=workflow.json

  # Export specific historical version
  n8n export:workflow --id=abc123 --version=xyz789
  --output=workflow-v1.json

  # Export published version
  n8n export:workflow --id=abc123 --published
  --output=published.json

  # Export all workflows in their published versions
  n8n export:workflow --all --published --output=workflows.json

  # Use -o alias for output
  n8n export:workflow --all -o workflows.json

  ### Examples
  ```bash
  # Export current draft (unchanged behavior)
  n8n export:workflow --id=abc123 --output=workflow.json

  # Export specific historical version
  n8n export:workflow --id=abc123 --version=xyz789
  --output=workflow-v1.json

  # Export published version
  n8n export:workflow --id=abc123 --published
  --output=published.json
```

 https://linear.app/n8n/issue/ADO-4775

## Review / Merge checklist

- [x] Tests included. 

### CODE REVIEW

File: packages/cli/src/commands/export/workflow.ts [hunk 1]
```diff
@@ -110,6 +120,52 @@ export class ExportWorkflowsCommand extends BaseCommand<z.infer<typeof flagsSche
...
+
+				workflow.nodes = workflowHistory.nodes;
+				workflow.connections = workflowHistory.connections;
+				workflow.versionId = workflowHistory.versionId;
+
+				if (workflowHistory.name !== null) {
+					workflow.name = workflowHistory.name;
+				}
+				if (workflowHistory.description !== null) {
+					workflow.description = workflowHistory.description;
```
dariacodes: I have a concern with this, because it overwrites workflow's name and description... I know we don't currently store version names and descriptions in workflow code, but maybe we should start doing that to avoid this issue. But for now, I wonder if we should keep workflow's name and description. I'll ask the product!

packages/cli/src/commands/export/workflow.ts
```diff
@@ -110,6 +120,52 @@ export class ExportWorkflowsCommand extends BaseCommand<z.infer<typeof flagsSche
...
+						versionId: targetVersionId,
+					},
+				});
+
+				if (!workflowHistory) {
+					this.logger.error(
+						`Version "${targetVersionId}" not found for workflow "${workflow.name}" (${workflow.id})`,
+					);
+					process.exit(1);
+					return;
```
dariacodes: For consistency, let's throw errors instead of using `process.exit()` as we already have a try-catch block that also logs "Error exporting workflows. See log messages for details.".
The only thing - please remove `this.logger.error(error.message);` in the catch block, because `UserError` messages are already displayed to the user by the error handling system, so we currently have duplication.

```suggestion
					throw new UserError(`Version "${targetVersionId}" not found for workflow "${workflow.name}" (${workflow.id})`);
```

packages/cli/src/commands/export/workflow.ts
```diff
@@ -110,6 +120,52 @@ export class ExportWorkflowsCommand extends BaseCommand<z.infer<typeof flagsSche
...
+			const workflow = workflows[0];
+			let targetVersionId: string;
+
+			if (flags.published) {
+				if (!workflow.activeVersionId) {
+					this.logger.error(
+						`Workflow "${workflow.name}" (${workflow.id}) is not published. Cannot export published version.`,
+					);
+					process.exit(1);
+					return;
```
dariacodes: (proposed a new wording as well)
```suggestion
					throw new UserError(
						`Workflow "${workflow.name}" (${workflow.id}) has no published version`,
					);
```

dariacodes (commented):
I tested it thoroughly and it works well! 🚀  
Left one concern and a couple of suggestions to improve user-facing logs

packages/cli/src/commands/export/workflow.ts
```diff
@@ -17,18 +17,16 @@ const flagsSchema = z.object({
 		)
 		.optional(),
 	id: z.string().describe('The ID of the workflow to export').optional(),
-	output: z
-		.string()
-		.alias('o')
-		.describe('Output file name or directory if using separate files')
-		.optional(),
+	output: z.string().describe('Output file name or directory if using separate files').optional(),
```
dariacodes: Any particular reason you removed the alias from here? Our users might be used to it

packages/cli/src/commands/export/workflow.ts
```diff
@@ -53,6 +53,16 @@ export class ExportWorkflowsCommand extends BaseCommand<z.infer<typeof flagsSche
 			flags.separate = true;
 		}
 
+		if (flags.version && flags.published) {
+			this.logger.info('Cannot use both --version and --published flags. Please specify one.');
+			return;
+		}
+
+		if ((flags.version || flags.published) && flags.all) {
```
dariacodes: By the way, why is it impossible to use `all` and `published`? What if we want to export all the published versions?

packages/cli/src/commands/export/workflow.ts
```diff
@@ -101,15 +124,102 @@ export class ExportWorkflowsCommand extends BaseCommand<z.infer<typeof flagsSche
...
+			}
+		}
+
+		// Handle --all --published: export all workflows in their published versions
+		if (flags.all && flags.published) {
+			const workflowsToExport = [];
+			for (const workflow of workflows) {
+				if (!workflow.activeVersionId) {
+					this.logger.warn(
+						`Skipping workflow "${workflow.name}" (${workflow.id}) - no published version`,
```
dariacodes: I'd remove this to not clutter the logs as enterprise instances can have lots of workflows

packages/cli/src/commands/export/workflow.ts
```diff
@@ -101,15 +124,102 @@ export class ExportWorkflowsCommand extends BaseCommand<z.infer<typeof flagsSche
...
+					);
+					continue;
+				}
+
+				// If published version is not the current draft, fetch it from history
```
dariacodes: I'm actually thinking that it makes sense to fetch history for all the workflows so that we fetch names and descriptions from workflow history, but we should do this in an optimized way to not send N queries (one for each workflow)

packages/cli/src/services/import.service.ts
```diff
@@ -158,12 +158,15 @@ export class ImportService {
 			// Always create workflow history for the current version
 			// This is needed to be able to activate the workflow later
 			for (const workflow of insertedWorkflows) {
+				const workflowHistory: any = workflow;
```
dariacodes: We should fix the types here

dariacodes (changes requested):
We need to optimize querying workflow history, as enterprise instances can have 10.000 workflows and we don't want to end up with 10.000 queries. I also checked that during import `name` and `description` are not set after all (upd.: see https://github.com/n8n-io/n8n/pull/25569/commits/ec58aebff0589a3b7b008e5137169220bd632f44)
Here is a PR fixing all the issues: https://github.com/n8n-io/n8n/pull/25569. Could you take a look?

dariacodes (approved):
🚢

PR #25429, opened 2026-02-06
feat(core): Add `workflow-version-updated` event to log streaming

dariacodes:
## Summary

- Add new `workflow-version-updated` event to log streaming that will be relayed on "Name version" action
- Fix `workflow-deactivated` event
    - It was sent with `activeVersionId` which is aways `null` - instead, we're now sending `deactivatedVersionId`
- Fix bug with not sending a deactivated event when activating a new version + sending the activated event before knowing activation went on successfully.

### How to test
- Pul the latest changes, checkout this branch and run `git rebase ado-4732-frontend-add-name-version-actions` (I want to merge it to master, that's why it's not initially based off of the frontend branch, but it's the only way to test it properly)
- Prepare a webhook link (e.g. [here](https://webhook.site/#!/view/db8b9d1b-a19f-433a-a00c-4c6f56a4da67))
- Go to your instance. Click `Settings` -> `Log Streaming`.
- Add a new destination of type `Webhook` and paste your URL (leave the `POST` method). Save and activate.
- Go to a workflow page, click `Cmd + S` and save a new version.
- See a new event (`n8n.audit.workflow.version.updated`) on the webhook page.

To see "deactivated" event, unpublish a published workflow.

Closes [ADO-4734](https://linear.app/n8n/issue/ADO-4734/add-version-updated-event-to-log-streaming)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/workflows/workflow.service.ts
+++ b/packages/cli/src/workflows/workflow.service.ts
@@ -488,7 +488,6 @@ export class WorkflowService {
 				activeVersionId: null,
 				activeVersion: null,
 			};
-			const previouslyActiveId = workflow.activeVersionId;
 			await this.workflowRepository.update(workflowId, rollbackPayload);
 
 			// Also set it in the returned data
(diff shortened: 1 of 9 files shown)


packages/cli/src/workflows/workflow.service.ts
```diff
@@ -653,11 +624,26 @@ export class WorkflowService {
 		this._validateNodes(workflowId, versionToActivate.nodes);
 		await this._validateSubWorkflowReferences(workflowId, versionToActivate.nodes);
 
-		if (wasActive) {
+		if (previousActiveVersionId) {
 			await this.activeWorkflowManager.remove(workflowId);
+
+			assert(previousActiveVersionId !== null);
```
CharlieKolb: Can you check this in the if condition? Or are empty strings ever valid `previousActiveVersionIds`?

CharlieKolb (dismissed):

packages/cli/src/workflows/workflow.service.ts
dariacodes: They are not, and it's an impossible situation to have an empty string there (because it's a foreign key). But this assert is not needed anymore after I refactored `wasActive` - removed it just now

CharlieKolb (approved):

PR #25451, opened 2026-02-06
fix(core): AI models don't explain images in ChatHub

autologie:
## Summary

This PR tweaks system prompt in ChatHub so models can explain shared images if they are capable of.

| Before | After |
|--------|--------|
| Screenshot 2026-02-06 at 13 07 36 | Screenshot 2026-02-06 at 13 07 43 | 

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -483,19 +483,24 @@ export class ChatHubWorkflowService {
 	}
 
 	getSystemMessageMetadata(timeZone: string) {
-		const now = inE2ETests ? DateTime.fromISO('2025-01-15T12:00:00.000Z') : DateTime.now();
+		if (inE2ETests) {
+			return '__e2e_system_prompt_placeholder__';
```
autologie, line 487: It's annoying to have to update e2e expectation files on every system prompt tuning

Cadiac: this is probably a good solution to avoid that 👍

Cadiac (approved):

PR #25457, opened 2026-02-06
ci: Skip e2e tests on private repository

ivov:
The e2e workflow pushes to `ghcr.io/n8n-io/runners` but the private repo's token [lacks write access](https://github.com/n8n-io/n8n-private/actions/runs/21756650573/job/62769128851) and linking it would expose security images. Since we run e2e anyway on merge, let's skip it on the private side.


shortstacked (approved):
I'll do the proper fix during the next week to push proper images to private as well so we don't need to cross repo boundaries at all.

PR #25463, opened 2026-02-06
feat(API): Add public API endpoints for execution tags

jonjames-n8n:
## Summary
Adds public API endpoints to programmatically tag n8n executions, matching the existing workflow tagging functionality.

## Changes
- Added `GET /executions/:id/tags` endpoint
- Added `PUT /executions/:id/tags` endpoint
- Added `executionTags` scope to API key permissions
- Full OpenAPI spec documentation
- 14 new tests added (47 total tests passing)

## Motivation
n8n already has internal execution annotation/tagging via `ExecutionAnnotation` and `AnnotationTagEntity`, but these were not exposed to the public API. This PR exposes the existing functionality to enable use cases like:
- Programmatically tagging executions as "dataset" to prevent pruning
- Automation workflows that need to categorize executions
- Integration with external systems that track execution metadata

## Testing
All existing tests pass (33), plus 14 new tests covering:
- GET /executions/:id/tags with valid execution
- GET with non-existent execution
- PUT /executions/:id/tags to add tags
- PUT to update existing tags
- Authorization checks

## Breaking Changes
None - this is purely additive functionality.


packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
```diff
@@ -196,4 +202,92 @@ export = {
...
+
+			return res.json(tags);
+		},
+	],
+	updateExecutionTags: [
+		apiKeyHasScope('executionTags:update'),
+		async (req: ExecutionRequest.UpdateTags, res: express.Response): Promise<express.Response> => {
+			const { id } = req.params;
+			const newTagIds = req.body.map((tag) => tag.id);
+			const sharedWorkflowsIds = await getSharedWorkflowIds(req.user, ['workflow:read']);
```
nikhilkuria: It should be `workflow:update` than `workflow:read` here.

packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
```diff
@@ -196,4 +202,92 @@ export = {
...
+
+			if (!sharedWorkflowsIds.length) {
+				return res.status(404).json({ message: 'Not Found' });
+			}
+
+			const execution = await Container.get(
```
nikhilkuria, line 236: This handler performs three separate DB operations (upsert annotation, overwrite tags, fetch updated tags) without a transaction. 
If something fails between the upsert and the overwriteTags, the annotation is created but tags aren't set.

nikhilkuria: Can you do something similar to how we handle workflow tags. 
The existing workflowTags handler uses a dbManager.transaction() wrapper in [workflows.service.ts:108](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/public-api/v1/handlers/workflows/workflows.service.ts#L106).

packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
```diff
@@ -196,4 +202,92 @@ export = {
...
+				throw error;
+			}
+
+			// Fetch updated tags to return
+			const updatedAnnotation = await Container.get(ExecutionAnnotationRepository).findOneOrFail({
+				where: { execution: { id } },
+				relations: ['tags'],
+			});
+
+			const tags = (updatedAnnotation.tags ?? []).map(
```
nikhilkuria: The same mapping is used for getExecutionTags as well. Can you extract this to a separate method?

packages/cli/src/public-api/types.ts
```diff
@@ -42,6 +42,8 @@ export declare namespace ExecutionRequest {
 	type Get = AuthenticatedRequest<{ id: string }, {}, {}, { includeData?: boolean }>;
 	type Delete = Get;
 	type Retry = AuthenticatedRequest<{ id: string }, {}, { loadWorkflow?: boolean }, {}>;
+	type GetTags = AuthenticatedRequest<{ id: string }>;
+	type UpdateTags = AuthenticatedRequest<{ id: string }, {}, AnnotationTagEntity[]>;
```
nikhilkuria: Can we define the UpdateTags type as 
```
type UpdateTags = AuthenticatedRequest<{ id: string }, {}, Array<{ id: string }>>;
```

We only use the id property from AnnotationTagEntity. So, it's better if we use an Array of strings instead

nikhilkuria (changes requested):

nikhilkuria (approved):

PR #25464, opened 2026-02-06
fix(editor): Gate dynamic credentials UI behind license check

cstuncsik:
## Summary

Dynamic credentials features were only gated by the DYNAMIC_CREDENTIALS env feature flag. Unlicensed users could see features they can’t use.

This adds a `useDynamicCredentials` composable that requires both the env feature flag AND an active dynamic-credentials backend module (which is license-gated via @BackendModule). All 8 UI locations that previously checked only the env flag now use this composable.

https://linear.app/n8n/issue/IAM-152/fe-feature-license-awareness-on-the-front-end

## Review / Merge checklist

- [ ] Tests included. 


BGZStephen (approved):
LGTM

guillaumejacquart (approved):
All good, thanks !
Just wondering if in the future we should do some "real" gating like presenting the disabled feature and inviting to upgrade

cstuncsik:
> All good, thanks ! Just wondering if in the future we should do some "real" gating like presenting the disabled feature and inviting to upgrade

Yes, I think we should and we will (I guess).
Product need to decide and design how.

Currently, we probably just want to hide the feature completely when there is no enterprise license, but the environment variable is set.

PR #25467, opened 2026-02-06
test: Align proxy service helper and add local dev service stack (no-changelog)

shortstacked:
## Summary

Two independent improvements to the testcontainer infrastructure:

**1. Proxy service alignment** — `ProxyServer` (MockServer API client) was the only service helper with a special standalone fixture in `base.ts`. Every other service (mailpit, gitea, kafka, etc.) follows the `services.X` pattern via the helper registry. This PR moves `ProxyServer` into the containers package and registers it as a proper service helper, so all tests use `services.proxy` consistently.

**2. `pnpm services` CLI** — Adds a `--services-only` mode to the test stack that starts service containers (postgres, redis, mailpit, etc.) without any n8n instances. Writes a `.env` file with host-compatible connection details so `pnpm dev` can connect to containerized services. Useful for local development against real databases without manual Docker setup.

### Why this approach

**Proxy alignment:** The `ProxyServer` class was duplicated in the playwright package with its own fixture wiring. Moving it to the containers package (where all other service helpers live) follows the established pattern: helper class + `createXHelper` factory + module augmentation on `ServiceHelpers` + registration in `helperFactories`. This means any new test just uses `services.proxy` like every other service — no special imports needed.

**Services CLI:** The existing `createN8NStack()` already supported `mains: 0, workers: 0` for service-only mode. The new CLI script is a thin wrapper that starts the stack, reads mapped ports from each container, and writes them as env vars to `.env` (which n8n CLI already loads via dotenv). Each service's `env()` method gained an `external` parameter to return host-compatible values (localhost + mapped port) instead of Docker-internal values.

### What changed for test authors

```diff
- import { ProxyServer } from '../services/proxy-server';
- // ... custom fixture setup
- const { proxyServer } = fixtures;
- await proxyServer.createExpectation(...)
+ const { services } = fixtures;
+ await services.proxy.createExpectation(...)
```

All 10 test files using `proxyServer` have been updated mechanically.

## How to test locally

**Proxy alignment:** Run existing proxy E2E tests — they should pass with no behavior change:
```bash
pnpm --filter=n8n-playwright test --grep @capability:proxy
```

**Services CLI:**
```bash
cd packages/testing/containers

# Start postgres for local dev
pnpm services --services postgres

# Check .env was written at repo root
cat ../../cli/bin/.env

# In another terminal — n8n connects to containerized postgres
pnpm dev

# Cleanup
pnpm services:clean
```

Requires Docker running locally. No special licenses or credentials needed.

## Review / Merge checklist

- [x] Tests included. 


shortstacked:
@claude Review please

mutdmour (approved):
LGTM

PR #25501, opened 2026-02-09
feat(editor): Show execution data in preview

schrothbn:
## Summary
Adds a new openExecutionPreview postMessage command that lets a parent window render a workflow canvas with per-node execution status indicators (success/error/canceled) without shipping actual execution output data. 
The utility function buildExecutionResponseFromSchema constructs a synthetic IExecutionResponse from a lightweight schema.

https://linear.app/n8n/issue/AI-1675/feature-load-workflow-preview-with-execution-data-loaded-and-error

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/views/NodeView.vue
+++ b/packages/frontend/editor-ui/src/app/views/NodeView.vue
@@ -101,7 +101,6 @@ import { useNpsSurveyStore } from '@/app/stores/npsSurvey.store';
 import { useTelemetry } from '@/app/composables/useTelemetry';
 import { useProjectsStore } from '@/features/collaboration/projects/projects.store';
 import { useNodeHelpers } from '@/app/composables/useNodeHelpers';
-import { useExecutionDebugging } from '@/features/execution/executions/composables/useExecutionDebugging';
 import { buildExecutionResponseFromSchema } from '@/features/execution/executions/executions.utils';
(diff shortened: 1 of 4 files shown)


packages/frontend/editor-ui/src/app/views/NodeView.vue
```diff
@@ -1151,6 +1154,45 @@ async function onOpenExecution(executionId: string, nodeId?: string) {
...
+
+	const workflow = json.workflow;
+	if (!workflow?.nodes || !workflow?.connections) {
+		canvasStore.stopLoading();
+		throw new Error('Invalid workflow object');
+	}
+
+	if (json.projectId) {
+		await projectsStore.fetchAndSetProject(json.projectId);
+	}
```
mutdmour, line 1175: why is this needed? since this is the preview instance, this endpoint might be blocked.

schrothbn: Was following the pattern in the other paths, but valid, maybe this would just fail silently for the other paths as well?

mutdmour: it just does not make sense for this context.

mutdmour: i.e. the preview component. there's no "project" context that we need to set

mutdmour (approved):

PR #25502, opened 2026-02-09
chore: Add telemetry for dynamic workflow executions

BGZStephen:
## Summary

For post execution events that include dynamic credentials, track all events, not just errors like currently happens.

IAM-250

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/telemetry/index.ts
+++ b/packages/cli/src/telemetry/index.ts
@@ -181,10 +181,15 @@ export class Telemetry {
 				executionTrackDataKey.count++;
 			}
 
+			if (properties.used_dynamic_credentials) {
+				this.track('Workflow execution with dynamic credentials', properties);
+			}
+
 			if (
 				!properties.success &&
 				properties.is_manual &&
-				properties.error_node_type?.startsWith('n8n-nodes-base')
+				properties.error_node_type?.startsWith('n8n-nodes-base') &&
+				!properties.used_dynamic_credentials
 			) {
 				this.track('Workflow execution errored', properties);
 			}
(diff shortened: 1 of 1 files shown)


packages/cli/src/telemetry/index.ts
```diff
@@ -181,10 +181,15 @@ export class Telemetry {
...
+			if (properties.used_dynamic_credentials) {
+				this.track('Workflow execution with dynamic credentials', properties);
+			}
+
 			if (
 				!properties.success &&
 				properties.is_manual &&
-				properties.error_node_type?.startsWith('n8n-nodes-base')
+				properties.error_node_type?.startsWith('n8n-nodes-base') &&
+				!properties.used_dynamic_credentials
```
phyllis-noester: why fo we not want to get the error notification for dynamic creds of node 'n8n-nodes-base'?

BGZStephen: to avoid double reporting, it's captured in the event above now that groups all dynamic credentials stuff together, or do you think double reporting should happen? I'm a little torn

BGZStephen: thinking about it, it may be weird that errors no longer go to the same event, I've added dynamic creds back to the error report but also let them exist in the new event too

phyllis-noester (approved):

packages/cli/src/telemetry/index.ts
phyllis-noester: sounds good! yea i think this way we know we're not breaking any existing behavior. I assume this telemetry event might somehow be connect to this https://n8n.metabaseapp.com/dashboard/229-base-quality-metrics-v2?error_threshold=98&last_n_days=180&total_executions_threshold=10 but not sure how the calculations happen might be on big query side 

(tracking events go to big query right?)

PR #25506, opened 2026-02-09
feat: Show Quick connect banner in the NDV

RomanDavydchuk:
## Summary

Show an incentive banner for Quick connect in the NDV if all of the following apply:
- Quick connect feature flag is enabled
- the node has credentials that support Quick connect
- the node does not have credentials configured
- the NDV is not in preview or read-only mode (i.e. is editable)

https://linear.app/n8n/issue/NODE-4368/implement-incentive-banner-and-quick-connect-button-in-node-details

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/integrations/quickConnect/composables/useQuickConnect.ts
```diff
@@ -7,13 +7,24 @@ type UseQuickConnectParams =
...
 	  }
 	| {
 			packageName?: never;
 			credentialType: string | Ref<string | undefined>;
+			credentialTypes?: never;
+	  }
+	| {
+			packageName?: never;
+			credentialType?: never;
+			credentialTypes: string[] | Ref<string[] | undefined>;
```
dlavrenuek: Instead of adding a new parameter we can use `credentialType` and say that you could either pass a single string or an array of credential types. Or we could also define that it must always be an array to have less conditions inside `useQuickConnect` (then `credentialTypes` would be a better name than `credentialType`). What do you think?

PS: coming from React considering reactive values adds quite an overhead in my brain. With a single value for credential type my intention of the hook was that it can work with
1. static values (for example credential type was passed from parent component and is already "final" when `useQuickConnect()` is called), like `useQuickConnect({ credentialType: "ok" })`
2. reactive values (for example where credential type is loaded inside component and therefore can also be undefined), similar [to node info view](https://github.com/n8n-io/n8n/blob/master/packages/frontend/editor-ui/src/features/settings/communityNodes/components/nodeCreator/CommunityNodeInfo.vue#L29)

If we say that you must always provide an array as credentialTypes, then we should probably consider `Ref<string[]>` and `Ref<string>[]`/`Array<Ref<string>>` as possible types? My preference would be to allow either a single value or an array, tho

dlavrenuek: I don't think that we need to consider `undefined` as a ref when working with arrays. When computing the array with `computed()` you can always return an empty array if something did not load yet, so I would suggest `string[] | Ref<string[]>` as the type here

dlavrenuek (changes requested):
Works as expected and UI looks good 🥇 I have added some suggestions on the typing interface for `useQuickConnect`.

packages/frontend/editor-ui/src/features/integrations/quickConnect/composables/useQuickConnect.ts
RomanDavydchuk: Updated the props to always accept an array, let me know if it looks good or if you have other comments

dlavrenuek (dismissed):
Perfect 🚀

dlavrenuek (approved):

PR #25447, opened 2026-02-06
feat: Quick connect banner for community nodes in node search view

dlavrenuek:
## Summary

Implements a quick connect banner promotion for community nodes in node search view.

Bildschirmfoto 2026-02-06 um 14 22 16

Also fixes wrong margin for the node info banner. Before:
Bildschirmfoto 2026-02-06 um 14 17 10

## Related Linear tickets, Github issues, and Community forum posts

- https://linear.app/n8n/issue/NODE-4367/implement-incentive-banner-and-quick-connect-information-in-node

## How to test

1. Start n8n with following env variable set, which is configured to display quick connect for the "Firecrawl" node

```
N8N_QUICK_CONNECT_OPTIONS='[{"packageName": "@mendable/n8n-nodes-firecrawl", "credentialType": "firecrawlApi", "text": "Firecrawl provides quick connect n8n users", "quickConnectType":"firecrawl"}]'
```

2. Once n8n runs enable the feature flag through developer console and reload the page:

```js
window.featureFlags.override('073_quick_connect', 'variant')
```
3. In a workflow click on add new node and search for "Firecrawl", the quick connect banner and tag should be displayed

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [x] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #25510, opened 2026-02-09
feat(editor): Add new empty state with quick connect for node credentials

elsmr:
## Summary

Add a new empty state for node credentials behind a feature flag. Nodes with managed OAuth credentials show a one-click "Connect to {provider}" button that creates a credential and runs the OAuth flow inline without opening the credential modal. Other credentials get a cleaner empty state with a "Set up credential" button.

feature flag: `073_quick_connect`

### ⚠️ Known limitation

When a user closes the OAuth popup mid-flow, a dangling credential is left. We clean up on component unmount and on explicit failure, but we cannot detect cross-origin window close. We could fix this on the backend by only creating the credential on OAuth callback, but this might be a major change as the callback is using credential ID to lookup the credential.

image

image

image

### Testing

Example `N8N_QUICK_CONNECT_OPTIONS` and `CREDENTIALS_OVERWRITE_DATA`
```bash
N8N_QUICK_CONNECT_OPTIONS='[{"packageName":"@mendable/n8n-nodes-firecrawl","credentialType":"firecrawlApi","text":"Firecrawl offers a free Basic plan to n8n users","quickConnectType":"api"}]' CREDENTIALS_OVERWRITE_DATA='{"googleSheetsOAuth2Api":{"clientId":"CLIENT_ID","clientSecret":"CLIENT_SECRET"}}'
```

https://linear.app/n8n/issue/NODE-4275

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/constants/experiments.ts
```diff
@@ -77,11 +77,7 @@ export const SETUP_PANEL = createExperiment('069_setup_panel', {
 	variant: 'variant',
 });
 
-export const QUICK_CONNECT_EXPERIMENT = {
-	name: '073_quick_connect',
-	control: 'control',
-	variant: 'variant',
-};
+export const QUICK_CONNECT_EXPERIMENT = createExperiment('073_quick_connect');
```
dlavrenuek, line 92: Nice, good to know that there is a helper for that.

packages/frontend/editor-ui/src/app/constants/experiments.ts
```diff
@@ -99,4 +95,5 @@ export const EXPERIMENTS_TO_TRACK = [
 	SIDEBAR_EXPANDED_EXPERIMENT.name,
 	EMPTY_STATE_EXPERIMENT.name,
 	SETUP_PANEL.name,
+	QUICK_CONNECT_EXPERIMENT.name,
```
dlavrenuek, line 113: ❓ What is this array used for?

packages/frontend/editor-ui/src/features/credentials/composables/useQuickConnect.ts
```diff
@@ -0,0 +1,32 @@
+import type { QuickConnectOption } from '@n8n/api-types';
+import { computed } from 'vue';
+
+import { useSettingsStore } from '@/app/stores/settings.store';
+
+/**
+ * Composable for quick connect detection.
+ * Used to determine when to show quick connect UI for credential types.
+ */
+export function useQuickConnect() {
```
dlavrenuek: There is already a reusable `useQuickConnect` hook, if I see it correctly it should provide the functionality you need (altho the interface and check changes a bit in [this PR](https://github.com/n8n-io/n8n/pull/25506/changes#diff-5830adbe9181cbbe5c70e089bc921e96548a360d29ef28261b5555a5f8f1cc5bR16))

packages/frontend/editor-ui/src/features/credentials/components/QuickConnectButton.vue
```diff
@@ -0,0 +1,67 @@
+<script setup lang="ts">
```
dlavrenuek, line 1: In this [PR](https://github.com/n8n-io/n8n/pull/25447/changes) I've implemented reusable quick connect composable and banner component in `features/integrations/quickConnect` but `features/credentials` looks like a better place. May be we should move the components you added here and what is already in `features/integrations/quickConnect` to `features/credentials/quickConnect`?

packages/frontend/editor-ui/src/features/credentials/components/QuickConnectButton.vue
```diff
@@ -0,0 +1,67 @@
...
+<template>
+	<N8nTooltip :disabled="!disabled || !disabledTooltip" placement="top">
+		<template #content>{{ disabledTooltip }}</template>
+		<span>
+			<GoogleAuthButton
+				v-if="isGoogleOAuthType(credentialTypeName)"
+				:disabled="disabled"
+				@click="$emit('click')"
+			/>
+			<N8nButton2
```
dlavrenuek: The quick connect button has border radius `--radius--2xs` which is visibly different from the `Execute step` button on the top and also all the input elements in the form, these have `--border`. The border radius seem to come from design system components, do you know if we can easily match the border radius for whole view?

dlavrenuek (changes requested):
First review with somme questions and suggestions

packages/frontend/editor-ui/src/features/credentials/components/QuickConnectButton.vue
dlavrenuek: We should also track the quick connect click events as specified in https://www.notion.so/n8n/User-clicked-quick-connect-button-dec217723c34464aa17204574e1c3f1c?v=bb5aef7984284f439aa93d66afb1ec13

Not sure if it should be inside the button component or rather in the parent view (probably in parent view since this button does not execute any logic on it's own)

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -498,6 +540,85 @@ async function onClickCreateCredential(type: ICredentialType | INodeCredentialDe
...
+	return hasQuickConnect(type.name, props.node.type) || hasManagedOAuthCredentials(type.name);
+}
+
+function showStandardEmptyState(type: INodeCredentialDescription): boolean {
+	if (isCredentialExisting(type)) return false;
+	// Standard empty state when no quick connect AND not managed OAuth
+	return !hasQuickConnect(type.name, props.node.type) && !hasManagedOAuthCredentials(type.name);
+}
+
+async function onQuickConnectSignIn(credentialTypeName: string) {
```
dlavrenuek, line 542: I'm wondering if the logic here should actually live mostly in the quick connect button, what do you think? Based on the [designs](https://www.figma.com/design/lHMYg2A8rBOklIrpA1XQ9x/Quick-Connect-Shaping?node-id=53-2113&m=dev) the button will be visible in the node details view but also in the create credentials view. If I see it correctly the second one is the `CredentialConfig.vue` component and does not include the `NodeCredentials` inside. But this can also be refactored once quick connect is implemented for the credential edit view

dlavrenuek: One a second though we should probably move the quick connect sign in logout outside this component. The next step would be to add integrations for different providers and it will blow up the code here

packages/frontend/editor-ui/src/app/constants/experiments.ts
elsmr: IIRC it tracks which experiments a user is part of, making it easier to query in Metabase

packages/frontend/editor-ui/src/features/credentials/components/QuickConnectButton.vue
elsmr: Good idea, will move

packages/frontend/editor-ui/src/features/credentials/components/QuickConnectButton.vue
elsmr: Maybe it could be part of the `useQuickConnect` composable? On the border radius, you're right, maybe I shouldn't be using the v2 button yet and instead create a new button variant of the old button

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
elsmr: Maybe the composable is a good place for it

dlavrenuek: I think it's a good idea to have the quick connect logic in the composable 👍

packages/frontend/editor-ui/src/features/credentials/composables/useCredentialOAuth.ts
```diff
@@ -0,0 +1,278 @@
...
+			}
+			if (credential.type === 'oAuth1Api' || parentTypes.includes('oAuth1Api')) {
+				return createResultOk(await credentialsStore.oAuth1Authorize(credential));
+			}
+		} catch (error) {
+			toast.showError(
+				error,
+				i18n.baseText('credentialEdit.credentialEdit.showError.generateAuthorizationUrl.title'),
+			);
+			return createResultError('api-error' as const);
```
dlavrenuek: 💡  There is no need for `as const` here. `'api-error'` already has the type `'api-error'` and not `string`

```suggestion
			return createResultError('api-error');
```

packages/frontend/editor-ui/src/features/credentials/composables/useCredentialOAuth.ts
```diff
@@ -0,0 +1,278 @@
...
+		}
+
+		return createResultError('no-url' as const);
```
dlavrenuek: 💡 
```suggestion
		return createResultError('no-url');
```

packages/frontend/editor-ui/src/features/credentials/composables/useCredentialOAuth.ts
```diff
@@ -0,0 +1,278 @@
...
+				{
+					id: '',
+					name: credentialType.displayName,
+					type: credentialTypeName,
+					data: {},
+				},
+				projectsStore.currentProject?.id,
+				undefined,
+				{ skipStoreUpdate: true },
+			);
```
dlavrenuek, line 239: ❌ If I see it correctly this creadential is created with "Allow HTTPRequest node domains"  being `all`, which provides security risks. While it is the default normally the credentials are created via the modal, so users can change it before saving the new credential. I suggest adding `allowedHttpRequestDomains: "none"` here (afaik simply in `data`)

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
```diff
@@ -0,0 +1,120 @@
...
+			// TODO: Implement quick connect flows here
+			return null;
+		}
+
+		if (!isOAuthCredentialType(credentialTypeName)) {
+			return null;
+		}
+
+		const credential = await createAndAuthorize(credentialTypeName);
+
```
dlavrenuek: 💡 The logic here is a bit confusing. First we check if quick connect is configured as option and execute custom quick connect and then there is a check if it's not oauth and returns null. I suggest to structure the if conditions based on positive outcome and not negative, so f.e.

```
if (getQuickConnectOption(credentialTypeName, nodeType)) {
...
} else if (isOAuthCredentialType(credentialTypeName)) {
...
}
return null;
```

It will not change the result, so I would assume that there is no need to change the tests

dlavrenuek (changes requested):
Great implementation, thank's a lot for the refactoring of existing composable 🎸 Added my final review. Comments marked with ❌ are required from my point of view, ones with 💡 are optional.

dlavrenuek (approved):
Perfect 🫰

PR #25526, opened 2026-02-09
feat(core): Expose externalSecretsStore to filter credentials by external secret provider key

alielkhateeb:
## Summary

Exposes optional query parameter `externalSecretsStore` to allow filtering credentials by a provider key.

Closes [LIGO-191](https://linear.app/n8n/issue/LIGO-191/be-filter-credentials-for-expressions-that-use-a-secret-provider)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/credentials/credentials.service.ts
+++ b/packages/cli/src/credentials/credentials.service.ts
@@ -50,6 +50,7 @@ import { OwnershipService } from '@/services/ownership.service';
 import { ProjectService } from '@/services/project.service.ee';
 import { RoleService } from '@/services/role.service';
 import { validateOAuthUrl } from '@/oauth/validate-oauth-url';
+import { getAllKeyPaths } from '@/utils';
 
 import { CredentialsFinderService } from './credentials-finder.service';
 import { validateExternalSecretsPermissions } from './validation';
@@ -103,6 +104,7 @@ export class CredentialsService {
 			includeData: true;
 			onlySharedWithMe?: boolean;
 			includeGlobal?: boolean;
+			externalSecretsStore?: string;
 		},
 	): Promise<Array<ICredentialsDecrypted<ICredentialDataDecryptedObject>>>;
 	async getMany(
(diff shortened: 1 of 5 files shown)


packages/cli/src/credentials/credentials.service.ts
```diff
@@ -166,6 +175,35 @@ export class CredentialsService {
...
+	private filterByExternalSecretsStore(
+		credentials: CredentialsEntity[],
+		externalSecretsStore: string,
+	): CredentialsEntity[] {
+		// matches either dot notation ($secrets.providerKey) or square bracket notation ($secrets['providerKey'])
+		const providerRegex = /\$secrets(?:\.([A-Za-z0-9_-]+)|\[['"]([^'"]+)['"]\])/g;
+		credentials = credentials.filter((credential) => {
+			const decryptedData = this.decrypt(credential, true);
+			const matchingSecretPaths = getAllKeyPaths(decryptedData, '', [], (value) => {
+				if (!value.includes('$secrets')) {
```
sandra0503, line 187: I think for performance reasons the early return should happen before decrypting and retrieving all key paths?

alielkhateeb: @sandra0503 That would be ideal, but unfortunately we have to decrypt in order to find the usages of "$secrets"

geemanjs (approved):

packages/cli/src/credentials/credentials.service.ts
sandra0503: Oh yes, of course regarding decrypting :D 
I was wondering if there's a shorter way to find to return (would `JSON.stringify` be an option our would we loose recursive information?)

alielkhateeb: Ah I see what you mean, I think it would work yes but there would be quite a few edge cases that would be inaccurate, e.g. if `data` after decryption is `{ "$secrets": "something" }` the condition would mistakenly think we have a secret there :D 
And there would be other edge cases so I think using the existing `getAllKeyPaths` with the value filter is safest as well as more maintainable in the future if we want to change any of the logic, then we only need to change this method's implementation and nothing else.

sandra0503: Makes sense! LGTM

sandra0503 (approved):

PR #25528, opened 2026-02-09
feat: Add Firecrawl handler to the Quick connect module

RomanDavydchuk:
## Summary

Add a Firecrawl handler that's responsible for interacting with the Firecrawl API to create the user under the hood and get the API key. Add tests, remove Sample handler that was used as an example

To test, make sure to configure `N8N_QUICK_CONNECT_OPTIONS` with proper value for `backendFlowConfig.secret`:
```json
[
  {
    "packageName": "@mendable/n8n-nodes-firecrawl",
    "credentialType": "firecrawlApi",
    "serviceName": "Firecrawl",
    "text": "Lorem ipsum",
    "quickConnectType": "backend",
    "backendFlowConfig": {
      "secret": "<SECRET>"
    }
  }
]
```

https://linear.app/n8n/issue/NODE-4404/implement-firecrawl-backend-handler

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/quick-connect/quick-connect.service.ts
elsmr: Can we make this a typeguard? like `config is ValidQuickConnectBackendConfig` then we can change the argument type of `getCredentialData` and won't need the `config.backendFlowConfig!`

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
```diff
@@ -0,0 +1,39 @@
...
+		user: User,
+	): Promise<ICredentialDataDecryptedObject> {
+		const secret = config.backendFlowConfig!.secret;
+		const response = await axios.post<FirecrawlCreateUserResponse>(
+			`${FIRECRAWL_API_BASE_URL}/admin/integration/create-user`,
+			{ email: user.email },
+			{
+				headers: {
+					Authorization: `Bearer ${secret}`,
+					'Content-Type': 'application/json',
```
elsmr, line 30: I was originally thinking this would be in the credential class for firecrawl, either as a function or a declarative version like we have for `test` or `authenticate`. I think this way is more than good, just wanted to know your opinion on it.

elsmr (commented):
Tested and works nicely 💯 

Some questions (not all related to this PR):

Should we also send `packageName` in the quick connect endpoint?

When quick connect succeeds, could we return the whole credential, same as the get credential endpoint (masked secrets), instead of only the ID? It could save an extra request from the frontend.

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
elsmr: On second thought, that would be in a community node in this case, more annoying for us to maintain

RomanDavydchuk: it's tricky, because each options has it's pros and cons. Here's what I think:
- keeping it as-is (a dedicated handler class) makes it easier to maintain, because all Quick connect related code is contained in a module
- but, for community nodes - it means that we have to have code in our core repo for a node that's not even there. I don't think it's an issue for a couple of partners, maybe even 10+, but for hundreds - that's not viable
- on the other hand, if a partner offers a frontend-based way of connecting with something like an SDK (like Pinecone), we have no other choice but to add it to our repo anyway, so we can't completely avoid that (at least at this stage)

given all this, I think it's fine to keep it as-is for now, if in the future it turns out that our pattern is not the best option - we can always change it. shouldn't be too hard since it's just Firecrawl for now. let me know what you think

elsmr (approved):

PR #25530, opened 2026-02-09
fix: Enable credentials experssions push and pull

### LINKED ISSUE #22015: Cannot push a credential using expressions that has been modified
### Bug Description

Hello,

When I create a credential (any type) and use expressions for the fields, I can push the credential into a Git repository.

But if I change an expression, I cannot push the credential again: I must modify its name to be able to push it.

### To Reproduce

1. Configure the n8n instance to be able to push into a Git repository. 
1. Create a credential of type `Azure Storage Shared Key API`.1. In `Account`, change to expression and put `{{ "test" }}`.
1. In `Key`, change to expression and put `{{ "test" }}`.
1. Click and the button `push` and push the credentials.
1. Edit the credential and chante the `Key` to `{{ "newtest" }}`.
1. Click and the button `push`.

❌ Current behaviour: cannot select the credential in order to push it.

### Expected behavior

Be able to select the credential in order to push it.

### Debug Info

# Debug info

## core

- n8nVersion: 1.118.2
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- consumerId: 4946b7eb-9f2e-497b-9431-286c0f8bdfe7

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2025-11-18T17:41:13.981Z

### Operating System

macOS 15.6.1

### n8n Version

1.118.2

### Node.js Version

22.21.0

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

ireneea:
## Summary

Fixes credentials with expressions not being pushable via source control.

**Root cause:** System couldn't identify or sync non-secret credential values (expressions, numbers, booleans).

### What changed

**Sanitization (`sanitizeCredentialData`):**
- Strips plain strings (secrets) → empty string
- Preserves expressions, numbers, booleans
- Removes `oauthTokenData` and nulls
- Uses `deepCopy` to preserve arrays (e.g., multi-header credentials)

**Merging (`mergeRemoteCrendetialDataIntoLocalCredentialData`):**
- On pull: merges remote expressions/numbers/booleans into local, preserves local secrets
- Arrays: merge by index when lengths match, use remote when they differ
- `isPlainObject` guard handles type mismatches safely

**Change detection (`hasSynchableCredentialDataChanged`):**
- Compares only synchable data so push modal correctly detects modified credentials

**Frontend:**
- Notices in push/pull modals explaining what gets synced

Screenshot 2026-02-10 at 08 45 58

### How to test
1. Create credential with expressions (e.g., `={{ $vars.API_KEY }}`)
2. Modify it and push — should show as modified and push successfully
3. Pull on another instance — expressions/numbers applied, secrets preserved
4. Test multi-value credential (e.g., `httpMultipleHeadersAuth`) — array structure preserved

https://linear.app/n8n/issue/LIGO-79
Fixes #22015

## Review / Merge checklist

- [x] Tests included.


packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -3032,6 +3033,7 @@
 	"settings.sourceControl.modals.pull.autoPublish.options.on.description": "Attempt to publish all workflows",
 	"settings.sourceControl.modals.pull.autoPublishing": "Auto-publish",
 	"settings.sourceControl.modals.pull.willBeArchived": "Archive",
+	"settings.sourceControl.modals.pull.modifiedCredentialsNotice": "Modified credentials will update expressions, numbers, and boolean values. Secret values are preserved locally.",
```
konstantintieber: What does "Secret values are preserved locally" mean?

ireneea: It means they are not replaced or overwritten. Really struggled with the wording for this. So if you raised the questions it means this is not clear enough and might need to be changed

konstantintieber: What does "secret value" refer to in this case?
Anything that's encrypted on the credential?

ireneea: A value that is stubbed. But the word stubbed felt too technical

konstantintieber: Which values will be stubbed? Sorry asking all the noob questions here.

ireneea: Everything that is not an `expression` an `number` or a `boolean` or in other words `strings` that are not `expressions`

konstantintieber: So this means "secret values will not be synced"?

ireneea: Correct. Could also go with

- secret values will not be imported
- secret values will not be synced
- secret values will not be pulled
- secret values will not be replaced
- secret values will not be updated
- ...
 
These are all examples of wording I heard, internally and from users tickets

konstantintieber: How about:
> Modified credentials will update expressions, numbers, and boolean values, any other values will not be synced, because they would be exposed as plain text in the connected git repository.

That helps me understand "the why" as a user.

ireneea: I like that for the `push` 👍, but do you think it also work for the `pull` message?

konstantintieber: True, for pull the text above won't work.
How about:
> Modified credentials will update expressions, numbers, and boolean values, any other values are not available to pull via environments, because ...
Ideas for because part:
- ... they are encrypted with a different encryption key.
- ... secret values cannot be synced.

I wonder: if we could support sharing encryption keys between two instances that are connected via environments, we could also put the encrypted values into the JSON we store on the git repo, right?

ireneea: Went for this:

> Modified credentials will update expressions, numbers, and boolean values, any other values are not available to pull via environments, because secret values cannot be synced.

packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
```diff
@@ -430,12 +439,27 @@ export class SourceControlImportService {
...
+			let data: Record<string, unknown> = {};
+			try {
+				const credentials = new Credentials(
+					{ id: local.id, name: local.name },
+					local.type,
+					local.data,
+				);
+				data = sanitizeCredentialData(credentials.getData());
+			} catch {
+				// Credential data may not be decryptable (e.g. empty or corrupted data)
```
geemanjs, line 455: Should we be doing something here? or just returning an empty object? Feels like a silent failure

geemanjs (approved):

packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
ireneea: This is totally a case of silent failure 😅.  Here is my thinking though. there are two predictable reasons I can think of where this can happen.

1. Then decrypted data is an invalid json
2. We can't decrypt the data (not sure but I think this could happen if the encryption key has changed)

Throwing an error here means that we won't be able to return the "status" and the users will be stuck. They can't pull or push unless they clean the data. Chances are they won't be able to do it through normal UI operations if the data is corrupted (haven't tested that though). 

On the other hand, failing silently here, means that "status" can be computed and the credential will be marked as "modified" and the user will be able to push or pull everything that is not stubbed. Which gives them an opportunity to set things up correctly.

I opted for the later. The downside is that on pull, corrupted credentials are either updated without an opportunity to fix them. And on push corrupted credential will be marked as modified. But the export will actually fail :(

I'm not really sure if this is the best way to deal with this so I'm open to doing it differently

konstantintieber (approved):
Whoop whoop 🥳

PR #25539, opened 2026-02-09
ci: Configure Sentry sourcemaps for releases

shortstacked:
## Summary

Sentry sourcemaps have never worked for the frontend because `RELEASE` never reached Vite during the release build. The `@sentry/vite-plugin` and `sourcemap: !!release` are both gated on `process.env.RELEASE`, but Turbo wasn't passing it through to child processes (`globalEnv`), and `SENTRY_AUTH_TOKEN` wasn't available during the build step for the plugin to authenticate uploads.

### Root cause

1. **`RELEASE` not in Turbo's `globalEnv`** — Turbo stripped it from child processes, so Vite never saw it. Also meant cached dev builds (no sourcemaps) could be served during release builds.
2. **`SENTRY_AUTH_TOKEN` not available during build** — Only injected into the `create-sentry-release` job, not the `publish-to-npm` build step where the Vite plugin runs.
3. **Redundant frontend upload** — The `getsentry/action-release` frontend step uploaded from a separate checkout without Debug IDs, while the deployed code also had no Debug IDs — so Sentry couldn't match either way.

### Changes

| Change | Why |
|--------|-----|
| Added `RELEASE` to `turbo.json` `globalEnv` | Turbo passes it to Vite AND caches release builds separately (sourcemaps + minification produce different output) |
| Added `SENTRY_AUTH_TOKEN` to `turbo.json` `globalPassThroughEnv` | Token reaches the Vite plugin without busting the cache (doesn't affect build output, only the upload side-effect) |
| Added `SENTRY_AUTH_TOKEN` to `publish-to-npm` build step env | `@sentry/vite-plugin` can now authenticate, inject Debug IDs, and upload sourcemaps during the build |
| Removed frontend `getsentry/action-release` step | Redundant — the Vite plugin handles frontend with Debug IDs (Sentry's recommended approach) |
| Bumped `getsentry/action-release` v1.7.0 → v3.5.0 | `version` param deprecated in v3 (now `release`), adds `environment` and `inject` params |
| Added `inject: false` for backend/task-runner | The action runs on a separate checkout from the deployed build — injecting Debug IDs here would mismatch with deployed code |
| Added `strip_common_prefix: true` | Removes CI runner paths from uploaded sourcemaps so file paths match runtime stack traces |
| Added `environment: production` | Enables environment filtering in Sentry |
| Added `fetch-depth: 0` on checkout | Allows associating commits with the Sentry release |

### Note on secrets

`SENTRY_AUTH_TOKEN` is a repo-level secret (not environment-scoped), so it's accessible from the `publish-to-npm` job which runs with `environment: npm`.

## Review / Merge checklist

- [ ] Tests included. 


Matsuuu (approved):
Looks good, thanks for spotting and fixing these! 💯

PR #25557, opened 2026-02-10
fix(HTTP Request Node): Use original URL on first pagination request

pemontto:
## Summary
In the HTTP Request node's `responseContainsNextURL` pagination mode, the Next URL expression is resolved and applied on the very first request, even though `$response` is empty at that point. This means the first request's URL is determined by the Next URL expression instead of the node's main URL field. Because `$response` has no data on the first iteration, users are forced to add a fallback in every Next URL expression to avoid invalid URL errors:

```js
{{ $response.headers["next-url"] || "https://api.example.com/original-endpoint" }}
```
This workaround is required on every workflow that uses `responseContainsNextURL` pagination, and it means duplicating the base URL inside the Next URL expression. It's error-prone (the fallback URL can drift out of sync with the actual URL field) and unintuitive since the UI description for Next URL refers to `$response` variables implying it's evaluated after a response exists.

### Why existing tests didn't catch this
The workflow.pagination.json test fixture contains responseContainsNextURL nodes with expressions like ={{ $response.headers["next-url"] }} and no fallbac, but this file is not referenced by any test runner. It's an unused fixture. There are no tests that actually exercise the responseContainsNextURL pagination code path, so the bug was never caught.

### The fix
The first request now uses the node's configured URL field. The Next URL expression is only applied from the second request onward, which matches the semantic intent, "Next URL" extracts a URL from a response that doesn't exist yet on the first iteration. The `|| fallback` workaround is no longer needed.

### Known related issue: query param duplication

`applyPaginationRequestData` uses `lodash.merge` to combine the original request options with the pagination overrides on every iteration. This means original query parameters (e.g. `api-version`) are merged into subsequent requests even when the Next URL already contains them as query parameters. This causes duplicate parameter errors on APIs like Microsoft Graph REST endpoints. This is pre-existing behaviour and not addressed in this PR.

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/__tests__/request-helper-functions.test.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/__tests__/request-helper-functions.test.ts
@@ -5,6 +5,7 @@ import type { Agent as HttpsAgent } from 'https';
 import { mock, mockDeep } from 'jest-mock-extended';
 import type {
 	IAllExecuteFunctions,
+	IExecuteFunctions,
 	IHttpRequestMethods,
 	IHttpRequestOptions,
 	INode,
@@ -22,6 +23,7 @@ import {
 	applyPaginationRequestData,
 	convertN8nRequestToAxios,
 	createFormDataObject,
+	getRequestHelperFunctions,
 	httpRequest,
 	invokeAxios,
 	parseRequestObject,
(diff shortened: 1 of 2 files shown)


pemontto:
For more context, with the current behaviour this obviously fake domain in the URL works just fine because it's never used:
pag-issue-2


pag-issue-1

PR #25571, opened 2026-02-10
feat(core): Support most tools on Chat hub

Cadiac:
## Summary

We're changing how tools can be added to agents on Chat hub, and now instead of users just being able to choose from couple of preconfigured search tools the whole inventory of n8n tools is made available to them.

In order to support this it must be possible for these users to configure tools without an active workflow canvas, and to do so I've made it possible to open this `ToolSettingsModal` that mimics NDV and allows users to configure the tool. This tool will then be attached to the agent that runs during Chat hub executions.

Users can define multiple copies of the same tool with different settings, and toggle them on and off. Tools defined this way are personal, and they will be used on their personal project (as all Chat hub base LLM chats & personal agent chats run on the personal project).

https://linear.app/n8n/issue/CHA-130

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
heymynameisrob: We have `<Dialog />` in `@n8n/design-system` which would be better to keep the shape

packages/frontend/editor-ui/src/features/ai/chatHub/components/ToolSettingsModal.vue
```diff
@@ -0,0 +1,156 @@
...
+	isValid.value = valid;
+}
+
+function handleNodeNameUpdate(name: string) {
+	nodeName.value = name;
+}
+</script>
+
+<template>
+	<Modal
```
heymynameisrob: We have `<Dialog />` in `@n8n/design-system` now

packages/frontend/editor-ui/src/features/ai/chatHub/components/ToolsSelector.vue
```diff
@@ -1,92 +1,247 @@
...
+
 .toolsButton {
 	display: flex;
 	align-items: center;
 	gap: var(--spacing--2xs);
-
-	&.transparentBg {
-		background-color: transparent !important;
-	}
+	border: none !important;
```
heymynameisrob: I know the design had this but I think it's better if we keep this the same. I was just messing around 😂

packages/frontend/editor-ui/src/features/ai/chatHub/components/ToolsSelector.vue
heymynameisrob, line 1: Image

When we have more than 3 would be good if the 3rd icon in the stack showed the remaining count  (e.g +3) so users know that others are applied still

packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
Cadiac: Changed these to use the N8nDialog component. This revealed some problems on that component though, I just went ahead and changed the background color & borders used there to match other components we have. Also touched the z-index so that these dialogs get drawn at a reasonable z-index & below modals.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ToolsSelector.vue
Cadiac: image

changed this. to make this work the tools on the right are always drawn on top now

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
```diff
@@ -0,0 +1,211 @@
...
+		logger,
+		migrationName,
+	}: MigrationContext) {
+		// Create the chat_hub_tools table with type and typeVersion columns
+		await createTable(table.tools)
+			.withColumns(
+				column('id').uuid.primary,
+				column('name').varchar(255).notNull,
+				column('type').varchar(255).notNull,
+				column('typeVersion').double.notNull,
```
Cadiac, line 51: Surprisingly we don't seem to have any float fields in our database?
This would be storing version numbers like "4.4", we don't have immediate use case for this but this might turn out to be very handy in future - but in theory this is also contained within the INode `definition` json column, and just denormalized here for easy viewing on icons and such.

What do you think, should I just drop the type and typeVersion fields, or keep them around for future (and enforce that they match the `definition` with application logic)?

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
```diff
@@ -0,0 +1,211 @@
...
+				column('ownerId').uuid.notNull,
+				column('definition').json.notNull,
```
Cadiac, line 53: This will contain full `INode` definition of the tool node.

packages/frontend/@n8n/design-system/src/components/N8nButton/Button.vue
```diff
@@ -352,6 +352,8 @@ $loading-overlay-background-color: rgba(255, 255, 255, 0);
 	display: inline-flex;
 	justify-content: center;
 	align-items: center;
+	/* maintain component height regardless of icon */
+	margin-block: -4px;
```
Cadiac: This change was something AI thought (?) necessary, I kind of want to undo this 🤔

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
```diff
@@ -83,6 +83,22 @@ const sizeClass = computed(() => sizeClasses[props.size]);
 /** ARIA Fallbacks: These are visually hidden but accessible to screen readers **/
 const needsFallbackTitle = computed(() => !!props.ariaLabel);
 const needsFallbackDescription = computed(() => !!props.ariaDescription);
+
```
Cadiac, line 86: The <N8nDialog> component is a new one without uses in the app yet, and turns out it had some early pains that I thought reasonable to just fix on this component.

Now modals can be opened from it, poppers, dropdowns etc work, its overlay doesn't get drawn on top of other modals, modals can be closed while a dialog is open, its background color is the same token as on modals..

packages/frontend/@n8n/design-system/src/components/N8nButton/Button.vue
Cadiac: Actually no, this was changed a month ago in the early version by Suguru..

Cadiac: Cancelled this change anyways as it didn't appear necessary.

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
CharlieKolb: I'd expect the typeVersion to be a string, or separate stores major and minor numbers 🤔 Double seems sketchy with floating point inaccuracies

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
CharlieKolb: Is that including all the parameters for all the node versions etc?

Not having read the rest of the PR yet, do we lazily load this on first chat hub interaction or is this gonna use up quite a lot of storage across all cloud instances? Probably looking at many TB here if I understand correctly? 🤔

packages/frontend/editor-ui/src/features/shared/nodeCreator/components/NodeCreator.vue
Cadiac, line 1: hm, these changes might not be necessary anymore - undoing

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
Cadiac: > Is that including all the parameters for all the node versions etc?

The `definition` is `INode` definition of the tool, and what we store might look like this

```
{
  "id": "64138f98-16f5-4bc3-9d4d-50efa03393a8",
  "name": "Get many bases in Airtable",
  "typeVersion": 2.1,
  "type": "n8n-nodes-base.airtableTool",
  "position": [0, 0],
  "parameters": {
    "descriptionType": "auto",
    "authentication": "airtableTokenApi",
    "resource": "base",
    "limit": 100,
    "operation": "getMany",
    "returnAll": "={{ /*n8n-auto-generated-fromAI-override*/ $fromAI('Return_All', ``, 'boolean') }}",
    "options": {}
  },
  "credentials": {
    "airtableTokenApi": {
      "id": "6PNqPfQ6LKwcy7JG",
      "name": "Airtable Personal Access Token account"
    }
  }
}
```

so not the full node, just the configuration to use it on a workflow

Cadiac: I'm not that worried about storage - these are a lot smaller than workflows (basically just single node fragments from it), and realistically users won't create that many of these, and now (unlike before) sessions / personal agents just reference these tool configurations created once, instead of creating copies and keeping them on `tools` as `INode[]` arrays like before 🤔

CharlieKolb: Ah sorry I was thinking of the node type properties. That's all good then 👍

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
Cadiac: `typeVersion` on INode's is a numeric field, it's from the version field of the node.

We're not going to have inaccuracies in storing javascript values like `4.4` into the field and back (as we're not doing math with them), i'm more worried of people putting something other than version strings there if its a text. but happy to change it into a short varchar too if we feel like it, or to just drop these extra fields until we have clear use for them, this was just something me and Surugu thought worth including here earlier

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
```diff
@@ -0,0 +1,211 @@
...
+				const tools = parseJson<ToolDefinition[]>(raw);
+				if (!Array.isArray(tools)) {
+					logger.warn(
+						`[${migrationName}] Tools column for ${entityType} ${entityId} is not an array. Skipping.`,
+					);
+					return [];
+				}
+				return tools;
+			} catch (error) {
+				logger.warn(
```
Cadiac, line 145: Data loss here is _okay_, if that means we can pass this migration. The existing tools aren't that useful and they should be quick to redefine.

packages/@n8n/db/src/migrations/common/1770000000000-CreateChatHubToolsTable.ts
CharlieKolb: Until someone starts naming their versions 100000.0001 for some reason :D But tbf this would also be an issue in the actual JS code, so don't mind me

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
```diff
@@ -142,15 +158,15 @@ const needsFallbackDescription = computed(() => !!props.ariaDescription);
 	transform: translate(-50%, -50%);
 	width: 100%;
 	padding: var(--spacing--lg);
+	border: var(--border);
```
heymynameisrob: Dont need this. The border is applied by the `box-shadow`

Cadiac: it looks buggy

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
```diff
@@ -142,15 +158,15 @@ const needsFallbackDescription = computed(() => !!props.ariaDescription);
...
 	border-radius: var(--radius--lg);
-	background-color: light-dark(var(--color--neutral-white), var(--color--neutral-800));
+	background-color: var(--dialog--color--background);
 	box-shadow:
-		inset 0 1.5px 0 light-dark(transparent, var(--border-color)),
```
heymynameisrob, line 148: Replace if we're removing `border`

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
```diff
@@ -142,15 +158,15 @@ const needsFallbackDescription = computed(() => !!props.ariaDescription);
 	transform: translate(-50%, -50%);
 	width: 100%;
 	padding: var(--spacing--lg);
+	border: var(--border);
 	border-radius: var(--radius--lg);
-	background-color: light-dark(var(--color--neutral-white), var(--color--neutral-800));
+	background-color: var(--dialog--color--background);
```
Cadiac: @heymynameisrob , 

Image

if I roll back the changes here this is what the dialog component looks like on ToolSettingsModal with dark theme. The AI override buttons are pretty dull and hard to see, even with the new button styles, and having used n8n this looks _very odd_, but if this is our new design I can roll this back and we can go with this...

packages/cli/src/modules/chat-hub/chat-hub-agent.service.ts
```diff
@@ -52,6 +55,16 @@ export class ChatHubAgentService {
 		return await this.chatAgentRepository.getManyByUserId(userId);
 	}
 
+	async getAgentsByUserIdAsDtos(userId: string): Promise<ChatHubAgentDto[]> {
+		const agents = await this.chatAgentRepository.getManyByUserId(userId);
+		return await Promise.all(
+			agents.map(async (agent) => {
```
CharlieKolb: Don't have full context, so excuse me if these are naive questions, but how many agents do we expect busy instances to have?

Could this become a harmfully high number of agents/async fns/db calls in a large instance where many admins have access to everyone's agents?

Do we need to load all tool ids for the generic getAgents endpoint or could we load the toolIds only for an actively selected agent?

It looks like we could change `getToolIdsForAgent` to support multiple agents also?

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
heymynameisrob: Keep the background color change, just not the border and shadow ones you did.

packages/cli/src/modules/chat-hub/chat-hub-agent.service.ts
Cadiac: No you're correct, this isn't a good way to do this.

this is probably good enough for monday's release (our current usage numbers are low, only handful of agents exist, they aren't shared), but this here is a messy implementation.

I'll try to see if I have time to make this more efficient.

CharlieKolb: I'd be happy which changing the called function to support a list of agentIds instead for now, reducing this to O(agents) calls from O(agents * tools)

packages/frontend/@n8n/design-system/src/components/N8nDialog/DialogContent.vue
Cadiac: image

With the background change the buggy looking border from box shadow (at the top) becomes visible on dark theme, unless we intended it to have this kind of mild 3D effect from it? Or at least my thought was that if I ship it looking like this there will be bug ticket opened next week about it missing the border on other sides.

image

Cadiac: But happy to just ship this as well, at least the fromAI buttons would have some contrast

Cadiac: anyways, changed back to the darker backgroudn override, lets go with this.

packages/cli/src/modules/chat-hub/chat-hub-agent.service.ts
Cadiac: changed this to a left join that gets the tool ids instead

CharlieKolb (dismissed):
🚀

tomi (dismissed):
🚀

tomi (approved):

PR #25575, opened 2026-02-10
refactor(ai-builder): Remove community node support from code builder

mutdmour:
## Summary
Remove community nodes from code-builder MVP
- Filter community nodes from code builder's `filterNodeTypes()` — only built-in nodes (`n8n-nodes-base.*`, `@n8n/*`) are now supported
- Remove `generateForPackage`/`removeForPackage` calls from `CommunityPackagesService` and its `NodeDefinitionGeneratorService` dependency
- Simplify `NodeDefinitionGeneratorService` to only resolve built-in definition directories
- Remove `nodeDefinitionsDir` property from `InstanceSettings`
- Remove `~/.n8n/node-definitions` fallback and debug `console.log` statements from `code-builder-get.tool.ts`
- Update `generate-types.ts` dev script to use local output path instead of `~/.n8n/node-definitions`

## Context
When the code builder agent calls `get_node_types` for a community node, it fails because the type definitions path was renamed (`generated-types` → `node-definitions`) without migration. Rather than fixing the path resolution, this removes community node support from the code builder entirely and cleans up all `~/.n8n/node-definitions` references.

**Kept:** `dist/node-definitions` references (built-in package build-time paths)
**Removed:** All `~/.n8n/node-definitions` references and community node definition generation

## Test plan
- [x] `packages/@n8n/ai-workflow-builder.ee` — 2047 tests pass (124 suites)
- [x] `packages/cli` community-packages — 190 tests pass
- [x] `packages/cli` node-definition-generator — 6 tests pass
- [x] `packages/core` instance-settings — 24 tests pass
- [x] `pnpm typecheck` — 85/85 tasks pass


packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-get.tool.ts
```diff
@@ -52,17 +51,14 @@ export function validatePathWithinBase(filePath: string, baseDir: string): boole
...
  */
 function getGeneratedNodesPaths(nodeDefinitionDirs?: string[]): string[] {
 	if (nodeDefinitionDirs && nodeDefinitionDirs.length > 0) {
 		return nodeDefinitionDirs.map((dir) => join(dir, 'nodes'));
 	}
 
-	// Default to ~/.n8n/node-definitions (same location as runtime and CLI)
-	const defaultTypesDir = join(homedir(), '.n8n', 'node-definitions');
-	return [join(defaultTypesDir, 'nodes')];
+	return [];
```
aalises, line 61: would it be worth it to log some warning here? now it's silently failing but maybe it's alright

mutdmour: there's a warning later in the flow

packages/cli/src/services/node-definition-generator.service.ts
aalises, line 1: Maybe now `NodeDefinitionGeneratorService` is not even needed anymore? it only has two methods, a one line utility and `getNodeDefinitionDirs()`... not sure if we're using it in a lot of places but seems like we could just use the methods as-is instead of encapsulating in a service

aalises (approved):

PR #25577, opened 2026-02-10
fix(editor): Fix bug with two consecutive saves

dariacodes:
### Problems
- Updating the workflow name or tags ("manual saves") triggered immediate saves without coordination with other save operations (so two manual saves in a row created a race condition)
- No single source of truth for "is a save in progress" 
    - Save coordination checked `uiStore.isActionActive.workflowSaving` first, then `autosaveStore.pendingAutoSave` (the actual promise)
- Conflict resolution (“force save”) could still race with other saves

### Solution
- Centralized save coordination: only one save runs at a time, tracked via `autosaveStore.pendingAutoSave` promise
    - Manual saves wait for in-progress saves to complete before proceeding
    - Autosaves skip if a save is already running (now aware of manual saves, not just other autosaves)
    - Conflict resolution saves bypass the wait entirely (using forceSave=true)
- Name and tag updates now schedule autosaves: Instead of triggering immediate saves, they:                                                                  
    - Update state optimistically                                                                                                                              
    - Mark the workflow as dirty                                                                                                                               
    - Schedule a debounced autosave                                                                                                                            
    - This allows multiple rapid changes to be batched into a single save
    - Removed updating tags after receiving the save response - they're now updated optimistically like all other workflow properties
- Simplified API: Removed name and tags parameters from saveCurrentWorkflow() - values are always read from current workflow state

Closes [ADO-4621](https://linear.app/n8n/issue/ADO-4621/bug-two-manual-saves-in-a-row-show-an-error)

## Review / Merge checklist

- [ ] Tests included. 


scdekov (approved):
Looks great!

PR #25598, opened 2026-02-10
fix(editor): Show existing credentials in template setup modal

RicardoE105:
## Summary

When importing a template via the "Set up template" modal, users only saw "Create new [credential]" buttons even when they already had matching credentials. This PR fixes three issues that caused this:

- **Remove broken `personal-only` filter**: The `CredentialPicker` in `SetupTemplateFormStep.vue` had a `personal-only` prop that filters credentials by `credential.homeProject?.type === 'personal'`. However, the `/credentials/for-workflow` backend endpoint does not return `homeProject` in its response, so `homeProject` was always `undefined` — filtering out every credential. Removing this prop is safe because the backend already enforces access control, only returning credentials the user has permission to use.
- **Load credentials during template import**: `useWorkflowInitialization.ts` was not loading credentials and credential types before the template setup modal opened, so the credential store was empty. Added `loadCredentials()` and `fetchCredentialTypes()` calls during template route initialization.
- **Prevent auto-assign toast on mount**: `NodeCredentials.vue` auto-selects the most recent credential on mount, which triggered the "credential auto-assigned to other nodes" toast. Added an `isUserAction` parameter to `onCredentialSelected` so auto-assign only runs for explicit user actions.

https://linear.app/n8n/issue/ADO-4782

## Review / Merge checklist

- [ ] Tests included.


packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -194,7 +194,7 @@ watch(
 			allOptions[0],
 		);
 
-		onCredentialSelected(mostRecentCredential.type, mostRecentCredential.id);
+		onCredentialSelected(mostRecentCredential.type, mostRecentCredential.id, false, false);
```
MiloradFilipovic: Two questions:
1. Why are we disabling auto-assign? Feels like it's very convenient and we had some customer requests to bring it to the setup modal also. We should check this with product
2. Would be nice to document these parameters:
```suggestion
		onCredentialSelected(
			mostRecentCredential.type,
			mostRecentCredential.id,
			false, // showAuthOptions
			false, // isUserAction
		);
```

MiloradFilipovic (dismissed):
Works well 🙌
We should just check about disabling auto assign with product.

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
MiloradFilipovic: Also, should we add some tests for this?

RicardoE105: We're not disabling auto-assign for user actions — it still works when a user explicitly picks a credential from the dropdown. The isUserAction = false here only prevents the auto-assign toast from firing on mount, when the watcher auto-selects the most recent credential. Without this, opening a node that has matching credentials immediately triggers the "credential auto-assigned to X other nodes" toast, which is confusing since the user didn't do anything.

RicardoE105: added a small test

MiloradFilipovic (approved):
🚢

PR #25602, opened 2026-02-10
fix(editor): Skip credential auto-assignment when node displayOptions don't match

RicardoE105:
## Summary

When selecting a credential (e.g. Header Auth) in the HTTP Request node, the system incorrectly auto-assigns it to other nodes that **support** the same credential type, even if those nodes don't have the credential **active** based on their current parameters.

For example, the Webhook node type supports `httpHeaderAuth`, but only when its `authentication` parameter is set to `'headerAuth'`. With default parameters (`authentication: 'none'`), the credential should not be auto-assigned. The toast "Added this credential to 1 other node(s)" appeared incorrectly.

**Root cause:** `assignCredentialToMatchingNodes` only checked if the node type *supports* the credential via `.some()`, without checking `displayOptions` against the node's current parameters.

**Fix:** Use `.find()` to get the credential description, then check `credentialDescription.displayOptions` against the node's parameters using the existing `nodeHelpers.displayParameter()` function. This matches the pattern already used in:
- `useNodeCredentialOptions.ts` (NDV credential display)
- `useNodeHelpers.ts` (credential issue checking)

https://linear.app/n8n/issue/ADO-4786

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/app/stores/workflows.store.test.ts
+++ b/packages/frontend/editor-ui/src/app/stores/workflows.store.test.ts
@@ -2123,6 +2123,71 @@ describe('useWorkflowsStore', () => {
 			expect(workflowsStore.workflow.nodes[1].credentials).toBeUndefined();
 		});
 
+		it('should not assign credential to nodes where displayOptions do not match current parameters', () => {
+			const credential = { id: 'cred-1', name: 'Header Auth Credential' };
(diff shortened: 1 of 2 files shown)


MiloradFilipovic (approved):
Nice, thanks for addressing this 👌

PR #25608, opened 2026-02-11
feat(ai-builder): Skip redundant node discovery in code-builder when approved plan is provided

mutdmour:
## Summary
When the code-builder receives an approved plan, the plan already contains suggested nodes per step. Instead of appending override instructions, dynamically build the mandatory workflow steps:
- Step 1 is replaced with a plan-specific version that reads the plan
- Step 2a (get_suggested_nodes) is omitted entirely
- Steps 2b, 2c, and 3 use plan-aware variants
- The get_suggested_nodes tool is excluded from LLM bindings

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-search.tool.ts
```diff
@@ -408,6 +408,48 @@ function formatDiscriminatorInfo(info: DiscriminatorInfo, nodeId: string): strin
...
+ */
+export function formatNodeResult(
+	nodeTypeParser: NodeTypeParser,
+	nodeId: string,
+	version?: number,
+): string | undefined {
+	const nodeType = nodeTypeParser.getNodeType(nodeId, version);
+	if (!nodeType) return undefined;
+
+	const resolvedVersion =
```
aalises, line 425: Should this version resolved be consistent with the one on the node type parser? (using the `max`)

mutdmour: yes that would be cleaner

packages/@n8n/ai-workflow-builder.ee/src/code-builder/handlers/chat-setup-handler.ts
```diff
@@ -255,4 +276,41 @@ export class ChatSetupHandler {
...
 	}
+
+	/**
+	 * Pre-fetch node info for the plan's suggestedNodes using direct NodeTypeParser lookup.
+	 * Returns the formatted result string, or undefined if not applicable.
+	 */
+	private prefetchSearchResults(plan: PlanOutput): string | undefined {
+		const nodeNames = extractNodeNamesFromPlan(plan);
+		if (nodeNames.length === 0) return undefined;
+		if (!this.nodeTypeParser) return undefined;
```
aalises: can it happen that this is undefined by mistake? (we didn't instance with node type parser)... if so maybe logging some warning would be nice, or forcing node type parser to be defined

aalises (approved):

PR #25609, opened 2026-02-11
fix(ai-builder): Improve plan mode UX: naming, question skipping, thinking states (no-changelog)

OlegIvaniv:
## Summary

Multiple fixes and improvements to the AI Builder's plan mode experience, covering workflow naming, question skipping, thinking states, and UI polish.

### Workflow naming in plan mode
- Stream the generated workflow name to the frontend immediately after creation. Previously, in plan mode the graph interrupted before `process_operations` ran, so the frontend never received the name.

### Question skipping
- The Next button is always enabled. When no answer is provided, it shows "Skip" (or "Skip & submit" on the last question) and marks the question as skipped.
- Skipped questions are excluded from the discovery context sent to the backend (already supported).

### Intro message persistence
- The questions message is now always rendered (collapsed to just the intro text after answers are submitted), so the intro message remains visible in the chat history and on session reload.

### Thinking/waiting state fixes
- **Thinking indicator preserved during streaming gaps:** After submitting answers, the "Thinking" indicator from `prepareForStreaming` is no longer cleared by empty stream chunks. The stream callback now only updates `builderThinkingMessage` when there's an explicit clear or new value.
- **"Workflow generated" label fixed:** The naming-only `workflow-updated` event (no nodes) no longer causes discovery tool groups to show "Workflow generated". Only `workflow-updated` events that follow tool messages in a region are counted.
- **Thinking placeholder for new turns:** The standalone thinking placeholder now shows for new turns even when old tool groups exist from previous turns.

### UI polish
- "Other" input stretches to full container width
- Radio options use `align-items: center`
- Submit button uses primary style on the last question when a valid answer is provided, label shortened to "Submit"
- Plan/Build mode selector is never disabled

### Code quality
- Refactored `PlanQuestionsMessage.vue`: extracted `__other__` magic string to constant, removed duplicate `getCurrentAnswer()`, consolidated answer initialization into single `watch({ immediate: true })`, simplified validation logic, fixed CSS variable usage.
## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/UserAnswersMessage.vue
```diff
@@ -48,7 +47,10 @@ function formatAnswer(answer: PlanMode.QuestionResponse): string {
 			<N8nText :bold="true" :class="$style.question">
 				{{ answer.question }}
 			</N8nText>
-			<N8nText color="text-light">{{ formatAnswer(answer) }}</N8nText>
+			<N8nText v-if="answer.skipped" :class="$style.skipped">
```
aalises: does the `style.skipped` exist (cannot see it), we can remove if so or add it back

packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/UserAnswersMessage.vue
```diff
@@ -20,8 +20,7 @@ interface Props {
 const props = defineProps<Props>();
 const i18n = useI18n();
 
-// Filter out skipped questions
-const displayedAnswers = computed(() => props.answers.filter((answer) => !answer.skipped));
+const displayedAnswers = computed(() => props.answers);
```
aalises: We can just use `props.answers` directly right?

packages/frontend/@n8n/design-system/src/components/AskAssistantChat/AskAssistantChat.vue
```diff
@@ -311,14 +316,12 @@ const showSuggestions = computed(() => {
...
 
-// Check if we have any thinking group (tool messages grouped into thinking blocks)
-const hasAnyThinkingGroup = computed(() => {
-	return normalizedMessages.value.some((msg) => msg.type === 'thinking-group');
-});
-
-// Show placeholder when streaming with loading message but no tool messages have arrived yet
+// Show placeholder when streaming with loading message but no active tool group for the current turn.
+// Check the last message — if it's not a thinking-group, the current turn has no tools yet.
 const showThinkingPlaceholder = computed(() => {
```
aalises, line 321: This assumes the thinking-group is always the last message for the current turn while tools are streaming. To double check that it's like this

OlegIvaniv: Yes, this is correct. `groupToolMessagesIntoThinking` guarantees it, a tool group is only active when there are no messages after it. Once a text response arrives, it becomes the last message, but we don't need the placeholder at that point anyway. The test `should not add thinking spinner to old completed group when new turn is streaming` covers this

aalises (dismissed):

aalises (approved):

PR #25626, opened 2026-02-11
perf(core): Optimize generate-node-defs with parallel writes and hash-based skip

mutdmour:
## Summary

Optimizes the `n8n-generate-node-defs` build step which was adding ~34s to every build.

- **Parallel file writes**: Batch all file writes (100 at a time via `Promise.all`) instead of sequential `await`s. The step was 61% CPU utilization (I/O bound), so parallelism gives a significant speedup.
- **Hash-based skip**: Compute SHA-256 of `nodes.json` content + SDK version, store as `.nodes-hash` sentinel. Skip generation entirely when hash matches on subsequent runs.
- **Unified bin entry**: Refactored `packages/core/bin/generate-node-defs` to import `generateNodeDefinitions()` from the CLI module, eliminating duplicated logic.

### Results (nodes-base, 418 nodes, ~11,845 files)

| Scenario | Before | After |
|----------|--------|-------|
| Clean generation | ~27s | **~9s** |
| Incremental (unchanged) | ~27s | **<1s** |

## Test plan

- [x] 10 unit tests pass (`pnpm jest generate-node-defs-cli`)
- [x] Hash skip: second run with unchanged nodes.json is near-instant
- [x] Hash regeneration: changing nodes.json triggers full regeneration
- [x] Hash regeneration: missing sentinel triggers full regeneration
- [x] Parallel writes: multiple nodes produce correct file count and content
- [x] `pnpm typecheck` passes
- [x] `pnpm build` passes from root

https://linear.app/n8n/issue/AI-2021

### DIFF
--- a/packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
+++ b/packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
@@ -92,10 +92,6 @@ export async function generateNodeDefinitions(
 		}
 	}
 
-	// Clean output directory before regenerating to remove stale files from deleted nodes
-	const nodesDir = path.join(outputDir, 'nodes');
-	await fs.promises.rm(nodesDir, { recursive: true, force: true });
-
 	const result = await orchestrateGeneration({ nodes, outputDir });
 
 	// Write hash sentinel after successful generation
--- a/packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
+++ b/packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
@@ -75,7 +75,7 @@ export async function generateNodeDefinitions(
 	try {
 		const existingHash = await fs.promises.readFile(hashFilePath, 'utf-8');
 		if (existingHash.trim() === inputHash) {
-			console.log(`Node definitions up to date (hash match), skipping generation.`);
+			console.log('Node definitions up to date (hash match), skipping generation.');
 			return;
 		}
 	} catch {
--- a/packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
+++ b/packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
@@ -92,6 +92,10 @@ export async function generateNodeDefinitions(
 		}
 	}
 
+	// Clean output directory before regenerating to remove stale files from deleted nodes
+	const nodesDir = path.join(outputDir, 'nodes');
+	await fs.promises.rm(nodesDir, { recursive: true, force: true });
+
 	const result = await orchestrateGeneration({ nodes, outputDir });
 
 	// Write hash sentinel after successful generation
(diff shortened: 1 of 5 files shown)


packages/@n8n/workflow-sdk/src/generate-types/generate-node-defs-cli.ts
```diff
@@ -47,6 +93,11 @@ export async function generateNodeDefinitions(
 	}
 
 	const result = await orchestrateGeneration({ nodes, outputDir });
```
aalises, line 95: In here I would consider cleaning the ooutput directory before regenerating, as I think that when nodes are removed from `nodes.json`, the hash changes so regeneration triggers, but only new files are written. Old node definition directories (e.g., for a deleted node) persist

packages/@n8n/workflow-sdk/src/generate-types/generate-types.ts
```diff
@@ -3782,16 +3782,17 @@ function convertNodeToToolVariant(node: NodeTypeDescription): NodeTypeDescriptio
...
+/** Batch size for parallel file writes to avoid file descriptor exhaustion */
+const WRITE_BATCH_SIZE = 100;
+
+/**
+ * Write files in parallel batches to stay within OS file descriptor limits.
+ */
+async function writeFilesInBatches(files: Array<{ path: string; content: string }>): Promise<void> {
+	for (let i = 0; i < files.length; i += WRITE_BATCH_SIZE) {
+		const batch = files.slice(i, i + WRITE_BATCH_SIZE);
+		await Promise.all(batch.map(async (f) => await fs.promises.writeFile(f.path, f.content)));
```
aalises: the async/await wrapper is redundant here I think as writeFile already returns a Promise, and Promise.all handles it. 

Can maybe simplify to: `await Promise.all(batch.map((f) => fs.promises.writeFile(f.path, f.content)));`

packages/@n8n/workflow-sdk/src/generate-types/generate-types.ts
```diff
@@ -3817,23 +3818,20 @@ async function generateVersionSpecificFiles(
...
-	let flatVersions = 0;
+
+	// Phase 1: Pure computation — generate all file contents in memory
+	const allDirs = new Set<string>();
+	const allWrites: Array<{ path: string; content: string }> = [];
+	// Track split plans that need their own directory creation + writes
+	const splitPlans: Array<{ baseDir: string; plan: Map<string, string> }> = [];
 
 	for (const [nodeName, nodes] of nodesByName) {
 		try {
```
aalises, line 3826: How do we handle partial write errors? if an error is thrown midway through processing, those partial entries still might end up in allWrites and get written. This could produce some half-baked node definition on disk

Maybe we can do the database transaction approach and store the writes in some local array, and only if all passed we "commit" them to write in one transaction at the end of the try block, instead of having half-pushes

aalises (dismissed):
thanks for the fast addressing of the feedback! 🚀

aalises (approved):

PR #25628, opened 2026-02-11
feat(core): Implement service and controllers for gradual publishing

schrothbn:
## Summary
Implement gradual publish on the backend. 
- Add endpoints to set and delete gradual publish settings
- Handle getting the right workflow version on different execution paths

### DIFF
--- a/packages/cli/src/workflows/resolve-execution-version.ts
+++ b/packages/cli/src/workflows/resolve-execution-version.ts
@@ -1,0 +1,56 @@
+import { UnexpectedError } from 'n8n-workflow';
+import type { INode, IConnections } from 'n8n-workflow';
+
+interface VersionSource {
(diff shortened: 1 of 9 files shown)


packages/cli/src/workflows/resolve-execution-version.ts
```diff
@@ -0,0 +1,56 @@
...
+				connections: gradualRolloutVersion.connections,
+				isRolloutVersion: true,
+			};
+		}
+	}
+
+	return {
+		nodes: activeVersion.nodes,
+		connections: activeVersion.connections,
+		isRolloutVersion: false,
```
nikhilkuria, line 54: I don't think we need to return this property

nikhilkuria (approved):

PR #25639, opened 2026-02-11
fix(core): Wrap webhook CRUD operations in transactions

cstuncsik:
## Summary

Webhook creation and deletion operations (`addWebhooks()` and `remove()`) now happen atomically within database transactions. This prevents orphaned webhook records when partial operations fail or race conditions occur.

**Changes:**
- `addWebhooks()`: Store all webhook entities in single transaction before creating external webhooks; on external failure, atomically roll back DB records
- `remove()`: Wrap DB cleanup in transaction after external webhook deletion
- `reconcileWebhooks()`: New startup method cleans orphaned webhook records for failed activations

All 64 tests pass (51 unit + 13 integration).

Resolves TODOs at:
- `packages/cli/src/active-workflow-manager.ts:187`
- `packages/cli/src/active-workflow-manager.ts:888`

## Review / Merge checklist

- [x] Tests included (64 tests: 51 unit + 13 integration, all passing)
- [ ] PR Labeled with `release/backport` if needed

### DIFF
--- a/packages/cli/src/active-workflow-manager.ts
+++ b/packages/cli/src/active-workflow-manager.ts
@@ -8,7 +8,7 @@ import {
 import { Logger } from '@n8n/backend-common';
 import { WorkflowsConfig } from '@n8n/config';
 import type { WorkflowEntity, IWorkflowDb } from '@n8n/db';
-import { WorkflowRepository } from '@n8n/db';
+import { WebhookEntity, WorkflowRepository, withTransaction } from '@n8n/db';
 import { OnLeaderStepdown, OnLeaderTakeover, OnPubSubEvent, OnShutdown } from '@n8n/decorators';
 import { Service } from '@n8n/di';
 import chunk from 'lodash/chunk';
@@ -108,6 +108,8 @@ export class ActiveWorkflowManager {
 
 		await this.addActiveWorkflows('init');
 
+		await this.reconcileWebhooks();
+
 		await this.externalHooks.run('activeWorkflows.initialized');
 	}
 
@@ -154,15 +156,16 @@ export class ActiveWorkflowManager {
 		activation: WorkflowActivateMode,
 	) {
 		const webhooks = WebhookHelpers.getWorkflowWebhooks(workflow, additionalData, undefined, true);
-		let path = '';
 
 		if (webhooks.length === 0) return false;
 
+		// Phase 1: Build webhook entities and normalize paths
+		const webhookEntities: WebhookEntity[] = [];
 		for (const webhookData of webhooks) {
 			const node = workflow.getNode(webhookData.node) as INode;
 			node.name = webhookData.node;
 
-			path = webhookData.path;
+			const path = webhookData.path;
 
 			const webhook = this.webhookService.createWebhook({
 				workflowId: webhookData.workflowId,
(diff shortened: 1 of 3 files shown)


packages/cli/src/active-workflow-manager.ts
```diff
@@ -183,45 +186,58 @@ export class ActiveWorkflowManager {
...
-				try {
-					await this.clearWebhooks(workflow.id);
-				} catch (error1) {
-					this.errorReporter.error(error1);
-					this.logger.error(
-						`Could not remove webhooks of workflow "${workflow.id}" because of error: "${error1.message}"`,
-					);
+		// Phase 2: Store all webhooks to DB in a single transaction
+		try {
+			await withTransaction(this.workflowRepository.manager, undefined, async (em) => {
```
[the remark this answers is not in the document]
cstuncsik: Intentional, the transaction boundary spans multiple webhook stores in a loop, so the caller controls when to commit/rollback. Moving it into webhookService.storeWebhook would lose the ability to batch multiple stores atomically.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -183,45 +186,58 @@ export class ActiveWorkflowManager {
...
+				for (const webhook of webhookEntities) {
+					await this.webhookService.storeWebhook(webhook, em);
```
afitzek, line 198: I think we should also include the await this.webhookService.createWebhookIfNotExists(workflow, webhookData, mode, activation); into the transaction, that seems to be the reason to introduce a transaction in the first place. Otherwise, it is just a single bulk insert with a cleanup later.

afitzek (commented):
I think the idea would be to add more into the transaction and avoid the manual cleanup in the catch therefore?

packages/cli/src/active-workflow-manager.ts
cstuncsik: Updated

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -134,14 +134,31 @@ export class WebhookService {
 			});
 		}
 
-		await this.webhookRepository.upsert(webhook, ['method', 'webhookPath']);
+		if (em) {
+			await em.upsert(WebhookEntity, webhook, ['method', 'webhookPath']);
```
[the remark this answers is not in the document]
cstuncsik: Good catch — fixed. Moved `cacheService.set()` after the DB write so a transaction rollback won't leave stale webhook entries in cache.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -183,45 +186,59 @@ export class ActiveWorkflowManager {
...
+			if (['init', 'leadershipChange'].includes(activation) && error.name === 'QueryFailedError') {
+				// n8n does not remove the registered webhooks on exit.
+				// This means that further initializations will always fail
+				// when inserting to database. This is why we ignore this error
+				// as it's expected to happen.
+				return true;
+			}
 
-				throw error;
+			if (error instanceof Error && error.name === 'QueryFailedError') {
```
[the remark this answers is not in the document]
cstuncsik: The `QueryFailedError` from the bulk transaction doesn't identify which specific insert caused the conflict, so `webhookEntities[0]?.node` is a best-effort fallback. Not practically fixable without catching per-insert errors, which would defeat the purpose of the batch transaction.

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -125,23 +125,40 @@ export class WebhookService {
...
-	async deleteWorkflowWebhooks(workflowId: string) {
+	async findAllWebhooks() {
+		return await this.webhookRepository.find();
+	}
+
+	async deleteWorkflowWebhooks(workflowId: string, em?: EntityManager) {
+		if (em) {
+			const webhooks = await em.findBy(WebhookEntity, { workflowId });
+			if (webhooks.length) {
+				void this.cacheService.deleteMany(webhooks.map((w) => w.cacheKey));
```
afitzek, line 156: Here is a very small window of a race condition: Inside the transaction we remove the cache entry for the webhook, while the transaction is not commited, the cache is setup again, by a different request reading the webhook from the database and filling the cache again. The webhooks that are deleted here, are only removed from the database on commit, meaning this can result in webhooks being in the cache that are deleted in the database.

packages/cli/src/webhooks/webhook.service.ts
```diff
@@ -125,23 +125,40 @@ export class WebhookService {
 		return await this.findCached(method, path);
 	}
 
-	async storeWebhook(webhook: WebhookEntity) {
+	async storeWebhook(webhook: WebhookEntity, em?: EntityManager) {
+		if (em) {
+			await em.upsert(WebhookEntity, webhook, ['method', 'webhookPath']);
```
afitzek, line 130: I think since we are running in a transaction here, and we write the cache in line 136, this can become the same race condition. If the transaction is rolled back, there can be entries in the cache that are not in the DB.

PR #25646, opened 2026-02-11
feat: Add tab scoped collaboration

dariacodes:
## Summary

Switch from user-level write-lock ownership to tab-level for collaboration. For all read-only tabs show that another tab is currently editing with a possibility to acquire editing:
image

Closes [ADO-4829](https://linear.app/n8n/issue/ADO-4829/feature-tab-scoped-collaboration)

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/collaboration/collaboration.state.ts
```diff
@@ -114,24 +135,34 @@ export class CollaborationState {
...
 		const cacheKey = this.formWriteLockCacheKey(workflowId);
-		const userId = await this.cache.get<User['id']>(cacheKey);
-		return userId ?? null;
+		const lockData = await this.cache.get<string>(cacheKey);
+
+		if (!lockData) {
+			return null;
+		}
+
+		return jsonParse<{ clientId: string; userId: string }>(lockData);
```
MiloradFilipovic: Maybe not valid, but since we switched to a new format is there a chance that users might have an existing write lock in cache (just `userId` as string) and this will throw?

packages/cli/src/collaboration/collaboration.service.ts
```diff
@@ -196,10 +226,7 @@ export class CollaborationService {
 
 	async broadcastWorkflowUpdate(workflowId: Workflow['id'], updatedByUserId: User['id']) {
 		const collaborators = await this.state.getCollaborators(workflowId);
-		// Filter out the user who made the update
-		const userIds = collaborators
-			.map((user) => user.userId)
-			.filter((userId) => userId !== updatedByUserId);
+		const userIds = collaborators.map((user) => user.userId);
```
MiloradFilipovic, line 229: Is this intentional, I see we are filtering out the current tab writer so just want to check do we need to send a message to all users in the first place?

MiloradFilipovic (dismissed):
Works well and the code looks good to me. Left a few comments so check out if they make sense.

packages/cli/src/collaboration/collaboration.state.ts
dariacodes: Good catch! I think the lock will expire by the time the customer upgrades their instance, but it's still better to add a fallback

packages/cli/src/collaboration/collaboration.service.ts
dariacodes: This is fine, because in `handleWorkflowUpdate` we are not sending them if the current tab is the editor

MiloradFilipovic (approved):
Thanks 🙌

PR #25648, opened 2026-02-11
fix(core): Handle Date objects in DateTime.fromISO() during full workflow execution

### LINKED ISSUE #25267: dateTime.format() value will display on output field but when executed returns invalid DateTime
### Bug Description

Hello, the previous [ticket](https://github.com/n8n-io/n8n/issues/23215) was closed due to inactivity, but the issue is still there.

DateTime.fromISO() works when editing a node, but doesn't work when full workflow is executed

### To Reproduce

1. Paste my simple test workflow

`{ "nodes": [ { "parameters": {}, "type": "n8n-nodes-base.manualTrigger", "typeVersion": 1, "position": [ 0, 0 ], "id": "89517ad5-18cc-43a2-8d60-23a6663fb776", "name": "When clicking ‘Execute workflow’" }, { "parameters": { "operation": "get", "dataTableId": { "__rl": true, "value": "RTUmXcXt5SSkmZr6", "mode": "list", "cachedResultName": "test2", "cachedResultUrl": "/projects/Gf5SWUTiJ2Xzg8or/datatables/RTUmXcXt5SSkmZr6" }, "limit": 1 }, "type": "n8n-nodes-base.dataTable", "typeVersion": 1, "position": [ 208, 0 ], "id": "4606fdc5-27a7-4bee-899b-35d3576ec415", "name": "Get row(s)" }, { "parameters": { "assignments": { "assignments": [ { "id": "3f58c229-f02b-4c92-8b10-0c7ecbc2b7bb", "name": "timestamp", "value": "={{ DateTime.fromISO($json.updatedAt).toFormat('dd/MM/yyyy HH:mm:ss') }}", "type": "string" }, { "id": "8b656ed7-bed8-4b48-89a8-b8e4480f774f", "name": "tt", "value": "={{ $json.updatedAt.toDateTime().setZone('Europe/Moscow').toFormat('dd/MM/yyyy HH:mm:ss') }}", "type": "string" } ] }, "includeOtherFields": true, "options": {} }, "type": "n8n-nodes-base.set", "typeVersion": 3.4, "position": [ 416, 0 ], "id": "69cf11a6-ef24-40e8-b73b-fe27b1adc5a4", "name": "Add timestamp" } ], "connections": { "When clicking ‘Execute workflow’": { "main": [ [ { "node": "Get row(s)", "type": "main", "index": 0 } ] ] }, "Get row(s)": { "main": [ [ { "node": "Add timestamp", "type": "main", "index": 0 } ] ] }, "Add timestamp": { "main": [ [] ] } }, "pinData": {}, "meta": { "instanceId": "435a1ecb9e606b9753a6ec0496715a8d59ac35a388e07f71b8ed8bb128b2d349" } }`

2. In the **get rows** node choose any table you have
3. Run the workflow
4. Check the output of the **add timestamp** node - you'll see "timestamp" field has "Invalid DateTime" output

Image
 
5. Open editor of the **add timestamp** node
6. Click "Execute step" button - now you have "timestamp" field with a proper datetime

Image

### Expected behavior

I expect DateTime.fromISO() to work properly both in editor and when I execute the full workflow

### Debug Info

# Debug info

## core

- n8nVersion: 2.6.3
- nodeJsVersion: 22.22.0
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: bf26dc67-2fd6-475f-9ae6-cdd0af7ee804

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/144.0.0.0 safari/537.36
- isTouchDevice: false

## security

- secureCookie: false

Generated at: 2026-02-03T21:01:25.477Z

### Operating System

Debian 12

### n8n Version

2.6.3

### Node.js Version

22.22.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

RajeshKumar11:
## Summary
Fixes #25267

- During full workflow execution, node output data is passed **by reference** between nodes without JSON serialization. This means `Date` objects (e.g. from the DataTable node) remain as `Date` instances when accessed via expressions like `{{ DateTime.fromISO($json.dateField) }}`.
- `DateTime.fromISO()` expects a string argument. When given a `Date` object, it internally calls `String(date)`, producing a non-ISO format like `"Wed Dec 12 2025 ..."`, which causes `DateTime.fromISO()` to return an **invalid DateTime**.
- In contrast, editor "Test Step" execution serializes data through a JSON round-trip, converting `Date` objects to ISO strings — which is why the same expression works correctly in the editor but fails during full execution.

## Solution
Wraps `DateTime` with a Proxy that intercepts `fromISO()` calls and converts `Date` objects to ISO strings before forwarding to the real `fromISO()`. This follows the **existing Proxy wrapper pattern** already used in the expression sandbox for `Error` (`createSafeError`) and `Object` (`createSafeObject`).

### Why this approach?
- **Minimal and targeted**: Only intercepts `fromISO()`, all other `DateTime` methods pass through unchanged
- **Consistent with codebase patterns**: Uses the same Proxy pattern as `createSafeError()` and `createSafeObject()`
- **No data mutation**: Does not modify the actual workflow data flowing between nodes
- **Handles the root cause**: Fixes the type mismatch at the point where users call `DateTime.fromISO()` in expressions

## Test plan
- [x] Added 6 tests covering:
  - `DateTime.fromISO()` with valid ISO string (existing behavior preserved)
  - `DateTime.fromISO()` with `Date` object (the bug fix)
  - `DateTime.fromISO()` with `Date` object and options parameter
  - Invalid input still returns invalid DateTime
  - Other `DateTime` static methods (`now()`, `fromMillis()`, `local()`) still work
  - ISO string and `Date` object produce identical results for the same timestamp
- [x] All 1912 existing tests pass
- [x] Pre-commit hooks (biome) pass


Joffcom:
Hi @RajeshKumar11 — thanks for the contribution! Before this can be reviewed, could you update the PR description to match our current [template](../blob/master/.github/pull_request_template.md)? Two sections are missing:

1. `## Related Linear tickets, Github issues, and Community forum posts` — you have `Fixes #25267` in the Summary, please move it under this dedicated heading.
2. `## Review / Merge checklist` — please paste the checklist from the template so reviewers can track readiness (your "Test plan" is great supplemental context but doesn't replace the merge checklist).

Once those are in we'll take another look. Thanks!

Joffcom:
Hey @RajeshKumar11, thanks a lot for digging into this and putting together a fix with tests, it is genuinely appreciated.

We looked into the underlying issue (#25267) and it turns out the root cause has already been resolved at the source: upgrading the Data Table node to v1.1 makes it emit properly serialised dates, so `DateTime.fromISO()` receives an ISO string as expected. The reporter has confirmed the upgrade fixed it for them.

Since the fix lives in the node itself, we would prefer not to also add an expression-layer proxy around `DateTime` (it wraps a very hot path for a case that no longer occurs). So we are going to close this one out.

Please do not let this put you off contributing, we would love to see more PRs from you. Thanks again!

PR #25658, opened 2026-02-11
feat(editor): In setup panel only mark credentials as completed if they are tested

RicardoE105:
## Summary

Previously, the setup panel marked credential cards as complete as soon as a credential was selected, without verifying that the credential actually works. This could mislead users into thinking their workflow is properly configured when credentials are invalid.

  Now, credential cards in the setup panel only show the checkmark after a background credential test
  succeeds:
  - When a credential is selected from the dropdown, a background test is triggered automatically
  - When a credential is saved/tested in the credential modal, the pending state is cleared on success
  - On page refresh, all pre-existing credentials are re-tested in the background
  - Failed or in-flight tests keep the card in an incomplete state

https://linear.app/n8n/issue/ADO-4817/only-mark-step-as-completed-if-credentials-work

  ## Review / Merge checklist

  ([conventions](../blob/master/.github/pull_request_title_conventions.md))
  - [x] Tests included.


MiloradFilipovic (changes requested):
I feel like this adds some unnecessary redundancy to get the job done:
- When a credential is set up in the `CredentialEdit` modal, it is being tested twice, both in the modal and in the composable in the background
- We seem to be tracking tested credentials state both in the `setupPanel.store` and `useWorkflowSetupPanel`

I would suggest a simpler approach:
- Instead of using `setupPanel.store` as a bridge between the modal and the setup panel, use the `credentials.store` to keep track of tested credentials and their test results
- This way we remove `credentialsPendingTest` completely and all of the new code to that store can be deleted
- When a credential is tested from UI, we update it's test status in the `credentials.store`
- `useWorkflowSetupState` composable listens to this store and marks cards as completed/uncompleted as credential status changes - doesn't re-test anything
- Keep the initial credentials check in the composable so we check them when the components load

We can also think about adding some loading state for the initial check, I think it's a bit unexpected for card statuses to change by themsleves

MiloradFilipovic (commented):
Code looks good, thanks for addressing the feedback 🙌
I still have an issue with oAuth credentials. Here, the setup was successful but nothing is checked in the setup panel:

https://github.com/user-attachments/assets/e9adac8f-8e55-46f5-969d-d043b99e73b8

packages/frontend/editor-ui/src/features/setupPanel/composables/useWorkflowSetupState.ts
```diff
@@ -196,6 +289,25 @@ export const useWorkflowSetupState = (nodes?: Ref<INodeUi[]>) => {
 		nodeHelpers.updateNodeCredentialIssuesByName(nodeName);
 	};
 
+	/**
+	 * When a credential is deleted, unset it from ALL nodes that reference it.
+	 * This is the symmetric counterpart to setCredential's auto-assignment.
+	 */
+	listenForCredentialChanges({
```
MiloradFilipovic, line 296: I think we should also unsubscribe from store once the composable is diposed:
```ts
import { computed, onScopeDispose, watch, type Ref } from 'vue';

const stopListeningForCredentialChanges = listenForCredentialChanges({ ... });

onScopeDispose(stopListeningForCredentialChanges);
```

MiloradFilipovic (approved):
Looks good 👌
Left a minor comment to prevent memory leaks.

PR #25662, opened 2026-02-12
fix(core): Workflow indexing tolerates missing workflowId on workflowCall

mfsiega:
## Summary

Tolerate a workflowCall dependency with empty parameteres.

Today, we produce an error if we can't find any kind of `workflowId` for the subworkflow call.

However, this case can arise when the node is first added to the canvas but hasn't been edited - the `parameters` look like `{ options: {} }`. When we detect this case we can safely ignore it. It's likely to happen more frequently now due to autosave.

NOTE: we might want to consider the intended behaviour for workflows with errors, and whether we want to index them at all (since they can't be executed). But for now, this will quiet the noisy errors in this particular common case.

https://linear.app/n8n/issue/CAT-2309

## Review / Merge checklist

- [ ] Tests included. 


tomi (approved):
🚀

PR #25663, opened 2026-02-12
fix(ai-builder): Handle InputTarget in WorkflowBuilder.to() for merge node connections

mutdmour:
## Summary

- Fix `WorkflowBuilder.to()` to handle `InputTarget` objects (e.g., `mergeNode.input(0)`), which previously fell through to the `NodeInstance` cast and produced broken connections with `node: undefined`
- Add `InputTarget` overload to `WorkflowBuilder` interface for type safety
- Add test covering the `.add(source).to(merge.input(n))` pattern

Also
- Fix WorkflowBuilder.add() and .to() to use the actual map key returned by addNodeWithSubnodes() instead of node.name, which is wrong when a node is auto-renamed due to duplicate names
- Add InputTarget handling to workflow-level to() method, enabling patterns like .add(process2).to(mergeNode.input(1))
Merge existing/new node paths in to() into a single idempotent path using addNodeWithSubnodes()

## Related Linear ticket

https://linear.app/n8n/issue/AI-2030

## Review / Merge checklist

- [x] Tests pass (4234/4234)
- [x] Typecheck clean
- [x] Pre-commit hooks pass

## Test plan

- [x] Existing merge tests still pass (no regressions)
- [x] New test verifies `.add(trigger).to(mergeNode.input(0))` pattern creates correct connections with proper input indices
- [x] Full workflow-sdk suite passes (86 suites, 4234 tests)


packages/@n8n/workflow-sdk/src/workflow-builder.ts
```diff
@@ -270,6 +298,33 @@ class WorkflowBuilderImpl implements WorkflowBuilder {
 			return this.handleNodeChain(nodeOrComposite);
 		}
 
+		// Handle InputTarget (e.g., mergeNode.input(1))
+		if (isInputTarget(nodeOrComposite)) {
```
aalises, line 302: Up there there's the same `isInputTarget(nodeOrComposite)` check that returns. Will this one ever be reached?

mutdmour: good point. will address in updated PR

PR #25671, opened 2026-02-12
fix(Schedule Trigger Node): clear stale recurrence data when schedule changes

### LINKED ISSUE #23711: Schedule trigger permanently drops valid triggers after changing schedule
### Bug Description

Some workflows with schedule triggers don't execute when they're meant to.

After diving into the code, I figured out the cause of the issue is `recurrenceCheck()`:
https://github.com/n8n-io/n8n/blob/1199904b15964a35cebaa6814d56aba5a9ecf761/packages/nodes-base/nodes/Schedule/GenericFunctions.ts#L39-L84

Called from the schedule trigger:
https://github.com/n8n-io/n8n/blob/1199904b15964a35cebaa6814d56aba5a9ecf761/packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts#L440-L442

I narrowed it down to `recurrenceCheck()` from `Executing cron for workflow` being logged but not `Received trigger for workflow`.

The workflow in question is set to every 3 hours, but was previously set to every hour. I assume this was what lead to `16` being set in the workflow's static data in `recurrenceRules`. After editing the schedule, the workflow never triggered again. As editing the schedule doesn't clear `staticData.recurrenceRules` and `3 + 16` is never going to equal a multiple of 3 (the cron "every 3 hours" generates means `hour` will always be a multiple of 3), `recurrenceCheck` always returns `false`:
https://github.com/n8n-io/n8n/blob/1199904b15964a35cebaa6814d56aba5a9ecf761/packages/nodes-base/nodes/Schedule/GenericFunctions.ts#L54-L59

Before diving into the code I also tried duplicating the workflow which still didn't fix the issue, nor did changing database or moving to queue mode, and as static data isn't visible to end users, there's no way for an end user to get out of this state unless they know about this bug and rename to node to refresh static data or create a new workflow and copy the nodes.

### To Reproduce

1. Create a workflow with a schedule trigger set to every 1 hour
2. After an execution where the hour is not a multiple of 3, edit the schedule to every 3 hours
3. Notice the workflow is never triggered again

### Expected behavior

The workflow executes.

### Debug Info

-

### Operating System

Docker

### n8n Version

2.1.4

### Node.js Version

Docker

### Database

PostgreSQL

### Execution mode

queue

### Hosting

self hosted

pierreeurope:
## Summary

Fix the Schedule Trigger permanently dropping valid triggers after a schedule change.

## Problem

When a user changes the schedule interval (e.g., from every 1 hour to every 3 hours), the `recurrenceRules` in the workflow's static data retain the `lastExecution` value from the previous configuration. Since `recurrenceCheck` computes `(intervalSize + lastExecution) % period`, the stale value causes the check to permanently return `false` — the trigger never fires again.

**Reproduction:**
1. Create a workflow with a schedule trigger set to every 1 hour
2. Let it execute at an hour that isn't a multiple of 3 (e.g., 16:00)
3. Change the schedule to every 3 hours
4. The workflow never triggers again because `(3 + 16) % 24 = 19`, which never matches a multiple-of-3 hour

## Fix

Compute a JSON fingerprint of the schedule configuration and store it in static data alongside `recurrenceRules`. When the trigger starts, compare the current configuration hash with the stored one. If they differ (schedule was changed), clear `recurrenceRules` so `recurrenceCheck` treats the next invocation as a fresh start (`lastExecution === undefined → return true`).

## Changes

- **`ScheduleTrigger.node.ts`**: Added schedule hash comparison and stale data clearing
- **`GenericFunctions.test.ts`**: Added test verifying recurrenceCheck succeeds after rules are cleared

## Related

Fixes #23711

### DIFF
--- a/packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
+++ b/packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
@@ -430,7 +430,17 @@ export class ScheduleTrigger implements INodeType {
 		const timezone = this.getTimezone();
 		const staticData = this.getWorkflowStaticData('node') as {
 			recurrenceRules: number[];
+			scheduleHash?: string;
 		};
+
+		// Compute a fingerprint of the current schedule configuration so we can
+		// detect when the user changes the schedule and clear stale recurrence
+		// data that would otherwise permanently prevent the trigger from firing.
+		const currentHash = JSON.stringify(intervals);
+		if (staticData.scheduleHash !== currentHash) {
+			staticData.recurrenceRules = [];
+			staticData.scheduleHash = currentHash;
+		}
 		if (!staticData.recurrenceRules) {
 			staticData.recurrenceRules = [];
 		}
--- a/packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
+++ b/packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
@@ -201,6 +201,22 @@ describe('recurrenceCheck', () => {
 		const result2 = recurrenceCheck(recurrence, recurrenceRules, 'UTC');
 		expect(result2).toBe(false);
 	});
+
+	it('should trigger on first call when recurrenceRules are cleared (schedule change)', () => {
+		// Simulate stale recurrence data from a previous schedule configuration
+		// being cleared when the schedule is changed — the first call after
+		// clearing should always succeed (lastExecution === undefined).
+		const recurrence: IRecurrenceRule = {
+			activated: true,
+			index: 0,
+			intervalSize: 3,
+			typeInterval: 'hours',
+		};
+		// Fresh recurrence rules (as if cleared after schedule change)
+		const recurrenceRules: number[] = [];
+		const result = recurrenceCheck(recurrence, recurrenceRules, 'UTC');
+		expect(result).toBe(true);
+	});
 });
 
 describe('intervalToRecurrence', () => {


packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -430,7 +430,17 @@ export class ScheduleTrigger implements INodeType {
...
+		// detect when the user changes the schedule and clear stale recurrence
+		// data that would otherwise permanently prevent the trigger from firing.
+		const currentHash = JSON.stringify(intervals);
+		if (staticData.scheduleHash !== currentHash) {
+			staticData.recurrenceRules = [];
+			staticData.scheduleHash = currentHash;
+		}
 		if (!staticData.recurrenceRules) {
 			staticData.recurrenceRules = [];
 		}
```
michael-radency, line 446: this block will be unreachable, we should remove it

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -430,7 +430,17 @@ export class ScheduleTrigger implements INodeType {
 		const timezone = this.getTimezone();
 		const staticData = this.getWorkflowStaticData('node') as {
 			recurrenceRules: number[];
+			scheduleHash?: string;
 		};
+
+		// Compute a fingerprint of the current schedule configuration so we can
+		// detect when the user changes the schedule and clear stale recurrence
+		// data that would otherwise permanently prevent the trigger from firing.
+		const currentHash = JSON.stringify(intervals);
```
michael-radency, line 439: naming is misleading - this is stringified object not hash

packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
```diff
@@ -201,6 +201,22 @@ describe('recurrenceCheck', () => {
...
+			activated: true,
+			index: 0,
+			intervalSize: 3,
+			typeInterval: 'hours',
+		};
+		// Fresh recurrence rules (as if cleared after schedule change)
+		const recurrenceRules: number[] = [];
+		const result = recurrenceCheck(recurrence, recurrenceRules, 'UTC');
+		expect(result).toBe(true);
+	});
```
michael-radency, line 219: this test does not cover this fix - it does cover `recurrenceCheck` util, fix meanwhile added to the node itself

michael-radency (changes requested):
Hello @pierreeurope , thanks for your PR, I did add some comments that needs to be addressed before we could accept it

PR #25673, opened 2026-02-12
fix(MySQL Node): Route parameter validation errors to error output

### LINKED ISSUE #25008: MySQL node sends parameter errors to the "Success" output rather than the "Error" output
### Bug Description

When using the MySQL Execute Query node with "On Error" set to "Continue (Using error output)", if you reference a parameter in the query but don't assign a query parameter to use, the error message is sent to the "Success" output instead of the "Error" output.

### To Reproduce

1. Create a MySQL node and open it
2. In the Parameters tab, Set the operation on the node to "Execute SQL"
3. Fill in the query as "$1"
4. In the Settings tab, set "On Error" to "Continue (using error output)"
5. Execute the workflow, the SQL node will fail but the output will be routed through the success end of the node

Image

[Bug Reproduction.json](https://github.com/user-attachments/files/24919788/Bug.Reproduction.json)

### Expected behavior

The error message should be routed through the error output as described by the setting name.

### Debug Info

# Debug info

## core

- n8nVersion: 2.2.5
- nodeJsVersion: 22.21.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: 64fa0aae-a37f-4913-945f-d4401b8c8b3b

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 168 hours
- maxCount: 0 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/144.0.0.0 safari/537.36 edg/144.0.0.0
- isTouchDevice: false

Generated at: 2026-01-28T19:19:38.023Z

### Operating System

Apline Linux 3.22

### n8n Version

2.2.5

### Node.js Version

22.21.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

pierreeurope:
## Summary

Fix parameter validation errors being sent to the success output instead of the error output when "On Error" is set to "Continue (using error output)".

## Problem

When a query references `$1` but no query parameters are provided, `prepareQueryAndReplacements` throws an error. This error occurs during query preparation in `executeQuery.execute()`, before the queries reach `runQueries()` — which is where `continueOnFail()` error handling lives. The uncaught error bubbles up through the router and appears on the success output.

## Fix

Wrap the per-item query preparation loop in a try/catch that checks `this.continueOnFail()`. When enabled, errors are returned as error items with proper `pairedItem` metadata for correct item mapping. When disabled, errors are re-thrown as before.

If all items fail during preparation, the function returns early with the error items without calling `runQueries`.

## Related

Fixes #25008

### DIFF
--- a/packages/nodes-base/nodes/MySql/v2/actions/database/executeQuery.operation.ts
+++ b/packages/nodes-base/nodes/MySql/v2/actions/database/executeQuery.operation.ts
@@ -47,10 +47,15 @@ export async function execute(
 	runQueries: QueryRunner,
 	nodeOptions: IDataObject,
 ): Promise<INodeExecutionData[]> {
-	let returnData: INodeExecutionData[] = [];
 	const items = replaceEmptyStringsByNulls(inputItems, nodeOptions.replaceEmptyStrings as boolean);
 
+	const errorItems: INodeExecutionData[] = [];
 	const queries: QueryWithValues[] = [];
+	// Track which original item index each query corresponds to.
+	// runQueries uses the query array index for pairedItem, so we need this
+	// to remap results back to the correct input items when some items are
+	// filtered out due to preparation errors.
+	const queryToItemIndex: number[] = [];
 
 	for (let i = 0; i < items.length; i++) {
 		try {
@@ -91,11 +96,12 @@ export async function execute(
 			}
 
 			queries.push(preparedQuery);
+			queryToItemIndex.push(i);
 		} catch (error) {
 			if (!this.continueOnFail()) {
-				throw error;
+				throw new NodeOperationError(this.getNode(), error as Error, { itemIndex: i });
 			}
-			returnData.push({
+			errorItems.push({
 				json: { message: (error as Error).message, error: { ...(error as Error) } },
 				pairedItem: { item: i },
 			});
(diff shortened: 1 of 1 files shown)


packages/nodes-base/nodes/MySql/v2/actions/database/executeQuery.operation.ts
```diff
@@ -53,43 +53,57 @@ export async function execute(
...
-			preparedQuery.values = parsedNumbers;
 		}
+	}
 
-		queries.push(preparedQuery);
+	if (queries.length === 0) {
+		return returnData;
 	}
 
 	returnData = await runQueries(queries);
```
AmitAnveri: This  assignment overwrites prior errors, if there are multiple input items and one of them succeeds without errors. The errors stored in catch block for the erroneous items is overwritten by this assignment.

Also in the same scenario with multiple items if there are some input items that are successful . They can get mapped to the wrong input item since some queries have been skipped and  runQueries maps input items based on new queries array index. The runQueries function assumes queries[i] came from input item i and this assumption is violated if some items are filtered out during preparation.

packages/nodes-base/nodes/MySql/v2/actions/database/executeQuery.operation.ts
```diff
@@ -53,43 +53,57 @@ export async function execute(
...
+				const parsedNumbers = preparedQuery.values.map((value) => {
+					return Number(value) ? Number(value) : value;
+				});
+				preparedQuery.values = parsedNumbers;
+			}
+
+			queries.push(preparedQuery);
+		} catch (error) {
+			if (!this.continueOnFail()) {
+				throw error;
```
AmitAnveri: I suggest using NodeOperationError. Using NodeOperationError will integrate better with the UI and gives item mapping context.

PR #25688, opened 2026-02-12
feat: Reload providers after connection and CRUD operations (no-changelog)

ireneea:
## Summary

This PR implements automatic provider reloading after connection changes and CRUD operations in the external secrets system. The changes ensure that provider connections stay in sync across the application when they are created, updated, or deleted.

**Key changes:**
- Added `syncProviderConnection` method to synchronize provider states after connection changes
- Enhanced `refreshProvider` logic in ExternalSecretsManager to handle provider reloading
- Made `broadcastReload` method public and invoke it after provider reloads
- Added comprehensive tests for both methods and CRUD operations

**How to test:**
1. Create/update/delete a secrets provider connection
2. Verify that the provider registry containing the secrets list refreshes automatically
3. Check that all instances receive the reload broadcast

To test that the cache is updated, you can use the completion endpoint `GET rest/secret-providers/completions/secrets/global`. The completion endpoint relies on the cached data to get the list of secrets.

https://linear.app/n8n/issue/LIGO-202

## Review / Merge checklist

- [ ] Tests included.


packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
```diff
@@ -128,6 +128,30 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
...
 	}
 
+	async syncProviderConnection(providerKey: string): Promise<void> {
+		await this.tearDownProviderConnection(providerKey);
+
+		const connection = await this.secretsProviderConnectionRepository.findOne({
+			where: { providerKey },
+		});
+
+		if (connection) {
```
konstantintieber, line 139: I wonder whether the case that the `connection` for the provided `providerKey` that was passed into a method called `syncProviderConnection` cannot be found should result in an error.

Meaning an early throw statement if connection is undefined.
Because if we couldn't find a connection there's also no reason to broadcastReload, right?

ireneea: This is specifically for the delete case, but maybe a comment would be useful to explain why there is this check

konstantintieber: Or we could in the delete case, not call `syncProviderConnection` but directly call `broadcastReload();`

konstantintieber: I see, broadcastReload is intended to be private. Maybe a dedicated public `onConnectionDeleted` then?

konstantintieber: So the `if (provider)` is also here because of this method being called when a connected provider is deleted?

ireneea: The `if (provider)` is useless is just to avoid typing issue because `this.providerRegistry.get` can return undefined

Do you think this looks better?

```typescript
const provider = this.providerRegistry.get(providerKey)!;
await this.secretsCache.refreshProvider(providerKey, provider);
```

ireneea: We could make `broadcastReload` that's not a big deal I think. The downside is that we are exposing more internal behaviour.

For the deletion we'll also need to expose the `tearDownProviderConnection` method

konstantintieber: > The if (provider) is useless is just to avoid typing issue because this.providerRegistry.get can return undefined

Keeping the if there is fine. Maybe with a comment saying that we expect this to always be defined if the connection exists.

ireneea: This would be the change in the `SecretsProvidersConnectionsService` if we don't want to use the common `syncProviderConnection` 


**Before**

```typescript
	async deleteConnection(providerKey: string): Promise<SecretsProviderConnection> {
		const connection = await this.repository.findOne({ where: { providerKey } });

		if (!connection) {
			throw new NotFoundError(`Connection with key "${providerKey}" not found`);
		}

		await this.projectAccessRepository.deleteByConnectionId(connection.id);
		await this.repository.remove(connection);

		await this.externalSecretsManager.syncProviderConnection(providerKey);

		return connection;
	}
```

**After**

```typescript
	async deleteConnection(providerKey: string): Promise<SecretsProviderConnection> {
		const connection = await this.repository.findOne({ where: { providerKey } });

		if (!connection) {
			throw new NotFoundError(`Connection with key "${providerKey}" not found`);
		}

		await this.projectAccessRepository.deleteByConnectionId(connection.id);
		await this.repository.remove(connection);

		await this.externalSecretsManager.tearDownProviderConnection(providerKey);
		this.externalSecretsManager.broadcastReload();

		return connection;
	}
```

Then we can change the `syncProviderConnection` to throw an error if `connection` doesn't exits.

```typescript
	async syncProviderConnection(providerKey: string): Promise<void> {
		await this.tearDownProviderConnection(providerKey);

		const connection = await this.secretsProviderConnectionRepository.findOne({
			where: { providerKey },
		});

		if (!connection) {
			throw new NotFoundError(`Connection with key "${providerKey}" not found`);
		}

		const settings = this.decryptSettings(connection.encryptedSettings);
		await this.setupProvider(
			connection.type,
			{ connected: connection.isEnabled, connectedAt: null, settings },
			providerKey,
		);

		if (provider) {
			await this.secretsCache.refreshProvider(providerKey, provider);
		}

		this.broadcastReload();
	}
```

How does this  seem to you

ireneea: Actually it's possible for the `provider` not to be set if the initialization fails

konstantintieber: The before is less code. I like that. So I'm leaning towards adding a comment next to `if(connection)` saying that this condition is here because the method is also called when deleting a connection.

ireneea: Sorted

packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts
```diff
@@ -128,6 +128,31 @@ export class ExternalSecretsManager implements IExternalSecretsManager {
...
+		const connection = await this.secretsProviderConnectionRepository.findOne({
+			where: { providerKey },
+		});
+
+		// Note: connection can be undefined if called after a delete operation
+		if (connection) {
+			const settings = this.decryptSettings(connection.encryptedSettings);
+			await this.setupProvider(
+				connection.type,
+				{ connected: connection.isEnabled, connectedAt: null, settings },
```
konstantintieber: With `isEnabled` being decommissioned in another ticket, should we already hardcode this here as `connected: true`?

ireneea: fixed 👍

konstantintieber (approved):

PR #25694, opened 2026-02-12
chore: Migrate codemirror packages to use vitest

Matsuuu:
## Summary

Migrating these packages to try and reduce flakiness caused by especially the codemirror-lang-sql's test suite in CI.

Part of https://linear.app/n8n/issue/CAT-1587/migrate-all-packages-to-use-vitest

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/codemirror-lang-html/package.json
+++ b/packages/@n8n/codemirror-lang-html/package.json
@@ -3,7 +3,7 @@
   "version": "1.2.0",
   "description": "HTML + n8n expression language support for CodeMirror 6",
   "scripts": {
-    "test": "cm-runtests",
+    "test": "vitest run",
     "build": "cm-buildhelper src/html.ts",
     "grammar:build": "lezer-generator src/grammar/html.grammar -o src/grammar/parser && rollup -c src/grammar/rollup.config.js",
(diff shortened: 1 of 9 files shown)


packages/@n8n/codemirror-lang-html/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.2.0",
   "description": "HTML + n8n expression language support for CodeMirror 6",
   "scripts": {
-    "test": "cm-runtests",
+    "test": "vitest run",
```
shortstacked, line 6: Unrelated to this PR since it was like this already. It turns out these probably don't run in the CI, since Turbo runs "pnpm test:unit" so we might need to update that in another PR!

shortstacked (approved):

packages/@n8n/codemirror-lang-html/package.json
Matsuuu: Yeah seems like the other packages have that but this doesn't. This also didn't rely on jest but instead the coedemirror custom test library so I guess it's from a different time. I'll put up a PR on fixing that later

PR #25701, opened 2026-02-12
feat: Add node creator methods

yehorkardash:
## Summary

This PR adds new methods that can be used to create chat model and memory nodes.
The methods are intended to be used as `class MyNode extends createChatMemoryNode(...) {}` so that the class is used. The reason for that is that a lot of eslint rules work with classes and using `const MyNode = createChatMemoryNode({...})` disables them.

Classes are casted to `UnextendableNodeType`

https://linear.app/n8n/issue/NODE-4431/add-node-creator-functions

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/ai-utilities/src/types/creators.ts
```diff
@@ -0,0 +1,53 @@
...
+export type GetChatMemoryFn = (
+	context: ISupplyDataFunctions,
+	itemIndex: number,
+) => Promise<ChatMemory>;
+export interface ChatMemoryNodeConfig extends AiNode {
+	getMemory: GetChatMemoryFn;
+	memoryOptions?: ChatMemoryOptions;
+}
+
+interface UndextendableNode {
```
DawidMyslak: Typo `UndextendableNode` vs `UnextendableNodeClass`.

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiSimple/LmChatOpenAiSimple.node.ts
```diff
@@ -17,8 +10,8 @@ export type ModelOptions = {
 	temperature?: number;
 };
 
-export class LmChatOpenAiSimple implements INodeType {
-	description: INodeTypeDescription = {
+export class LmChatOpenAiSimple extends createChatModelNode({
```
DawidMyslak: Ideally, we should just write `export const LmChatOpenAiSimple = createChatModelNode({ description, getModel });` 🥲.

We wouldn't need `UnextendableNodeType` then. I know you already checked, but which rules are affecting us? Can we not include those rules inside our SDK package?

One additional concern from my chat with Claude:
> One thing to consider: ESLint rules analyze the AST (the source code structure as the parser sees it - not runtime behavior). Since the consumer class body is empty `{}`, rules like `node-class-description-*` or `node-execute-block-*` can't see `description`/`supplyData`/`methods` anyway - those live inside the factory.

> If the main ESLint benefit is declaration-level rules (naming, filename), would `export const LmChatOpenAi = createChatModelNode(...)` work instead? No `extends` = no override risk = no need for `UnextendableNodeType`.

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiCustom/description.ts
```diff
@@ -0,0 +1,45 @@
+import { type INodeTypeDescription, NodeConnectionTypes } from 'n8n-workflow';
```
DawidMyslak, line 1: `decription.ts` file name typo

packages/@n8n/ai-utilities/src/creators/create-chat-model-node.ts
```diff
@@ -0,0 +1,19 @@
...
+import { supplyModel } from 'src/suppliers/supplyModel';
+import type { ChatModelNodeConfig, UnextendableNodeType } from 'src/types/creators';
+
+export const createChatModelNode = (chatModelNode: ChatModelNodeConfig) => {
+	const constructor = class ChatModelNode implements INodeType {
+		description = chatModelNode.description;
+		methods = chatModelNode.methods;
+		async supplyData(this: ISupplyDataFunctions, itemIndex: number) {
+			const model =
+				typeof chatModelNode.getModel === 'function'
```
DawidMyslak: Looks like type/runtime disagreement? The type says `getModel` is always a function.

DawidMyslak (changes requested):
Looks good, just blocking to meke sure the typos are resolved, 2 additional questions:
- Are `supplyMemory/supplyModel` still intended as public API or should they be internal now?
- Could we update the `README.md` as well to match newest API changes?

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiSimple/LmChatOpenAiSimple.node.ts
yehorkardash: Let me check how eslint rules can be changed

yehorkardash: I rechecked, some rules that I was concerned about are working, so I will revert it back to 
```typescript
export const LmChatOpenAiSimple = createChatModelNode({...})
```

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiCustom/description.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { openAiProperties } from '../properties';
+
```
michael-radency, line 4: should examples be in the main repository? I would add separate repository, something like `ai-community-node-samples` and include the link to `README`

yehorkardash: I'm using these examples for integration tests and manual testing, so I think they should be in monorepo. It's not the examples that will be in CLI examples.

packages/@n8n/ai-utilities/examples/nodes/LmChatOpenAiSimple/LmChatOpenAiSimple.node.ts
yehorkardash: Updated the implementation. Internally, the method returns a constructor, because n8n calls `new require('...').NodeName()` for every node it loads, so we have to return a class in the end.
```
export const createChatModelNode = (chatModelNode: ChatModelNodeConfig): unknown => {
	return class ChatModelNode implements INodeType {
		description = chatModelNode.description;
		methods = chatModelNode.methods;

		async supplyData(this: ISupplyDataFunctions, itemIndex: number) {
			const model = await chatModelNode.getModel(this, itemIndex);
			return supplyModel(this, model);
		}
	};
};
```
It's possible to still do `export class MyNode extends (createChatModelNode({...}) as any)`, unfortunately. If we want to prohibit that, we'd need to use Object.freeze() to protect methods like supplyData to be extended, but that's probably too much

PR #25707, opened 2026-02-12
feat: Update default Node.js version to v24

shortstacked:
## Summary

Drops Node 20 from the CI build list.
Updates the default Node.js version used across the project to v24.

This change involves updating the base Docker images, CI/CD workflows, and package engine constraints to align with the new Node.js version.

Removes references to Node.js v22 and adds Node.js v25 where applicable in CI/CD workflows.

## Review / Merge checklist

- [ ] Tests included. 


shortstacked:
@claude Review please

Matsuuu (dismissed):
LGTM. I've been running the system with node24 for the last few weeks and haven't had any problems so I'm assuming that this should be a safe transition. 

Should we signal this somehow in release notes or developer channels?

docker/images/n8n-base/Dockerfile
```diff
@@ -24,21 +24,11 @@ RUN apk add --no-cache busybox-binsh && \
         tzdata \
         ca-certificates \
         libc6-compat && \
-    # Install full-icu
-    npm install -g full-icu@1.5.0 && \
```
tomi, line 28: Why was this removed?

.github/workflows/docker-build-push.yml
```diff
@@ -7,7 +7,7 @@ name: 'Docker: Build and Push'
 
 env:
   NODE_OPTIONS: '--max-old-space-size=7168'
-  NODE_VERSION: '22.22.0'
+  NODE_VERSION: '24'
```
tomi: We should be explicit about version updates and not install any 24.x version. Otherwise we might bump the version while making some other completely unrelated change

docker/images/n8n-base/Dockerfile
shortstacked: It's been in my default since Node 18 I believe, so this was just legacy that needed to be removed.

.github/workflows/docker-build-push.yml
shortstacked: Yeah that makes sense. I've pinned to 24.13.1

tomi (approved):
🚀

PR #25714, opened 2026-02-12
fix(core): Disallow file uploads on Chat hub when chat trigger doesn't permit them

Cadiac:
## Summary

Chat trigger has optional file uploads option, which should control whether files can be sent to this trigger.

image

Chat hub hasn't been respecting this property, and this PR changes that. The accept hints for browser file picker are now based on mime types on the UI, and backend sends them to the UI, either based on the model (with base LLMs) or the workflow's chat trigger (workflow agents). Perhaps we'll make this configurable on personal agents later too.

Backend now also checks that the uploaded attachment matches this type.

https://linear.app/n8n/issue/CHA-125/bug-chat-hub-allows-file-attachment-when-disabled-in-workflow

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/chatHub/chat.utils.ts
```diff
@@ -452,6 +429,18 @@ export function createFakeAgent(
 	};
 }
 
+/**
+ * Enriches a MIME type accept string with the `.md` file extension.
+ * macOS file picker does not recognise `text/*` or `text/markdown` for
+ * Markdown files, so we add the explicit extension.
```
Cadiac, line 438: These are anyways just for browser UX, user can still pick any file regardless, but without this markdown files would stay grayed out :/ 
I'm expecting this isn't the only case, but the only major common case I found while testing it.
If this turns out to be problematic we might just want to remove the `accept` mechanism from the input and let backend error happen if they try something not accepted, but this should be good enough for now.

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -250,6 +250,81 @@ export class ChatHubWorkflowService {
...
+		}
+
+		return allowedFilesMimeTypes;
+	}
+
+	/**
+	 * Resolves the attachment upload policy for the given model.
+	 * Used to validate attachments before storage.
+	 */
+	async getAttachmentPolicy(
```
autologie, line 292: Hmm, chat-hub.models.service implements the same logic for returning model DTOs doesn't it?

packages/cli/src/modules/chat-hub/chat-hub.constants.ts
```diff
@@ -639,27 +653,71 @@ const MODEL_METADATA_REGISTRY: Partial<
...
+	const inputModalities =
+		modelOverride?.inputModalities ?? DEFAULT_INTERNAL_METADATA.inputModalities;
 
-	// Merge override with default metadata
 	return {
-		inputModalities: modelOverride.inputModalities ?? DEFAULT_MODEL_METADATA.inputModalities,
-		priority: modelOverride.priority,
+		inputModalities,
+		// LLM providers always allow file uploads
+		allowFileUploads: true,
```
autologie, line 712: This is because text files are treated as text message. I wonder if workflow agents can behave in the same manner.

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: Kind of? There's `extractModelFromWorkflow` on the model service, but it's using resolveAllowedMimeTypes from workflow service 🤔 

I refactored this a bit, making them both use the same common `chatHubWorkflowService.resolveWorkflowAttachmentPolicy` for this logic. Workflow service could also just call extractModelFromWorkflow after getting the workflow, but it does a bit of unnecessary logic so I feel like just providing a reusable `resolveWorkflowAttachmentPolicy` would be a reasonable way to go here.

packages/cli/src/modules/chat-hub/chat-hub.constants.ts
Cadiac: They probably could, but as they don't allow it through the regular chat trigger I feel like us allowing text files on workflow agents even if attachments aren't allowed would be wrong 🤔 

Our current way of not checking the field for workflow agents is wrong / different from how the chat trigger works outside chat hub...

autologie (dismissed):

autologie (approved):

PR #25716, opened 2026-02-12
fix(Send Email Node): Only set inline disposition for attachments referenced in HTML body

### LINKED ISSUE #24333: Send email node always mark images as inline attachments
### Bug Description

Send email node always marks images as inline attachments, regardless of whether they are referenced in the body of the message or not, and it is not possible to alter this behavior. Microsoft Exchange server leaves them in the message body, but MS Outlook or OWA clients do not display them.

### To Reproduce

Create binary data (f.e. jpeg image) with appropriate MimeType image/jpeg and name xxx.jpg.
Add them to attachment in Send email node.

### Expected behavior

Either explicitly allow the option to choose which ones should or should not be inline, or better yet, automatically scan the body of the message and mark only those that are referenced.

### Debug Info

# Debug info

## core

- n8nVersion: 2.4.0
- nodeJsVersion: 22.21.1
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- consumerId: 8ea44fd2-1f36-4b14-b6e1-f03ce5102bd7

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/143.0.0.0 safari/537.36 edg/143.0.0.0
- isTouchDevice: false

Generated at: 2026-01-14T15:27:30.137Z

### Operating System

Ubuntu 24.04

### n8n Version

2.4.0 

### Node.js Version

22.21.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

shtse8:
## Summary

Fixes #24333 — Send Email node always marks images as inline attachments.

## Problem

In the Send Email node v2, every attachment is unconditionally given a `cid` (Content-ID) property:

```ts
attachments.push({
    filename: binaryData.fileName || 'unknown',
    content: await this.helpers.getBinaryDataBuffer(itemIndex, propertyName),
    cid: propertyName, // ← always set
});
```

When nodemailer sees `cid` on an attachment, it sets `Content-Disposition: inline`, causing email clients like Microsoft Outlook and OWA to hide the attachment from the attachments list. The images appear embedded in the body (or not at all if not referenced).

**This is a v2 regression** — the v1 Send Email node never set `cid` on attachments.

## Fix

Auto-detect whether an attachment should be inline or regular:

```ts
// Only mark as inline (set cid) if the HTML body references this attachment
if (htmlBody.includes(`cid:${propertyName}`)) {
    attachment.cid = propertyName;
}
```

- If the HTML body references the attachment via `cid:<propertyName>` (e.g., ``), it is marked as inline
- Otherwise, it is sent as a regular attachment with proper `Content-Disposition: attachment`

## Backward Compatibility

✅ **Fully backward-compatible:**
- Users who embed images with `` continue to work exactly as before
- Users who just attach files now get correct attachment disposition
- No new options or UI changes needed — the fix is automatic

### DIFF
--- a/packages/nodes-base/nodes/EmailSend/test/v2/send.operation.test.ts
+++ b/packages/nodes-base/nodes/EmailSend/test/v2/send.operation.test.ts
@@ -58,7 +58,9 @@ describe('Test EmailSendV2, send operation', () => {
 				.mockReturnValueOnce('Test Subject')
 				.mockReturnValueOnce('html')
 				.mockReturnValueOnce({ attachments: 'file1, file2, file3', appendAttribution: false })
-				.mockReturnValueOnce('<p>Test HTML</p>');
+				.mockReturnValueOnce(
+					'<img src="cid:file1"><img src="cid:file2"><img src="cid:file3"><p>Test HTML</p>',
+				);
 
 			(mockExecuteFunctions.helpers.assertBinaryData as jest.Mock).mockImplementation(
 				(itemIndex: number, propertyName: string) => {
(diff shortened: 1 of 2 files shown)


michael-radency:
Hello @shtse8, thanks for your PR! We’re already adding a separate option for file attachments in https://github.com/n8n-io/n8n/pull/31071, as it’s more intuitive than relying on the `cid: `prefix and also allows passing `BinaryData` directly in both cases.

Therefore I’ll close this PR.

PR #25729, opened 2026-02-13
fix(core): Make health endpoint configurable to solve conflicts

dariacodes:
## Summary

It turns out our `/healthz` health endpoint can conflict with Google Cloud Platform's reserved paths (and potentially other cloud providers), so I added `N8N_ENDPOINT_HEALTH` env variable to make the health endpoint path configurable.                                                                                       
- Updated all servers (main, worker, task broker) to use the configured path                       
- Frontend now fetches the health endpoint path from backend settings and uses it for connectivity checks

Closes [ADO-4828](https://linear.app/n8n/issue/ADO-4828/community-issue-upgrading-to-v274-causes-all-workflow-editing-to-fail)

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/config/src/configs/scaling-mode.config.ts
```diff
@@ -3,9 +3,10 @@ import { Config, Env, Nested } from '../decorators';
 @Config
 class HealthConfig {
 	/**
-	 * Whether to enable the worker health check endpoints:
-	 * - `/healthz` (worker alive)
-	 * - `/healthz/readiness` (worker connected to migrated database and connected to Redis)
+	 * Whether to enable the worker health check endpoints.
+	 * The endpoint paths are configurable via N8N_ENDPOINT_HEALTH (default: 'health/live').
```
shortstacked: That's no longer the default so we should change it back

packages/cli/src/scaling/__tests__/worker-server.test.ts
```diff
@@ -56,6 +56,9 @@ describe('WorkerServer', () => {
 			credentials: {
 				overwrite: { endpoint: '' },
 			},
+			endpoints: {
```
shortstacked, line 59: We should add a test for the default when it's not passed to make sure it doesn't break

shortstacked (commented):
Minor test and doc comment

packages/cli/src/scaling/__tests__/worker-server.test.ts
dariacodes: This test is using a mocked config and just checks that you're going to use the endpoint defined in the config. To check whether you're going to use the default or the user defined value, we could add a test to `@n8/config`, but we don't do that for all our variables, so I think we should just trust it works.
Other tests are using the global config and if you change it in any test, it'll slip to other tests as well...

shortstacked (approved):

packages/@n8n/config/src/configs/endpoints.config.ts
```diff
@@ -135,4 +135,8 @@ export class EndpointsConfig {
 	/** Colon-delimited list of additional endpoints to not open the UI on. */
 	@Env('N8N_ADDITIONAL_NON_UI_ROUTES')
 	additionalNonUIRoutes: string = '';
+
+	/** Path for the health check endpoint. */
+	@Env('N8N_ENDPOINT_HEALTH')
+	health: string = 'healthz';
```
dlavrenuek, line 141: Looking at the code everywhere where the endpoint is used the path is calculated additionally as 
```
const path = `/${xxx.health}`;
```
only to add the leading `/`, so we could just make the default and configurable as `/healthz`

PR #25728, opened 2026-02-13
feat(editor): Update store logic for fetching global external secrets

sandra0503:
## Summary

Uses new endpoint to fetch global external secrets.
This is needed to split the autocompletion view between **project** external secrets and **global** external secrets.

Screenshot 2026-02-13 at 09 28 05

## Related Linear tickets, Github issues, and Community forum posts

Fixes [LIGO-219](https://linear.app/n8n/issue/LIGO-219/febe-autocompletion-bug-project-scoped-vaults-are-listed-under-global)

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [x] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #25730, opened 2026-02-13
feat(core): Old external secrets endpoint keeps returning only known provider keys (no-changelog)

konstantintieber:
## Summary

While there is a new endpoint to fetch global secrets with our new feature flag: https://github.com/n8n-io/n8n/pull/25728

I think we'd do ourselves a favor to ensure that the old endpoint doesn't return any providers that it didn't used to return in the past.

This PR limits the providers that can be returned from the `GET /rest/external-secrets/secrets` to the 5 hardcoded ones that were supported in the past.

https://linear.app/n8n/issue/LIGO-219/febe-autocompletion-bug-project-scoped-vaults-are-listed-under-global

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/secrets-cache.service.ts
+++ b/packages/cli/src/modules/external-secrets.ee/secrets-cache.service.ts
@@ -87,4 +87,30 @@ export class ExternalSecretsSecretsCache {
 
 		return result;
 	}
+
+	/**
+	 * Before we implemented project-scoped external secrets,
+	 * there was a hardcoded list of 5 supported provider names.
+	 * This method is only intended for the old GET /rest/external-secrets/secrets
+	 * to be backwards-compatible, by not including project-scoped secrets.
+	 */
+	getSecretsForVaultNamesBeforeProjectScopedRelease(): Record<string, string[]> {
+		const providers = this.registry.getAll();
+		const result: Record<string, string[]> = {};
+		const setOfPreviouslySupportedProviderNames = new Set([
+			'vault',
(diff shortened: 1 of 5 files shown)


packages/cli/src/modules/external-secrets.ee/secrets-cache.service.ts
```diff
@@ -87,4 +87,30 @@ export class ExternalSecretsSecretsCache {
...
 		return result;
 	}
+
+	/**
+	 * Before we implemented project-scoped external secrets,
+	 * there was a hardcoded list of 5 supported provider names.
+	 * This method is only intended for the old GET /rest/external-secrets/secrets
+	 * to be backwards-compatible, by not including project-scoped secrets.
+	 */
+	getSecretsForVaultNamesBeforeProjectScopedRelease(): Record<string, string[]> {
```
ireneea, line 97: What is the backward compatibility we're trying to maintain? If it's just the name that's fine. 
If we're trying to maintain the scope this might not be enough because connection with the old name could be re-scoped to a project

konstantintieber: From my testing today: if our project-scoped secrets feature flag is enabled, the `GET /rest/external-secrets/secrets` returns both global and project-scoped secrets.

It used to only return one of the 5 hardcoded secret providers.

konstantintieber: This might even be an access control concern, but didn't think this case through yet.

PR #25731, opened 2026-02-13
fix(OpenRouter Chat Model Node): Fix OpenRouter tool calls with empty arguments

OlegIvaniv:
## Summary

When using the **OpenRouter Chat Model** node with Anthropic models and an AI Agent that has a tool with no parameters, executing the tool fails with:

```
Failed to parse tool arguments from chat model response. Text: "[]".
SyntaxError: Unexpected end of JSON input
```

**Root cause:** Anthropic models accessed through OpenRouter's OpenAI-compatible API return empty/invalid `function.arguments` (`""`, `[]`, `null`) instead of `"{}"` for tools with no parameters. LangChain's `parseToolCall` does `JSON.parse("")` which throws, and the tool call ends up in `invalid_tool_calls` instead of `tool_calls`, breaking the agent output parser.

**Fix:** A `fetch` wrapper in the OpenRouter node intercepts API responses and normalizes invalid tool call arguments before LangChain sees them. This approach was chosen because:
- Subclassing `ChatOpenAI` doesn't work — `ChatOpenAI.withConfig()` (called by `bindTools`) hardcodes `new ChatOpenAI(this.fields)`, discarding any subclass
- The wrapper lives in the `configuration` object which is part of the original fields, so it survives all LangChain internal cloning

Normalization rules:
| Input | Output | Reason |
|---|---|---|
| `""`, `"  "`, `null` | `"{}"` | Empty — default to empty object |
| `[]`, `[1,2]` | `"{}"` | Arrays are not valid tool args |
| `{location: "NYC"}` | `'{"location":"NYC"}'` | Plain object — stringify to preserve data |
| `'{"city":"NYC"}'` | `'{"city":"NYC"}'` | Valid JSON string — unchanged |

https://linear.app/n8n/issue/AI-2029

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/llms/LmChatOpenRouter/LmChatOpenRouter.node.test.ts
+++ b/packages/@n8n/nodes-langchain/nodes/llms/LmChatOpenRouter/LmChatOpenRouter.node.test.ts
@@ -1,0 +1,326 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+
+/* eslint-disable @typescript-eslint/unbound-method */
+import { ChatOpenAI } from '@langchain/openai';
+import { makeN8nLlmFailedAttemptHandler, N8nLlmTracing, getProxyAgent } from '@n8n/ai-utilities';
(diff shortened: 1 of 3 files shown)


burivuhster (approved):
🚀

PR #25737, opened 2026-02-13
chore(core): Decouple workflow execution from edit permission

afitzek:
## Summary

Refactor the manual workflow execution flow to decouple execution from edit permissions:

- Backend: `POST /workflows/:workflowId/run` now always loads workflow data from the database, ignoring any `workflowData` sent in the request body. The `executeManually()` signature changes to accept `workflowData` as a separate parameter loaded by the controller. The `workflowData` field is removed from `ManualRunPayload` types.
- Frontend: `IStartRunData` replaces `workflowData: WorkflowData` with `workflowId: string`, stopping the frontend from sending the entire workflow object to the backend.
- Frontend: The canvas execute button is now visible in read-only mode when the user has `workflow:execute` scope, even without `workflow:update`. Templates, archived workflows, and demo mode are unaffected.

Example on user with `workflow:execute`, but without `workflow:update`.

image

closes https://linear.app/n8n/issue/IAM-281
closes https://linear.app/n8n/issue/IAM-282

## Review / Merge checklist

- [X] Tests included. 


packages/frontend/editor-ui/src/app/composables/useRunWorkflow.ts
```diff
@@ -319,7 +321,7 @@ export function useRunWorkflow(useRunWorkflowOpts: {
 				});
 
 			const startRunData: IStartRunData = {
-				workflowData,
+				workflowId: workflowData.id!,
```
guillaumejacquart, line 324: workflowData is coming from: 
`const workflowData = await workflowHelpers.getWorkflowDataToSave();`

Meaning that I think we used to send the current workflow data state (potentially changed) instead of the currently saved data in db when executing manuall.

Is it ok to now execute manually with the data saved in db. I think with auto save yes, but just wondering

guillaumejacquart: ah but there is autosave if state is dirty. Nevermind

packages/cli/src/workflows/workflow-execution.service.ts
```diff
@@ -173,22 +174,20 @@ export class WorkflowExecutionService {
 
 		// Case 3: Full execution from an unknown trigger.
 		if (isFullExecutionFromUnknownTrigger(payload)) {
-			const pinnedTrigger = payload.destinationNode
```
despairblue, line 176: nice one 👍🏾

despairblue (commented):
Logic seems solid. I didn't check the tests.

cstuncsik (approved):

PR #25750, opened 2026-02-13
fix: Bumping up package versions

shortstacked:
## Summary

Bumps up package versions for some packages and updates openvex.

## Review / Merge checklist

- [ ] Tests included. 


geemanjs (approved):

vex.openvex.json
```diff
@@ -3,7 +3,67 @@
...
+					"@id": "pkg:docker/n8nio/n8n",
+					"subcomponents": [
+						{
+							"@id": "pkg:apk/alpine/graphicsmagick@1.3.45-r0"
+						}
+					]
+				}
+			],
+			"status": "affected",
+			"action_statement": "WPG (WordPerfect Graphics) coder is compiled into Alpine's graphicsmagick package. However, WPG is an obsolete format from the 1980s with no legitimate use case in n8n workflows. Exploitation requires a workflow author to deliberately fetch and process a crafted WPG file via the Edit Image node."
```
Jubke, line 66: 🤔 Not sure this is reason enough to disregard it. I guess it's exactly the point that an n8n user with access could use this to read uninitialized memory on the n8n server. They control the images being processed, don't they?

But I guess that's back to not being able to update graphicsmagick, right?

shortstacked: Yeah, anything we do to interfere with it would be a breaking change unfortunately.

PR #25761, opened 2026-02-13
ci: Only tag current major version releases as prerelease

shortstacked:
## Summary

Legacy backport releases (e.g. `1.123.20`) were incorrectly tagged as **prerelease** on GitHub because the `prerelease` flag in `release-publish.yml` was hardcoded to `true`.

Now only `2.x` releases (current mainline) are marked as prerelease. Legacy `1.x` backport patches are created as full releases since they're already stable.

**Change:** `prerelease: true` → `prerelease: ${{ startsWith(needs.publish-to-npm.outputs.release, '2.') }}`

When n8n moves to 3.x, update `'2.'` to `'3.'`.

## Review / Merge checklist

- [x] Tests included.


Matsuuu (approved):
Yeah, I was wondering if this was by design or not. Good to see it fixed

PR #25766, opened 2026-02-13
feat(core): Add new params to claude task runner github action (no-changelog)

scdekov:
## Summary

Added parameters to the claude task runner action to make it more convenient to use from an n8n workflow

## Review / Merge checklist

- [ ] Tests included. 


shortstacked (approved):
Great changes but do we need the model part, it might be better to keep it consistent so we get consistent results across triage.
Also might need to test if we can drop the workflow pat token and just let us use the users somehow without mentioning it.

PR #25777, opened 2026-02-14
fix(GitLab & GitHub Triggers): Prevent duplicate webhooks on workflow re-publish

### LINKED ISSUE #25381: Webhook duplicates on each publish Gitlab
### Bug Description

n8n Version2.6.3

Since update about a week ago ( could have been only few versions, I update 2-3x a month) I am noticing that each publish will register the webhook in gitlab agian, with same URL and make the workflow be activated multiple times in the same second.

### To Reproduce

Use n8n Version2.6.3 and publish workflow with GL webhook multiple times

### Expected behavior

Only one webhook should be active per workflow / UUID

### Debug Info

# Debug info

## core

- n8nVersion: 2.6.3
- platform: npm
- nodeJsVersion: 20.19.2
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: 2bb6fb9a-e036-4857-b9fe-26efe698ef68

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 168 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/144.0.0.0 safari/537.36
- isTouchDevice: false

## security

- blockFileAccessToN8nFiles: false

Generated at: 2026-02-05T14:22:35.673Z

### Operating System

Ubuntu 22.04

### n8n Version

2.6.3

### Node.js Version

20.19.2

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

pranavsambyaI:
## Summary

Fixes #25381 - Webhook duplicates on each publish for GitLab and GitHub

## Bug Description

When a workflow with GitLab or GitHub triggers was published multiple times, a new webhook was registered each time with the same URL, causing duplicate workflow executions.

### GitLab Trigger
- Query existing webhooks via GET /projects/{id}/hooks before creating
- Check if webhook with matching URL already exists  
- Use existing webhook ID if found, otherwise create new

### GitHub Trigger  
- Query existing webhooks via GET /repos/{owner}/{repo}/hooks before creating
- Check if webhook with matching URL already exists
- Reuse existing webhook if events match
- Clear error if webhook exists with different events

## Testing

- [x] GitLab: Verified webhook lookup and reuse logic
- [x] GitHub: Verified webhook lookup and reuse logic
- [x] Both: Proper error handling for permission issues

## Related Issue

Fixes #25381

### DIFF
--- a/packages/nodes-base/nodes/Github/GithubTrigger.node.ts
+++ b/packages/nodes-base/nodes/Github/GithubTrigger.node.ts
@@ -509,6 +509,45 @@ export class GithubTrigger implements INodeType {
 				const endpoint = `/repos/${owner}/${repository}/hooks`;
 				const options = this.getNodeParameter('options') as { insecureSSL: boolean };
 
+				const webhookData = this.getWorkflowStaticData('node');
+
+				// Check for existing webhooks first to prevent duplicates
+				// See: https://github.com/n8n-io/n8n/issues/25381
+				let existingWebhooks;
+				try {
+					existingWebhooks = await githubApiRequest.call(this, 'GET', endpoint, {});
+				} catch (error) {
+					if (error.httpCode === '404') {
+						throw new NodeOperationError(
+							this.getNode(),
+							'Check that the repository exists and that you have permission to create the webhooks this node requires',
+							{ level: 'warning' },
+						);
+					}
+					throw error;
+				}
+
+				// Check if a webhook with the same URL already exists
+				for (const webhook of existingWebhooks as IDataObject[]) {
+					if ((webhook.config! as IDataObject).url! === webhookUrl) {
+						// Webhook with same URL found
+						if (JSON.stringify(webhook.events) === JSON.stringify(events)) {
+							// Same events, reuse existing webhook
+							webhookData.webhookId = webhook.id as string;
+							webhookData.webhookEvents = webhook.events as string[];
+							// Legacy webhook without secret - signature verification skipped
+							return true;
+						} else {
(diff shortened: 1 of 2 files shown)


pranavSambyalJSPY:
@Seblor Great question! Here's why the check is in `create` instead of `checkExists`:

**`checkExists` purpose:** Verifies if the webhook with the stored ID (in static data) still exists on GitLab/GitHub. It uses the webhook ID that was saved when the webhook was first created.

**The problem:** The webhook ID in static data can be lost/cleared in several scenarios:
- Workflow is exported/imported to a new instance
- n8n is restarted and static data is cleared  
- Workflow is duplicated
- Cache is cleared

In these cases, `checkExists` returns `false` (no webhookId in static data), so `create()` is called and creates a duplicate webhook.

**Why check in `create`:** We need to verify "is there ANY webhook with this URL?" regardless of whether we have its ID stored. This URL-based check prevents duplicates even when the stored webhook ID is missing.

**The flow now:**
1. `checkExists()` → Returns false (no stored webhookId)
2. `create()` → Checks GitLab/GitHub for ANY webhook with this URL
3. If found → Reuse it (no duplicate created)
4. If not found → Create new webhook

This ensures idempotency - publishing the same workflow multiple times never creates duplicates, even if static data is cleared between activations.

Does this make sense?

PR #25799, opened 2026-02-15
perf(core): Use streaming JSON parsing for large execution data

tomi:
## Summary

Replace synchronous `flatted.parse()` with `stream-json`-based async parsing for execution data to avoid blocking the event loop on large payloads.

### Problem

`flatted.parse()` calls `JSON.parse()` internally, which is synchronous and blocks the event loop for the entire duration. On large executions (18–75+ MB), this blocks the loop for **600ms to 2.8s**, starving all other requests and timers.

### Solution

- Add `parseFlatteAsync()` in `@n8n/backend-common` using `stream-json` to parse JSON in 64KB chunks, yielding to the event loop via `setImmediate` between chunks
- Add `parseFlatted()` wrapper that uses async parsing for payloads > 5MB, falling back to sync `flatted.parse()` for smaller payloads
- Update `ExecutionRepository.handleExecutionRunData()` to use the new async parser

### Libraries considered

| Library | Approach | Verdict |
|---|---|---|
| **stream-json** (selected) | SAX-style token stream with chunked `Readable` | 1.3–1.5x slower, p95 loop lag ~3ms |
| **yieldable-json** | Custom parser using ES6 generators, yields via `setImmediate` every ~512 ops | Not viable — 200x slower than sync at 50MB (94s vs 580ms) due to per-yield `setImmediate` overhead |
| **@sapientpro/json-stream** | Writable stream with async/await parser | Not viable — bug causes syntax errors on chunked writes >1MB, forcing single-write mode that blocks the loop. Also 3–4x slower |

#### 18 MB execution

| Metric | flatted (sync) | stream-json (async) |
|---|---|---|
| Parse time | ~585ms | ~880ms (1.5x) |
| Event loop p95 | 590ms (blocked) | **3.6ms** |
| Heap delta | 148 MB | **95 MB (36% less)** |
| GC total time | 67ms | 64ms |

#### 75 MB execution

| Metric | flatted (sync) | stream-json (async) |
|---|---|---|
| Parse time | ~2.8s | ~3.7s (1.3x) |
| Event loop p95 | 2.8s (blocked) | **2.9ms** |
| Heap delta | 596 MB | **422 MB (29% less)** |
| GC total time | 520ms | **370ms (29% less)** |

**Key takeaways:**
- **1.3–1.5x parse time overhead** — modest cost for async
- **p95 event loop lag drops from seconds to ~3ms** — the loop stays responsive throughout parsing
- **29–36% less peak heap** — incremental parsing avoids holding the raw JSON string and parsed intermediate array simultaneously
- **Less GC pressure** — more frequent but cheaper collections due to smaller working set

https://linear.app/n8n/issue/CAT-2352

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/backend-common/src/utils/flatted-async.ts
```diff
@@ -0,0 +1,174 @@
...
+function prepareFlatted(rawParsed: unknown[]): unknown[] {
+	return (applyPrimitivesDeep(rawParsed) as unknown[]).map(primitives);
+}
+
+/**
+ * Parse a flatted JSON string asynchronously using stream-json.
+ * Streams the string in 64KB chunks, yielding to the event loop between chunks
+ * via setImmediate, then resolves flatted references synchronously.
+ */
+export async function parseFlatteAsync(flattedString: string): Promise<unknown> {
```
mfsiega: Is this supposed to say `Flatted`?

packages/@n8n/backend-common/src/utils/flatted-async.ts
```diff
@@ -0,0 +1,174 @@
+/**
+ * Async flatted JSON parsing using stream-json.
+ *
+ * For large execution data, synchronous flatted.parse() blocks the event loop
```
mfsiega: I feel like this comment is maybe unnecessary, especially all the benchmark data?

packages/@n8n/backend-common/src/utils/flatted-async.ts
```diff
@@ -0,0 +1,174 @@
...
+import { ensureError } from 'n8n-workflow';
+import { Readable } from 'stream';
+import { parser } from 'stream-json';
+import Asm from 'stream-json/Assembler';
+
+// 64 KB chunks — selected via benchmarks as the sweet spot for event-loop
+// responsiveness vs scheduling overhead (p95 lag ~4 ms at 50 MB payload).
+const CHUNK_SIZE = 64 * 1024;
+
+// --- Flatted reference resolution (extracted from flatted@3.2.7) ---
```
mfsiega: Where does the extracted stuff end?

mfsiega (commented):
A few comments, generally 👏 

Nit: I might call this "streaming JSON parsing" rather than "async JSON parsing" (where I'd expect the approach to be more like what yieldable does)? Not sure if it's a real distinction.

packages/@n8n/backend-common/src/utils/flatted-async.ts
tomi: Yes 🤦 Fixed

packages/@n8n/backend-common/src/utils/flatted-async.ts
tomi: Removed it

packages/@n8n/backend-common/src/utils/flatted-async.ts
tomi: Added the end of the region

mfsiega (approved):
LGTM.

mfsiega (commented):
One comment I forgot to leave - if you want to be conservative, you could put this behind a feature flag and run it on Internal for a few days. But I don't think it's necessary.

PR #25810, opened 2026-02-16
feat(Agent): Add custom tracing metadata support in Agent Node

mjain:
## Summary

Adds support for user-defined tracing metadata in Agent-based nodes.
Users can now pass custom key–value pairs that are automatically attached to tracing spans during agent execution.

### Details

Introduces a Tracing Metadata parameter in Agent nodes

Metadata is merged into the tracing configuration when present

No behavior change when metadata is not provided (fully backward compatible)

### Benefit
While basic workflow execution data is already included in traces, there is currently no way to attach custom session or workflow-specific context (e.g. Slack thread ID, conversation ID, external request ID).

This makes it difficult to:

Correlate traces with external systems or sessions

Debug agent behavior across threaded or multi-session workflows

Enrich traces with business-level context

Custom tracing metadata allows users to attach this additional context directly to traces without requiring custom hooks or post-processing.

#### UI screenshot – Showing the new Tracing Metadata parameter

image
### LangSmith screenshot – Showing metadata attached to the trace
image

https://community.n8n.io/t/add-session-thread-tracking-metadata-to-openrouter-node-for-external-observability-broadcast/257260

## Review / Merge checklist

- [x] [Docs updated](https://github.com/n8n-io/n8n-docs) -https://github.com/n8n-io/n8n-docs/pull/4220.

- [x] Tests included. 


mjain:
> @mjain I see that Langsmith supports a lot of [OTel's Gen AI Attributes](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/), per [their documentation](https://docs.langchain.com/langsmith/trace-with-opentelemetry). Is there any need for / opportunity to follow that standard?
> 
> For example, rather than outputting the session id as an arbitrary name like `session_id`, should there be a standard way to output it as [`gen_ai.conversation.id`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-conversation-id) instead?


Thanks for the suggestion.

In this PR, We are not hardcoding any attribute names. it supports arbitrary key value metadata. If users prefer to follow OTel GenAI conventions (e.g., gen_ai.conversation.id), they can already do so using the metadata object.

Given that flexibility, I believe standardization would be better handled at a broader design or documentation level rather than enforced within this change.

packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/SqlAgent/execute.ts
```diff
@@ -58,7 +59,9 @@ export async function sqlAgentAgentExecute(
 				throw new NodeOperationError(this.getNode(), 'The ‘prompt’ parameter is empty.');
 			}
 
-			const options = this.getNodeParameter('options', i, {});
+			const options = this.getNodeParameter('options', i, {}) as {
+				tracingMetadata?: { values?: TracingMetadataEntry[] };
+			};
```
DeveloperTheExplorer:
```suggestion
			const options = this.getNodeParameter('options', i, {});
```

packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/SqlAgent/execute.ts
```diff
@@ -126,8 +129,13 @@ export async function sqlAgentAgentExecute(
 			}
 
 			let response: IDataObject;
+			const additionalMetadata = buildTracingMetadata(options.tracingMetadata?.values, this.logger);
```
DeveloperTheExplorer, line 132:
```suggestion
			const additionalMetadata = buildTracingMetadata((options.tracingMetadata as { values?: TracingMetadataEntry[] })?.values, this.logger);
```

packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/SqlAgent/execute.ts
DeveloperTheExplorer: This type enforcement causes errors in the type checking further down the function where `options.topK` and other properties of options is accessed (refer to lines 101-103) outside of the `tracingMetadata` type.

packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V3/helpers/tests/runAgent.test.ts
DeveloperTheExplorer, line 1: This test file contains excessive levels of mocking. Could you provide an explanation on why so much of the functions have been mocked and whether it's possible to test with the real functions?

DeveloperTheExplorer (changes requested):
While the work is greatly appreciated, the tests contain a lot of mocking. This to some extent defeats the purpose of the tests. Let me know if the mocks can be swapped out to actually test the functions/methods involved.

Also needs a minor type fix. Code change suggestion includes the change needed.

Thanks for the work!

DeveloperTheExplorer (approved):

PR #25814, opened 2026-02-16
feat(ai-builder): Implement workflow execution in evaluations (no-changelog)

schrothbn:
## Summary

Add an execution-based evaluator to the AI Workflow Builder evaluation harness. After the AI generates a workflow, the system:

1. Identifies service/API nodes that need mock data (nodes with credentials, HTTP Request, Webhook)
2. Generates realistic mock pin data for those nodes via a single LLM call, using Mermaid workflow diagrams and optional `__schema__` JSON Schemas for structural accuracy
3. Executes the workflow with the generated pin data — service nodes return pinned data while utility nodes (Set, If, Merge, etc.) run their real implementations
4. Reports success/failure as an `executionSuccess` metric

The workflow executor bypasses the DI container by loading compiled node implementations directly from `dist/known/nodes.json`, making it self-contained for the evaluation context.

https://linear.app/n8n/issue/AI-1977

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/evaluations/support/pin-data-generator.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/evaluations/support/pin-data-generator.ts
@@ -9,7 +9,6 @@
 import type { BaseChatModel } from '@langchain/core/language_models/chat_models';
 import { HumanMessage, SystemMessage } from '@langchain/core/messages';
 import { existsSync, readFileSync, readdirSync } from 'fs';
-import { join } from 'path';
 import {
 	jsonParse,
 	type IDataObject,
@@ -17,6 +16,7 @@ import {
 	type INodeTypeDescription,
 	type IPinData,
 } from 'n8n-workflow';
+import { join } from 'path';
 
 import type { SimpleWorkflow } from '../../src/types/workflow';
 import type { EvalLogger } from '../harness/logger';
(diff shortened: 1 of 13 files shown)


packages/@n8n/ai-workflow-builder.ee/evaluations/support/pin-data-generator.ts
```diff
@@ -0,0 +1,503 @@
...
+	operation?: string;
+	schema?: Record<string, unknown>;
+}
+
+// ---------------------------------------------------------------------------
+// Utility node deny-list
+// ---------------------------------------------------------------------------
+
+/** Node types that don't call external APIs and don't need pin data. */
+const UTILITY_NODE_TYPES = new Set([
```
OlegIvaniv, line 53: Instead of maintaining a list of these nodes couldn't we just say all nodes which don't have `credentials` in `INodeTypeDescription`?

OlegIvaniv (approved):

packages/@n8n/ai-workflow-builder.ee/evaluations/support/pin-data-generator.ts
schrothbn: I'll have a look separately to make sure that won't miss something if we do that (for when we also generate data for nodes without schema), good point.

PR #25817, opened 2026-02-16
fix(core): Validate GCP service account key on initialization

alielkhateeb:
## Summary

This PR adds early validation for the GCP service account key when initializing the GCP Secrets Manager provider.

Previously, if the stored service account key was invalid, empty, or malformed, we fell back to `{}` on JSON parse failure. This led to runtime uncaught errors from GCP due to missing `client_email` and `private_key`, which was blocking n8n instance startup.

With this change, the key is validated during initialization and a clear `UserError` is thrown if invalid, allowing other service providers to initialize normally.

## What’s Changed

- Validate that the service account key:
  - Is not empty or whitespace
  - Is valid JSON
  - Contains required fields: `client_email` and `private_key`
- Provide a clear error message referencing the Google Cloud Console service account JSON key file
- Prevent invalid stored GCP keys from blocking instance startup

## Why Only GCP?

GCP is currently the only provider requiring a JSON service account key at initialization. Other providers rely on tokens and do not require JSON parsing during startup.

## Tests

- Added 5 new test cases covering validation scenarios
- Updated existing tests to use a valid service account key helper

https://linear.app/n8n/issue/LIGO-201

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/providers/__tests__/gcp-secrets-manager.test.ts
+++ b/packages/cli/src/modules/external-secrets.ee/providers/__tests__/gcp-secrets-manager.test.ts
@@ -1,3 +1,4 @@
+import { UserError } from 'n8n-workflow';
 import { SecretManagerServiceClient } from '@google-cloud/secret-manager';
 import type { google } from '@google-cloud/secret-manager/build/protos/protos';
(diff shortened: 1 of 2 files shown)


MarcL (approved):
Nice! Thanks for fixing this.

PR #25819, opened 2026-02-16
feat(core): Add telemetry tracking for external secrets connections (no-changelog)

alielkhateeb:
## Summary

This PR adds telemetry tracking for external secrets connection lifecycle events (create, update, delete). Each event tracks the scope (global or project-level) to help understand how external secrets are being used across the platform.

**What this adds:**
- Telemetry tracking for `external-secrets-connection-created` events
- Telemetry tracking for `external-secrets-connection-updated` events (sharing/unsharing)
- Telemetry tracking for `external-secrets-connection-deleted` events
- Scope detection: `global` (no projects) vs `project` (associated with projects)
- Project IDs tracking to understand sharing patterns

**Properties tracked:**
- `user_id` - Who performed the action
- `vault_type` - Type of secrets provider (gcp, aws, vault, etc.)
- `scope` - `global` or `project` 
- `project_ids` - Number of projects the connection is shared with

**Tests:**
- 6 new test cases covering all three events with both global and project scopes
- All tests passing (79/79)

https://linear.app/n8n/issue/LIGO-203

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/events/relays/telemetry.event-relay.ts
+++ b/packages/cli/src/events/relays/telemetry.event-relay.ts
@@ -358,13 +358,11 @@ export class TelemetryEventRelay extends EventRelay {
 
 	private externalSecretsConnectionCreated({
 		userId,
-		providerKey,
 		vaultType,
 		projects,
 	}: RelayEventMap['external-secrets-connection-created']) {
 		this.telemetry.track('User created external secrets connection', {
 			user_id: userId,
-			provider_key: providerKey,
 			vault_type: vaultType,
 			scope: projects.length === 0 ? 'global' : 'project',
 			project_ids: projects.map((project) => project.id),
@@ -373,13 +371,11 @@ export class TelemetryEventRelay extends EventRelay {
 
 	private externalSecretsConnectionUpdated({
 		userId,
-		providerKey,
 		vaultType,
 		projects,
 	}: RelayEventMap['external-secrets-connection-updated']) {
 		this.telemetry.track('User updated external secrets connection', {
 			user_id: userId,
-			provider_key: providerKey,
 			vault_type: vaultType,
 			scope: projects.length === 0 ? 'global' : 'project',
 			project_ids: projects.map((project) => project.id),
@@ -388,13 +384,11 @@ export class TelemetryEventRelay extends EventRelay {
 
 	private externalSecretsConnectionDeleted({
 		userId,
-		providerKey,
 		vaultType,
 		projects,
 	}: RelayEventMap['external-secrets-connection-deleted']) {
 		this.telemetry.track('User deleted external secrets connection', {
 			user_id: userId,
-			provider_key: providerKey,
 			vault_type: vaultType,
 			scope: projects.length === 0 ? 'global' : 'project',
 			project_ids: projects.map((project) => project.id),
--- a/packages/cli/src/events/relays/telemetry.event-relay.ts
+++ b/packages/cli/src/events/relays/telemetry.event-relay.ts
@@ -164,7 +164,6 @@ export class TelemetryEventRelay extends EventRelay {
 		this.telemetry.track('Project settings updated', {
 			user_id: userId,
 			role,
-
 			members: members.map(({ userId: user_id, role }) => ({ user_id, role })),
 			project_id: projectId,
 		});
@@ -368,7 +367,7 @@ export class TelemetryEventRelay extends EventRelay {
 			provider_key: providerKey,
 			vault_type: vaultType,
 			scope: projects.length === 0 ? 'global' : 'project',
-			project_count: projects.length,
+			project_ids: projects.map((project) => project.id),
 		});
 	}
 
@@ -383,7 +382,7 @@ export class TelemetryEventRelay extends EventRelay {
 			provider_key: providerKey,
 			vault_type: vaultType,
 			scope: projects.length === 0 ? 'global' : 'project',
-			project_count: projects.length,
+			project_ids: projects.map((project) => project.id),
 		});
 	}
 
@@ -398,7 +397,7 @@ export class TelemetryEventRelay extends EventRelay {
 			provider_key: providerKey,
 			vault_type: vaultType,
 			scope: projects.length === 0 ? 'global' : 'project',
-			project_count: projects.length,
+			project_ids: projects.map((project) => project.id),
 		});
 	}
 
--- a/packages/cli/src/events/relays/telemetry.event-relay.ts
+++ b/packages/cli/src/events/relays/telemetry.event-relay.ts
@@ -89,6 +89,12 @@ export class TelemetryEventRelay extends EventRelay {
 			'external-secrets-provider-settings-saved': (event) =>
 				this.externalSecretsProviderSettingsSaved(event),
 			'external-secrets-provider-reloaded': (event) => this.externalSecretsProviderReloaded(event),
+			'external-secrets-connection-created': (event) =>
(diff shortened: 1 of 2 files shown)


packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -351,6 +357,51 @@ export class TelemetryEventRelay extends EventRelay {
...
+		providerKey,
+		vaultType,
+		projects,
+	}: RelayEventMap['external-secrets-connection-updated']) {
+		this.telemetry.track('User updated external secrets connection', {
+			user_id: userId,
+			provider_key: providerKey,
+			vault_type: vaultType,
+			scope: projects.length === 0 ? 'global' : 'project',
+			project_count: projects.length,
```
ireneea: This will always be `0` or `1` so not sure if it's even worth having.

There is a chance that the feature changes and we do allow multi-project 🤔. Not sure how likely that is though.

Would it be more valuable to have a `project` fields with project Ids maybe?

alielkhateeb: I relied on the schema having `projects` [here](https://github.com/n8n-io/n8n/blob/6e6ddb5d7737d3babc3d34b7111f318da9893d7b/packages/%40n8n/api-types/src/schemas/secrets-provider.schema.ts#L73) tbh, I thought a connection can already be added to multiple projects but maybe I misunderstood something

Having `projectIds` might be useful to have I agree, I'll add it

MarcL: Yes - I was thinking the project IDs would be useful. Also, these events aren't called for the global (existing) secrets management. Are we happy that 0 projects here means `global`?

alielkhateeb: After chatting with @ireneea we decided to go with `project_ids`, here are the changes 390eb4fa8211c75adf917e6607d91a21cf56ea8c

alielkhateeb: @MarcL As far as I understood from the code is that global and project based secret stores are created with the same endpoint, just either with or without projects selected, so to me this would be the indicator

I see an `isGlobal` property sent from the frontend but honestly I am skeptical about this prop as it's always true even if I do select a project in the share tab 😄

ireneea: Yes indeed this is the correct logic, we compute `global` based on the the fact that a connection has a project linked to it or not. 

But I supposed we could drop the `scope` property from the metric.

MarcL: Ah OK. Makes sense then. I had thought that it wasn't coming from the same event. Fine to approve if @ireneea is happy too.

alielkhateeb: @ireneea I thought about this too, but I think I'm held back by the acceptance criteria in the ticket `Every telemetry event will still contain a scope: 'global | 'project'.` 😂 

Technically the event is the same, either with `project_ids` or without, and we can filter on the length to determine the scope.

I can remove the scope too, if we agree here on that ❓

MarcL: Happy to remove the scope but worth checking with @fabian8n on Linear?

Also, do we want the project name in the telemetry event? I did add `ProjectSummary` which contains both ID and name for the log streaming events.

https://github.com/n8n-io/n8n/blob/1a8910e4bdb56d83d92a018f5a093e799e0cae13/packages/cli/src/events/maps/relay.event-map.ts#L26-L29

https://github.com/n8n-io/n8n/blob/1a8910e4bdb56d83d92a018f5a093e799e0cae13/packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts#L292-L301

ireneea: I'm also happy to remove if this is confirmed by product.

+1 for the project name being included

alielkhateeb: @MarcL @ireneea Agreed 💯  I asked [here](https://linear.app/n8n/issue/LIGO-203/be-add-telemetry-events#comment-8ac5da89)

packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -351,6 +356,51 @@ export class TelemetryEventRelay extends EventRelay {
...
+		providerKey,
+		vaultType,
+		projects,
+	}: RelayEventMap['external-secrets-connection-created']) {
+		this.telemetry.track('User created external secrets connection', {
+			user_id: userId,
+			provider_key: providerKey,
+			vault_type: vaultType,
+			scope: projects.length === 0 ? 'global' : 'project',
+			project_ids: projects.map((project) => project.id),
```
konstantintieber, line 368: I don't see what kind of insights we can gather relevant for telemtry data in our google bigquery by storing the auto-generated project IDs.
Did you have something specific in mind here?

packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -351,6 +356,51 @@ export class TelemetryEventRelay extends EventRelay {
...
 	}
 
+	private externalSecretsConnectionCreated({
+		userId,
+		providerKey,
+		vaultType,
+		projects,
+	}: RelayEventMap['external-secrets-connection-created']) {
+		this.telemetry.track('User created external secrets connection', {
+			user_id: userId,
```
konstantintieber, line 365: The `Telemetry` class used here automatically adds the `user_id` to each sent event. It doesn't have to be included here.

packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -351,6 +356,51 @@ export class TelemetryEventRelay extends EventRelay {
...
+			provider_key: providerKey,
```
konstantintieber: I would be vary to send the providerKey to our telemetry data. I can't come up with a telemetry question that requires us to know this, and I think it's good to only include properties where we know we need them, e.g. `vault_type` and `scope`

packages/cli/src/events/relays/telemetry.event-relay.ts
alielkhateeb: @konstantintieber for my own knowledge, where does it add it? 
I want to confirm before I go ahead and remove it from ALL the `track` calls in this file that send `userId` 😂 
Looked [here](https://github.com/n8n-io/n8n/blob/335dde7d1fcf032afb37a37600e1d4fa19e5c623/packages/cli/src/telemetry/index.ts#L232) but it seemed to me that it expects it as a property

packages/cli/src/events/relays/telemetry.event-relay.ts
alielkhateeb: I am more than happy to send only those, I think they're sufficient for now and maybe even `projects_count` can come later when/if we allow multiple projects, I have both versions ready to be pushed 😂 waiting for the discussion then I'll push whatever we agree on.

packages/cli/src/events/relays/telemetry.event-relay.ts
konstantintieber: My bad, it's actually not added automatically but destructured from the provided argument: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/telemetry/index.ts#L232

So your code is correct 👍 sorry for the confusion.

ireneea (approved):

PR #25822, opened 2026-02-16
feat: Return secrets count and secrets list in provider connection responses (no-changelog)

ireneea:
## Summary

Screenshot 2026-02-16 at 10 56 30

Enhances the external secrets provider connection API responses to include `secretsCount` and a `secrets` array (list of secret names):

- **Detail endpoint** (`GET /secret-providers/connections/:id`): Now returns both `secretsCount` and `secrets: [{ name }]`
- **List endpoint** (`GET /secret-providers/connections`): Now returns `secretsCount` (no secrets array, keeping it lightweight)
- **Project list endpoint** (`GET /secret-providers/projects/:id/connections`): Uses the lightweight list item DTO (with `secretsCount`, without settings or secrets array)
- Updated response types to include `secretsCount` in both DTOs and `secrets` in the full connection DTO
- Frontend: Removed unnecessary `?? 0` fallback for `secretsCount` since the backend now always provides it

## Testing

1. Connect to an external secrets providers with secrets
2. You should see the secrets count for that provider
3. Add new secrets to the external secret providers
4. Wait for the secret refresh / refresh the settings page and check that the count is updated

You can use `N8N_EXTERNAL_SECRETS_UPDATE_INTERVAL=30` to force the secret refresh every 30s  

https://linear.app/n8n/issue/LIGO-193

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -202,6 +202,7 @@ export class SecretsProvidersConnectionsService {
 			id: String(connection.id),
 			name: connection.providerKey,
 			type: connection.type as SecretsProviderType,
+			secretsCount: this.externalSecretsManager.getSecretNames(connection.providerKey).length,
```
alielkhateeb: Nit: `secretsCount: secretNames.length` having a `secretNames` const like below would make it more readable imo

konstantintieber (dismissed):

konstantintieber (approved):

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
ireneea: Fixed 👍

alielkhateeb (approved):

PR #25834, opened 2026-02-16
refactor(core): Remove TaskRunnersConfig.enabled since runners are always on

tomi:
## Summary

Task runners are always enabled in n8n, making the `enabled` property in `TaskRunnersConfig` redundant. This PR removes:

- The `enabled` property from the config class
- All conditional checks around task runner enablement
- The now-obsolete `TaskRunnersRule` breaking changes detector
- Dead code paths in Code/ToolCode nodes that handled disabled runners
- Test setup code that disabled runners

**Changes:** 19 files modified/deleted, 212 lines removed, simpler codebase with no conditional overhead.

https://linear.app/n8n/issue/CAT-2004

## Review / Merge checklist

- [x] Tests pass (nodes-base: 3218 tests passed)
- [x] No functional impact (refactor only)


packages/testing/playwright/services/workflow-api-helper.ts
```diff
@@ -289,6 +289,44 @@ export class WorkflowApiHelper {
 		return result;
 	}
 
+	/** Triggers a manual workflow execution via the REST API and waits for completion. */
+	async runManually(
```
despairblue: I think is not used anywhere? Will it be used in an upcoming PR?

packages/core/src/execution-engine/__tests__/workflow-execute.test.ts
```diff
@@ -537,77 +537,6 @@ describe('WorkflowExecute', () => {
 		});
 	});
 
-	//run tests on json files from specified directory, default 'workflows'
-	//workflows must have pinned data that would be used to test output after execution
-	describe('run test workflows', () => {
```
tomi, line 542: I replaced these with a simple e2e test that makes sure Code node runs in both modes fine. We have quite extensive unit test suite for task runners, so we don't necessarily need all these anymore

despairblue (dismissed):
added 2 commits, let me know if I messed something up

despairblue (approved):

Matsuuu (approved):

PR #25848, opened 2026-02-16
ci: Enable Ryuk cleanup for testcontainers

shortstacked:
## Summary
- Remove `TESTCONTAINERS_RYUK_DISABLED=true` from all CI workflows and npm scripts
- Ryuk handles container cleanup on process exit (including crashes/timeouts), providing more reliable cleanup than teardown alone

## Test plan
- [ ] E2E tests pass in CI (Ryuk starts and cleans up containers)
- [ ] Testcontainer-based integration tests (`test:postgres:tc`, `test:sqlite:migrations:tc`) still work


Matsuuu (commented):
So the reason for us disabling Ryuk in the first place was

> to avoid issues with Docker since it needs privileged access

Is this not an issue anymore or have we circled around it in some other way?

Matsuuu (approved):
Sounds good. Ship it

PR #25851, opened 2026-02-16
fix(editor): Add validation for secret JSON input fields

sandra0503:
## Summary

**Problem**: The GCP Secrets Manager service account key field was typed as string instead of json. Since JSON password fields render a plain text input (not the JSON editor), users could save invalid JSON without any feedback -- the save button remained enabled.

**Solution**:
Correct field type: Change the GCP serviceAccountKey parameter from type: 'string' to type: 'json', which accurately represents the expected input.

Add inline JSON validation: Introduce validateJsonPassword() in ParameterInput that validates input on JSON password fields and surfaces errors through the existing ParameterIssues UI.

Propagate validation state: Fix defineExpose in ParameterInput to expose displaysIssues as a reactive computed (was accidentally exposing a static boolean). Forward it through ParameterInputExpanded via useTemplateRef so parent components can read it.

Block save on validation errors: In useConnectionModal, collect displaysIssues from each parameter input ref and disable the save button when any field has validation issues.

## Screenshots

Screenshot 2026-02-16 at 17 44 44

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/composables/useConnectionModal.ee.ts
```diff
@@ -241,6 +252,9 @@ export function useConnectionModal(options: UseConnectionModalOptions) {
 		const hasPermission = isEditMode.value ? canUpdate.value : canCreate.value;
 		if (!hasPermission) return false;
 
+		// Check for validation errors (e.g., invalid JSON in password fields)
```
alielkhateeb: Nit: the comment doesn't add more info, I think the code is self-explanatory imho

packages/cli/src/modules/external-secrets.ee/providers/gcp-secrets-manager/gcp-secrets-manager.ts
```diff
@@ -21,7 +21,7 @@ export class GcpSecretsManager extends SecretsProvider {
 		{
 			displayName: 'Service Account Key',
 			name: 'serviceAccountKey',
-			type: 'string',
+			type: 'json',
```
alielkhateeb, line 24: This makes much more sense 👏

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/composables/useConnectionModal.ee.ts
```diff
@@ -49,6 +53,7 @@ export function useConnectionModal(options: UseConnectionModalOptions) {
 	const originalSettings = ref<Record<string, IUpdateInformation['value']>>({});
 	const isSaving = ref(false);
 	const didSave = ref(false);
+	const parameterInputRefs = ref<Record<string, ParameterInputRef | null>>({});
```
alielkhateeb: Few questions (Vue noob here):

- (Also a nit) Why the name `ParamaterInputRef`? I don't see `Ref` used in other ref names, which to me makes sense as it's kind of a repetition. Does `inputParameters` make sense? Or a
- What is this used for? Am I misunderstanding the code or is this only used to determine if we have a validation error or not?

Happy to huddle on this btw, I would personally benefit from it 😄

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/composables/useConnectionModal.ee.ts
sandra0503: True, will remove!

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/composables/useConnectionModal.ee.ts
sandra0503: Hi Ali, the naming might not have been very clear.

- I renamed the record to better reflect the purpose. It doesn't store Vue template refs, but rather snapshots of each parameter input's validation state (specifically the `displaysIssues` flag)
- Exactly, it's used to check whether there is a validation error or not

Let me know if we should go on a huddle! :)

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/composables/useConnectionModal.ee.ts
```diff
@@ -273,6 +281,14 @@ export function useConnectionModal(options: UseConnectionModalOptions) {
 		};
 	}
 
+	function setParameterValidationState(
+		propertyName: string,
+		el: Element | ComponentPublicInstance | null,
+	) {
+		if (!isComponentPublicInstance(el) || !('displaysIssues' in el)) return;
+		parameterValidationStates.value[propertyName] = !el.displaysIssues;
```
alielkhateeb, line 301: Is that supposed to be `!!el.displaysIssues` instead of `!el.displaysIssues`? At least it was `!!el.displaysIssues` in the previous code right?

sandra0503: I thought it's more intuitive if this contains whether the item is valid

alielkhateeb: Ah yes, got it!
And agreed, positive states / conditions are always better 👏

alielkhateeb (dismissed):
Nice. Tested and working as expected as well 👏 🚀 

Image

konstantintieber (dismissed):
Awesome

alielkhateeb (dismissed):
👏

konstantintieber (approved):

PR #25854, opened 2026-02-16
feat: Build pinecone quick connect integration flow

dlavrenuek:
## Summary

Implements the Quick Connect functionality for Pinecone

To test set the following env settings and start the n8n instance. The configuration will enable quick connect for the `Pinecone Vector Store` node and the `Pinecone Assistant` community node.

```
export N8N_QUICK_CONNECT_OPTIONS='[{"packageName": "@n8n/n8n-nodes-langchain", "credentialType": "pineconeApi", "serviceName":"Pinecone", "text": "Pinecone provides juicy benefits for n8n users", "quickConnectType":"pinecone", "config":{"integrationId":"n8n-dev"}},{"packageName": "@pinecone-database/n8n-nodes-pinecone-assistant", "credentialType": "pineconeAssistantApi", "serviceName":"Pinecone", "text": "Pinecone provides juicy benefits for n8n users", "quickConnectType":"pinecone", "config":{"integrationId":"n8n-dev"}}]'
export N8N_CROSS_ORIGIN_OPENER_POLICY='same-origin-allow-popups'
``` 

- https://linear.app/n8n/issue/NODE-4277/build-pinecone-integration-flow

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/quick-connect/quick-connect.config.ts
```diff
@@ -14,6 +14,7 @@ const quickConnectOptionSchema = z.object({
 	quickConnectType: z.string(),
 	serviceName: z.string(),
 	consentText: z.string().optional(),
+	config: z.record(z.string(), z.string()).optional(),
```
dlavrenuek: Using an arbitrary `Record<string, string>` type here since we don't know what possible options we will need for other providers

packages/cli/src/server.ts
```diff
@@ -393,6 +394,9 @@ export class Server extends AbstractServer {
 							preload: false,
 						}
 					: false,
+				crossOriginOpenerPolicy: {
+					policy: crossOriginOpenerPolicy,
+				},
```
dlavrenuek, line 399: This is required for the popup to allow sending messages (message with the created key) to n8n window.

dlavrenuek: Must be used in combination `N8N_CROSS_ORIGIN_OPENER_POLICY=same-origin-allow-popups` to work for pinecone flow

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
```diff
@@ -3,15 +3,37 @@ import { QUICK_CONNECT_EXPERIMENT } from '@/app/constants';
...
 import { computed } from 'vue';
 
 import type { ICredentialsResponse } from '../../credentials.types';
 import { useCredentialOAuth } from '../../composables/useCredentialOAuth';
+import { useCredentialsStore } from '../../credentials.store';
+import { useToast } from '@/app/composables/useToast';
+import { useI18n } from '@n8n/i18n';
+
+async function connectToPinecone(quickConnectOption: QuickConnectOption) {
+	const { ConnectPopup } = await import('@pinecone-database/connect');
```
dlavrenuek, line 16: The pinecone sdk is loaded asynchronously here, so it's not included/loaded into the normal bundle even if pinecone connect is not used

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
```diff
@@ -3,15 +3,37 @@ import { QUICK_CONNECT_EXPERIMENT } from '@/app/constants';
...
+import { useProjectsStore } from '@/features/collaboration/projects/projects.store';
 import { computed } from 'vue';
 
 import type { ICredentialsResponse } from '../../credentials.types';
 import { useCredentialOAuth } from '../../composables/useCredentialOAuth';
+import { useCredentialsStore } from '../../credentials.store';
+import { useToast } from '@/app/composables/useToast';
+import { useI18n } from '@n8n/i18n';
+
+async function connectToPinecone(quickConnectOption: QuickConnectOption) {
```
dlavrenuek: I did decide to implement the pinecone connect functionality here instead of a separate file, so it's not exported and can not be imported/used outside of the `useQuickConnect` hook

packages/cli/src/modules/quick-connect/quick-connect.config.ts
elsmr: Could we use a union type for `config` based on `quickConnectType`?

packages/cli/src/server.ts
elsmr: How will this work on cloud? Should this be set by default or does this open the door for security issues?

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
elsmr: Thanks for doing this :pray:

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
```diff
@@ -82,9 +116,36 @@ export function useQuickConnect() {
...
-			// TODO: Implement quick connect flows here
-			return null;
+		const quickConnectOption = getQuickConnectOption(credentialTypeName, nodeType);
+		if (quickConnectOption) {
+			const credentialType = credentialsStore.getCredentialTypeByName(credentialTypeName);
+			if (!credentialType) {
+				return null;
+			}
+
+			try {
```
elsmr, line 125: Should we warn users if their `crossOriginOpenerPolicy`will prevent this flow from working?

elsmr (commented):
Looks really good, tested it and it works as expected!

I'm only worried about the new env variable, does this mean that, by default, the quick connect flow will not work (until you set that env variable)?

packages/cli/src/modules/quick-connect/quick-connect.config.ts
dlavrenuek: Good idea, will add it 👍

packages/cli/src/server.ts
dlavrenuek: Let's say for quick connect the n8n instance (opener) opens a popup (popup)

Per default it is the old value `same-origin` (this is the value currently used), which only allows the popup to communicate with opener if the popup is on the same domain. `same-origin` also allows opening the n8n instance as a popup from a different window ONLY if the other window uses the same domain as the n8n instance. Since the pinecone integration flow opens the pinecone popup on pinecone domain it can not communicate with the opener n8n instance and can not sent the token to it.

The setting `same-origin-allow-popups` allows the popup to communicate with the opener but still prevents any other domain to open n8n in a popup. What the popup can access fro the opener is very limited, practically it is only the ability to send messages and to modify the url. The latter one could indeed be used for phishing but would require either pinecone to implement an attack on their servers (which we don't expect) or a supply chain attack where bad actors would hack pinecone and implement url redirect on their servers (we also don't expect that).

PS: as you can see in the code, you can not completely disable this header, the only available configuration options are `same-origin` (default) or `same-origin-allow-popups`

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
dlavrenuek: Quick connect is meant for cloud in first place, there we need to configure promotions on our side anyway (also for backend flows with secrets) and will also need to set the `crossOriginOpenerPolicy`. It is only required (currently) for pinecone. Since oauth uses a redirect url, the partner server always redirects to the redirect uri on n8n instance domain, which should have access to opener window from a popup. The backend flow does not open any popups, so the popup/opener security policy does not affect it anyway

packages/cli/src/modules/quick-connect/quick-connect.config.ts
dlavrenuek: Done. The backend option config type now does not make too much sense, but am refactoring it as part of firecrawl implementation in https://github.com/n8n-io/n8n/pull/25901 anyway.

elsmr (approved):

PR #25855, opened 2026-02-16
fix(core): Forward activationMode in multi-main webhook/trigger setup

tomi:
## Summary

In multi-main deployments, when the n8n Trigger node's "Active Workflow Updated" event is configured, it fails to fire because follower instances were always using 'activate' mode instead of the actual activation mode (e.g., 'update'). This fix forwards the `activationMode` through the pub/sub command so followers can correctly distinguish between initial activation and workflow updates.

Changes include:
- Forward `activationMode` in the `add-webhooks-triggers-and-pollers` pub/sub command payload
- Update handler to use the forwarded `activationMode` instead of hardcoded 'activate'
- Add workflow update helper method to the Playwright API service
- Add E2E tests to verify multi-main trigger behavior for both 'activate' and 'update' modes

https://linear.app/n8n/issue/CAT-2310

## Review / Merge checklist

- [x] Tests included

### DIFF
--- a/.github/workflows/test-e2e-reusable.yml
+++ b/.github/workflows/test-e2e-reusable.yml
@@ -138,7 +138,7 @@ jobs:
           # docker-build: build app + docker image locally
           # docker-pull: no build needed, image is pre-built
           # local: build app for local server
-          build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || (inputs.test-mode == 'local' && 'pnpm build' || '') }}
+          build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || (inputs.test-mode == 'local' && 'pnpm build' || 'pnpm build --filter n8n-playwright') }}
           enable-docker-cache: ${{ inputs.test-mode == 'docker-build' }}
         env:
           INCLUDE_TEST_CONTROLLER: ${{ inputs.test-mode == 'docker-build' && 'true' || '' }}
--- a/.github/workflows/test-e2e-reusable.yml
+++ b/.github/workflows/test-e2e-reusable.yml
@@ -138,7 +138,7 @@ jobs:
           # docker-build: build app + docker image locally
           # docker-pull: no build needed, image is pre-built
           # local: build app for local server
-          build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || (inputs.test-mode == 'local' && 'pnpm build' || 'pnpm build --filter n8n-playwright') }}
+          build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || 'pnpm build' }}
           enable-docker-cache: ${{ inputs.test-mode == 'docker-build' }}
         env:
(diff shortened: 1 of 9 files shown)


.github/workflows/test-e2e-reusable.yml
```diff
@@ -138,7 +138,7 @@ jobs:
           # docker-build: build app + docker image locally
           # docker-pull: no build needed, image is pre-built
           # local: build app for local server
-          build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || (inputs.test-mode == 'local' && 'pnpm build' || '') }}
+          build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || (inputs.test-mode == 'local' && 'pnpm build' || 'pnpm build --filter n8n-playwright') }}
```
shortstacked: You can drop the passed argument. By default the 'setup-nodejs' action calls pnpm build which would be cached from previous results. 
`build-command: ${{ inputs.test-mode == 'docker-build' && 'pnpm build:docker' || 'pnpm build' }}`

packages/testing/playwright/tests/e2e/nodes/n8n-trigger.spec.ts
```diff
@@ -0,0 +1,73 @@
+import { workflow, trigger, node } from '@n8n/workflow-sdk';
+import type { INode, IWorkflowBase } from 'n8n-workflow';
+import { nanoid } from 'nanoid';
+
+import { test, expect } from '../../../fixtures/base';
+
+type TriggerEventType = 'activate' | 'update';
+
+const makeN8nTriggerWorkflow = (events: TriggerEventType[]) => {
```
shortstacked, line 9: Nice to see this finally get used!

packages/testing/playwright/tests/e2e/nodes/n8n-trigger.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+test.describe(
+	'n8n Trigger node',
+	{
+		annotation: [{ type: 'owner', description: 'Catalysts' }],
+	},
+	() => {
+		test('should fire "activate" event when workflow is published', async ({ api }) => {
+			const wf = makeN8nTriggerWorkflow(['activate']);
+			const { workflowId, createdWorkflow } = await api.workflows.createWorkflowFromDefinition(
```
shortstacked, line 38: I can look at revisiting the workflow helper in a future PR to avoid having to cast.

shortstacked (dismissed):
Looks good. Minor comments on my side but not blockers by any means!

.github/workflows/test-e2e-reusable.yml
tomi: 👍 addressed

shortstacked (approved):

dariacodes (commented):
(upd.) The code looks good,  tested it with the workflow I described [here](https://linear.app/n8n/issue/CAT-2310/question-n8n-trigger-active-workflow-updated-not-firing-in-multi-main#comment-8b0c568a). I do not see `instance started` event for workers though, only the leader.

dariacodes (commented):
Another setup to test on: having webhook nodes instead of data tables

PR #25857, opened 2026-02-16
fixes/19902-mcp-server-hangs

Nitin75408:
## Summary

MCP Server workflows hang indefinitely when a tool implemented via call-n8n-workflow invokes a sub-workflow containing a JavaScript Code node in queue/scaling mode. This fix addresses a multi-layered race condition and cleanup failure across 4 production files.

Root Cause
Race condition in queue mode: execute-workflow.tool.ts used mcpService.createPendingResponse() to wait for execution results, but this promise was resolved by handleMcpResponse() which raced with finalizeExecution() in the PCancelable handler. When finalizeExecution() ran first and deleted execution data from the DB, handleMcpResponse couldn't find the data and silently returned — never resolving the pending promise.

Silent failure in handleMcpResponse: When execution data wasn't found in the DB (due to the race above) or when an error occurred, the method returned without resolving or rejecting the pending promise, causing an indefinite hang.

RPC call zombie promises in task-runner: When JS Code node tasks are cancelled or timed out, cancelTaskRequests() cleaned up dataRequests and nodeTypesRequests but not rpcCalls. JS runners use RPC calls to fetch input data on-demand (unlike Python which sends data upfront with the task settings), leaving unresolved promises that prevent execution completion.

Changes
execute-workflow.tool.ts: Always use activeExecutions.getPostExecutePromise() — the deterministic resolution path that works in both regular and queue mode — instead of conditionally using mcpService.createPendingResponse() in queue mode.

scaling.service.ts: Make handleMcpResponse always resolve/cleanup pending promises even when execution data is not found in DB or when an error occurs, preventing indefinite hangs.

mcp.service.ts: Add resolveOrCleanupPendingResponse() method as a defense-in-depth mechanism to clean up stale pending responses without double-resolution risk.

task-runner.ts: Add taskId to RPCCall interface and extend cancelTaskRequests() to also reject and clean up pending RPC calls for cancelled/timed-out tasks, preventing zombie promises from JS Code node executions.

How to test
Set up n8n in queue mode with OFFLOAD_MANUAL_EXECUTIONS_TO_WORKERS=true and N8N_RUNNERS_MODE=internal
Create a workflow with a Webhook trigger → JavaScript Code node, enable it for MCP
Execute the workflow via MCP Server tool — it should complete without hanging
Verify the same workflow with a Python Code node still works correctly
Verify non-MCP workflows are unaffected

Fixes #19902 

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/task-runner/src/js-task-runner/__tests__/task-runner.test.ts
+++ b/packages/@n8n/task-runner/src/js-task-runner/__tests__/task-runner.test.ts
@@ -273,6 +273,74 @@ describe('TestRunner', () => {
 			expect(runner.dataRequests.size).toBe(0);
 			expect(runner.nodeTypesRequests.size).toBe(0);
 		});
+
+		it('should reject pending RPC calls when task is cancelled', async () => {
+			runner = newTestRunner();
+
+			const taskId = 'test-task';
+			const task = newTaskState(taskId);
+			task.status = 'running';
+			runner.runningTasks.set(taskId, task);
+
+			const rpcReject = jest.fn();
+
+			runner.rpcCalls.set('rpc-1', {
+				callId: 'rpc-1',
+				taskId,
+				resolve: jest.fn(),
+				reject: rpcReject,
+			});
+
+			await runner.taskCancelled(taskId, 'test-reason');
+
+			expect(rpcReject).toHaveBeenCalledWith(
+				expect.objectContaining({
+					message: 'Task cancelled: test-reason',
+				}),
+			);
+
+			expect(runner.rpcCalls.size).toBe(0);
+		});
+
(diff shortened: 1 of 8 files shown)


garritfra:
> *This was generated by AI during triage.*

Closing — this is one of 24 PRs opened by this account in mid-February 2026 with formulaic, AI-generated descriptions ("This PR fixes a critical X bug…", "Root Cause / Solution / Implementation"), generic `Fixes/<issue-number>` titles, and large multi-file diffs that go well beyond a focused fix. The pattern across the batch strongly suggests automated, low-context contribution rather than a debugged real fix. If a genuine bug fix here matches a documented issue, please open a single focused PR with reproduction steps, a tight scope, and a clear changelog entry.

PR #25858, opened 2026-02-16
ci: Add CI filter action

shortstacked:
## Summary

Implements a new GitHub action for filtering CI jobs based on changed files.

This action provides two modes:

- `filter`: Determines which jobs to run based on a provided filter definition.
- `validate`: Checks the results of required jobs and fails if any of them failed or were cancelled.

The action uses gitignore-style patterns to match changed files and determine whether a job should be executed. It replaces the previous `dorny/paths-filter` action with a custom implementation that offers more flexibility and control.

The workflow file is updated to use the new action and defines filters for `ci`, `unit`, and `workflows`.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.github/actions/ci-filter/action.yml
+++ b/.github/actions/ci-filter/action.yml
@@ -1,5 +1,10 @@
 name: 'CI Filter'
-description: 'Filter CI jobs by changed files and validate results. Zero dependencies.'
+description: |
+  Filter CI jobs by changed files and validate results.
+
+  Modes:
+    - filter: Determines which jobs to run based on changed files and a provided filter definition.
+    - validate: Checks the results of required jobs and fails if any of them failed or were cancelled.
 
 inputs:
   mode:
--- a/.github/actions/ci-filter/action.yml
+++ b/.github/actions/ci-filter/action.yml
@@ -1,0 +1,34 @@
+name: 'CI Filter'
+description: 'Filter CI jobs by changed files and validate results. Zero dependencies.'
+
+inputs:
+  mode:
+    description: 'filter or validate'
+    required: true
+  filters:
+    description: 'Filter definitions (gitignore-style DSL)'
+    required: false
+  base-ref:
+    description: 'Base branch for diff. Auto-detected if not specified.'
+    required: false
+  job-results:
+    description: 'Job results from needs context as JSON (mode=validate)'
+    required: false
+
+outputs:
+  results:
+    description: 'JSON object: { "filter-name": true/false }'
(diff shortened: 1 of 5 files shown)


.github/actions/ci-filter/action.yml
```diff
@@ -0,0 +1,34 @@
+name: 'CI Filter'
+description: 'Filter CI jobs by changed files and validate results. Zero dependencies.'
```
Matsuuu: >  Zero dependencies.

Do we need this advertisement text :D

.github/actions/ci-filter/action.yml
```diff
@@ -0,0 +1,34 @@
...
+
+inputs:
+  mode:
+    description: 'filter or validate'
```
Matsuuu, line 11: The PR contains info on these modes

> This action provides two modes:
> filter: Determines which jobs to run based on a provided filter definition.
> validate: Checks the results of required jobs and fails if any of them failed or were cancelled.

Could this info be also appended to this description or into a comment?

.github/actions/ci-filter/ci-filter.mjs
```diff
@@ -0,0 +1,216 @@
...
+import { fileURLToPath } from 'node:url';
+import { resolve } from 'node:path';
+
+// --- Glob matching (dotfile-safe) ---
+
+/**
+ * Match a file path against a glob pattern.
+ * Unlike path.matchesGlob / standard POSIX globs, `**` matches dotfiles.
+ */
+export function matchGlob(filePath, pattern) {
```
Matsuuu, line 12: While being dependency-less, I wonder if we could just utilize `glob`, which we already have as a dependency at `.github/scripts`

Matsuuu: Not saying this wouldn't work, moreso that a standardized, largely adopted helper library usually helps us in the long run

Matsuuu (dismissed):
Looks good to me. Some comments / nits

.github/actions/ci-filter/ci-filter.mjs
shortstacked: From my understanding glob would help for doing file level matching, but in this case we are actually doing string level matching e.g 'mypath/**' not the actual files though.

Matsuuu: Ah yeah that might be true. Was just confused as it has the word Glob in the function name :D 

Let's proceed with the current implementation

Matsuuu (approved):
🚀

PR #25865, opened 2026-02-17
fix: Prevent node mention popup when typing email addresses

aalises:
## Summary

Kapture 2026-02-17 at 08 54 15

Fixes [AI-2058](https://linear.app/n8n/issue/AI-2058)

The `@` node mention dropdown was triggering on any `@` character typed in the AI builder chat input, including when typing email addresses (e.g. `eugene@n8n.io`). Additionally, pressing Escape to dismiss the popup would destructively delete the `@` and subsequent characters.

**Changes:**
- Only activate the mention dropdown when `@` is preceded by whitespace or is at the start of input — skip it when `@` follows a word character (as in email addresses)
- Make Escape non-destructive: closing the dropdown now always preserves the typed text


burivuhster (approved):
Looks good!

PR #25867, opened 2026-02-17
fix(AI Agent Node): Serialize AI Agent intermediateSteps messageLog to plain objects

### LINKED ISSUE #22779: If statement not working
### Bug Description

Image

How do I explain this?
The if statement is returning "true" as shown but during execution / runtime its failing as if it were false?

### To Reproduce

1. Add chat trigger and create an agent node and check intermediate steps.
2. Connect an if statement to the output and paste expression: `{{ JSON.parse($('Your Agent').first().json.intermediateSteps[0].observation).some(x => JSON.parse(x.response?.[0]?.text)?.[0].results != null) && !$('Your Agent').first().json.intermediateSteps.some(x => x.action.messageLog[0].kwargs.tool_calls[0].args.tool === 'SomeTool')}}` 
(annoying to parse so many times but ok)
3. Implement "SomeTool" that outputs `{ results: [] }` and connect it to the agent.

### Expected behavior

The if statement should fail as shown in the screenshot.
Even if it shows true its truly not because its failing.

### Debug Info

# Debug info

## core

- n8nVersion: 1.121.2
- platform: docker (cloud)
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: 20
- license: community

## storage

- success: all
- error: all
- progress: false
- manual: false
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 720 hours
- maxCount: 25000 executions

## client

- userAgent: mozilla/5.0 (x11; linux x86_64) applewebkit/537.36 (khtml, like gecko) chrome/140.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2025-12-04T22:08:37.774Z

### Operating System

ex. Ubuntu Linux 22.04

### n8n Version

1.121.2

### Node.js Version

ex. 22.16.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

aalises:
## Summary

When a workflow has an IF/Set node downstream of an AI Agent that references `intermediateSteps[].action.messageLog` properties, expressions evaluate to `null` during execution but work when the node is re-executed individually afterward.

**Root cause:** The AI Agent stores live LangChain `AIMessage` class instances in `messageLog`. These instances have properties like `content` and `tool_calls` directly accessible, but their `toJSON()` method wraps everything under `kwargs`. 

When `JSON.stringify` runs (for UI/DB storage), it calls `toJSON()`, producing the `kwargs`-wrapped format that users see and use to build expressions. At runtime, `kwargs` doesn't exist on the live instances, so those expressions return `null`.

**Fix:** Serialize `messageLog` entries to plain objects after agent execution completes (in both V2 and V3 finalization paths). This ensures the data structure is identical at runtime and in the UI. 

Properties like `content`, `tool_calls`, etc. are directly accessible without a `kwargs` wrapper.

- https://linear.app/n8n/issue/AI-1855
- Fixes #22779

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V2/execute.ts
+++ b/packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V2/execute.ts
@@ -16,6 +16,7 @@ import { jsonParse, NodeOperationError, sleep } from 'n8n-workflow';
 import type { IExecuteFunctions, INodeExecutionData, ISupplyDataFunctions } from 'n8n-workflow';
 import assert from 'node:assert';
 
+import { serializeIntermediateSteps } from '@utils/agent-execution';
 import { getPromptInputByType } from '@utils/helpers';
 import {
 	getOptionalOutputParser,
@@ -348,6 +349,12 @@ export async function toolsAgentExecute(
 					response.output as string,
 				);
 				response.output = parsedOutput?.output ?? parsedOutput;
+			}
+
+			// Serialize messageLog entries from LangChain class instances to plain objects
+			// so that downstream expressions see the same structure as the UI data browser.
+			if (response.intermediateSteps) {
+				serializeIntermediateSteps(response.intermediateSteps);
 			}
 
 			// Omit internal keys before returning the result.
--- a/packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V2/execute.ts
+++ b/packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V2/execute.ts
@@ -16,7 +16,6 @@ import { jsonParse, NodeOperationError, sleep } from 'n8n-workflow';
 import type { IExecuteFunctions, INodeExecutionData, ISupplyDataFunctions } from 'n8n-workflow';
 import assert from 'node:assert';
 
-import { serializeIntermediateSteps } from '@utils/agent-execution';
 import { getPromptInputByType } from '@utils/helpers';
 import {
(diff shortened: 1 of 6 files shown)


packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V2/execute.ts
```diff
@@ -350,6 +351,12 @@ export async function toolsAgentExecute(
 				response.output = parsedOutput?.output ?? parsedOutput;
 			}
 
+			// Serialize messageLog entries from LangChain class instances to plain objects
+			// so that downstream expressions see the same structure as the UI data browser.
+			if (response.intermediateSteps) {
+				serializeIntermediateSteps(response.intermediateSteps);
+			}
```
schrothbn: Don't think we need that here, we never got a report on v2 for this.

schrothbn (dismissed):
Don't think it's necessary in V2 as we don't have a report for that. It's using Langchain directly to track the steps while we build them in v3 ourselves, which probably caused this inconsistency. 
We might risk to break existing workflows there if we change it like that, depending on if something's actually different.  Best case we just don't change anything and add that step for no reason. Won't block it on that though.

But other than that LGTM.

packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V2/execute.ts
aalises: done! removed for v2

schrothbn (approved):

PR #25882, opened 2026-02-17
fix(Azure Cosmos DB Node): Preserve query parameter types instead of converting to strings

### LINKED ISSUE #23398: [CosmosDB Node] problem with query params timestamp converted to string
### Bug Description

Azure Cosmos DB Execute Query node in n8n returns an empty result set when using query parameters for a numeric timestamp, even though the exact same SQL query with hardcoded values works correctly in Azure Cosmos DB Data Explorer.

### To Reproduce

1. Use Azure Cosmos DB node – Execute Query in n8n.
2. Write a query with two parameters, e.g.:

`SELECT * FROM c WHERE c.projectId=@Param1 AND c.startDate=@Param2`

 3. Provide query parameters:

`[
  { "name": "@Param1", "value": "P12223" },
  { "name": "@Param2", "value": 1737062400000 }
]`

4. Execute the node.
5. The node returns an empty array.
6. Capturing the outgoing payload shows Cosmos DB receives the parameters with the timestamp parameter as a string, e.g.:
`{
  "query": "SELECT * FROM c where c.projectId=@Param1 and c.startDate=@Param2",
  "parameters": [
    { "name": "@Param1", "value": "P12223" },
    { "name": "@Param2", "value": "1737062400000" }  // <-- string!
  ]
}`
Because the timestamp becomes a string instead of a number, the query does not match any documents even though valid records exist.

### Expected behavior

The Cosmos DB node should send numeric query parameters (e.g., timestamp) as numbers, not strings. Example expected parameter payload:

`{ "name": "@Param2", "value": 1737062400000 }`

(without quotes around the number)

### Debug Info

# Debug info

## core

- n8nVersion: 1.117.3
- nodeJsVersion: 22.18.0
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- consumerId: ce1a5293-9e16-434d-a729-dcc7a8a4058a

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/143.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2025-12-18T10:07:40.147Z

### Operating System

Mac OS 26.1

### n8n Version

1.117.3

### Node.js Version

22.18.0

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

strzeluk:
## Summary

The `validateQueryParameters()` function in the Azure Cosmos DB node was converting all query parameter values to strings, regardless of their actual type. This caused numeric values (e.g. timestamps like `1737062400000`) to be sent as `"1737062400000"` instead of `1737062400000`, resulting in empty query results when comparing against numeric fields in Cosmos DB.

The fix adds type detection when parsing comma-separated parameter values:

* Numeric strings are converted to numbers
* `true`/`false` are converted to booleans
* `null` is converted to nullog
* Everything else remains a string

* Fixes [https://github.com/n8n-io/n8n/issues/23398](<https://github.com/n8n-io/n8n/issues/23398>)
* [NODE-4938](https://linear.app/n8n/issue/NODE-4938)

## Review / Merge checklist

- [X] PR title and summary are descriptive. ([conventions](<../blob/master/.github/pull_request_title_conventions.md>))
- [X] [Docs updated](<https://github.com/n8n-io/n8n-docs>) or follow-up ticket created.
- [X] Tests included.


packages/nodes-base/nodes/Microsoft/AzureCosmosDb/helpers/utils.ts
konstantintieber: I wonder whether it's safe to assume that a string that contains only numbers should always be sent as number.
I'm not familiar with Azure Cosmos DB. For reviewing the chance, it'd be helpful if you could share the API reference on their docs for example, to cross check that this is what they expect.

strzeluk: Azure Cosmos DB SQL API performs type-sensitive comparisons — if a field in the database stores a number and the query parameter is sent as a string, the query returns no results. This is exactly the bug described in #23398.

Relevant docs: [Parameterized queries - Azure Cosmos DB for NoSQL](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/parameterized-queries)

The queryParameters input accepts a comma-separated string (e.g. 1737062400000,active). Because this format carries no type metadata, there's no way to distinguish the string "1737062400000" from the number 1737062400000. The heuristic (parse as a number if possible) is the best we can do within the existing input format and directly fixes the reported failure case.

The edge case you're raising — a string field that happens to contain only digits — is a pre-existing limitation of the comma-separated input design, not something introduced by this fix. Such a user would have had no working path before this change either.

konstantintieber: > is a pre-existing limitation of the comma-separated input design, not something introduced by this fix

Is that really true?
I'm thinking of a zip code for example:

| User types `01234` → | Cosmos field = string `"01234"` | Cosmos field = number `1234` |
|---|---|---|
| **Before the PR** (always string) | ✅ matches (`"01234"` == `"01234"`) | ❌ no match (the reported bug) |
| **After the PR** (numeric heuristic) | ❌ no match (`1234` vs `"01234"`) | ✅ matches |

So before this change, strings that only contain digits, e.g. zip codes, would be sent properly to the cosmos query.
After this change, it will fail.

strzeluk: You're right, and I was wrong to claim it was a pre-existing limitation — the regression is real.

The root issue is that the comma-separated string format has no way to encode type intent. I see two paths forward:

Change the input to accept a JSON array, e.g. [1737062400000, "01234", "active"] — which gives users explicit control over types and fixes both cases cleanly
Keep the heuristic and document the trade-off in the field description (but this is clearly the worse option)
Option 1 seems like the right fix. Since numeric parameters never worked correctly before this PR, there shouldn't be any users with a working setup that relies on the comma-separated format for numbers — so changing the format isn't really a breaking change for anyone in practice. Happy to update the implementation if you agree with this direction.

alexander-gekov (commented):
Hey @strzeluk, thanks for the contribution!

Regarding the convo about queryParameters, I think going with Option 1 makes the most sense.

A few notes on the implementation:
* The comma-separated string field does work today for string-only parameters, so there are workflows relying on it. I’d suggest keeping backwards compatibility either by minor-versioning the node or simpler - adding a pure additive `queryParametersJson` field
* Please add unit test coverage for cases like "012345", large integers `(> Number.MAX_SAFE_INTEGER)`, booleans, nulls, empty values, and an empty parameters array and also one test for the existing comma-separated path to confirm it hasn’t regressed.

strzeluk:
Here's a summary of what was updated based on your feedback:

**Changes made:**

1. **Added `queryParametersJson` field** — new additive string field accepting a JSON array (e.g. `[1737062400000, "01234", true, null]`). Preserves types explicitly: numbers, booleans, null, and strings with leading zeros.

2. **Reverted `queryParameters` to string-only behavior** — removed the numeric heuristic, restoring backward compatibility for existing workflows.

3. **Updated `validateQueryParameters`** — checks `queryParametersJson` first; falls back to `queryParameters` (comma-separated strings) when not set.

4. **Unit test coverage added** for all cases you mentioned:
   - `"012345"` → preserved as string `"012345"` ✓
   - Large integers > `Number.MAX_SAFE_INTEGER` → precision loss documented as expected behavior ✓
   - Booleans, nulls, empty string values, empty array ✓
   - Invalid (non-array) JSON input → throws `NodeOperationError` ✓
   - Regression: comma-separated path now returns strings only ✓

**Question about `queryNumericParams.workflow.json`:**

I've updated this integration test to use `queryParametersJson: ["P12223", 1737062400000]` instead of the old comma-separated format (since the numeric heuristic no longer exists on `queryParameters`). This change is included in the current commit — but would you prefer it as a separate commit to keep the diff cleaner, or is it fine as part of this PR?

**Also:** If this implementation looks good to you, I'll add documentation updates for this node as well.

strzeluk:
Documentation for the new **Query Parameters (JSON)** field has been prepared in a separate PR: n8n-io/n8n-docs#4636

alexander-gekov (approved):
Looks good!

PR #25884, opened 2026-02-17
feat: Add connection state to secret stores response (no-changelog)

ireneea:
## Summary

image

- Added `state` property to the API response
- Implemented `getProviderProperties` method to fetch provider-specific connection information
- No need to change the frontend, there was already support for the state

### Testing:

1. Add connection with valid details -> Should not show any indicator on the screen
2. Add a connection with invalid details -> Should show a `Disconnected` label next to its 

https://linear.app/n8n/issue/LIGO-240

## Review / Merge checklist

- [x] Tests included (added comprehensive tests for all connection states)

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
+++ b/packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
@@ -203,12 +203,14 @@ export class SecretsProvidersConnectionsService {
 		connection: SecretsProviderConnection,
 	): SecretsProvidersResponses.ConnectionListItem {
 		const secretNames = this.externalSecretsManager.getSecretNames(connection.providerKey);
+		const connectionInstance = this.externalSecretsManager.getProvider(connection.providerKey);
 
 		return {
 			id: String(connection.id),
 			name: connection.providerKey,
 			type: connection.type as SecretsProviderType,
 			secretsCount: secretNames.length,
+			state: connectionInstance?.state ?? 'initializing',
 			projects: connection.projectAccess.map((access) => ({
 				id: access.project.id,
 				name: access.project.name,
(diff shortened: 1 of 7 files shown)


packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -220,15 +222,17 @@ export class SecretsProvidersConnectionsService {
 
 	toPublicConnection(connection: SecretsProviderConnection): SecretsProvidersResponses.Connection {
 		const decryptedSettings = this.decryptConnectionSettings(connection.encryptedSettings);
-		const { provider } = this.externalSecretsManager.getProviderWithSettings(connection.type);
-		const redactedSettings = this.redactionService.redact(decryptedSettings, provider.properties);
+		const properties = this.externalSecretsManager.getProviderProperties(connection.type);
```
ireneea, line 227: Wanted to get away from the "old" methods where there is an assumption that we are working with a fixed set of global providers. That's why I added a dedicated method just to get the "type" properties.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -220,15 +222,17 @@ export class SecretsProvidersConnectionsService {
...
+		const redactedSettings = this.redactionService.redact(decryptedSettings, properties);
 		const secretNames = this.externalSecretsManager.getSecretNames(connection.providerKey);
+		const connectionInstance = this.externalSecretsManager.getProvider(connection.providerKey);
 
 		return {
 			id: String(connection.id),
 			name: connection.providerKey,
 			type: connection.type as SecretsProviderType,
 			secretsCount: secretNames.length,
+			state: connectionInstance?.state ?? 'initializing',
```
ireneea, line 239: Using the same default state, as the `SecretsProvider`. This is in case the provider exists in the db but not yet in the registry. This can happen in multi-main when a node has not yet  handled the `reload-external-secrets-provider` event to update its registry

konstantintieber: Thx for the comment. Was wondering about the optional chaining here.
I think it'd be worth documenting that in a code comment.
Both in `toPublicConnectionListItem` and `toPublicConnection`, which have this optional chaining.

konstantintieber (dismissed):
Can't test it manually right now due to broken build, but the code looks good 👍

konstantintieber (approved):

PR #25885, opened 2026-02-17
feat(editor): Add review changes banner to AI builder

burivuhster:
## Summary

Adds a "Review Changes" banner to the AI builder chat that shows users how many nodes were edited by the AI in the latest interaction. Clicking it opens a diff modal comparing the workflow version before the AI edit with the current state.

**Key changes:**
- Added `latestRevertVersion` computed to `builder.store.ts` — finds the most recent message with a revert version (the version snapshot taken before the last AI edit)
- Created `useReviewChanges` composable — encapsulates version fetching, node diff computation, and diff modal opening logic
- Created `ReviewChangesBanner.vue` — compact banner showing edited node count with click-to-review
- Created `AIBuilderDiffModal.vue` — modal wrapper around the existing `WorkflowDiffView` component
- Gated behind `AI_BUILDER_REVIEW_CHANGES_EXPERIMENT` PostHog feature flag
- Banner is hidden when streaming, when no revert version exists, or when 0 nodes differ

**How to test:**
1. Enable the `075_ai_builder_review_changes ` feature flag
2. Open the AI builder and generate a workflow
3. Send a follow-up message that modifies existing nodes
4. Verify the "Review N changes" banner appears above the chat input
5. Click the banner to open the diff modal showing before/after comparison

https://linear.app/n8n/issue/AI-2001

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/features/ai/assistant/composables/useReviewChanges.ts
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	const editedNodesCount = computed(() => {
+		if (!cachedVersionLoaded.value) return 0;
+		const normalized = resolveNodeDefaults(cachedVersionNodes.value);
+		const diff = compareWorkflowsNodes(normalized, workflowsStore.workflow.nodes);
+		return [...diff.values()].filter((d) => d.status !== NodeDiffStatus.Eq).length;
+	});
+
+	const showReviewChanges = computed(() => {
```
aalises, line 88: The same version is already fetched in the watch on lline 35, but only nodes are cached there. This means every click on "Review changes"  makes a redundant API call for the same version.

Maybe we can consider caching the full version object (nodes + connections) in the watcher so `openDiffView` can reuse it without a second network round-trip. Not sure of the overhead that would add

packages/frontend/editor-ui/src/features/ai/assistant/composables/useReviewChanges.ts
```diff
@@ -0,0 +1,131 @@
...
+	const uiStore = useUIStore();
+	const posthogStore = usePostHog();
+	const i18n = useI18n();
+	const toast = useToast();
+
+	const isLoadingDiff = ref(false);
+	const cachedVersionNodes = ref<INode[]>([]);
+	const cachedVersionLoaded = ref(false);
+
+	watch(
```
aalises, line 32: Could be a race condition: if `latestRevertVersion` changes rapidly (e.g. user sends multiple messages in quick succession), two async fetches race and the last to resolve wins, which may not correspond to the latest version if a previous one ended later for some reason.

We can maybe capture version.id before the await and checking it still matches `builderStore.latestRevertVersion?.id` before writing to the cache

packages/frontend/editor-ui/src/features/ai/assistant/components/Agent/AIBuilderDiffModal.vue
```diff
@@ -0,0 +1,104 @@
...
+	builderStore.trackWorkflowBuilderJourney('user_opened_review_changes');
+});
+
+function handleBeforeClose() {
+	telemetry.track('Workflow diff view closed', { source: 'ai-builder-review' });
+	builderStore.trackWorkflowBuilderJourney('user_closed_review_changes');
+}
+
+function closeModal() {
+	handleBeforeClose();
```
aalises: HandleBeforeClose is also bound as `:before-close` on the Modal already

If the Modal fires its `before-close` callback during its close lifecycle, telemetry events (Workflow diff view closed + user_closed_review_changes) will be tracked twice when the user clicks the back button as we're also calling the handle before close

We can rely only on the before-close maybe

packages/frontend/editor-ui/src/features/ai/assistant/composables/useReviewChanges.ts
```diff
@@ -0,0 +1,131 @@
...
+				true,
+				false,
+				node,
+				nodeType,
+			);
+			return { ...node, parameters: resolved ?? {} };
+		});
+	}
+
+	const editedNodesCount = computed(() => {
```
aalises, line 81: Resolving the node defaults iterates over all the workflow right? on big workflows it can evaluate a lot and maybe become not that performant. 

Perhaps we can only compute when cachedVersionLoaded && streaming), or debounce it?

aalises (approved):

PR #25888, opened 2026-02-17
ci: Enable Sentry source maps in Docker builds

shortstacked:
## Summary
- Adds `RELEASE` and `SENTRY_AUTH_TOKEN` env vars to the Docker build's `setup-nodejs` step
- This enables the Sentry Vite plugin during Docker builds, injecting debug IDs into frontend JS bundles and uploading matching source maps
- Without this, frontend source maps in Sentry show "Uploaded Files Not Deployed" because the Docker build produced JS without debug IDs

## Context
The frontend Sentry Vite plugin (added this release) only activates when `RELEASE` is set. The NPM publish workflow sets this, but the Docker build — which produces the **actually deployed** code — did not. Since the JS in Docker images had no debug IDs, Sentry could never match browser errors to the uploaded source maps.

The previous CI approach (`Create a Sentry Release` job uploading frontend source maps) also never worked because Vite generates content-hashed filenames, and the CI job's build produced different hashes than the Docker build.

The `SENTRY_AUTH_TOKEN` is only used during the build step on the GitHub Actions runner — it is **not** passed as a Docker build arg and does not end up in the Docker image.

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -90,6 +90,9 @@ jobs:
         with:
           build-command: pnpm build:n8n
           enable-docker-cache: 'true'
+        env:
+          RELEASE: ${{ needs.determine-build-context.outputs.n8n_version }}
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
 
       - name: Determine Docker tags for all images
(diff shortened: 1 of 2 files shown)


Matsuuu (approved):
LGTM, Make sure to enable the correct backporting labels depending on which versions you want to backport this to

PR #25893, opened 2026-02-17
refactor(core): Replace Code nodes with TestDataNode in workflow test fixtures

tomi:
## Summary

Introduces a lightweight `TestDataNode` that provides static JSON data without requiring JavaScript sandbox execution. This improves test performance and clarity by replacing Code nodes in 66 workflow test fixtures. The new node converts JavaScript code strings to declarative JSON data structures, eliminating unnecessary JS evaluation overhead in tests that only need to provide static test data.

The motivation of this is to get rid of use of Code node as a dependency when testing other nodes. Relates to https://github.com/n8n-io/n8n/pull/25834

## Review / Merge checklist

- [x] Tests included.


dlavrenuek (approved):
Great improvement! I did fly through the test files to see if the new test data object is the same as previous code node style data, but did not check every single line there. I would assume that the tests will fail if the expected data is different 🥇

PR #25899, opened 2026-02-17
ci: Add DB integration tests to CI workflow instead of standalone

shortstacked:
## Summary

Introduces a new reusable workflow for running database integration tests. This workflow is triggered on pull requests that include changes to database-related files. It includes tests for SQLite and Postgres.

Removes the dedicated DB test workflow and integrates DB tests into the main CI workflow.

The new `db-tests` job is added to the `required-checks` job to ensure that DB tests pass before merging.

Same approach as before but keeps it all in one view so it's easier to see

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/backend-test-utils/src/test-db.ts
```diff
@@ -65,6 +68,22 @@ export async function terminate() {
...
+	if (testDbName && originalDatabase) {
+		const globalConfig = Container.get(GlobalConfig);
+		if (globalConfig.database.type === 'postgresdb') {
+			try {
+				globalConfig.database.postgresdb.database = originalDatabase;
+				const bootstrap = await new Connection(getBootstrapDBOptions()).initialize();
+				await bootstrap.query(`DROP DATABASE IF EXISTS "${testDbName}"`);
+				await bootstrap.destroy();
+			} catch {
+				// Best effort - don't fail tests over cleanup
```
despairblue, line 81: But maybe log it?

.github/workflows/test-db.yml
```diff
@@ -1,98 +0,0 @@
-name: "Test: DB Postgres"
-
-on:
-  schedule:
-    - cron: "0 0 * * *"
```
despairblue, line 5: I assume we don't really need this on a schedule, right? 
This could have caught issues introduced by changes outside of files in the filter list, but then again I don't think anybody looked into it when it did happen, so probably ok to remove it.

shortstacked: That was my thought process. I asked if anyone looked at it and no one seemed to. It would be caught on the next DB PR anyway which we do frequently enough.

despairblue (approved):

PR #25916, opened 2026-02-18
feat(core): Add mcp configs to claude gh action (no-changelog)

scdekov:
## Summary

* Dropped the need for passing personal access token as input and now using the `github.token`. Need to make sure this can still open mergable PRs
* Added install chromium step and configured playwright mcp
* Configured notion and linear MCPs

## Review / Merge checklist

- [ ] Tests included. 


.github/workflows/util-claude-task.yml
```diff
@@ -105,16 +95,47 @@ jobs:
             echo 'EOF'
           } >> "$GITHUB_ENV"
 
+      - name: Install Playwright browser
```
shortstacked: If we are doing this, I would actually go with doing the full project e.g build/install and then doing this from the Playwright project inside. This would give us the browser but it wouldn't give us the code base or the testing framework to use.

.github/workflows/util-claude-task.yml
```diff
@@ -131,7 +152,7 @@ jobs:
 
       - name: Push branch and create PR
         env:
-          GH_TOKEN: ${{ inputs.user_token }}
+          GH_TOKEN: ${{ github.token }}
```
shortstacked: This should be good, as long as when it gets triggered by a bot it still works.

.github/workflows/util-claude-task.yml
scdekov: Do you know how much overhead(time) will that add to the action? I've added it like this as we needed playwright MCP for the triage skill that was fetching video transcripts from loom. I think the build may be optional and triggered only if we're planning to reproduce the issue otherwise it's going to slow down this action unnecessarily. Maybe we look at another way to fetch those transcripts 🤔

.github/workflows/util-claude-task.yml
scdekov: Yeah, I suspect PRs may no be mergable as they'll be missing the PAT. I'll try it out

.github/workflows/util-claude-task.yml
scdekov: I think I like the idea of having an n8n workflow or something that CC can call here and not having to install playwright only or that.

.github/workflows/util-claude-task.yml
shortstacked: They should be mergable, the original approach done it this way. It will create it with your permission/pat so the PR should come from you.

.github/workflows/util-claude-task.yml
shortstacked: Regular build/install assuming cached would be around 1 minute. Would it make sense to have the video transcription in a callable workflow in n8n then. Something like "Call with ticket ID, get video, send for decoding/transcription, then update the ticket"?

scdekov: Let's see. It's definitely simpler approach but I was wondering for the video decoding specifically if there will be a difference between claude code doing the decoding and getting the frames vs doing the decoding externally and pass to CC the analysis of the video. Will keep this branch like this for a little more as it unblocks experimenting with the action and skill and then we decide how to move forward.

shortstacked: Oh that would be an interesting one to check alright!

shortstacked (dismissed):

shortstacked (approved):

PR #25921, opened 2026-02-18
fix: Only use displayOptions for determining credential required fields for update if the property is defined

BGZStephen:
## Summary

This PR addresses the fact that `displayOptions` is not always defined on a credential definition, which in turn, means any field that doesn't have that field, cannot be set up for credential overrides that essentially no longer require that field.

IAM-271

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/credentials/credentials.service.ts
+++ b/packages/cli/src/credentials/credentials.service.ts
@@ -950,7 +950,7 @@ export class CredentialsService {
 		const credentialProperties = this.credentialsHelper.getCredentialsProperties(type);
 		for (const property of credentialProperties) {
 			/*
-			 * displyaOpyions is possibly undefined, which causes displayParameter to default `true`, which is expected behaviur for UI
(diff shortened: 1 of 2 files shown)


packages/cli/src/credentials/credentials.service.ts
```diff
@@ -949,7 +949,16 @@ export class CredentialsService {
 		// check mandatory fields are present
 		const credentialProperties = this.credentialsHelper.getCredentialsProperties(type);
 		for (const property of credentialProperties) {
-			if (property.required && displayParameter(data, property, null, null)) {
+			/*
+			 * displyaOpyions is possibly undefined, which causes displayParameter to default `true`, which is expected behaviur for UI
```
guillaumejacquart:
```suggestion
			 * displayOpyions is possibly undefined, which causes displayParameter to default `true`, which is expected behavior for UI
```

guillaumejacquart (dismissed):
good catch !

guillaumejacquart (approved):

PR #25927, opened 2026-02-18
feat(core): Filter secrets provider autocomplete to connected vaults only

alielkhateeb:
## Summary

Improves the user experience when selecting secrets provider connections by filtering autocomplete suggestions to only show connections from providers that are successfully connected - hiding unsuccessfully connected providers from the list.

Previously, the autocomplete would show all connections regardless of whether their provider was connected, which could confuse users by showing options that couldn't actually be used.

**Changes:**
- Repository methods now accept optional `types` filter parameter
- Service uses provider registry to get list of connected providers  
- Autocomplete results are filtered to only connected provider types
- Added integration tests for the new filtering behavior

**Testing:**
1. Set up multiple secrets provider connections (e.g., AWS, HashiCorp Vault)
    a. Make sure at least one connection is not connected (e.g. use fake access token)
2. Open a credential or node that uses secrets
3. Type to trigger autocomplete for secrets provider selection
4. Verify only connections from connected providers appear in suggestions
5. Fix the invalid provider (use a correct token to successfully connect)
6. Verify that this connection now appear in autocomplete

Note: You can test this for global and project based secrets provider connections

https://linear.app/n8n/issue/LIGO-235

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
+++ b/packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
@@ -28,19 +28,23 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
 	 * Retrieves global connections, i.e., connections that are not assigned to any project.
 	 * A global connection has no associated entries in its projectAccess relation.
 	 *
-	 * @param typeFilter - (Optional) Limits results to connections of the specified provider type.
+	 * @param filters - (Optional) Filters to apply to the query.
+	 * @param filters.providerKeys - (Optional) Limits results to connections of the specified provider keys.
 	 * @returns Promise resolving to all matching SecretsProviderConnection entities.
 	 */
-	async findGlobalConnections({
-		providerKeys,
-	}: { providerKeys?: Array<SecretsProviderConnection['providerKey']> } = {}): Promise<
-		SecretsProviderConnection[]
-	> {
+	async findGlobalConnections(
+		filters: { providerKeys?: Array<SecretsProviderConnection['providerKey']> } = {},
+	): Promise<SecretsProviderConnection[]> {
+		const { providerKeys } = filters;
+		if (providerKeys && providerKeys.length === 0) {
+			return [];
+		}
+
 		const connectionQuery = this.createQueryBuilder('connection')
 			.leftJoin('connection.projectAccess', 'access')
 			.where('access.secretsProviderConnectionId IS NULL');
 
-		if (providerKeys && providerKeys.length > 0) {
+		if (providerKeys) {
 			connectionQuery.andWhere('connection.providerKey IN (:...providerKeys)', { providerKeys });
 		}
 
@@ -55,19 +59,25 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
 	 * not global connections (i.e., those without any project restriction).
 	 *
 	 * @param projectId - The ID of the project to filter on.
-	 * @param type - (Optional) The provider type to filter connections by.
+	 * @param filters - (Optional) Filters to apply to the query.
+	 * @param filters.providerKeys - (Optional) Limits results to connections of the specified provider keys.
 	 * @returns Promise resolving to all matching SecretsProviderConnection entities.
 	 */
 	async findByProjectId(
 		projectId: string,
-		{ providerKeys }: { providerKeys?: Array<SecretsProviderConnection['providerKey']> } = {},
+		filters: { providerKeys?: Array<SecretsProviderConnection['providerKey']> } = {},
 	): Promise<SecretsProviderConnection[]> {
+		const { providerKeys } = filters;
+		if (providerKeys && providerKeys.length === 0) {
+			return [];
+		}
+
 		const connectionQuery = this.createQueryBuilder('connection')
 			.innerJoinAndSelect('connection.projectAccess', 'projectAccess')
 			.leftJoinAndSelect('projectAccess.project', 'project')
 			.where('projectAccess.projectId = :projectId', { projectId });
 
-		if (providerKeys && providerKeys.length > 0) {
+		if (providerKeys) {
 			connectionQuery.andWhere('connection.providerKey IN (:...providerKeys)', { providerKeys });
 		}
 
--- a/packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
+++ b/packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
@@ -32,16 +32,16 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
 	 * @returns Promise resolving to all matching SecretsProviderConnection entities.
 	 */
 	async findGlobalConnections({
-		types,
-	}: { types?: Array<SecretsProviderConnection['type']> } = {}): Promise<
+		providerKeys,
+	}: { providerKeys?: Array<SecretsProviderConnection['providerKey']> } = {}): Promise<
 		SecretsProviderConnection[]
 	> {
 		const connectionQuery = this.createQueryBuilder('connection')
 			.leftJoin('connection.projectAccess', 'access')
 			.where('access.secretsProviderConnectionId IS NULL');
 
-		if (types && types.length > 0) {
-			connectionQuery.andWhere('connection.type IN (:...types)', { types });
+		if (providerKeys && providerKeys.length > 0) {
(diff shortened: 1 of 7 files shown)


packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -179,11 +181,21 @@ export class SecretsProvidersConnectionsService {
 	}
 
 	async getGlobalCompletions(): Promise<SecretsProviderConnection[]> {
-		return await this.repository.findGlobalConnections();
+		const connectedProviders = this.providerRegistry.getConnected();
+		const connectedProviderNames = connectedProviders.map((provider) => provider.name);
+		const connections = await this.repository.findGlobalConnections({
+			types: connectedProviderNames,
```
konstantintieber: Naming: Shouldn't this filter by providerKey, rather than type?

konstantintieber: What's "name" on the `SecretsProvider` from the registry is "providerKey" on the `SecretsProviderConnection` entity

alielkhateeb: @konstantintieber `type` here is the name of the connected vault, `providerKey` is the user input for the connection name, the terminology here is incredibly confusing tbh 😄

alielkhateeb: So from the `ProvierRegistry`, we get the providers that are actually connected, we extract the "names", then we use that to filter `SecretsProviderConnection` by `type`

konstantintieber: Hmm that's definitely confusing (I still don't get it, will try to understand it after lunch)

alielkhateeb: @konstantintieber Yup, very confusing. Here is an example:
image

Above is from the `ProviderRegistry`, below is `SecretsProviderConnection`, as you can see it's `vault` instead of `my-vault`, I would have expected this property to be called `type` instead of `name` in the provider registry, then it might have been a little bit less confusing 😂

konstantintieber: Video explaining why I don't think filtering by provider type will fully solve the issue that was raised in the ticket this PR belongs to:
https://www.loom.com/share/eeb490d8505048ae93a2fb5149c705d9

alielkhateeb: Problem is the provider registry is registering by provider type and not by provider name in all cases, we fallback to the `providerType` [here](https://github.com/n8n-io/n8n/blob/217c833f9d29e51a53fa4c1a6c10d25853ea55e6/packages/cli/src/modules/external-secrets.ee/external-secrets-manager.ee.ts#L341) in case a name was not provided. I am going to refactor this to always require a `providerKey` which is a synonym of "connection name".

konstantintieber: Each active connection is an instance of [SecretsProvider](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/modules/external-secrets.ee/types.ts#L44), which has `state` property.
When we determine which providerKeys to include the completions response, we should only include `SecretProviders` that have `state='connected'`.
I could see the `ExternalSecretsProviders` exposing a new method like `getAllConnectedProviders`, which filters out any providers that don't have this state.

alielkhateeb: `getConnected` from provider registry was not used anywhere 😅 and it was returning the "values" so there wasn't any clear indicator for me to filter by the provider key, but thanks to this comment, now I understand what was wrong in the existing methods and in this PR, I refactored `getConnected` and added a `getConnectedNames` here f1177bb03c56f78e33a6919aae864fd92e7df189

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -25,21 +25,53 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
...
+		}
+
+		return await connectionQuery.getMany();
 	}
 
-	async findByProjectId(projectId: string): Promise<SecretsProviderConnection[]> {
-		return await this.createQueryBuilder('connection')
+	/**
+	 * Finds all secrets provider connections assigned to a given project.
+	 * Optionally filters connections by provider type.
```
konstantintieber: This docblock is outdated

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
```diff
@@ -25,21 +25,53 @@ export class SecretsProviderConnectionRepository extends Repository<SecretsProvi
...
+	 * This returns only those connections explicitly linked to the project,
+	 * not global connections (i.e., those without any project restriction).
+	 *
+	 * @param projectId - The ID of the project to filter on.
+	 * @param type - (Optional) The provider type to filter connections by.
+	 * @returns Promise resolving to all matching SecretsProviderConnection entities.
+	 */
+	async findByProjectId(
+		projectId: string,
+		{ providerKeys }: { providerKeys?: Array<SecretsProviderConnection['providerKey']> } = {},
```
konstantintieber: From the function signature, I can't tell what the `providerKeys`'s purpose is. I have to read the `andWhere` SQL to understand it.
What do you think about calling the property `filterByProviderKeys`?

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -179,11 +182,19 @@ export class SecretsProvidersConnectionsService {
 	}
 
 	async getGlobalCompletions(): Promise<SecretsProviderConnection[]> {
-		return await this.repository.findGlobalConnections();
+		const connectedProviderKeys = this.providerRegistry.getConnectedNames();
+
+		return await this.repository.findGlobalConnections({
```
konstantintieber, line 187: Thinking out loud: if we'd store the information on whether a `SecretsProvider` is global or project-scoped in a new `SecretsProvider.isGlobal` property, we could get by without the extra SQL query, but get all the providerKeys we need here from the registry, right?

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
alielkhateeb: updated here 47113e4969b8ff9bf0a1b664cece5bc86e94d4dd

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
alielkhateeb: I have named the param to filters for clarity of the signature 47113e4969b8ff9bf0a1b664cece5bc86e94d4dd
That way it's clearer and we can extend this in the future when needed.

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
alielkhateeb: I agree. It is not obvious to me why we have two separate endpoints for global and project-scoped completions in the first place. Right now, we need to call two APIs just to determine the full set of completions, but as far as I understand, global connections should always be included.

If that’s the case, we could potentially refactor the API design to unify this into a single endpoint, which would also simplify things on the backend and almost surely result in a single SQL query.

However, within the scope of "fixing this bug" I would aim for that to be addressed in a separate ticket/PR, what do you think?

packages/@n8n/db/src/repositories/secrets-provider-connection.repository.ee.ts
alielkhateeb: Also found an edge case, where if you had only unsuccessful connections your completion endpoints would return all secret connections 💀 haha
Wrote a test for this case too and added an early return for when we filter by an empty array (Same commit above)

konstantintieber (dismissed):
Nice one

konstantintieber (approved):

PR #25930, opened 2026-02-18
fix(Webhook Node): Redact auth credential headers from Webhook execution output

guillaumejacquart:
## Summary
- When a Webhook node has authentication configured (Basic/Bearer/Header/JWT Auth), the auth headers in incoming requests were exposed in plain text in execution output data
- After auth validation succeeds, auth-related headers are now replaced with `**hidden**` before output is constructed
- Extracts `REDACTED` constant to `n8n-workflow` for shared use

### Screenshots
Before:
image

After:
image

It's the same when opening a node in the execution, the **hidden** value is shown the same way

We use the same **hidden** redacted value as for the http request node 

## Related ticket
https://linear.app/n8n/issue/IAM-278

## Review / Merge checklist
- [X] Tests included. 

### DIFF
--- a/packages/workflow/src/constants.ts
+++ b/packages/workflow/src/constants.ts
@@ -138,3 +138,6 @@ export const BINARY_IN_JSON_PROPERTY = '_files';
 
 export const BINARY_MODE_SEPARATE = 'separate';
 export const BINARY_MODE_COMBINED = 'combined';
+
+/** Replacement value for redacted sensitive data in execution output. */
+export const REDACTED = '**hidden**';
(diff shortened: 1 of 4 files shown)


packages/workflow/src/constants.ts
```diff
@@ -138,3 +138,6 @@ export const BINARY_IN_JSON_PROPERTY = '_files';
 
 export const BINARY_MODE_SEPARATE = 'separate';
 export const BINARY_MODE_COMBINED = 'combined';
+
+/** Replacement value for redacted sensitive data in execution output. */
+export const REDACTED = '**hidden**';
```
phyllis-noester, line 143: should we maybe share a screenshot of what that looks like in a workflow with JC and jan for feedback?

guillaumejacquart: I uploaded before / after screenshots

PR #25935, opened 2026-02-18
feat: Only sync scopes / roles as leader in multi-main

BGZStephen:
## Summary

This PR ensures role and scope syncing only happens on the leader with a multi-main setup, avoiding race conditions when starting up and populating roles and scopes.

IAM-298

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -221,6 +223,12 @@ export class AuthRolesService {
 	}
 
 	async init() {
+		// Only sync on the leader instance to avoid race conditions in multi-main setups
```
phyllis-noester: did you have a look at this? https://github.com/n8n-io/n8n/pull/25245/changes

looks like we're trying to achieve something similar here. is it possible to combine that?

BGZStephen: We are essentially, the reasoning here is:

* AuthRoleService is only initialised on a `main` instance, which seems reasonable
* The Init function could contain things that should only happen on the leader (roles and scope syncing) but may also include actions that are needed on both leaders and none leaders (such as building a cache)

This approach brings this flexibility, whether something is main is a bit of a different concern to whether that main instance is the leader

phyllis-noester: but at the moment this is not the case no? I'm not sure the roles service needs to know the InstanceSettings

BGZStephen: Not currently, not, but it's flexible for the future

It seems to be a pattern accross the codebase too:

* packages/cli/src/wait-tracker.ts
* packages/cli/src/active-workflow-manager.ts
* packages/cli/src/modules/insights/insights.service.ts
* packages/cli/src/services/pruning/workflow-history-compaction.service.ts

I don't feel too strongly either way, I can move this check higher and it would work too, just be more specific.

phyllis-noester (dismissed):
looks good! did you test it in multi main setup? looks pretty straight forward but I'm worried that if there is an issue we'd only find out once it reaches a customer (internal is not multi main right?)

packages/cli/src/commands/base-command.ts
```diff
@@ -146,7 +146,11 @@ export abstract class BaseCommand<F = never> {
 		// Initialize the auth roles service to make sure that roles are correctly setup for the instance.
 		// Only run on main instance - workers should not modify auth roles/scopes as they may have
 		// different code versions, and scope sync would incorrectly delete scopes they don't know about.
-		if (this.instanceSettings.instanceType === 'main') {
+		// In multi-main setups, only sync on the leader instance to avoid race conditions.
+		if (
+			this.instanceSettings.instanceType === 'main' &&
+			(!this.instanceSettings.isMultiMain || this.instanceSettings.isLeader)
```
guillaumejacquart: If i'm not mistaken, the isLeader field is set here: https://github.com/n8n-io/n8n/blob/iam-278-do-not-log-authcredentials-in-workflows/packages/cli/src/commands/start.ts#L296
which is ran after the base-command init function, so the value of `this.instanceSettings.isLeader` is not reliable I think (it's probably gonna be false or true for both instance)

guillaumejacquart (commented):
I'm not sure this work (see my comment)

packages/cli/src/commands/base-command.ts
BGZStephen: Move this to `start.ts` now instead to definitely await the leader setting.

packages/cli/src/commands/__tests__/start.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+		instanceSettings = mockInstance(InstanceSettings);
+		Container.set(InstanceSettings, instanceSettings);
+	});
+
+	/**
+	 * Helper function to simulate the conditional logic from start.ts
+	 * This mirrors the exact logic in start.ts lines 233-238
+	 */
+	async function simulateStartInitLogic() {
```
afitzek: These tests are not testing the actual implementation, but a logic that is disconnected from the actual implementation

BGZStephen: Should be better implemented now

afitzek (approved):

PR #25940, opened 2026-02-18
fix: Remove redundant builder-log check causing responder double-answering

aalises:
## Summary

- Fix responder double-answering after assistant in multi-turn `mergeAskBuild` sessions
- Remove redundant `!hasBuilderPhaseInLog(state.coordinationLog)` check from the responder short-circuit condition
- The `lastCompletedPhase === 'assistant'` check already guarantees builder didn't run after assistant in the current turn

### Root Cause

The `coordinationLog` uses an append-only reducer and persists across turns via the checkpointer. After a build turn (Turn 1), builder entries remain in the log. In Turn 2 (ask), `hasBuilderPhaseInLog` returns `true` on stale entries, causing the short-circuit to fail and the responder LLM to generate a second answer.


mike12345567 (approved):
LGTM - thanks for adding the turn information!

PR #25951, opened 2026-02-18
fix(editor): Prevent workflow-updated emission on assistant calls

aalises:
## Summary

When `MERGE_ASK_BUILD` is enabled, all chat messages route through the builder path, including assistant (help/debug) queries. This caused the `create_workflow_name` preprocessing step to emit `workflow-updated` messages on assistant calls, prematurely saving and naming pristine workflows.

### Root cause

`create_workflow_name` ran **before** supervisor routing, so every first message triggered name generation and a `workflow-updated` emission, regardless of whether it was headed to the builder or assistant subgraph.

### Fix

Move `create_workflow_name` from preprocessing to **after** supervisor routing. It now only runs on builder/discovery paths:

```
Before: check_state → create_workflow_name → supervisor → (builder | assistant)
After:  check_state → supervisor → create_workflow_name → (builder | discovery)
                                 → assistant  (no name generation)
```

## Related Linear tickets

https://linear.app/n8n/issue/AI-2066


mike12345567 (dismissed):
LGTM!

mike12345567 (approved):

PR #25959, opened 2026-02-18
fix(core): Always serve /healthz on task broker regardless of N8N_ENDPOINT_HEALTH

horiyee:
## Summary

This is a regression introduced by #25729, which made the health endpoint configurable across all servers including the task broker. The `task-runner-launcher` binary, however, hardcodes `/healthz` for the broker readiness check. When users set `N8N_ENDPOINT_HEALTH=health` — required on platforms such as Google Cloud Run that reserve `/healthz` at the ingress — the broker began serving health at `/health` while the launcher kept checking `/healthz`, receiving 404 and waiting forever. Code nodes would time out after 60 seconds.

The task broker runs on an internal port (5679) that is never publicly exposed, so `N8N_ENDPOINT_HEALTH` should not affect it. This fix hardcodes `/healthz` on the task broker, decoupling it from the external health endpoint configuration.

**Root cause:**
- Broker uses `globalConfig.endpoints.health`: [task-broker-server.ts#L173–L174](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/task-runners/task-broker/task-broker-server.ts#L173-L174)
- Launcher hardcodes `/healthz`: [check_until_broker_ready.go#L12](https://github.com/n8n-io/task-runner-launcher/blob/main/internal/http/check_until_broker_ready.go#L12)

Closes #25958

## Testing

Added a unit test to `task-broker-server.test.ts` that verifies `/healthz` returns 200 even when `N8N_ENDPOINT_HEALTH` is overridden (e.g. `endpoints: { health: '/health' }` in `GlobalConfig`).

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/task-runners/task-broker/__tests__/task-broker-server.test.ts
+++ b/packages/cli/src/task-runners/task-broker/__tests__/task-broker-server.test.ts
@@ -1,6 +1,7 @@
 import type { GlobalConfig } from '@n8n/config';
 import { mock } from 'jest-mock-extended';
 import { ServerResponse } from 'node:http';
+import request from 'supertest';
 import type WebSocket from 'ws';
 
 import type { TaskBrokerAuthController } from '@/task-runners/task-broker/auth/task-broker-auth.controller';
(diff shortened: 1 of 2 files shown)


shortstacked (approved):
Thanks for the PR.

PR #25960, opened 2026-02-18
fix(youtube): respect incoming binary mimeType for media uploads

cluster2600:
## What
Fixes the YouTube Upload node to use the incoming binary `mimeType` (e.g. `video/mp4`) instead of defaulting to `application/mp4`.

## Why
On recent n8n versions, uploads can fail because the YouTube API rejects `application/mp4` for MP4 media.

## How
- Prefer the incoming binary `mimeType` over stored metadata when preparing resumable uploads.
- Fall back to mime type detection from file extension when `mimeType` is missing.
- Adds a regression test to ensure binary `mimeType` overrides metadata.

## Testing
- Local run: `pnpm -r lint` and relevant test suite for `nodes-base` (plus targeted unit test).

## Checklist
- [x] Uses existing n8n request helpers
- [x] No breaking changes
- [x] Regression test added


cluster2600:
Triggered CLA recheck (https://cla-assistant.io/check/n8n-io/n8n?pullRequest=25960). It still shows pending; likely needs the commit email (maxime.grenu@gmail.com) added/verified on the GitHub account and the CLA signed.

PR #25961, opened 2026-02-18
fix(editor): Custom sticky text color

RicardoE105:
## Summary

Custom hex colors on sticky notes now adapt between light and dark mode, matching preset color
  behavior
  - At render time, the user's chosen hex is decomposed into hue + saturation via HSL, then lightness
  is set per theme (L:90%/75% light, L:15%/25% dark for background/border)
  - Text color auto-adjusts for contrast using WCAG 2.0 relative luminance, ensuring readability in
  both themes
  - The user's stored hex is unchanged (no data model change) — the normalization is purely visual at
  render time

  How it works

  The 7 preset sticky colors already use completely different lightness levels per theme (e.g.
  yellow-100 in light vs yellow-900 in dark) while keeping hue/saturation constant. Custom colors now
  follow the same pattern: the user's color choice effectively becomes a hue + saturation selection.

https://n8nio.slack.com/archives/C0848MV1RJL/p1771407402036599?thread_ts=1770302582.639859&cid=C0848MV1RJL

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/@n8n/design-system/src/components/N8nSticky/Sticky.vue
```diff
@@ -43,15 +43,16 @@ const getCustomColorStyles = (hexColor: string) => {
 		return {};
 	}
 
-	// Create lighter border for dark mode (80% lighter)
-	const lighterBorder = adjustColorLightness(hexColor, 80);
```
MiloradFilipovic, line 47: Can we remove `adjustColorLightness` method? If I see it correctly, it's not being used anymore after this PR.

packages/frontend/@n8n/design-system/src/components/N8nSticky/Sticky.vue
```diff
@@ -158,21 +159,27 @@ const onInputScroll = (event: WheelEvent) => {
...
+	--color--text--shade-1: var(--sticky--color--text--custom-dark);
 }
 
-// System dark mode (when theme is 'system')
 @media (prefers-color-scheme: dark) {
 	:global(body:not([data-theme='light'])) .customColor {
+		--sticky--color--background: var(--sticky--color--background--custom-dark);
 		--sticky--border-color: var(--sticky--border-color--custom-dark);
+		--sticky--color--text: var(--sticky--color--text--custom-dark);
+		--color--text--shade-1: var(--sticky--color--text--custom-dark);
```
MiloradFilipovic: We are overriding global variable here. Is this intentional?

MiloradFilipovic (commented):
Looks good overall. I left a couple of comments so take a look if they make sense.
Also, I know this has been trough design and product review but it bothers me that the picker popup no longer shows true background value. Here, I picked full white (255,255,255) and that is not what's applied to my sticky.

Image

I get it that it's a compromise so feel free to ignore

packages/frontend/@n8n/design-system/src/utils/colorUtils.ts
```diff
@@ -10,6 +29,136 @@ export function isValidHexColor(color: string): boolean {
...
+
+/**
+ * Normalizes a custom hex color for theme-appropriate display.
+ * Extracts hue and saturation from the input, then sets lightness to match
+ * the preset sticky note pattern (light backgrounds in light mode, dark in dark mode).
+ * @param hex - User's chosen hex color (#RRGGBB)
+ * @param isDark - Whether dark theme is active
+ * @returns Object with normalized background, border, and text hex colors
+ */
+export function normalizeCustomColorForTheme(
```
MiloradFilipovic: Is this one being used anywhere now? Should we delete it?

MiloradFilipovic (commented):
Looking good, thanks for addressing the feedback.
I left some comments on the dead code and there are some from the previous review so take a look

packages/frontend/@n8n/design-system/src/components/N8nSticky/Sticky.vue
RicardoE105: Yes, the override is intentional. The N8nInput textarea hardcodes color: var(--color--text--shade-1) in its CSS (Input.vue:546). There's no prop to pass a custom text color. So the only way to make the textarea text readable against a custom sticky background is to override --color--text--shade-1 on the sticky's root element.

RicardoE105: instead we could:

  2. Add a textColor prop to N8nInput. Cleaner but a bigger change to a shared component just for this use case.
  3. Override via unscoped CSS targeting the textarea directly. e.g. .sticky-textarea textarea { color: var(--sticky--color--text--custom) } in the unscoped <style> block. This avoids overriding the global variable name.

MiloradFilipovic (approved):
🛳️

PR #25970, opened 2026-02-19
feat: Quick connect integration flow for Firecrawl

dlavrenuek:
## Summary

Implements the Quick connect integration with Firecrawl

## To test

Run n8n with following env variable

```
export N8N_QUICK_CONNECT_OPTIONS='[{"packageName": "@mendable/n8n-nodes-firecrawl", "credentialType": "firecrawlApi", "consentText":"Confirm sending data to external service", "serviceName":"Firecrawl", "text": "Firecrawl!", "quickConnectType":"firecrawl", "backendFlowConfig":{"secret":"--secret--"}}]'
```

Activate quick connect feature flag locally via devtools console

```
window.featureFlags.override('073_quick_connect', 'variant')
```

Add the Firecrawl node to a workflow. Quick connect button is only displayed when there are no existing credentials

- https://linear.app/n8n/issue/NODE-4276/build-firecrawl-integration-flow
- https://linear.app/n8n/issue/NODE-4405/implement-consent-modal-and-call-the-api-to-create-the-credential-from

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/quick-connect/quick-connect.service.ts
```diff
@@ -1,108 +1,59 @@
...
-		});
-
-		return decryptedCredential;
 	}
 
-	private isValidBackendConfig(config: QuickConnectOption): config is QuickConnectBackendOption {
-		return (
-			typeof config.backendFlowConfig?.secret === 'string' && config.backendFlowConfig.secret !== ''
-		);
+	async registerHandlers() {
```
dlavrenuek, line 50: The initial implementation of `registerHandler` in the module did load all of the handlers even if quick connect was not enabled for these. Now only the handlers are loaded (and added to memory) which are actually configured/enabled

packages/cli/src/modules/quick-connect/quick-connect.service.ts
```diff
@@ -1,108 +1,59 @@
...
-	QuickConnectOption,
-} from './quick-connect.config';
-import { QuickConnectError } from './quick-connect.errors';
+const backendHandlers = {
+	firecrawl: async () => (await import('./handlers/firecrawl.handler')).FirecrawlHandler,
+};
 
 @Service()
 export class QuickConnectService {
+	private readonly handlers = new Map<string, IQuickConnectHandler>();
```
dlavrenuek, line 20: I did simplify the implementation and scope of this service a lot:
- handler registry service was technically equal to a simple map and was only used in this single service (also without an outlook to be used in other places). Removed the handler registry (as service)
- removed the credential creation here, only the service api key is generated here. Duplicating the business logic for credential creation does add potential for errors
- validation is now mostly declarative and not imperative (only one if condition if the handler for given type exists). The handler configuration is passed to handler on creation, so there is no need to search for the right option configuration on runtime

packages/cli/src/modules/quick-connect/handlers/handler.interface.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Interface for quick connect handlers.
+ * Handlers can be frontend-based or backend-based.
+ */
+export interface IQuickConnectHandler<ConfigOption = QuickConnectOption> {
+	/**
+	 * Sets provider configuration
+	 * @param config - Configured quick connect option
+	 */
+	setConfig(config: ConfigOption): void;
```
dlavrenuek, line 14: Ideally the handler would get the config in the constructor, but this is such a pain to type it with TS

packages/cli/src/modules/quick-connect/quick-connect.module.ts
```diff
@@ -27,9 +27,8 @@ export class QuickConnectModule implements ModuleInterface {
 	}
 
 	private async registerHandlers() {
-		const { QuickConnectHandlerRegistry } = await import('./handlers/quick-connect.handler');
-		const { FirecrawlHandler } = await import('./handlers/firecrawl.handler');
-		const registry = Container.get(QuickConnectHandlerRegistry);
-		registry.register(Container.get(FirecrawlHandler));
+		const { QuickConnectService } = await import('./quick-connect.service');
+		const quickConnectService = Container.get(QuickConnectService);
+		await quickConnectService.registerHandlers();
```
dlavrenuek, line 32: Moved handler registration to the quick connect service. It should not be in scope of the module (definition) to register handlers in my opinion

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
```diff
@@ -91,10 +83,32 @@ export function useQuickConnect() {
...
+				onCancel: reject,
+				integrationId: String(quickConnectOption.config.integrationId),
+			});
+
+			popup.open();
+		});
+	}
+
+	async function connectViaBackendFlow(quickConnectOption: QuickConnectOption) {
+		const { apiKey } = await getQuickConnectApiKey(rootStore.restApiContext, quickConnectOption);
```
dlavrenuek: In most of the code I see the pattern to call api calls through "stores'", but from my experience this approach is pretty outdated and is an anti-pattern nowadays

dlavrenuek: It was common ± 10 years ago due to react components and rendering exceptions in combination with async calls (exception when trying to access `this.state` after component unmount, f.e. when navigating to another route while the current one does an api request). Nowadays there is no technically need for it. Also there is no need for a strict store to api binding (for example we don't need a quick connect store?)

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
```diff
@@ -14,20 +13,21 @@ interface FirecrawlCreateUserResponse {
...
 		const response = await axios.post<FirecrawlCreateUserResponse>(
 			`${FIRECRAWL_API_BASE_URL}/admin/integration/create-user`,
-			{ email: user.email },
+			{ email },
 			{
 				headers: {
-					Authorization: `Bearer ${secret}`,
-					'Content-Type': 'application/json',
+					authorization: `Bearer ${secret}`,
+					contentType: 'application/json',
```
elsmr: Does this work? shouldn't it be `content-type`?

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
```diff
@@ -14,20 +13,21 @@ interface FirecrawlCreateUserResponse {
...
+		this.config = config;
+	}
 
-	async getCredentialData(
-		config: QuickConnectBackendOption,
-		user: User,
-	): Promise<ICredentialDataDecryptedObject> {
-		const secret = config.backendFlowConfig.secret;
+	async getApiKey({ email }: User) {
+		const secret = this.config.backendFlowConfig.secret;
```
elsmr: Should we check that this.config exists here?

packages/cli/src/modules/quick-connect/quick-connect.controller.ts
```diff
@@ -8,16 +8,12 @@ import { QuickConnectService } from './quick-connect.service';
 export class QuickConnectController {
 	constructor(private readonly quickConnectService: QuickConnectService) {}
 
-	@Post('/create')
-	async createCredential(
+	@Post('/api-key')
```
elsmr: I liked the old naming better because it doesn't limit us to only API keys. Could also be on `/` instead.

The response of the endpoint is `data: { apiKey: '***' }` so it's already compatible with sending other credential fields.

packages/cli/src/modules/quick-connect/quick-connect.service.ts
```diff
@@ -1,108 +1,59 @@
...
-			typeof config.backendFlowConfig?.secret === 'string' && config.backendFlowConfig.secret !== ''
-		);
+	async registerHandlers() {
+		for (const option of this.quickConnectConfig.options) {
+			const { quickConnectType } = option;
+			if (quickConnectType in backendHandlers) {
+				const Handler = await backendHandlers[quickConnectType as keyof typeof backendHandlers]();
+				const handler = Container.get(Handler);
+				// @ts-expect-error the configuration option type is not reduced to specific subtype
+				handler.setConfig(option);
```
elsmr: Low prio, but could we do a typeguard here instead?

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -530,7 +536,10 @@ async function onClickCreateCredential(type: ICredentialType | INodeCredentialDe
 
 function getServiceName(credentialTypeName: string): string {
 	const displayName = credentialTypeNames.value[credentialTypeName] ?? credentialTypeName;
-	return displayName.replace(/\s+OAuth2?\s+API$/i, '').replace(/\s+API$/i, '');
+	return displayName
+		.replace(/\s+OAuth2?\s+API$/i, '')
```
elsmr: There is an existing helper `getAppNameFromCredType`, I can also update it in my PR. I assume this one will go in first.

packages/frontend/editor-ui/src/features/credentials/quickConnect/composables/useQuickConnect.ts
elsmr: Here I would use all of `data` and merge it into the credential instead of picking the `apiKey` field

elsmr (changes requested):
Tested it and it works as expected!

I noticed the consent modal is missing some UI from the design:

Image

For me it's also fine if we update this in another PR.

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
dlavrenuek: There was a linting warning for the header name `Content-Type`, so I did change it assuming that axios will interpolate the name to proper header name. It does work, but this assumption might actually be wrong, will change it back.

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
dlavrenuek: The config is always set during handler registration and it's ensured by zod schema that  these match. But since the config is set separately and not in the constructor the type of config should actually include undefined. Added it and also a simple guard `this.config!.`, this will throw an exception if unset resulting in 500 http error. Imo 500 is also the correct error one since it's an unexpected server error, I would like not to add explicit error handling because it will not provide any additional value from my perspective.

packages/cli/src/modules/quick-connect/handlers/firecrawl.handler.ts
dlavrenuek: fixed

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
dlavrenuek: fixed

packages/cli/src/modules/quick-connect/quick-connect.service.ts
dlavrenuek: I did add a proper typecast

packages/cli/src/modules/quick-connect/quick-connect.controller.ts
dlavrenuek: good idea. I have also renamed this method and the handler methods to `getCredentialData` to correctly reflect what is returned

elsmr (approved):
🚢

PR #25973, opened 2026-02-19
feat(editor): Enable users to pick between managed and custom OAuth credentials

elsmr:
## Summary

- Enable users to pick between managed and custom OAuth credentials
- Improve credential modal design
  - New credential mode selector replacing the "Connect with" radio buttons
  - New docs callout
  - Save button changed
  - Added quick connect button (when enabled)
  - Design fixes
  - Save pending form changes when switching between auth modes
- Bug fixes
  - Show all available credentials in dropdown (not just selected `authentication`)
  - Don't change `authentication` on node until credential is saved.
  - Always use light theme icon on the quick connect button (enough contrast on white bg)

image

image

image

### Testing

- Creating and editing credentials with/without managed credentials (`CREDENTIALS_OVERWRITE_DATA='{"boxOAuth2Api":{"clientId":"test","clientSecret":"test"}}'`)
- Test with quick connect enabled (`N8N_QUICK_CONNECT_OPTIONS='[{"packageName":"@mendable/n8n-nodes-firecrawl","credentialType":"firecrawlApi","text":"Firecrawl offers a free Basic plan to n8n users","quickConnectType":"api","serviceName":"Firecrawl"}]'`)
- Test nodes with multiple available credential types
- Test from the credentials list page

https://linear.app/n8n/issue/NODE-4281/enable-users-to-pick-between-managed-cred-and-custom-oauth-on-cloud

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialConfig.vue
```diff
@@ -202,6 +205,33 @@ const assistantAlreadyAsked = computed<boolean>(() => {
...
+	const nodeType = ndvStore.activeNode?.type;
+
+	if (!nodeType) return appName.value;
+
+	const quickConnectOption = getQuickConnectOption(
+		credentialTypeName.value,
+		ndvStore.activeNode?.type,
+	);
+
+	if (quickConnectOption?.serviceName) {
```
dlavrenuek: In https://github.com/n8n-io/n8n/pull/25970/changes#diff-1ac438874231c37925bf920d1607b13e9c46bf36cc3c2a598b8c6a5624eb3ce6L6 i'm removing the `serviceName` from quick connect configuration because it is only needed in the frontend, where we already have the node name or app name. If you don't see a problem with that change then usage of `serviceName` should also be removed here

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -592,11 +615,50 @@ function onShareWithAllUsersUpdate(shareWithAllUsers: boolean) {
...
 
+function getCurrentModeKey(): string {
+	const base = credentialTypeName.value ?? '';
+	if (managedOAuthAvailable.value) {
+		return `${base}:${useCustomOAuth.value ? 'custom' : 'managed'}`;
+	}
+	return base;
+}
+
+function cacheCurrentData(): void {
```
dlavrenuek, line 614: When you have configured f.e. google docs (with managed credentials enabled) and created a "Service Account". Then you change the credential type to a different type (like managed oauth or custom oauth) and then back the data (service account) in the form is lost. `cacheCurrentData()` is called multiple times in this case (from `onManagedOAuthModeChange` and from `onAuthTypeChanged`). After the first call the generated `getCurrentModeKey()` will be wrong because it either contains the wrong credential type or wrong `custom`/`managed` mode

Image

dlavrenuek: I suggest firing only one event when changed and pass the new type and if it's managed or not as two parameters to `onAuthTypeChanged(type: string, isCustom: boolean)`

dlavrenuek (changes requested):
review so far

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -116,6 +117,13 @@ const requiredCredentials = ref(false); // Are credentials required or optional
 const contentRef = ref<HTMLDivElement>();
 const isSharedGlobally = ref(false);
 const isResolvable = ref(false);
+const useCustomOAuth = ref(false);
+const pendingAuthType = ref<string | null>(null);
+const credentialDataCacheKey = `credential_data_cache_${props.mode === 'new' ? 'new' : props.activeId}`;
+const credentialDataCache = useSessionStorage<Record<string, ICredentialDataDecryptedObject>>(
```
dlavrenuek: I was thinking about using the session storage as cache and it does not feel right. This way credentials are stored in the session store but these are never cleaned unless you close the window, even if you log out. Also there should be no need to use the session store, this data can simply be stored in a variable inside this component or is there a reason to use the session store?
If there is a reason, than an alternative would be to use a normal non-persistant store

dlavrenuek (changes requested):

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
elsmr: My idea was that you could leave the modal and come back. But maybe simply in component memory is better, the main point is to remember data when switching between auth types. When you leave the credential modal there is a warning modal anyway.

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -1207,21 +1179,28 @@ async function oAuthCredentialAuthorize() {
...
 }
 
-async function onAuthTypeChanged(type: string): Promise<void> {
+async function onAuthTypeChanged(payload: {
+	type: string;
+	useCustomOauth?: boolean;
+}): Promise<void> {
+	cacheCurrentData();
+	authError.value = '';
+	useCustomOAuth.value = payload.useCustomOauth ?? false;
```
dlavrenuek, line 1188: No need for a change here, just a tip on how you can eliminate the need for undefined check here by using default parameter value with destructuring:

```
async function onAuthTypeChanged({
	type,
	useCustomOauth = false,
}: {
	type: string;
	useCustomOauth?: boolean;
})
```

dlavrenuek (dismissed):
Great improvements 🚀

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialModeSelector.vue
```diff
@@ -217,6 +217,7 @@ function isSelected(option: Option['value']): boolean {
 }
 
 .dropdownContent {
+	width: auto;
```
dlavrenuek, line 220: 🎉

dlavrenuek (approved):

PR #25992, opened 2026-02-19
feat: Add assignments tab for custom roles with project visibility

cstuncsik:
## Summary

Adds a new **Assignments** tab to project roles showing where each role is assigned across projects. Admins can now see project membership details and reassign users before deleting a role. Improves the deletion flow with context-aware dialogs: roles in use show "Go to assignments", while unused roles show simple delete confirmation.

## Changes

- New `RoleAssignmentsTab` component lists projects with member counts and last assigned dates
- New `RoleProjectMembersModal` shows all members in a project with their role labels
- Backend endpoints: `GET /roles/{slug}/assignments` and `GET /roles/{slug}/assignments/{projectId}/members`
- Roles list now shows "projects assigned" (instead of users) with clickable navigation to assignments tab
- Danger zone updated with better copy and link to assignments
- Delete action from 3-dot menu now context-aware

https://linear.app/n8n/issue/IAM-222

Closes #25952

## Review / Merge checklist

- [x] Tests included (42 existing tests pass)
- [x] Lint and typecheck pass


packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -2441,6 +2443,19 @@
...
+	"projectRoles.sourceControl.table.projectsAssigned": "Projects assigned",
+	"projectRoles.assignments.emptyState": "This role is not assigned in any projects yet.",
+	"projectRoles.assignments.projectColumn": "Project",
+	"projectRoles.assignments.membersColumn": "Members assigned",
+	"projectRoles.assignments.lastAssignedColumn": "Last assigned",
+	"projectRoles.assignments.membersModal.title": "\"{projectName}\" members",
+	"projectRoles.assignments.membersModal.manageMembers": "Manage members",
+	"projectRoles.assignments.membersModal.cancel": "Cancel",
+	"projectRoles.action.delete.inUse.title": "{roleName}",
+	"projectRoles.action.delete.inUse.text": "You can't delete this role while it's assigned in {count} project. To delete it, first reassign users within projects to a different role. | You can't delete this role while it's assigned in {count} projects. To delete it, first reassign users within projects to a different role.",
```
phyllis-noester: why do we duplicate the text here?

phyllis-noester (dismissed):

packages/@n8n/api-types/src/dto/roles/role-assignments-response.dto.ts
```diff
@@ -0,0 +1,12 @@
+export interface RoleProjectAssignment {
```
afitzek: Is there a reason why we don't use zod schema for this, I think we use zod schemas for all other dtos? I think we try to use zod schema for all dtos, so that we can validate the response objects against these and strip additional properties, that should not go out to the FE for example. Also if we want to describe the API using openapi for example zod schema help to understand the types better.

packages/cli/src/controllers/role.controller.ts
```diff
@@ -36,6 +37,27 @@ export class RoleController {
...
 
+	@Get('/:slug/assignments/:projectId/members')
+	@GlobalScope('role:manage')
+	async getRoleProjectMembers(
+		_req: AuthenticatedRequest,
+		_res: Response,
+		@Param('slug') slug: string,
+		@Param('projectId') projectId: string,
+	): Promise<RoleProjectMembersResponse> {
+		return await this.roleService.getRoleProjectMembers(slug, projectId);
```
afitzek: Here we should ideally parse the results with the DTO schema. 🤔

packages/@n8n/db/src/repositories/role.repository.ts
```diff
@@ -60,6 +60,98 @@ export class RoleRepository extends Repository<Role> {
 		);
 	}
 
+	async findAllProjectCounts(): Promise<Record<string, number>> {
```
afitzek, line 63: Are the new functions in the file tested enough with the integration tests? 🤔

packages/cli/src/services/role.service.ts
```diff
@@ -1,3 +1,4 @@
+import type { RoleAssignmentsResponse, RoleProjectMembersResponse } from '@n8n/api-types';
```
afitzek, line 1: Are the changes in this file tested enough based on the integration test? 🤔

afitzek (commented):
I think we should use zod schemas for the DTOs and add tests for the repository and service changes.

packages/@n8n/api-types/src/dto/roles/role-assignments-response.dto.ts
```diff
@@ -0,0 +1,25 @@
...
+	projectId: z.string(),
+	projectName: z.string(),
+	projectIcon: z
+		.object({
+			type: z.string(),
+			value: z.string(),
+		})
+		.nullable(),
+	memberCount: z.number(),
+	lastAssigned: z.string().nullable(),
```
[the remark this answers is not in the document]
cstuncsik: lastAssigned comes from a SQL MAX(pr.createdAt) query — the database returns null when there are no matching rows (not undefined). 

Using .optional() would be wrong — the field is always present in the response, it’s just null when no assignment exists. This is the correct pattern for SQL aggregate results. The projectIcon field follows the same nullable() pattern for the same reason.

afitzek (approved):
Thanks for addressing the comments. 👍

PR #25998, opened 2026-02-19
fix(editor): Fetch tags when returning published wf version

nikhilkuria:
## Summary
Fixes a bug where workflow tags were being removed when publishing a workflow.

Root cause: The activateWorkflow method was not including tags in the relations when fetching the updated workflow after activation, causing the response to return tags: undefined which overwrote the frontend's cached tags.

Fix: Include tags in the relations query

https://linear.app/n8n/issue/ADO-4847/community-issue-bug-workflow-tags-are-removed-when-publishing-a

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/workflows/workflow.service.ts
+++ b/packages/cli/src/workflows/workflow.service.ts
@@ -705,9 +705,11 @@ export class WorkflowService {
 		}
 
 		// Fetch workflow again with workflowPublishHistory after activation to include the new entry
+		const tagsDisabled = this.globalConfig.tags.disabled;
 		const updatedWorkflow = await this.workflowRepository.findOne({
 			where: { id: workflowId },
 			relations: {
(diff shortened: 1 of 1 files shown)


packages/cli/src/workflows/workflow.service.ts
```diff
@@ -705,9 +705,11 @@ export class WorkflowService {
 		}
 
 		// Fetch workflow again with workflowPublishHistory after activation to include the new entry
+		const tagsDisabled = this.globalConfig.tags.disabled;
 		const updatedWorkflow = await this.workflowRepository.findOne({
 			where: { id: workflowId },
 			relations: {
+				tags: !tagsDisabled,
```
dariacodes, line 712: We do not need this relation here as activation shouldn't change the tags

PR #26001, opened 2026-02-19
feat: Migrate external secrets settings to entity-based storage (no-changelog)

ireneea:
## Summary

Adds a database migration that transfers existing external secrets provider settings from the legacy encrypted JSON blob stored in the `settings` table (`feature.externalSecrets`) to the new `secrets_provider_connection` table used by entity-based storage.

**How to test:**
1. Set `N8N_ENV_FEAT_EXTERNAL_SECRETS_MULTIPLE_CONNECTIONS` to true
1. Start n8n with an existing database that has external secrets configured (blob present in `settings` table under `feature.externalSecrets`)
2. Run migrations, verify providers appear in `secrets_provider_connection`
3. Confirm external secrets continue to work after migration

https://linear.app/n8n/issue/LIGO-112

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
+++ b/packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
@@ -70,6 +70,11 @@ export class MigrateExternalSecretsToEntityStorage1771500000000 implements Irrev
 		for (const providerName of providerNames) {
 			const providerData = allSettings[providerName];
 
+			if (!providerData.connected) {
+				logger.info(`[${migrationName}] Provider "${providerName}" is not connected, skipping`);
+				continue;
+			}
+
 			const existing: Array<{ providerKey: string }> = await runQuery(
 				`SELECT ${providerKeyCol} FROM ${connectionTable} WHERE ${providerKeyCol} = :providerKey;`,
 				{ providerKey: providerName },
--- a/packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
+++ b/packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
@@ -70,7 +70,6 @@ export class MigrateExternalSecretsToEntityStorage1771500000000 implements Irrev
 		for (const providerName of providerNames) {
 			const providerData = allSettings[providerName];
 
-			// Idempotency check: skip if provider already exists
 			const existing: Array<{ providerKey: string }> = await runQuery(
 				`SELECT ${providerKeyCol} FROM ${connectionTable} WHERE ${providerKeyCol} = :providerKey;`,
 				{ providerKey: providerName },
@@ -83,12 +82,8 @@ export class MigrateExternalSecretsToEntityStorage1771500000000 implements Irrev
 				continue;
 			}
 
-			// Re-encrypt just the settings object for the new entity
 			const encryptedSettings = this.cipher.encrypt(providerData.settings ?? {});
 
-			// Insert row: providerKey = type = providerName (per ticket spec)
-			// isEnabled, createdAt, updatedAt have DB defaults and are omitted
-			// No project_secrets_provider_access rows = global provider
 			await runQuery(
 				`INSERT INTO ${connectionTable} (${providerKeyCol}, ${typeCol}, ${encryptedSettingsCol}) VALUES (:providerKey, :type, :encryptedSettings);`,
 				{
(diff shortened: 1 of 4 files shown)


packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const EXTERNAL_SECRETS_DB_KEY = 'feature.externalSecrets';
+
+/**
+ * Migrates external secrets provider settings from a single encrypted JSON blob
+ * in the `settings` table to individual entity rows in the `secrets_provider_connection` table.
+ *
+ * Existing providers are migrated as global providers (no project access rows).
+ * The old settings row is preserved for backward compatibility.
```
ireneea, line 24: Hey team, I was a bit conflicted about keeping or removing the setting entry, in the end I kept it but this means that we'll need another migration to remove it. 

Let me know what you think

packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
```diff
@@ -0,0 +1,105 @@
...
+		const rows: Array<{ value: string }> = await runQuery(
+			`SELECT ${valueCol} FROM ${settingsTable} WHERE ${keyCol} = '${EXTERNAL_SECRETS_DB_KEY}';`,
+		);
+
+		if (rows.length === 0) {
+			logger.info(`[${migrationName}] No external secrets settings found, skipping`);
+			return;
+		}
+
+		const [row] = rows;
```
alielkhateeb: For my own understanding, how did we previously store external secrets in the settings table?
Asking because I assumed a row per external secret provider but the code implies a single row for all external secrets?

packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
alielkhateeb: TLDR; I personally think it’s safer not to delete the old row immediately.

Since we have self-hosted customers, we can’t fully control or predict the state of their data. If the migration results in missing keys or inconsistencies, keeping the old row gives us a fallback and the ability to ship a recovery fix. If we delete it now, we lose that recovery path.

packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
ireneea: The settings of all external providers were stored as an encrypted json blob inside the settings table under the `feature.externalSecrets`

```json
{
  "awsSecretsManager": { ... },
  "gcpSecretsManager": { ...},
  ...
}
```

ireneea: That's also why it was fixed to only 5 providers

alielkhateeb: Ok makes sense, and since we select by the primary key then it's guaranteed that one single row is returned.
Everything makes sense now 😂 Thank you!

ireneea: yup exactly 👍

alielkhateeb (dismissed):
Tested the migration, works as expected 👏 I really like how clear the code is 🔥

packages/@n8n/db/src/migrations/common/1771500000000-MigrateExternalSecretsToEntityStorage.ts
```diff
@@ -70,6 +70,11 @@ export class MigrateExternalSecretsToEntityStorage1771500000000 implements Irrev
 		for (const providerName of providerNames) {
 			const providerData = allSettings[providerName];
 
+			if (!providerData.connected) {
+				logger.info(`[${migrationName}] Provider "${providerName}" is not connected, skipping`);
+				continue;
+			}
+
```
alielkhateeb: That is even better and makes so much sense to do 👏

alielkhateeb (dismissed):

konstantintieber (dismissed):
Code looks good. I did not test the migration manually.
Plugging Andi's integration test infra for migrations (I very much enjoyed using that [here](https://github.com/n8n-io/n8n/blob/master/packages/cli/test/migration/1766064542000-add-workflow-publish-scope-to-project-roles.test.ts)):
https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/backend-test-utils/MIGRATION_TESTING.md

tomi (dismissed):
LGTM 👍 

Couple minor comments:
- nit: It's okay to use functions/methods in migrations instead of writing one long method
- nit: Could also have tests for the migration

tomi (approved):
Thank you for addressing the feedback 🫶

PR #26007, opened 2026-02-19
fix(core): Handle tool wrapper nodes without supplyData in queue mode MCP Trigger

aalises:
## Summary

In queue mode, the invike tool method only supported tool nodes with a native `supplyData` method (e.g. `@n8n/n8n-nodes-langchain.toolHttpRequest`). 

Tool wrapper nodes like`httpRequestTool` and `dataTableTool`, which are normalr nodes marked `usableAsTool` that get their description modified at runtime via `convertNodeToAiTool()` don't have `supplyData`. They use their `execute` method instead.

This adds a path in `invokeTool()` that calls `execute` directly for these nodes when `supplyData` is not available but the node has `AiTool` output, mirroring the existing behavior in `get-input-connection-data.ts`.

**How to test:** Set up an MCP Server Trigger with an HTTP Request Tool node in queue mode. Tool calls should now succeed instead of failing with "does not have supplyData method".

https://linear.app/n8n/issue/AI-2074

### DIFF
--- a/packages/cli/src/scaling/job-processor.ts
+++ b/packages/cli/src/scaling/job-processor.ts
@@ -527,7 +527,13 @@ export class JobProcessor {
 
 				const result = await nodeType.execute.call(context as unknown as IExecuteFunctions);
 
-				return result?.[0]?.flatMap((item: INodeExecutionData) => item.json);
+				const response = result?.[0]?.flatMap((item: INodeExecutionData) => item.json);
+
+				context.addOutputData(NodeConnectionTypes.AiTool, 0, [
+					[{ json: { response } as INodeExecutionData['json'] }],
+				]);
+
+				return response;
 			}
 
 			throw new UnexpectedError(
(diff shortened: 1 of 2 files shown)


packages/cli/src/scaling/job-processor.ts
```diff
@@ -509,16 +509,30 @@ export class JobProcessor {
...
+					throw new UnexpectedError(`Tool node "${sourceNodeName}" did not return a valid Tool`);
+				}
 
-			if (!tool || typeof tool.invoke !== 'function') {
-				throw new UnexpectedError(`Tool node "${sourceNodeName}" did not return a valid Tool`);
+				return await tool.invoke(validatedToolArgs);
 			}
 
-			const result = await tool.invoke(validatedToolArgs);
+			if (nodeType.execute && nodeType.description.outputs.includes(NodeConnectionTypes.AiTool)) {
```
aalises, line 523: this is the path that we were not considering before

burivuhster (approved):
LGTM!

PR #26011, opened 2026-02-19
feat(core): Split share and unshare into separate scopes

afitzek:
## Summary

Splits the single `workflow:share` / `credential:share` scope into separate `share` and `unshare` scopes (`workflow:unshare`, `credential:unshare`). This ensures users can remove existing shares even when the personal space sharing setting (`security.personalSpaceSharing`) is disabled.

Previously, disabling personal space sharing removed the `share` scope from the `project:personalOwner` role, which also prevented unsharing existing resources. Now, the unshare scopes remain in the `project:personalOwner` base definition regardless of the sharing setting.

**Changes:**

- Add `workflow:unshare` and `credential:unshare` to the scope registry and all role definitions that have the corresponding `share` scope
- Add unshare scopes to `PERSONAL_PROJECT_OWNER_SCOPES` base definition (not behind the sharing setting toggle)
- Modify the `PUT /workflows/:id/share` endpoint to check `workflow:share` for adding shares and `workflow:unshare` for removing shares
- Modify the `PUT /credentials/:id/share` endpoint to check `credential:share` for adding shares and `credential:unshare` for removing shares
- Add database migration to grant unshare scopes to existing custom roles that have the corresponding share scope

**How to test:**
1. Create a custom role with `workflow:share` scope — verify it also gets `workflow:unshare` after migration
2. Disable personal space sharing setting — verify a personal project owner can still remove existing workflow/credential shares but cannot add new ones
3. Enable personal space sharing setting — verify both sharing and unsharing work

https://linear.app/n8n/issue/IAM-274/split-share-and-unshare-scope

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/db/src/migrations/common/1771500000001-AddUnshareScopeToCustomRoles.ts
```diff
@@ -0,0 +1,89 @@
...
+	async down({ escape, runQuery }: MigrationContext) {
+		const roleScopeTableName = escape.tableName('role_scope');
+		const roleScopeScopeSlugColumn = escape.columnName('scopeSlug');
+
+		const deleteQuery = `
+			DELETE FROM ${roleScopeTableName}
+			WHERE ${roleScopeScopeSlugColumn} = :unshareScope
+		`;
+
+		await runQuery(deleteQuery, { unshareScope: 'workflow:unshare' });
```
guillaumejacquart, line 86: Shouldn't we remove the scope from the scope table as well ?

packages/cli/src/credentials/credentials.controller.ts
```diff
@@ -334,29 +334,44 @@ export class CredentialsController {
 		const credential = await this.credentialsFinderService.findCredentialForUser(
 			credentialId,
 			req.user,
-			['credential:share'],
+			['credential:read'],
```
guillaumejacquart, line 337: Does this mean we should implement some kind of scope dependency ? What about if the user has share scope but not read scope ?

guillaumejacquart (commented):
Great work, looks good ! 
2 minor comments

Also, will the new scope be available in the custom role UI in a followup PR ?

packages/@n8n/db/src/migrations/common/1771500000001-AddUnshareScopeToCustomRoles.ts
afitzek: This will happen in the AuthRolesService on sync.

packages/cli/src/credentials/credentials.controller.ts
afitzek: Yes I think so, for selecting them, we should have a dependency on the read scope. I wanted to add this on the next PR that adds this to the custom scope selection.

guillaumejacquart (approved):
LGTM, thanks for replying to comments

PR #26040, opened 2026-02-20
fix(editor): Fix bug with tags not being sent to backend

dariacodes:
## Summary

[#25327](https://github.com/n8n-io/n8n/pull/25327) introduced the workflow document store for handling tags, but it did not establish it as the single source of truth. Some parts of the codebase continued to rely on `workflowStore.tags`, which led to bugs where tags were not saved correctly (e.g. in `getWorkflowDataToSave`).

This PR makes the workflow document store the single source of truth for tags and removes the remaining functionality for setting tags in the workflow store.

### How to test

This bug is very flaky. What worked for me in the end (from time to time):
- Open a workflow with no tags and a set-up that makes it possible to publish
- Add 2 tags
- Publish
- Refresh and see that tags were not saved
However, this behavior is not directly related to the publish functionality - it's rather about tags not being set in `getWorkflowDataToSave` which is called both for PATCH and `activate` requests.

Closes [ADO-4847](https://linear.app/n8n/issue/ADO-4847/community-issue-bug-workflow-tags-are-removed-when-publishing-a)

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowDetails.vue
```diff
@@ -174,7 +174,6 @@ function onTagsBlur() {
 	if (workflowDocumentStore?.value) {
 		workflowDocumentStore.value.setTags(tags);
 	}
-	workflowState.setWorkflowTagIds(tags);
```
r00gm, line 177: happy to approve, just double checking on this

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.test.ts
```diff
@@ -326,6 +326,37 @@ describe('useWorkflowHelpers', () => {
...
+			workflowsStore.workflow.meta = {};
+			workflowsStore.pinnedWorkflowData = {};
+
+			const documentId = createWorkflowDocumentId(workflowId);
+			const workflowDocumentStore = useWorkflowDocumentStore(documentId);
+
+			Object.defineProperty(workflowDocumentStore, 'tags', {
+				value: tagIds,
+				writable: true,
+			});
```
r00gm, line 350:
```suggestion
			workflowDocumentStore.setTags(tagIds)
```

dariacodes: It doesn't work in the test because the test file uses the default `createTestingPinia()` configuration, which has `stubActions: true`

r00gm: that's odd, please add a comment there, to not block, will check it later

dariacodes: We're actually using this pattern (`Object.defineProperty`) a lot in other tests as well, but you're right that a comment won't hurt - just added

r00gm (approved):

packages/frontend/editor-ui/src/app/composables/useWorkflowHelpers.test.ts
r00gm: just for the record, in theory this could be solved in 3 ways.

createTestingPinia({ stubActions: false });
workflowDocumentStore.setTags(tagIds);

but it might break other tests in the file, or not.


or 
workflowDocumentStore.$patch({ tags: tagIds });


or 
workflowDocumentStore.tags = tagIds;

since even if it's readonly pinia should unwraps the readonly in testing mode

dariacodes: Yeah, I decided to not change `stubActions` as it might make the test very slow when we want it to just be a unit test.
The `$patch` is new to me! I see we have it in a couple of other places, but not sure what we should prefer here

r00gm: i would first try
workflowDocumentStore.tags = tagIds;

the beauty of testingPinia is the ability to set complex states for testing with ease.

if for some reason the readonly doesn't work like computed i would use $patch, since what you are testing here is that whatever is there is propagated, you don't really care about the actions or their side effects.

dariacodes: `workflowDocumentStore.tags = tagIds` doesn't update them unfortunately
Alright, I'll try to use `$patch` more, but won't update it now to not trigger the whole CI again 😢

dariacodes: oh wow, `$patch` doesn't work for me either! Maybe related to it being read-only?

r00gm: oh boy 😞 ! nvm

PR #26041, opened 2026-02-20
feat(core): Expand workflow-sdk test fixtures from 500 to 2000 workflows and fix codegen bugs

mutdmour:
## Summary
- Expand workflow-sdk test fixtures from ~500 to 2021 published workflow templates
- Fix multiple codegen roundtrip bugs, reducing skipped workflows from 22 to 10 (manifest) + 8 (SKIP_WORKFLOWS)
- Add error connection parsing in semantic graph, composite dispatch for `.onError()` handlers, and cycle detection in subnode collection

### Test fixtures expansion
- Fetch 2021 published workflow templates (up from ~500)
- Update zip file and manifest with all workflows
- Add `update-expected-warnings.ts` and `validate-expected-warnings.ts` scripts

### Codegen bug fixes
- **Error connections**: Parse `error` type connections in semantic graph and check them in `hasErrorOutput()`
- **Composite dispatch**: Dispatch `.onError()` composite targets (IfElseBuilder, SwitchCaseBuilder) in `addSingleNodeConnectionTargets` instead of skipping them
- **Missing composite targets**: Add `addMissingCompositeTargets()` in `toJSON()` to catch composites missed during chain processing, using a WeakSet to prevent infinite recursion
- **Circular subnodes**: Add cycle detection in `collectSubnodesAsVariables` and skip self-referencing subnodes
- **Non-ASCII variable names**: Generate fallback names for nodes with non-ASCII-only names
- **typeVersion normalization**: Handle string typeVersion values in semantic graph and workflow import
- **Flat array connections**: Normalize legacy `[node, type, index]` connection format
- **Placeholder values**: Clear placeholder values in list-mode resource locators during roundtrip normalization
- **Subnode indexes**: Connect subnodes based on index. 

## Test plan
- [ ] `pnpm test` in `packages/@n8n/workflow-sdk` — all 11,324 tests pass
- [ ] `pnpm typecheck` in `packages/@n8n/workflow-sdk` — clean
- [ ] Codegen roundtrip: 2040 tests pass (10 skipped in SKIP_WORKFLOWS, 10 skipped in manifest)


mutdmour:
@@cubic-dev-ai review PR

packages/@n8n/workflow-sdk/src/codegen/codegen-roundtrip.test.ts
```diff
@@ -9,18 +9,56 @@ import {
...
+import {
+	escapeNewlinesInExpressionStrings,
+	isPlaceholderValue,
+} from '../workflow-builder/string-utils';
+
+/** Workflow IDs where deep connection comparison is skipped (known codegen/parser gaps).
+ * Remaining patterns: merge VarRef leaks in SIB/composite downstream chains,
+ * multi-trigger fan-out ordering, and missing connections in complex convergence. */
+// prettier-ignore
+const SKIP_DEEP_CONNECTION_CHECK = new Set<string>([
```
mutdmour: will address in next PR

packages/@n8n/workflow-sdk/src/workflow-builder/subnode-utils.ts
```diff
@@ -91,29 +93,47 @@ function processSubnodesRecursively(
...
+				addNestedSubnode(subnodeOrArray[i], connectionType, i);
 			}
 		} else {
-			addNestedSubnode(subnodeOrArray, connectionType);
+			addNestedSubnode(subnodeOrArray, connectionType, 0);
+		}
+	};
+
+	// For types with a single input (all connections target index 0)
+	const addNestedSubnodeFlat = (
```
aalises: Only thing I would maybe look at is that thi `addNestedSubnodeFlat` and the `addSubnodeFlat` are very similar. Not sure if we can consolidate them

mutdmour: since I am touching this in next PR, will address there https://github.com/n8n-io/n8n/pull/26172/changes#diff-e88eea3fce2c55893a89dedbfd715ed29e7499f776d9a0c88b52ee1e92a755b0

mutdmour: merged the PR into this one. addressed here.

packages/@n8n/workflow-sdk/src/codegen/composite-builder.ts
```diff
@@ -301,11 +340,44 @@ function buildIfElse(node: SemanticNode, ctx: BuildContext): IfElseCompositeNode
 	const trueBranch = buildBranchTargets(trueBranchTargets, branchCtx, node.name, 0);
 	const falseBranch = buildBranchTargets(falseBranchTargets, branchCtx, node.name, 1);
 
+	// Build error handler if node has error output
+	let errorHandler: CompositeNode | undefined;
+	if (hasErrorOutput(node)) {
+		const errorTargets = getErrorOutputTargets(node);
+		if (errorTargets.length > 0) {
+			const firstErrorTarget = errorTargets[0];
```
burivuhster, line 348: Does it mean that we ignore 2nd+ connections from the same error output?

packages/@n8n/workflow-sdk/src/workflow-builder/node-builders/node-builder.ts
```diff
@@ -690,6 +674,13 @@ class IfElseBuilderImpl<TOutput = unknown> implements IfElseBuilder<TOutput> {
 				allNodes.push(node);
 			}
 		}
+		if (this.errorBranch) {
+			for (const node of extractNodesFromTarget(this.errorBranch)) {
+				if (!allNodes.some((n) => n.name === node.name)) {
+					allNodes.push(node);
+				}
+			}
+		}
```
burivuhster, line 683: If I understand it right: this means chained method calls will produce different logic:
`node.to().OnError()` vs `node.onError().to()`
Can it be an issue somehow?

burivuhster (approved):
Looks good! Ignore comments if not relevant

packages/@n8n/workflow-sdk/src/codegen/composite-builder.ts
mutdmour: Yes, 2nd+ connections from the same error output are currently ignored. The builder API actually supports multiple `.onError()` calls (it pushes to `_connections`, doesn't  overwrite), so the fix would be to loop over all error targets and emit deferred connections for each. That being said said, fan-out from error outputs seems to be very uncommon in practice. Will leave as is for now.

packages/@n8n/workflow-sdk/src/workflow-builder/node-builders/node-builder.ts
mutdmour: Good question. The order does produce different semantics, but that's by design — .onError() always applies to the current tail of the chain. So node.to(a).onError(handler) sets the error handler on a, while node.onError(handler).to(a) sets it on node itself. This matches the fluent API convention where each method operates on the current state. The codegen always emits  .onError() right after the node it belongs to, so the generated code will always have the correct ordering.

burivuhster: Great, if we have the order baked in the codegen - everything should be fine

PR #25889, opened 2026-02-17
refactor(core): Separate Expression from Workflow class

despairblue:
## Summary

This PR separates the Expression class from the Workflow class to enable future extraction into a standalone package for use in web workers and isolated-vm isolates.

### Changes

1. **Created WorkflowExpression adapter class**
   - Maintains backward-compatible API with 13-parameter method signatures
   - Handles WorkflowDataProxy creation internally
   - Delegates to simplified Expression methods
   - Exposes `convertObjectValueToString` delegate (frontend expects it via `workflow.expression.convertObjectValueToString(...)`)

2. **Refactored Expression class**
   - Constructor now accepts `timezone: string` instead of `Workflow` instance
   - Methods accept `IWorkflowDataProxyData` instead of creating WorkflowDataProxy internally
   - Removed methods: `getSimpleParameterValue`, `getComplexParameterValue` (moved to adapter)
   - Simplified signatures: `resolveSimpleParameterValue` (13 params → 3), `getParameterValue` (12 params → 3)
   - **Zero dependencies on Workflow class**

3. **Updated Workflow class**
   - Uses `WorkflowExpression` instead of `Expression`
   - External API unchanged: `workflow.expression.resolveSimpleParameterValue(...)` still works

4. **Updated test mocks**
   - Core package test mocks updated to use `WorkflowExpression` instead of `Expression`
   - All tests passing

### Architecture

**Before:**
```
Workflow → Expression (tightly coupled)
         ↓
    WorkflowDataProxy → evaluateExpression
```

**After:**
```
Workflow → WorkflowExpression (adapter) → Expression (decoupled)
                                        ↓
                                   WorkflowDataProxy → evaluateExpression
```

### Benefits

- Expression can be extracted to standalone package in future PR
- Expression can be used in web workers without loading entire Workflow class (once the WorkflowDataProxy is decoupled from Workflow as well)
- Expression can be used in isolated-vm isolates
- Cleaner separation of concerns
- Easier to test Expression in isolation

## Review / Merge checklist

- [x] PR title and summary are descriptive
- [x] Tests included
- [ ] Docs updated or follow-up ticket created (future extraction docs TBD)
- [ ] PR Labeled with `release/backport` (not needed - internal refactoring)

PR #26047, opened 2026-02-20
feat(core): Add package scaffold with public API types and architecture docs

despairblue:
## Summary

Registers the new `@n8n/expression-runtime` package in the monorepo. This PR contains **no implementation code** — only TypeScript interfaces, error classes, and architecture documentation. The goal is to let reviewers approve the API design before any implementation lands.

### What's included

**Type surface (`src/types/`):**
- `RuntimeBridge` — abstract interface all bridge implementations must satisfy (`initialize`, `execute`, `getDataSync`, `dispose`, `isDisposed`)
- `BridgeConfig` — memory limit, timeout, debug flag
- `IExpressionEvaluator` — main public evaluation interface
- `EvaluatorConfig` — bridge + optional observability + optional AST security hooks
- `WorkflowData`, `EvaluateOptions`
- Observability interfaces: `ObservabilityProvider`, `MetricsAPI`, `TracesAPI`, `LogsAPI`, `Span`
- Error classes: `ExpressionError`, `MemoryLimitError`, `TimeoutError`, `SecurityViolationError`, `SyntaxError`
- Runtime internals: `RuntimeHostInterface`, `RuntimeGlobals`, `RuntimeError`, `LazyProxyConfig`

**`src/index.ts`:** type-only re-exports; implementation exports (`ExpressionEvaluator`, `IsolatedVmBridge`) added in later PRs.

**Documentation:**
- `ARCHITECTURE.md` — three-layer design (Runtime / Bridge / Evaluator), data flow with mermaid sequence diagrams, known limitations, environment-specific implementations
- `README.md` — usage examples, API reference, security & performance notes
- `docs/architecture-diagram.mmd` — mermaid component diagram
- `docs/deep-lazy-proxy.md` — design of the lazy data loading system (small/large array metadata, callbacks, caching)
- `docs/implementation-phases.md` — phased delivery roadmap

**Build:** `pnpm build` (tsc only, no esbuild yet) and `pnpm typecheck` both pass.

### Subsequent PRs in this series

| PR | Content |
|----|---------|
| 2 | Runtime bundle: extension functions, deep lazy proxy (`src/runtime/index.ts`, `src/extensions/`, `esbuild.config.js`) |
| 3 | `IsolatedVmBridge`: V8 isolate management via `isolated-vm` |
| 4 | `ExpressionEvaluator`: tournament integration, expression code cache, integration tests |
| 5 | Workflow integration (`packages/workflow`) + Vite browser stub (`packages/frontend/editor-ui`) |

https://linear.app/n8n/issue/CAT-2275/phase-01-core-architecture-design

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/@n8n/expression-runtime/ARCHITECTURE.md
+++ b/packages/@n8n/expression-runtime/ARCHITECTURE.md
@@ -1,0 +1,426 @@
+# Expression Runtime Architecture
+
+This package provides a secure, isolated expression evaluation runtime that works across multiple execution environments (isolated-vm, Web Workers, and task runners).
+
+## Design Goals
+
+1. **Environment Agnostic**: Single codebase that works in Node.js (isolated-vm), browsers (Web Workers), and task runner processes
+2. **Security**: Expressions run in isolated contexts with memory limits and timeouts
+3. **Performance**: Lazy data loading, code caching, and efficient data transfer
+4. **Observability**: Built-in metrics, traces, and logs
+5. **Maintainability**: Clear separation of concerns with well-defined interfaces
+
+## Three-Layer Architecture
+
+The architecture is split into three distinct layers:
+
+```
+┌─────────────────────────────────────────────────────────┐
+│                     Host Process                        │
+│                                                         │
+│  ┌────────────────────────────────────────────────┐     │
+│  │         ExpressionEvaluator (Layer 3)          │     │
+│  │  - Public API                                  │     │
+│  │  - Tournament integration                      │     │
+│  │  - Code caching                                │     │
+│  │  - Observability                               │     │
+│  └────────────────┬───────────────────────────────┘     │
+│                   │                                     │
+│  ┌────────────────▼───────────────────────────────┐     │
+│  │           Bridge (Layer 2)                     │     │
+│  │  - IsolatedVmBridge (Phase 1.1)                │     │
+│  │  - WebWorkerBridge (Phase 2+)                  │     │
+│  │  - Task Runner Integration (TBD)               │     │
+│  └────────────────┬───────────────────────────────┘     │
+│                   │ IPC/Message Passing                 │
+└───────────────────┼─────────────────────────────────────┘
+                    │
+┌───────────────────▼─────────────────────────────────────┐
+│              Isolated Context                           │
+│                                                         │
+│  ┌────────────────────────────────────────────────┐     │
+│  │            Runtime (Layer 1)                   │     │
+│  │  - Runs inside isolation                       │     │
+│  │  - No Node.js dependencies                     │     │
+│  │  - Lazy loading proxies                        │     │
+│  │  - Helper functions ($json, $item, etc.)       │     │
+│  │  - lodash, Luxon                               │     │
+│  └────────────────────────────────────────────────┘     │
+│                                                         │
+└─────────────────────────────────────────────────────────┘
+```
+
+### Layer 1: Runtime (Isolated Context)
+
+**Location**: Runs inside the isolated context (isolate, worker, subprocess)
+
+**Purpose**: Provides the JavaScript execution environment for expressions
+
+**Key Components**:
+- **Lazy Loading Proxies**: Fetch data fields on-demand from host to avoid memory limits
+- **Helper Functions**: `$json`, `$item`, `$input`, `$`, etc.
+- **Libraries**: lodash, Luxon (bundled)
+- **No Node.js APIs**: Pure JavaScript only
+
+**Bundle**: IIFE format for isolated-vm, ESM for Web Workers
+
+### Layer 2: Bridge (Host Process)
+
+**Location**: Runs in the host process
+
+**Purpose**: Manages communication between host and isolated context
+
+**Key Components**:
+- **RuntimeBridge Interface**: Abstract interface for all bridge implementations
+- **IsolatedVmBridge**: Uses isolated-vm API for Node.js backend (Phase 1.1)
+- **WebWorkerBridge**: Uses postMessage API for browser (Phase 2+)
+- **Task Runner Integration**: TBD - May use IsolatedVmBridge locally or direct evaluation (Phase 2+)
+
+**Responsibilities**:
+- Initialize isolated context
+- Transfer code to context
+- Handle data requests from runtime (lazy loading)
+- Enforce memory limits and timeouts
+- Dispose of context when needed
+
+### Layer 3: Evaluator (Host Process)
+
+**Location**: Runs in the host process
+
+**Purpose**: Public API for expression evaluation
+
+**Key Components**:
+- **ExpressionEvaluator**: Main class used by workflow package
+- **Tournament Integration**: AST transformation and security validation
+- **Code Cache**: Cache transformed code (not evaluation results)
+- **Observability**: Emit metrics, traces, and logs
+
+**Responsibilities**:
+- Accept expression strings and workflow data
+- Transform expressions with Tournament
+- Cache transformed code to avoid re-transformation
+- Convert WorkflowData to WorkflowDataProxy for lazy loading
+- Use bridge to evaluate in isolated context
+- Handle errors gracefully
+- Emit observability data
+
+## Data Flow
+
+### Expression Evaluation Flow
+
+```mermaid
+sequenceDiagram
+    participant WF as Workflow
+    participant Eval as ExpressionEvaluator
+    participant Bridge as IsolatedVmBridge
+    participant Runtime as Runtime (Isolated)
+
+    WF->>Eval: evaluate(expr, data)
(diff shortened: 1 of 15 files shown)


packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
```diff
@@ -0,0 +1,245 @@
+# Deep Lazy Proxy
```
despairblue, line 1: Mainly here for claude code to keep context. I'll remove this or trim it later.

packages/@n8n/expression-runtime/docs/architecture-diagram.mmd
```diff
@@ -0,0 +1,92 @@
+%% Expression Runtime Architecture
```
despairblue, line 1: Mainly here for claude code to keep context. I'll update and inline this into the README later.

packages/@n8n/expression-runtime/docs/implementation-phases.md
```diff
@@ -0,0 +1,155 @@
+# Implementation Phases
```
despairblue, line 1: Mainly here for claude code to keep context. I'll remove this later.

packages/@n8n/expression-runtime/src/types/bridge.ts
```diff
@@ -0,0 +1,89 @@
+// ============================================================================
+// Phase 1.1: Bridge Interface (CORE - IMPLEMENT FIRST)
+//
+// This is the main interface all environments must implement.
+// Start here for CLI/backend (IsolatedVmBridge) or frontend (WebWorkerBridge).
```
despairblue, line 5: As of now it's unclear if this can really be re-used in a web worker. The current FE approach for offloading expression evaluations in the FE is offloading much more than just the expression engine to a web worker, the whole workflow instance is offloaded. 

To make this re-usable we also need to 
1. separate Workflow from Expression
2. WorkflowDataProxy from Workflow
3. move WorkflowDataProxy into the isolate/worker

This is future work, but has been started already here: https://github.com/n8n-io/n8n/pull/25889 

If it turns out that this is not possible I'll simplify this and remove the bridge abstraction, no need to have it in that case.

packages/@n8n/expression-runtime/src/types/evaluator.ts
```diff
@@ -0,0 +1,244 @@
...
+ */
+export interface EvaluateOptions {
+	/**
+	 * Custom timeout for this evaluation (in milliseconds).
+	 * Overrides the bridge's default timeout.
+	 */
+	timeout?: number;
+}
+
+// ============================================================================
```
despairblue, line 101: All of the Observability Interfaces may change once their implementation work has started.

packages/@n8n/expression-runtime/README.md
```diff
@@ -0,0 +1,307 @@
...
+pnpm test
+
+# Run tests in watch mode
+pnpm test:watch
+
+# Type check
+pnpm typecheck
+
+# Lint
+pnpm lint
```
despairblue, line 234: Adding linting is on my todo list.

packages/@n8n/expression-runtime/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+```bash
+# Install dependencies
+pnpm install
+
+# Build package
+pnpm build
+
+# Run tests
+pnpm test
```
despairblue, line 225: Tests will be added with the actual implementation.

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
```diff
@@ -0,0 +1,245 @@
...
+2. Calls `resetDataProxies()` in the isolate to create fresh lazy proxies for
+   `$json`, `$binary`, `$input`, `$node`, `$parameter`, `$workflow`, `$prevNode`
+3. Runs the tournament-transformed expression code with `this === __data`
+
+From the expression's perspective it just sees normal objects:
+
+```typescript
+// Inside an expression (runs in isolate):
+$json.user.email        // triggers getValueAtPath(['$json','user','email'])
+$json.items[150].id     // triggers getArrayElement(['$json','items'], 150)
```
tomi, line 51: Is the `id` intentionally omitted here?

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
```diff
@@ -0,0 +1,245 @@
...
+$items()                // triggers callFunctionAtPath(['$items'])
+```
+
+### Small vs. Large Array Threshold
+
+Arrays with ≤ 100 elements are transferred in full (with `__data`); larger arrays
```
tomi: How did you land on a number 100?

packages/@n8n/expression-runtime/ARCHITECTURE.md
```diff
@@ -0,0 +1,426 @@
...
+## Observability
+
+All layers emit metrics, traces, and logs:
+
+**Metrics**:
+- `expression.evaluation.count`
+- `expression.evaluation.duration_ms`
+- `expression.code_cache.hit` (transformed code cache)
+- `expression.code_cache.miss`
+- `expression.isolate.memory_mb`
```
tomi, line 401: 👏

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Access Patterns
+
+Best performance when:
+- Accessing few properties from large objects
+- Accessing specific array elements (not iterating entire array)
+- Accessing same properties multiple times (caching)
+
+Suboptimal performance when:
+- Iterating entire large arrays (`.map()`, `.filter()`)
```
tomi: Have you analyzed what are typical patterns to use in expressions?

packages/@n8n/expression-runtime/src/types/bridge.ts
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * IMPORTANT: This must be SYNCHRONOUS for lazy-loading proxies to work.
+	 * - IsolatedVmBridge: Uses ivm.Reference for synchronous callbacks
+	 * - WebWorkerBridge: Must pre-fetch data (no lazy loading in Phase 1)
+	 *
+	 * @param path - Property path (e.g., "user.email")
+	 * @returns Value at the path, or undefined if not found.
+	 *          Must be JSON-serializable.
+	 */
+	getDataSync(path: string): unknown;
```
tomi: If this is a string, we need to parse it into bits every time. Should it be an array instead? `["user", "email"]`?

tomi (commented):
Few questions, no objections really 👍

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
despairblue: Yeah, it's a bit simplified, what it will do is first get `items` with `getValueAtPath`, which will return a meta object describing it, not the actual value, then it goes on to ask the host for `[150]`, etc:

```
getValueAtPath (path: [ '$json', 'items' ]            )	-> { __isArray: true, __length: 200, __data: null }
getArrayElement(path: [ '$json', 'items' ], index: 150) -> { __isObject: true, __keys: [ 'id' ] }
getValueAtPath (path: [ '$json', 'items', '1', 'id' ] ) -> 2
```

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
despairblue: Completely arbitrarily chosen by claude. Honestly I think we should never transfer the full array, we can only linearly determine the length of it, not the size and thus not how much overhead serializing and deserializing would be.

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
despairblue: No, will do.

packages/@n8n/expression-runtime/src/types/bridge.ts
despairblue: Mhh, this may not be used at all. It was there for getting data synchronously which is what web workers would do. We can't access data asynchronously there, in `$json.items.user.id` `user` has to return an object, not a promise, but then again we don't need this function at all, the web worker can just use the WorkflowDataProxy directly. 

I'll remove it.

packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
despairblue: I will do that, but it should not block this PR I think.

tomi (approved):
🚀

PR #26053, opened 2026-02-20
fix(Gmail Node): Fix GetMany messages query string pollution with non-API parameters

aalises:
## Summary

Fix the Gmail node's `getAll` messages operation incorrectly merging `options` and the loop index `i` into the API query string via `Object.assign`.

**Before:**
```typescript
Object.assign(qs, prepareQuery.call(this, filters, i), options, i);
```

**After:**
```typescript
Object.assign(qs, prepareQuery.call(this, filters, i));
```

The `options` object contains internal n8n parameters (`dataPropertyAttachmentsPrefixName`, `downloadAttachments`) that are not Gmail API query parameters. Merging them into `qs` pollutes the query string sent to the Gmail API. The trailing `i` (loop index number) was also incorrectly passed as a source to `Object.assign`.

This caused failures when the Gmail node was used as a tool via the MCP Server Trigger, as the polluted query string led to faulty API requests. The thread `getAll` operation (line 729) already uses the correct pattern without `options` and `i`.

The `options` object is still properly accessed later in the function (line 427) for `dataPropertyAttachmentsPrefixName` — it just should not be merged into the API query string.

- https://linear.app/n8n/issue/AI-2082

### DIFF
--- a/packages/nodes-base/nodes/Google/Gmail/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/Google/Gmail/GenericFunctions.ts
@@ -366,8 +366,8 @@ export function prepareQuery(
 		} else {
 			qs.q = `is:${qs.readStatus}`;
 		}
-		delete qs.readStatus;
 	}
+	delete qs.readStatus;
 
 	if (qs.receivedAfter) {
 		qs.q = prepareTimestamp(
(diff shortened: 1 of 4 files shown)


burivuhster (approved):
🙌

PR #26063, opened 2026-02-20
feat(editor): Improve Google OAuth button design

elsmr:
## Summary

image
image
image

https://linear.app/n8n/issue/NODE-4403/update-google-oauth-button-design

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/design-system/src/components/N8nTooltip/Tooltip.vue
```diff
@@ -130,6 +130,10 @@ const handleOpenChange = (open: boolean) => {
 	gap: var(--spacing--2xs);
 }
 
+.trigger {
+	display: inline-block;
```
elsmr: Without this the tooltip does not inherit the size of the child elements, I checked some other tooltips and did not see side effects

elsmr: nevermind, it did have side effects, reverted

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/GoogleAuthButton.vue
```diff
@@ -1,52 +1,80 @@
...
-</script>
-
 <template>
-	<button
-		:class="$style.googleAuthBtn"
-		:title="i18n.baseText('credentialEdit.oAuthButton.signInWithGoogle')"
-		:style="googleAuthButtons"
-	/>
+	<button :class="$style.googleAuthBtn">
+		<svg
```
dlavrenuek, line 3: would be good to have it stored in a file and import it

dlavrenuek (dismissed):

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/GoogleAuthButton.vue
elsmr: I had it like this first, but having the svg as an `img` tag with `src` allows less styling flexibility

dlavrenuek: It should also be possible to import the contents of a svg and render it as an actual element rather than an image with src, but it's not that important

dlavrenuek (approved):

PR #26064, opened 2026-02-20
fix(editor): Design fixes for managed OAuth credentials

elsmr:
## Summary

- Fix cutoff pixel in save button
- Hide docs callout for credentials with managed OAuth
- Hide notice for credentials with managed OAuth

https://linear.app/n8n/issue/NODE-4488/design-follow-ups-pt2

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -1418,6 +1418,7 @@ const { width } = useElementSize(credNameRef);
 .mainContent {
 	flex: 1;
 	overflow: auto;
+	padding-left: var(--spacing--5xs);
```
elsmr: Kind of a hacky solution, I tried a bunch of things but it's pretty mysterious.

It's definitely connected the the `overflow: auto` on the parent container.

dlavrenuek: The spacing variables have relative size with `rem` which is based on the font size (afaik). The button is exactly 1px off, let's try it instead with a fixed `1px`?

dlavrenuek (dismissed):
Approving with a suggestion

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -1418,6 +1418,7 @@ const { width } = useElementSize(credNameRef);
 .mainContent {
 	flex: 1;
 	overflow: auto;
+	padding-left: 1px;
```
dlavrenuek: I've just realized that here the padding is added to `mainContent`, this will also offset all of the content like input fields and the input fields will be 1px to the right compared to the button. I also did not find where the problem comes from, but the padding should only be there for the button and not other elements

dlavrenuek (changes requested):
Sorry for the back and forth, didn't realize what element the padding was on on the first review

dlavrenuek (dismissed):

dlavrenuek (approved):

PR #26065, opened 2026-02-20
fix(editor): Workflow select filtering uses any of the keywords

nikhilkuria:
## Summary
Summary: The error workflow select filter used OR logic when joining multi-word search conditions, causing results to increase as more words were typed. Changed to AND logic so results correctly narrow with more specific queries.

https://linear.app/n8n/issue/ADO-4805/ux-error-workflow-select-filtering-increases-results-instead-of

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/collaboration/__tests__/collaboration.state.test.ts
CharlieKolb: is this related?

PR #26077, opened 2026-02-21
feat(core): Add runtime bundle and extension utilities

despairblue:
## Context

n8n expressions today run via `tournament` (an AST transformer) inside the main Node.js process. The goal of this PR series is to make expression evaluation available inside **`isolated-vm`** — a V8 isolate that is memory-isolated from the host process. This gives us a proper security boundary: a malicious expression cannot access Node.js APIs, the filesystem, or the host's memory.

The series is being landed incrementally:

| PR | What it adds |
|----|-------------|
| #26047 ✅ | Package scaffold: public types and architecture docs |
| **This PR** | The runtime bundle that runs _inside_ the isolate |
| PR 3 (upcoming) | `IsolatedVmBridge` — creates the isolate, loads the bundle, and exposes workflow data to it via callbacks |
| PR 4 (upcoming) | `ExpressionEvaluator` — the public API that drives the bridge, plus integration tests |

## What this PR adds

When an expression is evaluated inside an isolate, the isolate has **no access to the host process** — it cannot `require()` anything or call Node.js APIs directly. Every piece of code it needs must be bundled into a single self-contained script and injected at isolate startup.

This PR produces that script.

**`esbuild.config.js`** builds `src/runtime/index.ts` and all its imports into two bundle formats:
- `dist/bundle/runtime.iife.js` — an IIFE loaded into the isolate as a string via `isolated-vm`
- `dist/bundle/runtime.esm.js` — an ESM module for future Web Worker support

**`src/runtime/`** — the code that runs inside the isolate, split into four focused modules:

- `safe-globals.ts` — `SafeObject` and `SafeError` proxies that block dangerous methods (prototype pollution, stack trace manipulation) from expression code; `ExpressionError` for tournament-generated error handlers; `__sanitize()` for dynamic property access (guards against `obj["__proto__"]` etc.)
- `lazy-proxy.ts` — `createDeepLazyProxy()`, a `Proxy` that loads workflow data on demand across the isolate boundary. When an expression accesses `$json.name`, the proxy calls an `ivm.Reference` callback set by the bridge (PR 3) to fetch just that value — nothing is copied up front.
- `reset.ts` — `resetDataProxies()`, called by the bridge before each expression evaluation to clear proxy caches and re-expose `$json`, `$binary`, `$input`, etc. to `globalThis`.
- `index.ts` — entry point: declares global types, exposes `extend`/`extendOptional` and `DateTime` for tournament-transformed expressions, and wires the above modules to `globalThis`.

**`src/extensions/`** — see the duplication note below.

**`packages/workflow/src/extensions/*.ts`** — a 7-line comment prepended to each file documenting the mirror.

## Why is `src/extensions/` a copy of `packages/workflow/src/extensions/`?

n8n expressions support method calls like `$json.name.toUpperCase()` or `[1,2,3].sum()`. The non-native ones (`sum`, `toDateTime`, etc.) are provided by the extension framework in `packages/workflow/src/extensions/`. The isolate needs these too — but it cannot import them from the host, because all isolate code must be pre-bundled.

The natural fix would be to import them from `@n8n/expression-runtime` and have esbuild bundle them in. The problem is that `@n8n/expression-runtime` is **Vite-aliased to an empty stub** for browser builds, to exclude the `isolated-vm` native addon. Vite analyses dynamic imports statically — even a `IS_FRONTEND`-guarded `import('@n8n/expression-runtime')` in `packages/workflow` would cause the browser build to pull in the stub, meaning the extension functions would be missing at runtime in the editor.

The workaround is to keep a verbatim copy of the extension files inside `@n8n/expression-runtime/src/extensions/`, which esbuild bundles directly into the IIFE. Each mirrored file in `packages/workflow/src/extensions/` has a comment explaining this and pointing to the long-term fix: split `@n8n/expression-runtime` into a browser-safe subpath (extension utilities only, no `isolated-vm`) and a node-only subpath (the bridge and evaluator, Vite-stubbed). At that point, `packages/workflow` can import the extensions directly from the package and the duplication goes away.

## Why no tests in this PR?

**`src/extensions/`** — already covered by the existing test suite in `packages/workflow/src/extensions/`. Duplicating it would mean keeping two suites in sync.

**`src/runtime/`** — the interesting behaviour (`createDeepLazyProxy`, `resetDataProxies`) is the cross-isolate data transfer: the proxy calls `ivm.Reference` callbacks set by the bridge, which live in the host process. That transfer cannot happen without an actual isolate. Unit tests with mocked callbacks would only test that a mock was called — they would not exercise the serialisation, the `copy: true` transfer semantics, or the metadata protocol (`__isArray`, `__isObject`, `__isFunction`) that the bridge and runtime agree on.

The integration tests that cover the full path — bridge creates isolate → loads runtime bundle → sets callbacks → expression evaluates and produces the correct result — are planned for PR 4, once both the bridge and the evaluator exist.

## Verification

```
cd packages/@n8n/expression-runtime
node esbuild.config.js   # produces dist/bundle/runtime.iife.js and runtime.esm.js
pnpm typecheck           # passes
```

https://linear.app/n8n/issue/CAT-2280/phase-13-extension-framework

## Review / Merge checklist

- [x] Tests included. (deferred to PR 4 — see note above)


packages/@n8n/expression-runtime/docs/deep-lazy-proxy.md
```diff
@@ -155,29 +155,22 @@ Suboptimal performance when:
```
despairblue, line 155: Just changing some inconsistencies that were in the last PR.

packages/@n8n/expression-runtime/ARCHITECTURE.md
```diff
@@ -306,7 +306,7 @@ packages/@n8n/expression-runtime/
 
 **Future-Proofing**: Frontend will use Web Workers. Backend uses isolated-vm. Abstract bridge allows adding new environments without changing other layers.
```
despairblue, line 307: Just changing some inconsistencies that were in the last PR.

packages/@n8n/expression-runtime/src/extensions/array-extensions.ts
```diff
@@ -0,0 +1,705 @@
+import isEqual from 'lodash/isEqual';
```
despairblue, line 1: Everything is this folder is just copied from the workflow package. This is temporary, I just need to figure out a way for the workflow package to be able to import theses files without messing up the FE build later, then I can remove the files from the workflow package completely.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
mfsiega, line 1: I do feel like there could be some unit tests here? Of course we can cover it with integration tests as well, but I think we could add reasonable unit tests here?

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
```diff
@@ -0,0 +1,174 @@
...
+ * 1. When property accessed: Call __getValueAtPath([path]) to get metadata
+ * 2. Metadata indicates type: primitive, object, array, or function
+ * 3. For objects/arrays: Create nested proxy for lazy loading
+ * 4. For functions: Create wrapper that calls __callFunctionAtPath
+ * 5. Cache all fetched values in target to avoid repeated callbacks
+ *
+ * @param basePath - Current path in object tree (e.g., ['$json', 'user'])
+ * @returns Proxy object with lazy loading behavior
+ */
+export function createDeepLazyProxy(basePath: string[] = []): any {
```
mfsiega, line 21: It's a bit hard (for me) to review this in a vacuum without seeing how it's used. Adding test cases might help, unless I'm misunderstanding.

mfsiega (commented):
_Basically_ all looks fine but I found the lazy-proxy a bit hard to review in detail without seeing how it'll be used. If it's possible to add some unit tests that illustrate and provide some coverage that would be good, otherwise might be easiest to just combine it with the next PR after all (or break the extensions stuff into its own PR, since that seems straightforward).

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: Tests added: 5de27e2c2737db61066bfe863b8f0d03535bbba6

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: Tests added: 5de27e2c2737db61066bfe863b8f0d03535bbba6

despairblue:
@mfsiega I've added unit tests. When adding the integration tests in PR 4 I'll dedupe them and keep the ones that test things the integration tests cannot test. I'll do that because the unit tests won't serve a purpose if the integration tests cover the same thing while being equally fast. Keeping both is just going to slow down refactors.

packages/@n8n/expression-runtime/src/runtime/index.ts
mfsiega, line 1: I think these are generally good. Feel maybe slightly more than necessary but not a big deal.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
```diff
@@ -0,0 +1,161 @@
+// ============================================================================
+// Deep Lazy Proxy System
+// ============================================================================
+
+/**
```
mfsiega, line 5: Might consider linking to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy or similar. I'd never seen this before, not sure if it's a more obscure feature or I'm just ignorant.

packages/@n8n/expression-runtime/src/runtime/__tests__/lazy-proxy.test.ts
```diff
@@ -0,0 +1,495 @@
...
+		it('returns empty array for __path when basePath is default', () => {
+			const proxy = createDeepLazyProxy();
+			expect(proxy.__path).toEqual([]);
+			expect(mocks.getValueAtPath).not.toHaveBeenCalled();
+		});
+
+		it('returns basePath for __path when basePath is provided', () => {
+			const proxy = createDeepLazyProxy(['$json', 'user']);
+			expect(proxy.__path).toEqual(['$json', 'user']);
+			expect(mocks.getValueAtPath).not.toHaveBeenCalled();
```
mfsiega, line 86: Nit: this _seems_ like an unnecessary assertion (and similarly above and below)?

mfsiega (approved):
I know I'm a small-PR zealot but in this case I think the review might actually be more useful if you merge the changes in PR3 into this one.

But I'm also fine if you want to just go ahead, I think it will still be easily reviewable.

A few tiny comments on tests, feel free to leave for later.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: Funny enough the target of this proxy is in fact a proxy 😆 

https://github.com/n8n-io/n8n/blob/5de27e2c2737db61066bfe863b8f0d03535bbba6/packages/workflow/src/workflow-data-proxy.ts#L138

The workflow data proxy is also a proxy, so we're proxying a proxy into an isolate, proxies all the way down. 

I'll add a link to mdn in the next PR nonetheless.

packages/@n8n/expression-runtime/src/runtime/__tests__/lazy-proxy.test.ts
despairblue: So the assertion is necessary, I think. 
So for `__path` it does not matter, because it should not be accessed from the outside anyways, maybe we can also true privacy here (`#`).

For the `toString` it's actually important that we don't try to copy it into the isolate, since we can't copy native functions. That said, we may not need this at all given that we'll return a meta object that can be called through the bridge. I'll keep this in my todo list to see if I still need this exception or not.

despairblue: That's why I wasn't adding unit tests, in the end I need the integration tests to **prove** things work as expected.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: done in a3245a45805189d085dc90b35d7d76bc8c35dcd5

packages/@n8n/expression-runtime/src/runtime/__tests__/lazy-proxy.test.ts
despairblue: Moved away from having `__path` etc. props: 7d3bd9c18fa69853826c5c3617dcd0cc0ede225c

PR #26082, opened 2026-02-21
fix(task-runner): separate preamble lines and user code onto distinct lines to prevent V8 parser confusion

rergards:
## Summary

Fixes #26081

`createVmExecutableCode` joined all preamble statements with `'; '` onto a single line, placing user code on the same line via `VmCodeWrapper`. `runDirectly` similarly placed user code on the same line as the `async function() {` opening brace. V8's line-based token tracking misattributed tokens from the user's first line back to the preamble parse context, producing false `SyntaxError` for syntactically valid JavaScript patterns:

- Regex character classes: `/[:.]/g`
- Nested object literals: `{ options: { temperature: 0.3 } }`
- Multi-line `if/else` blocks

## What changed

`packages/@n8n/task-runner/src/js-task-runner/js-task-runner.ts`

```diff
-		`module.exports = async function VmCodeWrapper() {${code}\n}()`,
-	].join('; ');
+		`module.exports = async function VmCodeWrapper() {\n${code}\n}()`,
+	].join(';\n');

-		`with(context) { return (async function() {${code}\n})(); }`,
+		`with(context) { return (async function() {\n${code}\n})(); }`,
```

3 characters changed across 2 lines. Each preamble statement now gets its own line and user code starts on a fresh line in both `runInContext` (secure) and `runDirectly` (insecure) modes.

## Review / Merge checklist

- [ ] Docs updated or follow-up ticket created. _(no docs change needed — purely internal code generation)_
- [ ] Tests included. _(see below)_

## Test plan

Paste each snippet into a Code node (run-once mode) and execute — all should run without error after this fix:

```javascript
// Was: Unexpected token ']'
const date = new Date().toISOString().replace(/[:.]/g, '-');
return { date };
```

```javascript
// Was: Unexpected token '}'
const body = { model: 'llama3', options: { temperature: 0.3 } };
return { body };
```

```javascript
// Was: Unexpected token '}'
if (someCondition) {
  doA();
} else {
  doB();
}
return { ok: true };
```

Verified by patching the compiled JS in a running n8n 2.9.1 Docker container — all three cases pass after the fix on both default mode and `N8N_RUNNERS_INSECURE_MODE=true`.

### DIFF
--- a/packages/@n8n/task-runner/src/js-task-runner/__tests__/js-task-runner.test.ts
+++ b/packages/@n8n/task-runner/src/js-task-runner/__tests__/js-task-runner.test.ts
@@ -2153,4 +2153,74 @@ describe('JsTaskRunner', () => {
 			});
 		});
 	});
+
+	describe('createVmExecutableCode', () => {
+		type RunnerWithPrivate = { createVmExecutableCode: (code: string) => string };
+
+		it('joins preamble statements with newlines so each occupies its own line', () => {
+			const result = (defaultTaskRunner as unknown as RunnerWithPrivate).createVmExecutableCode(
+				'return 1',
+			);
+			const lines = result.split('\n');
+			// No line should be empty (no spurious blank lines from join)
+			expect(lines.every((l) => l.length > 0)).toBe(true);
+			// There must be more than one statement line (preamble + wrapper)
+			expect(lines.length).toBeGreaterThan(1);
+		});
+
+		it('starts user code on a fresh line inside VmCodeWrapper', () => {
+			const code = 'const x = /[:.]/g;';
+			const result = (defaultTaskRunner as unknown as RunnerWithPrivate).createVmExecutableCode(
+				code,
+			);
+			// The wrapper opening brace must be followed by a newline, not the code directly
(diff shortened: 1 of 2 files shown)


rergards:
Thanks for the review! Just pushed a follow-up commit (efa3276) that addresses both points:

**Unit tests** (`describe('createVmExecutableCode')`):
- Asserts each preamble statement is on its own line (no blank lines from the join)
- Asserts user code starts on a fresh line inside `VmCodeWrapper()` (`{\n${code}`)
- Asserts user code ends with `\n` before the closing `}()`

**Integration tests** (`describe('V8 parser edge cases – preamble line separation')`):
- Regex character class on first line: `/[:.]/g`
- Nested object literal on first line: `{ options: { temperature: 0.3 } }`
- Multiline `if/else` block

**Docstring** added to `createVmExecutableCode` explaining:
- What `@param code` expects (raw multi-line user JS)
- What `@returns` (newline-separated string for `runInContext`)
- Why the line separation matters (V8 token misattribution)

darshjme-codes:
Perfect! The test coverage looks solid now and the docstring really clarifies the intent. Thanks for the quick turnaround on the feedback. 🚀

Issue #22614, opened 2025-12-02
Cannot send Email using Gmail node


andig:
## Describe the problem/error/question

Error message when sending email

## What is the error message (if any)?

```json
{
  "errorMessage": "this.getNodeParameter(...).trim is not a function (item 0)",
  "errorDetails": {},
  "n8nDetails": {
    "nodeName": "Send a message",
    "nodeType": "n8n-nodes-base.gmail",
    "nodeVersion": 2.1,
    "resource": "message",
    "operation": "send",
    "itemIndex": 0,
    "time": "12/2/2025, 3:24:51 PM",
    "n8nVersion": "1.122.4 (Self Hosted)",
    "binaryDataMode": "default",
    "stackTrace": [
      "NodeOperationError: this.getNodeParameter(...).trim is not a function (item 0)",
      "    at ExecuteContext.execute (/usr/local/lib/node_modules/n8n/node_modules/n8n-nodes-base/nodes/Google/Gmail/v2/GmailV2.node.ts:836:11)",
      "    at WorkflowExecute.executeNode (/usr/local/lib/node_modules/n8n/node_modules/n8n-core/src/execution-engine/workflow-execute.ts:1074:31)",
      "    at WorkflowExecute.runNode (/usr/local/lib/node_modules/n8n/node_modules/n8n-core/src/execution-engine/workflow-execute.ts:1255:22)",
      "    at /usr/local/lib/node_modules/n8n/node_modules/n8n-core/src/execution-engine/workflow-execute.ts:1691:38",
      "    at /usr/local/lib/node_modules/n8n/node_modules/n8n-core/src/execution-engine/workflow-execute.ts:2307:11"
    ]
  }
}
```

## Please share your workflow/screenshots/recording

<img width="270" height="205" alt="Image" src="https://github.com/user-attachments/assets/974a52ce-8968-4d7d-b760-5524b839c8fa" />

### core

- n8nVersion: 1.122.4
- platform: npm
- nodeJsVersion: 24.11.1
- nodeEnv: undefined
- database: sqlite
- executionMode: regular
- concurrency: -1
- license: enterprise (production)

### storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

### pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

### client

[...]

PR #26105, opened 2026-02-23
fix: Gmail node crashes when non-string value passed to message/sendTo fields

dlavrenuek:
## Summary

The Gmail node was crashing with `this.getNodeParameter(...).trim is not a function` when an expression in the **Message** (or any email address field) resolved to a non-string type — for example a number from pinned-data like `$json.code`.

Root cause: two helper functions in `GenericFunctions.ts` assumed their inputs were always strings:

1. **`prepareEmailBody`** — called `.trim()` directly on the raw return value of `getNodeParameter('message', ...)`. If the expression result was a number (or other non-string), this threw a `TypeError`.
2. **`prepareEmailsInput`** — accepted `input: string` but was called without prior coercion. Passing a non-string caused `.split(',')` to fail in the same way.

**Fix:**
- `prepareEmailBody`: coerce the message to a string with `String(rawMessage)` before calling `.trim()`.
- `prepareEmailsInput`: change parameter type from `string` to `unknown` and coerce with `String(input)` before calling `.split(',')`.

No behavioural change for users who pass normal string values. Non-string values are now safely converted, so e.g. a numeric body `42` is sent as the string `"42"` instead of crashing.

- https://linear.app/n8n/issue/NODE-4122
- https://github.com/n8n-io/n8n/issues/22614

## Review / Merge checklist

- [x] Tests included.
- [ ] PR Labeled with `release/backport` (if urgent fix that needs to be backported)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,3 +175,4 @@ When implementing features:
 - Always reference the Linear ticket in the PR description,
   use `https://linear.app/n8n/issue/[TICKET-ID]`
 - always link to the github issue if mentioned in the linear ticket.
+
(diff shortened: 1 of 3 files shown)


RomanDavydchuk:
Closing in favor of #33800

PR #26112, opened 2026-02-23
fix(editor): Respect displayOptions when determining response mode for Chat on WF canvas

Cadiac:
## Summary

Resolve node option parameters respecting their displayOptions, so that the correct defaults get loaded. We have different default value for `responseMode` depending on the value of `availableInChat` (streaming or lastNode), which lead to the chat widget being in wrong mode when Chat hub was enabled on the chat trigger.

Instead of just making the defaults logic match here I made this actually resolve the parameters from the node, which should prevent this from breaking again in the future if/when we change the node.

https://linear.app/n8n/issue/CHA-149/bug-chat-widget-streaming-on-wf-canvas-is-broken-with-chat-hub-enabled

## Review / Merge checklist

- [ ] Tests included. 


schrothbn (approved):
Thanks 👍

PR #26133, opened 2026-02-23
fix: Use instance origin for waiting forms and webhooks CORS headers instead of wildcard

### LINKED ISSUE #23262: Form trigger always looping

michael-radency:
## Summary

Waiting webhooks(and forms) could break for users behind reverse proxies that strip `Access-Control-Allow-Origin: * `responses

Fix: `applyCors` now uses the configured instance origin instead of * when the request has no real origin (CSP-sandboxed form pages). The origin is derived from `WEBHOOK_URL` (falls back to `N8N_HOST/N8N_PROTOCOL/N8N_PORT`).

Required: `WEBHOOK_URL` must be set to the public-facing n8n URL for the fix to take effect

https://linear.app/n8n/issue/NODE-4138/community-issue-form-trigger-always-loop
fixes  https://github.com/n8n-io/n8n/issues/23262

## Review / Merge checklist

- [ ] Tests included. 


Matsuuu:
Can we get some eyes on this PR? 🙏🏼 

There's a community GItHub Issue and a Forum report on this and would be great to get it merged

packages/cli/src/webhooks/waiting-webhooks.ts
```diff
@@ -46,13 +47,18 @@ import { preserveInputOverride } from '@/workflow-helpers';
...
 	constructor(
 		protected readonly logger: Logger,
 		protected readonly nodeTypes: NodeTypes,
 		private readonly executionRepository: ExecutionRepository,
 		private readonly webhookService: WebhookService,
 		private readonly instanceSettings: InstanceSettings,
-	) {}
+		urlService: UrlService,
+	) {
+		this.instanceOrigin = new URL(urlService.getWebhookBaseUrl()).origin;
```
DawidMyslak, line 60: Nit: if `getWebhookBaseUrl()` returns something `new URL()` can't parse, this throws at construction time with an error. A try/catch falling back to `'*'` might be worth adding. Unless this is not possible?

DawidMyslak (commented):
Thanks for working on this @michael-radency!

I have a concern about the CORS interaction with the CSP sandbox. 

`getWebhookSandboxCSP()` deliberately excludes `allow-same-origin`, so sandboxed form pages have origin `null`. With this change the response becomes `Access-Control-Allow-Origin: <instanceOrigin>` instead of `*`.

Per the CORS spec a specific origin must match the request origin exactly - `null` !== `https://n8n.example.com`, so the browser should still reject the response?

Could you confirm this was tested end-to-end with a sandboxed form page behind a reverse proxy?

Also, `findAccessControlOptions` still returns `allowedOrigins: '*'`, while the actual response now uses `instanceOrigin`. These should probably be consistent?

packages/cli/src/webhooks/waiting-webhooks.ts
michael-radency: thanks @DawidMyslak , putting this to draft to test with proxy setup

PR #26137, opened 2026-02-23
feat(core): Support Chat hub on workflow canvas

Cadiac:
## Summary

Add support for new floating chat panel for manual chat hub executions directly on the canvas.

https://github.com/user-attachments/assets/48157361-6c6a-4c3e-9196-49727f4af667

When feature is enabled and chat trigger has "Make Available in n8n Chat Hub" enabled the chat doesn't open in the bottom left corner widget and instead this floating panel is used. For now this only works with chat hub - but this widget might become the default later.

Unlike the regular manual chat widget this one has history (personal, these sessions don't appear on regular chat hub), and it renders the agent as it would be shown on Chat hub once published for your instances other n8n users.

https://linear.app/n8n/issue/CHA-147

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/chat-hub/chat-hub-event-relay.service.ts
```diff
@@ -14,7 +14,7 @@ export class ChatHubEventRelay {
 	) {
 		this.logger = this.logger.scoped('chat-hub');
 
-		this.logger.info('initializing ChatHub event relay...');
+		this.logger.debug('initializing ChatHub event relay...');
```
Cadiac, line 17: this was maybe a bit noisy at info so I switched it to debug

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
```diff
@@ -158,6 +167,41 @@ export class ChatHubController {
...
 	}
 
+	/**
+	 * Send a message using the draft (unpublished) workflow version.
+	 * Requires workflow:execute — not available to chat-only users.
+	 * Passes pushRef header so the execution sends canvas events.
+	 */
+	@ProjectScope('workflow:execute')
+	@Post('/conversations/send/manual/:workflowId')
+	async sendMessageManual(
```
Cadiac, line 177: This works different to the regular manual executions, where the user provides the workflow to execute on the payload. This means that in order for this to work the workflow has to have been (auto)saved and we execute the latest saved draft - but that seems reasonable? I have a feeling that now that autosave is out we probably want to change the regular manual executions endpoint to work this way too, or can you think of cases where being able to pass in the workflow instead of the lastest draft is actually desired?

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
```diff
@@ -183,6 +227,40 @@ export class ChatHubController {
 		};
 	}
 
+	@ProjectScope('workflow:execute')
```
Cadiac, line 230: as these require different permissions I made these separate endpoints. Chat only users shouldn't be able to execute these.

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -479,46 +552,175 @@ export class ChatHubService {
...
+			message,
+			processedAttachments,
+		);
+	}
+
+	/**
+	 * Send a human message using the draft (unpublished) version of a workflow.
+	 * Requires workflow:execute permission. Passes pushRef for canvas execution events.
+	 */
+	async sendHumanMessageManual(
```
Cadiac, line 586: I'm not the biggest fan of basically replicating the logic of these by having second set of functions for the manual mode - but on the other hand I didn't want to introduce more complexity to the regular functions.

how do you feel, should I try to combine these?

packages/cli/src/modules/chat-hub/chat-hub.service.ts
```diff
@@ -666,6 +869,161 @@ export class ChatHubService {
...
+		}
+
+		if (!result?.workflow) {
+			return;
+		}
+
+		const { workflow } = result;
+
+		// Broadcast message edit to all user connections for cross-client sync
+		await this.chatStreamService.sendMessageEdit({
```
Cadiac, line 999: probably actually not so relevant here...

packages/frontend/editor-ui/src/app/composables/useClipboard.ts
```diff
@@ -10,11 +10,26 @@ export function useClipboard({
...
+		copied,
+		isSupported,
+		text,
+	} = useClipboardCore({
 		legacy: true,
 	});
 
+	// Find the correct navigator at copy-time so it works even when the
+	// pop-out window opens after the composable was created.
+	async function copy(value: string) {
```
Cadiac, line 24: This was necessary as otherwise the "Copy session ID" button would fail, as it tried to copy using the main window which wasn't focused.

packages/frontend/editor-ui/src/app/App.vue
```diff
@@ -88,6 +89,7 @@ const setLayoutRef = (el: Element) => {
 };
 
 useExposeCssVar('--toast--offset', toastBottomOffset);
+useExposeCssVar('--toast--right', toastRightOffset);
```
Cadiac, line 92: the toast would cover the chat on right panel. Initially this PR implemented chat on that panel so I added a mechanism to offset the toasts to not overlap the chat there, but then that panel got scrapped.

However I think keeping this mechanism is still a good call, we don't really want the toasts to overlap the AIWFB / assistant either, so kept this.

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatPrompt.vue
Cadiac, line 1: this is a bit of a mess, I'm not happy how much I had to change to support this new ... "compact" inline mode for the prompt. should I try to clean this up?

packages/frontend/editor-ui/src/features/ai/chatHub/composables/useChatSession.ts
Cadiac, line 1: I tried to extract the common logic from ChatView & this new chat into here to not have to repeat it all

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatPrompt.vue
autologie: Looks like most of the logic is common. Most approachable refactor would be something like
```
<ChatPromptCompact v-if="compact" />
<ChatPromptFull v-else />
```
does this work?

Cadiac: that could be the way to go, I'll look into making this better this morning

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
autologie: I think it makes sense, refactoring the existing endpoint would also be a nice simplification.

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
autologie: Given the specific use case and different permission model, having `/manual/:workflowId/` as the common namespace prefix for these new endpoints feels cleaner to me.

packages/frontend/editor-ui/src/app/App.vue
autologie: Nice!

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
```diff
@@ -385,7 +391,7 @@ export const useChatStore = defineStore(STORES.CHAT_HUB, () => {
 		try {
 			const cursor = reset ? undefined : (sessions.value?.nextCursor ?? undefined);
 			const [response] = await Promise.all([
-				fetchSessionsApi(rootStore.restApiContext, CHAT_SESSIONS_PAGE_SIZE, cursor),
+				fetchSessionsApi(rootStore.restApiContext, CHAT_SESSIONS_PAGE_SIZE, cursor, options.type),
 				new Promise((resolve) => setTimeout(resolve, options.minLoadingTime ?? 0)),
 			]);
```
autologie, line 406: I found a bug that after showing the session list in floating chat UI, conversation list isn't populated in ChatHub sidebar. Probably session list has to be separate state per type (manual or not).

Cadiac: thanks, I'll see if I can replicate this and fix that..

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatPrompt.test.ts
```diff
@@ -0,0 +1,259 @@
...
+			});
+
+			const textarea = getByRole('textbox');
+			await user.type(textarea, 'Compact message');
+			await user.keyboard('{Enter}');
+
+			expect(emitted().submit[0]).toEqual(['Compact message', []]);
+		});
+
+		it('shows missing agent callout in compact mode', () => {
```
autologie: I think callout is not relevant in compact mode

packages/frontend/@n8n/design-system/src/components/N8nFloatingWindow/FloatingWindow.vue
autologie, line 1: Cool, perhaps have a story for discoverability? We should definitely reuse this for similar features.

packages/frontend/@n8n/design-system/src/components/N8nFloatingWindow/FloatingWindow.vue
```diff
@@ -0,0 +1,416 @@
...
+			@mousedown="onResizeMouseDown('se', $event)"
+		/>
+		<div
+			:class="[$style.resizeHandle, $style.resizeSW]"
+			@mousedown="onResizeMouseDown('sw', $event)"
+		/>
+
+		<div :class="$style.header" @mousedown="onHeaderMouseDown">
+			<div :class="$style.headerLeft">
+				<slot name="header-icon" />
```
autologie, line 295: Can we add `defineSlots()`? That would help figuring out what slots are available.

packages/cli/src/modules/chat-hub/chat-hub.controller.ts
Cadiac: Changed the endpoints to have a common prefix,
```
POST /conversations/manual/:workflowId/send
POST /conversations/manual/:workflowId/:sessionId/messages/:messageId/edit
POST /conversations/manual/:workflowId/:sessionId/messages/:messageId/regenerate
```

packages/frontend/editor-ui/src/features/ai/chatHub/components/ChatPrompt.vue
Cadiac: Changed this, now ChatPrompt renders either the full or compact component. Callouts are implemented as a slot.

packages/cli/src/modules/chat-hub/chat-hub.service.ts
autologie: I think we can benefit from grouping arguments into logically related clusters. Repeated logic itself isn't that complicated, but since we pass many things around, it looks more bloated than it actually is. Of course something we can refactor in a separate PR

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
Cadiac: I think I fixed this, introducing a secondary state per type was rather verbose, instead I made this aware of the last type fetched and to start fresh if the type is different than before

autologie: Thanks, it seems to work

packages/frontend/@n8n/design-system/src/components/N8nFloatingWindow/FloatingWindow.vue
Cadiac: Added a story 👍

packages/frontend/@n8n/design-system/src/components/N8nFloatingWindow/FloatingWindow.vue
Cadiac: Added defineSlots

autologie (dismissed):
🚀

autologie (approved):

autologie (approved):

PR #26142, opened 2026-02-23
feat(core): Add IsolatedVmBridge (no-changelog)

despairblue:
## Context

n8n expressions today run via `tournament` (an AST transformer) inside the main Node.js process. The goal of this PR series is to make expression evaluation available inside **`isolated-vm`** — a V8 isolate that is memory-isolated from the host process. This gives us a proper security boundary: a malicious expression cannot access Node.js APIs, the filesystem, or the host's memory.

The series is being landed incrementally:

| PR | What it adds |
|----|-------------|
| #26047 ✅ | Package scaffold: public types and architecture docs |
| #26077 ✅ | The runtime bundle that runs _inside_ the isolate |
| **This PR** | `IsolatedVmBridge` — creates the isolate, loads the bundle, and exposes workflow data to it via callbacks |
| PR 4 (upcoming) | `ExpressionEvaluator` — the public API that drives the bridge, plus integration tests |

## What this PR adds

`IsolatedVmBridge` is the component that owns the V8 isolate. It:

1. **Creates a persistent isolate** (128 MB memory limit) in the constructor
2. **Loads the runtime bundle** (`dist/bundle/runtime.iife.js` from PR 2) into the isolate context on `initialize()`
3. **Registers three `ivm.Reference` callbacks** before each evaluation so the in-isolate lazy proxy can fetch workflow data from the host without pre-copying it:
   - `__getValueAtPath(path[])` — returns a primitive, array metadata `{__isArray, __length}`, object metadata `{__isObject, __keys}`, or function metadata
   - `__getArrayElement(path[], index)` — returns a single element (or its metadata) from an array
   - `__callFunctionAtPath(path[], ...args)` — invokes a function at the given path in the host data and returns the result
4. **Calls `resetDataProxies()`** in the isolate to initialise `$json`, `$binary`, `$input`, etc. as fresh lazy proxies backed by those callbacks
5. **Runs the tournament-transformed expression** with `this === __data`, so `this.$json.email` resolves through the proxy

**Arrays are always lazy** — only the length is transferred. Elements are fetched individually on demand via `__getArrayElement`. There is no small-array threshold.

**Script compilation is cached** — once an expression is compiled to an `ivm.Script`, it is stored in a `Map` keyed by the expression string. Subsequent evaluations of the same expression skip recompilation.

### Isolation setup in `initialize()`

Broken into four focused private methods, each independently verifiable in debug mode:

- `loadVendorLibraries()` — evaluates the runtime bundle in the isolate context and verifies `DateTime` and `extend` are available on `globalThis`
- `verifyProxySystem()` — checks that `createDeepLazyProxy`, `resetDataProxies`, `SafeObject`, `SafeError`, and `__data` are all present
- `injectErrorHandler()` — injects the `E()` global needed by tournament's try-catch wrapping: re-throws security violations, swallows `TypeError`s (failed prototype-attack attempts), re-throws everything else

### New dependency: `isolated-vm ^6.0.2`

`isolated-vm` is a native Node.js addon (compiled from C++ at install time). It is added to `pnpm.onlyBuiltDependencies` in the root `package.json` so `pnpm install` compiles it correctly.

## Why no tests in this PR?

The interesting behaviour of this bridge is the round-trip: isolate creates a proxy → proxy fires an `ivm.Reference` callback → callback navigates host-side data → value is copied back into the isolate. Unit tests with mocked callbacks would only verify that a mock was called, not the actual serialisation, `copy: true` transfer semantics, or the `__isArray`/`__isObject` metadata protocol.

Integration tests covering the full path (bridge creates isolate → loads runtime bundle → sets callbacks → expression evaluates to the correct result) are planned for PR 4, once `ExpressionEvaluator` exists to drive the bridge.

## Verification

```
cd packages/@n8n/expression-runtime
pnpm build       # tsc + esbuild — both succeed
pnpm typecheck   # passes
```

https://linear.app/n8n/issue/CAT-2311

## Review / Merge checklist

- [x] Tests included. (deferred to PR 4 — see note above)


packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,486 @@
...
+				};
+			}
+
+			// Primitive element
+			return element;
+		});
+
+		// Callback 3: Call function at path with arguments
+		// Used when expressions invoke functions from workflow data
+		const callFunctionAtPath = new ivm.Reference((path: string[], ...args: unknown[]) => {
```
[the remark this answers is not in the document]
despairblue: This will be completely locked down in a later PR.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,486 @@
...
+			let fn: unknown = data;
+			for (const key of path) {
+				fn = (fn as Record<string, unknown>)?.[key];
+			}
+
+			if (typeof fn !== 'function') {
+				throw new Error(`${path.join('.')} is not a function`);
+			}
+
+			// Execute function in host context
```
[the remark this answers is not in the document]
despairblue: fixed in 261ea387bd434989adee74f0c44286f8b3d0a597, but I'm not sure we need this to be honest, I may remove this in the future

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,486 @@
...
+	private async loadVendorLibraries(): Promise<void> {
+		if (!this.context) {
+			throw new Error('Context not initialized');
+		}
+
+		try {
+			// Load runtime bundle (includes vendor libraries + proxy system)
+			// Path: dist/bundle/runtime.iife.js
+			const runtimeBundlePath = path.join(__dirname, '../../dist/bundle/runtime.iife.js');
+			const runtimeBundle = fs.readFileSync(runtimeBundlePath, 'utf-8');
```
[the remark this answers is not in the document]
despairblue: fixed in 1390f37848d2934b8f61c07267e0fa48d6e991c1

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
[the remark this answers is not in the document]
despairblue: nevermind, fixed in b22efbd093d44c40a4c39215fa91891a568df5db, but proper checks will still be implemented later when we can test them properly

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,493 @@
...
+	// Maps expression code -> compiled ivm.Script
+	private scriptCache = new Map<string, ivm.Script>();
+
+	// Active ivm.Reference callbacks — released before each re-registration
+	// to prevent reference accumulation across execute() calls
+	private valueAtPathRef?: ivm.Reference;
+	private arrayElementRef?: ivm.Reference;
+	private callFunctionRef?: ivm.Reference;
+
+	constructor(config: BridgeConfig = {}) {
```
afitzek, line 39: Nit: I think in packages/@n8n/expression-runtime/src/types/bridge.ts we have a comment, that has these default values basically on an interface level. Does it make sense in this case to have a interface level default object, that is used by implementations to set up the default values?

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,493 @@
...
+				copy: true,
+			});
+
+			if (this.config.debug) {
+				console.log('[IsolatedVmBridge] Expression executed successfully');
+			}
+
+			return result;
+		} catch (error) {
+			const errorMessage = error instanceof Error ? error.message : String(error);
```
afitzek, line 451: In packages/@n8n/expression-runtime/src/types/evaluator.ts we define different types of errors MemoryLimitError, TimeoutError, etc. is the `script.runSync` supposed to throw these errors, or are we supposed to translate these errors here?

edit: Seems like there is a RuntimeError, that should be thrown here, translated from codes that are send from the bridge.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,493 @@
...
+			return value;
+		});
+
+		// Callback 2: Get array element at index
+		// Used by array proxy when accessing numeric indices
+		const getArrayElement = new ivm.Reference((path: string[], index: number) => {
+			// Navigate to array
+			let arr: unknown = data;
+			for (const key of path) {
+				arr = (arr as Record<string, unknown>)?.[key];
```
afitzek, line 324: Should we check here also of arr being undefined or null similar to the getValueAtPath loop?

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,493 @@
...
+	 * @private
+	 */
+	private registerCallbacks(data: Record<string, unknown>): void {
+		if (!this.context) {
+			throw new Error('Context not initialized');
+		}
+
+		// Callback 1: Get value/metadata at path
+		// Used by createDeepLazyProxy when accessing properties
+		const getValueAtPath = new ivm.Reference((path: string[]) => {
```
afitzek, line 277: It would be great if we could have unit tests for these functions. Can we have a function that creates a function for use with data in scope, something like:

`function getValueAtPath(data: Record<>) {
  return (path: string[]) => {....}
}`

Than we could write unit tests for these logic of these functions.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -0,0 +1,493 @@
...
+ *
+ * This bridge creates a V8 isolate with:
+ * - Hard memory limit (128MB default)
+ * - No access to Node.js APIs
+ * - Timeout enforcement
+ * - Complete isolation from host process
+ *
+ * Context reuse pattern: Create isolate/context once, reset state between evaluations.
+ */
+export class IsolatedVmBridge implements RuntimeBridge {
```
afitzek, line 22: It would be great if we could implement test cases for some expressions, also testing memory limit and runtime enforcement. 

Maybe we can develop a test suite against the RuntimeBridge and have it execute against different implementations in the future as well.

afitzek (approved):

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: done in 5ea4d0eddee4be4af8d08bbe4c06373cbb21afb2

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: done in 19f220f58460670cc635e3d0c890f2dac7c39054

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: done in 5deebd0a20e073c9598f41fab8c1e451b73b4d59

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: Most of this is excercised by integration tests in https://github.com/n8n-io/n8n/pull/26230

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: In PR 5 this will be integrated into into the workflow package and it will have to pass all tests, so we can re-use the expression tests there to prove correctness.

PR #26145, opened 2026-02-23
fix(editor): Add aria-label to icon-only buttons missing accessible labels

aalises:
## Summary

Adds `aria-label` attributes to all icon-only `N8nButton` and `N8nIconButton` instances that were missing accessible labels, eliminating the dev-mode console warning:

> [N8nButton] Icon-only buttons should have an accessible label. Add aria-label or title attribute.

All labels use i18n — no hardcoded strings. Labels were reviewed for accessibility quality (descriptive action verbs, context-specific where needed).


sandra0503 (dismissed):

konstantintieber (approved):
Great initiative!

PR #26148, opened 2026-02-23
fix(editor): Change node list shortcut from Tab to N to unlock keyboard navigation

Tuukkaa:
## Summary
- Changed the node creator panel shortcut from `Tab` to `N` to fix keyboard navigation being broken by the Tab key opening the nodes panel
- Added a coachmark tooltip that appears when users press Tab, informing them of the new shortcut
- The coachmark is dismissible and persists the dismissed state so it only shows once
- Removed Tab from watched keyboard navigation keys and the NodeCreator close handler
- Encapsulated coachmark event bus logic in a dedicated composable for clean separation of concerns

https://linear.app/n8n/issue/DS-457

## Review / Merge checklist
- [ ] Tests included.
- [ ] PR Labeled with release/backport (if the PR is an urgent fix that needs to be backported)


heymynameisrob (approved):
LGTM!

nit: Arguably could add test case for the keyboard shortcut but if there already isn't one think It's OK to leave it.

PR #26155, opened 2026-02-23
test: Add Helm chart E2E testing via K3s + testcontainers

shortstacked:
## Summary

- Deploys n8n via its Helm chart into a real K3s cluster (inside Docker via `@testcontainers/k3s`) and runs Playwright E2E tests against it
- Uses **NodePort routing** instead of `kubectl port-forward` for reliable traffic delivery under sustained test load
- Supports **standalone** (SQLite) and **queue** (PostgreSQL + Redis + workers) deployment modes
- Manual dispatch workflow with dropdowns for mode and chart ref

## New Files

| File | Purpose |
|------|---------|
| `packages/testing/containers/helm-stack.ts` | Core module: K3s lifecycle, image preloading, Helm install, NodePort routing |
| `packages/testing/containers/helm-start-stack.ts` | CLI wrapper (`pnpm stack:helm`) with `--mode`, `--image`, `--chart-ref` flags |
| `.github/workflows/test-e2e-helm.yml` | Manual dispatch workflow with K8s diagnostics on failure |
| `packages/testing/containers/HELM-TESTING.md` | Prerequisites, architecture diagram, troubleshooting guide |

## Why NodePort over port-forward

`kubectl port-forward` silently stops routing traffic after ~6 concurrent connections. The process stays alive, no error is emitted, but HTTP requests return nothing — causing cascade test timeouts. NodePort uses kube-proxy for stateless per-connection routing, which is reliable under sustained load.

## What this enables

This infrastructure validates Kubernetes-specific deployment concerns that can't be tested with testcontainers alone:

| Concern | Example | Why K8s is needed |
|---------|---------|-------------------|
| **readOnlyRootFilesystem** | npm cache/tmp dirs must redirect to writable volumes ([#26032](https://github.com/n8n-io/n8n/pull/26032)) | SecurityContext is a K8s-only concept |
| **Shared PVC for community nodes** | Mount a ReadWriteMany PVC at the community nodes path so install-once is available across all pods in queue mode | PVC access modes, mount propagation, and multi-pod sharing only exist in K8s |
| **Helm chart value correctness** | Verify that chart defaults actually produce a working deployment | Catches misconfigured env vars, missing secrets, bad resource limits |
| **Queue mode pod coordination** | Workers pick up jobs, webhooks route correctly, leader election works | Requires real K8s service discovery and DNS |
| **Security contexts & RBAC** | Non-root UID, dropped capabilities, read-only filesystem | Enforced by kubelet, not Docker |

### Future: Playwright fixture integration

The `createHelmStack` API is designed to plug into the existing Playwright fixture system (same pattern as chaos tests). A test file could use `test.use({ capability: { helmMode: 'queue', helmOverrides: { ... } } })` to spin up a Helm stack with custom overrides per test suite — enabling infrastructure-level E2E tests alongside existing feature tests.

## Test plan

- [ ] Run manual dispatch with `standalone` mode — verify 35/35 building-blocks tests pass
- [ ] Run manual dispatch with `queue` mode — verify deployment succeeds (queue mode has tighter resource constraints)
- [ ] Local: `pnpm stack:helm --image n8nio/n8n:latest` starts stack, Ctrl+C cleans up


shortstacked:
> @shortstacked this is still in draft, but you asked for review. Is it ready?

Hey @krider2010, yeah it's ready for review. I kept it as draft since it's referencing your charts on a branch, but also I wanted your thoughts on the approach here of testing it with postgres/redis bundled in. Our official charts don't have it, but it's the only way I can make it testable.

packages/testing/containers/helm-stack.ts
```diff
@@ -0,0 +1,405 @@
...
+
+	// License env vars from host (same pattern as testcontainers stack)
+	let envIdx = 5;
+	const licenseTenantId = process.env.N8N_LICENSE_TENANT_ID;
+	if (licenseTenantId) {
+		flags.push(
+			`--set config.extraEnv[${envIdx}].name=N8N_LICENSE_TENANT_ID --set-string config.extraEnv[${envIdx}].value=${licenseTenantId}`,
+		);
+		envIdx++;
+	}
```
krider2010: Just a comment that adding an env var without updating this `envIdx` this would break.

Something we could consider doing perhaps?

```
const extraEnvs: Array<{ name: string; value: string }> = [
    { name: 'E2E_TESTS', value: 'true' },
    { name: 'NODE_ENV', value: 'development' },
    // ...
];
if (licenseTenantId) extraEnvs.push({ name: 'N8N_LICENSE_TENANT_ID', value: licenseTenantId });

const envFlags = extraEnvs.flatMap((env, i) => [
    `--set config.extraEnv[${i}].name=${env.name}`,
    `--set-string config.extraEnv[${i}].value=${env.value}`,
]);
```

packages/testing/containers/package.json
```diff
@@ -18,6 +18,8 @@
     "stack:clean:containers": "docker ps -aq --filter 'name=n8n-stack-*' | xargs -r docker rm -f 2>/dev/null",
     "stack:clean:networks": "docker network ls --filter 'label=org.testcontainers=true' -q | xargs -r docker network rm 2>/dev/null",
     "stack:clean:all": "pnpm run stack:clean:containers && pnpm run stack:clean:networks",
+    "stack:helm": "tsx ./helm-start-stack.ts",
+    "stack:helm:clean": "docker ps -aq --filter 'name=*k3s*' | xargs -r docker rm -f 2>/dev/null",
```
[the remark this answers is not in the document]
krider2010: Thanks cubic! I was also going to flag this.

.github/workflows/test-e2e-helm.yml
```diff
@@ -0,0 +1,183 @@
...
+      - name: Run Building Blocks E2E Tests
+        working-directory: packages/testing/playwright
+        run: |
+          N8N_BASE_URL="${{ env.N8N_BASE_URL }}" \
+          RESET_E2E_DB=true \
+          npx playwright test \
+            --project=e2e \
+            tests/e2e/building-blocks/ \
+            --workers=1 \
+            --retries=0 \
```
krider2010: `--workers=1` makes sense for a single-pod deployment. But `-retries=0` is unusual for E2E. Is this intentional to surface real issues during early development?

.github/workflows/test-e2e-helm.yml
```diff
@@ -0,0 +1,183 @@
...
+          IMAGE_TAG: ci-${{ github.run_id }}
+          RUNNERS_IMAGE_BASE_NAME: ghcr.io/${{ github.repository_owner }}/runners
+
+  helm-e2e:
+    name: "Helm E2E (${{ inputs.mode || 'queue' }})"
+    needs: build
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    timeout-minutes: 30
+    permissions:
+      contents: read
```
krider2010, line 70: Just again possibly worth noting, that this works for now as the images are public. If we ever change that we'd need to update here.

krider2010 (dismissed):
Just some comments. Be interesting to see this up and running!

.github/workflows/test-e2e-helm.yml
shortstacked: I'll add a retry for transient issues that come up, but hopefully it won't be needed.

.github/workflows/test-e2e-helm.yml
shortstacked: Yeah that make sense, if that's the case we can swap over and pull the tests from the other repo instead.

krider2010 (approved):
Nice!

PR #26167, opened 2026-02-24
feat(Chat Node): Return chat message instead of input data if Chat is not waiting for user input

michael-radency:
## Summary

Added new node version 1.3 - if node is not waiting, a message will be returned instead of input data as in earlier versions

https://linear.app/n8n/issue/NODE-4041/the-respond-to-chat-output-seems-to-be-wrong

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/__test__/Chat.node.test.ts
+++ b/packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/__test__/Chat.node.test.ts
@@ -280,6 +280,34 @@ describe('Test Chat Node', () => {
 			expect(memory.chatHistory.addUserMessage).toHaveBeenCalledWith('user message');
 		});
 
+		it('v1.3 should return [[data]] (original message) instead of input data when not waiting for reply', async () => {
+			const chatNode = mock<INode>({
+				name: 'Chat',
+				type: CHAT_NODE_TYPE,
+				parameters: {},
+				typeVersion: 1.3,
+			});
+			const message = { json: { chatInput: 'user message' } };
+			const differentInputData = [{ json: { chatInput: 'other input' } }];
+			mockExecuteFunctions.getInputData.mockReturnValue(differentInputData);
+			mockExecuteFunctions.getNode.mockReturnValue(chatNode);
+			mockExecuteFunctions.getNodeParameter.mockImplementation((parameterName) => {
+				switch (parameterName) {
+					case 'operation':
+						return 'send';
+					case 'options':
+						return { memoryConnection: false };
+					default:
+						return undefined;
+				}
+			});
+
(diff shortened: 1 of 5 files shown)


packages/cli/src/chat/utils.ts
```diff
@@ -70,3 +70,12 @@ export function shouldResumeImmediately(lastNode: INode) {
 
 	return false;
 }
+
+export function getLastNodeMessage(execution: IExecutionResponse, lastNode: INode) {
+	if (lastNode.type !== CHAT_NODE_TYPE) return '';
```
ShireenMissi, line 75: should this handle CHAT_TOOL_NODE_TYPE as well?

packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/__test__/Chat.node.test.ts
```diff
@@ -280,6 +280,34 @@ describe('Test Chat Node', () => {
...
+						return { memoryConnection: false };
+					default:
+						return undefined;
+				}
+			});
+
+			const result = await chat.onMessage(mockExecuteFunctions, message);
+
+			expect(result).toEqual([[message]]);
+			expect(result).not.toEqual([differentInputData]);
```
ShireenMissi: This is not needed as we already asserting the result equals [[message]]
```suggestion
```

ShireenMissi (changes requested):

packages/cli/src/chat/utils.ts
michael-radency: no, there is separate handling for tool

packages/cli/src/chat/chat-service.ts
```diff
@@ -150,7 +155,7 @@ export class ChatService {
 			session.connection.send(N8N_CONTINUE);
 			const data: ChatMessage = {
 				action: 'sendMessage',
-				chatInput: '',
+				chatInput: getLastNodeMessage(execution, lastNode),
```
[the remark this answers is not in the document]
michael-radency: chat input should be shown to user, currently in the chat only, this mimics behavior of the node when chat is waiting for user input mode

packages/cli/src/chat/utils.ts
```diff
@@ -70,3 +70,12 @@ export function shouldResumeImmediately(lastNode: INode) {
 
 	return false;
 }
+
+export function getLastNodeMessage(execution: IExecutionResponse, lastNode: INode) {
```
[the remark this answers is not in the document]
michael-radency: message is added to the node's output, it was already displayed in the chat, this make send message behave consistently with send and wait operation

ShireenMissi (approved):
I tested locally and it works as expected 👌

PR #26170, opened 2026-02-24
chore(core): Enforce golden rules via linting and structural tests (no-changelog)

aalises:
## Summary

Applies [harness engineering](https://openai.com/index/harness-engineering/) principles to the n8n codebase: instead of relying on agents/developers reading documentation, the **lint rules catch violations and the error messages teach the fix**. The agent writes code, lints, reads the error, and self-corrects. No docs needed.

### What changed

**Fixed existing rules:**
- **`no-plain-errors`**: was auto-fixing `Error` → `ApplicationError`, directly contradicting AGENTS.md which says `ApplicationError` is deprecated. Now detects both `throw new Error()` and `throw new ApplicationError()`, removed auto-fix, error messages guide toward `UnexpectedError`/`OperationalError`/`UserError`. Also added to the plugin recommended config (previously missing from many packages).

**Promoted existing rules:**
- **`no-explicit-any`**: promoted from `warn` to `error` in base config (all packages). Was only `error` in frontend.
- **`consistent-type-assertions`**: changed from style-only enforcement to `assertionStyle: 'never'` as `warn` for non-test files. Effectively new enforcement of the "no `as` casting" principle.

**New lint rules:**
- **`no-invalid-data-testid`** (`error`): catches spaces in `data-testid` attribute values in Vue templates
- **`no-hardcoded-ui-text`** (`warn`): catches hardcoded strings in Vue template UI attributes (`label`, `placeholder`, `title`, etc.) instead of i18n calls
- **`no-cross-boundary-import`** (`error`): enforces the package dependency flow at lint time. Fires immediately when an import crosses an architectural boundary (e.g., `nodes-base` importing from `n8n-core`, `editor-ui` importing from `n8n-core`). Centralized boundary map covering 6 package layers.

**New structural test:**
- **`architecture.test.ts`**: validates all package boundaries in one pass. Found 3 real pre-existing violations in `nodes-base` (importing from `n8n-core`) that are baselined for cleanup.

**New documentation:**
- **`docs/principles/golden-rules.md`**: single reference for all 9 golden rules with enforcement status, fix instructions, and the dependency flow diagram
- **AGENTS.md refactored**: replaced scattered inline principles with consolidated "Golden Rules" section linking to the reference doc

### Enforcement summary

| # | Rule | Severity | Enforcement |
|---|------|----------|-------------|
| 1 | No `any` | `error` | `@typescript-eslint/no-explicit-any` |
| 2 | No `as` casting | `warn` | `@typescript-eslint/consistent-type-assertions` |
| 3 | Shared types in api-types | convention | — |
| 4 | No ApplicationError | `error` | `n8n-local-rules/no-plain-errors` |
| 5 | i18n for UI text | `warn` | `n8n-local-rules/no-hardcoded-ui-text` |
| 6 | CSS variables only | convention | — |
| 7 | Single data-testid | `error` | `n8n-local-rules/no-invalid-data-testid` |
| 8 | Traversal utilities | convention | — |
| 9 | Package boundaries | `error` | `n8n-local-rules/no-cross-boundary-import` + `architecture.test.ts` |

### How it works for agents

The key insight: **lint error messages ARE the agent's documentation**. When a rule fires, the message tells you what's wrong, why, and how to fix it:

```
error  Do not throw `ApplicationError` directly — it is deprecated.
       Use `UnexpectedError`, `OperationalError`, or `UserError` instead.
       Import from the appropriate package. See AGENTS.md for guidance.
```

```
error  Do not import "n8n-core" from this package — it violates the
       architectural boundary. See docs/principles/golden-rules.md
       for the expected dependency flow.
```

The agent doesn't need to read AGENTS.md first, as the guardrails catch it and teach the fix in the error message itself.


docs/principles/golden-rules.md
```diff
@@ -0,0 +1,135 @@
...
+> wrong, why, and how to fix it. Follow the message.
+
+---
+
+## 1. No `any` type
+
+**Use proper types or `unknown` — never `any`.**
+
+- **Enforced by:** `@typescript-eslint/no-explicit-any` = `error` in base config
+- **Applies to:** All packages (frontend and backend)
```
shortstacked: Could we exclude tests from this, sometimes to write types just for testing can be counter productive, especially in the e2e tests.

packages/@n8n/api-types/src/dto/workflows/base-workflow.dto.ts
```diff
@@ -38,6 +38,7 @@ export const workflowStaticDataSchema = z.preprocess(
 			try {
 				return JSON.parse(val);
 			} catch {
+				// eslint-disable-next-line n8n-local-rules/no-plain-errors -- Zod preprocessor catches thrown errors internally
```
shortstacked: Is this something we could use a baseline for instead? https://web.dev/blog/eslint-baseline-integration

packages/@n8n/crdt/eslint.config.mjs
```diff
@@ -1,4 +1,9 @@
 import { defineConfig } from 'eslint/config';
 import { nodeConfig } from '@n8n/eslint-config/node';
 
-export default defineConfig(nodeConfig);
+export default defineConfig(nodeConfig, {
+	rules: {
+		// Low-level CRDT package — no dependency on n8n-workflow where error classes live
+		'n8n-local-rules/no-plain-errors': 'off',
```
shortstacked: I think this is still valid though. How will the CRDT distinguish between errors that get sent through to Sentry or ones that don't?

packages/@n8n/eslint-config/src/architecture.test.ts
```diff
@@ -0,0 +1,191 @@
+/**
+ * Structural test: Package Boundary Enforcement
```
shortstacked, line 2: When does this test get run/enforced?

aalises: on the `Backend Unit Tests` job afaik

packages/@n8n/api-types/src/dto/workflows/base-workflow.dto.ts
aalises: Good idea! will make it use a baseline

packages/@n8n/crdt/eslint.config.mjs
aalises: will move them to warn!

shortstacked (dismissed):
🔥

PR #26171, opened 2026-02-24
chore(editor): Improve workflow builder enter/exit animation

Tuukkaa:
## Summary

Improves the AI assistant sidebar open/close animation to be smoother and less janky:

- **Width-based animation**: Replaced the `SlideTransition` component with a custom `Transition` using `width` property for smoother enter/exit, with separate durations (175ms enter, 150ms exit)
- **Deferred grid resize**: Delayed updating `appGridDimensions` until after the slide-in animation completes, preventing the canvas from resizing mid-animation
- **Overflow hidden on aside**: Added `overflow: hidden` to the `.aside` layout element to prevent content from visually overflowing during the animation
- **Simplified toggle logic**: Collapsed a redundant `if/else` in `toggleAssistantMode` into a single call using the already-computed `newMode`

https://linear.app/n8n/issue/AI-2093

## Review / Merge checklist

- [ ] Tests included.


packages/frontend/editor-ui/src/app/constants/ai.ts
```diff
@@ -1,7 +1,8 @@
 export const ASK_AI_MAX_PROMPT_LENGTH = 600;
 export const ASK_AI_MIN_PROMPT_LENGTH = 15;
 export const ASK_AI_LOADING_DURATION_MS = 12000;
-export const ASK_AI_SLIDE_OUT_DURATION_MS = 200;
+export const ASK_AI_SLIDE_IN_DURATION_MS = 175;
+export const ASK_AI_SLIDE_OUT_DURATION_MS = 150;
```
burivuhster, line 5: Should this be unified with e.g. nodes panel slide animation?

burivuhster (approved):
Looks good! Unification of all sliding panels may be postponed as a tech debt

packages/frontend/editor-ui/src/app/constants/ai.ts
Tuukkaa: Yeah we could do that although typically duration is chosen based on how long distance an element travels. We might want to make the builder panel a much wider in the future since it becomes the main way of building things. Wider panel needs longer duration so coupling the two together might not be ideal

burivuhster: Good point

PR #26176, opened 2026-02-24
fix(core): Resolve multi-main startup race condition in AuthRolesService

afitzek:
## Summary

Fixes a multi-main startup race condition where concurrent n8n instances running `AuthRolesService.syncScopes()` against the same Postgres database hit duplicate key constraint violations.

- Introduces a centralized `DbLockService` in `@n8n/db` that wraps Postgres advisory locks (`pg_advisory_xact_lock`) inside transactions, with a `DbLock` enum to prevent lock ID collisions
- Supports an optional `timeoutMs` on `withLock` (via `SET LOCAL lock_timeout`) and a non-blocking `tryWithLock` (via `pg_try_advisory_xact_lock`) that fails fast if the lock is already held
- On SQLite, advisory locks are skipped — the transaction alone provides serialization
- Refactors `AuthRolesService.init()` to use `DbLockService.withLock(DbLock.AUTH_ROLES_SYNC, ...)` instead of repository-level calls
- Removes the leader-only guard in `start.ts` — all main instances now call `AuthRolesService.init()`, and the advisory lock serializes them safely
- Lock contention throws `OperationalError` (timeout exceeded or try-lock busy)

## Test plan

- [x] 10 unit tests for `DbLockService` (`withLock` + `tryWithLock`, Postgres/SQLite, timeout, error propagation)
- [x] 8 integration tests against live database (transaction behavior, advisory lock serialization, timeout, try-lock contention)
- [x] 27 existing `AuthRolesService` unit tests updated and passing
- [x] 4 existing `start.ts` unit tests updated (non-leader now calls init)
- [x] Typecheck and lint clean on both `@n8n/db` and `cli` packages

closes https://linear.app/n8n/issue/IAM-298

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/services/auth.roles.service.ts
+++ b/packages/@n8n/db/src/services/auth.roles.service.ts
@@ -12,21 +12,23 @@ import {
 
 // eslint-disable-next-line n8n-local-rules/misplaced-n8n-typeorm-import
 import { In } from '@n8n/typeorm';
+import type { EntityManager } from '@n8n/typeorm';
 
-import { Scope } from '../entities';
-import { RoleRepository, ScopeRepository, SettingsRepository } from '../repositories';
+import { Role, Scope, Settings } from '../entities';
+import { DbLock, DbLockService } from './db-lock.service';
 
 @Service()
 export class AuthRolesService {
 	constructor(
 		private readonly logger: Logger,
-		private readonly scopeRepository: ScopeRepository,
-		private readonly roleRepository: RoleRepository,
-		private readonly settingsRepository: SettingsRepository,
+		private readonly dbLockService: DbLockService,
 	) {}
 
-	private async syncScopes() {
-		const availableScopes = await this.scopeRepository.find({
+	private async syncScopes(tx: EntityManager) {
+		const scopeRepo = tx.getRepository(Scope);
+		const roleRepo = tx.getRepository(Role);
+
+		const availableScopes = await scopeRepo.find({
 			select: {
 				slug: true,
 				displayName: true,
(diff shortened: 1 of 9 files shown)


packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -222,8 +228,10 @@ export class AuthRolesService {
 
 	async init() {
 		this.logger.debug('Initializing AuthRolesService...');
-		await this.syncScopes();
-		await this.syncRoles();
+		await this.dbLockService.withLock(DbLock.AUTH_ROLES_SYNC, async (tx) => {
```
guillaumejacquart, line 231: Why not use the tryWithLock ? I feel like it would be faster, and would fail only if another main instance is doing the job ? 
Also, why no catching the `OperationalError` in case of timeout ? Do we want this to prevent the instance from starting ?

guillaumejacquart (commented):
Just a few questions

packages/@n8n/db/src/services/auth.roles.service.ts
afitzek: The `OperationalError` is only thrown when the timeout parameter is set, which it is not in this call. The thinking for not using tryWithLock is that I want all instances to sync there changes to avoid any potential miss behaviors. The downside is that instances might wait for each other if they reach this point at the same time, but since this is a few ms, it should be fine during the bootstrap process.

packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -222,8 +228,10 @@ export class AuthRolesService {
...
+			await this.syncScopes(tx);
```
phyllis-noester, line 232: just to make sure I understand: the reason why we want to do this whenever a new instance starts is that the instance start could be related to a deployment that either added or removed scopes. correct?

afitzek: Yes correct, this approach safes us from adding a DB migration for every scope that we want to add.

phyllis-noester: given that we don't do rolling updates, there is no scenario where there could be a race condition and the scopes remain in the old state right?
I was trying to think of edge cases e.g.:
- a container restarts for non deployment related reason
- we deploy (miliseconds after)
- container restart locks the table
- deploy does not sync roles
=> scopes remain in old state


but I assume that would only be possible if we did rolling updates and even then it would be super unlikely

afitzek: If the container restarts it would acquire the advisory lock, and sync its roles, the deployment (the new versions), would reach this point and wait for the advisory lock to be released. One of the new instances would acquire the lock and sync its roles, the other still wait. So one after another passes this point.

phyllis-noester: ah  i see, that's essentially the answer to guillaumes question :)

guillaumejacquart (approved):

PR #26180, opened 2026-02-24
refactor(core): Extract and reuse containsExpression string utility

alielkhateeb:
## Summary

Extracts the `containsExpression` string-checking function into a shared utility in `packages/cli/src/utils.ts` and reuses it in two places:

- **`source-control-helper.ee.ts`**: removes the local `stringContainsExpression` function and replaces calls with the shared utility
- **`redaction.service.ee.ts`**: replaces the inline `!value.startsWith('=')` check (which only tested the leading `=`) with `containsExpression`, making the expression-detection logic consistent across both modules

Tests updated to cover the new exported utility.

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/__tests__/utils.test.ts
+++ b/packages/cli/src/__tests__/utils.test.ts
@@ -6,6 +6,7 @@ import {
 	getAllKeyPaths,
 	isWorkflowIdValid,
 	setMicrosoftObservabilityDefaults,
+	containsExpression,
 } from '../utils';
 
 describe('shouldAssignExecuteMethod', () => {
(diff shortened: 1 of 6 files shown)


packages/cli/src/__tests__/utils.test.ts
```diff
@@ -331,3 +332,51 @@ describe('setMicrosoftObservabilityDefaults', () => {
 		expect(process.env.ENABLE_A365_OBSERVABILITY_EXPORTER).toBe('true');
 	});
 });
+
+describe('containsExpression', () => {
```
ireneea, line 336: Nice 👍

ireneea (approved):
Love this 🚀

PR #26188, opened 2026-02-24
fix(Notion Node): Skip fetching children of unsupported block types

romeobalta:
## Summary

The Notion API returns meeting transcription blocks as `type: "unsupported"` with
`has_children: true`. When "Fetch Nested Blocks" is enabled in the "Get Many Child Blocks"
action, the node tries to recursively fetch children of all blocks, including unsupported ones.
The API rejects these requests, causing the entire execution to fail.

Add `unsupported` to the skip list alongside `child_page` in `notionApiRequestGetBlockChildrens`,
so these blocks are still included in the output but their children are not recursively fetched.

## Review / Merge checklist

- [x] Tests included. 


ShireenMissi (approved):
Tested locally and it works as expected ✅

PR #26189, opened 2026-02-24
fix: Enable external secrets multi-connection to vault 

ireneea:
## Summary

Replaces `useEnvFeatureFlag` checks (`EXTERNAL_SECRETS_FOR_PROJECTS`, `EXTERNAL_SECRETS_MULTIPLE_CONNECTIONS`) with backend-driven module settings exposed via `settingsStore.moduleSettings['external-secrets']`.

## How to test

**env vars**
- `N8N_ENV_FEAT_EXTERNAL_SECRETS_FOR_PROJECTS`
- `N8N_ENV_FEAT_EXTERNAL_SECRETS_MULTIPLE_CONNECTIONS`

### Without env vars  

1. Start n8n without any of the environment variable listed above set
2. Check that multi-connection is enabled by default  

- The new external external settings page should be displayed by default
- You should be able to create a new connection
- You should be able to use connection secrets in credentials

### With  `N8N_ENV_FEAT_EXTERNAL_SECRETS_MULTIPLE_CONNECTIONS` set to `false`

1. Start n8n with `N8N_ENV_FEAT_EXTERNAL_SECRETS_MULTIPLE_CONNECTIONS` set to `false`
2. Check that multi-connection is disabled

- You should see the old external settings page (with the five static providers and the "set up" button)
- You should still be able to use secrets in credentials

### With  `N8N_ENV_FEAT_EXTERNAL_SECRETS_FOR_PROJECTS` set to `true`

1. Start n8n with `N8N_ENV_FEAT_EXTERNAL_SECRETS_FOR_PROJECTS` set to `false`
2. Check that project-scope feature is enabled

- You should be able to share external secrets with project and see the external secrets section in the project settings tab

https://linear.app/n8n/issue/LIGO-280

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectExternalSecrets.vue
+++ b/packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectExternalSecrets.vue
@@ -7,7 +7,7 @@ import { useSecretsProvidersList } from '@/features/integrations/secretsProvider
 import type { SecretProviderConnection } from '@n8n/api-types';
 import { useUIStore } from '@/app/stores/ui.store';
 import { SECRETS_PROVIDER_CONNECTION_MODAL_KEY, VIEWS } from '@/app/constants';
-import { useEnvFeatureFlag } from '@/features/shared/envFeatureFlag/useEnvFeatureFlag';
+import { useSettingsStore } from '@/app/stores/settings.store';
 import { useRouter } from 'vue-router';
 
 import {
@@ -30,9 +30,9 @@ const router = useRouter();
 const projectsStore = useProjectsStore();
 const uiStore = useUIStore();
 const rbacStore = useRBACStore();
+const settingsStore = useSettingsStore();
 const secretsProviders = useSecretsProvidersList();
 const secretsProviderConnection = useSecretsProviderConnection();
-const envFeatureFlag = useEnvFeatureFlag();
 
 interface ConnectionRow {
 	id: string;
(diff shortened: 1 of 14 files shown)


packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectExternalSecrets.vue
```diff
@@ -50,8 +50,8 @@ const expandedConnections = ref<Set<string>>(new Set());
 const currentPage = ref(0);
 const itemsPerPage = ref(5);
 
-const isFeatureEnabled = computed(() =>
-	envFeatureFlag.check.value('EXTERNAL_SECRETS_FOR_PROJECTS'),
+const isFeatureEnabled = computed(
+	() => settingsStore.moduleSettings['external-secrets']?.forProjects ?? false,
```
konstantintieber, line 54: Do you know how I would change these moduleSettings on a cloud instance once I ssh into it?
Asking since I know how to do this with an env var but not with moduleSettings.

konstantintieber: Ah nevermind, I see you're wiring up the env vars with module settings in `packages/cli/src/modules/external-secrets.ee/external-secrets.module.ts -> settings()`

ireneea: Module settings are run time logic based so you can't directly affect their values.

In this case the value is based on our backend logic which is 

- if no env var -> return `true`
- if there is an env var -> use the env var value

So you could still set the env var to inderectly affect the module setting value

konstantintieber (approved):
Learned something new about our backend today. Thx for migrating from pure env vars to moduleSettings!
Really handy pattern.

PR #26191, opened 2026-02-24
fix(editor): Resolve credential modal expressions without workflow co…

sandra0503:
…ntext

## Summary

When the credentials modal is opened from the NDV (which sets an active node), expressions like $secrets were incorrectly resolved using the workflow context instead of resolveWithoutWorkflow. This happened because the condition only checked for the absence of an active node, which is never true when the modal is opened from the NDV.

Add an explicit isCredentialsModalOpen() check so credential expressions always use resolveWithoutWorkflow, regardless of whether an active node exists.

Screenshot 2026-02-24 at 14 38 23

Fixes [LIGO-241](https://linear.app/n8n/issue/LIGO-241/fe-bug-bash-secrets-that-are-resolved-show-as-red)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/shared/editors/composables/useExpressionEditor.ts
+++ b/packages/frontend/editor-ui/src/features/shared/editors/composables/useExpressionEditor.ts
@@ -32,6 +32,7 @@ import { highlighter } from '../plugins/codemirror/resolvableHighlighter';
 import { closeCursorInfoBox } from '../plugins/codemirror/tooltips/InfoBoxTooltip';
 import type { Html, Plaintext, RawSegment, Resolvable, Segment } from '@/app/types/expressions';
 import { getExpressionErrorMessage, getResolvableState } from '@/app/utils/expressions';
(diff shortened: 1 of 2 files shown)


geemanjs (approved):

packages/frontend/editor-ui/src/features/shared/editors/composables/useExpressionEditor.ts
```diff
@@ -354,7 +355,10 @@ export const useExpressionEditor = ({
 					...expressionLocalResolveContext.value,
 					additionalKeys: toValue(additionalData),
 				});
-			} else if (!ndvStore.activeNode && toValue(targetNodeParameterContext) === undefined) {
+			} else if (
+				isCredentialsModalOpen() ||
+				(!ndvStore.activeNode && toValue(targetNodeParameterContext) === undefined)
```
konstantintieber, line 360: So the second condition here describes the case of `isNodeDetailsViewOpen`?

PR #26228, opened 2026-02-25
feat(editor): Add favoriting for projects, folders, workflows and data tables

CharlieKolb:
## Summary

Add the ability to favorite resources. Discoverability is poor right now, but we'll adress that in follow-ups

n/a

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/favorites/database/entities/user-favorite.entity.ts
```diff
@@ -0,0 +1,22 @@
+import { User } from '@n8n/db';
+import { Column, Entity, Index, ManyToOne, PrimaryGeneratedColumn, Unique } from '@n8n/typeorm';
+
+@Entity('user_favorites')
+@Unique(['userId', 'resourceId', 'resourceType'])
+@Index(['userId', 'resourceType'])
```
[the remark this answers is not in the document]
CharlieKolb: We want to optimize reads, not writes - page loads and fetches are very frequent compared to deletes

RicardoE105 (commented):
Nice work 🚀 

Couple of comments:  

issue (blocking): Missing DB index for resource cleanup queries

  deleteByResourceId and deleteByResourceIds query WHERE resourceId = ? AND resourceType = ? without a userId. The
  migration only has an index on userId and a unique constraint on (userId, resourceId, resourceType) — neither covers
  this query pattern.

  These cleanup queries run on every workflow, project, folder, and data table deletion (via the event relay). On
  instances with many users and favorites, this means a full table scan per resource deletion.

   .withIndexOn('userId')
  +.withIndexOn(['resourceType', 'resourceId'])
   .withUniqueConstraintOn(['userId', 'resourceId', 'resourceType']);

  ---
  suggestion (non-blocking): Parallelize independent queries in getEnrichedFavorites

  After findByUser and getAccessibleProjects complete (which must be sequential), the workflow, data table, and folder
  enrichment blocks are independent of each other. Currently they run sequentially (up to 4 more awaits). These could run
   in parallel:

  const [workflows, sharedWorkflows, dataTables, folders] = await Promise.all([
      workflowIds.length > 0 ? this.workflowRepository.findByIds(...) : [],
      workflowIds.length > 0 ? this.sharedWorkflowRepository.find(...) : [],
      dataTableIds.length > 0 ? this.dataTableRepository.find(...) : [],
      folderIds.length > 0 ? this.folderRepository.find(...) : [],
  ]);

  ---
  nitpick: star was added to both updatedIconSet and deprecatedIconSet in icons.ts. Per AGENTS.md, new code should only
  use updatedIconSet. The deprecatedIconSet entry is unnecessary.

  ---
  thought: If another user deletes a resource that the current user has favorited, the event relay cleans the DB but the
  frontend store retains the stale entry. The user sees "Unfavorite", clicks it, and the API returns 404 (which shows as
  a toast). An idempotent delete (return success even if not found) or try/catch in toggleFavorite would handle this more
   gracefully.

CharlieKolb:
/size-limit-override

This PR was already mostly reviewed and only needed feedback addressed - will split up future ones more 👍

RicardoE105:
1.	Is it intentional that credentials cannot be marked as favorites?
2.	If a workflow is marked as a favorite, what happens when its original project is deleted and the workflow is moved to another project? Should it remain marked as a favorite in the table?

packages/@n8n/db/src/migrations/common/1776150756000-CreateFavoritesTable.ts
```diff
@@ -0,0 +1,25 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
```
RicardoE105, line 1: looks like we need to update the timestamp here

packages/cli/src/modules/favorites/favorites.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+	constructor(private readonly favoritesService: FavoritesService) {}
+
+	@Get('/')
+	async getFavorites(req: AuthenticatedRequest) {
+		return await this.favoritesService.getEnrichedFavorites(req.user.id);
+	}
+
+	@Post('/')
+	async addFavorite(req: AuthenticatedRequest, _res: unknown, @Body body: AddFavoriteDto) {
+		return await this.favoritesService.addFavorite(req.user.id, body.resourceId, body.resourceType);
```
RicardoE105, line 25: question: should we validate the resource exist before inserting it? else you can make a post here with anything and save it to the database.

typecheck.log
```diff
@@ -0,0 +1,2265 @@
+
+> n8n-monorepo@2.16.0 typecheck /Users/charliekolb/dev/n8n
```
RicardoE105: I guess it was an overlook to commit this

packages/frontend/editor-ui/vite.config.mts
```diff
@@ -34,6 +34,7 @@ const alias = [
 	},
 	// Ensure bare imports resolve to sources (not dist)
 	{ find: '@n8n/i18n', replacement: resolve(packagesDir, 'frontend', '@n8n', 'i18n', 'src') },
+	{ find: '@n8n/stores', replacement: resolve(packagesDir, 'frontend', '@n8n', 'stores', 'src') },
```
RicardoE105: this change should probably not be part of this PR

packages/cli/src/modules/favorites/favorites.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	@Delete('/:resourceType/:resourceId')
+	async removeFavorite(
+		req: AuthenticatedRequest,
+		_res: unknown,
+		@Param('resourceId') resourceId: string,
+		@Param('resourceType') resourceType: string,
+	) {
+		if (!isFavoriteResourceType(resourceType)) {
```
RicardoE105, line 35: Should this validation be in the DTO?

packages/frontend/editor-ui/src/app/stores/favorites.store.ts
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	function reset() {
+		favorites.value = [];
+		initialized.value = false;
+	}
+
+	return {
+		favorites,
+		favoriteIds,
```
RicardoE105: This is exported but is it used?

packages/@n8n/db/src/migrations/common/1776150756000-CreateFavoritesTable.ts
CharlieKolb: Ugh the other migration just rounded up to a future timestamp, that's so annoying :'D

typecheck.log
CharlieKolb: Oh this was an accident during the merge conflict - had some local files and I always do `git add . && git commit` for merge conflicts only. Nice catch 🙏

packages/cli/src/modules/favorites/favorites.controller.ts
CharlieKolb: Good point!

packages/cli/src/modules/favorites/favorites.controller.ts
CharlieKolb: This is mostly to get the actual type narrower - not sure if we should use params or a body with delete?

RicardoE105 (approved):
🚀

CharlieKolb:
Ah sorry, to address your other comments:

> Is it intentional that credentials cannot be marked as favorites?

Yes, I felt that these shouldn't need frequent enough access to support, maybe I'm missing a use case?

> If a workflow is marked as a favorite, what happens when its original project is deleted and the workflow is moved to another project? Should it remain marked as a favorite in the table?

If it keeps its ID it will keep the favorite status, otherwise it'll be gone

PR #26229, opened 2026-02-25
fix: Add missing chat tool input logs

michael-radency:
## Summary

fixes issue when input logs was not displayed for chat tool
image
image

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/chat/chat-execution-manager.ts
```diff
@@ -139,6 +145,31 @@ export class ChatExecutionManager {
...
+			runExecutionData.executionData!.nodeExecutionStack[0].node.rewireOutputLogTo =
+				NodeConnectionTypes.AiTool;
+
+			const lastNodeExecuted = runExecutionData.resultData.lastNodeExecuted as string;
+			const runDataArray = runExecutionData.resultData.runData[lastNodeExecuted];
+			if (runDataArray?.length > 0) {
+				const entryToPop = runDataArray[runDataArray.length - 1];
+				// Preserve inputOverride so the LLM's input stays visible in logs after resume.
+				const preservedInputOverride = entryToPop?.inputOverride;
+				runDataArray.pop();
```
dlavrenuek: You can simple use pop to retrieve the element and remove it from the array.

```suggestion
				const entryToPop = runDataArray.pop();
				// Preserve inputOverride so the LLM's input stays visible in logs after resume.
				const preservedInputOverride = entryToPop?.inputOverride;
```

Can `runDataArray` actually have `undefined` as an entry? If not then .pop() can not return undefined here doe to the outer check `if (runDataArray?.length > 0) {`, which is not recognized by TS apparently, so instead of optional checks like `entryToPop?.inputOverride;` you could also do `const entryToPop = runDataArray.pop()!;` (with !) which will ensure that this entry is not undefined also for TS (it will throw an error if it's undefined afaik).

dlavrenuek (changes requested):

packages/cli/src/chat/chat-execution-manager.ts
michael-radency: thanks @dlavrenuek , updated as suggested

dlavrenuek (approved):
Works as expected 🚀

PR #26245, opened 2026-02-25
fix(core): Skip onboarding screen for SAML users with IdP-provided names

guillaumejacquart:
## Summary

Fixes IAM-311: Users logging in via SAML SSO with just-in-time provisioning should no longer see the "Finish account setup" screen if their name is already provided by the SAML identity provider.

Previously, the onboarding screen was always shown for new JIT-provisioned users. Now it applies the same conditional check used for existing users—only showing onboarding when firstName or lastName is missing from the IdP.

https://linear.app/n8n/issue/IAM-311/saml-sso-direct-login-without-setup-screen

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/sso-saml/saml.service.ee.ts
+++ b/packages/cli/src/modules/sso-saml/saml.service.ee.ts
@@ -254,7 +254,7 @@ export class SamlService {
 					return {
 						authenticatedUser: newUser,
 						attributes,
-						onboardingRequired: true,
+						onboardingRequired: !newUser.firstName || !newUser.lastName,
 					};
 				}
 			}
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/sso-saml/saml.service.ee.ts
```diff
@@ -254,7 +254,7 @@ export class SamlService {
 					return {
 						authenticatedUser: newUser,
 						attributes,
-						onboardingRequired: true,
+						onboardingRequired: !newUser.firstName || !newUser.lastName,
```
afitzek, line 257: Do we need the same also for OIDC?

afitzek (approved):

packages/cli/src/modules/sso-saml/saml.service.ee.ts
guillaumejacquart: Good question. I asked product about it, we need to align:
- do we still want this onboarding step to be displayed if user is missing first name / last name in SSO info. If yes we need to implement it on the OIDC service as well
- if no, we can remove this from SAML

PR #26248, opened 2026-02-25
feat: Update existing MCP tools

dariacodes:
## Summary

- Remove MCP requirement for published workflows
- Extend the existing tools to return both current and active versions
- Add manual mode to executions
- Update search workflows tool
    - Remove nodes to have smaller output
    - Show all workflows and return `availableInMCP` as a param
- Update `role.service.ts` method detecting whether an entity is a workflow or credential
    - Before we were checking a deprecated field `active` - now we check for the existence of any of these: `activeVersionId`, `versionId`, `triggerCount`

Closes [ADO-4865](https://linear.app/n8n/issue/ADO-4865/feature-introduce-manual-executions-for-mcp)
Closes [ADO-4878](https://linear.app/n8n/issue/ADO-4878/feature-update-mcp-search-workflows-tool)

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/mcp/tools/execute-workflow.tool.ts
```diff
@@ -38,6 +40,11 @@ const ERROR_KEYS_TO_IGNORE = ['stack', 'node'];
 
 const inputSchema = z.object({
 	workflowId: z.string().describe('The ID of the workflow to execute'),
+	mode: z
+		.enum(['manual', 'production'])
+		.optional()
+		.default('production')
+		.describe('Execution mode: manual runs current version; production runs published (active)'),
```
MiloradFilipovic: I think it would help to add more details about when to use which mode, so agents can have more context when deciding.
Something like: `Execution mode: use manual mode to test the current version of the workflow. Use production mode to execute published version.`
Also, if we rename the property to `executionMode`, we don't need to add `Execution mode:` to the description.

packages/cli/src/modules/mcp/tools/search-workflows.tool.ts
```diff
@@ -51,6 +56,10 @@ const outputSchema = {
 					.nullable()
 					.describe('The number of triggers associated with the workflow'),
 				nodes: z.array(nodeSchema).describe('List of nodes in the workflow'),
+				activeNodes: z
```
MiloradFilipovic: Why are we exposing nodes trough this tool?
Since we were talking about removing the mcp access filter from this tool (so all workflows are searchable), exposing sensitive data here will defeat the purpose of the mcp access toggle.
This will also greatly increase the results payload size.

packages/cli/src/modules/mcp/tools/webhook-utils.ts
```diff
@@ -54,7 +54,7 @@ export const getTriggerDetails = async (
 	endpoints: WebhookEndpoints,
 ): Promise<string> => {
 	if (supportedTriggers.length === 0) {
-		return 'This workflow does not have a trigger node that can be executed via MCP.';
+		return 'No MCP input trigger node detected. This workflow can still be executed in manual mode.';
```
MiloradFilipovic: What is `MCP input trigger node`? If this means supported MCP triggers, I suggest we rephrase this.

packages/frontend/editor-ui/src/app/components/WorkflowProductionChecklist.vue
```diff
@@ -96,16 +94,13 @@ const isMcpAccessEnabled = computed(() => {
 	return settingsStore.moduleSettings.mcp?.mcpAccessEnabled ?? false;
 });
 
-const isWorkflowEligibleForMcpAccess = computed(() => {
```
MiloradFilipovic, line 99: Maybe I am missing something, but I wasn't expecting changes to this component since it pops up after workflows are published.

packages/cli/src/modules/mcp/tools/execute-workflow.tool.ts
```diff
@@ -447,6 +520,14 @@ const getPinDataForTrigger = async (
 	}
 };
 
+const getSupportedTriggerNamesForMode = (mode: z.infer<typeof inputSchema>['mode']): string[] => {
```
MiloradFilipovic: Nit: I feel like this method is a bit awkward to read, maybe something like this would be cleaner:
```ts
const names = Object.values(SUPPORTED_MCP_TRIGGERS);
if (mode === 'manual') {
  names.unshift('Manual Trigger');
}
return names;
```

MiloradFilipovic (commented):
Works well overall 🙌
I left a few comments so take a look at what makes sense.

packages/cli/src/modules/mcp/tools/search-workflows.tool.ts
dariacodes: What do you mean by sensitive data? We're already exposing nodes here (which are nodes of an active version), and now that we want to expose both in MCP, it looks like here we should either expose both current version and active version nodes or neither, no?

packages/cli/src/modules/mcp/tools/webhook-utils.ts
dariacodes: This is a bit tricky, because it means there are no production triggers supported by MCP, and we can't just say "there are no triggers supported by MCP" because we also support a manual trigger now. Shall we use "production" then?

packages/frontend/editor-ui/src/app/components/WorkflowProductionChecklist.vue
dariacodes: Yeah, I got confused with the manual execution, thinking that we can do it with any workflow. Will revert this

packages/cli/src/modules/mcp/tools/execute-workflow.tool.ts
dariacodes: I created a separate type, so this method is now much cleaner

packages/cli/src/modules/mcp/tools/search-workflows.tool.ts
dariacodes: upd.: We decided to keep minimum info here i.e. remove everything I added here + remove nodes

packages/cli/src/services/role.service.ts
```diff
@@ -245,11 +245,11 @@ export class RoleService {
 			return entity;
 		}
 
-		if (!('active' in entity) && !('type' in entity)) {
+		if (!('versionId' in entity) && !('type' in entity)) {
 			throw new UnexpectedError('Cannot detect if entity is a workflow or credential.');
 		}
 
-		const entityType = 'active' in entity ? 'workflow' : 'credential';
+		const entityType = 'versionId' in entity ? 'workflow' : 'credential';
```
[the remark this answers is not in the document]
dariacodes: The field `active` is deprecated, and all these `ListQueryDb.Workflow` fields are optional, so this issue already exists in the repo. But in practice, `applySelect()` always includes `versionId` in the default select if no custom field is selected. I'll add other fallbacks in a new commit, and I think it'll be a better solution

packages/cli/src/modules/mcp/tools/search-workflows.tool.ts
```diff
@@ -50,11 +48,11 @@ const outputSchema = {
 					.number()
 					.nullable()
 					.describe('The number of triggers associated with the workflow'),
-				nodes: z.array(nodeSchema).describe('List of nodes in the workflow'),
 				scopes: z.array(z.string()).describe('User permissions for this workflow'),
 				canExecute: z
 					.boolean()
 					.describe('Whether the user has permission to execute this workflow'),
+				availableInMCP: z.boolean().describe('Whether the workflow is available for MCP execution'),
```
MiloradFilipovic: Nit: Maybe we can be more precise in this description. Since `availableInMCP` makes workflow available for other tools also maybe we can say something like: `Whether the workflow is visible to MCP tools`

MiloradFilipovic (dismissed):
Looking good, thanks for addressing the feedback!
Left a minor comment about `availableInMCP` description but good to go even without it.

packages/cli/src/modules/mcp/tools/webhook-utils.ts
MiloradFilipovic: Yeah, new version looks good to me

MiloradFilipovic (approved):
Thanks 🙌

PR #26249, opened 2026-02-25
fix: Improve chat message button handling

dlavrenuek:
## Summary

Improves chat action button rendering

https://linear.app/n8n/issue/NODE-4374

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/@n8n/chat/src/components/MessageWithButtons.vue
```diff
@@ -15,6 +15,15 @@ defineProps<{
 
 const clickedButtonIndex = ref<number | null>(null);
 
+const isSameDomain = (link: string): boolean => {
+	try {
+		const url = new URL(link, window.location.href);
+		return url.origin === window.location.origin;
```
yehorkardash: Can those urls include n8n webhook urls? Afaik buttons usually can lead to `form-waiting` or `form` endpoint.
Those endpoints can be modified by some env variable(don't remember the name) and it will be different from frontend url. Users may use a proxy for webhooks, which can lead to button not rendering even if it's technically their domain

dlavrenuek: They can. I just figured that this can be different from the web ui url apparently and am looking how we can get this value in the chat components

dlavrenuek: fixed

dlavrenuek: did run some tests locally to confirm that the links are always generated with the configured webhook url, so there is no need to allow the current origin

packages/frontend/@n8n/chat/src/components/MessageWithButtons.vue
```diff
@@ -13,8 +15,19 @@ defineProps<{
 	}>;
 }>();
 
+const chatOptions = useOptions();
 const clickedButtonIndex = ref<number | null>(null);
 
+const isValidOrigin = (link: string): boolean => {
```
yehorkardash: Would maybe make it computed function

packages/frontend/@n8n/chat/src/components/MessageWithButtons.vue
```diff
@@ -33,7 +46,10 @@ const onClick = async (link: string, index: number) => {
 		<div :class="$style.buttons">
 			<template v-for="(button, index) in buttons" :key="button.text">
 				<Button
-					v-if="clickedButtonIndex === null || index === clickedButtonIndex"
+					v-if="
```
yehorkardash: Would be nice to have `clickedButtonIndex === null || index === clickedButtonIndex` as a variable with a clear name, because it's not clear what is it for

yehorkardash (dismissed):
LGTM, just two nitpicks

packages/frontend/@n8n/chat/src/components/MessageWithButtons.vue
dlavrenuek: This is indeed not optimal. I did extract the check into the new function and renamed it to `isButtonVisible`. The `clickedButtonIndex.value === null || index === clickedButtonIndex.value` is still there since a proper description would be "if no buttons clicked or current button is clicked" and converting it to multiple conditions would not make it better, imo

packages/frontend/@n8n/chat/src/components/MessageWithButtons.vue
dlavrenuek: I did check what the chatOptions are and the result is not reactive, so a computed value would also never be updated

yehorkardash (approved):

PR #26255, opened 2026-02-25
chore(core): Integrate execution redaction service in execution service

afitzek:
## Summary

Introduce the `ExecutionRedaction` interface and `ExecutionRedactionServiceProxy` and wire them into the execution service. This is a structural extraction from a larger redaction feature PR — no functional redaction logic is included.

### Changes

- **New `ExecutionRedaction` interface** and `ExecutionRedactionOptions` type — defines the contract for execution data processing
- **New `ExecutionRedactionServiceProxy`** — proxy service that delegates to a real implementation when one is registered, otherwise returns executions unmodified
- **New `ExecutionRedactionQueryDto`** — Zod schema for the `redactExecutionData` query parameter, using the established `booleanFromString` helper
- **New `findIfSharedUnflatten()`** on `ExecutionRepository` — fetches unflattened execution data for shared workflows
- **Modified `ExecutionService.findOne()`** — now fetches unflattened data, passes through the redaction proxy, and re-stringifies the result
- **Adapted `ExecutionRedactionService` stub** — implements the new `ExecutionRedaction` interface (still a no-op)
- **Updated `RedactionModule`** — wires the proxy on init, renamed env var to `N8N_ENV_FEAT_EXECUTION_REDACTION`
- **Registered `'redaction'` module** in `MODULE_NAMES` and `defaultModules`

### How to test

1. Build and run `pnpm typecheck` — all types resolve
2. Run affected tests — all pass:
   - `packages/cli/src/executions/__tests__/execution-redaction-proxy.service.test.ts`
   - `packages/cli/src/executions/__tests__/execution.service.test.ts`
   - `packages/cli/src/modules/redaction/__tests__/redaction.module.test.ts`
   - `packages/cli/src/modules/redaction/executions/__tests__/execution-redaction.service.test.ts`
3. Verify no behavioral change — the proxy is a passthrough when no redaction implementation is registered

https://linear.app/n8n/issue/IAM-177/integrate-service-into-execution-controllers

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/executions/execution.service.ts
```diff
@@ -133,7 +140,24 @@ export class ExecutionService {
...
+		const redactQuery = ExecutionRedactionQueryDtoSchema.safeParse(req.query);
+		if (redactQuery.success) {
+			redactExecutionData = redactQuery.data.redactExecutionData;
+		}
+
+		const processedExecution = await this.executionRedactionServiceProxy.processExecution(
+			execution,
+			{
+				user: req.user,
+				redactExecutionData,
```
guillaumejacquart, line 153: Why do we pass this boolean here ? Couldn't we just say that if it's false, we return the execution without processing it ?

packages/cli/src/executions/execution.service.ts
```diff
@@ -133,7 +140,24 @@ export class ExecutionService {
...
+			},
+		);
+
+		return {
+			...execution,
+			data: stringify(processedExecution.data),
```
guillaumejacquart, line 159: I thought we would try and do the redaction inside the flattening / stringifying process to prevent full parsing / redacting / stringifying. But maybe that was before we decided to redact the full output and not field values

guillaumejacquart (commented):
2 small comments

packages/cli/src/executions/execution.service.ts
afitzek: No, because we need to check the permissions and policies, this happens in the redaction service.

packages/cli/src/executions/execution.service.ts
afitzek: I don't think that it is feasible to redact on the flattened string tbh.

packages/cli/src/executions/execution.service.ts
guillaumejacquart: ah, do you mean especially if the value of redactExecutionData is false, we need to make sure that the user has the right to reveal ? 
I'm wondering if this permission check should live on the redaction service ?

afitzek: Yes, so basically here (https://github.com/n8n-io/n8n/pull/26268/changes#diff-3f26a12cd2ddeb7c0238866b48543518f2eac7d885d7ed5e57d36df6a6fcb6b1R51) we are handling the logic for the redaction. I think the redaction module is the right place for this, since I think this is the core use case of the module. :) Otherwise the logic is scattered around different services. So here we are basically just always running it through to the redaction service to give it the change to handle it.

packages/cli/src/executions/execution.service.ts
guillaumejacquart: Yes, I thought we already parsed the execution data when fetching one execution, I did not realize we parse from the frontend then ?
Otherwise, we might have been able to pass a reviver (specified by the redaction service or something) to the parse method: https://github.com/WebReflection/flatted/blob/main/index.js#L72

afitzek: The problem is that we need data from the flatten info to get the redaction policy to make a decision if we want to redact or not. So to safe memory we would have to find this information from the execution context in the flattened string to make a decision, but if we have to redact, we would not even save that much memory, because the redaction runs in place and mutates the data.

Also there is a migration running for the runtime data: https://github.com/n8n-io/n8n/pull/26255/changes#diff-13aab488c338e81849a1783dcc6021b51e17a0443e21dc8e5ed5ad3ff60fa8e6R1147 which we would have to take into account before operating on the flattened string to potentially read the correct redaction policy for the execution.

guillaumejacquart (approved):

PR #26262, opened 2026-02-25
fix(Monday.com Node): Migrate to API version 2026-07

### LINKED ISSUE #26071: Monday nodes using deprecated API-Version
### Bug Description

Since the 15th of Feb I have been having issues with subitems and board relationships not being available in the returned data from Monday when using any function that returns items;
 - Get
 - Get Many
 - Get by Column Value

Previously related items were returned via the value:
E.g from successful execution prior to 15th of Feb.
`{
  "id": "subtasks_mkt4epz0",
  "text": null,
  "type": "subtasks",
  "value": "{\"linkedPulseIds\":[{\"linkedPulseId\":2708163991}]}",
  "column": {
    "title": "Subitems",
    "archived": false,
    "description": null,
    "settings_str": "{\"allowMultipleItems\":true,\"itemTypeName\":\"column.subtasks.title\",\"displayType\":\"BOARD_INLINE\",\"boardIds\":[2065244174]}"
  }
}`

Same out today
`{
  "id": "subtasks_mkt4epz0",
  "text": null,
  "type": "subtasks",
  "value": null,
  "column": {
    "title": "Subitems",
    "archived": false,
    "description": null,
    "settings_str": "{\"allowMultipleItems\":true,\"itemTypeName\":\"column.subtasks.title\",\"displayType\":\"BOARD_INLINE\",\"boardIds\":[2065244174]}"
  }
}`

Monday.com effectively said "Don't use the Monday nodes, use HTTP Requests" which is a significant amount of work for me in 100+ workflows and implementing pagination etc etc while my business is at a stand still is chaos.

My Research indicates that the API-Version being used by n8n is deprecated ( https://developer.monday.com/api-reference/docs/api-versioning ) so I'm assuming that's the issue.

I have checks using Postman and upgrading the API-Version to 2026-01 and using the below query
`{
  "query": "query { items (ids: [2076829050]) { id name created_at state column_values { id text type value ... on BoardRelationValue { display_value } ... on SubtasksValue { display_value subitems { id name } } column { title archived description settings_str } } } }"
}`

I get an array of subtasks;
`                "column_values": [
                    {
                        "id": "subtasks_mkt4epz0",
                        "text": null,
                        "type": "subtasks",
                        "value": null,
                        "display_value": "Elham Bidarigh - Shift",
                        "subitems": [
                            {
                                "id": "2708163991",
                                "name": "Elham Bidarigh - Shift"
                            }
                        ],
                        "column": {
                            "title": "Subitems",
                            "archived": false,
                            "description": null,
                            "settings_str": "{\"allowMultipleItems\":true,\"itemTypeName\":\"column.subtasks.title\",\"displayType\":\"BOARD_INLINE\",\"boardIds\":[2065244174]}"
                        }
                    },`

### To Reproduce

1. Have a Monday board with subtasks and connected board columns
2. Use the Monday.com > Get Item node and add the pulseId for the item with subtasks/connected boards
3. Check the output

### Expected behavior

Returned data still contains the `"value": "{\"linkedPulseIds\":[{\"linkedPulseId\":2708163991}]}",` or at least a list of sub tasks in an array.

### Debug Info

# My primary n8n

# Debug info

## core

- n8nVersion: 1.121.3
- platform: npm
- nodeJsVersion: 22.17.0
- nodeEnv: undefined
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: e13efa3d-11d5-4578-b416-fa3d3890b1e3

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/144.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-02-21T00:31:29.201Z

# My test local n8n

# Debug info

## core

- n8nVersion: 2.8.3
- nodeJsVersion: 22.22.0
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- consumerId: d257fffc-5d9f-45f4-be74-264ee2d0b61a

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/144.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-02-21T00:32:39.076Z

### Operating System

Ububtu

### n8n Version

1.121.3

### Node.js Version

22.17.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

c2nprds:
## Summary

Migrate the Monday.com node from the deprecated API version `2023-10` to the current stable version `2026-07`.

Monday.com no longer supports version `2023-10`. When a node sends a removed version, Monday.com serves the request with the oldest supported version (maintenance version) instead. That silent bump already exposed the node to server-side breaking changes it never adopted — most notably the `value` field returning `null` for board relation, dependency, subtasks, and mirror columns (changed in `2025-04`). This PR pins the node to a supported version and updates the GraphQL queries to read the documented replacement fields, so those columns return data again.

This PR does **not** introduce breaking changes. The output changes below reflect Monday.com API changes that already shipped; workflows relying on the old fields were already broken because `2023-10` is deprecated.

### Changes

- Update the `API-Version` header from `2023-10` to `2026-07` in `GenericFunctions.ts` and `MondayComApi.credentials.ts`
- Add inline fragments to `column_values` for the columns whose `value` now returns `null`, for `boardItem:get`, `getAll`, and `getByColumnValue`:
  - `BoardRelationValue`: `display_value`, `linked_item_ids`
  - `DependencyValue`: `display_value`, `linked_item_ids`
  - `MirrorValue`: `display_value`
  - `SubtasksValue`: `display_value`, `subitems { id name }`
- Extract the shared `column_values` selection into a single `columnValuesFragment` constant instead of duplicating it across the three queries
- Add unit tests for all modified operations, API version header verification, and cursor-based pagination

### Behavior changes (already shipped by Monday.com, not introduced here)

These affect `boardItem:get`, `boardItem:getAll`, and `boardItem:getByColumnValue`:

| Field | Details |
|-------|---------|
| `column_values[].value` returns `null` | For board relation, dependency, subtasks, and mirror columns, per the Monday.com `2025-04` change. The data now comes from the typed fragments below. |
| `display_value` on column values | Text representation of the column, available on `BoardRelationValue`, `DependencyValue`, `MirrorValue`, and `SubtasksValue`. |
| `linked_item_ids` on column values | IDs of linked items, available on `BoardRelationValue` and `DependencyValue`. |
| `subitems` on `SubtasksValue` | Subitems (`{ id name }`) are read from the subtasks column value, keeping them linked to the column that holds them. |

### References

- [Monday.com API versioning](https://developer.monday.com/api-reference/docs/api-versioning) (current stable version: `2026-07`)
- [`value` field now returns `null` on connect boards, dependency, and subtasks columns](https://developer.monday.com/api-reference/changelog/value-field-now-returns-null-on-connect-boards-dependency-and-subtasks-columns)
- [BoardRelationValue fields](https://developer.monday.com/api-reference/reference/connect)
- [DependencyValue fields](https://developer.monday.com/api-reference/reference/dependency)
- [MirrorValue fields](https://developer.monday.com/api-reference/reference/mirror)
- [Subitems column reference](https://developer.monday.com/api-reference/reference/subitems-column)

Fixes #26071

## Review / Merge checklist

  Follow-up needed: response structure changed.
  - `value` returns `null` for board relation, dependency, subtasks, and mirror columns
  - New `display_value` / `linked_item_ids` fields on typed column values, and `subitems` on `SubtasksValue`
- [x] Tests included.

### DIFF
--- a/packages/nodes-base/credentials/MondayComApi.credentials.ts
+++ b/packages/nodes-base/credentials/MondayComApi.credentials.ts
@@ -34,7 +34,7 @@ export class MondayComApi implements ICredentialType {
 	test: ICredentialTestRequest = {
 		request: {
 			headers: {
-				'API-Version': '2023-10',
+				'API-Version': '2026-01',
 				'Content-Type': 'application/json',
 			},
 			baseURL: 'https://api.monday.com/v2',
--- a/packages/nodes-base/nodes/MondayCom/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/MondayCom/GenericFunctions.ts
@@ -19,7 +19,7 @@ export async function mondayComApiRequest(
 
 	let options: IRequestOptions = {
 		headers: {
-			'API-Version': '2023-10',
+			'API-Version': '2026-01',
 			'Content-Type': 'application/json',
 		},
 		method: 'POST',
--- a/packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
+++ b/packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
@@ -596,13 +596,27 @@ export class MondayCom implements INodeType {
 										name
 										created_at
 										state
+										subitems {
+											id
+											name
+										}
 										column_values {
 											id
 											text
 											type
 											value
+											... on BoardRelationValue {
+												display_value
+												linked_item_ids
+											}
+											... on DependencyValue {
+												display_value
+												linked_item_ids
+											}
+											... on MirrorValue {
+												display_value
+											}
 											column {
-
 												title
 												archived
 												description
@@ -629,11 +643,26 @@ export class MondayCom implements INodeType {
 							name
 							created_at
 							state
+							subitems {
+								id
+								name
+							}
 							column_values {
 								id
 								text
 								type
 								value
+								... on BoardRelationValue {
+									display_value
+									linked_item_ids
+								}
+								... on DependencyValue {
+									display_value
+									linked_item_ids
+								}
+								... on MirrorValue {
+									display_value
+								}
 								column {
 									title
 									archived
@@ -690,11 +719,26 @@ export class MondayCom implements INodeType {
 							board {
 								id
 							}
+							subitems {
+								id
+								name
+							}
 							column_values {
 								id
 								text
 								type
 								value
+								... on BoardRelationValue {
+									display_value
+									linked_item_ids
+								}
+								... on DependencyValue {
+									display_value
+									linked_item_ids
+								}
+								... on MirrorValue {
+									display_value
+								}
 								column {
 									title
 									archived
--- a/packages/nodes-base/nodes/MondayCom/test/MondayCom.node.test.ts
+++ b/packages/nodes-base/nodes/MondayCom/test/MondayCom.node.test.ts
@@ -1,0 +1,253 @@
+import { NodeTestHarness } from '@nodes-testing/node-test-harness';
+import nock from 'nock';
+
+describe('MondayCom', () => {
+	const mondayApiUrl = 'https://api.monday.com';
+	const apiVersion = '2026-01';
+
+	/**
+	 * Shared column_values mock data aligned with API version 2026-01.
+	 * - BoardRelationValue: value is null, use display_value and linked_item_ids
+	 * - DependencyValue: value is null, use display_value and linked_item_ids
+	 * - MirrorValue: value is null, use display_value
+	 * Ref: https://developer.monday.com/api-reference/reference/connect
+	 * Ref: https://developer.monday.com/api-reference/reference/dependency
+	 * Ref: https://developer.monday.com/api-reference/reference/mirror
+	 */
+	const columnValuesWithFragments = [
+		{
+			id: 'status',
+			text: 'Working on it',
+			type: 'status',
+			value: '{"index":1}',
+			column: {
+				title: 'Status',
+				archived: false,
+				description: null,
+				settings_str: '{}',
+			},
+		},
+		{
+			id: 'connect_boards',
+			text: null,
+			type: 'board_relation',
+			value: null,
+			display_value: 'Related Item 1',
+			linked_item_ids: ['5566778899'],
+			column: {
+				title: 'Connected Board',
+				archived: false,
+				description: null,
+				settings_str: '{}',
+			},
+		},
+		{
+			id: 'subtasks_col',
+			text: null,
+			type: 'subtasks',
+			value: null,
+			column: {
+				title: 'Subitems',
+				archived: false,
+				description: null,
+				settings_str: '{}',
+			},
+		},
+		{
+			id: 'dependency_col',
+			text: null,
+			type: 'dependency',
+			value: null,
+			display_value: 'Dep Item 1',
+			linked_item_ids: ['4433221100'],
+			column: {
+				title: 'Dependencies',
+				archived: false,
+				description: null,
+				settings_str: '{}',
+			},
+		},
+		{
+			id: 'mirror_col',
+			text: null,
+			type: 'mirror',
+			value: null,
+			display_value: 'Mirrored Value',
+			column: {
+				title: 'Mirror',
+				archived: false,
+				description: null,
+				settings_str: '{}',
+			},
+		},
+	];
+
+	const itemData = {
+		id: '1234567890',
+		name: 'Test Item',
+		created_at: '2026-02-20T10:00:00Z',
+		state: 'active',
+		subitems: [{ id: '111222333', name: 'Subtask A' }],
+		column_values: columnValuesWithFragments,
+	};
+
+	/** Checks that the GraphQL query uses the documented field structure */
+	function hasExpectedQueryStructure(body: Record<string, unknown>): boolean {
+		const query = body.query as string;
+		return (
+			query.includes('subitems') &&
+			query.includes('... on BoardRelationValue') &&
+			query.includes('... on DependencyValue') &&
+			query.includes('... on MirrorValue') &&
+			query.includes('display_value') &&
+			query.includes('linked_item_ids')
+		);
+	}
+
+	describe('boardItem:get - returns subitems and board relation data', () => {
+		beforeAll(() => {
+			nock(mondayApiUrl)
+				.post('/v2/', (body: Record<string, unknown>) => {
+					const query = body.query as string;
+					return query.includes('items (ids: $itemId)') && hasExpectedQueryStructure(body);
+				})
+				.reply(200, { data: { items: [itemData] } });
+		});
+
+		new NodeTestHarness().setupTests({
+			workflowFiles: ['get.workflow.json'],
+		});
+	});
+
+	describe('boardItem:getAll - returns subitems and board relation data', () => {
+		beforeAll(() => {
+			nock(mondayApiUrl)
+				.post('/v2/', (body: Record<string, unknown>) => {
+					const query = body.query as string;
+					return query.includes('items_page') && hasExpectedQueryStructure(body);
+				})
+				.reply(200, {
+					data: {
+						boards: [
+							{
+								groups: [
+									{
+										id: 'group1',
+										items_page: {
+											cursor: null,
+											items: [itemData],
+										},
+									},
+								],
+							},
+						],
+					},
+				});
+		});
+
+		new NodeTestHarness().setupTests({
+			workflowFiles: ['getAll.workflow.json'],
+		});
+	});
+
+	describe('boardItem:getByColumnValue - returns subitems and board relation data', () => {
+		beforeAll(() => {
+			nock(mondayApiUrl)
+				.post('/v2/', (body: Record<string, unknown>) => {
+					const query = body.query as string;
+					return query.includes('items_page_by_column_values') && hasExpectedQueryStructure(body);
+				})
+				.reply(200, {
+					data: {
+						items_page_by_column_values: {
+							cursor: null,
+							items: [
+								{
+									...itemData,
+									board: { id: '9876543210' },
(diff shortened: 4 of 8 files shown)


c2nprds:
@rootInfluence Thanks for the feedback. I've updated the PR description to include a "Breaking changes for existing workflows" section that documents the output changes users may encounter.

packages/nodes-base/credentials/MondayComApi.credentials.ts
```diff
@@ -34,7 +34,7 @@ export class MondayComApi implements ICredentialType {
 	test: ICredentialTestRequest = {
 		request: {
 			headers: {
-				'API-Version': '2023-10',
+				'API-Version': '2026-01',
```
dlavrenuek: This API version is now also deprecated, let's use `2026-07`
```suggestion
				'API-Version': '2026-07',
```

packages/nodes-base/nodes/MondayCom/GenericFunctions.ts
```diff
@@ -19,7 +19,7 @@ export async function mondayComApiRequest(
 
 	let options: IRequestOptions = {
 		headers: {
-			'API-Version': '2023-10',
+			'API-Version': '2026-01',
```
dlavrenuek: Also here

```suggestion
			'API-Version': '2026-07',
```

packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
```diff
@@ -596,13 +596,27 @@ export class MondayCom implements INodeType {
 										name
 										created_at
 										state
+										subitems {
+											id
+											name
+										}
```
dlavrenuek: Version 2025-04 made value return null on connect boards, dependency and subtasks columns. The PR adds a fragment for the first two but not for subtasks. A subtasks column therefore matches none of the three fragments and falls back to the interface fields only. The test fixture in this PR shows the result: subtasks_col has value: null, text: null and no replacement field.

The documented replacement for a subtasks column is subitems_ids or subitems, queried on the column value. Item-level subitems returns the same items but loses the link to the column that holds them.

If I see it correctly this should be removed and instead we can add following in the `column_values`


```
... on SubtasksValue {
    display_value
    subitems {
        id
        name
    }
}
```

https://developer.monday.com/api-reference/changelog/value-field-now-returns-null-on-connect-boards-dependency-and-subtasks-columns

packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
```diff
@@ -596,13 +596,27 @@ export class MondayCom implements INodeType {
...
+												display_value
+												linked_item_ids
+											}
+											... on DependencyValue {
+												display_value
+												linked_item_ids
+											}
+											... on MirrorValue {
+												display_value
+											}
```
dlavrenuek: The same block appears three times, that's 11 (18 if we add `... on SubtasksValue {...`) lines x3 code duplication, let's extract it to a constant to be reused

dlavrenuek (changes requested):
Hi @c2nprds, thank you for the contribution and sorry for the long waiting time.

In general the changes are solid, we need to make some adjustments to the query and also the description.
The description states breaking changes, while there are none. I looked through Monday.com api docs and found that they move the api version to the latest supported one, when using a removed version, this means that there the api already had breaking changes that the node did not pick up and what happens in this PR is fixing some of them.

According to https://developer.monday.com/api-reference/docs/api-versioning the currently supported version is `2026-07`, I did look through the changes to `2025-04` (one that would be set automatically afaik) and there is nothing that we must consider.

One change that could be added here, but could also be considered out of scope: the error reporting did change in [`2025-01`](https://developer.monday.com/api-reference/changelog/breaking-change-consistent-error-format), application level errors such as failed mutations now return status 200 with `errors` property in the response.  This means that we should add additional check and throw an error - but it fixes another problem, so totally optional here.

packages/nodes-base/credentials/MondayComApi.credentials.ts
c2nprds: Done — using 2026-07.

packages/nodes-base/nodes/MondayCom/GenericFunctions.ts
c2nprds: Done — using 2026-07.

packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
c2nprds: Done. Removed the item-level `subitems` and query the subtasks column through `... on SubtasksValue { display_value subitems { id name } }` instead, per the docs. Fixtures updated to match.

packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
c2nprds: Done — extracted into a shared `columnValuesFragment` constant reused by the get, getAll and getByColumnValue queries.

c2nprds:
@dlavrenuek Thanks for the review. Pushed the changes:

- Bumped the API version to `2026-07` (current supported version).
- Reworked subtasks handling: removed the item-level `subitems` field and added `... on SubtasksValue { display_value subitems { id name } }` to `column_values`, so the subitems stay linked to the column that holds them.
- Extracted the shared `column_values` selection into a single constant reused by the three queries.
- Included the optional `2025-01` error handling: `mondayComApiRequest` now throws when the API returns HTTP 200 with an `errors` array.
- Updated the description — agreed there are no breaking changes introduced here; the `value` = null behaviour already shipped in `2025-04` and the node had not adopted the replacement fields. Reworded accordingly.

Tests and workflow fixtures updated to match.

packages/nodes-base/nodes/MondayCom/MondayCom.node.ts
```diff
@@ -25,6 +25,45 @@ interface IGraphqlBody {
...
+		linked_item_ids
+	}
+	... on DependencyValue {
+		display_value
+		linked_item_ids
+	}
+	... on MirrorValue {
+		display_value
+	}
+	... on SubtasksValue {
```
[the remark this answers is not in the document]
c2nprds: Intentional, not a regression. The item-level `subitems` field was only added earlier in this same PR (never released), so there are no existing consumers relying on it. This follows the maintainer's review: subitems are now read from the subtasks column via `... on SubtasksValue { display_value subitems { id name } }`, which keeps them linked to the column that holds them (item-level subitems loses that link). The data is still returned, under `column_values`.

packages/nodes-base/nodes/MondayCom/GenericFunctions.ts
```diff
@@ -30,16 +30,38 @@ export async function mondayComApiRequest(
...
 	} catch (error) {
 		throw new NodeApiError(this.getNode(), error as JsonObject);
 	}
+
+	// Since API version 2025-01 application-level errors (e.g. a failed mutation)
+	// are returned with HTTP 200 and an `errors` array in the body, so surface
+	// them explicitly instead of returning a response with missing data.
+	// https://developer.monday.com/api-reference/changelog/breaking-change-consistent-error-format
+	const errors = response?.errors as Array<{ message?: string }> | undefined;
+	if (errors?.length) {
```
dlavrenuek: I did some testing with very old boards, where some of them were apparently broken and returned partial data, which according to the docs is the behavior now:

> Partial data: Your query may return partial data and throw errors. In that case, you will receive both the data and errors objects. If a field throws an error, it will return null.

This means that to not add breaking changes we must also check if data is present and only if it's absent throw an error.

```suggestion
	if (errors?.length && !response?.data) {
```

dlavrenuek (changes requested):
Thank you for addressing my suggestions @c2nprds and also so quickly 🥇 Works very well overall, I only found one issue with new error handling and broken data/server errors on Monday side - small fix and we are good to merge!

packages/nodes-base/nodes/MondayCom/GenericFunctions.ts
dlavrenuek: This would mean also one additional test for the error path (one with data, one without)

c2nprds: Done. Now throws only when the response has errors and no data, so partial responses (data + errors) pass through unchanged.

c2nprds: Added a test for the partial-data path (data present, not thrown) alongside the existing no-data path.

dlavrenuek (approved):
Looks great 🎉

c2nprds:
@dlavrenuek All the review feedback is addressed and pushed (858ee14):

- API version → `2026-07`; subtasks read via `... on SubtasksValue { display_value subitems { id name } }`; shared `column_values` selection extracted into one constant.
- Application-level errors: `mondayComApiRequest` now throws only when the response has an `errors` array and no `data`, so partial responses (data + errors) pass through. Tests cover both the no-data and partial-data paths.

Ready for another review when you have time.

Heads-up on CI: the `Required Reviews` check failed at its `Set pending status` step with `Resource not accessible by integration (HTTP 403)` — the run's token can't write a commit status for a PR from a fork. It looks unrelated to this change; flagging in case it needs attention on the CI side.

PR #26268, opened 2026-02-25
chore(core): Implement redaction logic in redaction module

afitzek:
## Summary

Implement execution data redaction logic in `ExecutionRedactionService`, replacing the previous stub.

The `processExecution` method handles three cases based on the `redactExecutionData` query parameter:

1. **`true`** — Always redact (user explicitly requests redacted view)
2. **`false`** — User explicitly requests unredacted data; if policies allow it, we send unredacted data, otherwise checks `execution:reveal` scope via `canUserReveal`. Returns unmodified if allowed, throws `ForbiddenError` otherwise
3. **`undefined`** (default) — Evaluates the workflow's redaction policy:
   - `none` → return unmodified
   - `non-manual` → return unmodified only for manual executions, redact otherwise
   - `all` → always redact

Policy is resolved from `runtimeData.redaction.policy` (captured at execution time), falling back to `workflowData.settings.redactionPolicy`, defaulting to `none`.

`canUserReveal` checks the `execution:reveal` scope against all projects the workflow is shared with.

`applyRedaction` mutates execution data in place — replaces node output JSON with `{}`, removes binary data, and sets a `{ redacted: true }` marker on each item.

https://linear.app/n8n/issue/IAM-178/if-data-redaction-is-enabled-on-a-workflow-production-execution-data
https://linear.app/n8n/issue/IAM-182/implement-redactexecutiondatatrue-support
https://linear.app/n8n/issue/IAM-186/implement-redactexecutiondataundefined-behavior
https://linear.app/n8n/issue/IAM-184/implement-redactexecutiondatafalse-logic

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
+++ b/packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
@@ -1,11 +1,21 @@
 import { Logger } from '@n8n/backend-common';
-import type { IExecutionDb } from '@n8n/db';
+import { type IExecutionDb, SharedWorkflowRepository, type User } from '@n8n/db';
 import { Service } from '@n8n/di';
 
 import type {
 	ExecutionRedaction,
 	ExecutionRedactionOptions,
 } from '@/executions/execution-redaction';
+import {
+	INodeExecutionData,
+	ITaskDataConnections,
+	WorkflowExecuteMode,
+	WorkflowSettings,
+} from 'n8n-workflow';
+import { userHasScopes } from '@/permissions.ee/check-access';
+import { ForbiddenError } from '@/errors/response-errors/forbidden.error';
+
+const MANUAL_MODES: ReadonlySet<WorkflowExecuteMode> = new Set(['manual']);
 
 /**
  * Service responsible for redacting sensitive data from executions.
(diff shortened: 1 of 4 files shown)


phyllis-noester (approved):
looks good to me but i also lack some context from prs that i didn't review in the meantime

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
```diff
@@ -35,30 +48,120 @@ export class ExecutionRedactionService implements ExecutionRedaction {
...
-		this.logger.debug('Checking reveal permissions', {
-			userId,
-			executionId,
+	private async canUserReveal(user: User, execution: IExecutionDb): Promise<boolean> {
+		const sharedWorkflows = await this.sharedWorkflowRepository.findBy({
+			workflowId: execution.workflowId,
 		});
 
-		// Stub implementation: return false (no reveal permission)
+		for (const sw of sharedWorkflows) {
```
guillaumejacquart, line 97: I'm wondering if you could use the workflow service here:
```ts
const foundWorkflowWithPermission = await this.workflowFinderService.findWorkflowForUser(workflowId, user, [
	'execution:reveal',
]);
return !!foundWorkflowWithPermission;
```

guillaumejacquart (commented):
1 small comment

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
afitzek: Maybe, I'll look into this in the next PR. 👍

PR #26270, opened 2026-02-25
refactor(core): Implement decorator-based auto-discovery for breaking change rules

guillaumejacquart:
## Summary

This refactoring replaces the manual array-based registration of breaking change rules with a decorator-based auto-discovery system. Rules are now automatically discovered the decorator and registered via a metadata service, eliminating the need for manual barrel files and significantly simplifying the addition of new rules.

**Changes:**
- Introduced `@BreakingChangeRule({ version: 'v3' })` decorator pattern (matching existing `@BackendModule` conventions)
- Created `BreakingChangeRuleMetadata` service for automatic rule registration
- Migrated all 17 v2 rules from `@Service()` to `@BreakingChangeRule()`
- Removed manual barrel file (`rules/v2/index.ts`)
- Added integration test verifying rule discovery with minimal mocking
- Updated module README with new architecture and simplified rule addition workflow

**Testing:**
All changes include proper test coverage: 141 tests pass (138 original + 3 new discovery tests), lint and typecheck all pass.

## Review / Merge checklist

- [x] Tests included and passing
- [x] Docs (README) updated
- [ ] This is not a breaking change


guillaumejacquart:
> Hey hey I think I lack some context - is this an outcome from the last BE session? And when is the breaking change decorator used? is that if i introduce a breaking change to v2 (e.g. because a critical fix requires a breaking change). and what's the impact of this decorator on the customer? that they can see a breaking change has been introduced in the release notes?

Hey ! It was an outcome from AI on [this ticket ](https://linear.app/n8n/issue/IAM-284/remove-v2-migration-report-from-instance-settings) to improve the breaking change "engine". It's around the migration report (called breaking changes module in the backend). It list and try to detect all breaking changes if the users where to upgrade to version X. The decorator is just useful to add a breaking change "rule" without having to import and handle an array of rules

packages/cli/src/modules/breaking-changes/__tests__/breaking-changes.rule-discovery.test.ts
```diff
@@ -0,0 +1,38 @@
...
+describe('Breaking change rules auto-discovery', () => {
+	it('should register all rules', () => {
+		const metadata = Container.get(BreakingChangeRuleMetadata);
+		expect(metadata.getEntries().length).toBeGreaterThan(0);
+	});
+
+	it('should resolve all registered rules with valid metadata from the DI container', () => {
+		const metadata = Container.get(BreakingChangeRuleMetadata);
+
+		for (const entry of metadata.getEntries()) {
```
afitzek, line 16: How would this test check if there are all rules found without a list of all rules? It the getEntries function returns the empty array this will also pass.

packages/@n8n/decorators/src/breaking-change-rule/breaking-change-rule-metadata.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import type { IBreakingChangeRule } from './types';
+
+type RuleEntry = {
+	class: Constructable<IBreakingChangeRule>;
+	version: BreakingChangeVersion;
+};
+
+@Service()
+export class BreakingChangeRuleMetadata {
```
afitzek, line 9: Having this rule metadata service in the module restricts all rules to live in the module. Not sure what is the benefit in this case. It restricts all rules to live in the module, because we cannot reference module files from outside the module.

packages/cli/src/modules/breaking-changes/breaking-change-rule.decorator.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Container, Service, type Constructable } from '@n8n/di';
+
+import { BreakingChangeRuleMetadata } from './breaking-changes.rule-metadata.service';
+import type { IBreakingChangeRule } from './types';
+
+type BreakingChangeRuleOptions = {
+	version: BreakingChangeVersion;
+};
+
+export const BreakingChangeRule =
```
afitzek: Same here, having the interfaces defined in the decorator package would allow us to define rules everywhere in the code base, and keep the rules local to their deprecated functionality.

cstuncsik (commented):
@guillaumejacquart the PR looks good, my agent just warned about this:

  `packages/cli/src/modules/breaking-changes/breaking-changes.module.ts:8` now owns the `await import('./
  rules')` side effect, while `packages/cli/src/modules/breaking-changes/breaking-changes.service.ts:56`
  assumes rule metadata is already populated. That is correct for the current module-init path, but it
  creates a hidden ordering dependency: if `BreakingChangeService` is ever resolved outside
  `BreakingChangesModule.init()`, detection can silently run with an empty registry. I’d treat that as a
  test gap rather than a blocker and add one integration test around module initialization.

cstuncsik (approved):

PR #26288, opened 2026-02-26
fix(core): Fix Python and JS task runners for Windows platform

### LINKED ISSUE #26286: n8n npm installation for Windows is broken for Python node
### Bug Description

After installing npm package of n8n in Windows 10 I have created an Workflow which contains a Python code node. This node repeated giving me error that - "Python runner unavailable: Virtual environment is missing from this system...". Further investigation reveals that - 
1. In task-runner-process-py.js file '.venv/bin/python' path is hardcoded which is a LINUX path. 
2. The path '@n8n/task-runner-python' is missing from the npm installation but this is used in the JS file.

As a result Python script is not running in the workflow.

### To Reproduce

1. Install n8n npm package for Windows
2. run n8n server
3. Create a workflow with a Python code node.
4. Try to run a test run of the default Python code.

### Expected behavior

Python script should run properly without any error like - "Python runner unavailable: Virtual environment is missing from this system
Internal mode is intended only for debugging."

### Debug Info

1. In command prompt run the command - 
type "C:\Users\<username>\AppData\Roaming\npm\node_modules\n8n\dist\task-runners\task-runner-process-py.js"
See the hardcoded paths mentioned in "Bug Description".

### Operating System

Windows 10

### n8n Version

2.9.2

### Node.js Version

24.11.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

shortstacked:
## Summary
- Fix hardcoded Unix venv path (`.venv/bin/python` → platform-aware `Scripts/python.exe` on Windows)
- Fix `python3` command not existing on Windows (uses `python` instead)
- Verify `python --version` returns Python 3 on Windows (where `python` may resolve to Python 2)
- Fix `HOME` env var not set on Windows for both Python and JS task runners (falls back to `USERPROFILE`)

## Related Linear ticket(s), issues(s), and documentation

Closes #26286
https://linear.app/n8n/issue/GHC-7019

## Review / Merge checklist
- [x] Tests are passing (7/7 task runner tests pass)
- [x] No Linux/Docker impact — all changes are behind `process.platform === 'win32'` checks
- [x] Pre-commit hooks pass (biome)


packages/cli/src/task-runners/task-runner-process-py.ts
```diff
@@ -65,7 +65,8 @@ export class PyTaskRunnerProcess extends TaskRunnerProcessBase {
 	 */
 	static async checkRequirements(): Promise<'python' | 'venv' | null> {
 		try {
-			await asyncExec('python3 --version', { timeout: 5000 });
+			const pythonCmd = process.platform === 'win32' ? 'python' : 'python3';
```
tomi: If we allow also `python` command, we need to make sure it is actually python version 3 and not 2

shortstacked: Good catch! Thanks

tomi (approved):

PR #26310, opened 2026-02-26
chore(core): Introduce instance registry module

afitzek:
## Summary

This lays the foundation for the instance registry module.

closes https://linear.app/n8n/issue/IAM-325/foundations-types-and-module-scaffolding

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/api-types/src/instance-registry-types.ts
+++ b/packages/@n8n/api-types/src/instance-registry-types.ts
@@ -1,0 +1,49 @@
+import { z } from 'zod';
+
+// Versioned schema (V1 with all fields from RFC)
+const InstanceRegistrationSchemaV1 = z
+	.object({
+		schemaVersion: z.literal(1),
+		instanceKey: z.string(),
+		hostId: z.string(),
+		instanceType: z.enum(['main', 'worker', 'webhook']),
+		instanceRole: z.enum(['leader', 'follower', 'unset']),
+		version: z.string(),
+		registeredAt: z.number(),
+		lastSeen: z.number(),
+	})
+	.passthrough();
+
+// Discriminated union for future schema versions
+export const instanceRegistrationSchema = z.discriminatedUnion('schemaVersion', [
+	InstanceRegistrationSchemaV1,
+]);
+
+export type InstanceRegistration = z.infer<typeof instanceRegistrationSchema>;
+
+const ClusterVersionMismatchSchema = z.object({
+	versions: z.array(
+		z.object({
+			version: z.string(),
+			instances: z.array(
+				z.object({
+					instanceKey: z.string(),
+					hostId: z.string(),
+					instanceType: z.enum(['main', 'worker', 'webhook']),
+					instanceRole: z.enum(['leader', 'follower', 'unset']),
+				}),
+			),
+		}),
+	),
+});
+
+// Version mismatch type (for REST API response)
+export type ClusterVersionMismatch = z.infer<typeof ClusterVersionMismatchSchema>;
+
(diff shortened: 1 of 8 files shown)


packages/@n8n/api-types/src/instance-registry-types.ts
```diff
@@ -0,0 +1,49 @@
+import { z } from 'zod';
+
+// Versioned schema (V1 with all fields from RFC)
+const InstanceRegistrationSchemaV1 = z
+	.object({
+		schemaVersion: z.literal(1),
+		instanceKey: z.string(),
+		hostId: z.string(),
+		instanceType: z.enum(['main', 'worker', 'webhook']),
+		instanceRole: z.enum(['leader', 'follower', 'unset']),
```
phyllis-noester, line 10: what's unset? when do you get into that state?

packages/cli/src/modules/instance-registry/storage/instance-storage.interface.ts
```diff
@@ -0,0 +1,66 @@
+import type { InstanceRegistration } from '@n8n/api-types';
+
+/**
+ * Storage abstraction for instance registrations
+ * Implementations: RedisInstanceStorage, MemoryInstanceStorage
+ */
+export interface InstanceStorage {
+	/** Tagged union discriminator for runtime type checking */
+	kind: 'redis' | 'memory';
```
phyllis-noester, line 9: type?

packages/cli/src/modules/instance-registry/instance-registry.module.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Tracks all n8n processes (main, worker, webhook) in distributed deployments.
+ * Provides cluster visibility, version mismatch detection, and health checks.
+ *
+ * Runs on all instance types (no instanceTypes filter).
+ */
+@BackendModule({ name: 'instance-registry' })
+export class InstanceRegistryModule implements ModuleInterface {
+	async init() {
+		const logger = Container.get(Logger);
```
phyllis-noester, line 17: this is empty bc just the foundation correct?

phyllis-noester (approved):
It looks alright to me, but I think it's quite difficult to review properly because there is no business logic yet. e.g. i couldn't say whether something is missing or incorrect in packages/cli/src/modules/instance-registry/instance-registry.types.ts because these types aren't used yet. I think going forward it would be better that we spit work into small increments that add user value or are testable, otherwise it;s hard to give actionable feedback on the increments

packages/cli/src/modules/instance-registry/storage/instance-storage.interface.ts
afitzek: Yes there will be a redis and an inmemory storage for now

packages/@n8n/api-types/src/instance-registry-types.ts
afitzek: Ideally not, I just wanted to have the flexibiltiy in case there is ambiguity in the determination

phyllis-noester: should we use `none` or something more generic or otherwise allow it to be undefined?

packages/cli/src/modules/instance-registry/storage/instance-storage.interface.ts
phyllis-noester: ah i just suggested to call this type because I've never seen kind used in this context. it's nit pick but I'm just wary we don't introduce too much custom logic

PR #26326, opened 2026-02-27
chore(core): Add redactionInfo object to execution data responses

afitzek:
## Summary

Adds an optional `redactionInfo` field to `IRunExecutionDataV1` (in `packages/workflow`) that travels with the execution data object itself. When an execution's data is redacted, this field describes the redaction so that callers know whether data was withheld and whether they can request unredacted access.

**New type (`packages/workflow`):**
```typescript
export interface RedactionInfo {
  isRedacted: boolean;
  reason: string;      // 'workflow_redaction_policy' | 'user_requested'
  canReveal: boolean;  // true if the requester is allowed to see unredacted data
}
```

**`canReveal` logic:** A user can reveal unredacted data when either:
- The workflow's redaction policy implicitly allows it (`policy === 'none'`, or `policy === 'non-manual'` and the execution mode is `manual`); **or**
- The user holds the `execution:reveal` scope (global owners/admins, project admins).

This also fixes a gap in the `redactExecutionData=false` (explicit reveal) path: users without the `execution:reveal` scope were previously blocked with a 403 even when the policy would not have redacted the data in the first place.

**Implementation details:**
- `applyRedaction()` in `ExecutionRedactionService` now accepts a `canReveal` parameter and writes `execution.data.redactionInfo`
- A new `policyAllowsReveal()` private method encapsulates the policy-based check
- `canUserReveal()` is simplified to delegate to `WorkflowFinderService.findWorkflowForUser()` with the `execution:reveal` scope
- Non-redacted paths leave `redactionInfo` as `undefined`

https://linear.app/n8n/issue/IAM-181

## Review / Merge checklist

- [X] Tests included.

### DIFF
--- a/packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
+++ b/packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
@@ -1,5 +1,5 @@
 import { Logger } from '@n8n/backend-common';
-import { type IExecutionDb, SharedWorkflowRepository, type User } from '@n8n/db';
+import { type IExecutionDb, type User } from '@n8n/db';
 import { Service } from '@n8n/di';
 
 import type {
@@ -12,8 +12,8 @@ import {
 	WorkflowExecuteMode,
 	WorkflowSettings,
 } from 'n8n-workflow';
-import { userHasScopes } from '@/permissions.ee/check-access';
 import { ForbiddenError } from '@/errors/response-errors/forbidden.error';
+import { WorkflowFinderService } from '@/workflows/workflow-finder.service';
 
 const MANUAL_MODES: ReadonlySet<WorkflowExecuteMode> = new Set(['manual']);
 
@@ -25,7 +25,7 @@ const MANUAL_MODES: ReadonlySet<WorkflowExecuteMode> = new Set(['manual']);
 export class ExecutionRedactionService implements ExecutionRedaction {
 	constructor(
 		private readonly logger: Logger,
-		private readonly sharedWorkflowRepository: SharedWorkflowRepository,
+		private readonly workflowFinderService: WorkflowFinderService,
 	) {}
 
 	/**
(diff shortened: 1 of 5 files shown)


packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
```diff
@@ -50,10 +50,13 @@ export class ExecutionRedactionService implements ExecutionRedaction {
 	): Promise<IExecutionDb> {
 		if (options.redactExecutionData === true) {
 			// user wants redacted data, this is always fine!
-			this.applyRedaction(execution);
+			const canReveal =
+				this.policyAllowsReveal(execution) || (await this.canUserReveal(options.user, execution));
```
guillaumejacquart, line 54: I don't understand why we check the policy. The policy is there to know whether or not the data is redacted, not whether or not we can reveal, no ?

guillaumejacquart (commented):
1 question

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
afitzek: There is an edge case: If a user does not have the correct permission, but the policy does not automatically redact it, the use can reveal it, because it is not enforced by the policy.

phyllis-noester: How is the policy set again? afaik we decided to only set a policy on project/instance in a future and start by jsut evaluating the workflow setting

afitzek: This policy is set in the workflow settings and taken there during workflow execution and stored in the execution context, so that changing the policy in the workflow setting does not affect the redaction behavior of an execution, but the policy is persisted to the value at execution time.

guillaumejacquart (approved):

PR #26354, opened 2026-02-27
feat(core): Track dynamic credential resolution per node execution in ITaskData

afitzek:
## Summary

Track when a node's credential was resolved dynamically (via the EE dynamic credentials module) by adding an optional `usedDynamicCredentials` flag to `ITaskData`.

### How it works

During workflow execution, `CredentialsHelper.getDecrypted()` calls `DynamicCredentialsProxy.resolveIfNeeded()`, which delegates to `DynamicCredentialService.resolveIfNeeded()`. Previously, that call returned the raw decrypted credential object, making it impossible to tell whether a _real_ dynamic resolution occurred (i.e. `resolver.getSecret()` succeeded) vs. a fallback to static data.

This PR introduces a `{ data, wasDynamic }` result type throughout the resolution chain so the flag can propagate upward cleanly:

1. **`CredentialResolutionResult`** — new return type for `ICredentialResolutionProvider.resolveIfNeeded()` carrying both the data and a `wasDynamic: boolean`. Only the successful `resolver.getSecret()` path in `DynamicCredentialService` sets `wasDynamic: true`; all fallback/non-resolvable paths return `false`.

2. **`additionalData` side-channel** — `IWorkflowExecuteAdditionalData` gets a new optional `currentNodeUsedDynamicCredentials` field. `CredentialsHelper` writes `true` to it when `wasDynamic` is set; `WorkflowExecute` resets it to `false` before each node and reads it when building `ITaskData`.

3. **`ITaskData.usedDynamicCredentials`** — the per-node-run flag stored in execution run data. Absent (i.e. `undefined`) when no dynamic resolution occurred, keeping serialised payloads lean.

### Changes

| Package | File | Change |
|---|---|---|
| `workflow` | `interfaces.ts` | Add `usedDynamicCredentials?` to `ITaskData`; add `currentNodeUsedDynamicCredentials?` to `IWorkflowExecuteAdditionalData` |
| `cli` | `credential-resolution-provider.interface.ts` | Add `CredentialResolutionResult` type; update `resolveIfNeeded()` return type |
| `cli` | `dynamic-credentials-proxy.ts` | Propagate new return type through proxy |
| `cli` | `dynamic-credential.service.ts` | Return `wasDynamic: true` only on successful `getSecret()` |
| `cli` | `credentials-helper.ts` | Capture `wasDynamic`, write to `additionalData` |
| `core` | `workflow-execute.ts` | Reset flag before node; write to `ITaskData` |

closes https://linear.app/n8n/issue/IAM-308/check-validity-of-dynamic-cred-metrics

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/credentials/credential-resolution-provider.interface.ts
+++ b/packages/cli/src/credentials/credential-resolution-provider.interface.ts
@@ -14,6 +14,12 @@ export type CredentialResolveMetadata = {
 	isResolvable: boolean;
 };
 
+export type CredentialResolutionResult = {
+	data: ICredentialDataDecryptedObject;
+	/** True only when the credential was actually resolved via a dynamic resolver (not a fallback to static data). */
+	wasDynamic: boolean;
+};
+
 /**
  * Interface for credential resolution providers.
  * Implementations can provide dynamic credential resolution logic.
@@ -27,7 +33,7 @@ export interface ICredentialResolutionProvider {
 	 * @param staticData The decrypted static credential data
 	 * @param additionalData Additional workflow execution data for context and settings
 	 * @param canUseExternalSecrets Whether the credential can use external secrets for expression resolution
-	 * @returns Resolved credential data (either dynamic or static)
+	 * @returns Resolved credential data and a flag indicating whether dynamic resolution occurred
 	 */
 	resolveIfNeeded(
 		credentialsResolveMetadata: CredentialResolveMetadata,
@@ -35,5 +41,5 @@ export interface ICredentialResolutionProvider {
 		executionContext?: IExecutionContext,
 		workflowSettings?: IWorkflowSettings,
 		canUseExternalSecrets?: boolean,
-	): Promise<ICredentialDataDecryptedObject>;
+	): Promise<CredentialResolutionResult>;
 }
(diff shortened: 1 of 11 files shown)


packages/cli/src/credentials/credential-resolution-provider.interface.ts
```diff
@@ -14,6 +14,12 @@ export type CredentialResolveMetadata = {
 	isResolvable: boolean;
 };
 
+export type CredentialResolutionResult = {
+	data: ICredentialDataDecryptedObject;
+	/** True only when the credential was actually resolved via a dynamic resolver (not a fallback to static data). */
+	wasDynamic: boolean;
```
phyllis-noester: nit pick but I think this should be `isDynamic` I haven't seen as use past tense anywhere

packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -809,7 +809,9 @@ export class TelemetryEventRelay extends EventRelay {
 			is_manual: false,
 			version_cli: N8N_VERSION,
 			success: false,
-			used_dynamic_credentials: !!runData?.data?.executionData?.runtimeData?.credentials,
```
phyllis-noester, line 812: did this just include everything execution that contained a credential

afitzek: Yes, the condition was as soon as the credential context is set, we count it as a dynamic execution.

guillaumejacquart (approved):
Nice work !

PR #26355, opened 2026-02-27
fix: Implement raw query parameters field for Microsoft SQL node

dlavrenuek:
## Summary

Add `Query Parameters` field to Microsoft SQL Node `Execute Query` operation. This allows passing user controlled input parameters to the query through a prepared statement rather than expressions/string replacement.

https://linear.app/n8n/issue/NODE-4445

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Microsoft/Sql/GenericFunctions.ts
```diff
@@ -259,8 +259,21 @@ export async function executeSqlQueryAndPrepareResults(
 	pool: mssql.ConnectionPool,
 	rawQuery: string,
 	itemIndex: number,
+	queryValues: Array<string | number | IDataObject> = [],
```
ShireenMissi: boolean is missing

```suggestion
	queryValues: Array<string | number | boolean | IDataObject> = [],
```

packages/nodes-base/nodes/Microsoft/Sql/MicrosoftSql.node.ts
```diff
@@ -279,7 +305,31 @@ export class MicrosoftSql implements INodeType {
...
+						queryReplacement = queryReplacement.split(',').map((entry) => entry.trim());
+					}
+					if (queryReplacement !== '' && !Array.isArray(queryReplacement)) {
+						// convert non-string single expression values to arrays
+						queryReplacement = [queryReplacement];
+					}
+
+					if (Array.isArray(queryReplacement)) {
+						queryValues = queryReplacement;
+					} else if (queryReplacement) {
```
ShireenMissi: This `else if` branch can never be reached and the error will never throw. By the time execution reaches here,
  `queryReplacement` is either `''` (falsy, skips the branch) or has already been wrapped into an array at line 319

ShireenMissi (commented):
I tested locally and it works as expected 👏
left a few comments

packages/nodes-base/nodes/Microsoft/Sql/GenericFunctions.ts
dlavrenuek: fixed

packages/nodes-base/nodes/Microsoft/Sql/MicrosoftSql.node.ts
dlavrenuek: When I first looked at this code I also though that it looks strange but then I seemed to have found a case where this branch is executed. However now I can't remember what it was and also can't see a case where this branch would be reached. Removed it

ShireenMissi (approved):
LGTM 🎉

PR #26364, opened 2026-02-28
fix: Merge by sql update

michael-radency:
## Summary

implementation update

https://linear.app/n8n/issue/NODE-4518

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Merge/v3/helpers/sandbox-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+	tableData: unknown[][],
+	query: string,
+): Promise<IDataObject[]> {
+	// UUID per invocation so concurrent calls sharing the singleton context
+	// don't collide on alasql.databases.
+	const dbId = randomUUID();
+
+	// Table data is embedded as a JSON literal so no live objects cross the boundary.
+	const script = `(function() {
+		const __rows = JSON.parse(${JSON.stringify(JSON.stringify(tableData))});
```
DawidMyslak, line 56: The double-serialization pattern is clever but non-obvious. A one-line comment explaining why it's double-wrapped (to ensure data enters the isolate as a parsed JSON literal, never as live objects) would help future maintainers 👍.

packages/nodes-base/nodes/Merge/v3/helpers/sandbox-utils.ts
```diff
@@ -0,0 +1,72 @@
+import { readFile } from 'node:fs/promises';
+import { randomUUID } from 'node:crypto';
+
+import ivm from 'isolated-vm';
+
+import type { IDataObject } from 'n8n-workflow';
+
+// Singleton – recreated only after resetSandboxCache() (tests) or isolate disposal.
```
DawidMyslak, line 8: Let's consider this hypothetical scenario: Workflow A runs `CREATE FUNCTION foo(x) RETURNS NUMBER BEGIN RETURN x * 2 END; SELECT foo(val) FROM input1`. Because the singleton context shares a single `alasql` instance across the entire process, that `foo` function now lives permanently on `alasql.fn`. Later, Workflow B (possibly owned by a different user) runs `SELECT foo(salary) FROM input1` - and it succeeds, even though B never defined `foo`.

Note that this is a regression from the previous implementation, which froze `alasql.fn` via `Object.freeze(alasql.fn)` to prevent exactly this. The new code doesn't carry over that freeze. A simple fix would be to add `Object.freeze(alasql.fn)` after loading the bundle:
```typescript
await sandboxContext.eval(await readFile(alasqlBundlePath, 'utf-8'));
await sandboxContext.eval('Object.freeze(alasql.fn)');
```

Not a security concern since everything stays inside the isolate, but it's a correctness regression worth addressing.

packages/nodes-base/nodes/Merge/test/v3/combineBtSql.test.ts
```diff
@@ -1,186 +1,271 @@
...
+} from '../../v3/helpers/sandbox-utils';
+
+describe('combineBySql sandbox (isolated-vm)', () => {
+	beforeEach(() => {
+		resetSandboxCache();
+	});
+
+	afterEach(() => {
+		resetSandboxCache();
+	});
```
DawidMyslak: Resetting the sandbox cache in both `beforeEach` and `afterEach` means every single test pays the cost of loading the alasql bundle from disk into a fresh isolate. How about using `beforeAll/afterAll` for the isolate and only resetting between describe blocks that need it?

packages/nodes-base/nodes/Merge/test/v3/combineBtSql.test.ts
```diff
@@ -1,186 +1,271 @@
...
-			// Verify both are frozen after calling freezeAlasql
-			expect(Object.isFrozen(mockAlasql.fn)).toBe(true);
-			if (mockAlasql.yy) {
-				expect(Object.isFrozen(mockAlasql.yy)).toBe(true);
-			}
+			// Database count must not grow – each invocation cleans up after itself
+			expect(countAfter).toBe(countBefore);
 		});
 	});
 });
```
DawidMyslak, line 322: Could we add a proper test covering our update?  

Something like:
```typescript
it('should not allow arbitrary code execution via SQL expressions', async () => {
   const context = await loadAlaSqlSandbox();
   await expect(
     runAlaSqlInSandbox(
       context,
       [[{ id: 1 }]],
       `SELECT {}.constructor->constructor->[call]('','return {result: process.getBuiltinModule("child_process").execSync("ls").toString()}') as toJSON`,
     ),
   ).rejects.toThrow();
 });
 ```

DawidMyslak (changes requested):

DawidMyslak (approved):
👍

PR #26366, opened 2026-02-28
test: Move test discovery and orchestration into janitor

shortstacked:
## Summary
- Moves test discovery (spec enumeration + capability extraction) from Playwright `--list` + regex into janitor's AST-based analyzer
- Moves test orchestration (capability-aware shard distribution) into janitor, replacing the standalone `distribute-tests.mjs` logic
- Wires impact analysis into orchestration — when a PR only changes Playwright files, CI runs only the affected tests instead of the full E2E suite

## Key changes
- **Test discovery** (`janitor discover`): AST-based spec discovery with configurable skip tags, replacing `playwright test --list` + regex parsing
- **Orchestration** (`janitor orchestrate`): Capability-aware bin-packing with metrics-based load balancing, now with `--impact` and `--base` flags for filtered runs
- **CI integration**: `playwright-only` detection flows from paths-filter → `distribute-tests.mjs --impact --base` → `janitor orchestrate --impact`
- **Bug fixes**: `test.fixme('title', opts, callback)` wrapper pattern now correctly detected as skipped; `--shard-index` NaN bypass fixed

## Test plan
- [ ] Unit tests pass (janitor source changes trigger unit test filter)
- [ ] E2E tests pass (full suite — this PR triggers both unit and e2e filters)
- [ ] Typecheck passes
- [ ] Lint passes
- [ ] Verify `pnpm janitor discover --json` output matches expected spec count
- [ ] Verify `pnpm janitor orchestrate --shards=4 --impact --base=origin/master --json` returns filtered results
- [ ] Verify impact-only path activates correctly on a Playwright-only PR (manual test after merge)


shortstacked:
> rootInfluence

Hey @rootInfluence ,
Worth noting that this is from an internal developer and updates all the workflows as expected. Given that's the case there is no migration path needed.

packages/testing/playwright/scripts/distribute-tests.mjs
```diff
@@ -2,38 +2,26 @@
...
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const ROOT_DIR = path.resolve(__dirname, '../../../..');
-const METRICS_PATH = path.join(ROOT_DIR, '.github/test-metrics/playwright.json');
 const PLAYWRIGHT_DIR = path.resolve(__dirname, '..');
-const DEFAULT_DURATION = 60000; // 1 minute default (accounts for container startup)
-const E2E_PROJECT = 'multi-main:e2e';
-const CONTAINER_STARTUP_TIME = 22500; // 22.5 seconds average (heavier stacks with extra services take longer)
-const MAX_GROUP_DURATION = 5 * 60 * 1000; // 5 minutes - split groups larger than this
+const JANITOR_CLI = path.resolve(__dirname, '..', '..', 'janitor', 'dist', 'cli.js');
```
Matsuuu, line 22: I wonder as this is also a dependency on this package, could we just go 

```javascript
const JANITOR_CLI = path.resolve(__dirname, '..', 'node_modules', '@n8n', 'playwright-janitor', '', 'dist', 'cli.js);
```

Not that the above code doesn't work, it just binds the code inside the package to monorepo structures outside of the package.

`node_modules/@n8n/playwright-janitor` symlinks to `packages/@n8n/testing/janitor` anyways so it's just semantics.

packages/testing/janitor/src/cli.ts
```diff
@@ -402,6 +406,101 @@ function runRules(options: CliOptions): void {
...
+	const analyzer = new TestDiscoveryAnalyzer(project);
+	const report = analyzer.discover();
+
+	console.log(JSON.stringify(report, null, 2));
+}
+
+async function runOrchestrate(options: CliOptions): Promise<void> {
+	const config = getConfig();
+
+	if (!options.shards || options.shards < 1) {
```
Matsuuu, line 421: Is there a reason we don't have a "sane default" for shards when omitted?

The examples in the markdown files use 14, should we use that as a default?

Also a side note, how did we end up with 14? Doesn't our e2e matrix (`test-e2e-ci-reusable.yml`) have 16 as their matrix?

Matsuuu (approved):
Quite clear changes, just a couple of nits / questions to resolve before merging

PR #26370, opened 2026-02-28
fix(store): improve WebSocket client ID on tab reload and duplication

luka-mimi:
## Problem

1. Each browser tab needs a unique `clientId` to maintain its own WebSocket
   connection with the n8n backend, so execution push messages are routed to
   the correct tab.
2. Tab duplication copies `sessionStorage`, meaning the duplicated tab inherits
   the same `n8n-client-id` as the original. Both tabs then share one
   `clientId`, causing push messages to be misrouted between them.
3. Page reload must reuse the existing `clientId`, otherwise in-flight
   execution messages from the backend — already addressed to the old ID —
   are lost.
4. The old code unconditionally reused the stored ID, making it impossible to
   tell a reload (reuse needed) from a duplication (new ID needed).

## Fix

Detect navigation type using
`performance.getEntriesByType('navigation')[0].type`:

- `"reload"` → reuse the stored ID (keeps push messages alive)
- anything else → generate a new ID (breaks the shared-ID problem on
  duplication)

## Result

| Scenario       | Before              | After       |
| -------------- | ------------------- | ----------- |
| First visit    | New ID ✅           | New ID ✅   |
| Page reload    | Reuse ID ✅         | Reuse ID ✅ |
| Tab duplication | Reuse ID (conflict) ❌ | New ID ✅ |

https://github.com/n8n-io/n8n/issues/26369
fixes #<26369>

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/@n8n/stores/src/__tests__/useRootStore.test.ts
+++ b/packages/frontend/@n8n/stores/src/__tests__/useRootStore.test.ts
@@ -1,0 +1,134 @@
+import { randomString } from 'n8n-workflow';
+import { createPinia, setActivePinia } from 'pinia';
+
+import { useRootStore } from '../useRootStore';
+
+vi.mock('n8n-workflow', () => ({
+	randomString: vi.fn(() => 'DEFAULTMOCK'),
+	setGlobalState: vi.fn(),
+}));
+
+vi.mock('../metaTagConfig', () => ({
+	getConfigFromMetaTag: vi.fn(() => null),
+}));
+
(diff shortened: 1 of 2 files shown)


Joffcom:
Hey @luka-mimi 

Don't forget to add tests so we can pass it on for review.

luka-mimi:
> Hey @luka-mimi
> 
> Don't forget to add tests so we can pass it on for review.

I have added the test cases. However, I don’t think these test cases can fully simulate browser behavior. This bug only occurs when duplicating a tab in the browser.

PR #26376, opened 2026-02-28
feat(lineMessagingApi Node): Add LINE Messaging API node and deprecate Line Notify node

4geru:
## Summary

- Deprecate the **Line Notify** node (service shut down March 31, 2025): mark as hidden and update the notice
- Add new **LINE Messaging API** node with the following capabilities:
  - **Message** resource:
    - Reply – respond to a webhook event with a reply token
    - Push – send a message to a user
    - Broadcast – send a message to all users
    - Multicast – send a message to multiple users at once
  - **Custom API Call** – make any authenticated request to the LINE API
- Add new **LINE Messaging API Trigger** node:
  - Listens for webhook events: Message, Follow, Unfollow, Postback
- Add **LINE API credentials** supporting Production and Test environments (channel access token + channel secret)

### `LINE API` (Credential)

Supports both **Production** and **Test** channel credentials (Channel Access Token + Channel Secret).

スクリーンショット 2026-03-01 8 06 56

---

### Node UI

スクリーンショット 2026-03-01 8 06 16

### Execution Result

スクリーンショット 2026-03-01 8 05 18

- [LINE Node API is archived - n8n Community](https://community.n8n.io/t/line-node-api-is-archived/270710)
- [Line Notify service shutdown announcement](https://notify-bot.line.me/closing-announce)

## Review / Merge checklist

- [x] Tests included.
- [ ] PR labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)


Joffcom:
Hey @4geru 

Removing a node isn't always a simple process if we were to merge this it would possibly break any workflows that are still using it without issue so we need to deprecate it.

To do that we would set the node as hidden and add a notice to it then in a future version like v3 we would remove the code.

If they are adding a new api then we would update the existing node to use that as a new version.

PR #26389, opened 2026-03-01
fix(core): Handle engine requests from sub-agents in makeHandleToolInvocation

MrAliHasan:
## Summary

Resolves #26373

Sub-agents connected via `AgentToolV3` fail when their tools (MCP Client Tool, Vector Store in retrieve-as-tool mode, HITL nodes) return an `EngineRequest`:

```
Error: The Tool attempted to return an engine request, which is not supported in Agents
```

**Root cause**: `makeHandleToolInvocation()` executes sub-agents inside a synchronous tool callback. When the sub-agent's tools need to communicate with external services (MCP, vector stores, etc.), they return an `EngineRequest` to request execution from the workflow loop. But `makeHandleToolInvocation` cannot relay this to the `WorkflowExecute.processRunExecutionData()` loop — `mapResult()` rejects the `EngineRequest` with an error string.

## Solution

Execute sub-agent tool nodes **in-process** when an `EngineRequest` is detected, then re-invoke the sub-agent with the assembled `EngineResponse`.

### New functions

| Function | Purpose |
|---|---|
| `executeSubAgentTools()` | Executes tool nodes referenced in `EngineRequest.actions` in-process using `SupplyDataContext`. Handles recursive `EngineRequest`s from nested tools with a depth limit of 10. |
| `resolveSubAgentEngineRequests()` | Loops the engine request → tool execution → engine response → sub-agent re-invocation cycle with an iteration limit of 50. |

### Modified functions

| Function | Change |
|---|---|
| `makeHandleToolInvocation()` | Added 5 optional parameters for workflow context (`workflow`, `additionalData`, `mode`, `closeFns`, `abortSig`). Detects `EngineRequest` returns and delegates to `resolveSubAgentEngineRequests()`. |
| `getInputConnectionData()` | Passes workflow context to `makeHandleToolInvocation()` at the call site. |

### Architecture

```
Parent Agent
  └─ makeHandleToolInvocation()
       └─ Sub-Agent.execute()
            └─ Returns EngineRequest {actions: [...]}
                 └─ resolveSubAgentEngineRequests()  ← NEW
                      └─ executeSubAgentTools()      ← NEW
                           └─ ToolNode.execute() (in-process)
                      └─ Sub-Agent.execute(engineResponse)
                           └─ Returns final data [[{json: ...}]]
```

## Design Decisions

1. **Backward compatible**: All new parameters are optional. Without workflow context, the original error message is preserved — no breaking changes for existing callers.

2. **Safety guards**: Two layers of protection against infinite loops:
   - `MAX_ENGINE_REQUEST_ITERATIONS = 50` for the main request/response loop
   - `MAX_RECURSION_DEPTH = 10` for nested `EngineRequest`s within tool execution

3. **Error propagation**: If a tool node throws during in-process execution, the error is captured as `ITaskData` with `executionStatus: 'error'` and returned to the sub-agent via `EngineResponse`, allowing the agent to handle or report the error gracefully.

4. **Complexity management**: Extracted the engine request loop into `resolveSubAgentEngineRequests()` to keep `makeHandleToolInvocation()`'s cyclomatic complexity under the 27 limit (final: 26).

## Testing

- Updated existing engine request test to verify backward compatibility (fallback error when no workflow context)
- Added new test: `should execute sub-agent tools in-process when engine request is returned with workflow context`
- **All 158 tests pass**, zero new lint errors

## Related Issue

Fixes #26373

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/__tests__/get-input-connection-data.test.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/__tests__/get-input-connection-data.test.ts
@@ -633,12 +633,13 @@ describe('makeHandleToolInvocation', () => {
 		]);
 	});
 
-	it('should handle engine requests and return a warning message', async () => {
+	it('should handle engine requests with fallback error when no workflow context is provided', async () => {
 		const mockContext = mock<IExecuteFunctions>();
 		contextFactory.mockReturnValue(mockContext);
 		const mockResult: EngineRequest = { actions: [], metadata: {} };
 		execute.mockResolvedValueOnce(mockResult);
 
+		// Without workflow context, engine requests fall back to the error message
 		const handleToolInvocation = makeHandleToolInvocation(
 			contextFactory,
 			connectedNode,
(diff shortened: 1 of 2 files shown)


MrAliHasan:
Addressed all Aikido bot feedback:

**1. IDOR concern (`action.nodeName`)** — This is a false positive. The `EngineRequest` is generated server-side by the sub-agent's own `execute()` call within the trusted n8n execution engine — `action.nodeName` is **not** user-supplied input. This is consistent with how `WorkflowExecute.processRunExecutionData()` handles the same `EngineRequest.actions` pattern (see [workflow-execute.ts L1649](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/workflow-execute.ts#L1649)). Added a code comment documenting this. Node existence is validated with an explicit `ApplicationError` throw.

**2. `iteration++` in while condition** — ✅ Fixed. Moved to its own line inside the loop body.

**3. `runIdx` naming** — ✅ Fixed. Renamed to `runIndex` for clarity.

fq393:
2.21.4  There is still this report wrong

DeveloperTheExplorer:
Closing in favour of PR #31072. Thanks for the work and bringing up this issue.

PR #26404, opened 2026-03-02
fix(core): Make n8n-workflow a peer dependency in ai-utilities

DawidMyslak:
## Summary

When published to npm, `@n8n/ai-utilities` had `n8n-workflow` as a **direct dependency**, which allowed npm/pnpm to install a separate nested copy. Community nodes using `@n8n/ai-node-sdk` would then end up with **two instances** of `n8n-workflow` at runtime — one from the host (n8n) and one nested inside `@n8n/ai-utilities` — causing type mismatches and `instanceof` failures.

This PR moves `n8n-workflow` from `dependencies` to `peerDependencies` in `@n8n/ai-utilities`, so a single host-provided instance is used throughout.

https://linear.app/n8n/issue/NODE-4532

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/ai-node-sdk/package.json
```diff
@@ -33,6 +33,9 @@
   "dependencies": {
     "@n8n/ai-utilities": "workspace:*"
   },
+  "peerDependencies": {
```
yehorkardash: Do we need it here, if ai-utilities defines n8n-workflow dependency?

DawidMyslak: Ah, since community node templates already declare their own `"n8n-workflow": "*"` peer dependency independently, you're right - the SDK doesn't need to duplicate that.

yehorkardash (approved):

PR #26416, opened 2026-03-02
fix(core): Implement data tables name collision detection on pull

MiloradFilipovic:
## Summary
Detect name collisions when pulling data tables and stopping the pull process in that case (mimicking [the same logic from tags pull](https://github.com/n8n-io/n8n/blob/b883b3c41c635295047f5b2d3ec08fe59604ab68/packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts#L1082)).

image

Fixes ADO-4860

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
```diff
@@ -1350,6 +1350,17 @@ export class SourceControlImportService {
 
 				const targetProjectId = targetProject.id;
 
+				// Check for name collision: a local table with the same name but different ID
+				const existingByName = await this.dataTableRepository.findOne({
+					where: { name: dataTable.name },
+					select: ['id'],
+				});
```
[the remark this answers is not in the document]
nikhilkuria: This is a valid suggestion

packages/cli/src/modules/source-control.ee/source-control-status.service.ee.ts
```diff
@@ -626,6 +622,33 @@ export class SourceControlStatusService {
 			}
 		}
 
+		for (const local of dataTablesLocal) {
```
nikhilkuria: why are these changes needed?

MiloradFilipovic: This is the actual name collision detection. Earlier loops only catch ID collisions but we can probably merge these.

packages/cli/src/modules/source-control.ee/source-control-status.service.ee.ts
```diff
@@ -598,16 +598,35 @@ export class SourceControlStatusService {
 					updatedAt: new Date().toISOString(),
 					owner: local.ownedBy ?? undefined,
 				});
+
+				// Check for cross-ID name collision (different tables sharing the same name)
+				const nameCollision = dataTablesRemote.find(
```
nikhilkuria: I think we need to check for projects here as well
https://www.loom.com/share/04f648b67ab24356b35aa6797912770e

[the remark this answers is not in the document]
MiloradFilipovic: Thanks, makes sense

MiloradFilipovic: Yeah we are missing that but I think the issue here is a bit deeper. Looks like we cannot safely distinguish between deleted and never pushed tables. Let me try a few approaches but I think we need a bit more involved solution here.

MiloradFilipovic: Ok, I pushed an alternative way to address this, take a look if it makes sense. Also @RicardoE105 would appreciate if you can take a look as well

nikhilkuria: When there is a name collision on a (project, data table) pair, I think the whole pull operation should fail. It seems it completes partially now
https://www.loom.com/share/cb8798fd04384539b660f8c02a7e44ff

nikhilkuria: We can leave it as it is. The important part is to avoid the loss of data in the prod data table.

nikhilkuria (dismissed):

nikhilkuria (dismissed):

nikhilkuria (approved):

rehevkor5:
This breaks the pull whenever someone:

- deletes something
- creates something new with the same name
- pushes + pulls

That is a normal thing which people can do (there's nothing unusual about it, nothing warns you or anything), and if any of our n8n users does it then env pull is broken for everyone.

The pull logic should be smart enough to perform the operations in the right order:

1. delete the thing that was deleted
2. then create the new thing

If you do that, there's no conflict and everything works.

rehevkor5:
Or, if names are actually the ids (and not the synthetic key, which is weird but whatever), then make it impossible to create anything which has the same name as anything else that previously existed. That would prevent these conflicts from occurring in the first place.

PR #26418, opened 2026-03-02
feat(editor): Allow switching versions in workflow history diff view

alielkhateeb:
## Summary

Adds interactive version selectors to the workflow history diff view, allowing users to switch between any two versions for comparison without navigating away.

https://www.loom.com/share/9073407f512d4ac08aa371d77794675e

**Changes:**
- Added `WorkflowHistoryVersionSelect` component: a dropdown select for choosing workflow history versions, grouped by date with status indicators (latest/published/default)
- Added `WorkflowHistoryVersionDot` component: a visual dot indicator showing version status
- Added `useWorkflowHistoryVersionOptions` composable: manages version option state, labels, and publish metadata
- Updated `WorkflowHistoryDiff.vue`: replaced static version labels with interactive dropdowns; added race-condition protection via request IDs when loading versions; watch-based reactivity for version changes
- Updated `WorkflowHistoryList`, `WorkflowHistoryListItem`, and `WorkflowHistoryContent` to pass available versions down to the diff view

https://linear.app/n8n/issue/LIGO-270

## Review / Merge checklist

- [ ] Tests included.


packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryPublishedTooltip.vue
```diff
@@ -0,0 +1,91 @@
...
+<template>
+	<N8nTooltip
+		:disabled="!publishInfo"
+		:placement="props.placement"
+		:show-after="300"
+		:offset="props.offset"
+		:content-class="$style.tooltipContent"
+	>
+		<template #content>
+			<div v-if="publishInfo">
```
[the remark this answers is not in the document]
alielkhateeb: Resolved 7e16909a6a120d2cc1d48b7e9b446672f8bff526

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryVersionSelect.vue
```diff
@@ -0,0 +1,190 @@
...
+			groupOptions.push(option);
+		} else {
+			groups.set(key, [option]);
+		}
+	}
+
+	return Array.from(groups.entries()).map(([dateLabel, options]) => ({ dateLabel, options }));
+});
+
+const filteredGroupedOptions = computed(() => {
```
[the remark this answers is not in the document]
alielkhateeb: Might be beneficial but I'll wait for other opinions too, I can extract this into something like this:
```ts
const filteredGroupedOptions = computed(() => {
	if (!filter.value) {
		return groupedOptions.value;
	}

	const normalizedFilter = filter.value.toLowerCase().trim();
	const groupsWithFilteredOptions = groupedOptions.value.map((group) => ({
		...group,
		options: group.options.filter((option) => {
			const optionIsMatching = option.label.toLowerCase().includes(normalizedFilter);
			return optionIsMatching;
		}),
	}));

	return groupsWithFilteredOptions.filter((group) => group.options.length > 0);
});
```

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowVersionStatusIndicator.vue
```diff
@@ -0,0 +1,37 @@
+<script setup lang="ts">
```
sandra0503, line 1: Nit: it took me a while to figure out what this represents, maybe a better naming would be `WorkflowVersionIndicator` or something like that?

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryPublishedTooltip.vue
```diff
@@ -0,0 +1,79 @@
...
+
+defineSlots<{
+	default: () => unknown;
+}>();
+
+const props = withDefaults(
+	defineProps<{
+		label: string;
+		status: WorkflowHistoryVersionStatus;
+		publishInfo?: {
```
konstantintieber: From the component name I expected it to be shown only on published versions.
But reading here that `publishInfo` is an optional prop, I understand that it's on any workflow history entry.
What do you think about calling the component `WorkflowHistoryItemTooltip.vue`?

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryVersionSelect.vue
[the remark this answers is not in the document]
sandra0503: I don't think this is an issue. However there could be small performance gain by moving this into a separate function (currently { ...group } always creates a new object reference). Is group potentially very large?

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryVersionSelect.vue
```diff
@@ -0,0 +1,200 @@
...
+
+.select :global(.el-input--prefix .el-input__inner) {
+	padding-left: 22px;
+}
+
+.workflow-history-version-select-dropdown :global(.el-select-dropdown) {
+	min-width: 320px !important;
+}
+
+.workflow-history-version-select-dropdown :global(.el-select-group__wrap::after) {
```
konstantintieber, line 216: Would be nice if we could get by without these global overrides here.
These selectors will definitely break once we replace the element-plus select with the reka one.

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryUpgradeFooter.vue
```diff
@@ -0,0 +1,49 @@
+<script setup lang="ts">
```
sandra0503, line 1: opinion: feel free to ignore, but for this use case I would personally avoid the overhead of creating/maintaining a new component. Ideally we would have something in design system for that

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
```diff
@@ -23,29 +27,63 @@ const i18n = useI18n();
...
 
 const isLoading = ref(true);
 const sourceWorkflow = ref<IWorkflowDb>();
 const targetWorkflow = ref<IWorkflowDb>();
 const sourceLabel = ref('');
 const targetLabel = ref('');
+const selectedSourceVersionId = ref(props.sourceWorkflowVersionId);
+const selectedTargetVersionId = ref(props.targetWorkflowVersionId);
+const currentWorkflowVersionId = ref<string>();
+const activeWorkflowVersionId = ref<string | undefined>();
```
konstantintieber: As non-author I find myself pondering: what's the difference between `currentWorkflowVersionId` and `activeWorkflowVersionId`. Is `activeWorkflowVersionId` the `publishedWorkflowVersionId`?

I deduce that from `activeWorkflowVersionId.value = workflow.activeVersionId` down below.
The word `active` is the term we used before workflow publishing became a thing in Q4 2025.

When writing new code today, I like to use `published` over `active` for variable/method names.

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
```diff
@@ -23,29 +27,63 @@ const i18n = useI18n();
...
+
+const loadRequestId = ref(0);
```
konstantintieber, line 43: What is the `loadRequestId` needed for, when there is also an `isLoading.value`?
Is there a use case for multiple requests being in flight at the same time?

konstantintieber (commented):
Overall looks good. Just added some comments to better my understanding of the implementation.

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryUpgradeFooter.vue
alielkhateeb: Yes I agree that we should have something like that in the design system, but for now I extracted it to use it in both places (Workflow history list & Workflow Diff select dropdown) without having to maintain it in both places

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryPublishedTooltip.vue
alielkhateeb: Yeah good observation 🚀 
I'll think about either renaming this or changing it to actually require `publishInfo` prop

alielkhateeb: @konstantintieber I refactored to change the prop to required, this made more sense to me 
Benefit I saw in this approach is when there is no publish info, then we won't even render the tooltip, if I kept the prop optional we would be rendering it in "disabled" state 

Really good catch 👏

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
alielkhateeb: @konstantintieber I had the same issue, let me onboard you on my thought process 😂 

The reason I was hesitant about `published` is that multiple versions can have the status "published", but only one version is actually active, which is the latest published workflow version.

While writing this, I think `latestPublishedWorkflowVersionId` would be the most explicit and accurate name 😂 

If you agree, I’ll rename it accordingly. Otherwise, if we intentionally use `publishedWorkflowVersionId` to refer to the **latest** published version, I’m also fine sticking with that .. just point and I'll shoot 😄

konstantintieber: > The reason I was hesitant about published is that multiple versions can have the status "published"

Really? When is it the case that multiple versions are published?
Isn't only the version that's referenced in the `activeVersionId` the published one?

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
alielkhateeb: It's a race-condition guard. Since `loadComparedVersions`'s Promise.all can resolve unreliably the `loadRequestId` guard will prevent cases where a delayed response is received after the response from the most recent request, I hope this is not as badly explained as I feel it is haha

konstantintieber: Sounds like it'd be worth adding a code comment for it, for the next reader

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryListItem.vue
```diff
@@ -163,19 +162,7 @@ onMounted(() => {
...
-				<template v-if="mainTooltipUser">
-					{{ mainTooltipUser }}
-				</template>
-				<template v-if="mainTooltipFormattedDate">
-					<template v-if="mainTooltipUser">, </template>
-					{{ mainTooltipFormattedDate }}
-				</template>
-			</div>
-		</template>
+	<component :is="wrapperProps ? WorkflowHistoryPublishedTooltip : 'span'" v-bind="wrapperProps">
```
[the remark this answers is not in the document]
alielkhateeb: I am amused by these review comments 😂 quite good
9ca65df11ffe09e8387de420fbc0560978746d82

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowHistoryVersionSelect.vue
alielkhateeb: @konstantintieber happy to hear ideas, I couldn't really think of one tbh :( 

I followed what has been done [here](https://github.com/n8n-io/n8n/blob/9ca65df11ffe09e8387de420fbc0560978746d82/packages/frontend/editor-ui/src/features/settings/apiKeys/components/ApiKeyScopes.vue#L162) and I honestly thought the select with dropdown would be an n8n design system component, but having to play around with styles like that didn't feel right.

I changed the main component to use `N8nSelect` instead of `ElSelect` though, this made sense to me
8d4c4774d143d02348c526f18bf46c2df010d0f1

packages/frontend/editor-ui/src/features/workflows/workflowHistory/components/WorkflowVersionStatusIndicator.vue
alielkhateeb: I love this .. `indicator` is a much clearer term, I renamed it to `WorkflowVersionStatusIndicator` here 50031adde6bbe8de691f1ac0554e9b13f23fec0b

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
alielkhateeb: @konstantintieber Yes, in this screenshot you can see 2 published versions, latest is the one with the Green indicator (Another Published Version), but that does not cancel the fact that "With Webhook" is also published - just not latest.
image

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
alielkhateeb: Maybe I am very used to this kind of stale response guards in the UI but that doesn't mean that it's intuitively obvious what this `loadRequestId` does. Added the comment here :) 9da33beb6cbf05dfdfac811de47b93e260f5e9ea

packages/frontend/editor-ui/src/features/workflows/workflowHistory/views/WorkflowHistoryDiff.vue
konstantintieber: I still don't buy it sorry :D what would be the sense in having two published versions?
The version that shows the tooltip in your screenshot looks like it just has `(Published)` in its version name.
But the actual published version is only the one from 11:44:32

There can only be one published version, because that's the version that a productive trigger like webhook or schedule would use when executing a workflow. If there would be two published versions of the same workflow, how would the execution engine decide which one to use?

alielkhateeb: > I still don't buy it sorry :D

That's ok 😄 let's reach the end of the tunnel together 🤝 

> how would the execution engine decide which one to use?

Simply Latest published 😄 

> The version that shows the tooltip in your screenshot looks like it just has (Published) in its version name.

Incorrect, this "(Published)" is dynamically added because `With webhook` has been published before, by me at `11:23:50` this is literally the purpose of the tooltip, to show information about _ALL previously published versions_. 
Every publish creates a new entry in `workflow_publish_history`, so we keep a complete history of all published versions.
Here are both versions `Another Published Version` and `With Webhook` from a joined table between `workflow_history` and `workflow_publish_history` joined on `version ID`

image

You can see above, `With webhook` was activated, then deactivated when `Another Published Version` was "activated" AKA "published" 😄 

Does that paint a clearer picture? Happy to huddle about this to avoid further long and boring comments such as this one (Sorry 😂)

alielkhateeb: Also for clarity from the requirements/designs, [here](https://www.figma.com/design/LWHcX3cJQ1yuC6UvFn6000/%F0%9F%94%83-Version-history?node-id=214-21382&t=5GhhHIdm9dVlWUrL-4) you can see `Version 0f12389 (Published)` with the green indicator, and also `Version 0f12381` which was _previously_ published and we show the publish metadata in the tooltip

konstantintieber: But that means the older version is not actually published, but "used to be published", correct?

Thx for the figma link.
I think displaying `(Published)` on historically published versions in the tooltip here can be confusing for users.
I don't see that specified in the figma designs.

@RobTF9 what do you think as the designer of the publish indicator?

konstantintieber: I just did some testing on my cloud instance running n8n version 2.10.

In the following screenshot, `eafeffb2` was the last published version, and `9f220ae9` before that.
In the list view itself and in the name of the version, there is no `(Published)` text. Only the green indicator is the source of this information, and the "grey empty dot" means "formerly published".
image

When I however hover a formerly published version, then only the tooltip will include the "Publish" term:
image

That behaviour, I can understand.
But automatically adding a `(Published)` text to every version name and tooltip text in the version list, because they used to be published, I now feel the green indicator and this `(Published)` text are competing for the same meaning, even though they don't have the same meaning.
And the most important thing to me is that users understand the "currently published version".
I can see that it can be useful to know "what is the last published version", if you want to revert to that. In the existing UI, the "grey empty dot" is the visual indicator for this. But I feel a `(Published)` text is not implicit enough about this being a "formerly published" version.

konstantintieber: Feedback by @alielkhateeb: actually the "grey empty dot" is displayed on any version that was named, as well as formerly published versions.

konstantintieber: @JanOstrowka chimed in that "formerly published versions" should not include the "(Published)" text.

alielkhateeb: @konstantintieber Action points from the call:
- Will rename to `publishedWorkflowVersionId`
- Will remove "(Published)" from previously published versions, keeping it on the latest published version

sandra0503:
LGTM!

sandra0503 (approved):

PR #26423, opened 2026-03-02
feat(core): Add validation on workflow with dynamic credentials

guillaumejacquart:
## Summary

Adds pre-publication validation for workflows that use dynamic (resolvable) credentials, blocking activation when the required runtime dependencies are not configured.

Two checks are enforced:

  - Resolver check — each dynamic credential must have a resolver configured. The resolver can be
  set either on the credential itself (resolverId) or as a workflow-level fallback
  (workflowSettings.credentialResolverId). If neither is present, activation is blocked with a
  clear error naming the affected credentials.
  - Identity extractor check — at least one trigger node must have context establishment hooks
  configured. These hooks extract user identity from incoming requests (e.g. bearer tokens from
  HTTP headers), which is required for the resolver to determine which credential values to return.
   Triggers without hooks (e.g. schedule, polling) cannot establish this context.

  Both checks run at workflow activation time, before the workflow goes live, preventing silent
  runtime failures when dynamic credentials cannot be resolved.

https://linear.app/n8n/issue/IAM-354/add-validation-on-workflow-with-dynamic-credentials

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -184,6 +190,76 @@ export class WorkflowValidationService {
...
+					credentialIds.add(credData.id);
+				}
+			}
+		}
+
+		if (credentialIds.size === 0) {
+			return { isValid: true };
+		}
+
+		const resolvableCredentials = await this.credentialsRepository.find({
```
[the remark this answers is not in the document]
guillaumejacquart: I think it's ok, because the validation is happening under the hood. I think it's not irrelevant that someone with publish rights to the workflow tries to publishes it, and that triggers an under the hood call to all the workflow credentials (even the one the caller does not have rights to read) just to add the validation business logic

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -661,6 +661,7 @@ export class WorkflowService {
 		await this._detectWebhookConflicts(workflow, versionToActivate);
 
 		this._validateNodes(workflowId, versionToActivate.nodes, versionToActivate.connections);
+		await this._validateDynamicCredentials(workflowId, versionToActivate.nodes, workflow.settings);
```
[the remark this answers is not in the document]
guillaumejacquart: Same as before I think

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -184,6 +190,76 @@ export class WorkflowValidationService {
...
+		const credentialsWithoutResolver = resolvableCredentials.filter((c) => !c.resolverId);
+		if (!workflowResolverId && credentialsWithoutResolver.length > 0) {
+			const credNames = credentialsWithoutResolver.map((c) => `"${c.name}"`).join(', ');
+			return {
+				isValid: false,
+				error: `Cannot publish workflow: dynamic credentials (${credNames}) require a resolver to be configured.`,
+			};
+		}
+
+		const hasExtractorHook = nodes.some((node) => {
```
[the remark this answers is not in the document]
guillaumejacquart: This is ok, it's still a rather small logic inside

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -184,6 +190,76 @@ export class WorkflowValidationService {
...
+			}
+		}
+
+		if (credentialIds.size === 0) {
+			return { isValid: true };
+		}
+
+		const resolvableCredentials = await this.credentialsRepository.find({
+			where: { id: In([...credentialIds]), isResolvable: true },
+			select: ['id', 'name', 'resolverId'],
```
phyllis-noester, line 222: i think i missed this: did we move the resolver to the credential itself?

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -184,6 +190,76 @@ export class WorkflowValidationService {
...
+		});
+
+		if (resolvableCredentials.length === 0) {
+			return { isValid: true };
+		}
+
+		// A credential is covered if it has its own resolver OR the workflow has a defined resolver
+		const workflowResolverId = workflowSettings?.credentialResolverId;
```
phyllis-noester, line 232: i guess we only need to get const credentialsWithoutResolver = resolvableCredentials.filter((c) => !c.resolverId); if no workflow ersolver exists no?

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -184,6 +190,76 @@ export class WorkflowValidationService {
...
+		const credentialsWithoutResolver = resolvableCredentials.filter((c) => !c.resolverId);
+		if (!workflowResolverId && credentialsWithoutResolver.length > 0) {
+			const credNames = credentialsWithoutResolver.map((c) => `"${c.name}"`).join(', ');
+			return {
+				isValid: false,
```
phyllis-noester: i think we should also check for extractor hook before returning so it's easier to debug for the workflow developer as they get all problems back

packages/cli/src/workflows/workflow-validation.service.ts
guillaumejacquart: it has always been there just in case at some point we set a resolver on the credential level. Although not used for now (since it's set on the workflow settings). I'm just making sure here that it's taken into account for future compatibility

packages/cli/src/workflows/workflow-validation.service.ts
guillaumejacquart: Good idea ! I accumulate the errors and send them all at once now. THanks

packages/cli/src/workflows/workflow-validation.service.ts
guillaumejacquart: Good catch, thanks

packages/cli/src/workflows/workflow-validation.service.ts
```diff
@@ -184,6 +190,83 @@ export class WorkflowValidationService {
...
+				}
+			}
+		}
+
+		if (credentialIds.size === 0) {
+			return { isValid: true };
+		}
+
+		const resolvableCredentials = await this.credentialsRepository.find({
+			where: { id: In([...credentialIds]), isResolvable: true },
```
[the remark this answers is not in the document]
guillaumejacquart: already replied on that

afitzek (approved):

PR #26424, opened 2026-03-02
feat(core): Add SSRF protection config

tomi:
## Summary

- Add `SsrfProtectionConfig` to `@n8n/config` and wire it into `GlobalConfig` as `ssrfProtection`.
- Add SSRF env-backed settings, including default blocked ranges, allow lists, DNS cache TTL, and DNS cache size.
- Export `SsrfProtectionConfig` and `SSRF_DEFAULT_BLOCKED_IP_RANGES` from the config package index.
- Add unit tests covering defaults, env parsing, numeric overrides, and fallback behavior for invalid numeric values.

- https://linear.app/n8n/issue/CAT-2383

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/config/src/configs/ssrf-protection.config.ts
+++ b/packages/@n8n/config/src/configs/ssrf-protection.config.ts
@@ -1,3 +1,5 @@
+import { z } from 'zod';
+
 import { CommaSeparatedStringArray } from '../custom-types';
 import { Config, Env } from '../decorators';
 
@@ -28,6 +30,25 @@ export const SSRF_DEFAULT_BLOCKED_IP_RANGES: readonly string[] = Object.freeze([
 	'203.0.113.0/24',
 ]);
 
+/**
+ * Parses comma-separated blocked ranges, expands `default` (case-insensitive)
+ * to the built-in blocked ranges, and removes duplicates while preserving order.
+ */
+const blockedIpRangesSchema = z.string().transform((input) => {
+	const values = input
+		.split(',')
+		.map((value) => value.trim())
+		.filter((value) => value.length > 0);
+
+	const expanded = values.flatMap((value) =>
+		value.toLowerCase() === 'default' ? SSRF_DEFAULT_BLOCKED_IP_RANGES : value,
+	);
+
+	return [...new Set(expanded)];
+});
+
+const positiveNumberSchema = z.coerce.number().gt(0);
+
 @Config
 export class SsrfProtectionConfig {
 	/** Whether SSRF protection is enabled for nodes making HTTP requests to user-controllable targets. */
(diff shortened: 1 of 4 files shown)


packages/@n8n/config/src/configs/ssrf-protection.config.ts
```diff
@@ -0,0 +1,74 @@
...
+	/** Whether SSRF protection is enabled for nodes making HTTP requests to user-controllable targets. */
+	@Env('N8N_SSRF_PROTECTION_ENABLED')
+	enabled: boolean = false;
+
+	/**
+	 * Comma-separated CIDR ranges to block, or 'default' for the standard set.
+	 * Use `resolvedBlockedIpRanges` at runtime instead of reading this directly.
+	 */
+	@Env('N8N_SSRF_BLOCKED_IP_RANGES')
+	blockedIpRanges: string = 'default';
```
dlavrenuek: I imagine that admins would also want the option to add blocked ips without specifying the full default range. For example I am an admin and know that some of the internal pods run on `100.0.0.0`, then it would be good to have the configuration accept `default, 100.0.0.0`.

dlavrenuek: It is also possible to transform the input with zod schema instead of using `resolvedBlockedIpRanges`, this should also avoid errors when someone tries to use `config.blockedIpRanges` instead of `config.resolvedBlockedIpRanges`, an example can be found [here](https://github.com/n8n-io/n8n/blob/ab77c5fc5727beed6d60d45c6190d74c33956072/packages/cli/src/modules/quick-connect/quick-connect.config.ts#L54). Following is untested but should work:

```ts
const blockedIpRangesSchema = z.string().pipe(
	z.preprocess((input: string) => input
      .split(',')
      .map((value) => value.trim()) // <-- may be cast to lower case here to avoid errors with `DEFAULT`
      .flatMap((value) => value === 'default' ? SSRF_DEFAULT_BLOCKED_IP_RANGES : value)
      .filter(Boolean),
    z.array(z.string())),
);

...
	@Env('N8N_SSRF_BLOCKED_IP_RANGES', blockedIpRangesSchema)
	blockedIpRanges: string[] = SSRF_DEFAULT_BLOCKED_IP_RANGES;
```

What do you think?

packages/@n8n/config/src/configs/ssrf-protection.config.ts
```diff
@@ -0,0 +1,74 @@
...
+	@Env('N8N_SSRF_ALLOWED_IP_RANGES')
+	allowedIpRanges: CommaSeparatedStringArray<string> = [];
+
+	/** Comma-separated hostname patterns to allow. Supports wildcards: *.n8n.internal */
+	@Env('N8N_SSRF_ALLOWED_HOSTNAMES')
+	allowedHostnames: CommaSeparatedStringArray<string> = [];
+
+	/** Maximum DNS cache TTL in seconds. Default: 300. */
+	@Env('N8N_SSRF_DNS_CACHE_MAX_TTL_SECONDS')
+	dnsCacheMaxTtlSeconds: number = 300;
```
dlavrenuek: Looking at https://github.com/n8n-io/n8n/pull/26482/changes#diff-43a83e5af0e55ff4cc53ab2b223af375bfccbf46131985187ec92eb310e1a928R50 `ttl` should be > 0 but there is no restriction for the configuration property, so one could configure `N8N_SSRF_DNS_CACHE_MAX_TTL_SECONDS=0`. Should we add a schema to enforce that it's > 0? Something like (untested, not sure if [`coerce`](https://zod.dev/api?id=coercion) is needed or string to number conversion is built in)

```suggestion
	@Env('N8N_SSRF_DNS_CACHE_MAX_TTL_SECONDS', z.coerce.number().gt(0))
	dnsCacheMaxTtlSeconds: number = 300;
```

packages/@n8n/config/src/configs/ssrf-protection.config.ts
tomi: That's a good idea, I'll add it 👍

tomi: Also a good idea 👍 I added it

tomi:
Thank you for the feedback @dlavrenuek 🫶 Addressed the comments

dlavrenuek (approved):
Thank you for addressing my comments and the clean implementation!

PR #26430, opened 2026-03-02
fix(core): Don't include last human message twice on Chat hub message regeneration

Cadiac:
## Summary

When user regenerated messages on Chat hub we accidentally included the last human message both oh the chat history and as a new human message triggering that turn.

Instead don't include the human message on the history, as it gets replayed which inserts it into the memory again.

https://linear.app/n8n/issue/CHA-83/bug-duplicate-user-message-sent-on-regeneration

## Review / Merge checklist

- [ ] Tests included. 


autologie (approved):
Thank you

PR #26435, opened 2026-03-02
feat(editor): Make it possible to pick Chat hub model by ID even if no models are loaded

Cadiac:
## Summary

On providers such as Azure when we can't list available models we should still make it possible for users to choose the model to use manually by entering the ID of the model. Our logic was faulty and hid the option even if user had defined credentials if entering those credentials didn't lead to the list being populated.

Now users can configure their Azure credentials, and then manually enter the identifier of the model to use (ie. `gpt-5-mini`) using this menu option and use it.

image

This still isn't the ideal solution as these custom options aren't saved on the backend. For these providers admin can go to Chat settings, find the provider and edit it, adding the desired models manually under "Limit Models" to make them show up for all users of the instance.

image

cha-123-bug-azure-ai-foundry-microsoft-foundry-doesnt-work-on-chat

## Review / Merge checklist

- [ ] Tests included. 


autologie (approved):
👍

PR #26470, opened 2026-03-03
feat(core): Support data tables in workflow indexer (no-changelog)

CharlieKolb:
## Summary

Also track data table dependencies. Precursor for https://github.com/n8n-io/n8n/tree/dt_dependencies

https://linear.app/n8n/issue/ADO-4917

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/workflow-index/__tests__/workflow-index.service.test.ts
```diff
@@ -388,6 +388,232 @@ describe('WorkflowIndexService', () => {
...
+					}),
+					expect.objectContaining({
+						dependencyKey: 'table-4',
+						dependencyInfo: expect.objectContaining({ nodeId: 'node-4' }),
+					}),
+				]),
+			);
+		});
+
+		it('should skip dataTableId with expression-based values', async () => {
```
mfsiega: A thought I've had: it would be cool to also update the dependencies at runtime, so you could identify these (as long as you see them, of course).

packages/cli/src/modules/workflow-index/__tests__/workflow-index.service.test.ts
```diff
@@ -388,6 +388,232 @@ describe('WorkflowIndexService', () => {
...
+							dependencyType: 'dataTableId',
+							dependencyKey: 'My Table',
+							dependencyInfo: { nodeId: 'node-1', nodeVersion: 1, mode: 'name' },
+						}),
+					]),
+				}),
+			);
+		});
+
+		it('should extract dataTableId from all data-table-related node types', async () => {
```
mfsiega, line 391: Nit: It slightly feels like this test could replace the three preceding tests, and those others are a bit redundant?

packages/cli/src/modules/workflow-index/workflow-index.service.ts
```diff
@@ -244,6 +245,34 @@ export class WorkflowIndexService {
 		}
 	}
 
+	private static readonly DATA_TABLE_NODE_TYPES = new Set([
```
mfsiega: Out of curiosity - elsewhere we've used something like a simpler const for this sort of pattern. Is there an advantage here?

mfsiega (commented):
A few small comments, haven't finished reviewing the code carefully yet but will finish later today!

packages/cli/src/modules/workflow-index/workflow-index.service.ts
CharlieKolb: Fair point, this could be useful elsewhere also - I was surprised we didn't have a global value here already actually

packages/cli/src/modules/workflow-index/__tests__/workflow-index.service.test.ts
CharlieKolb: Yes, same with sub-workflows using expressions to call different sub-workflows (You can implement "interfaces" where users can bring their own custom sub-workflows). But that's out of scope here I think 😅

mfsiega (approved):
LGTM 👏

PR #26477, opened 2026-03-03
fix: Devcontainer build failures from hardened base image (no-changelog)

shortstacked:
## Summary
- Switch devcontainer base from `n8nio/base:24` to `node:24-alpine` — the production base image removes `apk-tools`, making package installation impossible
- Replace `npm install -g pnpm` with `corepack enable` — pnpm version is managed by `packageManager` in `package.json` via the existing `postCreateCommand`
- Add `libc-utils` for `getconf`, required by VS Code server's `check-requirements.sh`
- Add all OS dependencies from the base image (`openssl`, `tini`, `tzdata`, `ca-certificates`, etc.) to maintain parity

## Test plan
- [x] `docker build -f .devcontainer/Dockerfile .devcontainer/` builds successfully
- [ ] Open devcontainer via VS Code and verify it starts without errors
- [ ] Run `pnpm install` and `pnpm build` inside the container

Related: https://linear.app/n8n/issue/CAT-2506

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,9 +1,15 @@
-FROM n8nio/base:24
+ARG NODE_VERSION=24
 
-RUN apk add --no-cache --update openssh sudo shadow bash
+FROM node:${NODE_VERSION}-alpine
+
+ARG NODE_VERSION
+
+RUN apk add --no-cache \
+    openssh sudo shadow bash libc-utils \
+    git openssl graphicsmagick tini tzdata ca-certificates libc6-compat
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,9 +1,15 @@
-FROM n8nio/base:24
```
shortstacked, line 1: We can let the dev container use a generic image, there's no need for it to go through a hardened image.

shortstacked:
@claude review please

despairblue (approved):

PR #26482, opened 2026-03-03
feat(core): Add in-memory DNS cache with TTL validation (no-changelog)

tomi:
## Summary

- Adds an `InMemoryDnsCache` service in `packages/cli` using `cache-manager` memory storage with byte-size accounting, LRU eviction, and per-entry TTL capped by SSRF config.
- Enforces `ttl > 0` in `set()` to prevent caching DNS entries with non-positive TTLs.
- Adds unit tests for cache misses, set/get behavior, TTL expiration, max-TTL capping, TTL validation errors, LRU eviction, and clear/reset behavior.

https://linear.app/n8n/issue/CAT-2384

## Review / Merge checklist

- [x] Tests included.


dlavrenuek (dismissed):
Very clean implementation. Great that the cache max size is also considered to prevent DDOS by cache size exploding in memory 🚀

packages/cli/src/services/ssrf/ssrf-blocked-ip.error.ts
```diff
@@ -0,0 +1,17 @@
+import { UserError } from 'n8n-workflow';
+
+/**
+ * Error thrown when a resolved address is denied by SSRF IP policy.
+ */
+export class SsrfBlockedIpError extends UserError {
+	readonly ip: string;
+	readonly hostname?: string;
+
+	constructor(ip: string, hostname?: string) {
```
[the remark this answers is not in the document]
dlavrenuek: That looks like a valid issue @tomi. Do we even need to have following in this class at all?

```ts
	readonly ip: string;
	readonly hostname?: string;
```

packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,202 @@
...
+		return async (hostname, options, onResult) => {
+			try {
+				const resolved = await this.secureLookupAsync(hostname, options);
+
+				if (options.all) {
+					onResult(null, resolved);
+					return;
+				}
+
+				const first = resolved[0];
```
dlavrenuek, line 145: I assume that there will always be at least one IP resolved or an error thrown before

tomi: Yes:

```ts
> dns.lookup("asjdkasjhdjkashdkjsa.com", (err, record) => { console.log(err, record) })
GetAddrInfoReqWrap {
  callback: [Function (anonymous)],
  family: 0,
  hostname: 'asjdkasjhdjkashdkjsa.com',
  oncomplete: [Function: onlookup]
}
> Error: getaddrinfo ENOTFOUND asjdkasjhdjkashdkjsa.com
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:111:26)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'asjdkasjhdjkashdkjsa.com'
} undefined
```

packages/cli/src/services/ssrf/ssrf-blocked-ip.error.ts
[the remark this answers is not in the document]
tomi: Thanks, fixed now

packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,202 @@
...
+		const { hostname } = parsed;
+
+		if (this.allowedHostnameMatcher.matches(hostname)) {
+			return { allowed: true };
+		}
+
+		const cleanIp = this.normalizeIpInHostname(hostname);
+		if (isIP(cleanIp)) {
+			return this.validateAddress(cleanIp);
+		}
```
dlavrenuek, line 86: There are two different validation implementations, one in `validateUrl` and another one in `secureLookupAsync`, this makes it hard to follow/compare if the logic is the same and brings more potential for errors when the logic is changed/extended. I suggest reusing `secureLookupAsync()` inside here.

Just from the code it looks like the logic here will be different `secureLookupAsync` if the ip is on the blocklist and allowed list when used directly in the url like `http://10.0.0.1/some-endpoint`. But then `secureLookupAsync` should only get one ip from `this.dnsResolver.lookup()`, so the result should be the same. The result will only be different if lookup can return multiple ips for a given ip (not hostname) where only one of them is allowed, but I'm not sure if that's even possible

packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,202 @@
...
+		if (ips.length === 0) {
+			return { allowed: false, reason: 'DNS resolution failed', hostname };
+		}
+
+		for (const ip of ips) {
+			const result = this.validateAddress(ip.address);
+			if (!result.allowed) {
+				return result;
+			}
+		}
```
dlavrenuek, line 99: Given a scenario where we use the default blocklist but add `10.0.0.1` to allowed ips because we have some database there that is allowed to be used, the hostname will be f.e. `database.k8s`. If `database.k8s` resolves to `10.0.0.1` but also to an ipv6 which is on the blocklist, then it will be blocked while it should have passed the validation, right? May be we need to distinguish between "ip is not configured" - resulting in true, "ip is explicitly blocked" - overrides ip is not configured resulting in false, "ip is explicitly allowed" - overrides ip is blocked resulting in true. What do you think?

dlavrenuek (changes requested):
The different networking level blocking logic really requires some heavy brain energy. Some questions and suggestions disguised as questions from my side.

packages/cli/src/services/ssrf/ssrf-protection.service.ts
tomi: I think for a scenario like this you would want to allowlist the hostname `database.k8s`. In that case it should always be allowed no matter what the IP is. But good point about providing clearer feedback about why it fails. I'll clarify that

dlavrenuek (approved):
🚀

PR #26488, opened 2026-03-03
chore(core): Integrate ExecutionRedactionService into all execution data endpoints

afitzek:
## Summary

Extends the `ExecutionRedactionService` integration (previously only applied to `GET /executions/:id` and `PATCH /executions/:id`) to cover all endpoints that return full execution node output data.

**New endpoints covered:**
- `GET /workflows/:workflowId/executions/last-successful` — applies policy-based redaction; supports the `redactExecutionData` query parameter for explicit redact/reveal
- `POST /executions/:id/retry` (internal API) — applies redaction to the retry response before returning, respecting the `redactExecutionData` query parameter
- `GET /api/v1/executions/:id` (public API) — applies policy-based redaction when `includeData=true`
- `GET /api/v1/executions` (public API) — applies policy-based redaction to each execution in the list when `includeData=true`
- `POST /api/v1/executions/:id/retry` (public API) — inherits redaction automatically via `ExecutionService.retry()`

**Type improvements:** Introduced a `RedactableExecution` structural type that captures the minimal shape the redaction service actually needs (`mode`, `workflowId`, `data`, `workflowData.settings`). This replaces the previous `IExecutionDb` coupling and allows callers with `IExecutionResponse` or other compatible types to pass executions without unsafe casts.

> **Breaking change (opt-in):** When `N8N_ENV_FEAT_EXECUTION_REDACTION=true` is set and the workflow's redaction policy is not `none`, the public API endpoints will now return redacted node output data consistent with the internal API. This is the intended behavior for operators who have enabled the feature flag.

https://linear.app/n8n/issue/IAM-187

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
+++ b/packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
@@ -1,4 +1,4 @@
-import type { IExecutionDb } from '@n8n/db';
+import type { IExecutionBase } from '@n8n/db';
 import { ExecutionRepository } from '@n8n/db';
 import { Container } from '@n8n/di';
 import type express from 'express';
@@ -14,8 +14,15 @@ import { QueuedExecutionRetryError } from '@/errors/queued-execution-retry.error
 import { NotFoundError } from '@/errors/response-errors/not-found.error';
 import { EventService } from '@/events/event.service';
 import { ExecutionPersistence } from '@/executions/execution-persistence';
+import type { RedactableExecution } from '@/executions/execution-redaction';
 import { ExecutionRedactionServiceProxy } from '@/executions/execution-redaction-proxy.service';
 import { ExecutionService } from '@/executions/execution.service';
+
+function isRedactableExecution(
+	execution: IExecutionBase,
+): execution is IExecutionBase & RedactableExecution {
+	return 'data' in execution && 'workflowData' in execution;
+}
 
 import type { ExecutionRequest } from '../../../types';
 import { apiKeyHasScope, validCursor } from '../../shared/middlewares/global.middleware';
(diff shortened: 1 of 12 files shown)


packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
```diff
@@ -148,6 +163,18 @@ export = {
 			const count =
 				await Container.get(ExecutionRepository).getExecutionsCountForPublicApi(filters);
 
+			if (includeData) {
+				const redactionProxy = Container.get(ExecutionRedactionServiceProxy);
+				await Promise.all(
```
BGZStephen: `processExecution` doesn't seem to currently be `async`, have we made the interface expect a Promise because of future expectations? Otherwise might be worth removing the async bits.

BGZStephen (dismissed):

packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
afitzek: I think `processExecution` is `async`, the proxy function is async, because the implementation function is also async. The reason for this is, that we want to be able to determine if the user can reveal execution data, which requires a db query. Maybe I'm missing something?

BGZStephen: Ah yes, ignore me, I didn't see the `|| (await`

packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
```diff
@@ -148,6 +163,18 @@ export = {
...
+					executions.filter(isRedactableExecution).map(
+						async (execution) =>
+							await redactionProxy.processExecution(execution, {
```
guillaumejacquart: I think there might be performance impact on this one, no ? Since it's calling canUserReveal and thus WorkflowFinderService.findWorkflowForUser under the hood. 
Would it be worth having some kind of multiple executions redaction process, that would for instance get all the user reveal permissions in a single query ?

guillaumejacquart (commented):
Nice work ! I like the new RedactableExecution type :)
1 comment

guillaumejacquart (approved):
nice work, thanks !

PR #26489, opened 2026-03-03
feat: Add option to hide credential overwrites

Joffcom:
## Summary
There are times where we might want to deprecate a managed credential on n8n cloud, To handle this and make the transition easier we are adding a new env option that accepts a credential type, If this type is in the overwrites we will keep existing credentials working but won't allow new ones to be created.

## Review / Merge checklist

- [x] Tests included. 


elsmr (approved):
Tested, all works as expected

PR #26490, opened 2026-03-03
test: Add contract test for n8n node vs public API coverage

shortstacked:
## Summary
- Adds a manifest-based contract test that ensures the n8n node (`packages/nodes-base/nodes/N8n/`) stays in sync with the public API OpenAPI spec (`packages/cli/src/public-api/v1/`)
- When a new endpoint is added to the spec, CI fails until the endpoint is classified in `n8n-api-coverage.json` as `covered`, `gap`, or `excluded`
- Also catches stale manifest entries when endpoints are removed from the spec

## Context

Currently 15 of 59 API endpoints are covered by the node. The remaining 44 are baselined as `gap` so the test passes from day one while catching all future drift.

## Test plan
- [x] All 3 test cases pass: tracked endpoints, stale detection, excluded reason validation
- [x] Verified failure output when an endpoint is missing from the manifest
- [x] Verified failure output when a stale entry exists in the manifest
- [ ] CI passes


shortstacked:
@claude review please

heymynameisrob (approved):

PR #26492, opened 2026-03-03
fix: Log streaming config works with proxy

phyllis-noester:
## Summary

https://www.loom.com/share/d2c4a158170e44638b957d967f93fcbf

https://linear.app/n8n/issue/IAM-61/log-streaming-proxy-config-not-working

## Review / Merge checklist

- [x] Tests included. 


packages/workflow/src/message-event-bus.ts
```diff
@@ -90,9 +90,12 @@ const webhookParameterOptionsSchema = z
 		queryParameterArrays: z.enum(['indices', 'brackets', 'repeat']).optional(),
 		redirect: z
 			.object({
-				followRedirects: z.boolean().optional(),
-				maxRedirects: z.number().int().positive().optional(),
+				redirect: z.object({
```
afitzek, line 93: Do we have the same issue for redirects?

afitzek (approved):

packages/workflow/src/message-event-bus.ts
phyllis-noester: yea, from what i can see redirect is also defined as fixedcollection in line with the proxy but that was not reflected in zod. I tested it (that it;s configurable now) and it saved successfully

PR #26493, opened 2026-03-03
test: Add external secrets provider connection tests

sandra0503:
## Summary

Covers two flows: creating a global connection and referencing it in a credential with the delete/confirmation flow, and creating a global connection scoped to a project via the sharing tab then using it in a project credential.

Run (with docker built):
```
pnpm --filter=n8n-playwright test:container:sqlite --grep "Secret Providers Connections"
```

## Recording:
https://www.loom.com/share/1b354c8a92c840efb3a9d562702cc4f6

Closes [LIGO-204](https://linear.app/n8n/issue/LIGO-204/add-e2e-tests-for-external-secret-providers)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
+++ b/packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
@@ -1,0 +1,184 @@
+import { expect, test } from '../../../../fixtures/base';
+
+test.use({ capability: 'external-secrets' });
+
+// LocalStack can take time to start up, and connection tests add latency
+test.setTimeout(180_000);
+
+// LocalStack test credentials (AWS_ENDPOINT_URL in the n8n container redirects all SDK calls to LocalStack)
+const LOCALSTACK_AWS_SETTINGS = {
+	region: 'us-east-1',
+	authMethod: 'iamUser',
+	accessKeyId: 'test',
+	secretAccessKey: 'test',
+} as const;
+
+test.describe(
+	'Secret Providers Connections UI @capability:external-secrets @licensed',
+	{
+		annotation: [{ type: 'owner', description: 'Lifecycle & Governance' }],
+	},
+	() => {
+		test.beforeEach(async ({ n8n, services }) => {
+			await n8n.api.enableFeature('externalSecrets');
+			await services.localstack.secretsManager.clear();
+		});
+
+		/**
+		 * Creates a global connection via the UI settings page, verifies the success
+		 * callout and global badge, creates a credential referencing a secret from
+		 * the provider, then deletes the connection through the UI delete flow.
+		 */
+		test('should create a global connection via UI and reference it in a credential', async ({
+			n8n,
+			services,
+		}) => {
+			const providerName = 'awsGlobal';
+
+			await services.localstack.secretsManager.createSecret('apikey', 'supersecret');
+
+			await n8n.navigate.toExternalSecrets();
+			await n8n.secretsProviderSettings.waitForProvidersLoaded();
+			await expect(n8n.secretsProviderSettings.getEmptyState()).toBeVisible();
+
+			await n8n.secretsProviderSettings.clickEmptyStateAddButton();
+			await n8n.secretsProviderConnectionModal.waitForModal();
+
+			await n8n.secretsProviderConnectionModal.fillConnectionName(providerName);
+			await n8n.secretsProviderConnectionModal.selectProviderType('AWS Secrets Manager');
+			await n8n.secretsProviderConnectionModal.fillProviderField(
+				'region',
+				LOCALSTACK_AWS_SETTINGS.region,
+			);
+			await n8n.secretsProviderConnectionModal.fillProviderField(
+				'accessKeyId',
+				LOCALSTACK_AWS_SETTINGS.accessKeyId,
+			);
+			await n8n.secretsProviderConnectionModal.fillProviderField(
+				'secretAccessKey',
+				LOCALSTACK_AWS_SETTINGS.secretAccessKey,
+			);
+
+			await n8n.secretsProviderConnectionModal.save();
+			await expect(n8n.secretsProviderConnectionModal.getSuccessCallout()).toBeVisible();
+			await n8n.secretsProviderConnectionModal.close();
+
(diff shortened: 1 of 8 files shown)


packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+// LocalStack can take time to start up, and connection tests add latency
+test.setTimeout(180_000);
+
+// LocalStack test credentials (AWS_ENDPOINT_URL in the n8n container redirects all SDK calls to LocalStack)
+const LOCALSTACK_AWS_SETTINGS = {
+	region: 'us-east-1',
+	authMethod: 'iamUser',
+	accessKeyId: 'test',
+	secretAccessKey: 'test',
+} as const;
```
geemanjs, line 14: Should these values be coming from `services.localstack` or are they always the same? or does it just accept any credential?

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+			const credentialName = 'Secret Ref Header Auth';
+			await n8n.navigate.toCredentials();
+			await n8n.credentials.emptyListCreateCredentialButton.click();
+			await n8n.credentials.selectCredentialType('Header Auth');
+			await n8n.credentials.credentialModal.waitForModal();
+			await n8n.credentials.credentialModal.renameCredential(credentialName);
+			// plain text, not expression mode
+			await n8n.credentials.credentialModal.fillField('name', 'Authorization');
+			// expression mode, referencing the external secret
```
geemanjs, line 81: Should we check here that the credential modal has the options for our new provider? Feels like it could be a "freebie"

geemanjs: To do this we might need to force a "Reload secrets" before going here

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+			await n8n.credentials.credentialModal.close();
+
+			await expect(n8n.credentials.cards.getCredential(credentialName)).toBeVisible();
+
+			await n8n.navigate.toExternalSecrets();
+			await n8n.secretsProviderSettings.waitForProvidersLoaded();
+			await n8n.secretsProviderSettings.selectCardAction(providerName, 'Delete');
+			await n8n.deleteSecretsProviderModal.waitForModal();
+
+			// The credential references this provider, so the user must type the exact provider
```
geemanjs, line 99: Am I right in understanding that because they have used it in a credential they have to fill out the name?

geemanjs (commented):
I know we descoped it from the ticket but I feel there is a lot of value in a test that "misconfigures" a provider so we can validate the error state

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
sandra0503: Yes, exactly. It's additional check to prevent unwanted deletions

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
sandra0503: Which options do you mean exactly?

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
geemanjs: This bit 😄 
https://github.com/n8n-io/n8n/blob/master/packages/testing/containers/services/localstack.ts#L87-L94

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
geemanjs: The "expression" autocomplete has the "secert provider name" and then the "secret key" available

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
sandra0503: This is a copy from the values in services.localstack. I tested and it accepts any credential.
I agree it would be great if we could validate these and test error states. Do you know what's missing for that in localstack set-up?

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
sandra0503: Ah, got you!

sandra0503: I would prefer to keep the autocomplete out of this test. The critical path is that a project-scoped secret can be referenced in a credential. Testing the autocompletion pop-up could make the test more fragile (popup coming up and disappearing again).

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
sandra0503: It was a bit tricky to enforce IAM credential check in localstack, but I found a way to trigger the error state by using a wrong region - let me know what you think

geemanjs (approved):

packages/testing/playwright/tests/e2e/settings/external-secrets/secret-providers-connections-ui.spec.ts
geemanjs: Awesome - Nice work 😄

sandra0503:
Thanks for reviewing :)

PR #26495, opened 2026-03-03
fix(editor): Add data to a data table by csv upload

nikhilkuria:
## Summary

- Adds ability to import CSV files into existing data tables via a new modal (Actions > Import CSV)
- CSV column names must exactly match table column names; unrecognized columns block import
- System columns (id, createdAt, updatedAt) in CSV are silently ignored
- Empty CSV values are converted to null to avoid type validation issues with datetime/boolean columns
- 

https://linear.app/n8n/issue/ADO-4896/feature-add-data-to-a-data-table-by-csv-upload

## Review / Merge checklist

- [ ] Tests included. 


RicardoE105 (commented):
## Review

Nice feature! The backend service extraction is clean and the CSV parser refactoring is solid. A few things I noticed:

### PR Title
The title uses `fix(editor)` but this is a new feature — should be `feat(editor): Add data to a data table by CSV upload`.

### Permission check mismatch (DataTableActions.vue)

The frontend disables the action based on `dataTable.update`:
```typescript
disabled: !dataTableStore.projectPermissions.dataTable.update || props.isReadOnly,
```
But the backend endpoint uses `@ProjectScope('dataTable:writeRow')`. These should be consistent — the frontend should check `dataTable.writeRow`.

### Missing error handling after refactor (data-table.service.ts)

After extracting logic to `DataTableCsvImportService`, the `try/catch` in `importCsvToExistingTable` was removed and only `finally` remains. If `validateAndBuildRowsForExistingTable` throws a `FileUploadError`, it bubbles up unwrapped. The controller only handles `DataTableValidationError` and `DataTableNotFoundError` — a `FileUploadError` would fall through to the generic `InternalServerError` catch. Consider whether you want to handle `FileUploadError` explicitly in the controller or re-add a catch block in the service.

### No client-side file size validation

The modal doesn't check file size before uploading. Users could attempt uploading very large files. Consider adding a client-side guard.

### Minor nits
- The `!important` on `input[type='file'] { display: none !important; }` is unnecessary since CSS modules already scope this.
- `isUploadHovered` for icon color could be done with pure CSS `:hover` instead of JS event handlers.
- Missing `data-test-id` on the upload drop zone — will make E2E testing harder.
- Frontend test coverage is minimal — only tests rendering, disabled state, and cancel. Missing tests for column matching/mismatch callout display, error states, and import success flow.

### Memory concern (not blocking)

`parseFileWithData` loads the entire CSV into memory. This is the same pattern as `parseFileData`, but since import-to-existing is user-facing and may handle larger files, streaming/batching could be a good follow-up.

RicardoE105 (commented):
left a few comments

packages/cli/src/modules/data-table/csv-parser.service.ts
```diff
@@ -42,6 +42,59 @@ export class CsvParserService {
 
 	private readonly TYPE_INFERENCE_SAMPLE_SIZE = 100;
 
+	private createParserOptions(hasHeaders: boolean, onColumnNames: (names: string[]) => void) {
+		return {
+			columns: hasHeaders
+				? (header: string[]) => {
+						const trimmed = header.map((h) => h.trim());
```
scdekov: Should we just do that in the parent function and pass the trimmed headers to `createParserOptions`?

packages/cli/src/modules/data-table/csv-parser.service.ts
```diff
@@ -53,59 +106,24 @@ export class CsvParserService {
...
+			const parser = parse(
+				this.createParserOptions(hasHeaders, (names) => {
+					columnNames = names;
+				}),
+			)
 				.on('data', (row: Record<string, string> | string[]) => {
 					rowCount++;
+					const normalized = this.normalizeRow(row, hasHeaders, columnNames);
+					if (!normalized) return;
+					columnNames = normalized.columnNames;
```
scdekov: Seems a little odd to update the column names after each row.

packages/cli/src/modules/data-table/data-table-csv-import.service.ts
```diff
@@ -0,0 +1,168 @@
...
+	private async buildColumnMappingForNewTable(
+		fileId: string,
+		hasHeaders: boolean,
+		tableColumns: DataTableColumn[],
+		dtoColumns?: CreateDataTableDto['columns'],
+	): Promise<Map<string, string>> {
+		const columnMapping = new Map<string, string>();
+
+		const hasCsvColumnNames = dtoColumns?.some((c) => c.csvColumnName);
+		if (hasCsvColumnNames && dtoColumns) {
```
scdekov: ` && dtoColumns` is redundant

packages/cli/src/modules/data-table/data-table.service.ts
```diff
@@ -79,14 +76,19 @@ export class DataTableService {
...
+					tableColumns,
 					dto.columns,
 				);
-				await this.fileCleanupService.deleteFile(dto.fileId);
+
+				if (rows.length > 0) {
+					await this.insertRows(result.id, projectId, rows);
+				}
+
+				await this.csvImportService.cleanupFile(dto.fileId);
```
scdekov: Should this be in a finally block?

scdekov (dismissed):
Works very well 🚀

packages/cli/src/modules/data-table/csv-parser.service.ts
nikhilkuria: Good point. Doing some light refactors to avoid this.

RicardoE105 (approved):

PR #26502, opened 2026-03-04
fix(core): Fix cache get for Redis missing key placeholders and set for falsy values

alielkhateeb:
## Summary

Fixes two bugs in the cache service:

1. **`set` ignores falsy values** — the previous guard `!value` would skip caching `0`, `false`, `""`, etc. Changed to only skip `undefined` and `null`.
2. **`get` returns `[]` as a cache hit after Redis restart** — after a cache loss (e.g. Redis restart), some adapters return `[]` for missing keys. Added `isValueAMissingKeyPlaceholder` to detect this case by checking the key's TTL (`-2` means key does not exist in Redis), ensuring a proper cache miss is returned instead.

Clean up - Remove `getMany`:

1. `getMany` in `cacheService` was unused, only in tests. This is removed here as well as `malformedRefreshValueError` which consequently because unused anywhere too.

> [!NOTE]
> I genuinely believe this will fix the issue, but I want to test it
> I was not able to reproduce the issue because I must be doing something wrong in running queue mode, I followed the steps and I have n8n running in queue mode (Docker), but I can't figure out what am I missing, help doing so is greatly appreciated

https://linear.app/n8n/issue/LIGO-300

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/cli/src/services/cache/cache.service.ts
+++ b/packages/cli/src/services/cache/cache.service.ts
@@ -6,6 +6,7 @@ import { jsonStringify, UserError } from 'n8n-workflow';
 
 import { MalformedRefreshValueError } from '@/errors/cache-errors/malformed-refresh-value.error';
 import { UncacheableValueError } from '@/errors/cache-errors/uncacheable-value.error';
+import { REDIS_TTL_KEY_MISSING } from '@/services/cache/cache.constants';
 import type {
 	TaggedRedisCache,
 	TaggedMemoryCache,
@@ -102,7 +103,7 @@ export class CacheService extends TypedEmitter<CacheEvents> {
 	async set(key: string, value: unknown, ttl?: number) {
 		if (!this.cache) await this.init();
 
-		if (!key || !value) return;
+		if (!key || value === undefined || value === null) return;
 
 		if (this.cache.kind === 'redis' && !this.cache.store.isCacheable(value)) {
 			throw new UncacheableValueError(key);
@@ -189,7 +190,9 @@ export class CacheService extends TypedEmitter<CacheEvents> {
 
 		const value = await this.cache.store.get<T>(key);
 
-		if (value !== undefined) {
+		const hasValue = value !== undefined;
+		const cacheHit = hasValue && !(await this.isValueAMissingKeyPlaceholder(key, value));
+		if (cacheHit) {
 			this.emit('metrics.cache.hit');
 
 			return value;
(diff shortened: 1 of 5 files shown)


packages/cli/src/services/cache/cache.service.ts
```diff
@@ -102,7 +103,7 @@ export class CacheService extends TypedEmitter<CacheEvents> {
 	async set(key: string, value: unknown, ttl?: number) {
 		if (!this.cache) await this.init();
 
-		if (!key || !value) return;
+		if (!key || value === undefined || value === null) return;
```
alielkhateeb, line 105: This is a bug I noticed in the code, `!value` will evaluate to `true` on falsy values such as `0`, `''` and `false`, which is incorrect if the cached value would be one of these.

MarcL (commented):
Review with no comments for testing Linear.

packages/cli/src/services/cache/cache.service.ts
```diff
@@ -102,7 +102,7 @@ export class CacheService extends TypedEmitter<CacheEvents> {
...
```
geemanjs, line 105: Should we be doing the same check with key?

packages/cli/src/services/cache/cache.service.ts
```diff
@@ -189,7 +189,9 @@ export class CacheService extends TypedEmitter<CacheEvents> {
 
 		const value = await this.cache.store.get<T>(key);
 
-		if (value !== undefined) {
+		const hasValue = value !== undefined;
```
geemanjs, line 192: Should this also be `value !== undefined && value !== null`

geemanjs (approved):

packages/cli/src/services/cache/cache.service.ts
alielkhateeb: @geemanjs I would really doubt that there would be a scenario where we want to set the key to `false`, `0` or `''`. 
But I see your point that it still might happen, I just can't see it happening 😄

alielkhateeb: Also the `key` is controlled by us, unlike the value which might come from the consumer/user

packages/cli/src/services/cache/cache.service.ts
alielkhateeb: That is unnecessary, the return type of `cache.store.get` is `Promise<T | undefined>`, so we only need to guard it against `undefined` to ensure we have a value

PR #26511, opened 2026-03-04
fix(Extract from File Node): Skip empty lines in CSV parsing to prevent errors

elsmr:
## Summary

Fixes the Extract From CSV node failing when a CSV file has trailing empty lines. This regression was introduced in v1.1 of the node (n8n 1.120.0) when `failOnCsvBufferError` was added.

**Two fixes:**

1. **Added `skip_empty_lines: true`** to the `csv-parse` options: empty lines are silently skipped instead of generating "Invalid Record Length" errors
2. **Fixed promise handling**: replaced `readable` event with `end` event. The `readable` event fires on the first available chunk (not when parsing completes), which could miss trailing errors. Added `parser.resume()` to drain the Transform stream's readable side so `end` fires properly.

Test workflow in Linear

https://linear.app/n8n/issue/NODE-4528

## Review / Merge checklist

- [x] Tests included.


michael-radency (approved):
LGTM

PR #26516, opened 2026-03-04
feat(core): Agent file knowledge in ChatHub (no-changelog)

autologie:
## Summary

This PR adds the new feature in ChatHub personal agent to answer questions based on PDF files that the user has shared. The feature is built on RAG pattern, allowing AI models to make semantic search on the documents in vector store.

### Setup

To make this feature available in an n8n instance, owner/admin must choose the vector store and embedding model provider in Chat settings page and setup credential for them.

Screenshot 2026-03-04 at 11 01 41

At the moment, three vector store providers are supported:
- PGVector (PostgreSQL with pgvector extension); also works for Supabase
- Qdrant
- Pinecone

### Usage

Once the setup is complete, users in the n8n instance can add files to their personal agent. Then, simply talk to the agent - it will generate response based on contents of the files.

Screenshot 2026-03-04 at 11 10 40

### Limitations

Only text PDF file is supported for now. Pictures, diagrams in PDF are not recognized.

### Technical design

Semantic search is implemented by adding a vector store tool to the agent in chat reply workflow. To ensure users cannot access documents that other users have uploaded, new purpose-built vector store nodes and corresponding credentials are introduced.

| Provider | Namespace mapping |
|--------|--------|
| **PGVector** | Table name has the format `{prefix}_{userId}` where prefix is a fixed value set in the credential and users cannot modify |
| **Pinecone** | Namespace has the format `{prefix}_{userId}`, similar to PGVector |
| **Qdrant** | All documents for all users are indexed in the same collection specified in the credential, and access control is implemented with metadata filter. (Considered the approach of collection per user, but decided against it because Qdrant cloud has the upper limit to collections: 1000)  |

User ID is taken from execution context and throws exception if it's missing (e.g. manual execution). These credentials appear in the credential list but don't in "create new credential" flow. Nodes are hidden from node creator.

### Data lifecycle

Uploaded PDF files belong to the workflow execution for indexing documents, and deleted once completed. Documents in vector store are deleted through
- Targeted deletion in updating agent
- On deleting agent
- On deleting the user who owns the agent

https://linear.app/n8n/issue/CHA-100/feature-file-knowledge

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/ai-utilities/src/utils/vector-store/createVectorStoreNode/types.ts
```diff
@@ -48,13 +49,17 @@ export interface VectorStoreNodeConstructorArgs<T extends VectorStore = VectorSt
 				payload: IDataObject | string | undefined,
 			) => Promise<NodeParameterValueType>;
 		};
+		credentialTest?: {
```
autologie, line 52: To implement the extra check for PGVector to check if pgvector extension is enabled

packages/@n8n/api-types/src/chat-hub.ts
```diff
@@ -493,8 +502,8 @@ export class ChatHubCreateAgentRequest extends Z.class({
 	name: z.string().min(1).max(128),
 	description: z.string().max(512).optional(),
 	icon: agentIconOrEmojiSchema,
+	systemPrompt: z.string().default(''),
```
autologie, line 505: Making optional because tools and files can be the reason to create personal agent on its own, without any specific system prompt

packages/cli/src/modules/chat-hub/chat-hub-agent.service.ts
```diff
@@ -58,16 +78,6 @@ export class ChatHubAgentService {
 		return await this.chatAgentRepository.getManyByUserId(userId);
 	}
 
-	async getAgentsByUserIdAsDtos(userId: string): Promise<ChatHubAgentDto[]> {
```
autologie, line 61: Cleaning up unused GET /agents endpoint and its implementation

packages/cli/src/modules/chat-hub/chat-hub-event-relay.service.ts
```diff
@@ -0,0 +1,34 @@
+import { Logger } from '@n8n/backend-common';
+import { Service } from '@n8n/di';
+
+import { EventService } from '@/events/event.service';
+
+import { ChatHubAgentService } from './chat-hub-agent.service';
+
+@Service()
+export class ChatHubEventRelay {
```
autologie, line 9: For deleting documents in vector store on user deletion

packages/cli/src/modules/chat-hub/chat-hub-upload.middleware.ts
```diff
@@ -0,0 +1,28 @@
+import { GlobalConfig } from '@n8n/config';
+import { Service } from '@n8n/di';
+import type { RequestHandler } from 'express';
+import multer from 'multer';
+
+@Service()
+export class ChatHubUploadMiddleware {
```
autologie, line 7: Using multipart upload instead of base64 in JSON payload to reduce peak memory usage

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -76,7 +79,7 @@ export class ChatHubWorkflowService {
 		private readonly workflowRepository: WorkflowRepository,
 		private readonly sharedWorkflowRepository: SharedWorkflowRepository,
 		private readonly chatHubAttachmentService: ChatHubAttachmentService,
-		private readonly chatHubAgentService: ChatHubAgentService,
+		private readonly chatHubAgentRepository: ChatHubAgentRepository,
```
autologie, line 82: The ChatHub agent service needs ChatHub workflow service as dependency for indexing documents. This change is to avoid dependency cycle.

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -87,7 +90,9 @@ export class ChatHubWorkflowService {
 	}
 
 	async deleteChatWorkflow(workflowId: string): Promise<void> {
-		await this.workflowRepository.delete(workflowId);
+		if (process.env.SKIP_CHAT_WORKFLOW_CLEANUP !== 'true') {
```
autologie: To facilitate debugging and testing locally

packages/frontend/editor-ui/src/features/ai/chatHub/components/AgentEditorModal.vue
```diff
@@ -282,107 +400,174 @@ async function onDelete() {
 			</div>
 		</template>
 		<template #content>
-			<div :class="$style.content">
-				<N8nInputLabel
-					input-name="agent-name"
-					:label="i18n.baseText('chatHub.agent.editor.name.label')"
-					:required="true"
+			<div :class="$style.contentWrapper">
```
autologie, line 414: Added wrapper to show D&D overlay

packages/@n8n/api-types/src/chat-hub.ts
Cadiac: Image

perhaps we should change the placeholder text to Enter system prompt (optional) then too?

packages/@n8n/nodes-langchain/credentials/ChatHubVectorStorePGVectorApi.credentials.ts
```diff
@@ -0,0 +1,21 @@
+import type { ICredentialType, INodeProperties } from 'n8n-workflow';
+
+export class ChatHubVectorStorePGVectorApi implements ICredentialType {
```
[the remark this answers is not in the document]
Cadiac: there appears to be test on the extension being enabled, I think this is false positive

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
Cadiac: should we make this a permanent feature and move it to n8n/packages/@n8n/config/src/configs/chat-hub.config.ts?

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -1410,4 +1491,251 @@ ${artifactsText}
...
+					metadata: {
+						metadataValues: [{ name: 'agentId', value: agentId }],
+					},
+				},
+			},
+			type: settings.vectorStore.nodeType,
+			typeVersion: 1,
+			position: [800, 496],
+			id: uuidv4(),
+			name: 'Vector Store',
```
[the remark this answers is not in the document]
Cadiac: Could be worth fixing for consistency

packages/@n8n/chat-hub/src/constants.ts
```diff
@@ -0,0 +1,53 @@
...
+		name: '@n8n/n8n-nodes-langchain.embeddingsCohere',
+		version: 1,
+	},
+	mistralCloud: {
+		name: '@n8n/n8n-nodes-langchain.embeddingsMistralCloud',
+		version: 1,
+	},
+};
+
+export const DEFAULT_SEMANTIC_SEARCH_SETTINGS: ChatHubSemanticSearchSettings = {
```
Cadiac, line 45: We don't filter this list by enabled providers, but I guess that's... to be expected? it would be basically only the admin changing these settings anyways.

packages/core/src/execution-engine/node-execution-context/node-execution-context.ts
```diff
@@ -76,6 +76,10 @@ export abstract class NodeExecutionContext implements Omit<FunctionsBase, 'getCr
 		return this.additionalData.executionId!;
 	}
 
+	getUserId() {
+		return this.additionalData.userId;
```
Cadiac: I guess this mechanism is my main worry in this PR. I did originally plan to do something similar for the simple memory but that plan was cancelled. Here we don't support our end users using these nodes relying on this on canvas, so the issue of this not getting a value on non-manual/chat executions shouldn't matter.

I'm not sure if this field has strictly speaking been a security feature, and now we're kind of turning it into one. The push at n8n lately has been towards dynamic credentials, but I suppose using them to prove the identity would be also cumbersome - did you look into it?

In practice I think this will work but I wonder if people have opinions about it.

There was something about this field not working on multi-main/queue mode in some case, but perhaps that had to do with waits - we should test that this works on multi-main / queue mode

Cadiac: Right, the issue I discovered earlier was
> additionalData is lost when execution enters waiting state and is later resumed, as we don't appear to persist it

which... likely doesn't affect this, as these workflows never enter waiting state, unless some tool causes that, and I don't think we support such tools?

packages/@n8n/chat-hub/src/constants.ts
autologie: Yes I think this can be separate. The provider that the admin wants people to use for conversation might not support embeddings, and keeping it flexible won't harm

packages/core/src/execution-engine/node-execution-context/node-execution-context.ts
autologie: > I'm not sure if this field has strictly speaking been a security feature, and now we're kind of turning it into one.

This is fair, let me check how dynamic credentials work and maybe benefit from it or the same mechanism

packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
autologie: This flag isn't intended for production use - do we make such env vars a config?

Cadiac: that's a good question, but us having "undocumented" env vars scattered accross the codebase seems worse to me 🤔 but I guess we already have some of them. Perhaps we should at least name this following those config env conventions, now the name is very generic?

packages/frontend/editor-ui/src/features/ai/chatHub/chat.store.ts
```diff
@@ -1249,6 +1278,68 @@ export const useChatStore = defineStore(STORES.CHAT_HUB, () => {
...
+			return 'notShared';
+		}
+
+		return undefined;
+	});
+
+	const semanticSearchReadiness = computed<SemanticSearchReadiness>(() => ({
+		isReadyForCurrentUser:
+			(!vectorStoreIssue.value || vectorStoreIssue.value === 'notShared') &&
+			(!embeddingIssue.value || embeddingIssue.value === 'notShared'),
```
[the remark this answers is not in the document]
autologie: This is intentional. Having access to the credential object itself confirms the current object is owned by the current user or shared with them. The change is to allow current user to upload files without sharing the credential with others.

packages/core/src/execution-engine/node-execution-context/node-execution-context.ts
autologie: Turns out that utilizing the dynamic credential infra or similar mechanism is tricky when the node's action is invoked from outside of an execution. As per talked about in SU, just replaced `getUserId` in `NodeExecutionContext` with a direct reference to `additionalData.userId` for now. 928b65ffbf35de4205d360f6bba9d18686df4dc7

packages/@n8n/nodes-langchain/nodes/vector_store/ChatHubVectorStorePGVector/ChatHubVectorStorePGVector.node.ts
```diff
@@ -0,0 +1,178 @@
...
+	payload: IDataObject | string | undefined,
+): Promise<NodeParameterValueType> {
+	const { filter } = (typeof payload === 'string' ? jsonParse(payload) : (payload ?? {})) as {
+		filter: Record<string, string | string[]>;
+	};
+
+	const credentials = await this.getCredentials<ChatHubVectorStorePGVectorApiCredentials>(
+		'chatHubVectorStorePGVectorApi',
+	);
+	const tableName = getUserScopedSlot(this, credentials.tableNamePrefix);
```
Cadiac, line 46: We still lack the identifier escaping on table names produced by this, meaning that prefix looks to still be vulnerable to injections?

autologie: `getUserScopedSlot` replaces all non alphanumeric characters with `_`, are there still possibilities where the final table name can be exploitable?

Cadiac: sorry I misread that code, I thought we were doing it only for the userId.
in that case I think we're safe 👍

Cadiac (dismissed):
Implementation looks good, the couple of issues we've discussed on DM's remain but 🚀

packages/cli/src/modules/chat-hub/chat-hub-agent.service.ts
```diff
@@ -175,8 +187,248 @@ export class ChatHubAgentService {
...
+		const fileItem = agent.files.find((f) => f.id === fileKnowledgeId);
+		if (!fileItem) {
+			throw new NotFoundError('File not found');
+		}
+
+		try {
+			await this.deleteEmbeddingsByKnowledgeId(user, agentId, fileItem.id);
+		} catch (error) {
+			this.logger.warn(
+				`Could not delete embeddings from vector store for file "${fileItem.fileName}" (agentId: ${agentId}, knowledgeId: ${fileItem.id}). Proceeding with file deletion.`,
```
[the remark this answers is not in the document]
autologie: This is intentional; the error is expected if the instance admin has switched vector store provider before deleting documents (no corresponding index/collection/table in the new vector store yet). In that case, the admin is responsible of a manual cleanup of the old vector store if necessary, and deleting files from an agent should be allowed.

Cadiac (approved):
👍

PR #26524, opened 2026-03-04
fix(core): Clean up resolver references on deletion

cstuncsik:
## Summary

When a credential resolver is deleted, workflows that reference it keep the stale ID, causing execution failures and display issues. This fix adds cleanup of workflow settings on deletion, shows affected workflows in a warning dialog, and handles existing stale IDs on the frontend.

## How to test

1. Create a credential resolver and assign it to 2+ workflows via Workflow Settings
2. Delete the resolver — verify a warning dialog appears listing the affected workflow names
3. Confirm deletion — verify the resolver is removed and workflows no longer reference it
4. Open one of the previously affected workflows' settings — verify the resolver dropdown shows no selection (not a stale ID)
5. Create another resolver, assign it to a workflow, then delete the resolver via API/DB directly (simulating a stale reference)
6. Open that workflow's settings — verify the stale resolver ID is cleared on load

https://linear.app/n8n/issue/IAM-145

## Review / Merge checklist

- [x] Tests included (backend + frontend comprehensive coverage)
- [x] All tests passing
- [x] Lint and typecheck passing


packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver.service.ts
```diff
@@ -155,8 +156,21 @@ export class DynamicCredentialResolverService {
 		return this.withDecryptedConfig(saved);
 	}
 
+	/**
+	 * Finds workflows that reference a credential resolver by ID.
+	 * @throws {DynamicCredentialResolverNotFoundError} When resolver is not found
+	 */
+	async findAffectedWorkflows(id: string): Promise<Array<{ id: string; name: string }>> {
```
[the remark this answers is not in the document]
cstuncsik: The endpoint is protected by `@GlobalScope('credentialResolver:read')` which is the same permission level as other resolver endpoints. The returned data is limited to workflow id/name only. Filtering by per-workflow access would be incorrect here — this is an admin-level operation (deleting a resolver) that needs to show all impacted workflows regardless of the caller’s workflow-level permissions.

packages/@n8n/db/src/repositories/workflow.repository.ts
```diff
@@ -125,6 +125,48 @@ export class WorkflowRepository extends Repository<WorkflowEntity> {
 		return count > 0;
 	}
 
+	async findByCredentialResolverId(
```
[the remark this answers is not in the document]
cstuncsik: WorkflowRepository is the canonical location for all workflow database operations in this codebase. The existing file already contains similar JSON-field query methods (e.g., `hasAnyWorkflowsWithErrorWorkflow`). Extracting 2 small methods into a separate module would break the established pattern and add indirection. If the team decides to split the repository in the future, these methods would move along with the rest.

packages/frontend/editor-ui/src/features/resolvers/composables/useCredentialResolvers.ts
```diff
@@ -48,11 +58,50 @@ export function useCredentialResolvers() {
...
 	const confirmAndDeleteResolver = async (resolver: CredentialResolver): Promise<boolean> => {
+		let affectedWorkflows: CredentialResolverAffectedWorkflow[] = [];
+		try {
+			affectedWorkflows = await getCredentialResolverWorkflows(
+				rootStore.restApiContext,
+				resolver.id,
+			);
+		} catch (error) {
+			// Fall back to standard confirm dialog if fetching affected workflows fails
+			console.warn('Failed to fetch affected workflows for resolver deletion', error);
```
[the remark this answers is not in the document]
cstuncsik: This is a non-critical graceful fallback. The user still gets a functional deletion dialog, just without the affected workflows list. Reporting to Sentry would generate noise for an expected degradation path (e.g., network hiccup, resolver already deleted). `console.warn` provides sufficient observability for debugging without polluting error tracking.

packages/@n8n/db/src/repositories/workflow.repository.ts
```diff
@@ -125,6 +125,71 @@ export class WorkflowRepository extends Repository<WorkflowEntity> {
...
+			});
+		}
+
+		return await qb.getMany();
+	}
+
+	/**
+	 * Finds IDs of active workflows that reference a credential resolver.
+	 */
+	async findActiveByCredentialResolverId(resolverId: string): Promise<string[]> {
```
guillaumejacquart, line 139: <mzybe we could refactor this method with the previous one, as they both are doing almost the same thing

packages/frontend/editor-ui/src/features/resolvers/composables/useCredentialResolvers.ts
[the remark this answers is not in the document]
afitzek: Do we want any console.log statements in the FE?

afitzek (commented):
I think @guillaumejacquart point on the refactor is good.

packages/@n8n/db/src/repositories/workflow.repository.ts
```diff
@@ -125,6 +125,67 @@ export class WorkflowRepository extends Repository<WorkflowEntity> {
...
+
+	private addCredentialResolverFilter(
+		qb: SelectQueryBuilder<WorkflowEntity>,
+		resolverId: string,
+		method: 'where' | 'andWhere' = 'where',
+	): void {
+		const dbType = this.globalConfig.database.type;
+
+		if (dbType === 'postgresdb') {
+			qb[method]("workflow.settings ->> 'credentialResolverId' = :resolverId", { resolverId });
```
BGZStephen, line 156: Do we know how performant this query will be at scale? Querying by nested JSON isn't particularly good over large data sets, which could be compounded by the fact workflow settings are not indexed either

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver.service.ts
```diff
@@ -160,8 +178,32 @@ export class DynamicCredentialResolverService {
...
+		const { manager } = this.repository;
+		await manager.transaction(async (trx) => {
+			await this.workflowRepository.clearCredentialResolverId(id, trx);
+			await trx.remove(existing);
+		});
+
 		this.logger.debug(`Deleted credential resolver "${existing.name}" (${id})`);
+
+		// Reactivate affected active workflows so they pick up the cleared settings
+		await Promise.all(
```
BGZStephen: A couple of general comments:

* affectedWorkflows is unbounded, so it's potentially problematic given there are async operations involved
* What are the ramifications of a workflow failing to be reactivated? Is a log enough? I'm assuming this is a background task so this could cause workflows to enter an odd state without being able to notify system users

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
```diff
@@ -90,6 +92,27 @@ export class CredentialResolversController {
 		}
 	}
 
+	@Get('/:id/workflows')
+	@GlobalScope('credentialResolver:read')
+	async getAffectedWorkflows(
```
BGZStephen, line 97: There is a data leak possible with this endpoint, which is:

* `credentialResolver:read` is a permission widely available to admin, owner and member instance roles.
* With this permission, any user can call this endpoint and gain information on workflows that are not accessible to them (id, name)

It is OK for a background system task not to care about filters, but not for API endpoints to expose data that is not generally visible to users

BGZStephen (changes requested):

packages/@n8n/db/src/repositories/workflow.repository.ts
cstuncsik: Only runs on admin delete (rare), not a hot path. Same unindexed JSON pattern already used elsewhere in this file (e.g., `hasAnyWorkflowsWithErrorWorkflow`). Lter we can add a functional index if resolver adoption grows.

packages/cli/src/modules/dynamic-credentials.ee/services/credential-resolver.service.ts
cstuncsik: Improved on this a bit in https://github.com/n8n-io/n8n/pull/26524/changes/e857b8965a7ce965566a6873fa339eaf65d0b700

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers.controller.ts
cstuncsik: As far as I know `credentialResolver:read` is only granted to `global:owner` and `global:admin`
Members only get `credentialResolver:list`

BGZStephen: oops, you're right, ignore me!

afitzek (approved):

BGZStephen (approved):

PR #26527, opened 2026-03-04
fix(editor): Remove light-dark and use mixin (no-changelog)

heymynameisrob:
## Summary
When your n8n theme differs from your system theme, some variables stick to what your system theme is.

LightningCSS rewrites light-dark() into helper vars: var(--lightningcss-light, <light>) var(--lightningcss-dark, <dark>).
That produces two colors separated by a space. LightningCSS seems to incorrectly set the `--lightningcss-light` as `initial` so it's always chosen. 

This seems to happen when transforming the scss that isn't co-located in component styles. For example, `<Button />` uses `light-dark()` and works fine.

For now, just use mixin to fix the problem and will explore more on the root cause.

## Review / Merge checklist

- [ ] Tests included. 


konstantintieber (approved):
Tested manually by setting light theme on n8n settings while using dark theme on OS level

PR #26534, opened 2026-03-04
fix(Form Node): Improve form description processing

DawidMyslak:


packages/nodes-base/nodes/Form/utils/formNodeUtils.ts
```diff
@@ -29,7 +29,7 @@ export const renderFormNode = async (
 		title = context.evaluateExpression(`{{ $('${trigger?.name}').params.formTitle }}`) as string;
 	}
 
-	const description = options.formDescription ?? '';
+	const description = handleNewlines(sanitizeHtml(options.formDescription ?? ''));
```
DawidMyslak, line 32: Same as https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Form/utils/utils.ts#L634 👍

PR #26537, opened 2026-03-04
feat: Implement SSRFProtectionService with URL/IP validation (no-changelog)

tomi:
## Summary

Implements SSRF protection with URL/IP validation, DNS resolution with caching, and IP range blocking.

We want to use `dns.lookup` instead of `dns.resolve` to keep the resolving backwards compatible. `dns.lookup` does not give the IP TTLs. For this reason we only cache them locally for a very short time (1s). We have to see if this actually has any benefit in actual usage, or if we should remove it altogether.

https://linear.app/n8n/issue/CAT-2385

## Review / Merge checklist

- [x] Tests included (631 lines of test coverage)
- [ ] PR Labeled with `release/backport` (if applicable)


packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,204 @@
...
+
+		return { allowed: true };
+	}
+
+	/**
+	 * Create a custom DNS lookup function for injection into HTTP clients.
+	 * This is the single validation point that prevents TOCTOU attacks by
+	 * validating IPs at connection time rather than before the request.
+	 */
+	createSecureLookup(): (
```
dlavrenuek: I don't quite understand how this method is supposed to be used in the code. Let's say we have our http request helpers and would add an interception method for axios through `createSecureLookup()` - why do we expose the lookup options? Shouldn't we always retrieve both ipv4 and ipv6 addresses and if not how do I know which one I want to have? If I would implement url validation checks using this service, I would rather use `validateUrl()`, which does not care about options, or?

packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,204 @@
...
+	 */
+	async validateRedirect(redirectUrl: string): Promise<SsrfCheckResult> {
+		return await this.validateUrl(redirectUrl);
+	}
+
+	/**
+	 * Normalize IPv6 bracket notation from a URL hostname.
+	 * E.g. `[::1]` → `::1`, `127.0.0.1` → `127.0.0.1`.
+	 */
+	private normalizeIpInHostname(hostname: string): string {
```
dlavrenuek, line 164: Is it even possible to have an url like `http://[127.0.0.1]/test` ?

packages/cli/src/services/ssrf/dns-resolver.ts
```diff
@@ -0,0 +1,106 @@
...
+		const cached = await this.dnsCache.get(cacheKey);
+		if (cached) return cached;
+
+		const existing = this.inFlightByCacheKey.get(cacheKey);
+		if (existing) return await existing;
+
+		const lookupPromise = (async () => {
+			const addresses = await this.doLookup(hostname, normalized);
+			if (addresses.length > 0) {
+				await this.dnsCache.set(cacheKey, addresses, LOOKUP_CACHE_TTL_SECONDS);
```
dlavrenuek, line 54: I can see that we don't get the ttl due to usage of `dns.lookup` and not `dns.resolve`. This looks like a valid strategy to me, but it also means that the configured `SsrfProtectionConfig.dnsCacheMaxTtlSeconds` will never be used assuming that it will always be greater than LOOKUP_CACHE_TTL_SECONDS (1 sec). Should we rather remove the ttl parameter from `dnsCache.set` and always use the configured `dnsCacheMaxTtlSeconds`?

packages/cli/src/services/ssrf/dns-resolver.ts
```diff
@@ -0,0 +1,106 @@
...
+			}
+
+			return addresses;
+		})();
+
+		this.inFlightByCacheKey.set(cacheKey, lookupPromise);
+
+		return await lookupPromise.finally(() => {
+			this.inFlightByCacheKey.delete(cacheKey);
```
dlavrenuek, line 63: Nice implementation 👍

packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,204 @@
...
+
+	/**
+	 * Create a custom DNS lookup function for injection into HTTP clients.
+	 * This is the single validation point that prevents TOCTOU attacks by
+	 * validating IPs at connection time rather than before the request.
+	 */
+	createSecureLookup(): (
+		hostname: string,
+		options: LookupOptions,
+		onResult: (
```
dlavrenuek: Is there a reason to use a callback method rather than a promise?

packages/cli/src/services/ssrf/ssrf-protection.service.ts
```diff
@@ -0,0 +1,204 @@
...
+			error: NodeJS.ErrnoException | null,
+			address: string | LookupAddress[],
```
dlavrenuek: A bit of nitpicking: due to the union the type of `address` must always be narrowed in the code by `Array.isArray(address)` or similar while it depends on the provided options. Just from typing perspective you can define a `SecureLookup` with multiple signatures

```ts

type SecureLookup = {
    (
        hostname: string,
		options: { all: true } & LookupOptions,
		onResult: (
			error: NodeJS.ErrnoException | null,
			address: LookupAddress[],
        ) => void
    ): void,
    (
        hostname: string,
		options: { all?: false } & LookupOptions,
		onResult: (
			error: NodeJS.ErrnoException | null,
			address: string,
            family?: number
        ) => void
    ): void 
}

...

createSecureLookup(): SecureLookup
```

When using in the code you will know which type the address have if options are passed as a constant type

```ts
declare const secureLookup: SecureLookup;

secureLookup("test", {}, (error, address) => {
    address.toLowerCase(); // no error because address is a string
})
secureLookup("test", {all: false}, (error, address) => {
    address.toLowerCase(); // no error because address is a string
})
secureLookup("test", {all: true}, (error, address) => {
    address.toLowerCase(); // error because address is an array
})
```

[Example playground](https://www.typescriptlang.org/play/?ssl=28&ssc=2&pln=8&pc=1#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgGQggGsBXMAeTBgQiQGc4BvAKDnbiwFsEAbATwD8ALjhJSXAEZo4AHzgAiAJIAFAG4AWBXMWq1ANm3zSSACbB0yYKYDcbDgAtkMeiLETpUHSfOWk1uw44TF5eN0kiXmBMJG8zCytbe3Zocyg3BQQwTUsoehgjRSyDXPzChTU0SUwaLkKfBP8koMqoatrwyOjY43i-AJYAXxYWGH4wPABlYGxSKGBCEnI4AF5mZLgACg2ghwh8pEwuYFF8qGQAcwAaFgBIW4hqWgZRJmDQ0RgoUjxBuAAyAhEMiUJ50eg3e50ABKwHopF4MFE2yCqI4aCg0FEADkIOYAFKTAB0AFEoFAkBASSBcGCeu5QpD7phTKZ5vR6KJFiCAIKs9n0ADaAF0bmi0dw+EJROIpGgdhwAJSrAB8cDUEAQpg2itEGq1Yo4KPFewORxOcDOlyZjxo4Ne7zCoiwvHovwBQKWoLtDBtSFh8MRyIVaIxWLguIJxLJFKpNOAdJ04kZd2Z-LhnMtX2tIdRkoEbllnlzypWav12qCuvVmtMcCGI3M2F4mHmcGw4Pgbtm8255FE0x7C2B5DsLG7c2HXs2ClQZSuzEGC82YagC5ZbIzpbVrCCG4FRJgEEIAHc0ABhTBuzaKmxwAD097EEDgq7g0mwmFIbuC6Y5iEYTAs3OJALiGRVxxmSc+zAGc5wKBcmBCXhnRCN0ly2Vd1z-eht3WPccMPY8IDPKBL2vW8HyfSlX3JaB3xmL8f33DMAOCYDLnAyChxguC4QQ5hkM+b5gAwlc6LXX9Nw5PDdw4FiOSI08LyvYAbzvR9aMxLwPyYvAFMYBBANiVsoEwfhwKAA)

packages/cli/src/services/ssrf/hostname-matcher.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Patterns like `*.n8n.internal` match any subdomain
+ * (e.g. `api.n8n.internal`, `deep.sub.n8n.internal`), but not the bare domain.
+ */
+export class HostnameMatcher {
+	private readonly parsed: ParsedPattern[];
+
+	constructor(patterns: readonly string[]) {
+		this.parsed = patterns
+			.map((p) => this.normalizeHostname(p))
+			.filter((p) => p.length > 0)
```
dlavrenuek: I saw following somewhere in the code recently, which I really like (Java static method style)


```suggestion
			.filter(Boolean)
```

packages/cli/src/services/ssrf/ip-range-blocklist-builder.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface BlocklistBuildResult {
+	blocklist: BlockList;
+	issues: BlocklistIssue[];
+}
+
+/**
+ * Builds a `net.BlockList` from a list of CIDR IP ranges.
+ * Invalid entries are skipped and collected into an `issues` array for the caller to log.
+ */
+export function buildBlocklist(cidrRanges: readonly string[]): BlocklistBuildResult {
```
dlavrenuek: This is used for both blocklist and allowlist and the naming is confusing when looking at the code f.e. [here](https://github.com/n8n-io/n8n/pull/26537/changes#diff-5d07baaf6890e153e7e7ace37a9457ac4f0c84f9fd26401580571b8f3e5329baR59) `this.allowedIps = allowed.blocklist;`. Should we make the name more generic like `ip-range-builder` and return `{ list, issues }`?

dlavrenuek (changes requested):

packages/cli/src/services/ssrf/ssrf-protection.service.ts
tomi: The bracket notation is for [connecting to ipv6 addresses](https://www.ietf.org/rfc/rfc2732.txt). So `http://[127.0.0.1]/test` is not valid URL, since `127.0.0.1` is not a valid IPv6 address

packages/cli/src/services/ssrf/ssrf-protection.service.ts
tomi: It's meant to be a 1:1 replacement for nodejs `dns.lookup`. Nodejs and axios allow specifying a custom lookup function e.g. for `socket.connect` and `axios.request`. This function will return a fn that implements that interface.

We expose the lookup options so nodejs internals can decide which IP family it wants to resolve. Which IP nodejs will actually want to resolve depends on a few configuration options:
- Is the family specified directly (either in [`socket.connect`](https://nodejs.org/docs/latest-v24.x/api/net.html#socketconnectoptions-connectlistener) or `axios.request`)
- Is autoselect configured for nodejs (either by [`dns.setDefaultAutoSelectFamily`](https://nodejs.org/docs/latest-v24.x/api/net.html#netsetdefaultautoselectfamilyvalue) or given directly to `socket.connect`)

I'll clarify this behaviour in the comment

packages/cli/src/services/ssrf/dns-resolver.ts
tomi: Yeah, unfortunately I realized this only after implementing the caching. I think I'd rather just remove the cache TTL option altogether and perhaps even the entire cache. I don't think it's a good strategy to have a long TTL for all lookups despite what the actual TTL is. It would cause too many issues.

packages/cli/src/services/ssrf/ssrf-protection.service.ts
tomi: See https://github.com/n8n-io/n8n/pull/26537#discussion_r2885494881

packages/cli/src/services/ssrf/ssrf-protection.service.ts
tomi: See https://github.com/n8n-io/n8n/pull/26537#discussion_r2885494881 and [`dns.lookup`](https://nodejs.org/docs/latest-v24.x/api/dns.html#dnslookuphostname-options-callback). `dns.lookup` returns a single IP unless all IPs are requested

packages/cli/src/services/ssrf/hostname-matcher.ts
tomi: It is more concise, but also more implicit. You have to deduct what are we filtering here for, and understand the coalescing rules of strings. I don't have too strong opinions so happy to change it 👍

packages/cli/src/services/ssrf/ip-range-blocklist-builder.ts
tomi: Good point! I'll clarify it

packages/cli/src/services/ssrf/ssrf-protection.service.ts
dlavrenuek: Thanks a lot for the clarification, that makes sense. Then the type should be rather `typeof dns.lookup` (with original dns being used here and not promise based) instead of a custom type, this way the intention should be more clear and it will be less error prone (for example due to changes in the nodejs interface, which is very unlikely). But I don't have a too strong opinion here

packages/cli/src/services/ssrf/ssrf-protection.service.ts
dlavrenuek: Good to know, I learned something new!

packages/cli/src/services/ssrf/ssrf-protection.service.ts
dlavrenuek: Oh, you did already change the signature to the native lookup function, very nice 🎉

dlavrenuek (approved):
Very nice 🚀

PR #26543, opened 2026-03-04
feat(editor): Show redacted state in execution viewer with reveal flow

cstuncsik:
## Summary

Display redacted execution data state in the NDV (Node Details View). When execution data is redacted by workflow policy, both input and output panels show a lock icon with an explanation and an optional "Reveal data" button for authorized users. Clicking reveal shows a confirmation modal, then fetches unredacted data via `?redactExecutionData=false`.

Backend: Force-redact executions that used dynamic credentials (`runtimeData.credentials`) with `reason: 'dynamic_credentials'` and `canReveal: false`. Attempting to reveal such executions returns `ForbiddenError`.

### How to test

1. Enable the redaction module and set a workflow's redaction policy to redact all executions
2. Run the workflow, then open the execution in the NDV
3. Verify both input and output panels show a lock icon with "Data redacted" title and a description linking to workflow settings
4. As a user with `execution:reveal` scope, verify the "Reveal data" button appears
5. Click "Reveal data" → confirm in the warning modal → verify data appears in both panels
6. As a user without `execution:reveal` scope, verify no reveal button is shown
7. Trigger an execution using dynamic credentials → verify the redacted state shows a distinct message ("This execution used dynamic credentials...") with no reveal button
8. Verify pinned data still displays normally even when the execution is redacted
9. Verify nodes that didn't run show "Node hasn't run" instead of the redacted state
10. Verify nodes with errors show the error state instead of the redacted state

https://linear.app/n8n/issue/IAM-197

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
```diff
@@ -97,17 +104,33 @@ export class ExecutionRedactionService implements ExecutionRedaction {
 		// reveal path (redactExecutionData === false). NodeDefinedFieldRedactionStrategy
 		// always runs — node-declared sensitive fields are never revealable.
 		for (const execution of executions) {
+			const hasDynCreds = this.hasDynamicCredentials(execution);
 			const policyAllowsReveal = this.policyAllowsReveal(execution);
-			const userCanReveal = policyAllowsReveal || revealableIds.has(execution.workflowId);
+			// Dynamic credential executions can never be revealed regardless of permissions
+			const userCanReveal = hasDynCreds
```
guillaumejacquart, line 133: Maybe we could instead add the hasDynamicCredentials to the RedactionContext, and use that information to set the proper reason on the redaction strategies

packages/frontend/editor-ui/src/features/execution/executions/composables/useExecutionRedaction.ts
```diff
@@ -0,0 +1,61 @@
...
+
+	const canReveal = computed(() => redactionInfo.value?.canReveal === true);
+
+	const isDynamicCredentials = computed(
+		() => redactionInfo.value?.reason === 'dynamic_credentials',
+	);
+
+	async function revealData() {
+		const warningHtml = [
+			`<p>${i18n.baseText('ndv.redacted.revealModal.warning')}</p>`,
```
guillaumejacquart: Is this common pattern ? Isn't there a way to make that tsx instead ?

guillaumejacquart (commented):
2 comments, but looks good

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
```diff
@@ -190,6 +211,14 @@ export class ExecutionRedactionService implements ExecutionRedaction {
 		return pipeline;
 	}
 
+	/**
+	 * Returns true when the execution used dynamic credential resolution.
+	 * Such executions must always be redacted with canReveal = false.
+	 */
+	private hasDynamicCredentials(execution: RedactableExecution): boolean {
+		return Boolean(execution.data.executionData?.runtimeData?.credentials);
```
afitzek: This actually checks is an execution context with credentials was created, this is for example true for every chathub workflow execution. We do track for each execution if a dynamic resolution actually happened during runtime here: https://github.com/n8n-io/n8n/blob/e4e929229bddd85c100988ecbe5b134e11dea514/packages/workflow/src/interfaces.ts#L2751 Maybe we want to use this one?

afitzek (commented):
I think this is good, just the way we determine if an execution is dynamic, should be adapted in my opinion.

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
afitzek: This is how we determine this for a telemetry event https://github.com/n8n-io/n8n/blob/e4e929229bddd85c100988ecbe5b134e11dea514/packages/cli/src/events/relays/telemetry.event-relay.ts#L817

packages/workflow/src/interfaces.ts
```diff
@@ -3056,7 +3056,7 @@ export type WorkflowActivateMode =
 export namespace WorkflowSettings {
 	export type CallerPolicy = 'any' | 'none' | 'workflowsFromAList' | 'workflowsFromSameOwner';
 	export type SaveDataExecution = 'DEFAULT' | 'all' | 'none';
-	export type RedactionPolicy = 'none' | 'all' | 'non-manual';
+	export type RedactionPolicy = 'none' | 'all' | 'non-manual' | 'manual-only';
```
afitzek, line 3059: What is the reason for a 'manual-only' mode in terms of redaction?

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
afitzek: I think this should be addressed.

packages/workflow/src/interfaces.ts
cstuncsik: The `manual-only` policy exists because we have two independent selects in Workflow Settings. One for production executions and one for manual executions. The four combinations (redact/default × redact/default) map to `all`, `non-manual`, `manual-only`, and `none`. Without `manual-only` we couldn’t represent the “redact manual but not production” state.

afitzek (approved):
Nice 🚀

PR #26546, opened 2026-03-04
chore(core): Redact sensitive node output fields via configurable strategy pipeline

afitzek:
## Summary

Refactors `ExecutionRedactionService` into a composable strategy pipeline and adds a mechanism for node authors to declare output fields that must always be redacted — regardless of workflow policy, user permissions, or execution mode.

### Problem

Webhook nodes pass raw HTTP request data (including `Authorization` and `Cookie` headers) directly into workflow execution data. Any user with `execution:read` scope could read these values. There was no way to prevent this without redacting the entire execution.

### Solution

**New type system** (`packages/workflow`):
- `IRedactedFieldMarker` — a sentinel value that replaces redacted fields in `item.json`, carrying `{ __redacted: true, reason, canReveal }` so the frontend can render appropriate UI
- `sensitiveOutputFields?: string[]` on `INodeTypeDescription` — dot-notation paths (e.g. `headers.authorization`) that a node declares as always-sensitive

**Strategy pipeline** (`packages/cli`):
- `IExecutionRedactionStrategy` interface — each strategy receives the execution and a `RedactionContext` and owns its own mutations
- `FullItemRedactionStrategy` — extracts the existing item-clearing logic; included by the orchestrator when policy or an explicit request requires it
- `NodeDefinedFieldRedactionStrategy` — walks `sensitiveOutputFields` paths in `item.json` for every node that declares them; runs in all cases including on the "reveal" path, since these fields are **never revealable**
- `ExecutionRedactionService` now acts as an orchestrator: resolves `userCanReveal` once, builds the pipeline, and delegates all mutations to strategies

**Fail-closed behaviour**: if a node's type cannot be resolved at redaction time (e.g. a community node was uninstalled), `NodeDefinedFieldRedactionStrategy` wipes the entire node output conservatively rather than skipping it.

**Reference implementation** (`packages/nodes-base`):
- Webhook node declares `sensitiveOutputFields: ['headers.authorization', 'headers.cookie']`

### How it works end-to-end

1. A webhook fires with an `Authorization: Bearer <token>` header
2. The execution is stored with the raw header value
3. When the execution is retrieved, `NodeDefinedFieldRedactionStrategy` replaces `item.json.headers.authorization` with `{ __redacted: true, reason: 'node_defined_field', canReveal: false }`
4. Even if the requesting user is a global owner with `execution:reveal`, this field is not restored — `canReveal: false` is unconditional for node-declared fields

### Example screenshot

image

https://linear.app/n8n/issue/IAM-278

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/redaction/executions/strategies/node-defined-field-redaction.strategy.ts
+++ b/packages/cli/src/modules/redaction/executions/strategies/node-defined-field-redaction.strategy.ts
@@ -1,6 +1,6 @@
 import { Logger } from '@n8n/backend-common';
 import { Service } from '@n8n/di';
-import type { INodeExecutionData, IRedactedFieldMarker } from 'n8n-workflow';
+import type { INodeExecutionData, IRedactedFieldMarker, ITaskDataConnections } from 'n8n-workflow';
 
 import type { RedactableExecution } from '@/executions/execution-redaction';
 import { NodeTypes } from '@/node-types';
@@ -20,14 +20,23 @@ export class NodeDefinedFieldRedactionStrategy implements IExecutionRedactionStr
 	) {}
 
 	async apply(execution: RedactableExecution, _context: RedactionContext): Promise<void> {
-		const sensitiveFieldsMap = this.buildSensitiveFieldsMap(execution);
-		if (sensitiveFieldsMap.size === 0) return;
+		const { sensitiveFields, unknownNodes } = this.buildSensitiveFieldsMap(execution);
+		if (sensitiveFields.size === 0 && unknownNodes.size === 0) return;
 
 		const runData = execution.data.resultData.runData;
 		if (!runData) return;
 
 		for (const nodeName of Object.keys(runData)) {
-			const fieldPaths = sensitiveFieldsMap.get(nodeName);
+			if (unknownNodes.has(nodeName)) {
+				for (const taskData of runData[nodeName]) {
+					if (taskData.data) {
+						this.redactAllOutputs(taskData.data);
+					}
+				}
+				continue;
+			}
+
+			const fieldPaths = sensitiveFields.get(nodeName);
 			if (!fieldPaths?.length) continue;
 
 			for (const taskData of runData[nodeName]) {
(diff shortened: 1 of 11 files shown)


packages/cli/src/modules/redaction/executions/strategies/node-defined-field-redaction.strategy.ts
```diff
@@ -0,0 +1,157 @@
...
+				// Fail fast — path does not exist, nothing to redact
+				return;
+			}
+			current = next;
+		}
+
+		const lastSegment = segments[segments.length - 1];
+		if (!(lastSegment in current)) return;
+
+		const marker: IRedactedFieldMarker = {
```
guillaumejacquart, line 150: I'm just wondering, this could potentially change the type of the field value, right ? Like it should have been a string, but is now an object with the redacted info. 
I'm just wondering if this could have impacts or trigger regression if a field has an expected value

packages/cli/src/modules/redaction/executions/strategies/node-defined-field-redaction.strategy.ts
```diff
@@ -0,0 +1,157 @@
...
+			__redacted: true,
+			reason: 'node_defined_field',
+			canReveal: false,
```
guillaumejacquart, line 153: just wondering if we want to make some fields revealable in the future. For instance i see we're redacting the headers.cookie field, but it could contain non auth cookies useful for debugging

guillaumejacquart (commented):
Looks good ! 
2 comments / questions

packages/cli/src/modules/redaction/executions/strategies/node-defined-field-redaction.strategy.ts
afitzek: Sure, we can make some revealable in the future, but currently no plans for this.

packages/cli/src/modules/redaction/executions/strategies/node-defined-field-redaction.strategy.ts
afitzek: These redacted values should only be used for display, so we will have to adapt the FE to look for this marker and display it. I don't think this would have any impact on executions, if this is your concern.

guillaumejacquart (approved):

PR #26550, opened 2026-03-04
fix(core): Correct `process.version` in expression sandbox

### LINKED ISSUE #26390: Expression custom evaluation for process.version returns PID instead of version
### Bug Description

When trying to return the expression `{{ process.version }}`, it returns the `process.pid` instead of resolving the `process.version`

The line [expression.ts](https://github.com/n8n-io/n8n/blob/master/packages/workflow/src/expression.ts#L390) shows that is simply retrurning the .pid instead of .version and I don't think that this is intended?

### To Reproduce

Use the following expression somewhere E.g Edit Fields:
```json
{
  "my_field_2": {{ process.version }}
}
```
Image

### Expected behavior

It should return the `process.version` if that wasn't intended

### Debug Info

Tested in n8n cloud FYI

# Debug info

## core

- n8nVersion: 2.9.4
- platform: docker (cloud)
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: 5

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 168 hours
- maxCount: 2500 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/145.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-03-01T23:31:44.541Z

### Operating System

N/A

### n8n Version

2.9.4

### Node.js Version

N/A

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

manus-pi:
## What

Fix `process.version` in the expression sandbox returning `process.pid` instead of the Node.js version string.

## Why

A copy-paste error in the sandbox `process` object definition caused `version` to be assigned `process.pid` (a number) instead of `process.version` (a semver string like `v24.13.1`).

Fixes #26390

## How

Single-character change in `packages/workflow/src/expression.ts`:

```diff
-  version: process.pid,
+  version: process.version,
```

## Testing

```bash
cd packages/workflow && vitest run test/expression.test.ts
```

Added regression tests that assert:
- `{{ process.version }}` matches `/^v\d+\.\d+\.\d+/`
- `{{ process.version }}` is not equal to `process.pid`
- `{{ typeof process.pid }}` is `'number'`

All 176 tests pass.

## Breaking Changes

None. This restores the documented behaviour — `process.version` should return the Node.js version string.

### DIFF
--- a/packages/workflow/src/expression.ts
+++ b/packages/workflow/src/expression.ts
@@ -387,7 +387,7 @@ export class Expression {
 						pid: process.pid,
 						ppid: process.ppid,
 						release: process.release,
-						version: process.pid,
+						version: process.version,
 						versions: process.versions,
 					}
 				: {};
--- a/packages/workflow/test/expression.test.ts
+++ b/packages/workflow/test/expression.test.ts
@@ -163,6 +163,13 @@ describe('Expression', () => {
 			expect(evaluate('={{Symbol(1).toString()}}')).toEqual(Symbol(1).toString());
 		});
 
+		it('should expose correct process properties in sandbox', () => {
+			expect(evaluate('={{process.version}}')).toMatch(/^v\d+\.\d+\.\d+/);
+			expect(evaluate('={{typeof process.pid}}')).toBe('number');
+			expect(evaluate('={{process.version}}')).not.toBe(process.pid);
+			expect(evaluate('={{process.version}}')).toBe(process.version);
+		});
+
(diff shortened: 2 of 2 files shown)


ivov (approved):
Thanks for your contribution! Once CI is green we'll bring it in.

PR #26558, opened 2026-03-04
fix(ai-builder): Don't use autoGenerate for UUID columns for WFB persistence migration

mike12345567:
## Summary

With managed Postgres installations like Supabase for Azure PSQL we have been encountering issues with the recent AI WFB persistence migration.

Example: https://community.n8n.io/t/function-uuid-generate-v4-does-not-exist/272516

The way we seem to be handling this now is generating UUIDs at the application layer to remove the need for auto-generation - I've also added comments to the TypeORM functions we should avoid using.

## Review / Merge checklist

- [ ] Tests included. 


tomi (approved):
🚀

PR #26574, opened 2026-03-05
feat: Allow users with custom project scope to access secrets completions (no-changelog)

ireneea:
## Summary

Users with a custom project role that includes the `externalSecret:list` scope can now access secrets provider completions. The can access both global secrets (if they have the scope in any project) and project-scoped secrets (if they have the scope in that specific project).

Previously, only global scope holders could access global secrets, blocking legitimate project-scoped custom role users.

## Test

**Setup**

1. create a global connections with secrets
2. create project-scoped connection with secrets
3. Add a custom role that has `externalSecret:list`
4. Add a `userA` to your project and assign them the new custom role
5. Add a `userB` to your project as a viewer  

**Test**

- `userA` should have access to global and project-scope secrets in when typing a `$secrets.` expression in a credential field 
- `userB` **should not** have access to global and project-scope secrets in when typing a `$secrets.` expression in a credential field 
- instance admins should have access to global and project-scope secrets in when typing a `$secrets.` expression in a credential field 

> NOTE saving the credential with `userA` will not work. There is a separate ticket to update the validation (LIGO-313)

https://linear.app/n8n/issue/LIGO-290

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/external-secrets.ee/secrets-providers-completions.controller.ee.ts
```diff
@@ -51,16 +54,26 @@ export class SecretProvidersCompletionsController {
 	}
 
 	@Get('/secrets/global')
-	@GlobalScope('externalSecret:list')
-	@ProjectScope('externalSecret:list')
-	async listGlobalSecrets(): Promise<SecretCompletionsResponse> {
+	async listGlobalSecrets(
```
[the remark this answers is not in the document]
ireneea: Added a comment

packages/frontend/editor-ui/src/features/integrations/externalSecrets.ee/externalSecrets.ee.store.ts
```diff
@@ -121,7 +130,14 @@ export const useExternalSecretsStore = defineStore('externalSecrets', () => {
...
 		}
 
-		if (rbacStore.hasScope('externalSecret:list')) {
+		const project =
+			projectsStore.currentProject?.id === projectId
+				? projectsStore.currentProject
+				: projectsStore.myProjects.find((p) => p.id === projectId);
+		if (
+			rbacStore.hasScope('externalSecret:list') ||
+			project?.scopes?.includes('externalSecret:list') === true
```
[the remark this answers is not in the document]
ireneea: This comment is technically valid. But this is a pre-existing issue where project-scoped secrets are never cleared/reset. 

I have created a dedicated [project](https://linear.app/n8n/project/secrets-memory-usage-optimisation-9e3b0acf9f16/overview) to optimise usage of secrets list overall and this item has been mentionened there

packages/cli/test/integration/external-secrets/secrets-providers-completions.api.test.ts
```diff
@@ -106,6 +139,20 @@ describe('Secret Providers Completions API', () => {
 					expect(response.body.message).toBe(FORBIDDEN_MESSAGE);
 				}
 			});
+
+			it('should allow member with custom externalSecret:list role to list global secrets', async () => {
```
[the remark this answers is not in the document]
ireneea: The Authorization tests intentionally only verify status codes (200/403) — the same pattern used by the owner/admin/member tests above them. Response body verification is covered in the separate “with global connections” and “with existing project connections” describe blocks.

packages/cli/src/modules/external-secrets.ee/secrets-providers-completions.controller.ee.ts
```diff
@@ -50,17 +53,34 @@ export class SecretProvidersCompletionsController {
...
+	 */
 	@Get('/secrets/global')
-	@GlobalScope('externalSecret:list')
-	@ProjectScope('externalSecret:list')
-	async listGlobalSecrets(): Promise<SecretCompletionsResponse> {
+	async listGlobalSecrets(
+		req: AuthenticatedRequest,
+		res: Response,
+	): Promise<SecretCompletionsResponse | undefined> {
+		const hasAccess = await this.accessCheckService.userHasGlobalScopeOrAnyProjectScope(
```
konstantintieber: The logic that global secrets should be accessible if a user has `externalSecret:list` on any project they are a member of seems arbitrary to me.
Can you explain why we check for this specifically?

Given that these secrets are global, how about we tie it to `credential:create` or `credential:edit` instead?

EDIT:
If I had to explain to a user (in docs for example), that if they have a member user, who is project admin for one project (and this project role has `externalSecret:list`), and project editor for three other projects, that that user can use global secrets in credentials of all four projects, ONLY AS LONG as they have the scope externalSecret:list in one of the projects. Meaning once they're no longer the admin for that one project, they can't reference global secrets in the other anymore.
I wouldn't know how to explain it.

ireneea: The requirement is that if a user can use project scoped secrets they should also be able to use global secrets. Using secrets is controlled by the `externalSecret:list` that's why it's used here.

Using `externalSecret:list` will allow us to have the opt in toggle, because we'll be able to only affect the relevant scope dynamically.

konstantintieber: Update for complete data: that is not the product requirement.
Here is the actual one (copied from https://n8nio.slack.com/archives/C0A7M5T04KW/p1772723636791489)

> All users with credential edit or write permissions be able to use global secrets

packages/cli/src/modules/external-secrets.ee/secrets-providers-completions.controller.ee.ts
```diff
@@ -52,15 +52,34 @@ export class SecretProvidersCompletionsController {
...
+	 * As a result any user who can create or edit a project credential can use the global secrets.
+	 */
+	@Get('/secrets/global/:projectId')
+	@ProjectScope('externalSecret:list')
+	async listGlobalSecretsForProject(
+		_req: AuthenticatedRequest,
+		_res: Response,
+		@Param('projectId') projectId: string,
+	): Promise<SecretCompletionsResponse> {
+		this.logger.debug('Listing global secrets for project', { projectId });
```
konstantintieber, line 77: Do I see this correctly that the only reason we're passing in the `projectId` param here is to include it in the debug logs?

ireneea: The real key aspect of the `projectId` is to be part of the url `/secrets/global/:projectId` for the access control. (see js doc above)

But yes in this case having it accessible in this method is just about having it in the log. Happy to remove it though

konstantintieber: What confuses me is why we now have two separate endpoints for fetching completions for global secrets. I'd expect there's just a single one.

ireneea: Agreed, now that they both have the same access control it does make sense to have only one endpoint. 

Is it ok if we do that as a follow-up task?

konstantintieber: It would be easier to review the code if it would already be in that state in this PR, since I'd have to actively "ignore" the parts that I don't approve of yet.

But if you want an early approve and it's behind feature flag anyway, I can provide that.

packages/cli/src/modules/external-secrets.ee/secrets-providers-completions.controller.ee.ts
```diff
@@ -52,15 +52,34 @@ export class SecretProvidersCompletionsController {
...
+	 *
+	 * On the system role the externalSecret:list scope is compatible with the credential create and edit scopes.
+	 * As a result any user who can create or edit a project credential can use the global secrets.
+	 */
+	@Get('/secrets/global/:projectId')
+	@ProjectScope('externalSecret:list')
+	async listGlobalSecretsForProject(
+		_req: AuthenticatedRequest,
+		_res: Response,
+		@Param('projectId') projectId: string,
```
konstantintieber, line 75: Would be nice to have a comment here that the `@param projectId` is only required for `@ProjectScope` decorator to work.

packages/cli/src/modules/external-secrets.ee/secrets-providers-completions.controller.ee.ts
konstantintieber: I understand now that the `projectId` is needed for the `@ProjectScope` decorator to work.

Will commence manual testing of this PR.

konstantintieber (approved):

PR #26586, opened 2026-03-05
chore: Apply biome formatting to playwright package (no-changelog)

shortstacked:
## Summary
- Applied biome formatter to the playwright package, fixing 127 files of formatting drift
- Added missing `format` and `format:check` scripts to `package.json` so `turbo run format:check` in CI now includes this package

## Why this was needed
The playwright package (and all `packages/testing/*` packages) never had `format` / `format:check` scripts. CI runs `pnpm format:check → turbo run format:check`, which only executes in packages that declare the script — so playwright was silently skipped. Lefthook covers staged files locally, but full-tree formatting was never enforced.

## Test plan
- [ ] CI `format:check` step passes (this is the main validation)
- [ ] No functional changes — formatting only


tomi (approved):
🚀

PR #26588, opened 2026-03-05
feat(core): Move enum check constraints to Column class (no-changelog)

tomi:
## Summary

Refactors the migration DSL to support declaring enum check constraints at the column level instead of the table level. This improves the API by:
- Making enum values discoverable alongside the column definition
- Enabling enum constraints when adding columns to existing tables via AddColumns
- Consolidating enum constraint logic into a shared buildEnumChecks() helper

## Changes
- Add withEnumCheck() method to Column class
- Update CreateTable and AddColumns to extract and apply enum checks from columns
- Deprecate table-level withEnumCheck() on CreateTable (backward compatible)
- Add 10 unit tests covering column-level enum checks, SQL generation, and quote escaping
- Update 2 existing migrations to use the new column-level API

### DIFF
--- a/packages/@n8n/db/src/migrations/dsl/table.ts
+++ b/packages/@n8n/db/src/migrations/dsl/table.ts
@@ -1,9 +1,29 @@
-import type { TableForeignKeyOptions, TableIndexOptions, QueryRunner } from '@n8n/typeorm';
+import type { Driver, TableForeignKeyOptions, TableIndexOptions, QueryRunner } from '@n8n/typeorm';
 import { Table, TableCheck, TableColumn, TableForeignKey, TableUnique } from '@n8n/typeorm';
 import { UnexpectedError } from 'n8n-workflow';
 import LazyPromise from 'p-lazy';
 
 import { Column } from './column';
+
+function buildEnumChecks(
+	columns: Column[],
+	prefix: string,
+	tableName: string,
+	driver: Driver,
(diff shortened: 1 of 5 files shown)


packages/@n8n/db/src/migrations/dsl/table.ts
```diff
@@ -67,6 +87,7 @@ export class CreateTable extends TableOperation {
 		return this;
 	}
 
+	/** @deprecated Declare enum checks on the column instead: `column('name').varchar().withEnumCheck([...])` */
 	withEnumCheck(columnName: string, values: string[]) {
```
ireneea: Since the migrations using this have been updated is it worth completely removing this?

tomi: Good point, let's remove it 👍

tomi: Done — removed `withEnumCheck` method and the `enumChecks` field from `CreateTable` entirely in e8fb670.

ireneea (approved):

packages/@n8n/db/src/migrations/dsl/table.ts
ireneea: Thanks 🙏

guillaumejacquart (approved):
Thanks for taking this initiative 🚀

PR #26602, opened 2026-03-05
ci: Normalize input paths for impact analysis orchestration (no-changelog)

shortstacked:
## Summary
- Rename 3 generic test workflow fixtures to descriptive names (`Test_workflow_1.json` → `manual-trigger-with-code.json`, etc.) fixing 3 janitor `test-data-hygiene` violations
- Fix path normalization bug where `orchestrate --impact` returned 0 affected specs due to mismatched path formats between impact analyzer output and discover output
- Add `resolveInputPaths` utility that handles absolute, rootDir-relative, and repo-root-relative path inputs consistently
- Make `ImpactAnalyzer.toRelative()` consistent with `toAbsolute()` by always normalizing through absolute path conversion

## Test plan
- [x] 331/331 janitor unit tests pass (including 4 new path normalization tests + 7 new `resolveInputPaths` tests)
- [x] `pnpm janitor orchestrate --shards=14 --impact --files=<rootDir-relative>` → correct count
- [x] `pnpm janitor orchestrate --shards=14 --impact --files=<repo-root-relative>` → correct count
- [x] `pnpm janitor orchestrate --shards=14 --impact --files=<absolute>` → correct count
- [x] Build and typecheck clean


packages/testing/janitor/src/utils/paths.ts
```diff
@@ -95,6 +95,28 @@ export function getRelativePath(absolutePath: string): string {
...
+ *
+ * Handles three input formats:
+ *   1. Absolute paths → returned as-is
+ *   2. Repo-root-relative paths (e.g. from `git diff`) → resolved against gitRoot
+ *   3. rootDir-relative paths → resolved against rootDir
+ *
+ * The heuristic: resolve against gitRoot first. If the result falls within
+ * rootDir, it's a repo-root-relative path. Otherwise treat it as rootDir-relative.
+ */
+export function resolveInputPaths(files: string[], gitRoot: string): string[] {
```
Matsuuu: Can we just resolve `gitRoot` from the config at this level? It would remove the unnecessary passing of `getGitRoot(config.rootDir)` to every single call of this function.

Currently I think the only reason we pass it here is for tests to be able to pass something extra to it, but I think we have a few approaches here that could improve QoL:

1. Make `gitRoot` optional, default to getGitRoot

```javascript
export function resolveInputPaths(files: string[], gitRoot?: string): string[] {
    if (!gitRoot) {
        gitRoot = getGitRoot(getConfig().rootDir);
    }
```

2. Mock the `getGitRoot` function in tests to make it return what the test wants.

Matsuuu (changes requested):

packages/testing/janitor/src/utils/paths.ts
shortstacked: Done! Thanks

Matsuuu (approved):

PR #26618, opened 2026-03-05
fix(Filter node): misleading description

### LINKED ISSUE #26564: Misleading description for the Filter node
### Bug Description

The filter node has a description of "Remove items matching a condition", but the behaviour of the node is to keep items matching the conditions set within the node. I would recommend rewording the description to something like "Keep items matching a condition" or "Keep only items matching a condition".

### To Reproduce

Click the + in a workflow to add a node
Type "filter" in the search box
Read the description of the Filter node

### Expected behavior

I expect the description to match the node's behaviour

### Debug Info

Not needed

### Operating System

Oracle Linux

### n8n Version

2.4.6

### Node.js Version

24.x.x

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

04cb:
Fixes #26564. Updates the Filter node description from 'Remove items matching a condition' to 'Keep items matching a condition' to accurately reflect the node's behavior.

### DIFF
--- a/packages/nodes-base/nodes/Filter/Filter.node.ts
+++ b/packages/nodes-base/nodes/Filter/Filter.node.ts
@@ -12,7 +12,7 @@ export class Filter extends VersionedNodeType {
 			icon: 'fa:filter',
 			iconColor: 'light-blue',
 			group: ['transform'],
-			description: 'Remove items matching a condition',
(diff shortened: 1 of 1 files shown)


krisn0x (dismissed):
Thank you, looks good

PR #26624, opened 2026-03-05
feat: Add lint rule to disallow httpRequest with manual authentication

garritfra:
## Summary

Adds a new ESLint rule `no-http-request-with-manual-auth` that flags community nodes manually handling authentication instead of using `this.helpers.httpRequestWithAuthentication()`.

What it catches:

```
// ❌ flagged — credentials fetched and httpRequest called in the same function
async function apiRequest() {
  const credentials = await this.getCredentials('myApi');
  const options = { headers: { Authorization: `Bearer ${credentials.apiKey}` } };
  return this.helpers.httpRequest(options);
}

// ✅ correct
async function apiRequest() {
  return this.helpers.httpRequestWithAuthentication.call(this, 'myApi', options);
}
```

Detection approach: Function-scope stack — if `this.getCredentials()` and `this.helpers.httpRequest()` both appear in the same function body, every httpRequest call is reported. Nested functions are checked independently. See the rule file for notes on alternatives considered and the known false-positive edge case.

The rule is added to both `recommended` and `recommendedWithoutN8nCloudSupport` configs as error.

None

## Review / Merge checklist

- [x] Tests included. 


garritfra:
Test coverage dropped only in the index files. Hope that's okay; I don't want to bloat up the PR with tests for these unrelated files. 🙈

packages/@n8n/eslint-plugin-community-nodes/src/rules/no-http-request-with-manual-auth.ts
```diff
@@ -0,0 +1,104 @@
...
+	return (
+		node.callee.type === AST_NODE_TYPES.MemberExpression &&
+		node.callee.object.type === AST_NODE_TYPES.ThisExpression &&
+		node.callee.property.type === AST_NODE_TYPES.Identifier &&
+		node.callee.property.name === method
+	);
+}
+
+/** Matches: this.helpers.methodName(...) */
+function isThisHelpersMethodCall(node: TSESTree.CallExpression, method: string): boolean {
```
elsmr: nit: The `no-deprecated-workflow-functions` rule has a similar util, maybe we could reuse? Also some of the logic is shared with `isThisMethodCall`. Maybe something worth extracting to `ast-utils`.

elsmr (commented):
Nice one! great tests too

Could you also add some md docs to `docs/rules` for this rule? We link to these from lint errors

packages/@n8n/eslint-plugin-community-nodes/src/rules/no-http-request-with-manual-auth.ts
garritfra: Good call! Addressed in https://github.com/n8n-io/n8n/pull/26624/changes/0d267644729dcbcc11185dd765bc29605983e82c

elsmr (approved):

PR #26625, opened 2026-03-05
feat(core): Add workflow:unpublish project scope

cstuncsik:
## Summary

This PR completes the implementation of the `workflow:unpublish` scope that was partially added in a prior effort. It fixes several issues: instance admins can now unpublish workflows (fixing the bug reported in the Slack channel), the custom role UI now shows a single "Publish workflow" checkbox that controls both publish and unpublish permissions, and all permission checks throughout the frontend now correctly gate on `workflow:unpublish` instead of `workflow:publish` or `workflow:update`.

https://linear.app/n8n/issue/IAM-127
fixes #26509

### 1. Instance Admin Unpublish (Global Scope Fix)
- Create a published workflow and log in as an instance admin (global:owner role)
- Open the workflow and verify the "Unpublish" button in the header is enabled
- Click "Unpublish" → verify it succeeds and workflow is unpublished
- Also test keyboard shortcut `Ctrl+U` to unpublish

### 2. Custom Role: Single Publish/Unpublish Checkbox
- Navigate to Project Settings → Roles
- Create a new custom role or edit an existing one
- In the Permissions tab, scroll to Workflow section
- Verify you see ONE "Publish" checkbox (not separate publish/unpublish)
- Check the checkbox → verify the role gets both `workflow:publish` and `workflow:unpublish` scopes
- Uncheck it → verify both scopes are removed

### 3. Workflow Card Unpublish Permission
- In the Workflows list, find a published workflow
- Hover over it and open the context menu (three dots)
- Verify "Unpublish" action appears and is enabled for a user with `workflow:unpublish` scope
- Create a role without `workflow:unpublish` and assign it to a user, then verify the "Unpublish" action is disabled

### 4. Keybinding Permission
- In a published workflow, press `Ctrl+U` with a user that has `workflow:unpublish` → should unpublish
- Remove `workflow:unpublish` from their role and try again → button should be disabled


packages/frontend/editor-ui/src/features/project-roles/projectRoleScopes.ts
```diff
@@ -64,8 +63,37 @@ export const SCOPES: Record<ProjectResource, ProjectRoleScope[]> = Object.fromEn
...
+ * state when an admin edits unrelated permissions.
+ */
+export function normalizeCoupledScopes(scopes: string[]): string[] {
+	const hasPublish = scopes.includes('workflow:publish');
+	const hasUnpublish = scopes.includes('workflow:unpublish');
+
+	if (hasPublish && !hasUnpublish) {
+		return [...scopes, 'workflow:unpublish'];
+	}
+	if (hasUnpublish && !hasPublish) {
```
BGZStephen, line 95: I'm guessing personal space policies blocking publishing of personal space workflows still works even with this?

afitzek: Yeah, this should only effect custom roles for projects, if we ever want to allow editing of buildin roles, we need to keep this in mind.

afitzek (approved):
🚀

packages/frontend/editor-ui/src/features/project-roles/projectRoleScopes.ts
cstuncsik: Yes, the normalization only applies to custom roles saved through the role editor UI. Built-in roles like `PERSONAL_PROJECT_OWNER_SCOPES` are defined in code and unaffected. 
`workflow:publish` is still conditionally added via `PERSONAL_SPACE_PUBLISHING_SETTING`.

So, personal space policies blocking publishing still work correctly, the personal project owner will have the `workflow:unpublish`, without `workflow:publish` as before.

PR #26626, opened 2026-03-05
chore(core): Polyfill crypto and URLSearchParams in VM expression engine isolate

afitzek:
## Summary

The VM expression engine runs expressions inside a sandboxed V8 isolate via `isolated-vm`. The isolate has no access to Web APIs, causing two expression extension functions to fail when `N8N_EXPRESSION_ENGINE=vm`:

- `Array.randomItem()` — internally uses `crypto.getRandomValues()` → throws `crypto is not defined`
- `Object.urlEncode()` — uses `new URLSearchParams()` → throws `URLSearchParams is not defined`

This fix adds conditional polyfills to the runtime bundle (`src/runtime/index.ts`) that activate only when the APIs are absent (inside the isolate) and are no-ops in jsdom/Node.js environments where these APIs already exist.

**`crypto.getRandomValues`:** `Math.random()` is used intentionally. A bridge to Node.js `crypto` via `ivm.Reference` was considered but rejected — an attacker-controlled `length` argument could allocate unbounded memory in the host process, bypassing the isolate's memory limit. `randomItem()` is non-security-critical (random array element selection), so `Math.random()` is appropriate.

**`URLSearchParams`:** A minimal polyfill implementing `constructor(Record<string, string>)` and `toString()`, which is the only interface used. `encodeURIComponent` is an ECMAScript built-in available in all V8 contexts.

https://linear.app/n8n/issue/CAT-2538

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/expression-runtime/src/extensions/array-extensions.ts
+++ b/packages/@n8n/expression-runtime/src/extensions/array-extensions.ts
@@ -1,12 +1,17 @@
 import isEqual from 'lodash/isEqual';
+import random from 'lodash/random';
 import uniqWith from 'lodash/uniqWith';
 
 import type { Extension, ExtensionMap } from './extensions';
 import { ExpressionExtensionError } from './expression-extension-error';
 import { compact as oCompact } from './object-extensions';
 
+// DIVERGENCE from packages/workflow/src/extensions/array-extensions.ts:
+// The original uses crypto.getRandomValues() which is a Web API unavailable
+// inside the V8 isolate. lodash/random (already bundled) is used instead — both
+// are backed by Math.random() and randomItem() is non-security-critical.
 function randomInt(max: number): number {
-	return crypto.getRandomValues(new Uint32Array(1))[0] % max;
+	return random(0, max - 1);
 }
 
 function first(value: unknown[]): unknown {
--- a/packages/@n8n/expression-runtime/src/extensions/object-extensions.ts
+++ b/packages/@n8n/expression-runtime/src/extensions/object-extensions.ts
@@ -84,8 +84,19 @@ export function compact(value: object): object {
 	return newObj;
 }
 
+// DIVERGENCE from packages/workflow/src/extensions/object-extensions.ts:
+// The original uses URLSearchParams which is a Web API unavailable inside the
+// V8 isolate. encodeURIComponent is an ECMAScript built-in available in all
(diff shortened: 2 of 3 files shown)


packages/@n8n/expression-runtime/src/runtime/index.ts
```diff
@@ -34,6 +34,57 @@ declare global {
...
+// Web API Polyfills (for V8 isolate which lacks Web APIs)
+// ============================================================================
+
+// Polyfill crypto.getRandomValues for V8 isolate (no Web APIs available).
+// Math.random() is intentionally used — randomItem() is non-security-critical.
+// A bridge to Node.js crypto.getRandomValues was considered but rejected: an
+// attacker-controlled `length` argument could allocate unbounded memory in the
+// host process, bypassing the isolate's memory limit.
+if (typeof crypto === 'undefined') {
+	(globalThis as any).crypto = {
```
despairblue: Instead of doing this, how about we replace the use of crypto to generate a random in in the extension file with `_.random`?

Turning this

https://github.com/n8n-io/n8n/blob/f4d786ea07fcfa9744a36d308ea8f44fa79dc143/packages/@n8n/expression-runtime/src/extensions/array-extensions.ts#L8-L10

into this:

```ts
// DIVERGENCE from packages/workflow/src/extensions/array-extensions.ts:
// The original uses crypto.getRandomValues() which is a Web API unavailable
// inside the V8 isolate. lodash/random (already bundled) is used instead — both
// are backed by Math.random() and randomItem() is non-security-critical.
function randomInt(max: number): number {
	return random(max);
}
```

That way we don't have to add any global.

packages/@n8n/expression-runtime/src/runtime/index.ts
```diff
@@ -34,6 +34,57 @@ declare global {
...
+		},
+	};
+}
+
+// Polyfill URLSearchParams for V8 isolate.
+// Only constructor(Record<string, string>) + toString() are needed —
+// the sole usage is object-extensions.ts urlEncode().
+// encodeURIComponent is an ECMAScript built-in, available in all V8 contexts.
+if (typeof URLSearchParams === 'undefined') {
+	(globalThis as any).URLSearchParams = class {
```
despairblue: Same with this, we can probably rewrite `urlEncode` to use `encodeURIComponent` directly.

despairblue (approved):

PR #26634, opened 2026-03-05
feat(core): Implements telemetry group identify calls

Joco-95:
## Summary
Implements telemetry (Rudderstack & PostHog) group calls to group instance users under common instance group for feature flag targeting & analytics segmentation. Includes general instance settings & licensing information.

Maintains legacy telemetry (Rudderstack identify call under `instanceId` on server startup) for cross-compatibility.

### DIFF
--- a/packages/cli/src/events/relays/telemetry.event-relay.ts
+++ b/packages/cli/src/events/relays/telemetry.event-relay.ts
@@ -1,3 +1,4 @@
+import { LicenseState } from '@n8n/backend-common';
 import { GlobalConfig } from '@n8n/config';
 import {
 	CredentialsRepository,
@@ -52,6 +53,7 @@ export class TelemetryEventRelay extends EventRelay {
 		readonly eventService: EventService,
 		private readonly telemetry: Telemetry,
 		private readonly license: License,
+		private readonly licenseState: LicenseState,
 		private readonly globalConfig: GlobalConfig,
 		private readonly instanceSettings: InstanceSettings,
 		private readonly binaryDataConfig: BinaryDataConfig,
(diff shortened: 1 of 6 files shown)


packages/cli/src/posthog/index.ts
```diff
@@ -40,6 +40,37 @@ export class PostHogClient {
...
+			groupKey: instanceId,
+			properties,
+			...(distinctId && { distinctId }),
+		});
+	}
+
+	identify({
+		distinctId,
+		properties,
+	}: { distinctId: string; properties: Record<string | number, any> | undefined }): void {
```
romeobalta: Should we avoid `any`?

```suggestion
	}: { distinctId: string; properties: Record<string | number, unknown> | undefined }): void {
```

packages/cli/src/telemetry/index.ts
```diff
@@ -63,6 +63,52 @@ export class Telemetry {
 		private readonly errorReporter: ErrorReporter,
 	) {}
 
+	// PostHog groupIdentify only accepts flat objects with string or number values, function sanitizes objects to match that format.
+	sanitizeTelemetryProperties(
+		obj: Record<string, any>,
```
romeobalta, line 68:
```suggestion
		obj: Record<string, unknown>,
```

packages/cli/src/telemetry/index.ts
```diff
@@ -202,21 +248,61 @@ export class Telemetry {
...
+	}: {
+		userId?: string;
+		traits?: Record<string, string | number>;
+	}): void {
+		const { instanceId } = this.instanceSettings;
+		if (!instanceId) return;
+
+		if (this.postHog) {
+			this.postHog.groupIdentify({
+				...(userId && { distinctId: `${userId}#${instanceId}` }),
```
romeobalta: Just double-checking this is correct, I know getFeatureFlags does `instanceId # userId` and here the order is reversed.

```ts
const fullId = [this.instanceSettings.instanceId, user.id].join('#');
```

romeobalta (commented):
Claude suggestions:

1. **No unit tests for `sanitizeTelemetryProperties`** — it's mocked out in tests. Has non-trivial logic (recursive flattening, depth limiting, type coercion) that deserves direct coverage.
2. **No unit tests for new `License` methods** — date arithmetic edge cases (null manager, NaN dates, past dates).
3. **`groupType`: 'company' is a magic string** — consider a constant.
4. **Test assertions are loose** — `expect.any(String)` would pass even with wrong values.

packages/cli/src/events/relays/telemetry.event-relay.ts
```diff
@@ -1444,3 +1581,33 @@ export class TelemetryEventRelay extends EventRelay {
 
 	// #endregion
 }
+
+function getSemanticVersioning(versionCli: string): {
```
juusotn8n: I think we have the `semver` package available, could use that to parse the version string

packages/cli/src/telemetry/index.ts
Joco-95: Thanks for catching this!

juusotn8n (approved):

PR #26646, opened 2026-03-06
feat: External secrets access based on system roles (no-changelog)

ireneea:
## Summary

- Enforce project-role-based access control on secrets provider connections, users’ write access is now computed from their project role rather than global permissions alone
- Add `secretsProvider` scopes to the role/permissions system, including project-level connection sharing scopes
- Update the connection modal (frontend) to pass project context when creating/updating provider connections
- Users with global admin/owner access bypass project-scope checks; project members get scoped permissions based on their role

resolves [LIGO-311](https://linear.app/n8n/issue/LIGO-311/open-secrets-update-repository-queries-for-role-based-access)

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/external-secrets.ee/secret-provider-access-check.service.ee.ts
```diff
@@ -1,11 +1,114 @@
-import { SecretsProviderConnectionRepository } from '@n8n/db';
+import type { User } from '@n8n/db';
```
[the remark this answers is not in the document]
ireneea: tests have been added now

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -130,7 +132,11 @@ export class SecretsProvidersConnectionsService {
 		await this.repository.save(connection);
 
 		if (updates.projectIds !== undefined) {
-			await this.projectAccessRepository.setProjectAccess(connection.id, updates.projectIds);
+			await this.projectAccessRepository.setProjectAccess(
+				connection.id,
+				updates.projectIds,
+				'secretsProviderConnection:user',
+			);
```
alielkhateeb: I have a concern here, `updateConnection` is called whether it happened from a project scope or a global scope, yet we always "downgrade" the role to `secretsProviderConnection:user` is that intentional?

If I understand the code correctly, a project owned connection that is to be updated will not be owned by the project anymore

packages/cli/src/modules/external-secrets.ee/secrets-providers-project.controller.ee.ts
```diff
@@ -109,25 +129,47 @@ export class SecretProvidersProjectController {
...
 	async deleteConnection(
-		_req: AuthenticatedRequest,
+		req: AuthenticatedRequest,
 		res: Response,
 		@Param('projectId') projectId: string,
 		@Param('providerKey') providerKey: string,
 	) {
 		this.logger.debug('Deleting connection for project', { projectId, providerKey });
+
+		await this.accessCheckService.assertConnectionAccess({
```
alielkhateeb, line 166: Just a question for my own understanding, why do we have the decorator `@ProjectScope('externalSecretsProvider:delete')` as well as the `assertConnectionAccess` call with the same `requiredScope`? Same question applies to all other occurrences of `assertConnectionAccess`

ireneea: The decorator checks the user access, and the custom method checks the project access

The decorator will use the user role to see if the have global scope, if not it will check if the user project role grant them the scope. 

The custom method will check the relation between the project and the vault. For example a project could use a global vault but it's not allowed for its users to edit the vault

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
ireneea: Yeah that's a good point. It's only possible to set the project ids via the global endpoint, and the idea is that if the instance admin shares a vault with a project it's always shared with a 'user' role.

Let me check if I can there could be a better way to make this more explicit

ireneea: @alielkhateeb I've split the logic into two methods now `updateGlobalConnection` and `updateProjectConnection`

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -104,11 +122,43 @@ export class SecretsProvidersConnectionsService {
...
+				.filter((e) => !desiredProjectIds.has(e.projectId))
+				.map((e) => e.projectId);
+
+			// Add access for newly added projects with user role
+			// Existing projects keep their current role (e.g. owner)
+			const entriesToAdd = updates.projectIds
+				.filter((id) => !existingProjectIds.has(id))
+				.map((projectId) => ({
+					projectId,
+					role: 'secretsProviderConnection:user' as SecretsProviderAccessRole,
```
alielkhateeb: `as const` would provide better type-safety here because with type cast it would accept any value (e.g. `foo:bar`) and bypass the TS error assuming it is a `SecretsProviderAccessRole`. With `as const` TS will complain if it wasn't a possible value of `SecretsProviderAccessRole`
```suggestion
					role: 'secretsProviderConnection:user' as const,
```

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -173,6 +216,14 @@ export class SecretsProvidersConnectionsService {
 		return connection;
 	}
 
+	private async findConnectionOrFail(providerKey: string): Promise<SecretsProviderConnection> {
+		const connection = await this.repository.findOne({ where: { providerKey } });
+		if (!connection) {
+			throw new NotFoundError(`Connection with key "${providerKey}" not found`);
+		}
+		return connection;
+	}
```
alielkhateeb, line 225: This is really nice 👏

packages/frontend/editor-ui/src/features/integrations/secretsProviders.ee/composables/useConnectionModal.ee.ts
```diff
@@ -7,12 +7,13 @@ import { useRBACStore } from '@/app/stores/rbac.store';
...
 import { i18n } from '@n8n/i18n';
 import type { Scope } from '@n8n/permissions';
+import { getResourcePermissions } from '@n8n/permissions';
 import { useProjectsStore } from '@/features/collaboration/projects/projects.store';
 import type { ProjectSharingData } from '@/features/collaboration/projects/projects.types';
 import { isComponentPublicInstance } from '@/app/utils/typeGuards';
 import { useSettingsStore } from '@/app/stores/settings.store';
 
-export type ConnectionProjectSummary = { id: string; name: string };
+export type ConnectionProjectSummary = { id: string; name: string; role?: string };
```
alielkhateeb: Should we restrict the role type to a string union instead of any string?

ireneea: Nice one, I've updated the api types directly instead of using this inline type 👍

alielkhateeb (approved):
Really nice 🚀 
Tested and working as expected 👏

PR #26675, opened 2026-03-06
fix(core): Don't fail 2.x migration for workflows without history record

konstantintieber:
## Summary

Disclaimer: this fix is fully AI generated.

### Root Cause
workflow_entity.versionId and workflow_history can become out of sync due to silent error handling in the saveVersion() method (packages/cli/src/workflows/workflow-history/workflow-history.service.ts):

```ts
try {
    await repository.insert({...});
} catch (e) {
    const error = ensureError(e);
    this.logger.error(`Failed to save workflow history version...`, { error });
    // No rethrow — silently continues
}
```
When a workflow is saved with node/connection changes, n8n:

Generates a new versionId
Calls saveVersion() — which inserts into workflow_history
Updates workflow_entity.versionId
If step 2 fails (e.g. due to a transient DB error, a constraint violation, disk pressure, etc.), the error is swallowed. Step 3 still executes. The result: workflow_entity.versionId points to a UUID that has no matching record in workflow_history.

This is a known recurring issue — the codebase already contains two separate backfill migrations (1762763704614-BackfillMissingWorkflowHistoryRecords and 1765448186933-BackfillMissingWorkflowHistoryRecords) specifically to repair this out-of-sync state.

When ActivateExecuteWorkflowTriggerWorkflows1763048000000 later runs and tries to set activeVersionId = workflow.versionId, it hits the FK constraint because that versionId was never persisted to workflow_history.

### Fix Description
The else branch (no nodes modified) in ActivateExecuteWorkflowTriggerWorkflows blindly sets activeVersionId = workflow.versionId without verifying the versionId exists in workflow_history. Before performing that update, a workflow_history record must be conditionally inserted for the workflow's current versionId — only if one does not already exist. This is the same defensive pattern used in the two existing backfill migrations. The fix uses INSERT ... SELECT ... WHERE NOT EXISTS which is compatible with both PostgreSQL and SQLite.

The if (nodesModified) branch is not affected — it always generates a fresh versionId and creates the history record itself before touching workflow_entity.

https://linear.app/n8n/issue/N8N-9775/

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
+++ b/packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
@@ -189,6 +189,21 @@ export class ActivateExecuteWorkflowTriggerWorkflows1763048000000 implements Irr
 						},
 					);
 				} else {
+					// Ensure a workflow_history record exists for this versionId.
+					// Imported workflows may have a versionId that was never recorded in this instance's history.
+					await runQuery(
+						`INSERT INTO ${historyTableName} (${versionIdColumn}, ${workflowIdColumn}, ${authorsColumn}, ${nodesColumn}, ${connectionsColumn}, ${createdAtColumn}, ${updatedAtColumn}) SELECT :versionId, :workflowId, :authors, :nodes, :connections, :createdAt, :updatedAt WHERE NOT EXISTS (SELECT 1 FROM ${historyTableName} WHERE ${versionIdColumn} = :versionId)`,
+						{
+							versionId: workflow.versionId,
+							workflowId: workflow.id,
+							authors: 'system migration',
+							nodes: JSON.stringify(nodes),
+							connections: workflow.connections,
+							createdAt: new Date(),
+							updatedAt: new Date(),
+						},
+					);
+
 					// No nodes modified, just activate with existing versionId
 					await runQuery(
 						`UPDATE ${tableName} SET ${activeColumn} = :active, ${activeVersionIdColumn} = :versionId WHERE ${idColumn} = :id`,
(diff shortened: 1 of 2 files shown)


packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
```diff
@@ -189,6 +189,21 @@ export class ActivateExecuteWorkflowTriggerWorkflows1763048000000 implements Irr
...
 					);
 				} else {
+					// Ensure a workflow_history record exists for this versionId.
+					// Imported workflows may have a versionId that was never recorded in this instance's history.
+					await runQuery(
+						`INSERT INTO ${historyTableName} (${versionIdColumn}, ${workflowIdColumn}, ${authorsColumn}, ${nodesColumn}, ${connectionsColumn}, ${createdAtColumn}, ${updatedAtColumn}) SELECT :versionId, :workflowId, :authors, :nodes, :connections, :createdAt, :updatedAt WHERE NOT EXISTS (SELECT 1 FROM ${historyTableName} WHERE ${versionIdColumn} = :versionId)`,
+						{
+							versionId: workflow.versionId,
+							workflowId: workflow.id,
+							authors: 'system migration',
```
konstantintieber, line 199: Not sure if we want to add an author like "system migration" here.

packages/@n8n/db/src/migrations/common/1763048000000-ActivateExecuteWorkflowTriggerWorkflows.ts
```diff
@@ -189,6 +189,21 @@ export class ActivateExecuteWorkflowTriggerWorkflows1763048000000 implements Irr
 						},
 					);
 				} else {
+					// Ensure a workflow_history record exists for this versionId.
```
geemanjs, line 192: Will we cause problems with a system that has already upgraded here? 

In tools I've used in the Java world they store a checksum of the migration file contents in the database - and if that checksum changes it gets itself in a right tizz as it knows the "database" is out of sync with the historical migrations. 

I kind of hope what we do here is less clever than that 👀

konstantintieber:
Closing this one since https://github.com/n8n-io/n8n/pull/22974 already added a backfill migration for workflows that don't have a corresponding row in the `workflow_history` table.

PR #26684, opened 2026-03-06
feat(editor): Add opt-in toggle for external secrets system roles

ireneea:
## Summary

image

- Add backend settings endpoint to enable/disable external secrets scopes for system roles (admin/editor)
- Add `systemRolesEnabled` setting that dynamically syncs project role scopes via `AuthRolesService`
- Add toggle in External Secrets settings page (visible when role-based access is enabled); disabling prompts confirmation
- Expose `systemRolesEnabled` in frontend settings via module settings

https://linear.app/n8n/issue/LIGO-312

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/external-secrets.ee/external-secrets-settings.controller.ee.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	@Post('/')
+	@GlobalScope('externalSecretsProvider:update')
+	async updateSettings(
+		_req: Request,
+		_res: Response,
+		@Body body: UpdateExternalSecretsSettingsDto,
+	) {
+		if (body.systemRolesEnabled !== undefined) {
```
geemanjs: The value here could be `null` and we would pass that through 🤔? Or does the @Body flag coerce/force it to be a `boolean`?  

Reading the service if null came through it would get NullPointerException

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.service.ee.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	async setSystemRolesEnabled(enabled: boolean): Promise<void> {
+		await this.settingsRepository.upsert(
+			{
+				key: EXTERNAL_SECRETS_SYSTEM_ROLES_ENABLED_SETTING.key,
+				value: enabled.toString(),
+				loadOnStartup: true,
+			},
+			['key'],
+		);
```
geemanjs, line 22: This is such an annoying thing to have to do and such a "gotcha" 😆 . I'm guessing we do this because at startup time we overwrite the values? So need to check the setting and make sure the roles are added 🤦

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.controller.ee.ts
```diff
@@ -0,0 +1,45 @@
...
+				res,
+				new ForbiddenError('Role-based access for external secrets is not enabled'),
+			);
+			return;
+		}
+		next();
+	}
+
+	@Post('/')
+	@GlobalScope('externalSecretsProvider:update')
```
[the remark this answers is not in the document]
ireneea: The license check is controlled at the module level: See `packages/cli/src/modules/external-secrets.ee/external-secrets.module.ts`

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.controller.ee.ts
ireneea: That's because there is already a zod validation happening based on `UpdateExternalSecretsSettingsDto` where `systemRolesEnabled` is defined as an optional boolean. 

But actually this seems totally unnecessary in this context, I'm going to make the property required 👍

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.service.ee.ts
ireneea: Actually I just checked and we don't need need to set `loadOnStartup` to true this is only needed when we want to have the settings be part of `config`

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.controller.ee.ts
geemanjs: I wasn't sure if `null` was allowed in an optional - I think that makes sense 👍

packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -134,11 +135,35 @@ export class AuthRolesService {
 		return scopes;
 	}
 
+	private async getExternalSecretsSystemRolesScopes(
```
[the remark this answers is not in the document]
ireneea: added the unit tests 👍

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.service.ee.ts
geemanjs: I thought the roles get overridden at startup?

Given we're adding the roles into the database below here why do we need the setting change at all? 🤔

geemanjs: Maybe i'm getting confused with how roles operate here 👀

ireneea: You're correct the roles and scope get synced on startup and this happens in `AuthRolesService`. In our case we've added the `getExternalSecretsSystemRolesScopes` to deal with our use case. 

But that has nothing to do with the `loadOnStartup`. The `loadOnStartup` flag is specifically used to add the setting to the `config` which can be re-used in other places. But we don't need that because we directly read the setting row in our `getExternalSecretsSystemRolesScopes` we don't rely on the config

geemanjs: Ahh ok.. 

But won't our "setting" of the "role" scopes get removed at startup?

```typescript
		for (const [roleSlug, scopes] of Object.entries(roleScopeMap)) {
			if (enabled) {
				await this.roleService.addScopesToRole(roleSlug, scopes);
			} else {
				await this.roleService.removeScopesFromRole(roleSlug, scopes);
			}
		}
	}
```

ireneea: It will depend on the value of the setting stored in the db. 

If at startup it's enabled the scopes will be added to the roles if they are not already set
if at startup it's disabled, the scopes will be removed from the roles if they had them

Same thing at run time:

When enabled,  the scopes will be added to the roles
When disabled,  the scopes will be removed from the roles


It's all driven by the value in the settings row regardless if it's at startup or the setting is changed at run time

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -3043,6 +3043,15 @@
 	"settings.externalSecrets.provider.closeWithoutSaving.confirm": "Keep editing",
 	"settings.externalSecrets.docs": "https://docs.n8n.io/external-secrets/",
 	"settings.externalSecrets.docs.use": "https://docs.n8n.io/external-secrets/#use-secrets-in-n8n-credentials",
+	"settings.externalSecrets.systemRoles.title": "Enable external secrets for project roles",
+	"settings.externalSecrets.systemRoles.description": "Allow project admins and editors to access external secrets within their projects.",
```
konstantintieber: This sounds like it could use more detail.
Project admins will not just be able to access external secrets within their projects, but also create new vaults (To be clear: I think that's a good thing, and something that shouldn't even be opt-in).

ireneea: updated the description based on PM feedback 👍

packages/cli/src/modules/external-secrets.ee/external-secrets-settings.service.ee.ts
ireneea: Put back `loadOnStartup: true`

geemanjs (dismissed):

packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -134,11 +135,35 @@ export class AuthRolesService {
...
+		roleSlug: string,
+		tx: EntityManager,
+	): Promise<string[]> {
```
konstantintieber, line 143: I think we should inform the IAM team, as owners of authorization, about this change.
As this is the first time we're extending conditional scope functionality based on settings in the `Settings` entity.

packages/@n8n/db/src/services/auth.roles.service.ts
```diff
@@ -150,6 +175,9 @@ export class AuthRolesService {
 		if (roleSlug === PROJECT_OWNER_ROLE_SLUG) {
 			scopes.push(...(await this.getPersonalOwnerSettingsScopes(tx)));
 		}
+
+		// External secrets system roles scopes
+		scopes.push(...(await this.getExternalSecretsSystemRolesScopes(roleSlug, tx)));
```
konstantintieber: The `updateScopesBasedOnSettings` method is called for any kind of role inside `AuthRolesService.syncRoles`.

The previously existing code inside `updateScopesBasedOnSettings` makes sure that that logic only runs for the `PROJECT_OWNER_ROLE_SLUG` project role and that's visible inside the implementation of `updatesScopesBasedOnSettings`.

As I read this method as it is now, I wondered whether the `getExternalSecretsSystemRolesScopes` just runs for every role.
Once I looked into its implementation, I found the `if (scopesForRole) {`, which is true if the role slug is present as property inside the `EXTERNAL_SECRETS_SYSTEM_ROLES_ENABLED_SETTING.roleScopeMap` object.

I think it would be nice if we extract that `scopesForRole` check into the `updateScopesBasedOnSettings`, so that as another developer reading that method, I can understand more quickly which roles will be affected.

konstantintieber (approved):

packages/@n8n/db/src/services/auth.roles.service.ts
ireneea: Done ✅

PR #26691, opened 2026-03-06
fix(editor): Show relevant node in workflow activation errors

elsmr:
## Summary

- If workflow activation fails because of an error in a node, display the relevant node
- Include error description in a collapsible "Show details" UI
- Make this error consistent with the workflow execution error
- Refactor `toast.showError` arguments to use options object

image
image

https://linear.app/n8n/issue/NODE-4466

## Review / Merge checklist

- [ ] Tests included.


packages/frontend/editor-ui/src/app/composables/usePushConnection/handlers/workflowFailedToActivate.ts
```diff
@@ -21,10 +21,17 @@ export async function workflowFailedToActivate(
...
-			interpolate: { newStateName: 'activated' },
-		}) + ':',
-	);
+	const node = data.nodeId ? workflowsStore.getNodeById(data.nodeId) : undefined;
+	const title = i18n.baseText('workflowActivator.showError.title', {
+		interpolate: { newStateName: 'activated' },
+	});
+	const nodeMessage = node
+		? i18n.baseText('workflowActivator.showError.nodeError', {
+				interpolate: { nodeName: node.name },
```
DawidMyslak: Tiny suggestion: this node-name resolution block (resolve `nodeId` → node → formatted message) is duplicated almost verbatim in `useWorkflowActivate.ts`. Might be nice to extract a small shared helper like `formatActivationNodeMessage(nodeId)` to keep the two paths in sync — totally non-blocking though, just a maintainability thought! 🙂

packages/frontend/editor-ui/src/app/composables/useToast.ts
```diff
@@ -145,21 +143,32 @@ export function useToast() {
...
+		e: Error | unknown,
+		title: string,
+		options?: { message?: string; description?: string },
+	) {
 		const error = e as NotificationErrorWithNodeAndDescription;
+		const message = options?.message;
+		const description = options?.description ?? error.description;
+		const nodeLine =
+			!message && error.node?.name
+				? `${i18n.baseText('workflowActivator.showError.nodeError', { interpolate: { nodeName: error.node.name } })}<br/>`
```
DawidMyslak: Just a thought: the `nodeLine` fallback here (reading `error.node?.name`) uses the activation-specific i18n key `workflowActivator.showError.nodeError`, but `showError` is the general-purpose error toast used across the whole app. In practice I think this path might be unreachable for activation errors (they arrive as `ResponseError` without a `node` property), so it may only fire for other error types that happen to have `.node`.

Might be worth either removing this fallback (since both activation callers already construct the node message explicitly) or using a more generic i18n key. Not blocking at all — the current behavior is correct, just flagging for future clarity! 😊

DawidMyslak (dismissed):

DawidMyslak (dismissed):

DawidMyslak (approved):

PR #26694, opened 2026-03-06
fix(core): Default property handling for HITL tool

alexander-gekov:
## Summary

* When a HITL tool (Gmail/Discord/etc) is added it will show 'Approval options' with the default options set.
* The check in workflow package helper `getSendAndWaitProperties` now checks if options are empty (including `{}`)
* For multiple values the user still can manually delete them
image

https://linear.app/n8n/issue/NODE-4268/show-approve-and-disapprove-by-default-for-hitl-tools

## Review / Merge checklist

- [x] Tests included. 


RomanDavydchuk (changes requested):
This PR does indeed make it show approve/disapprove by default for HITL tools, but it has a couple of issues:
1. Changing the `default` value of a node parameter is a breaking change. It's because if a user has configured the node and didn't change that parameter's value (i.e. left it as default), changing `default` will affect that node. If we do need to change the default value, we usually introduce a new node version and show the parameter with new default value for the new version only (using `displayOptions: { show: '@version': [{  _cnd: { gte: '<new version>' } }] }`) and show the parameter with old default value for older versions (using `displayOptions: { show: '@version': [{  _cnd: { lt: '<new version>' } }] }`). This case, however, is tricky because these parameters are shared among all of the nodes with HITL operation, each of which has different versions. I think this is the reason why it was mentioned that implementing this change would be tricky.
P.S. You can see how this happens in practice: create a workflow with a HITL node (does not have to be a tool), for example Telegram. Leave the "Approval options" as default. Publish the workflow. If you trigger it on `master` you should see a message with just "Approve" (current default). If, however, you trigger it on this branch, you should see a message with both "Approve" and "Decline" buttons (new default). But you didn't do any changes to the workflow, so this is a breaking change from the user's POV
2. Making the item in the collection non-deletable works for this use case, but I can imagine that there are nodes that have `fixedCollection` with default value, but it's totally valid to not have it filled in. I'm going to be honest, I'm not sure if there is such node _right now_, but it's possible. Assuming we are fine with this change, I think that we better check in with design whether it's a good change to the UX. And if it is, would be nice to not show the trash icon for the case when the item is not deletable, since it doesn't actually do anything

This change is a lot trickier than it initially looks, and to be honest I'm not sure what's the best solution for this myself. Let me know if you have any questions and we can discuss more ideas on this

PR #26697, opened 2026-03-06
fix(Airtable Node): Revert flattening output from search and get operations

krisn0x:
## Summary

In v2 of the Airtable node, we started flattening the output form the Search and Get operations. If the table contains an `id` column it will overwrite the built-in `id` of the airtable element which users might need for deletion operations. This PR brings back the old behavior which nests all table fields under `fields`,  while the built-in `id` remains outside.

https://linear.app/n8n/issue/NODE-4030/bug-airtable-node-search-operation-not-returning-element-id-rec

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/Airtable/v2/actions/record/get.operation.ts
+++ b/packages/nodes-base/nodes/Airtable/v2/actions/record/get.operation.ts
@@ -85,8 +85,7 @@ export async function execute(
 					[responseData] as IRecord[],
 					options.downloadFields as string[],
 				);
-				legacyFlattenRecordOutputs(itemWithAttachments, nodeVersion);
-				returnData.push(...itemWithAttachments);
+				returnData.push(...legacyFlattenRecordOutputs(itemWithAttachments, nodeVersion));
 				continue;
 			}
 
--- a/packages/nodes-base/nodes/Airtable/v2/actions/record/get.operation.ts
+++ b/packages/nodes-base/nodes/Airtable/v2/actions/record/get.operation.ts
@@ -8,11 +8,7 @@ import type {
 
 import { updateDisplayOptions, wrapData } from '../../../../../utils/utilities';
 import type { IRecord } from '../../helpers/interfaces';
-import {
-	flattenOutput,
-	legacyFlattenRecordOutputs,
-	processAirtableError,
-} from '../../helpers/utils';
+import { legacyFlattenOutput, processAirtableError } from '../../helpers/utils';
 import { apiRequest, downloadRecordAttachments } from '../../transport';
 
 const properties: INodeProperties[] = [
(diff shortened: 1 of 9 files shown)


packages/nodes-base/nodes/Airtable/v2/actions/record/get.operation.ts
```diff
@@ -80,14 +85,18 @@ export async function execute(
...
 			}
 
-			const executionData = this.helpers.constructExecutionMetaData(
-				wrapData(flattenOutput(responseData as IDataObject)),
-				{ itemData: { item: i } },
-			);
+			const output =
+				nodeVersion < 2.2
+					? flattenOutput(responseData as IDataObject)
+					: (responseData as IDataObject);
```
dlavrenuek: We should reuse the `legacyFlattenRecordOutputs` method here, to not split version check logic over multiple files

packages/nodes-base/nodes/Airtable/v2/actions/record/search.operation.ts
```diff
@@ -210,14 +210,14 @@ export async function execute(
...
-				returnData.push(...itemWithAttachments);
+				returnData.push(...legacyFlattenRecordOutputs(itemWithAttachments, nodeVersion));
 				continue;
 			}
 
 			let records = responseData.records;
 
 			records = (records as IDataObject[]).map((record) => ({
-				json: flattenOutput(record),
+				json: nodeVersion < 2.2 ? flattenOutput(record) : record,
```
dlavrenuek: Also here `legacyFlattenRecordOutputs` should be used

packages/nodes-base/nodes/Airtable/v2/helpers/utils.ts
```diff
@@ -90,3 +95,11 @@ export const flattenOutput = (record: IDataObject) => {
 		...(fields as IDataObject),
 	};
 };
+
+export function legacyFlattenRecordOutputs(
```
dlavrenuek: The original `flattenOutput` is still exported, which invites for it being used directly. Right now it is also only used with version checks that should use `legacyFlattenRecordOutputs` instead. I suggest removing `export` from `flattenOutput`

dlavrenuek: After writing other comments I did realize that this method works on an array of items while the old `flattenOutput` worked with record data directly which can be a reason why in some places `flattenOutput` is still called. I would still advice to have only one method where the version is checked and the format is adjusted. The easiest way would be to rename the old `flattenOutput` to `legacyFlattenOutput`, pass the version as a required parameter to it and let it return either the record or a flatted record. Otherwise `legacyFlattenRecordOutputs` could also be changed to work on the record itself rather than on an array

dlavrenuek (changes requested):

packages/nodes-base/nodes/Airtable/v2/helpers/utils.ts
krisn0x: makes sense, simplified it

dlavrenuek (approved):
🎸

PR #26706, opened 2026-03-06
feat(core): Delete secrets provider connections on project deletion

alielkhateeb:
## Summary

https://www.loom.com/share/fc7b50ab077c440781e5116389665173

- Implements \`deleteConnectionsByProjectId\` method in \`SecretsProvidersConnectionsService\` to remove all secrets provider connections associated with a project
- Integrates this cleanup into \`ProjectService.deleteProject\` to ensure connections are cleaned up when a project is deleted
- Adds API tests covering the connection deletion behavior during project deletion

### Test scenarios covered

- **Instance admin creates project + connection, project admin deletes project** → connection remains with \`isEnabled = false\` and access rows removed
- **Project-owned connection** (role = \`secretsProviderConnection:owner\`) → deleting the project removes both the connection and its access rows
- **Instance admin creates and deletes** → connection and access are both removed

https://linear.app/n8n/issue/LIGO-323

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/external-secrets.ee/__tests__/secrets-providers-connections.service.ee.test.ts
alielkhateeb, line 1: Reason why I created a new test file is because I think it's better isolation this way, instead of trying to adapt all test cases setup in the existing `project.api.test.ts` it's cleaner to split as it also sets up a different set of `endpointGroups`

packages/cli/test/integration/external-secrets/project-deletion.api.test.ts
```diff
@@ -0,0 +1,206 @@
...
+		const projectConnection = await connectionRepository.findOneByOrFail({
+			providerKey: 'project-owner-connection',
+		});
+		await projectAccessRepository.update(
+			{
+				projectId: project.id,
+				secretsProviderConnectionId: projectConnection.id,
+			},
+			{ role: 'secretsProviderConnection:owner' },
+		);
```
alielkhateeb: I believe we will allow the customers to do this in the future through the API, but for the sake of this PR, I had to manually update the database to reflect the scenario anticipated to happen in the future
Of course my expectation here is to let me know if this is wrong in any way 😂

packages/cli/src/modules/insights/__tests__/insights.service.integration.test.ts
alielkhateeb, line 1: All changes here are just to anchor the start date so that when the tests run close to end of day (like now) they wouldn't break, this will save us time when we have to ship a hot fix around midnight on our cloud some time in the future 😂 
[Failing run for reference](https://github.com/n8n-io/n8n/actions/runs/22785966225/job/66102895070?pr=26706)

packages/cli/src/modules/external-secrets.ee/__tests__/secrets-providers-connections.service.ee.test.ts
ireneea: Makes sense 👍

ireneea: I assume you're talking about this file `packages/cli/test/integration/external-secrets/project-deletion.api.test.ts`

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -363,6 +364,43 @@ export class SecretsProvidersConnectionsService {
...
+	 */
+	async cleanupConnectionsForProjectDeletion(projectId: string, user: User): Promise<void> {
+		const accessEntries = await this.projectAccessRepository.findByProjectId(projectId);
+		const providerKeysToSync = new Set<string>();
+
+		for (const access of accessEntries) {
+			providerKeysToSync.add(access.secretsProviderConnection.providerKey);
+
+			if (
+				hasGlobalScope(user, 'externalSecretsProvider:delete') ||
```
ireneea: I'll double check if we even need to have an access control check for this use case. We might want to delete the the connection owned by the project regardless of who is deleting the project. 

TBC...

packages/cli/test/integration/external-secrets/project-deletion.api.test.ts
```diff
@@ -0,0 +1,206 @@
...
+	});
+
+	test('project admin owns connection, deleting project removes both connection and access entries', async () => {
+		const instanceAdmin = await createAdmin();
+		const projectAdmin = await createMember();
+		const project = await createTeamProject('Project Admin Team', projectAdmin);
+
+		await testServer
+			.authAgentFor(instanceAdmin)
+			.post('/secret-providers/connections')
```
ireneea:
```suggestion
			.post(`/secret-providers/projects/${project.id}/connections`)
```

This will make the connection be owned by the project instead of relying on the repository 😉

packages/cli/test/integration/external-secrets/project-deletion.api.test.ts
ireneea: Added a comment above to showcase how to create the 'owner' link by default ☝️

packages/cli/src/modules/external-secrets.ee/__tests__/secrets-providers-connections.service.ee.test.ts
alielkhateeb: @ireneea Yes sorry, It was a late brain talking 😂

packages/cli/test/integration/external-secrets/project-deletion.api.test.ts
alielkhateeb: @ireneea Ah it's done? Nice, I'll adjust the code

alielkhateeb: @ireneea 3e2506426c563c96028556a863fae90384275aab

ireneea (dismissed):

packages/cli/test/integration/external-secrets/project-deletion.api.test.ts
```diff
@@ -0,0 +1,313 @@
...
+
+		await projectAccessRepository.update(
+			{
+				projectId: project.id,
+				secretsProviderConnectionId: connection.id,
+			},
+			{
+				role: 'secretsProviderConnection:owner',
+			},
+		);
```
alielkhateeb: @ireneea If I don't update manually here, test wouldn't make sense, I didn't figure out how owner access is created as [here](https://github.com/n8n-io/n8n/blob/e1fd9479b7c8066ed8b76c5531521e36d6ce8ce2/packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts#L72) we don't pass the role in any way

Please let me know what am I missing here

packages/cli/test/integration/external-secrets/project-deletion.api.test.ts
```diff
@@ -0,0 +1,313 @@
...
+
+		await projectAccessRepository.update(
+			{
+				projectId: project.id,
+				secretsProviderConnectionId: ownerConnection.id,
+			},
+			{
+				role: 'secretsProviderConnection:owner',
+			},
+		);
```
alielkhateeb: @ireneea Same goes here

packages/workflow/package.json
```diff
@@ -69,6 +69,7 @@
     "transliteration": "2.3.5",
     "xml2js": "catalog:",
     "zod": "catalog:",
-    "jsonrepair": "catalog:"
+    "jsonrepair": "catalog:",
+    "uuid": "catalog:"
```
alielkhateeb, line 73: This was causing issues in the tests, apparently this was missed in another PR :)

ireneea (approved):
Tested locally and works as expected 👍

PR #26743, opened 2026-03-08
feat: Add Save Execution node for runtime execution save override

shortstacked:
## Summary
- Adds a new **Save Execution** node that overrides the workflow-level "Save successful executions" setting at runtime
- Uses `ITaskMetadata` (`this.setMetadata()`) — a first-class typed channel, not `$execution.customData`
- Hooks check the override at the save decision point in both regular and scaling main paths
- Zero overhead for workflows that don't use the node (guard resolved once at hook creation time)

Related Request:
https://community.n8n.io/t/do-we-have-a-save-execution-node/14884

## Test plan
- [x] Unit tests for SaveExecution node (metadata set correctly, items pass through)
- [x] Unit tests for execution lifecycle hooks (force save, force discard, fallback — both regular and scaling main)
- [x] E2E Playwright tests via webhook triggering (force save overrides `none` setting, force discard overrides `all` setting)
- [x] Manual smoke test in UI


shortstacked:
Abandoned

PR #26767, opened 2026-03-09
fix(core): Use correct published nodes for workflow indexing

mfsiega:
## Summary

Use the correct published nodes for workflow indexing.

Today we incorrectly build the published index based on the draft nodes. This fixes that.

Also stop pulling columns we don't need - these are potentially very large.

https://linear.app/n8n/issue/CAT-2601/use-the-published-nodes-to-build-the-published-index

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/workflow-index/workflow-index.service.ts
```diff
@@ -100,8 +100,9 @@ export class WorkflowIndexService {
 					await this.updateIndexForDraft(workflow);
 				} else {
 					// We know activeVersionId is not null here because the finder only returns workflows
-					// that have a published version.
-					await this.updateIndexForPublished(workflow, workflow.activeVersionId!);
+					// that have a published version. Use the published version's nodes, not the draft nodes.
+					const publishedNodes = workflow.activeVersion?.nodes;
```
CharlieKolb, line 103: I guess this should be `workflow.activeVersion!.nodes` to reflect the comment?

CharlieKolb (dismissed):
LGTM

packages/cli/src/modules/workflow-index/workflow-index.service.ts
mfsiega: If we're checking publishedNodes (which should also never be null) I think I'd rather just remove the comment, which seems outdated?

tomi (approved):
LGTM 🚀

PR #26783, opened 2026-03-09
fix(core): Exclude disabled secret connections from autocomplete, resolution, and registry

ireneea:
## Summary

Excludes disabled external secret provider connections from credential autocomplete, credential resolution, and the provider registry. This prevents disabled integrations from appearing or being used in workflows.

## Testing

1. Manually set isEnabled to false in the db
2. Check that the secrets of that connections are no longer available for completion in the credential form 

https://linear.app/n8n/issue/LIGO-342

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/external-secrets.ee/secrets-providers-project.controller.ee.ts
```diff
@@ -145,15 +144,4 @@ export class SecretProvidersProjectController {
 		await this.connectionsService.getConnectionAccessibleFromProject(providerKey, projectId);
 		return await this.connectionsService.testConnection(providerKey, req.user.id);
 	}
-
-	@Post('/:projectId/reload')
-	@ProjectScope('externalSecretsProvider:sync')
-	async reloadConnectionSecrets(
```
ireneea, line 151: Removed this because it's not needed anymore. This was only added for the initial design where there was a "Refresh all" button, but this is no longer needed

konstantintieber: Thx for the hint

konstantintieber (approved):
Code looks good, did not test it manually (all my local checkouts right now are on 1.x)

PR #26785, opened 2026-03-09
fix(editor): Show version name instead of current changes label for named versions

alielkhateeb:
## Summary

https://www.loom.com/share/4be06b87afd74f4587677f6b75916ff3

When a workflow history version has a name (e.g. it was published/saved with a label), the version list was still showing \"Current Changes\" for the active version instead of its name. This fix prioritizes displaying the version name when available, falling back to \"Current Changes\" only for unnamed current versions.

## Review / Merge checklist

- [x] Tests included.


dariacodes (dismissed):
👍

dariacodes (approved):

PR #26788, opened 2026-03-09
fix(editor): Command bar wasn't finding any workflows

scdekov:
## Summary

There was a condition added on the backend related to scopes that was determining if the entity being fetched is workflow or not based on the fields being selected.

Background: https://github.com/n8n-io/n8n/pull/26248 changed this behavior.

https://linear.app/n8n/issue/ADO-4929/command-bar-doesnt-see-workflows

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/shared/commandBar/composables/useWorkflowNavigationCommands.ts
+++ b/packages/frontend/editor-ui/src/features/shared/commandBar/composables/useWorkflowNavigationCommands.ts
@@ -82,16 +82,7 @@ export function useWorkflowNavigationCommands(options: {
 			// Search workflows by name with minimal fields
 			const nameSearchPromise = workflowsListStore.searchWorkflows({
 				query: trimmed,
-				select: [
-					'id',
-					'name',
-					'active',
-					'versionId',
-					'ownedBy',
-					'parentFolder',
-					'isArchived',
-					'description',
-				],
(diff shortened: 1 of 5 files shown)


dariacodes (dismissed):
Thanks for fixing this! Missed this initially. Shall we patch this?

packages/frontend/editor-ui/src/features/shared/commandBar/composables/useWorkflowNavigationCommands.ts
```diff
@@ -82,7 +82,16 @@ export function useWorkflowNavigationCommands(options: {
 			// Search workflows by name with minimal fields
 			const nameSearchPromise = workflowsListStore.searchWorkflows({
 				query: trimmed,
-				select: ['id', 'name', 'active', 'ownedBy', 'parentFolder', 'isArchived', 'description'],
+				select: [
+					'id',
+					'name',
+					'active',
```
dariacodes: Why are we selecting `active` without `activeVersionId` here by the way? Could we refactor to not select `active` at all? We will remove this field at some point, so could be a nice prep. But for now it might be enough to just add `activeVersiondId` which will also make it of workflow type

scdekov: Good point, will update it. We were selecting active because the command bar only cares about this. Also that was added before publishing

dariacodes (approved):

PR #26789, opened 2026-03-09
fix(editor): Preserve resource mapper values during floating node navigation

scdekov:
## Summary

When navigating between connected Data Table nodes (or any nodes using resource mapper) via the NDV floating node arrows, the resource mapper field values were being cleared. This was a community-reported bug.

**Root cause:** The parent component (`ParameterInputList`) recomputes `dependentParametersValues` asynchronously via a `throttledWatch` (200ms). When switching nodes, the `ResourceMapper` component would see the stale dep values from the previous node transition to the correct values for the new node. Its dependency watcher mistook this for the user changing a dependent parameter and cleared all field values.

**Fix:**
1. Added `:key="node?.name"` on `ResourceMapper` to force a clean remount when switching nodes
2. Added a `watch(node)` that immediately nullifies `dependentParametersValues` when the active node changes, so the ResourceMapper mounts with `null` dep values. When the `throttledWatch` later recomputes the correct values, the watcher sees a `null → correctValue` transition which is naturally skipped by the existing `oldValue !== null` guard

**How to test:**
1. Create two Data Table nodes connected to each other
2. Configure both with different field IDs and fill in resource mapper values
3. Open the second node's NDV, use the floating navigation arrow to go to the first node
4. Verify the first node's values are preserved
5. Navigate back to the second node and verify its values are preserved

https://linear.app/n8n/issue/ADO-4918

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInputList.vue
+++ b/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInputList.vue
@@ -251,6 +251,20 @@ throttledWatch(
 	{ throttle: 200, immediate: true },
 );
 
+// When the active node changes (e.g. via floating node navigation arrows),
+// immediately nullify dependentParametersValues in the cached items.
+// The throttledWatch above will recompute them asynchronously, but until then
+// the ResourceMapper component would see stale dep values from the previous node.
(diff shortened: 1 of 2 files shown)


dariacodes (approved):
Seems to be working correctly now 🚀

PR #26798, opened 2026-03-09
feat(core): Allow manual configuration of HashiCorp Vault KV mount path and version

MarcL:
## Summary

Enterprise customers whose Vault tokens lack `sys/mounts` read access cannot use n8n's external secrets integration because auto-discovery fails. This PR adds optional `kvMountPath` and `kvVersion` settings to the HashiCorp Vault external secrets provider, allowing users to bypass `sys/mounts` auto-discovery entirely.

**Changes:**
- Added `kvMountPath` (string) and `kvVersion` (v1/v2 dropdown) fields to the Vault provider UI
- Extracted `discoverKvMounts()` to unify mount discovery — returns manual config or auto-discovered mounts with zero branching in `update()`
- Extracted `testSecretAccess()` to unify connection testing — validates the configured KV path directly instead of `sys/mounts` when manual path is set
- Added `normalizeKvPath()` helper to ensure trailing slash consistency
- Comprehensive test coverage: 16 tests covering auto-discovery, manual KV v1/v2 paths, path normalisation, 403/500 error handling

**How to test:**
1. Set up a Vault instance with a KV v2 store (e.g. `secret/`)
2. Create a token **without** `sys/mounts` access but with read/list on `secret/data/*` and `secret/metadata/*`
3. Configure Vault external secrets in n8n with the restricted token, setting KV Mount Path to `secret/` and KV Version to `v2`
4. Verify connection test succeeds and secrets are loaded
5. Verify that leaving KV Mount Path blank still uses auto-discovery as before

https://linear.app/n8n/issue/LIGO-90

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/providers/vault.ts
+++ b/packages/cli/src/modules/external-secrets.ee/providers/vault.ts
@@ -27,6 +27,10 @@ interface VaultSettings {
 	// AppRole
 	roleId: string;
 	secretId: string;
+
+	// Manual KV configuration (bypasses sys/mounts auto-discovery)
+	kvMountPath?: string;
+	kvVersion?: string;
 }
 
 interface VaultResponse<T> {
@@ -211,6 +215,31 @@ export class VaultProvider extends SecretsProvider {
 					authMethod: ['appRole'],
 				},
 			},
+		},
+
+		// Manual KV configuration
+		{
+			displayName: 'KV Mount Path (optional)',
+			name: 'kvMountPath',
+			type: 'string',
+			default: '',
+			required: false,
+			noDataExpression: true,
+			placeholder: 'e.g. secret/',
+			hint: 'Specify the KV engine mount path to skip sys/mounts auto-discovery. Leave blank to auto-detect.',
+		},
+		{
+			displayName: 'KV Version',
+			name: 'kvVersion',
+			type: 'options',
+			default: '2',
+			required: false,
+			noDataExpression: true,
+			options: [
+				{ name: 'v1', value: '1' },
+				{ name: 'v2', value: '2' },
+			],
+			hint: 'Only used when KV Mount Path is specified.',
 		},
 	];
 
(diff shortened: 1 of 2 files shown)


geemanjs (commented):
This PR follows the same pattern set before in the vault.ts file - but I don't recall seeing # in use to denote private methods elsewhere in the n8n codebase 👀 - I don't see it as a problem but thought it's worth mentioning

geemanjs (dismissed):

gambletan (commented):
This is a well-structured feature addition. The `#discoverKvMounts` abstraction cleanly separates manual vs auto-discovery paths, and the test coverage is comprehensive.

A few specific observations:

1. **`#normalizeKvPath` trailing slash** — Good defensive handling. Worth noting that this also handles the case where a user passes `secret` without a trailing slash, which is tested. However, if a user passes something like `secret//` (double trailing slash), the current logic wouldn't normalize that. Probably an edge case not worth handling, but a `trimEnd('/')` + append `/` pattern would be more robust.

2. **`kvVersion` default fallback** — In `#discoverKvMounts`, when `kvMountPath` is set, you fall back to `kvVersion ?? '2'`. The UI default is also `'2'`. This is consistent, but the `kvVersion` field in the UI has `required: false` — consider whether this should be `required: true` with a `displayOptions.show` conditional on `kvMountPath` being non-empty, so the user is prompted to choose a version when they specify a path.

3. **Error message consistency** — In the original code, the sys/mounts failure message was `"Couldn't list mounts but wasn't a permissions issue."` (missing "it"). The new code says `"Couldn't list mounts but it wasn't a permissions issue."` — good, this fixes a minor grammar issue.

4. **`#testSecretAccess` for KV v1** — For v1, the list URL is `${normalizedPath}?list=true` without the `metadata/` segment. This correctly mirrors the `update()` flow. The test for v1 validation (`should validate KV v1 path without metadata segment`) confirms this path works.

5. **Test helpers** — `vaultSettingsWithKvPath` and `tokenLookupResponse` are nice test utilities. Consider whether `tokenLookupResponse` could replace the inline token setup in the existing tests above for consistency.

Solid work — the feature enables a meaningful use case (restricted Vault tokens without `sys/mounts` access) with clear code and good test coverage.

packages/cli/src/modules/external-secrets.ee/providers/vault.ts
```diff
@@ -487,29 +570,10 @@ export class VaultProvider extends SecretsProvider {
...
-			}
-
-			return [true];
-		} catch (e) {
-			if (axios.isAxiosError(e)) {
-				if (e.code === 'ECONNREFUSED') {
+			return await this.testSecretAccess();
+		} catch (error) {
+			if (axios.isAxiosError(error)) {
+				if (error.code === 'ECONNREFUSED') {
```
geemanjs, line 576: Not directly related to this PR - but should we be better handling error responses from vault here - and pushing more detail back up to the user?

I think that is outside the scope of this PR - maybe we should raise a ticket?

geemanjs (approved):

packages/cli/src/modules/external-secrets.ee/providers/vault.ts
MarcL: Yes - agreed. I'll raise a ticket.

LIGO-367

PR #26811, opened 2026-03-10
perf(core): Use a Map() instead of a generic object for activeExecutions

OscarBataille:
## Summary

I performed a light refactor of active-executions.ts to:
1. Use a Map() instead of a generic object (that was used as a Map). This was done in order to avoid useless hidden class  transitions in V8 (https://v8.dev/blog/fast-properties), and avoid the corresponding Garbage Collection.
2. Remove unncessary assignments by directly iterating over the Map's keys when possible.

The performance gain is very small, but it is good practice.

None

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/permissions/src/utilities/combine-scopes.ee.ts
+++ b/packages/@n8n/permissions/src/utilities/combine-scopes.ee.ts
@@ -14,18 +14,21 @@ import type { Scope, ScopeLevels, MaskLevels } from '../types.ee';
  * }, { sharing: ['workflow:read'] });
  */
 export function combineScopes(userScopes: ScopeLevels, masks?: MaskLevels): Set<Scope> {
-	const maskedScopes: ScopeLevels = Object.fromEntries(
-		Object.entries(userScopes).map((e) => [e[0], [...e[1]]]),
-	) as ScopeLevels;
+	const result = new Set<Scope>();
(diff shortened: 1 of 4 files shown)


gambletan (commented):
Nice refactor — `Map` is the right data structure here and eliminates the prototype chain concerns of plain objects.

A couple of observations:

1. **Deleting from a Map during iteration** — In `shutdown()`, you iterate with `for (const [executionId, { status }] of this.activeExecutions)` and then call `this.activeExecutions.delete(executionId)` inside the loop body (via `stopExecution` or directly). While ES6 Maps do support safe deletion during `for...of` iteration (per spec), this relies on a subtlety that can surprise future contributors. A brief comment would help, or you could collect IDs to delete and batch them after the loop, similar to how `toCancel` is already used for the concurrency control removal.

2. **`getActiveExecutions()` improvement** — Since you're already using `Map`, you could simplify the loop with destructuring: the current code uses `for (const [id, data] of this.activeExecutions)` which is already clean, nice.

3. **Minor: the shutdown polling loop** — The `while (this.activeExecutions.size !== 0)` is cleaner than the old `Object.keys` approach since it avoids re-creating an array each iteration. Good improvement.

Overall this is a clean, straightforward migration. The behavioral semantics are preserved correctly. 👍

PR #26817, opened 2026-03-10
fix(HTTP Request Node): Improve invalid JSON field error messages

tomi:
## Summary

Improve HTTP Request V3 JSON validation errors by returning field-specific messages for body, query, and headers JSON inputs.
Refactor duplicated JSON parsing/validation code into a shared helper.
Add/extend tests to cover valid and invalid JSON parameter handling and update the workflow test fixture for the new error text.

Before
image

After
image

https://linear.app/n8n/issue/CAT-2576/better-error-message-if-param-is-not-valid-json-in-http-request-node

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/HttpRequest/V3/HttpRequestV3.node.ts
+++ b/packages/nodes-base/nodes/HttpRequest/V3/HttpRequestV3.node.ts
@@ -3,6 +3,7 @@ import type {
 	IBinaryKeyData,
 	IDataObject,
 	IExecuteFunctions,
+	INode,
 	INodeExecutionData,
 	INodeType,
 	INodeTypeBaseDescription,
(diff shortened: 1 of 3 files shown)


packages/nodes-base/nodes/HttpRequest/V3/HttpRequestV3.node.ts
```diff
@@ -1149,3 +1130,27 @@ export class HttpRequestV3 implements INodeType {
...
+function parseJsonParameter(
+	node: INode,
+	jsonString: string,
+	fieldName: string,
+	itemIndex: number,
+): IDataObject {
+	try {
+		JSON.parse(jsonString);
+
+		return jsonParse(jsonString);
```
garritfra: Is `JSON.parse(jsonString)` still necessary if we're returning `jsonParse(jsonString)` anyway?

tomi: Actually looking at `jsonParse` it behaves exactly the same as `JSON.parse` when no options are given. So good call, let's change it 👍

garritfra (approved):
LGTM! 🙂

PR #26836, opened 2026-03-10
fix(Help Scout Node): Fix pagination query param duplication and modifiedSince handling

### LINKED ISSUE #38104: Helpscout pagination and filtering issues
### Bug Description

There are two issues that are breaking helpscout workflows:

1. Pagination query param duplication — helpscoutApiRequestAllItems is passing query params on every request, including paginated ones where the _links.next.href URL already contains them. Helpscout throws errors when it gets the duplicate params.

2. modifiedSince option handling —The date picker doesn't format dates correctly when converting from UI filters in n8n to the actual API calls. It can return an array which is sent to Helpscout as-is rather than just getting the date.

### To Reproduce

1. Make a flow that uses helpscout's search node and request all items. It blows up.
2. Add a conversation or customer search and apply a modifiedSince filter. It blows up.

### Expected behavior

The nodes return results rather than failing.

### Debug Info

n/a

### Operating System

all

### n8n Version

2.x

### Node.js Version

all

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

mh-avant:
## Summary

Two related bugs in the Help Scout node:

1. **Pagination query param duplication** — `helpscoutApiRequestAllItems` was passing query params on every request, including paginated ones where the `_links.next.href` URL already contains them. This caused duplicate/conflicting params and unexpected API behavior. Fix: clear `qs` when following a pagination URI.

2. **`modifiedSince` option handling** — The date picker can return `modifiedSince` as an array. When passed through `Object.assign(qs, options)`, the array value was sent as-is to the API. Fix: extract the first element before assigning to `qs`, for both conversation and customer `getAll` operations.

## Review / Merge checklist

- [x] Tests included.

Fixes #38104


gambletan (commented):
Good catch on the pagination query param duplication — passing `qs` alongside a pagination URI that already contains those params would double them up.

A few observations:

1. **`GenericFunctions.ts` fix is clean** — The `const requestQuery: IDataObject = uri ? {} : query;` pattern is a clear and correct way to avoid the duplication. The comment explains the intent well.

2. **`modifiedSince` array handling** — The `Array.isArray(value) ? value[0] : value` pattern in `HelpScout.node.ts` handles the date picker returning an array, which is pragmatic. However, this same logic is duplicated in two places (conversation `getAll` at line ~245 and customer `getAll` at line ~366). Consider extracting a small helper like `normalizeModifiedSince(value)` to avoid the duplication and make it easier to maintain if the date picker behavior changes.

3. **`limit` handling in `helpscoutApiRequestAllItems`** — Looking at the existing code, the `limit` check (`if (limit && returnData.length >= limit) break;`) reads from `query.limit`. Since on paginated requests you now pass `requestQuery = {}`, the limit is still read from the original `query` parameter (not `requestQuery`), so limit behavior is preserved. Good.

4. **Test coverage is thorough** — The `GenericFunctions.test.ts` covers first request params, paginated empty params, limit behavior, and multi-page accumulation. Nice.

Overall this is a solid bug fix with good test coverage.

PR #26838, opened 2026-03-10
fix: MCP - Resolve webhook id during workflow creation and update

dariacodes:
## Summary

When workflows are created or updated via MCP tools, webhook nodes are missing `webhookId`. Without it, webhook registration falls back to a path built from `{workflowId}/{nodeName}/{path}`. Moreover, this fallback is unreachable for the default webhook trigger node name because `encodeURIComponent` produces `%20` for spaces in node names, but Express decodes the URL to a literal space, so they never match. As a result, webhooks created or edited through MCP were permanently 404.

This PR introduces a shared `resolveNodeWebhookId` utility and uses it in both MCP tools. It also refactors the frontend and AI workflow builder, which previously implemented similar logic, to use this utility and eliminate code duplication.

### How to test

Ask MCP client to create a workflow with a GET webhook trigger and to publish it. Check that production URL is not returning 404 anymore.

Closes [ADO-4932](https://linear.app/n8n/issue/ADO-4932/bug-mcp-workflow-create-and-update-webhook-id-is-not-set)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/mcp/tools/workflow-builder/create-workflow-from-code.tool.ts
+++ b/packages/cli/src/modules/mcp/tools/workflow-builder/create-workflow-from-code.tool.ts
@@ -1,5 +1,5 @@
 import { type User, type ProjectRepository, WorkflowEntity } from '@n8n/db';
-import { resolveNodeWebhook } from 'n8n-workflow';
+import { resolveNodeWebhookId } from 'n8n-workflow';
 import z from 'zod';
 
 import { MCP_CREATE_WORKFLOW_FROM_CODE_TOOL, CODE_BUILDER_VALIDATE_TOOL } from './constants';
@@ -127,7 +127,7 @@ export const createCreateWorkflowFromCodeTool = (
 			for (const node of newWorkflow.nodes) {
 				try {
 					const desc = nodeTypes.getByNameAndVersion(node.type, node.typeVersion);
-					resolveNodeWebhook(node, desc.description);
+					resolveNodeWebhookId(node, desc.description);
 				} catch {
 					// Node type not found, skip
 				}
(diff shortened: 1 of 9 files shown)


packages/workflow/src/node-helpers.ts
```diff
@@ -1101,6 +1102,19 @@ export function getNodeWebhookUrl(
...
+/**
+ * Assigns a webhookId to a node if its type has webhook definitions
+ * and the node doesn't already have one.
+ */
+export function resolveNodeWebhookId(
+	node: Pick<INode, 'webhookId'>,
+	nodeTypeDescription: Pick<INodeTypeDescription, 'webhooks'>,
+): void {
+	if (nodeTypeDescription.webhooks?.length && !node.webhookId) {
+		node.webhookId = uuid();
```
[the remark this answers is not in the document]
dariacodes: All existing consumers already mutate

packages/cli/src/modules/mcp/tools/workflow-builder/create-workflow-from-code.tool.ts
```diff
@@ -123,6 +124,15 @@ export const createCreateWorkflowFromCodeTool = (
 				meta: { ...workflowJson.meta, aiBuilderAssisted: true },
 			});
 
+			for (const node of newWorkflow.nodes) {
+				try {
+					const desc = nodeTypes.getByNameAndVersion(node.type, node.typeVersion);
+					resolveNodeWebhookId(node, desc.description);
+				} catch {
+					// Node type not found, skip
```
scdekov, line 132: Seems legit that we don't need this catch. `resolveNodeWebhookId` doesn't really throw if the type is not found

scdekov (approved):

packages/cli/src/modules/mcp/tools/workflow-builder/create-workflow-from-code.tool.ts
dariacodes: We need it, because `getByNameAndVersion` can throw and `resolveNodeWebhookId` depends on it. AI didn't like that we are silently catching errors, but they are all validation errors which we hope to have fixed before while using the validation tool

scdekov: Got it

PR #25147, opened 2026-02-02
feat(core): Add MCP queue mode and multi-main support

aalises:
demo: [https://www.loom.com/share/11695991d948464d83714929eb01d043](<https://www.loom.com/share/11695991d948464d83714929eb01d043>)<br>full writeup: [https://www.notion.so/n8n/MCP-Trigger-Queue-Mode-Multi-Main-Support-2fc5b6e0c94f80f6952eca821a250e03](<https://www.notion.so/n8n/MCP-Trigger-Queue-Mode-Multi-Main-Support-2fc5b6e0c94f80f6952eca821a250e03>)

## Summary

Added support for MCP Trigger and MCP Service in **queue mode** and **multi-main** deployments, enabling distributed execution of MCP tool calls while maintaining session state and reliable response delivery.

### Queue Mode Incompatibility

In queue mode, workflows execute on workers via Redis job queues. When a worker completes an MCP tool call, it has no direct connection to the MCP client—the response must travel back through the main process.

```mermaid
graph TB
    subgraph "Queue Mode Challenge"
        Client[MCP Client]
        Main[Main Process]
        Redis[(Redis Queue)]
        Worker[Worker]
    end

    Client -->|POST| Main
    Main -->|Enqueue Job| Redis
    Redis -->|Dequeue| Worker
    Worker -.->|??? How to respond ???| Client
```

### Multi-Main Load Balancing

With load balancing, POST requests may land on a different main than the one holding the SSE session.

```mermaid
graph TB
    subgraph "Multi-Main Challenge"
        Client[MCP Client]
        LB[Load Balancer]
        Main1[Main 1 - Has SSE Session]
        Main2[Main 2 - No Session]
    end

    Client -->|SSE Connect| Main1
    Client -->|POST Tool Call| LB
    LB -->|Round Robin| Main2
    Main2 -.->|Session not found!| Client
```

### Architecture Overview

```mermaid
flowchart TB
    subgraph "MCP Client"
        MC[AI Agent]
    end

    subgraph "Main Layer"
        M1[Main 1]
        M2[Main 2]
        MSM[McpServerManager]
    end

    subgraph "Message Bus"
        BQ[(Bull Queue - Jobs)]
        PS[(Redis PubSub - Responses)]
    end

    subgraph "Worker Layer"
        W1[Worker]
    end

    MC -->|1. Tool Call| M1
    M1 -->|2. Enqueue Job| BQ
    BQ -->|3. Dequeue| W1
[...]

PR #26876, opened 2026-03-11
fix(core): Skip task runner startup on main in offload mode

tomi:
## Summary

When running in queue mode with `OFFLOAD_MANUAL_EXECUTIONS_TO_WORKERS=true`, this change disables task runner startup in the `start` command.
Task runners used to be always enforced on main instances, due to MCP trigger node not supporting running the executions on workers. That was fixed in #25147 so we shouldn't need this anymore.

https://linear.app/n8n/issue/CAT-2596

## Review / Merge checklist

- [ ] Tests included.


mfsiega (approved):
LGTM.

PR #26889, opened 2026-03-11
feat(core): Add workflow execution duration histogram metric

shortstacked:
## Summary
- Adds `n8n_workflow_execution_duration_seconds` Prometheus histogram metric
- Enables PromQL-based latency percentiles (p50/p95/p99) decoupled from execution persistence
- Labels: `status` (success/failed), `mode` (manual/trigger/webhook/etc.), optional `workflow_id`
- 16 histogram buckets covering 5ms to 600s (fast webhooks to slow batch jobs)
- Gated by `N8N_METRICS_INCLUDE_WORKFLOW_EXECUTION_DURATION` (default: `true` when metrics enabled)

## PromQL examples
```promql
# p95 execution duration
histogram_quantile(0.95, rate(n8n_workflow_execution_duration_seconds_bucket[5m]))

# Average execution duration
rate(n8n_workflow_execution_duration_seconds_sum[5m]) / rate(n8n_workflow_execution_duration_seconds_count[5m])

# Per-workflow p95 (when workflow_id label enabled)
histogram_quantile(0.95, rate(n8n_workflow_execution_duration_seconds_bucket{workflow_id="123"}[5m]))
```

## Test plan
- [x] Unit tests: histogram registration, observation with correct labels, disabled behavior, missing data handling (8 tests)
- [x] Typecheck passes on `@n8n/config` and `cli` packages
- [ ] Integration: verify histogram appears in `/metrics` scrape output
- [ ] Verify no double-counting in queue mode (histogram observed on main only)

### DIFF
--- a/packages/cli/src/metrics/prometheus-metrics.service.ts
+++ b/packages/cli/src/metrics/prometheus-metrics.service.ts
@@ -10,7 +10,7 @@ import promBundle from 'express-prom-bundle';
 import { DateTime } from 'luxon';
 import { InstanceSettings } from 'n8n-core';
 import { EventMessageTypeNames, jsonParse } from 'n8n-workflow';
-import promClient, { type Counter, type Gauge } from 'prom-client';
+import promClient, { type Counter, type Gauge, type Histogram } from 'prom-client';
 import semverParse from 'semver/functions/parse';
 
 import { N8N_VERSION } from '@/constants';
@@ -37,6 +37,8 @@ export class PrometheusMetricsService {
 
 	private readonly gauges: Record<string, Gauge<string>> = {};
 
+	private readonly histograms: Record<string, Histogram<string>> = {};
+
 	private readonly prefix = this.globalConfig.endpoints.metrics.prefix;
 
 	private readonly includes: Includes = {
@@ -46,6 +48,8 @@ export class PrometheusMetricsService {
 			cache: this.globalConfig.endpoints.metrics.includeCacheMetrics,
 			logs: this.globalConfig.endpoints.metrics.includeMessageEventBusMetrics,
 			queue: this.globalConfig.endpoints.metrics.includeQueueMetrics,
+			workflowExecutionDuration:
+				this.globalConfig.endpoints.metrics.includeWorkflowExecutionDuration,
 			workflowStatistics: this.globalConfig.endpoints.metrics.includeWorkflowStatistics,
 		},
 		labels: {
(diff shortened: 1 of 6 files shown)


shortstacked:
@claude Review please

packages/cli/src/metrics/prometheus-metrics.service.ts
```diff
@@ -342,6 +347,43 @@ export class PrometheusMetricsService {
...
+		if (this.includes.labels.workflowId) labelNames.push('workflow_id');
+
+		this.histograms.workflowExecutionDuration = new promClient.Histogram({
+			name: this.prefix + 'workflow_execution_duration_seconds',
+			help: 'Workflow execution duration in seconds.',
+			labelNames,
+			buckets: [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30, 60, 120, 300, 600],
+		});
+
+		this.eventService.on('workflow-post-execute', ({ runData, workflow }) => {
```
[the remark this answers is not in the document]
shortstacked: Looking at execution-lifecycle-hooks.ts, the hook registration per mode tells the story:

getLifecycleHooksForScalingWorker (line 569): Does NOT call hookFunctionsWorkflowEvents — workers never emit workflow-post-execute
getLifecycleHooksForScalingMain (line 596): DOES call hookFunctionsWorkflowEvents (line 607) — the main process emits workflow-post-execute after the worker reports back via JobFinishedMessage

mfsiega (commented):
One general question - IIRC the "execution time" that gets reported today excludes time spent waiting. This is quite confusing and I think the simpler calculation here is better, but it will also be confusing to have two different metrics.

Is that something we thought about? I also have some memory that there was weird behaviours around waiting/resuming/subworkflows but that might be less relevant.

packages/cli/src/metrics/prometheus-metrics.service.ts
```diff
@@ -342,6 +347,43 @@ export class PrometheusMetricsService {
...
+			if (!runData?.stoppedAt) return;
+
+			const durationSeconds = (runData.stoppedAt.getTime() - runData.startedAt.getTime()) / 1000;
+			const labels: Record<string, string> = {
```
mfsiega, line 374: Just for my own understanding, why these labels and not others?

mfsiega (approved):
LGTM. I don't have so much context on our general story around metrics so I can't say how much sense this makes in the big picture, but it looks reasonable.

packages/cli/src/metrics/prometheus-metrics.service.ts
shortstacked: Since we want the histogram to measure wall clock duration of start -> stop. If we add in other labels like, failed/cancelled it means we would have to bucket those as well. So I was aiming for the smallest/simplest grouping.

PR #26893, opened 2026-03-11
feat(core): Make workflow index batch size configurable via env var

mfsiega:
## Summary

Expose the workflow indexing batch size as an environment variable.

The queries support a configurable batch size but it wasn't previously exposed. This can be a way to manage memory usage.

## Related Linear ticket
https://linear.app/n8n/issue/CAT-2599

## Review / Merge checklist
- [x] Tests pass
- [x] Follows n8n config conventions

### DIFF
--- a/packages/@n8n/config/src/configs/workflows.config.ts
+++ b/packages/@n8n/config/src/configs/workflows.config.ts
@@ -23,6 +23,10 @@ export class WorkflowsConfig {
 	@Env('N8N_WORKFLOWS_INDEXING_ENABLED')
 	indexingEnabled: boolean = true;
 
+	/** Number of workflows to process per batch during dependency indexing on startup. */
+	@Env('N8N_WORKFLOW_INDEX_BATCH_SIZE')
+	indexingBatchSize: number = 100;
+
 	/** Whether to use the workflow publication service. Still under development. */
 	@Env('N8N_USE_WORKFLOW_PUBLICATION_SERVICE')
 	useWorkflowPublicationService: boolean = false;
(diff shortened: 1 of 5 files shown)


packages/@n8n/config/src/configs/workflows.config.ts
```diff
@@ -23,6 +23,10 @@ export class WorkflowsConfig {
 	@Env('N8N_WORKFLOWS_INDEXING_ENABLED')
 	indexingEnabled: boolean = true;
 
+	/** Number of workflows to process per batch during dependency indexing on startup. */
+	@Env('N8N_WORKFLOW_INDEX_BATCH_SIZE')
+	indexingBatchSize: number = 100;
```
tomi: Maybe we should be very conservative about this and use something like 10 by default?

tomi (dismissed):
One comment but LGTM 👌 Remember to update docs

packages/@n8n/config/src/configs/workflows.config.ts
mfsiega: Seems reasonable.

packages/@n8n/config/test/config.test.ts
```diff
@@ -195,6 +195,7 @@ describe('GlobalConfig', () => {
 			callerPolicyDefaultOption: 'workflowsFromSameOwner',
 			activationBatchSize: 1,
 			indexingEnabled: true,
+			indexingBatchSize: 10,
```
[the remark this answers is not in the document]
mfsiega: That's true, but I don't think it's so terrible to make this change at the same time if we believe that's what we want the new batch size to be.

afitzek (approved):
👍

PR #26895, opened 2026-03-11
feat: Custom auth redacts values correctly

BGZStephen:
## Summary

This PR allows the JSON custom auth value to be redacted without being marked as a `password` in credential definitions, allowing the value to appear unmasked until save, at which point it is masked.

IAM-165

Before save:

image

After save: 

image

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/credentials/credentials.service.ts
```diff
@@ -777,7 +777,15 @@ export class CredentialsService {
 			return props;
 		};
 		const properties = getExtendedProps(credType);
-		return this.redactValues(copiedData, properties);
+		const redacted = this.redactValues(copiedData, properties);
+		// Custom Auth: mask JSON after save (not marked as secret; redacted by type + field)
+		if (credential.type === 'httpCustomAuth' && 'json' in redacted) {
```
guillaumejacquart: Maybe this should be inside the `redactValues` function ?

packages/workflow/src/constants.ts
```diff
@@ -14,6 +14,9 @@ export const CODE_EXECUTION_MODES = ['runOnceForAllItems', 'runOnceForEachItem']
 // Arbitrary value to represent an empty credential value
 export const CREDENTIAL_EMPTY_VALUE = '__n8n_EMPTY_VALUE_7b1af746-3729-4c60-9b9b-e08eb29e58da';
 
+// Value used when redacting sensitive credential data for the client (never sent decrypted)
+export const CREDENTIAL_BLANKING_VALUE = '__n8n_BLANK_VALUE_e5362baf-c777-4d57-a609-6eaf1f9e87f6';
```
guillaumejacquart, line 18: Should we remove the one in the cli package ?

guillaumejacquart (commented):
Small comments

packages/cli/src/credentials/credentials.service.ts
BGZStephen: Yeah, I was on the fence with this one, as the redactValues doesn't currently get passed in the credential type, but I don't have a strong opinion either way, let me move it there

packages/workflow/src/constants.ts
BGZStephen: let me take a look

guillaumejacquart (approved):

PR #26896, opened 2026-03-11
feat(core): Read from workflow_published_version for webhooks/triggers/pollers (no-changelog)

mfsiega:
## Summary

Read from the new workflow_published_version table when triggering a workflow.

This is behind a feature flag. The PR should be a functional no-op when the flag is off.

This makes that table the source-of-truth. It adds a database lookup for triggers and pollers which didn't previously exist (though we always had this for webhooks), but it makes it so they don't need to be un-published and re-published. It also allows for atomically updating multiple triggers.

In a subsequent PR we'll add a caching layer -- similar to what webhooks use today -- to make lookups faster.

https://linear.app/n8n/issue/CAT-2191

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/db/src/repositories/workflow-published-version.repository.ts
```diff
@@ -24,4 +24,19 @@ export class WorkflowPublishedVersionRepository extends Repository<WorkflowPubli
 		});
 		return record?.publishedVersionId ?? null;
 	}
+
+	/**
+	 * Loads the published version record with the related workflow entity
+	 * (including shared/project relations) and the workflow history version
+	 * (which contains the published nodes/connections).
+	 */
+	async getPublishedVersionWithRelations(workflowId: string) {
```
mfsiega: Should we describe the type we're returning?

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,96 @@
...
+	name: string;
+	nodes: INode[];
+	connections: IConnections;
+	staticData: IDataObject | undefined;
+	settings: IWorkflowSettings | undefined;
+	shared: SharedWorkflow[];
+}
+
+/**
+ * Service for loading the currently-published workflow data.
```
mfsiega: I think we want a comment to distinguish that this is the source of truth for what _actually_ executes in production, whereas the `workflow_entity` table reflects how the user has configured their workflow. Since workflow publication is asynchronous, these may differ briefly.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -460,6 +491,31 @@ export class ActiveWorkflowManager {
 		executeErrorWorkflow(workflowData, fullRunData, mode);
 	}
 
+	/**
+	 * Load fresh workflow data for execution. Re-reads the published version
```
mfsiega: I think the way this is documented highlights the differences from how it works today, but that's not really necessary. We can just call this "loadPublishedWorkflowData".

I think the fallback is interesting, but we should leave it as a todo. The purpose of this indirection to load the workflow data is to allow atomically updating all triggers for a workflow at once - the fallback defeats that. So for now, this should simply retrieve the latest nodes/connections/etc from the published workflow version table.

We should document that the workflow we pass in is used for everything else.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -680,11 +781,19 @@ export class ActiveWorkflowManager {
 				);
 			}
 
+			// When the flag is on, trigger/poller emit callbacks re-read the published
+			// version from the DB so they pick up updates without deactivate/reactivate.
+			// When the flag is off, they use the in-memory workflowData (same as before).
+			const resolveWorkflowData = this.workflowsConfig.useWorkflowPublicationService
+				? async () => await this.loadFreshWorkflowData(dbWorkflow)
+				: undefined;
```
mfsiega: It's unnecessarily confusing that we pass undefined here, and then use the default value. Let's make the argument required and always pass something.

packages/cli/src/workflows/workflow-published-data.service.ts
mfsiega, line 1: I'm missing tests for this file. Should there be some?

packages/cli/test/integration/workflows/workflow-published-data.service.test.ts
```diff
@@ -0,0 +1,127 @@
...
+		// Should have the alternate nodes, NOT the original activeVersion nodes
+		expect(result!.nodes).toEqual(
+			expect.arrayContaining([expect.objectContaining({ name: 'Alternate Node' })]),
+		);
+		expect(result!.nodes).not.toEqual(
+			expect.arrayContaining([expect.objectContaining({ type: 'n8n-nodes-base.scheduleTrigger' })]),
+		);
+	});
+
+	test('should fall back to activeVersion when published_version table has no record', async () => {
```
mfsiega: This is not the right behaviour. In the new world, gated by the feature flag, the published_version table is the source of truth.

packages/cli/test/integration/workflows/workflow-published-data.service.test.ts
```diff
@@ -0,0 +1,127 @@
...
+		const result = await workflowPublishedDataService.getPublishedWorkflowData(workflow.id);
+
+		// Should fall back to the activeVersion
+		expect(result).not.toBeNull();
+		expect(result!.nodes).toEqual(
+			expect.arrayContaining([expect.objectContaining({ type: 'n8n-nodes-base.scheduleTrigger' })]),
+		);
+	});
+
+	test('should pick up changes when published_version is updated without reactivation', async () => {
```
mfsiega: Let's get rid of this test for now, it isn't useful.

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Service for loading the currently-published workflow data from the
+ * `workflow_published_version` table.
+ *
+ * Callers should only use this service when the feature flag
+ * (`N8N_USE_WORKFLOW_PUBLICATION_SERVICE`) is enabled. When the flag is off,
+ * callers should continue using the legacy `activeVersion` relation directly.
+ *
+ * If no record exists in the published version table (e.g. for workflows
```
mfsiega: This is the wrong behaviour.

packages/cli/src/__tests__/active-workflow-manager.test.ts
```diff
@@ -138,19 +141,7 @@ describe('ActiveWorkflowManager', () => {
 			Object.assign(instanceSettings, { isLeader: true });
 		});
 
-		test('should use active version when calling executeErrorWorkflow on activation failure', async () => {
```
mfsiega, line 252: Why are there changes in these tests? These should be no-op, right? Since the feature flag is off?

packages/cli/src/__tests__/active-workflow-manager.test.ts
```diff
@@ -261,11 +240,15 @@ describe('ActiveWorkflowManager', () => {
...
 					mode,
 					activation,
+					async () => workflowData,
 				);
 				const context = getTriggerFunctions(workflow, node, additionalData, mode, activation);
 
 				context.emit(triggerData);
 
+				// emit is now async (loads fresh workflow data), so wait for promises to settle
+				await new Promise((resolve) => setTimeout(resolve, 0));
```
mfsiega, line 391: Why does this need to move? It was already below, wasn't it?

packages/cli/src/webhooks/__tests__/live-webhooks.test.ts
```diff
@@ -29,6 +34,8 @@ describe('LiveWebhooks', () => {
 	const webhookService = mock<WebhookService>();
 	const nodeTypes = mock<NodeTypes>();
 	const workflowStaticDataService = mock<WorkflowStaticDataService>();
+	const workflowsConfig = mock<WorkflowsConfig>({ useWorkflowPublicationService: true });
```
mfsiega: Since useWorkflowPublicationService is off by default, it seems like we should prefer to cover that code path? If we want specific tests for the new path, let's add them, but otherwise try to leave things unchanged where possible.

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -94,48 +107,75 @@ export class LiveWebhooks implements IWebhookManager {
...
+			ownerProjectId = workflowData.shared.find(
+				(share) => share.role === 'workflow:owner',
+			)?.projectId;
+			activeWorkflowData = { ...workflowData, nodes, connections };
 		}
 
-		const { nodes, connections } = workflowData.activeVersion;
-
-		// Create a clean workflowData object with only activeVersion nodes/connections
-		// This prevents any downstream code from accidentally using the draft nodes
```
mfsiega, line 122: Do we not still want this? Or does it just happen above for free?

packages/cli/src/workflows/__tests__/workflow-published-data.service.test.ts
```diff
@@ -0,0 +1,81 @@
...
+describe('WorkflowPublishedDataService', () => {
+	const workflowPublishedVersionRepository = mock<WorkflowPublishedVersionRepository>();
+	let service: WorkflowPublishedDataService;
+
+	beforeEach(() => {
+		jest.clearAllMocks();
+		service = new WorkflowPublishedDataService(mockLogger(), workflowPublishedVersionRepository);
+	});
+
+	test('should return published data when record exists', async () => {
```
mfsiega, line 18: Let's add a comment that this test just verifies that we hit the repository and return the data.

packages/cli/src/workflows/__tests__/workflow-published-data.service.test.ts
```diff
@@ -0,0 +1,81 @@
...
+		const result = await service.getPublishedWorkflowData('wf-1');
+
+		expect(result).not.toBeNull();
+		expect(result!.id).toBe('wf-1');
+		expect(result!.name).toBe('Published Name');
+		expect(result!.nodes).toBe(nodes);
+		expect(result!.connections).toBe(connections);
+	});
+
+	test('should use workflow name when published version has no name', async () => {
```
mfsiega: We should _always_ use the workflow name. The published version name is a different concept (like "v1", "v2", "draft 17 mar") from the workflow name (which is more like "Customer Notification Flow"). This seems like wrong behaviour.

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,61 @@
...
+		if (!record?.publishedVersion || !record.workflow) {
+			this.logger.warn(`Published version record not found for workflow "${workflowId}"`);
+			return null;
+		}
+
+		const { publishedVersion, workflow } = record;
+
+		return {
+			id: workflowId,
+			name: publishedVersion.name ?? workflow.name,
```
mfsiega: This is a bug. The name of the workflow is the name of the workflow, period. The version name is a different thing.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -302,13 +325,17 @@ export class ActiveWorkflowManager {
...
 				void this.workflowStaticDataService.saveStaticData(workflow);
-				const executePromise = this.workflowExecutionService.runWorkflow(
-					workflowData,
-					node,
-					data,
-					additionalData,
-					mode,
-					responsePromise,
+
+				const executePromise = resolveWorkflowData().then(
```
mfsiega, line 331: In the future I think we'll stop using a callback, and always directly call into the workflow published data service. So we should leave a comment that this indirection is temporary to allow switching based on the feature flag, and we'll directly use the workflow published data service in the future; so the code will get simpler.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -344,6 +371,7 @@ export class ActiveWorkflowManager {
 		additionalData: IWorkflowExecuteAdditionalData,
 		mode: WorkflowExecuteMode,
 		activation: WorkflowActivateMode,
+		resolveWorkflowData: () => Promise<IWorkflowBase>,
```
mfsiega, line 380: We should add a comment here as well that we use this callback so we can use a feature flag to switch between the current way of holding the workflow data in memory vs retrieving it via the workflow published data service - in the future, this parameter will go away.

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -44,15 +48,24 @@ export class LiveWebhooks implements IWebhookManager {
...
+				webhook.workflowId,
+			);
+			nodes = publishedData?.nodes;
+		} else {
+			const workflowData = await this.workflowRepository.findOne({
+				where: { id: webhook.workflowId },
+				select: ['nodes'],
+			});
+			nodes = workflowData?.nodes;
+		}
```
tomi: This seems quite independent block. It could be its own function right?

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,61 @@
...
+ * `workflow_entity` table (which reflects how the user has configured their
+ * workflow).
+ *
+ * Callers should only use this service when the feature flag
+ * (`N8N_USE_WORKFLOW_PUBLICATION_SERVICE`) is enabled. When the flag is off,
+ * callers should continue using the legacy `activeVersion` relation directly.
+ *
+ * TODO: Add a caching layer to avoid a DB lookup on every trigger/poll/webhook
+ * execution. The cache should be invalidated when the published version changes
+ * (e.g. from the outbox consumer).
```
tomi: Nice documentation 👌

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,61 @@
...
+	constructor(
+		private readonly logger: Logger,
+		private readonly workflowPublishedVersionRepository: WorkflowPublishedVersionRepository,
+	) {}
+
+	async getPublishedWorkflowData(workflowId: string): Promise<PublishedWorkflowData | null> {
+		const record =
+			await this.workflowPublishedVersionRepository.getPublishedVersionWithRelations(workflowId);
+
+		if (!record?.publishedVersion || !record.workflow) {
```
tomi, line 31: Is this a realistic outcome in some scenarios or something that is a bug? I.e. should we report an error here or not?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -94,48 +107,75 @@ export class LiveWebhooks implements IWebhookManager {
...
+		let nodes: INode[];
+		let connections: IWorkflowBase['connections'];
+		let activeWorkflowData: IWorkflowBase;
+		let workflowName: string;
+		let staticData: IWorkflowBase['staticData'];
+		let settings: IWorkflowBase['settings'];
+		let isActive: boolean;
+		let ownerProjectId: string | undefined;
+
+		if (this.workflowsConfig.useWorkflowPublicationService) {
```
tomi: This makes this method really long and hard to follow. Could we move some parts to separate methods?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -238,39 +240,59 @@ export class ActiveWorkflowManager {
...
-		if (!workflowData.activeVersion) {
-			throw new UnexpectedError('Active version not found for workflow', {
-				extra: { workflowId },
+		let nodes: INode[];
+		let connections: IWorkflowBase['connections'];
+		let name: string;
+		let staticData: IWorkflowBase['staticData'];
+		let settings: IWorkflowBase['settings'];
+
+		if (this.workflowsConfig.useWorkflowPublicationService) {
```
tomi: This addition makes this method really long. Could we move it to a separate method?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -680,11 +797,19 @@ export class ActiveWorkflowManager {
 				);
 			}
 
+			// When the flag is on, trigger/poller emit callbacks re-read the published
+			// version from the DB so they pick up updates without deactivate/reactivate.
+			// When the flag is off, they use the in-memory workflowData (same as before).
+			const resolveWorkflowData = this.workflowsConfig.useWorkflowPublicationService
+				? async () => await this.loadPublishedWorkflowData(dbWorkflow)
```
tomi: Did I understand correctly that we are always loading the _currently_ published version here? Shouldn't we lock the workflow version ID here to a specific one? Otherwise it might lead to cases where the workflow has been activated with version A, but when the trigger event happens the published version has changed to version B. I can see this being a problem when for example that specific trigger has been disabled/removed in version B. Am I understanding this correctly?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -527,14 +599,35 @@ export class ActiveWorkflowManager {
...
 			);
 
-			if (!dbWorkflow.activeVersion) {
-				throw new UnexpectedError('Active version not found for workflow', {
-					extra: { workflowId: dbWorkflow.id },
-				});
+			let errorNodes: INode[];
+			let errorConnections: IWorkflowBase['connections'];
+
+			if (this.workflowsConfig.useWorkflowPublicationService) {
```
tomi: Same comment about code readability here

packages/cli/src/active-workflow-manager.ts
```diff
@@ -622,23 +715,47 @@ export class ActiveWorkflowManager {
...
 
 		try {
-			if (['init', 'leadershipChange'].includes(activationMode) && !dbWorkflow.activeVersion) {
-				this.logger.debug(
-					`Skipping workflow ${formatWorkflow(dbWorkflow)} as it is no longer active`,
-					{ workflowId: dbWorkflow.id },
+			let nodes: INode[];
+			let connections: IWorkflowBase['connections'];
+
+			if (this.workflowsConfig.useWorkflowPublicationService) {
```
tomi: Same comment about long method here

tomi (commented):
Nice work, well documented 👏 

Most of the comments are mainly about adding more lines into already long methods. I know these methods are not that readable to begin with, but would be nice to aim to at least not make it worse.

And on a more serious side also one question about the behaviour with the flag on

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -200,6 +197,71 @@ export class LiveWebhooks implements IWebhookManager {
...
+		}
+
+		const { nodes, connections, workflow } = publishedData;
+		return {
+			nodes,
+			connections,
+			workflowName: publishedData.name,
+			staticData: publishedData.staticData,
+			settings: publishedData.settings,
+			isActive: true,
```
mfsiega: Let's leave a comment - since we're getting the data from the published workflow data, we know the workflow is active.

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -200,6 +197,71 @@ export class LiveWebhooks implements IWebhookManager {
...
+		}
+	}
+
+	private async loadWebhookExecutionData(workflowId: string): Promise<WebhookExecutionData> {
+		return this.workflowsConfig.useWorkflowPublicationService
+			? await this.loadFromPublishedVersion(workflowId)
+			: await this.loadFromActiveVersion(workflowId);
+	}
+
+	private async loadFromPublishedVersion(workflowId: string): Promise<WebhookExecutionData> {
```
mfsiega: Let's leave a comment: "This is the new path for the workflow publication service. Behind a flag, disabled by default."

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -200,6 +197,71 @@ export class LiveWebhooks implements IWebhookManager {
...
+			workflowName: publishedData.name,
+			staticData: publishedData.staticData,
+			settings: publishedData.settings,
+			isActive: true,
+			ownerProjectId: findOwnerProjectId(publishedData.shared),
+			activeWorkflowData: { ...workflow, nodes, connections },
+		};
+	}
+
+	private async loadFromActiveVersion(workflowId: string): Promise<WebhookExecutionData> {
```
mfsiega: Leave a comment: "This is the old path, before the workflow publication service. Currently the default."

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,76 @@
...
+	staticData: IDataObject | undefined;
+	settings: IWorkflowSettings | undefined;
+	shared: SharedWorkflow[];
+	/**
+	 * The underlying workflow entity loaded alongside the published version
+	 * (in the same query, with shared/project relations). Callers that need
+	 * more than the projected fields above should use this instead of
+	 * issuing a separate query.
+	 */
+	workflow: WorkflowEntity;
```
mfsiega: Are we loading the nodes and connections etc twice here? What about pinned data?

packages/cli/src/workflows/workflow-published-data.service.ts
mfsiega: I think this represents a case that should never happen. The only thing that should read this is a trigger firing a workflow, and a trigger should only *try* to read this if the flag is enabled. Similarly, we should stop the triggers *before* remove the record. So I'd say this is a "should never happen" and we should use the error reporter to report the error.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -500,6 +511,107 @@ export class ActiveWorkflowManager {
...
+		};
+	}
+
+	/**
+	 * Resolve the active nodes and connections for a workflow.
+	 * Returns `null` when no published data / no active version exists
+	 * (i.e. the workflow was deactivated between the DB query and now).
+	 * Callers decide how to handle `null` -- skip or throw.
+	 */
+	private async resolveActiveNodesAndConnections(
```
mfsiega: I'm not so sure about this. Previously it seems like this was a simple object access. Now it turns into a database query. Is it really necessary that we re-query the database here, given the usage?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -739,11 +857,25 @@ export class ActiveWorkflowManager {
 					);
 				}
 
+				// When the flag is on, trigger/poller emit callbacks re-read the published
+				// version from the DB so they pick up updates without deactivate/reactivate.
+				// When the flag is off, they use the in-memory workflowData (same as before).
+				//
+				// Note: we intentionally load the _latest_ published version at trigger/poll
```
mfsiega: Change this note as follows:

```
Note: we intentionally load the latest published version when the trigger fires so all triggers are updated at the same time without any downtime. The workflow publication service is responsible for ensuring that removed/disabled triggers in a new workflow version are deactivated before updating the published version.
```

packages/cli/src/active-workflow-manager.ts
mfsiega: Left a comment that aimed to clarify this, but it's definitely not simple and straightforward. We can discuss more if we want.

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,84 @@
...
+
+	async getPublishedWorkflowData(workflowId: string): Promise<PublishedWorkflowData | null> {
+		const record =
+			await this.workflowPublishedVersionRepository.getPublishedVersionWithRelations(workflowId);
+
+		// This should not happen: only triggers read from this service, and they
+		// only do so when the flag is on; the publication service stops triggers
+		// before deleting the record. If we hit this, we have a real bug.
+		if (!record?.publishedVersion || !record.workflow) {
+			this.errorReporter.error(
```
mfsiega, line 32: I think this should automatically log, maybe? Or if we don't want it to log, we have to pass a param to suppress it?

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,84 @@
...
+				new UnexpectedError('Published version record not found for workflow', {
+					extra: { workflowId },
+				}),
+			);
+			this.logger.warn(`Published version record not found for workflow "${workflowId}"`);
```
mfsiega: Should this be `error`, given we reported an error above?

packages/cli/src/__tests__/active-workflow-manager.test.ts
```diff
@@ -377,11 +382,16 @@ describe('ActiveWorkflowManager', () => {
 					additionalData,
 					mode,
 					activation,
+					async () => workflowData,
 				);
 				const context = getTriggerFunctions(workflow, node, additionalData, mode, activation);
 
 				context.emit(triggerData);
 
+				// Even with the flag off, emit goes through resolveWorkflowData().then(...)
```
mfsiega: I don't think we actually need this comment.

packages/cli/src/__tests__/active-workflow-manager.test.ts
```diff
@@ -417,11 +425,15 @@ describe('ActiveWorkflowManager', () => {
 					additionalData,
 					mode,
 					activation,
+					async () => workflowData,
 				);
 				const context = getTriggerFunctions(workflow, node, additionalData, mode, activation);
 
 				context.emit(triggerData, undefined, undefined, 'wf-1:node-1:1700000000000');
 
+				// resolveWorkflowData().then(...) adds a microtask tick before runWorkflow.
```
mfsiega: Nor this one.

packages/cli/src/__tests__/active-workflow-manager.test.ts
mfsiega: Let's delete this comment.

packages/cli/src/__tests__/active-workflow-manager.test.ts
mfsiega: Let's delete this comment.

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -24,6 +26,23 @@ import type {
 	WebhookRequest,
 } from './webhook.types';
 
+interface WebhookExecutionData {
```
mfsiega: What is this interface for? Why are we adding it now, and what does it represent? Does it belong in a `types.ts` file or similar somewhere? What's the convention?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -24,6 +26,23 @@ import type {
...
+	connections: IWorkflowBase['connections'];
+	workflowName: string;
+	staticData: IWorkflowBase['staticData'];
+	settings: IWorkflowBase['settings'];
+	isActive: boolean;
+	ownerProjectId: string | undefined;
+	activeWorkflowData: IWorkflowBase;
+}
+
+function findOwnerProjectId(
```
mfsiega: Is there a `helpers` or `utils` somewhere this would belong better?

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,81 @@
...
+import type { IConnections, IDataObject, INode, IWorkflowSettings } from 'n8n-workflow';
+
+export interface PublishedWorkflowData {
+	id: string;
+	name: string;
+	nodes: INode[];
+	connections: IConnections;
+	staticData: IDataObject | undefined;
+	settings: IWorkflowSettings | undefined;
+	shared: SharedWorkflow[];
```
mfsiega: Why are these in the top-level object? Do they not just come from the workflow entity?

packages/cli/src/workflows/workflow-published-data.service.ts
```diff
@@ -0,0 +1,81 @@
...
+	WorkflowPublishedVersionRepository,
+	type SharedWorkflow,
+	type WorkflowEntity,
+} from '@n8n/db';
+import { Service } from '@n8n/di';
+import { ErrorReporter } from 'n8n-core';
+import { UnexpectedError } from 'n8n-workflow';
+import type { IConnections, IDataObject, INode, IWorkflowSettings } from 'n8n-workflow';
+
+export interface PublishedWorkflowData {
```
mfsiega: Maybe a comment explaining what this interface represents? Then the lengthy comment belong can be removed, because it will be clear.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -308,6 +311,11 @@ export class ActiveWorkflowManager {
 		additionalData: IWorkflowExecuteAdditionalData,
 		mode: WorkflowExecuteMode,
 		activation: WorkflowActivateMode,
+		// Temporary: this callback lets us switch between reading from
```
mfsiega: Instead of saying "temporary", say "TODO(CAT-3202):" and below as well.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -583,8 +636,11 @@ export class ActiveWorkflowManager {
 				});
 			}
 
-			const { nodes, connections } = dbWorkflow.activeVersion;
-			const workflowForError = { ...dbWorkflow, nodes, connections };
+			const workflowForError = {
+				...dbWorkflow,
+				nodes: dbWorkflow.activeVersion.nodes,
+				connections: dbWorkflow.activeVersion.connections,
+			};
```
mfsiega: This looks like a pure style change that's unrelated to this PR? If so, revert.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -678,23 +734,22 @@ export class ActiveWorkflowManager {
...
-					`Skipping workflow ${formatWorkflow(dbWorkflow)} as it is no longer active`,
-					{ workflowId: dbWorkflow.id },
-				);
-
-				return added;
-			}
-
-			// Get workflow data from the active version
 			if (!dbWorkflow.activeVersion) {
+				if (['init', 'leadershipChange'].includes(activationMode)) {
```
mfsiega: This also looks like a pure style change, unrelated to the PR. Revert?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -177,7 +158,7 @@ export class LiveWebhooks implements IWebhookManager {
 				void WebhookHelpers.executeWebhook(
 					workflow,
 					webhookData,
-					activeWorkflowData, // Use activeWorkflowData instead of workflowData
```
mfsiega, line 180: Why are we losing this comment?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -46,15 +50,11 @@ export class LiveWebhooks implements IWebhookManager {
 	async findAccessControlOptions(path: string, httpMethod: IHttpRequestMethods) {
 		const webhook = await this.findWebhook(path, httpMethod);
 
-		const workflowData = await this.workflowRepository.findOne({
-			where: { id: webhook.workflowId },
-			select: ['nodes'],
-		});
+		const nodes = await this.getWorkflowNodes(webhook.workflowId);
```
mfsiega: This seems like a behavioural change - previously we were always reading the draft nodes, now we read the latest published version. Is that intentional? Why?

It does maybe seem like a bug -- live webhooks will use the access control options according to the latest draft version, not the published version -- but fixing it seems out of scope for this PR?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -98,46 +98,27 @@ export class LiveWebhooks implements IWebhookManager {
...
-		}
-
-		if (!workflowData.activeVersion) {
-			throw new NotFoundError(
-				`Active version not found for workflow with id "${webhook.workflowId}"`,
-			);
-		}
-
-		const { nodes, connections } = workflowData.activeVersion;
+		const { workflow: workflowEntity, version } = await this.loadWebhookExecutionData(
```
mfsiega: How about we leave the variable name unchanged, to minimize the diff?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -1,5 +1,6 @@
 import { Logger } from '@n8n/backend-common';
```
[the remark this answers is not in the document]
mfsiega: This is a pre-existing bug, that we always read the draft nodes. I opened https://linear.app/n8n/issue/CAT-3225/bug-findaccesscontroloptions-uses-draft-nodes-instead-of-active to fix it separately. If this PR lands before that fix, then we can include the new codepath in the fix as well.

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -200,6 +185,43 @@ export class LiveWebhooks implements IWebhookManager {
...
+	): Promise<{ workflow: WorkflowEntity; version: WorkflowHistory }> {
+		const publishedData =
+			await this.workflowPublishedDataService.getPublishedWorkflowData(workflowId);
+		if (publishedData === null) {
+			throw new NotFoundError(`Published version not found for workflow with id "${workflowId}"`);
+		}
+		return { workflow: publishedData.workflow, version: publishedData.publishedVersion };
+	}
+
+	// Old path, before the workflow publication service. Currently the default.
```
tomi: nit: Could be a JSDoc

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -200,6 +185,43 @@ export class LiveWebhooks implements IWebhookManager {
...
 
+	private async loadWebhookExecutionData(
+		workflowId: string,
+	): Promise<{ workflow: WorkflowEntity; version: WorkflowHistory }> {
+		return this.workflowsConfig.useWorkflowPublicationService
+			? await this.loadFromPublishedVersion(workflowId)
+			: await this.loadFromActiveVersion(workflowId);
+	}
+
+	// New path for the workflow publication service. Behind a flag, disabled by default.
```
tomi: nit: Could be a JSDoc

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -200,6 +185,43 @@ export class LiveWebhooks implements IWebhookManager {
...
+	private async loadFromPublishedVersion(
+		workflowId: string,
+	): Promise<{ workflow: WorkflowEntity; version: WorkflowHistory }> {
+		const publishedData =
+			await this.workflowPublishedDataService.getPublishedWorkflowData(workflowId);
+		if (publishedData === null) {
+			throw new NotFoundError(`Published version not found for workflow with id "${workflowId}"`);
+		}
+		return { workflow: publishedData.workflow, version: publishedData.publishedVersion };
```
tomi: `version` might be a bit vague even tho it is deductable from the context. Could we name it `publishedVersion`?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -137,7 +122,7 @@ export class LiveWebhooks implements IWebhookManager {
 			settings: workflowData.settings,
 		});
 
-		const ownerProjectId = workflowData.shared.find(
+		const ownerProjectId = workflowData.shared?.find(
```
tomi, line 125: Hmm shouldn't every workflow have at least one `shared` relation? If it doesn't, that would mean the workflow has not been shared into any project, not even someones personal one 🤔 I think it would be better to throw if that is the case, because then our invariants wouldn't hold. But since this has been the case already way before this, it's okay to leave it, jsut mentioning for thoroughness

packages/cli/src/active-workflow-manager.ts
```diff
@@ -370,15 +389,21 @@ export class ActiveWorkflowManager {
...
-						node,
-						data,
-						additionalData,
-						mode,
-						responsePromise,
-						deduplicationKey,
+				// TODO(CAT-3202): resolves workflow data via callback so we
+				// can feature-flag between in-memory data and the published data
+				// service. Once the flag is removed, we'll call the service directly.
+				const executePromise = resolveWorkflowData()
```
tomi, line 395: I wonder if this can cause issues in some cases 🤔  The `emit` is called by the `trigger` function, and that is tied to specific workflow version. I think there's now a possibility that the workflow version of trigger node might differ from the rest of the workflow, since we always load the workflow data from DB here. Ideally these should be synced: trigger and rest of the workflow always match the same version

tomi (commented):
One concern about whether we are executing the correct workflow version from now on

tomi (dismissed):

tomi (approved):

tomi (approved):

PR #26912, opened 2026-03-11
feat(editor): Display workflow, credential and data table dependencies

CharlieKolb:
## Summary

Expose resource dependencies on cards and data table details view

https://linear.app/n8n/issue/ADO-4916

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/api-types/src/schemas/dependency.schema.ts
```diff
@@ -0,0 +1,13 @@
+export type DependencyType = 'credentialId' | 'dataTableId' | 'workflowCall' | 'workflowParent';
+
+export type DependencyResourceType = 'workflow' | 'credential' | 'dataTable';
+
+export interface ResolvedDependency {
```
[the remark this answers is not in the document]
CharlieKolb: In this case it's also optional for the case where a user doesn't have access to it

packages/frontend/editor-ui/src/features/credentials/components/CredentialCard.vue
```diff
@@ -186,6 +191,13 @@ function moveResource() {
 		</div>
 		<template #append>
 			<div :class="$style.cardActions" @click.stop>
+				<DependencyPill
```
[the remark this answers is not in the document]
CharlieKolb: Imo this is a trade-off where we'd prefer an accurate count (which indicates a resource is unsafe to delete). Not showing a matching count in the dropdown would be misleading, and "leaking" the id seems preferred

packages/@n8n/api-types/src/schemas/dependency.schema.ts
```diff
@@ -0,0 +1,13 @@
+export type DependencyType = 'credentialId' | 'dataTableId' | 'workflowCall' | 'workflowParent';
```
dariacodes, line 1: I was quite confused to see `Id` in the first two and then generic ones in the second two. It mixes resource references with workflow relationship types, while we already have `DependencyResourceType` for resources. I wonder if we could separate resource type from dependency relation, e.g.:
```ts
resourceType: 'workflow' | 'credential' | 'dataTable'
relation: 'reference' | 'workflowCall' | 'workflowParent'
```
Or at least remove `Id`, though that would be a bit less accurate than the structure above.
Also `workflowCall` vs `workflowParent` feels a bit unfortunate naming-wise - parent implies child, while call would imply something like caller/callee.

CharlieKolb: the first three are taken from the workflow-dependency db schema which I think this should stay close to. I agree call vs parent is unfortunate, but `workflowCall` and `workflowCalle` seems too similar too

dariacodes: ah, I see... what about Child / Parent then?

CharlieKolb: Well I wanted to stay aligned with the workflow index dependency types where possible, do you think this should take priority?

packages/cli/src/modules/workflow-index/workflow-dependency-query.service.ts
```diff
@@ -0,0 +1,261 @@
...
+			}
+		}
+		return result;
+	}
+
+	async getResourceDependencies(
+		resourceIds: string[],
+		resourceType: DependencyResourceType,
+		user: User,
+		enrichResources = true,
```
dariacodes: Looks like this could be called `resolveNames` or are you planning to extend this later? I'd add an annotation to this to highlight that when it's false, only ids are returned and we do not check access to dependencies

packages/frontend/editor-ui/src/app/components/DependencyPill.vue
```diff
@@ -0,0 +1,217 @@
...
+	} else if (dep.type === 'workflowCall' || dep.type === 'workflowParent') {
+		const href = router.resolve({ name: VIEWS.WORKFLOW, params: { name: dep.id } }).href;
+		window.open(href, '_blank');
+	} else if (dep.type === 'dataTableId' && dep.projectId) {
+		const href = router.resolve({
+			name: DATA_TABLE_DETAILS,
+			params: { projectId: dep.projectId, id: dep.id },
+		}).href;
+		window.open(href, '_blank');
+	}
```
dariacodes: Shall we use switch here? It could improve readability

dariacodes (dismissed):
Works really well! 🚀 
Left a few minor comments, nothing blocking

packages/cli/src/modules/workflow-index/workflow-dependency-query.service.ts
CharlieKolb: Yeah this is kept generic as it may also include other fields in the future (e.g. projectId for data tables) Will add an annotation 👍

dariacodes (approved):

PR #26938, opened 2026-03-12
fix(core): Fix MCP validate_workflow tool output schema warnings

dariacodes:
## Summary

The `validate_workflow_code` tool declares its output schema as `warnings: string[]`, but the handler was passing `ValidationWarning` objects directly into `structuredContent`, which caused the -32602 error. It only happens when warnings are present.

This PR changes the output format to an object.

Closes [ADO-4935](https://linear.app/n8n/issue/ADO-4935/feature-mcp-validate-workflow-tool-errors-out-sometimes)

## Review / Merge checklist

- [ ] Tests included. 


nikhilkuria (dismissed):

packages/cli/src/modules/mcp/tools/workflow-builder/validate-workflow-code.tool.ts
```diff
@@ -75,7 +75,7 @@ export const createValidateWorkflowCodeTool = (
 			};
 
 			if (result.warnings.length > 0) {
-				response.warnings = result.warnings;
+				response.warnings = result.warnings.map((w) => w.message);
```
nikhilkuria: Does it help the MCP client to have info on Nodename and parameterPath as well?

dariacodes: Good point! I think the right fix would be to change the output format to object format. On it

nikhilkuria (approved):
Looking good 👍

PR #26939, opened 2026-03-12
ci: Reduce noisy Codecov PR comments and fix double bundle upload

shortstacked:
## Summary
- Suppress Codecov PR comments unless there is a coverage drop or uncovered patch (`require_changes`)
- Lower bundle change comment threshold from 100Kb to 50Kb
- Disable Codecov Vite plugin on release builds to prevent double uploads caused by `@vitejs/plugin-legacy`

## Test plan
These have to be done after it's merged to master and the config takes affect.
- [ ] Verify Codecov comments only appear on PRs with coverage drops or uncovered patches
- [ ] Verify bundle analysis still uploads on non-release CI builds

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,7 +8,8 @@ bundle_analysis:
 
 comment:
   require_bundle_changes: bundle_increase
-  bundle_change_threshold: 100Kb
+  bundle_change_threshold: 50Kb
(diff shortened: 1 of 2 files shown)


Cadiac (approved):
🙏 thanks for keeping these nice

PR #26943, opened 2026-03-12
feat: Add `@n8n/cli`: a client CLI to manage n8n from the terminal

aalises:
## Summary

Introduces `@n8n/cli` (`n8n-cli`), a lightweight client CLI for managing n8n instances entirely from the terminal. It talks to the n8n REST API and requires zero server-side dependencies — just an instance URL and an API key.

### What it covers

| Domain | Commands |
|--------|----------|
| **Workflows** | `list`, `get`, `create`, `update`, `delete`, `activate`, `deactivate`, `tags`, `transfer` |
| **Executions** | `list`, `get`, `retry`, `stop`, `delete` |
| **Credentials** | `list`, `get`, `schema`, `create`, `delete`, `transfer` |
| **Projects** | `list`, `get`, `create`, `update`, `delete`, `members`, `add-member`, `remove-member` |
| **Tags** | `list`, `create`, `update`, `delete` |
| **Variables** | `list`, `create`, `update`, `delete` |
| **Data Tables** | `list`, `get`, `create`, `delete`, `rows`, `add-rows`, `update-rows`, `upsert-rows`, `delete-rows` |
| **Users** | `list`, `get` |
| **Other** | `login`, `audit`, `source-control pull`, `config show/set-url/set-api-key`, `skill install` |

### Design principles

- **Zero dependencies** beyond `@oclif/core` — uses native `fetch`, no HTTP libraries or server stack
- **Composable output** — table, JSON, and ID-only formats; auto-switches to JSON when stdout is piped; built-in `--jq` filtering
- **Unix-friendly** — data to stdout, errors to stderr, meaningful exit codes, `--quiet` and `--no-header` flags for scripting
- **AI agent friendly** — self-documenting via `--help`, ships with a skill definition (`skill install`) that teaches Claude Code, Cursor, and Windsurf how to use the CLI
- **Config flexibility** — config file (`~/.n8n-cli/config.json`), environment variables (`N8N_URL`, `N8N_API_KEY`), or inline flags; resolved in that priority order

### Included

- Full source under `packages/@n8n/cli/src/` (54 command files + client, config, output helpers)
- VitePress documentation site (`docs/`)
- AI agent skill definition (`skills/n8n-cli/SKILL.md`)

Closes https://linear.app/n8n/issue/ADO-4966/feature-merge-the-first-production-version-of-n8n-cli


shortstacked:
> > Hey @aalises few thoughts on this, it looks great but just wanted to raise a few concerns.
> > **Why a CLI over the API directly?**
> > Everything here wraps existing API calls — what's the value-add over consumers just hitting the API? With a discovery endpoint (e.g. `GET /api/v1/discover`) that returns available endpoints scoped to the caller's permissions, any HTTP client or AI agent could introspect what's available at runtime. That would:
> > 
> > * Let agents know what they can do without burning tokens guessing or hitting 403s
> > * Always stay in sync with the API (no drift between CLI and server)
> > * Work for any consumer, not just this CLI
> > * Remove the need for a 268-line static skill file that has to be manually kept up to date
> > 
> > Genuine question — is there a use case this solves that the API + discovery can't?
> > **Testing**
> > This is 54 commands and an HTTP client with no tests. It'd be great to see at least:
> > 
> > * Unit tests on the HTTP client and output formatting
> > * Contract/integration tests against the API to catch drift — if an endpoint changes shape, consumers won't notice until something silently breaks
> > 
> > **Dependency footprint**
> > The PR describes "zero dependencies beyond `@oclif/core`" but oclif pulls in a significant transitive tree (~900 lines of lockfile changes). The existing CLI uses a lightweight custom stack (`yargs-parser` + `zod` + decorators). Was oclif a deliberate choice over extending the existing CLI infrastructure, or would a thinner approach work here?
> 
> Thanks for the review and concerns @shortstacked ! This was born as an experimentation/greenfield project. I've posted it here so ADORE can pick it up and take a look, so nothing is carved in stone 🗿
> 
> Regarding the `oclif`, we might as well use the lightweight custom stack. Just used what seemed the state-of-the-art way of doing semantic APIs.
> 
> And **Why a CLI over the API directly?** some thoughts that we discussed with @dariacodes, also general enough on why CLIs are favoured these days vs. MCPs
> 
> **1. Self-documenting via --help**: An AI agent can run n8n-cli --help or n8n-cli workflow --help and immediately learn every available operation, flag, and expected format. With raw API, the agent needs to either have the API docs in context or guess at endpoints/payloads.
> 
> **2. Pagination is handled for you:** paginate() auto-fetches all pages. With curl, the agent would need to manually parse nextCursor and loop. That's error-prone boilerplate every time.
> 
> **3. Composability with Unix pipes:** n8n-cli workflow list --active --format=id-only | xargs -I{} n8n-cli workflow deactivate {} vs. the equivalent multi-step curl + jq + loop. Agents are very good at chaining CLI commands but bad at managing multi-step HTTP flows with state.
> 
> **4. Output format control (--format=json|table|id-only)**: The agent picks the format that suits the task. id-only for piping, json for parsing, table for display. Raw API always returns the full JSON payload. Auth is configured once: n8n-cli config set-api-key — done. No need to thread X-N8N-API-KEY headers and base URLs through every single curl call.
> 
> **5. Error handling with exit codes + stderr:** Exit code 2 = auth failure, 1 = general error. Agents can branch on $?. Errors go to stderr, data to stdout — no contamination. With curl, the agent has to parse HTTP status codes from JSON response bodies.
> 
> **6. Universal agent support:** Every AI agent (Claude Code, Cursor, Windsurf, Copilot) can run bash. Not every agent supports MCP. And curl is universally available but verbose — the CLI is the sweet spot between accessibility and ergonomics.
> 
> so basically
> 
> * Output format negotiation (table/json/id-only/--jq)
> * Automatic pagination with limits
> * Pipe-friendly output (stdout/stderr separation, exit codes)
> * The compound operations that come from xargs-style chaining
> 
> The SKILL is added as an example/optimisation.

I wasn't talking about an MCP server, more so just to add something like "publicapi/v1/discover" which would return a list of endpoints the API token would have access to. That way it never drifts from the API on what is actually available and an agent wouldn't burn tokens trying CLI commands it doesn't have access to. The drift is probably one of the biggest concerns though because this couples the API and CLI together so they can drift.

This would mean that it's also portable, it's not tied to local only it can be inferred with across the wire so I could let Claude/Curser/Windsurf use any version of n8n remotely/locally without needing the extra overhead.

dariacodes (approved):

PR #26946, opened 2026-03-12
chore: Update CODEOWNERS with ci-admins

Matsuuu:
## Summary

Add n8n-io/ci-admins as the CODEOWNERS of .github/workflows and .github/scripts

https://linear.app/n8n/issue/CAT-2552

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,3 @@
 packages/@n8n/db/src/migrations/ @n8n-io/migrations-review
+.github/workflows @n8n-io/ci-admins
+.github/scripts @n8n-io/ci-admins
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1 +1,3 @@
 packages/@n8n/db/src/migrations/ @n8n-io/migrations-review
+.github/workflows @n8n-io/ci-admins
+.github/scripts @n8n-io/ci-admins
```
tomi, line 3: Should it be entire `.github/`? There's at least `.github/actions` that should be included

Matsuuu: Updated with actions and poutine-rules.

I don't want to own the whole directory as we are putting Claude instructions there too, as well as issue templates

shortstacked (approved):
I think the scope of this is correct, and it's better than what we have so far.

PR #26970, opened 2026-03-13
feat(Perplexity Node): Update Perplexity node for full API coverage

kesku:
## Summary

Update the Perplexity node from v1 (chat-only) to v2 with full Perplexity API coverage. The node now supports `version: [1, 2]` with `defaultVersion: 2`, preserving backward compatibility for existing v1 workflows.

### New Resources

- **Agent** (`POST /v1/responses`): Create responses using third-party models (OpenAI, Anthropic, Google, etc.) via Perplexity's Agent API. Supports dynamic model loading from `/v1/models`, system instructions, tools (`web_search`, `fetch_url`), structured output via JSON schema, fallback models, and reasoning configuration.
- **Search** (`POST /search`): Get raw, ranked web search results with filtering by date range, domain, language, recency, and country. Returns structured result objects with titles, URLs, snippets, and dates.
- **Embeddings** (`POST /v1/embeddings` and `POST /v1/contextualizedembeddings`): Generate vector embeddings for text using `pplx-embed-v1-4b` or `pplx-embed-v1-0.6b` models. Includes a contextualized embedding operation for document-chunk-aware embeddings.

### Chat Completions Enhancements

| Parameter | Description |
|-----------|-------------|
| `search_mode` | Web, Academic, or SEC filings search |
| `search_after_date_filter` / `search_before_date_filter` | Date range filters for search results |
| `last_updated_after_filter` / `last_updated_before_filter` | Last-updated date filters |
| `search_language_filter` | ISO 639-1 language codes |
| `image_domain_filter` / `image_format_filter` | Image result filtering |
| `enable_search_classifier` / `disable_search` | Search behavior control |
| `reasoning_effort` | Minimal, Low, Medium, High |
| `language_preference` | Response language preference |
| `stop` | Stop sequences |
| `response_format` | Structured output via JSON schema |
| `web_search_options` | Advanced search configuration |

### Bug Fixes & Cleanup

- **Fix `search_recency_filter` parameter name**: The v1 node was sending `search_recency` which is not a valid API parameter. Corrected to `search_recency_filter`.
- **Remove deprecated `sonar-reasoning` model**: This model returns HTTP 400 from the Perplexity API. Removed from the model dropdown.
- **Add `year` option** to search recency filter (was missing from v1).

### Testing

All endpoints verified against the live Perplexity API:
- `GET /v1/models` - Dynamic model list loading
- `POST /chat/completions` - All v1 and v2 parameters
- `POST /v1/responses` - Agent API with/without tools, structured output
- `POST /search` - With date, domain, language, and recency filters
- `POST /v1/embeddings` - Standard embeddings
- `POST /v1/contextualizedembeddings` - Context-aware embeddings
- Error handling verified for invalid models, missing parameters across all endpoints

Unit tests added for Agent, Search, and Embeddings resources using `NodeTestHarness` with nock mocks.

## Review / Merge checklist

- [x] Tests included.


packages/nodes-base/nodes/Perplexity/GenericFunctions.ts
```diff
@@ -39,3 +41,91 @@ export async function sendErrorPostReceive(
...
+			message: errorMessage,
+			description:
+				'Refer to the Embeddings API documentation at https://docs.perplexity.ai/api-reference/embeddings-post for valid parameters.',
+		});
+	}
+	return data;
+}
+
+export async function getModels(this: ILoadOptionsFunctions): Promise<INodeListSearchResult> {
+	const filter = (this.getFilterValue?.('search', '') ?? '') as string;
```
dlavrenuek: `this.getFilterValue` is not a valid property. The filter is passed to the `listSearch` as an optional parameter in  https://github.com/n8n-io/n8n/pull/26970/changes#diff-dd2d6575427a204aec5375050386b30ce8660932c6569763121eec95be5d566cR110


```suggestion
export async function getModels(this: ILoadOptionsFunctions, filter: string = ''): Promise<INodeListSearchResult> {
```

packages/nodes-base/nodes/Perplexity/Perplexity.node.ts
```diff
@@ -45,8 +56,60 @@ export class Perplexity implements INodeType {
...
+			...embeddings.description,
+			...search.description,
 		],
 	};
+
+	methods = {
+		listSearch: {
+			async getAgentModels(this: ILoadOptionsFunctions): Promise<INodeListSearchResult> {
+				return await getModels.call(this);
+			},
```
dlavrenuek: There should be no need to specify this method and call `getModels`, I suggest just doing


```suggestion
			getAgentModels,
```

and adjust the import at the top to 

```ts
import { getModels as getAgentModels } from './GenericFunctions';
```

Since `getModels` is not used anywhere than else I suggest simply renaming this method to `getAgentModels` to have consistent naming - now it is called with `getAgentModels` via the search endpoint but is named `getModels` in the code

packages/nodes-base/nodes/Perplexity/GenericFunctions.ts
```diff
@@ -39,3 +41,91 @@ export async function sendErrorPostReceive(
...
+		const error = (errorBody?.error ?? errorBody) as JsonObject;
+		const errorMessage =
+			typeof error.message === 'string'
+				? error.message
+				: (response.statusMessage ?? 'An unexpected issue occurred');
+
+		throw new NodeApiError(this.getNode(), errorBody, {
+			message: errorMessage,
+			description:
+				'Refer to the Agent API documentation at https://docs.perplexity.ai/api-reference/responses-post for valid parameters.',
```
dlavrenuek: The link redirects to https://docs.perplexity.ai/api-reference/agent-post, is it possible that the docs have been updated since these changes were implemented? I would recommend using the new url here.

packages/nodes-base/nodes/Perplexity/GenericFunctions.ts
```diff
@@ -39,3 +41,91 @@ export async function sendErrorPostReceive(
...
+		throw new NodeApiError(this.getNode(), errorBody, {
+			message: errorMessage,
+			description:
+				'Refer to the Search API documentation at https://docs.perplexity.ai/api-reference/search-post for valid parameters.',
+		});
+	}
+	return data;
+}
+
+export async function embeddingsErrorPostReceive(
```
dlavrenuek: The `agentErrorPostReceive`, `searchErrorPostReceive` and `embeddingsErrorPostReceive` are all identical except for the docs url, let's reduce the code duplication with a generic error handler. for example (I've cleaned up the code a bit with destructuging. Also no idea why status Code is converted to a string all over our code base, that should not be needed):

```ts
const createErrorHandler = (description: string) =>
	async function searchErrorPostReceive(
		this: IExecuteSingleFunctions,
		data: INodeExecutionData[],
		{ statusCode, body, statusMessage }: IN8nHttpFullResponse,
	): Promise<INodeExecutionData[]> {
		if (statusCode >= 400 && statusCode <= 599) {
			const errorBody = body as JsonObject;
			const error = (errorBody?.error ?? errorBody) as JsonObject;
			const errorMessage =
				typeof error.message === 'string'
					? error.message
					: (statusMessage ?? 'An unexpected issue occurred');

			throw new NodeApiError(this.getNode(), errorBody, {
				message: errorMessage,
				description,
			});
		}
		return data;
	};

export const agentErrorPostReceive = createErrorHandler(
	'Refer to the Agent API documentation at https://docs.perplexity.ai/api-reference/agent-post for valid parameters.',
);

export const searchErrorPostReceive = createErrorHandler(
	'Refer to the Search API documentation at https://docs.perplexity.ai/api-reference/search-post for valid parameters.',
);

export const embeddingsErrorPostReceive = createErrorHandler(
	'Refer to the Embeddings API documentation at https://docs.perplexity.ai/api-reference/embeddings-post for valid parameters.',
);
```

What do you think?

packages/nodes-base/nodes/Perplexity/descriptions/agent/Agent.resource.ts
```diff
@@ -0,0 +1,39 @@
...
+			{
+				name: 'Create Response',
+				value: 'createResponse',
+				action: 'Create a response',
+				description:
+					'Create a response using the Agent API with third-party models, presets, and tools',
+				routing: {
+					request: {
+						method: 'POST',
+						url: '/v1/responses',
```
dlavrenuek: In the docs this endpoint is defined as `/v1/agent`, seeing that https://docs.perplexity.ai/api-reference/responses-post is redirected to https://docs.perplexity.ai/api-reference/agent-post probably means that the `/v1/responses` is outdated/deprecated, is it correct? If yes, then I would use the new endpoint

dlavrenuek (changes requested):
Thank you for the changes @kesku, great extension of the Perplexity node capabilities! 
There is an error in the code preventing the node (package) from being built. You can do `pnpm build` in the repository root or nodes package to confirm that it works. Also added some code improvement suggestions and a question regarding the agents api.

packages/nodes-base/nodes/Perplexity/descriptions/agent/Agent.resource.ts
kesku: `/v1/responses` is an alias of `/v1/agent` and both continue to work for the sake of client compatibility, there is no difference

Will update to reference agent over responses though, same as what was flagged in `Perplexity/GenericFunctions.ts`

packages/nodes-base/nodes/Perplexity/descriptions/embeddings/createEmbedding.operation.ts
```diff
@@ -0,0 +1,95 @@
...
+				},
+			},
+			{
+				displayName: 'Encoding Format',
+				name: 'encoding_format',
+				type: 'options',
+				default: 'float',
+				options: [
+					{ name: 'Float', value: 'float' },
+					{ name: 'Base64 Int8', value: 'base64_int8' },
```
dlavrenuek: It's strange that cubic only found the problem in `createContextualized` but not here. For the embeddings float also does not exist at least in the docs: https://docs.perplexity.ai/api-reference/embeddings-post

```suggestion
				default: 'base64_int8',
				options: [
					{ name: 'Base64 Int8', value: 'base64_int8' },
```

dlavrenuek (changes requested):

packages/nodes-base/nodes/Perplexity/descriptions/embeddings/createEmbedding.operation.ts
kesku: mb! was suppsed to include in 67b1140aa2ea05f6a7dcac2789379591cc862284 since it flagged that in only one file too

dlavrenuek (approved):

PR #26972, opened 2026-03-13
fix(AWS Bedrock Chat Model Node): Extract region from modelName ARNs

schrothbn:
## Summary
When using the AWS Bedrock Chat Model node with an inference profile specified as a full ARN (e.g. arn:aws:bedrock:eu-west-3:123456:inference-profile/eu.amazon.nova-pro-v1:0), the node ignored the region in the ARN and always used the credential’s region instead. This caused AccessDenied errors for users whose AWS organizations restrict which regions can be used via Service Control Policies.

The fix parses the region from the ARN when one is provided and uses it for the BedrockRuntimeClient and ChatBedrockConverse configuration. When the model is specified as a short ID (e.g. amazon.nova-pro-v1:0 or eu.amazon.nova-pro-v1:0), the credential region is used as before.

https://linear.app/n8n/issue/AI-2174

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/llms/LmChatAwsBedrock/LmChatAwsBedrock.node.ts
+++ b/packages/@n8n/nodes-langchain/nodes/llms/LmChatAwsBedrock/LmChatAwsBedrock.node.ts
@@ -238,10 +238,18 @@ export class LmChatAwsBedrock implements INodeType {
 			maxTokensToSample: number;
 		};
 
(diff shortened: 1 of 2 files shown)


DeveloperTheExplorer (approved):
Looks good but I'm a bit confused as to how to properly handle this. If the user puts a region in their credentials that is different from the ARN url, perhaps we should at least put out a warning? May lead to some confusion.

PR #26983, opened 2026-03-13
fix(core): Handle external hook file paths on Windows

tomi:
## Summary

Main fix is:
Add `EXTERNAL_HOOK_FILES_SEPARATOR` and parse `EXTERNAL_HOOK_FILES` to support Windows drive-letter paths

To implement this cleanly:
- Extend config decorators to support deferred instance transforms and centralize primitive env coercion in `coerce.ts`

And refactored this to use the new mechanism:
- Move JWT refresh timeout validation to env parsing and add regression tests for external hooks/default config shape updates.

- https://linear.app/n8n/issue/CAT-106

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/config/src/configs/external-hooks.config.ts
```diff
@@ -1,9 +1,14 @@
...
 import { Config, Env } from '../decorators';
 
 @Config
 export class ExternalHooksConfig {
-	/** Paths to files that define external lifecycle hooks. Colon-separated for multiple files. */
-	@Env('EXTERNAL_HOOK_FILES')
-	files: ColonSeparatedStringArray = [];
+	/** Separator character for EXTERNAL_HOOK_FILES. Defaults to ':'. Use ';' on Windows. */
+	@Env('EXTERNAL_HOOK_FILES_SEPARATOR')
+	separator: string = ':';
```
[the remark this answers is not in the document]
tomi: We can't change the default as it could break existing usage

packages/@n8n/config/src/coerce.ts
```diff
@@ -0,0 +1,24 @@
+/** Parse a string env value to a number. Returns `undefined` if parsing fails. */
+export function toNumber(value: string): number | undefined {
+	const parsed = Number(value);
```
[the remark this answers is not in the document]
tomi: This was the behaviour before. As much as I'd like to change it, it might break existing usage

packages/@n8n/config/src/coerce.ts
```diff
@@ -0,0 +1,24 @@
...
+	return isNaN(parsed) ? undefined : parsed;
+}
+
+/** Parse a string env value to a boolean. Returns `undefined` if parsing fails. */
+export function toBoolean(value: string): boolean | undefined {
+	const lower = value.toLowerCase();
```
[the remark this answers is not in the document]
tomi: This is how it worked before. We don't want to change the behaviour here

packages/@n8n/config/src/configs/user-management.config.ts
```diff
@@ -125,13 +126,23 @@ export class UserManagementConfig {
...
+				return 0;
+			}
+
+			if (hours >= self.jwtSessionDurationHours) {
+				console.warn(INVALID_JWT_REFRESH_TIMEOUT_WARNING);
+				return 0;
+			}
+			return hours;
+		},
+	)
```
despairblue: This approach reduces type safety, we could return anything from this function and then the type space and the value space would be out of sync. The old sanitize approach didn't have that issue.

Can we make this type safe?

tomi: Excellent point. Unfortunately there's no clean way to do this. Couple options:
1. Label the types manually (like in this PR)
2. Convert it to an object where it is required:
```ts
@Env('MY_ENV',
	{
		type: Number,
		transform: (value, self) => {
			return 2
		}
	}
```
3. Revert back to sanitize

@despairblue WDYT?

packages/@n8n/config/src/decorators.ts
```diff
@@ -9,9 +9,15 @@ import * as coerce from './coerce';
...
-type InstanceTransform = (value: string, config: any) => unknown;
+type PropertyType =
+	| NumberConstructor
+	| BooleanConstructor
+	| StringConstructor
+	| DateConstructor
+	| Class;
+type InstanceTransform<TConfig = unknown, TValue = unknown> = {
+	bivarianceHack(value: string, config: TConfig): TValue;
+}['bivarianceHack'];
```
despairblue: Interesting thing, I tried to look into this and understand what it's doing and how it's applied here and I think some LLM went a bit overboard, unless I'm missing something. Because it seems it's not actually being put to use here. 

Am I missing something?

btw claude code was also very adamant about it being necessary until I pressed it on.

despairblue (commented):
I'm still not 100% happy with it.

The runtime type checking only works if we assign the type.

This works:
```ts
		@Config
		class TestConfig {
			@Env('TEST_VALUE', () => 43)
			value: number = 42;
		}
		const config = Container.get(TestConfig);
		expect(config.value).toBe(43);
```

This fails silently (`config.value` is 42, not 43):
```ts
		@Config
		class TestConfig {
			@Env('TEST_VALUE', () => 43)
			value = 42;
		}
		const config = Container.get(TestConfig);
		expect(config.value).toBe(42);
```

This worked fine with the sanitize function. I don't want to block this, I know this is because the type of `ExternalHooksConfig.files` depends on the content of `ExternalHooksConfig.separator`, so I'll approve it for now. 

I have to jump on something else now, otherwise I would have put in more thought about how to resolve this with less complexity. These decorators are a little too fancy, having 2 types and one function doing all the checking and transformation would have been easier to maintain I think.

I pushed 2 additional commits with a a little clean up.

despairblue (dismissed):

packages/@n8n/config/src/decorators.ts
tomi: Yeah it is LLM generated. I tried it without it, but it then fails to be able to assign it in the decorator implementaiton

despairblue (approved):
😻 🚀

PR #26984, opened 2026-03-13
chore: Improve gmail trigger

endor:
## Summary

Reduce out-of-memory (OOM) risk when the Gmail Trigger polls many or very large emails by:
1. switching to a streaming parser
2. keeping only minimal per-message data for dedupe/timestamp
3. processing messages in batches of 20

### Steps taken

1. **Minimal data in `allFetchedMessages`**  
   The trigger only needs `id`, `internalDate`, `date`, and `headers` per message for `lastTimeChecked` and `possibleDuplicates`. We now push that minimal object instead of the full Gmail message (including `raw`), so the large raw payload is no longer retained for every message and can be GC’d after parsing.

2. **Streaming parser (PostalMime)**  
   Replaced mailparser’s `simpleParser` with PostalMime, which can parse from a stream. Raw body is decoded and parsed in chunks instead of loading the full decoded string and building copies in memory. A small adapter (`postalMimeToMailparserShape`) keeps the same output shape for the rest of the Gmail node (mailparser-style addresses, `textAsHtml`, etc.) when not using the trigger’s simplified output.

3. **Explicit fetch limit and batched processing**  
   - Set `maxResults = 100` explicitly per poll (Gmail API default; keeps behaviour stable and documented).  
   - Process messages in batches of 20: fetch and parse up to 20, append to `responseData` and minimal data to `allFetchedMessages`, then move to the next batch. Peak memory is capped to roughly “20 full messages at a time” instead of up to 100.  
   - `lastTimeChecked` and `possibleDuplicates` are still computed once at the end from all fetched minimal data.

4. **Trigger output format**  
   The trigger calls `parseRawEmail` with `outputFormat: 'simple'` so it continues to return the same simplified shape (no mailparser-specific fields). Other Gmail nodes keep using the mailparser-compatible shape via the adapter.

### Measurable outcome (dataset-dependent)

With my test setup (fetching 100 messages, 3 of them ~27MB in size):
* Baseline (mailparser, full messages in memory): OOM during/after poll; end-of-poll heap ~400 MB.
* Minimal data only (mailparser): End-of-poll heap ~12% lower; still OOM (at a higher heap).
* PostalMime + minimal data: OOM eliminated. End-of-poll heap ~25% lower than minimal+mailparser; per-message parse heap ~25% lower than mailparser. Peak during poll well below OOM range.
* PostalMime + minimal + batching (e.g. 20 per batch): No OOM; peak heap capped when many large messages are spread across the poll; end-of-poll heap in the same range as without batching for this run.

https://linear.app/n8n/issue/NODE-4595/gmail-oom-on-cloud

## Review / Merge checklist

- [X] Tests included. 


packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -27,6 +27,12 @@ import type {
 	MessageListResponse,
 } from './types';
 
+/** Max messages to fetch per poll. Matches the Gmail API default (current implicit limit). */
+const MAX_MESSAGES_PER_POLL = 100;
```
Joffcom, line 31: Should this be something that a user can set? Same with batch size below.

packages/nodes-base/nodes/Google/Gmail/GenericFunctions.ts
```diff
@@ -1,6 +1,7 @@
 import isEmpty from 'lodash/isEmpty';
 import { DateTime } from 'luxon';
-import { simpleParser } from 'mailparser';
+import { Readable } from 'stream';
+import PostalMime from 'postal-mime';
```
Joffcom, line 4: Does this change the "shape" of the data output which may break emails? What impact does this have on the GMail action node as well?

packages/nodes-base/package.json
```diff
@@ -927,6 +927,7 @@
     "lossless-json": "1.0.5",
     "luxon": "catalog:",
     "mailparser": "3.9.3",
+    "postal-mime": "2.7.3",
```
Joffcom, line 930: Should we run this past Declan / QA to make sure they are happy with this dependency?

Joffcom (commented):
Overall looks good to me, Just a few questions to think about.

packages/nodes-base/package.json
endor: Sure. ping @shortstacked.

packages/nodes-base/nodes/Google/Gmail/GenericFunctions.ts
endor: So the idea was for `toMailparserShape` to make sure we still get the same output as before and the tests still pass, so I guess it matches.

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
endor: Maybe, but that feels like a separate change, since they weren't able to set this before.

packages/nodes-base/package.json
shortstacked: If we are adding postal-mime can we remove mailparser?

packages/nodes-base/nodes/Google/Gmail/GenericFunctions.ts
```diff
@@ -143,62 +145,138 @@ export async function googleApiRequest(
...
 	}
 
-	const json = Object.assign({}, mailBaseData, responseData, {
+	const fromFormatted = formatAddress(email.from);
+	const toFormatted =
+		email.to
+			?.map((a: PostalMimeAddress) => formatAddress(a))
+			.filter(Boolean)
+			.join(', ') ?? '';
+	const json: IDataObject = {
```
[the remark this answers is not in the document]
shortstacked: Agree with the above, isn't this a breaking change to existing consumers? Would this be a v3 of the trigger node?

packages/nodes-base/package.json
endor: We're still using it in the `EmailReadImap` node.

packages/nodes-base/nodes/Google/Gmail/GenericFunctions.ts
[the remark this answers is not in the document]
endor: So the idea was to have the output be the same as before. It seems I missed some fields. Now added.

BerniWittmann:
As this discussion around this PR (especially around maintaining the output format) looks a bit stale, I have created https://github.com/n8n-io/n8n/pull/28470 (https://linear.app/n8n/issue/NODE-4595) which adds V2 to the Gmail Trigger Node which sets a `maxResults` property to be able to limit the memory consumption. 
My idea was to have this out first to help improve the current situation, while you guys can decide on this PR here :)

PR #26990, opened 2026-03-13
fix(core): Retry multi-main follower license check during startup

tomi:
## Summary

Adds a startup license-check retry path for multi-main followers so they can wait for the leader-written cert instead of failing immediately.
Followers now retry with exponential backoff and `license.reload()`, while leaders still fail fast when multi-main is not licensed.
Includes regression tests for follower success-after-retry, follower failure-after-retries, and leader no-retry behavior.

https://linear.app/n8n/issue/CAT-2572

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -8,7 +8,7 @@ import { McpServer } from '@n8n/n8n-nodes-langchain/mcp/core';
 import glob from 'fast-glob';
 import { createReadStream, createWriteStream, existsSync } from 'fs';
 import { mkdir } from 'fs/promises';
-import { jsonParse, type IWorkflowExecutionDataProcess } from 'n8n-workflow';
+import { jsonParse, sleep, type IWorkflowExecutionDataProcess } from 'n8n-workflow';
 import path from 'path';
 import replaceStream from 'replacestream';
 import { pipeline } from 'stream/promises';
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/start.ts
```diff
@@ -312,6 +312,30 @@ export class Start extends BaseCommand<z.infer<typeof flagsSchema>> {
 		}
 	}
 
+	/**
+	 * Ensures the multi-main license entitlement is present. Followers retry with
+	 * exponential backoff because the leader may not have written the cert to the
+	 * database yet when the follower starts up.
```
tomi, line 318: This is a bit of a working theory. If you know what else might cause the issue, LMK!

packages/cli/src/commands/start.ts
```diff
@@ -312,6 +312,30 @@ export class Start extends BaseCommand<z.infer<typeof flagsSchema>> {
...
+	 */
+	private async ensureMultiMainLicensed() {
+		if (this.license.isMultiMainLicensed()) return;
+
+		if (!this.instanceSettings.isLeader) {
+			const maxRetries = 5;
+			for (let attempt = 1; attempt <= maxRetries; attempt++) {
```
afitzek, line 325: If the leader would change in this loop, to this instance, I assume it would not recover, but this is not new behavior either.

afitzek (approved):
Looks good to me. 🚀

PR #26993, opened 2026-03-13
feat(core): Serialize Luxon types at V8 isolate boundary

despairblue:
## Summary

Luxon DateTime/Duration/Interval objects lose their class identity when returned from the V8 isolate via `copy: true` (structured clone). This PR adds a `__prepareForTransfer` function that runs inside the isolate after expression evaluation, converting Luxon types to ISO strings before the result crosses the boundary.

Also exposes `Duration` and `Interval` on `globalThis` in the runtime bundle so expressions like `Interval.after(...)` work inside the isolate.

**Key design decisions:**
- Conversion happens inside the isolate where Luxon objects still have working methods — no duck-typing needed on the host side
- Only plain objects are recursively walked; structured-cloneable types (Date, Map, Set, RegExp) pass through untouched
- Chaining within expressions is unaffected — DateTime stays a real instance inside the isolate, only converted at the exit boundary
- Tests use shared `asDateTime`/`asDuration`/`asInterval` helpers that normalize results for both engines

https://linear.app/n8n/issue/CAT-2541

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/expression-runtime/src/runtime/index.ts
+++ b/packages/@n8n/expression-runtime/src/runtime/index.ts
@@ -1,10 +1,11 @@
-import { DateTime } from 'luxon';
+import { DateTime, Duration, Interval } from 'luxon';
 
 import { extend, extendOptional } from '../extensions/extend';
 
 import { SafeObject, SafeError, ExpressionError } from './safe-globals';
 import { createDeepLazyProxy } from './lazy-proxy';
 import { resetDataProxies } from './reset';
+import { __prepareForTransfer } from './serialize';
 
 // Augment globalThis with runtime properties
 declare global {
@@ -29,6 +30,9 @@ declare global {
 
 		// Expression engine globals
 		var DateTime: typeof import('luxon').DateTime;
+		var Duration: typeof import('luxon').Duration;
+		var Interval: typeof import('luxon').Interval;
+		var __prepareForTransfer: (value: unknown) => unknown;
 		var extend: typeof import('../extensions/extend').extend;
 		var extendOptional: typeof import('../extensions/extend').extendOptional;
 
@@ -58,6 +62,8 @@ declare global {
 globalThis.extend = extend;
 globalThis.extendOptional = extendOptional;
 globalThis.DateTime = DateTime;
+globalThis.Duration = Duration;
+globalThis.Interval = Interval;
 
 // ============================================================================
 // Expose security globals and runtime functions
@@ -69,6 +75,7 @@ globalThis.SafeError = SafeError;
 
 globalThis.createDeepLazyProxy = createDeepLazyProxy;
 globalThis.resetDataProxies = resetDataProxies;
+globalThis.__prepareForTransfer = __prepareForTransfer;
 
 // Initialize empty __data object (populated by resetDataProxies before each evaluation)
 globalThis.__data = {};
--- a/packages/@n8n/expression-runtime/src/runtime/serialize.ts
+++ b/packages/@n8n/expression-runtime/src/runtime/serialize.ts
@@ -1,5 +1,11 @@
 import { DateTime, Duration, Interval } from 'luxon';
 
+/** Type guard: plain object with Object.prototype or null prototype. */
+function isPlainObject(value: object): value is Record<string, unknown> {
+	const proto = Object.getPrototypeOf(value);
+	return proto === Object.prototype || proto === null;
+}
+
 /**
  * Prepare a value for transfer across the V8 isolate boundary.
  *
@@ -36,13 +42,12 @@ export function __prepareForTransfer(value: unknown): unknown {
 
 	// Only walk plain objects — structured-cloneable types like Date, Map, Set,
 	// RegExp, Error, typed arrays survive copy:true with prototypes intact.
-	const proto = Object.getPrototypeOf(value);
-	if (proto !== Object.prototype && proto !== null) return value;
+	if (!isPlainObject(value)) return value;
 
 	// Plain object — walk values
 	const result: Record<string, unknown> = {};
 	for (const key of Object.keys(value)) {
-		result[key] = __prepareForTransfer((value as Record<string, unknown>)[key]);
+		result[key] = __prepareForTransfer(value[key]);
 	}
 	return result;
 }
--- a/packages/@n8n/expression-runtime/src/runtime/serialize.ts
+++ b/packages/@n8n/expression-runtime/src/runtime/serialize.ts
@@ -1,3 +1,5 @@
+import { DateTime, Duration, Interval } from 'luxon';
+
 /**
  * Prepare a value for transfer across the V8 isolate boundary.
  *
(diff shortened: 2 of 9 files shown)


packages/@n8n/expression-runtime/src/runtime/index.ts
```diff
@@ -53,6 +59,7 @@ globalThis.SafeError = SafeError;
 
 globalThis.createDeepLazyProxy = createDeepLazyProxy;
 globalThis.resetDataProxies = resetDataProxies;
+globalThis.__prepareForTransfer = __prepareForTransfer;
```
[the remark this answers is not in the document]
despairblue: Valid concern, but this applies to all runtime globals (`DateTime`, `SafeObject`, `resetDataProxies`, `createDeepLazyProxy`, etc.) — they're all assigned the same way. Fixing just this one would be inconsistent. Created [CAT-2614](https://linear.app/n8n/issue/CAT-2614) to address all of them together.

packages/@n8n/expression-runtime/src/runtime/serialize.ts
```diff
@@ -0,0 +1,43 @@
...
+	// Luxon Interval -> ISO string (e.g., "2024-01-01T00:00/2024-01-02T00:00")
+	if ((value as any).isLuxonInterval) return (value as any).toISO();
+
+	// Array — walk elements
+	if (Array.isArray(value)) return value.map(__prepareForTransfer);
+
+	// Only walk plain objects — structured-cloneable types like Date, Map, Set,
+	// RegExp, Error, typed arrays survive copy:true with prototypes intact.
+	const proto = Object.getPrototypeOf(value);
+	if (proto !== Object.prototype && proto !== null) return value;
```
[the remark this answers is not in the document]
despairblue: n8n expressions should not produce Maps or Sets, as they don't survive (de)serialization to/from the db.

packages/workflow/test/expression.test.ts
```diff
@@ -88,12 +88,14 @@ describe('Expression', () => {
...
-			);
+			const intervalResult = asInterval(evaluate('={{Interval.after(new Date(), 100)}}'));
+			expect(intervalResult).toBeInstanceOf(Interval);
+			expect(intervalResult.length('milliseconds')).toEqual(100);
 			vi.useRealTimers();
 
-			expect(evaluate('={{Duration.fromMillis(100)}}')).toEqual(Duration.fromMillis(100));
+			const durationResult = asDuration(evaluate('={{Duration.fromMillis(100)}}'));
+			expect(durationResult).toBeInstanceOf(Duration);
+			expect(durationResult.toMillis()).toEqual(100);
```
despairblue, line 98: The test is being rewritten to still assert that global built-ins from the allow-list can be used, while still making sure that it passes with both engines. 

Whereas the old engine will actually return an interval or a duration and the new engine will return a string, which makes more sense, because this is also what it will be serialized into when stored in the database, and also returning an interval or a duration is forbidden by certain nodes like the set node — it will actually throw because it does type validation after an expression.

packages/workflow/test/ExpressionExtensions/string-extensions.test.ts
```diff
@@ -276,13 +276,17 @@ describe('Data Transformation Functions', () => {
...
-			expect(evaluate('={{ "Wed, 21 Oct 2015 07:28:00 GMT".toDateTime() }}')).toBeInstanceOf(
+			expect(
+				asDateTime(evaluate('={{ "Wed, 21 Oct 2015 07:28:00 GMT".toDateTime() }}')),
+			).toBeInstanceOf(DateTime);
+			expect(asDateTime(evaluate('={{ "2008-11-11".toDateTime() }}'))).toBeInstanceOf(DateTime);
+			expect(asDateTime(evaluate('={{ "1-Feb-2024".toDateTime() }}'))).toBeInstanceOf(DateTime);
+			expect(asDateTime(evaluate('={{ "1713976144063".toDateTime("ms") }}'))).toBeInstanceOf(
+				DateTime,
+			);
+			expect(asDateTime(evaluate('={{ "31-01-2024".toDateTime("dd-MM-yyyy") }}'))).toBeInstanceOf(
```
despairblue, line 287: The test is being rewritten to still assert that global built-ins from the allow-list can be used, while still making sure that it passes with both engines. 

Whereas the old engine will actually return an interval or a duration and the new engine will return a string, which makes more sense, because this is also what it will be serialized into when stored in the database, and also returning an interval or a duration is forbidden by certain nodes like the set node — it will actually throw because it does type validation after an expression.

packages/workflow/test/ExpressionExtensions/date-extensions.test.ts
despairblue, line 1: The test is being rewritten to still assert that global built-ins from the allow-list can be used, while still making sure that it passes with both engines. 

Whereas the old engine will actually return an interval or a duration and the new engine will return a string, which makes more sense, because this is also what it will be serialized into when stored in the database, and also returning an interval or a duration is forbidden by certain nodes like the set node — it will actually throw because it does type validation after an expression.

packages/@n8n/expression-runtime/src/runtime/serialize.ts
```diff
@@ -0,0 +1,48 @@
...
+ * identity and arrive on the host as plain objects.
+ *
+ * This function recursively walks a value and converts types that don't
+ * survive structured clone into their string representations. It runs
+ * inside the isolate before the result is transferred.
+ *
+ * Note: JS Date objects survive structured clone with prototype intact
+ * (Date is a standard structured-cloneable type) and are not converted.
+ *
+ * Circular references are not handled — expression results should not
```
tomi, line 23: Nice comment 👌

packages/@n8n/expression-runtime/src/runtime/serialize.ts
```diff
@@ -0,0 +1,48 @@
...
+		return iso === 'Invalid Interval' ? null : iso;
+	}
+
+	// Array — walk elements
+	if (Array.isArray(value)) return value.map(__prepareForTransfer);
+
+	// Only walk plain objects — structured-cloneable types like Date, Map, Set,
+	// RegExp, Error, typed arrays survive copy:true with prototypes intact.
+	const proto = Object.getPrototypeOf(value);
+	if (proto !== Object.prototype && proto !== null) return value;
```
tomi: This could be moved to a typeguard (e.g. `isPlainObject`) that ensures `value` is `Record<string, unknown>`. Then we wouldn't need the cast on line 45

tomi (commented):
Nice 👏 Just one minor suggestion

packages/@n8n/expression-runtime/src/runtime/serialize.ts
despairblue: Done in 769a0b0b19 — extracted `isPlainObject` type guard so the cast is no longer needed.

tomi (approved):
🚀

PR #27014, opened 2026-03-13
feat(API): Add GET /api/v1/discover endpoint for capability discovery

shortstacked:
## Why

AI agents interacting with the n8n public API currently need to fetch and parse the full OpenAPI spec (107 KB / ~25,000 tokens) just to understand what they can do. Most of that spec is irrelevant to a given API key's scopes. This endpoint reduces that to a single, scoped JSON response:

| Approach | Size | Tokens | Calls |
|----------|------|--------|-------|
| Full OpenAPI spec | 107 KB | ~25,000 | 1 + client-side parsing |
| Discover (full-scope key) | 6.9 KB | ~1,700 | 1 |
| Discover + filtered + schemas | ~1 KB | ~150 | 1 |

An agent can now create a workflow in 2 calls (discover → create) instead of downloading the spec, parsing YAML, resolving `$ref`s, and figuring out which endpoints their key has access to.

## Summary
- Adds `GET /api/v1/discover` — a scoped, machine-readable capability map that returns only the endpoints, operations, and request schemas the caller's API key has access to
- Supports query parameter filtering: `?resource=workflow`, `?op=create`, `?include=schemas` — combinable for minimal responses (~150 tokens)
- Self-describing: response includes a `filters` field listing available filter params and their valid values, so agents learn how to use the endpoint from its first response
- Uses runtime middleware introspection (`__apiKeyScope` tagging) for zero-maintenance scope detection — new endpoints are automatically discoverable
- Results cached in memory after first parse
- Works on both community and enterprise editions

## How it works
1. Parses the OpenAPI spec and introspects handler middleware chains to detect scope requirements
2. Filters endpoints by the caller's API key scopes
3. Groups by resource tag with operations and endpoint details
4. Applies optional `?resource=` and `?op=` filters to narrow the response
5. Optionally resolves all `$ref` references recursively (including nested schemas like workflow → nodes → node)

## Query parameters
| Parameter | Example | Effect |
|-----------|---------|--------|
| `resource` | `?resource=workflow` | Only return endpoints for that resource |
| `op` | `?op=create` | Only return endpoints matching that operation |
| `include` | `?include=schemas` | Inline fully-resolved request body schemas |

All three can be combined: `?resource=workflow&op=create&include=schemas`

## Response shape
```json
{
  "data": {
    "scopes": ["workflow:create", "workflow:read"],
    "resources": {
      "workflow": {
        "operations": ["create", "read"],
        "endpoints": [
          {
            "method": "POST",
            "path": "/api/v1/workflows",
            "operationId": "createWorkflow",
            "requestSchema": { "type": "object", "properties": { ... } }
          }
        ]
      }
    },
    "filters": {
      "resource": { "description": "Filter to a specific resource", "values": ["workflow", "tags", ...] },
      "op": { "description": "Filter to a specific operation", "values": ["create", "read", ...] },
      "include": { "description": "Include additional data", "values": ["schemas"] }
    },
    "specUrl": "/api/v1/openapi.yml"
  }
}
```

## Files changed
- **New:** `handlers/discover/` — handler, service, OpenAPI spec, unit tests
- **Modified:** `global.middleware.ts` — always tags scope middleware with `__apiKeyScope` (read-only annotation, no behavior change to enforcement)
- **Modified:** `openapi.yml` — registers `/discover` path and `Discover` tag
- **New:** Integration tests, E2E tests, Playwright helper

## Test plan
- [x] 33 unit tests (service: scope filtering, resource grouping, schema inclusion/exclusion, nested $ref resolution, circular ref protection, resource/op filtering, self-describing filters; handler: auth, query parsing, param normalization)
- [x] 12 integration tests (full HTTP round-trip with real DB, scope-filtered responses, resource filter)
- [x] 4 scoped integration tests (op filtering, combined filters — requires license)
- [x] 4 E2E tests (Playwright, end-to-end against running instance)
- [x] Typecheck passes


packages/cli/src/public-api/v1/handlers/discover/discover.service.ts
```diff
@@ -0,0 +1,289 @@
...
+ * within a schema object, inlining the resolved definitions. This handles
+ * nested refs (e.g. workflow → nodes → node) so the returned schema is
+ * fully self-contained. A `seen` set prevents circular reference loops.
+ */
+function resolveRefs(
+	obj: Record<string, unknown>,
+	componentSchemas: Record<string, unknown>,
+	seen = new Set<string>(),
+): Record<string, unknown> {
+	if (typeof obj.$ref === 'string' && obj.$ref.startsWith(COMPONENTS_SCHEMA_PREFIX)) {
```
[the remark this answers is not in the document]
shortstacked: source schemas don’t have nested file $refs; #/components/schemas/ refs are recursively resolved

packages/cli/src/public-api/v1/handlers/discover/discover.service.ts
```diff
@@ -0,0 +1,356 @@
...
+	const allEndpoints = await parseEndpointsFromSpec();
+	const scopeSet = new Set(callerScopes);
+	const includeSchemas = options?.includeSchemas === true;
+	const scopesEnabled = options?.scopesEnabled !== false;
+
+	// When scopes are not licensed (community edition), show all endpoints
+	// since all API keys have unrestricted access regardless of stored scopes
+	const filtered = scopesEnabled
+		? allEndpoints.filter((ep) => ep.scope === null || scopeSet.has(ep.scope))
+		: allEndpoints;
```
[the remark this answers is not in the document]
shortstacked: Not a scopesEnabled regression — community keys already have all role scopes stored (resolveScopesForUser gives getApiKeyScopesForRole when scopes aren't licensed), so the scope filter passes for sourceControl etc. regardless of the flag.

The real gap is that discover doesn't filter by license features (e.g. isLicensed middleware). Intentional for now — the actual endpoint returns 403 with a clear 'feature not licensed' message, which is self-explanatory. We can add license-aware filtering as a follow-up if it causes confusion.

packages/cli/src/public-api/v1/handlers/discover/discover.service.ts
```diff
@@ -0,0 +1,356 @@
...
+
+	return resolveRefs(schema, componentSchemas);
+}
+
+async function parseEndpointsFromSpec(): Promise<EndpointInfo[]> {
+	if (!cachedEndpointsPromise) {
+		cachedEndpointsPromise = _parseEndpointsFromSpec().catch((e) => {
+			cachedEndpointsPromise = undefined;
+			throw e;
+		});
```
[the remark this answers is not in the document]
shortstacked: Valid point — parse failures here are permanent (local YAML files). Removing the catch so a failed promise stays cached and isn't retried.

packages/cli/src/public-api/v1/handlers/discover/spec/paths/discover.yml
```diff
@@ -0,0 +1,96 @@
...
+        Include additional data. Use "schemas" to inline request body schemas
+        per endpoint, eliminating the need to fetch the full OpenAPI spec.
+    - name: resource
+      in: query
+      required: false
+      schema:
+        type: string
+      description: >
+        Filter to a specific resource (e.g. "workflow", "tags", "credential").
+    - name: op
```
guillaumejacquart: just wondering is `operation` would be a more explicit name. Not sure I see the value of shortening it

packages/cli/src/public-api/v1/handlers/discover/discover.handler.ts
```diff
@@ -0,0 +1,49 @@
...
+		async (
+			req: AuthenticatedRequest<{}, {}, {}, { include?: string; resource?: string; op?: string }>,
+			res: express.Response,
+		): Promise<express.Response> => {
+			const apiKey = firstString(req.headers['x-n8n-api-key']);
+			if (!apiKey) {
+				return res.status(401).json({ message: 'Unauthorized' });
+			}
+
+			const apiKeyRecord = await Container.get(ApiKeyRepository).findOne({
```
guillaumejacquart, line 34: (nit) I'm just wondering if we couldn't pass the scopes (only required field it seems) from the getApiKeyScopeMiddleware, which already fetch the API Key and get the scopes from it.

packages/cli/src/public-api/v1/handlers/discover/discover.service.ts
```diff
@@ -0,0 +1,353 @@
...
+
+const COMPONENTS_SCHEMA_PREFIX = '#/components/schemas/';
+
+/**
+ * Recursively resolve all `$ref: '#/components/schemas/<name>'` references
+ * within a schema object, inlining the resolved definitions. This handles
+ * nested refs (e.g. workflow → nodes → node) so the returned schema is
+ * fully self-contained. A `seen` set prevents circular reference loops.
+ */
+export function resolveRefs(
```
guillaumejacquart: I feel like lots of functions in this file around parsing the specs could be simplified a lot if we used `@apidevtools/json-schema-ref-parser`, which is already a shared dependency with `express-openapi-validator`. For instance, the `dereference` method already loads all the inner "refs" 

Also more generally speaking about how many times we load the specs file. index.ts already loads openapi.yml via yamljs for Swagger UI (with its own cache), and express-openapi-validator parses it again internally for request validation. The discover service would add  a third independent parse + cache (cachedEndpointsPromise).

guillaumejacquart (commented):
Some comments

packages/cli/src/public-api/v1/handlers/discover/discover.service.ts
shortstacked: Brilliant, I didn't even realize we had the parser already that makes this way smaller, thanks!

packages/cli/src/public-api/v1/handlers/discover/spec/paths/discover.yml
shortstacked: That's fair, I don't think tokens need to be that short. And being clearer makes more sense.

guillaumejacquart (approved):
Looks good to me

PR #21202, opened 2025-10-26
feat(core): Use active version instead of current version (no-changelog)

dariacodes:
## Summary

This PR decouples the active version from the current version of the workflow:

- Add a foreign (nullable) key `activeVersionId` to the workflow entity
    - It does not currently replace the `active` flag, but we will tackle this as tech debt later.
- Automatic executions now always run the populated active version of the workflow. Manual executions still use the data that is sent.
- Update public API endpoints with keeping the current behavior
    - Update `activeVersion` with the current version when updating and activating workflows
    - Return `activeVersion` for GET endpoints (it'll be the same as the current version until user starts using V2 endpoints)
    - Remove `excludePinnedData` from activate endpoint (it was unused and using "GET" type was semantically incorrect)
- Update the logic to always first add a record to workflow history and only then update the workflow entity, so that `activeVersion` is set once along with the workflow entity update

### Active Version Usage

These are now using the active workflow version:

1. Workflow Activation [`active-workflow-manager.ts` - [add](https://github.com/n8n-io/n8n/blob/4ea2757df67519c0b7e1de19a35c55d74ed93266/packages/cli/src/active-workflow-manager.ts#L540)]
    - When activating workflows, the system now uses `activeVersion` nodes and connections instead of current version.
2. Webhook Cleanup [`active-workflow-manager.ts` - [clearWebhooks](https://github.com/n8n-io/n8n/blob/4ea2757df67519c0b7e1de19a35c55d74ed93266/packages/cli/src/active-workflow-manager.ts#L248)]
    - When deactivating workflows and removing webhooks, the system now uses `activeVersion` to properly identify and clean up the correct webhooks.
3. Activation Error Handling [`active-workflow-manager.ts` - [activateWorkflow](https://github.com/n8n-io/n8n/blob/4ea2757df67519c0b7e1de19a35c55d74ed93266/packages/cli/src/active-workflow-manager.ts#L462)]
    - When workflow activation fails and throws an error workflow needs to execute, it now uses `activeVersion` instead of current version.
4. Production Webhook Execution [`live-webhooks.ts` - [executeWebhook](https://github.com/n8n-io/n8n/blob/8527e4b6f1b35472ac472e5c3857cb0e072acb0a/packages/cli/src/webhooks/live-webhooks.ts#L71)]
    - Production webhooks now use `activeVersion` instead of current version.

### What I tested and how

- For public endpoints, I created workflows with an HTTP Request node that sends a payload to activate or update another workflow.
[...]

PR #27017, opened 2026-03-13
fix(core): Old version of workflow still active after pulling [1.x]

konstantintieber:
## Summary

Hint to reviewer: This PR is deployed to two test instances (for positive test) as well as one test instance with the latest state of 1.x (for negative test).
See this slack message for more details: https://n8nio.slack.com/archives/C0A9PM073QW/p1773822807520939

This PR targets 1.x only, it is not needed for 2.x.

Fixes a regression on the environments feature introduced in https://github.com/n8n-io/n8n/pull/21202/ (v 1.122.2)

Video demo of the unexpected behaviour: https://www.loom.com/share/4e3efc2aa6934399bb2d4ffe4b80a83c

In the 1.x world, where a workflow can either by active or inactive, pulling a new version of an active workflow used to have that new version be the new active one.
PR #21202 (linked above) resulted in the new version being pulled, but the old version of the workflow being the one that was still active. As a 1.x user, I have no easy way to detect that my active workflow is actually using an older version for production executions.

The only known workaround is to deactivate and activate an active workflow after pulling.

**For a follow-up:** Should we build the means to inform admins of n8n instances using versions 1.122.2 to 1.123.25 that they have active workflows where the active version is not the latest version of the workflow, and provide them with an API to set the `activeVersionId` of all diverged active workflows to the latest version of each workflow?

https://linear.app/n8n/issue/LIGO-331/be-environments-pulled-active-workflow-on-1x-will-keep-executing-older

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
+++ b/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
@@ -1292,7 +1292,7 @@ export class SourceControlImportService {
 	): Promise<void> {
 		if (!importedWorkflow.versionId || !importedWorkflow.nodes || !importedWorkflow.connections) {
 			this.logger.debug('Skipping workflow history - missing versionId, nodes, or connections');
-			return undefined;
+			return;
 		}
 
 		// Fetch user for author info
--- a/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
+++ b/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
@@ -639,6 +639,7 @@ export class SourceControlImportService {
 		};
 	}
 
+	// eslint-disable-next-line complexity
 	async importWorkflowFromWorkFolder(candidates: SourceControlledFile[], userId: string) {
 		const personalProject = await this.projectRepository.getPersonalProjectForUserOrFail(userId);
 		const candidateIds = candidates.map((c) => c.id);
@@ -698,11 +699,12 @@ export class SourceControlImportService {
 			);
 			if (upsertResult?.identifiers?.length !== 1) {
 				throw new UnexpectedError('Failed to upsert workflow', {
+					// TODO: why is this nullish coalescing operator here, when importedWorkflows that don't have an id are skipped via "continue;" above?
 					extra: { workflowId: importedWorkflow.id ?? 'new' },
 				});
 			}
 
-			const savedVersionId = await this.saveOrUpdateWorkflowHistory(importedWorkflow, userId);
+			await this.saveOrUpdateWorkflowHistory(importedWorkflow, userId);
 			const localOwner = allSharedWorkflows.find(
 				(w) => w.workflowId === importedWorkflow.id && w.role === 'workflow:owner',
 			);
@@ -715,22 +717,32 @@ export class SourceControlImportService {
 				repository: this.sharedWorkflowRepository,
 			});
 
-			if (importedWorkflow.active && !savedVersionId) {
-				throw new UnexpectedError(
-					'Failed to create workflow history entry for new active version',
+			if (
+				importedWorkflow.active &&
+				existingWorkflow?.activeVersionId &&
+				importedWorkflow.versionId
+			) {
+				// Ensure the pulled version of a previously active version is now the active one
+				await this.activateImportedWorkflow(
 					{
-						extra: { workflowId: importedWorkflow.id ?? 'new' },
+						workflowId: importedWorkflow.id,
+						versionIdToActivate: importedWorkflow.versionId,
 					},
+					userId,
 				);
+			} else if (importedWorkflow.isArchived && existingWorkflow?.activeVersionId) {
+				await this.activeWorkflowManager.remove(existingWorkflow.id);
+				await this.workflowPublishHistoryRepository.addRecord({
+					workflowId: existingWorkflow.id,
+					versionId: existingWorkflow.activeVersionId,
+					event: 'deactivated',
+					userId,
+				});
+				await this.workflowRepository.updateActiveState(existingWorkflow.id, false);
 			}
 
-			await this.activateImportedWorkflowIfAlreadyActive(
-				{ existingWorkflow, importedWorkflow, versionIdToActivate: savedVersionId! },
-				userId,
-			);
-
 			importWorkflowsResult.push({
-				id: importedWorkflow.id ?? 'unknown',
+				id: importedWorkflow.id,
 				name: candidate.file,
 			});
 		}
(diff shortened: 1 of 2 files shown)


packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -1335,11 +1339,13 @@ export class SourceControlImportService {
 					importedWorkflow.id,
 				);
 			}
+			return importedWorkflow.versionId;
 		} catch (error) {
 			this.logger.error(
 				`Failed to save/update workflow history for workflow ${importedWorkflow.id}`,
 				{ error: ensureError(error) },
 			);
+			return undefined;
```
geemanjs: Should we propagate this error up? Silently failing feels like we may end up in an inconsistent state later - but it's probably better we know / report back to the user earlier.

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -1280,10 +1284,10 @@ export class SourceControlImportService {
 	private async saveOrUpdateWorkflowHistory(
 		importedWorkflow: IWorkflowToImport,
 		userId: string,
-	): Promise<void> {
+	): Promise<string | undefined> {
```
geemanjs: I'm not a big fan of returning `undefined` from a service call 👀 to me the action either succeeded (response) or didn't (error). The caller to the service should then be deciding what to do with that information

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
konstantintieber: I agree. It's kind of weird how we have an existing unit test about [should handle activation errors gracefully](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/modules/source-control.ee/__tests__/source-control-import.service.ee.test.ts#L939).

The only actual error we throw on the `master` branch in this code right now is when the db upsert statement failed "Failed to upsert workflow".

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
konstantintieber: Agreed. I did it to be as backwards compatible as possible for the workflow history part of importing a workflow failing to be as "graceful as possible"

geemanjs (dismissed):

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
konstantintieber: Once I understood this method in detail, I changed its return signature to `Promise<void>` again in https://github.com/n8n-io/n8n/pull/27017/commits/654e244556d93f741ef65f8220075833d58167af

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
konstantintieber: Added todo comments for errors that are silently logged on the server rather than being communicated to the user on pull:
- https://github.com/n8n-io/n8n/pull/27017/commits/f4ddd57706e868054ab021ec42ea035a506d3562
- https://github.com/n8n-io/n8n/pull/27017/commits/f4ddd57706e868054ab021ec42ea035a506d3562

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -639,6 +639,7 @@ export class SourceControlImportService {
 		};
 	}
 
+	// eslint-disable-next-line complexity
```
konstantintieber, line 641: I had to add this since the amount of if statements in this method has increased for the case of handling "pulling of previously active workflow that is now archived" and "pulling of previously active workflow whose activeVersionId now needs to be updated" in two separate code branches, rather than in one -> while eslint rule says complexity is too high now, I find the code easier to reason about now.

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -746,52 +766,41 @@ export class SourceControlImportService {
...
+				didPublish = true;
 			}
 		} catch (e) {
+			// TODO: this operation failing should be told to the user in the app, rather than just being logged
 			const error = ensureError(e);
-			this.logger.error(`Failed to activate workflow ${existingWorkflow.id}`, { error });
+			this.logger.error(`Failed to activate workflow ${workflowId}`, { error });
 		} finally {
-			// update the versionId of the workflow to match the imported workflow
-			await this.workflowRepository.update(
```
konstantintieber, line 777: @ireneea I removed this `workflowRepository.update` call here, because I have found that it redundantly sets the versionId of the imported workflow to the `versionId` from the tracked workflow.json file in git.
Notifying you, since you added this update statement here: https://github.com/n8n-io/n8n/commit/65942c3b7b1a643e415efcaba7b582d829623160#diff-9372037b7cadf07a667f84f65e8ace7bdb924915f484d2aebfe5249d68c036baR635

Why do I deem it redundant?
Because higher above inside the `importWorkflowFromWorkFolder` method, there is a [workflowRepository.upsert()](https://github.com/n8n-io/n8n/blob/c2757ae9e467c70a81ba14a40920c6983fbd432c/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts#L693) call, that will already set the `versionId` of the imported workflow to the versionId from the json file.

Would be great if you can take a look as well and maybe add why the `workflowRepository.update({ id: existingWorkflow.id }, { versionId: importedWorkflow.versionId })` should still remain in addition to the earlier `upsert`.

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
```diff
@@ -746,52 +765,39 @@ export class SourceControlImportService {
...
-				this.logger.debug(`Reactivating workflow id ${existingWorkflow.id}`);
-				await this.activeWorkflowManager.add(existingWorkflow.id, 'activate');
-				didAdd = true;
-			}
+			this.logger.debug(`Deactivating workflow id ${workflowId}`);
+			await this.activeWorkflowManager.remove(workflowId);
+
+			// try activating the imported workflow
+			this.logger.debug(`Reactivating workflow id ${workflowId}`);
+			await this.workflowRepository.updateActiveState(workflowId, true);
```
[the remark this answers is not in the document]
konstantintieber: This should rather be fixed by surfacing the error to the user.
There is no known case to me where workflow activation would fail here.

packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts
ireneea: Agreed, at the time when this was added, there was no logic in `importWorkflowFromWorkFolder` to set the `versionId` now that this is the case we no longer need that extra update here. So all good for removing it 👍

ireneea (approved):
All good, tested this and works as expected nice fix 🚀

PR #27036, opened 2026-03-14
feat(core): Add server-side search, pagination, and filtering to GET /projects

r00gm:
## Summary

Add `skip`, `take`, `search`, `type`, and `activated` query parameters to the `GET /projects` endpoint to support server-side pagination, search, and filtering.

- **Admin users** (`project:read` scope): queries all projects via `findAllProjectsAndCount`
- **Non-admin users**: queries accessible projects (all personal + team projects where user is a member) via `getAccessibleProjectsAndCount`, using a `DISTINCT` subquery to avoid duplicate rows
- **Activation-aware ordering**: team projects → activated personal projects → pending personal projects, alphabetically within each group
- **Backward-compatible response**: when `take` or `skip` params are present, returns `{ count, data }` envelope; otherwise returns a bare array preserving existing behavior

### New query parameters

| Param | Type | Default | Description |
|-------|------|---------|-------------|
| `skip` | number | `0` | Offset for pagination |
| `take` | number | — | Limit for pagination (max 250) |
| `search` | string | — | Case-insensitive name search (LIKE %search%) |
| `type` | `personal` \| `team` | — | Filter by project type |
| `activated` | `true` \| `false` | — | When true, excludes personal projects of pending users |

https://linear.app/n8n/issue/IAM-393/add-server-side-search-pagination-and-filtering-to-project-list

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/api-types/src/dto/project/list-projects-query.dto.ts
+++ b/packages/@n8n/api-types/src/dto/project/list-projects-query.dto.ts
@@ -1,0 +1,36 @@
+import { z } from 'zod';
+
+import { Z } from '../../zod-class';
+import { MAX_ITEMS_PER_PAGE, paginationSchema } from '../pagination/pagination.dto';
+
+const searchValidator = z.string().trim().optional();
+
+const typeValidator = z.enum(['personal', 'team']).optional();
+
+const activatedValidator = z
+	.enum(['true', 'false'])
+	.optional()
+	.transform((val) => (val === undefined ? undefined : val === 'true'));
+
+// Custom take validator that defaults to undefined (no limit) for backward
+// compatibility — existing callers that omit `take` receive all projects.
+// When provided, the value is capped at MAX_ITEMS_PER_PAGE.
+const takeValidator = z
+	.string()
+	.optional()
+	.transform((val) => (val ? parseInt(val, 10) : undefined))
+	.refine((val) => val === undefined || (!isNaN(val) && Number.isInteger(val)), {
+		message: 'Param `take` must be a valid integer',
+	})
+	.refine((val) => val === undefined || val >= 0, {
+		message: 'Param `take` must be a non-negative integer',
+	})
+	.transform((val) => (val !== undefined ? Math.min(val, MAX_ITEMS_PER_PAGE) : undefined));
+
+export class ListProjectsQueryDto extends Z.class({
+	...paginationSchema,
+	take: takeValidator,
+	search: searchValidator,
+	type: typeValidator,
+	activated: activatedValidator,
+}) {}
--- a/packages/@n8n/api-types/src/dto/project/list-projects-query.dto.ts
+++ b/packages/@n8n/api-types/src/dto/project/list-projects-query.dto.ts
@@ -1,9 +1,22 @@
 import { z } from 'zod';
 
 import { Z } from '../../zod-class';
-import { MAX_ITEMS_PER_PAGE, paginationSchema } from '../pagination/pagination.dto';
+import { MAX_ITEMS_PER_PAGE } from '../pagination/pagination.dto';
 
 const searchValidator = z.string().trim().optional();
+
+// Custom skip validator that defaults to undefined (not 0) for backward
+// compatibility — existing callers that omit pagination params receive a bare array.
+const skipValidator = z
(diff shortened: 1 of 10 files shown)


packages/@n8n/api-types/src/dto/project/list-projects-query.dto.ts
```diff
@@ -0,0 +1,49 @@
+import { z } from 'zod';
+
+import { Z } from '../../zod-class';
+import { MAX_ITEMS_PER_PAGE } from '../pagination/pagination.dto';
+
+const searchValidator = z.string().trim().optional();
+
+// Custom skip validator that defaults to undefined (not 0) for backward
+// compatibility — existing callers that omit pagination params receive a bare array.
+const skipValidator = z
```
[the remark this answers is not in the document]
r00gm: Not a real issue for us. parseInt('10abc') → 10 is the same behavior used in the shared paginationSchema across the entire codebase. Fixing it only here would be inconsistent. This is a pre-existing pattern, not something we introduced.

packages/cli/test/integration/database/repositories/project.repository.test.ts
```diff
@@ -114,6 +114,230 @@ describe('ProjectRepository', () => {
...
+		});
+
+		it('does not produce duplicate rows from multiple project relations', async () => {
+			const owner = await createOwner();
+			const member = await createMember();
+
+			// Member has a relation to the team project (via linkUserToProject)
+			// This tests that the subquery deduplicates correctly
+			const team = await createTeamProject('Shared Team', owner);
+			await linkUserToProject(member, team, 'project:editor');
```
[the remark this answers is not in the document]
r00gm: fixed https://github.com/n8n-io/n8n/pull/27036/changes/5b7cd4c1ac2b2e15a1ddbd287ecf15012b12cdc1

packages/@n8n/db/src/repositories/project.repository.ts
```diff
@@ -47,10 +47,123 @@ export class ProjectRepository extends Repository<Project> {
...
+	): Promise<[Project[], number]> {
+		// Build a subquery that finds the IDs of accessible projects, avoiding
+		// duplicate rows from the LEFT JOIN on projectRelations which would
+		// produce incorrect counts with getManyAndCount().
+		const idsQuery = this.createQueryBuilder('p')
+			.select('DISTINCT p.id', 'id')
+			.leftJoin('p.projectRelations', 'pr')
+			.where(
+				new Brackets((qb) => {
+					qb.where('p.type = :personalType', { personalType: 'personal' }).orWhere(
```
afitzek, line 72: Just to be sure, we allow users to see all personal projects right?

packages/@n8n/db/src/repositories/project.repository.ts
```diff
@@ -47,10 +47,123 @@ export class ProjectRepository extends Repository<Project> {
...
+					qb.where('p.type != :personalTypeFilter', {
+						personalTypeFilter: 'personal',
+					}).orWhere('creator.password IS NOT NULL');
+				}),
+			);
+		}
+
+		const query = this.createQueryBuilder('project')
+			.leftJoin('project.creator', 'creator')
+			.where(`project.id IN (${idsQuery.getQuery()})`);
```
afitzek, line 101: Is there an option to use the subquery without a string interpolation?

afitzek (approved):
Just two small questions, but feel free to ignore them.

packages/@n8n/db/src/repositories/project.repository.ts
r00gm: yes, it's needed for sharing with other users

packages/@n8n/db/src/repositories/project.repository.ts
r00gm: AFAIK

No, TypeORM doesn’t offer a way to avoid string interpolation for `WHERE IN (subquery)`. The getQuery() + setParameters() pattern is the standard approach. It’s also what `execution.repository.ts `uses. The interpolated string is safe because `getQuery()` returns generated SQL (not user input), and all user values go through parameterized `setParameters()`.

Issue #22452, opened 2025-11-28
When placing custom nodes into the ~/.n8n/custom folder, the node icon cannot be displayed.


ichat001:
### Bug Description

When placing custom nodes into the ~/.n8n/custom folder, the node icon cannot be displayed. As shown in the image below

<img width="759" height="386" alt="Image" src="https://github.com/user-attachments/assets/b44a9016-7cfb-417b-ba51-bd69447ce8a4" />

The incorrect generated path is:
**Docker:**
/icons/CUSTOM//home/node/.n8n/custom/n8n-nodes-pwsh/dist/Pwsh/pwsh.svg
**Windows:**
/icons/CUSTOM/C:\Users\Administrator\.n8n\custom\n8n-nodes-pwsh\dist\Pwsh\pwsh.svg

The correct path should be:
/icons/CUSTOM/n8n-nodes-pwsh/dist/Pwsh/pwsh.svg

It seems that the `getIconPath` function in the following link has an issue:
https://github.com/n8n-io/n8n/blob/master/packages/core/src/nodes-loader/directory-loader.ts#L397

### To Reproduce

1.Copy any community node into the ~/.n8n/custom folder.

2.Restart n8n.

3.Locate the node and observe its icon.

### Expected behavior

It should display correctly using the proper path.

### Debug Info

no

### Operating System

linux 22.04, windows10

### n8n Version

1.122.3

### Node.js Version

22.16

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

Joffcom:
Hey @ichat001 

I am fairly sure we have another issue raised for this, I will dig it out later and get them linked.

ichat001:
> Hey [@ichat001](https://github.com/ichat001)
> 
> I am fairly sure we have another issue raised for this, I will dig it out later and get them linked.


https://github.com/n8n-io/n8n/blob/master/packages/core/src/nodes-loader/directory-loader.ts#L406
I tried modifying the code on line 406. It seems to have resolved the issue, but it’s not very elegant.

Previously:

```javascript
return `icons/${this.packageName}/${iconPath}`;
```

After:

```javascript
return `icons/${this.packageName}/${path.relative(this.directory, iconPath).replace(/\\/g, '/')}`;
```

PR #27056, opened 2026-03-15
fix(core): Fix custom node icon path resolution

### LINKED ISSUE #26882: Node icon broken when placing custom node into ~/.n8n/custom folder
### Bug Description

With at least N8N version 2.11.2, bug #22452 is back: The icons of nodes put into `~/.n8n/custom` cannot be loaded.

The icon path is e.g. `http://localhost:5678/icons/CUSTOM//home/cweiske/.n8n/custom/node_modules/n8n-nodes-nextcloud-tables/dist/nodes/NextcloudTables/nextcloud-tables.svg`

This was supposed to be fixed with #23029 but apparently got broken again.

### To Reproduce

1. Install custom node with: `cd ~/.n8n/custom && npm link n8n-nodes-nextcloud-tables`
2. Restart n8n
3. Edit workflow, add new node. Search for "nextcloud tables"
4. See the icon is broken

### Expected behavior

Icon shows

### Debug Info

# Debug info

## core

- n8nVersion: 2.11.2
- nodeJsVersion: 24.14.0
- nodeEnv: undefined
- database: sqlite
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (x11; linux x86_64; rv:148.0) gecko/20100101 firefox/148.0
- isTouchDevice: false

Generated at: 2026-03-11T10:04:34.800Z

### Operating System

Debian 13

### n8n Version

2.11.2

### Node.js Version

24.14.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

mvanhorn:
## Summary

Custom nodes installed in `~/.n8n/custom` have broken icons in v2.11.2+. This is a regression of issue #22452, which was previously fixed in PR #23029.

**Root cause:** `DirectoryLoader.getIconPath()` in `packages/core` constructs icon URLs by embedding the raw filesystem path. When `CustomDirectoryLoader` loads nodes using `fast-glob` with `absolute: true`, the `filePath` is an absolute path like `/home/user/.n8n/custom/node_modules/pkg/dist/nodes/Name/NodeName.node.js`. The resulting icon URL becomes `/icons/CUSTOM//home/user/.n8n/custom/...` — an absolute path with a leading double slash.

**Fix:**

1. **`packages/core/src/nodes-loader/directory-loader.ts`** - `getIconPath()` now computes a relative path from `this.directory` before embedding it in the icon URL. The security check (`isContainedWithin`) still validates against the absolute path.

2. **`packages/cli/src/load-nodes-and-credentials.ts`** - `resolveIcon()` now first tries to resolve the URL path as a relative path (the new format). If the resolved path is not within the loader directory (old absolute-path format), it falls back to the legacy behavior for backward compatibility.

**Result:** Custom node icon URLs will be well-formed relative paths, e.g. `/icons/CUSTOM/node_modules/pkg/dist/nodes/Name/icon.svg`, resolving correctly whether accessed directly or through a reverse proxy.

Fixes #26882

This contribution was developed with AI assistance (Claude Code).


packages/core/src/nodes-loader/directory-loader.ts
```diff
@@ -419,13 +419,18 @@ export abstract class DirectoryLoader {
 	private getIconPath(icon: string, filePath: string) {
 		const iconPath = path.join(path.dirname(filePath), icon.replace('file:', ''));
 
-		if (!isContainedWithin(this.directory, path.join(this.directory, iconPath))) {
+		// Resolve to absolute path for security validation (relative to loader directory, not CWD)
+		const absoluteIconPath = path.resolve(this.directory, iconPath);
```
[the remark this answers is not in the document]
mvanhorn: Tests for this already exist in this PR - see `should produce relative icon URLs when filePath is absolute (custom node regression)` at line 801 and `should produce relative iconBasePath when filePath is absolute and icon uses expression` at line 823.

garritfra (commented):
Thanks for tracking this regression down and providing a well-structured fix — the root cause explanation is clear and the two-part approach (fix the path at the source in `directory-loader.ts`, keep a legacy fallback in `resolveIcon`) is the right way to handle backward compat.

One thing missing before we can merge: a test in `load-nodes-and-credentials.test.ts` for the **new relative-URL format**. The existing `N8N_CUSTOM_EXTENSIONS` tests only exercise the old malformed double-slash URLs (`/icons/CUSTOM//home/...`). After your fix, `directory-loader.ts` will emit clean relative URLs like `/icons/CUSTOM/node_modules/pkg/dist/nodes/Name/icon.svg`, and there's currently no test that confirms `resolveIcon` maps those back to the correct absolute filesystem path.

Something along these lines would cover it:

```typescript
it('should resolve new relative-path custom icon URL correctly', () => {
  const result = instanceCustom.resolveIcon(
    packageNameCustom,
    `${pathPrefixCustom}/node_modules/some-pkg/dist/nodes/MyNode/icon.svg`,
  );
  expect(result).toBe(`${dirCustom}/node_modules/some-pkg/dist/nodes/MyNode/icon.svg`);
});
```

Once that's in and CI is green, this looks good to merge.

mvanhorn:
Thanks @garritfra. Pushed 640e0a66 with two changes:

1. Added the relative-path custom icon URL test you suggested in `packages/cli/src/__tests__/load-nodes-and-credentials.test.ts` using the existing `instanceCustom` / `packageNameCustom` / `pathPrefixCustom` / `dirCustom` fixtures.
2. Updated the expression-icon fixture in `packages/core/src/nodes-loader/__tests__/directory-loader.test.ts:828` so it satisfies the current `Icon` type. The old `'=file:{{$parameter.icon}}.svg'` literal stopped matching `Icon | undefined` after upstream tightened the type; the new `'={{"file:" + $parameter.icon + ".svg"}}'` form is a valid `isExpression()`-prefixed value that yields the same final URL.

E2E shard failures (token-exchange, OIDC, log-streaming) appear to be pre-existing flakes unrelated to this PR. Ready for re-review once the Typecheck and Backend Unit Tests jobs come back green.

garritfra:
Thanks you! Yeah, the E2E tests are a bit flaky at the moment, I'll take care of that

mvanhorn:
Thanks for picking up the E2E side! Ping me if a re-run on this PR's branch helps once the flakiness is smoothed.

Qodo-Free-For-OSS:
Hi, DirectoryLoader now builds `iconUrl`/`iconBasePath` using `path.relative()`, which returns backslash-separated paths on Windows. Those backslashes get embedded into URL strings and can cause the `/icons/...` route to not match, resulting in missing icons (404).

**Severity:** action required | **Category:** reliability

**How to fix:** Normalize URL path separators

> ## Issue description
> `path.relative()` returns OS-specific separators. On Windows, it emits backslashes, which are then embedded in `iconUrl` and `iconBasePath` and used as HTTP URL paths. This can break `/icons/...` route matching and lead to 404 icons.
> 
> ## Issue Context
> - `DirectoryLoader.getIconPath()` returns `icons/${packageName}/${relativeIconPath}`.
> - `fixIconPaths()` sets `iconBasePath` similarly for expression icons.
> - These values are later used as URL paths in the frontend and served by an Express `/icons/...` route.
> 
> ## Fix Focus Areas
> - packages/core/src/nodes-loader/directory-loader.ts[419-433]
> - packages/core/src/nodes-loader/directory-loader.ts[448-452]
> 
> ## Implementation notes
> - After computing `relativeIconPath` / `relativeDirPath`, replace backslashes with forward slashes before embedding in the URL:
>   - `relativeIconPath.replace(/\\/g, '/')`
>   - `relativeDirPath.replace(/\\/g, '/')`
> - Consider adding/adjusting unit tests to cover Windows-style paths (e.g. using `path.win32` inputs or conditional assertions).

---
*Spotted by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source projects.*

packages/cli/src/load-nodes-and-credentials.ts
```diff
@@ -223,14 +223,29 @@ export class LoadNodesAndCredentials {
...
+		if (isCustom) {
+			const relativePath = resolvePath(urlFilePath);
+			if (isContainedWithin(loader.directory, relativePath)) {
+				filePath = relativePath;
+			} else {
+				filePath = resolvePathCustomLegacy(urlFilePath);
+			}
+		} else {
+			filePath = resolvePath(urlFilePath);
+		}
```
garritfra, line 256: This block is what's failing the existing `should return file path if url contains "C:" with absolute custom windows file path` test on Linux CI.

Root cause: on Linux, `path.resolve(loader.directory, "C:/…")` doesn't recognise `C:` as a drive letter and prefixes both sides with `process.cwd()`, so `isContainedWithin` falsely reports "contained" and we skip the legacy branch.

Worth flagging: the `isContainedWithin`-as-router heuristic is almost always `true` (it only fails for `..` traversals), so it doesn't really distinguish legacy from new URLs. Naming the actual distinction reads cleaner and avoids this bug class:

```suggestion
		// Custom nodes used to encode absolute filesystem paths in icon URLs,
		// producing malformed URLs like `/icons/CUSTOM//home/user/.n8n/...`.
		// They now use relative paths (fixed in directory-loader.ts), but we
		// keep the legacy absolute-path fallback for backward compatibility.
		let filePath: string;
		if (isCustom) {
			const isLegacyAbsolute = isWindowsFilePath(urlFilePath) || urlFilePath.startsWith('/');
			filePath = isLegacyAbsolute ? resolvePathCustomLegacy(urlFilePath) : resolvePath(urlFilePath);
		} else {
			filePath = resolvePath(urlFilePath);
		}
```

The final `isContainedWithin` at the end of `resolveIcon` still guards against path traversal, so the security check is unchanged.

mvanhorn: Pushed 8fdf6c0e:

1. Replaced the `isContainedWithin`-as-router heuristic with explicit `isLegacyAbsolute` detection (`isWindowsFilePath || startsWith('/')`) in `load-nodes-and-credentials`, matching your suggestion.

2. Normalized backslashes from `path.relative()` before embedding in URL strings in `directory-loader.ts` (Qodo also flagged this), so Windows-emitted backslashes don't leak into HTTP URL paths.

3. Added regression coverage for Windows-style separators in `directory-loader.test`.

Local `pnpm install` hit a registry timeout in this environment, so I haven't run the test locally; CI will validate. Happy to fix-forward if it surfaces anything.

garritfra (approved):
LGTM!

mvanhorn:
Approval received, thanks @garritfra. Let me know if a rebase is needed before merge lands.

PR #27064, opened 2026-03-15
fix(core): Reject reserved system column names when creating data tables

RicardoE105:
## Summary

CSV import of a data table fails with a vague "There is already an entry with this name" error when the CSV contains a column named "id", "ID", "createdAt", or "updatedAt". These names conflict with system columns that are automatically created for every data table.

**Changes:**
- Add case-insensitive system column name validation in `DataTableRepository.createDataTable()` — this was missing entirely, so the error only surfaced as a raw DB constraint violation
- Fix existing `validateNotSystemColumn()` in `DataTableColumnRepository` to be case-insensitive (it only matched exact case before)
- Fix frontend `AddDataTableModal` validation to also use case-insensitive comparison, so users see the error in the preview before submitting

**How to test:**
1. Create a CSV file with a column named "id" or "ID"
2. Go to Overview > Data tables > Create data table
3. Click Import CSV, check "My CSV file contains a header row"
4. Verify the column shows a clear validation error about reserved system column names
5. Verify creating a table with columns named "createdAt", "updatedAt" (any case) also shows the error

https://linear.app/n8n/issue/ADO-4946

## Review / Merge checklist

- [x] Tests included.


CharlieKolb (commented):
I can create and use a data table with a `CreatedAt` column on internal. Are you sure this issue isn't in the CSV logic somewhere?

RicardoE105:
@CharlieKolb Yeah, that was my first instinct too but it's actually a SQLite-only thing. Internal runs Postgres, where quoted `"CreatedAt"` and `"createdAt"` are different identifiers, so both columns happily coexist (I spun up pg 16 locally to double check). On SQLite the parser treats column identifiers case insensitively, so the CREATE TABLE blows up at parse time with `duplicate column name: Id` once you add a user column like `Id` next to the lowercase system `id` PK + `createdAt`/`updatedAt` from `withTimestamps` (`data-table-ddl.service.ts:33`).

So it's not really a CSV thing, manual creation with `Id` would fail the same way on SQLite, CSV is just where folks hit it most often. Going case insensitive on validation keeps the behavior the same across engines.

CharlieKolb (approved):

PR #27066, opened 2026-03-15
perf(core): Make Wait node fully durable by removing in-memory execution path

shortstacked:
## Summary

Removes the dual-execution-path behaviour from the Wait node. Previously, waits shorter than 65 seconds ran entirely in-memory via `setTimeout` and were never persisted to the database. This made them invisible to crash recovery, multi-main failover, and the WaitTracker entirely.

**What changed:**
- **Wait node** (`Wait.node.ts`): removed the `< 65s` in-memory branch. All time-based waits now call `putToWait` immediately, regardless of duration.
- **ExecutionRepository** (`execution.repository.ts`): `getWaitingExecutions()` now uses a DB-server-clock-anchored 15-second lookahead window (`NOW() + INTERVAL '15 seconds'` / `datetime('now', '+15 seconds')`) via `createQueryBuilder`. Added `getServerTime()` to fetch the DB server's current timestamp (PostgreSQL: `CURRENT_TIMESTAMP(3)`, SQLite: `STRFTIME`).
- **WaitTracker** (`wait-tracker.ts`): poll interval reduced from 60s → 5s. `triggerTime` is now computed relative to the DB server clock (via a 60s-TTL cache with elapsed-time interpolation) rather than `Date.now()`, eliminating inter-instance clock skew from timer precision. Logs a warning when skew exceeds 2s.
- **PrometheusMetricsService** (`prometheus-metrics.service.ts`): added `n8n_db_clock_skew_ms` gauge, scraped live on each Prometheus pull.

**Why:** The 65s threshold existed because the old 60s poll interval made DB-persisted short waits resume late. Reducing the poll to 5s and adding a 15s lookahead window eliminates the need for the in-memory path entirely. The trade-off is up to ~5s of jitter on short waits in exchange for full crash/restart durability.

**Blast radius:** Narrow — only affects time-based Wait node resume and WaitTracker scheduling. No schema changes, no API changes. Safe to revert with a single commit revert; in-flight waiting executions survive revert (they resume ~60s late via the old poll cycle).

**How to test:**
1. Create a workflow: Manual Trigger → Wait (15s) → Set node
2. Execute — verify execution enters `waiting` status in DB immediately (not after 15s)
3. Verify it resumes at ~15s (±5s acceptable)
4. Kill n8n mid-wait, restart — verify it resumes after restart
5. Scrape `/metrics` and confirm `n8n_db_clock_skew_ms` gauge is present

## Review / Merge checklist

- [ ] Tests included.


SlugVortex:
## :white_check_mark: DriftWatch Risk Analysis — APPROVED

| Metric | Value |
|--------|-------|
| **Risk Score** | **46/100** (Medium) |
| **Blast Radius** | 13 files, 3 services (score: 31) |
| **Deployment Weather** | :partly_sunny: 20/100 |
| **Pipeline Duration** | 29.3s |

### :dart: Bottom Line
Moderate risk — review recommended before deploying.

### :speech_balloon: Negotiator Says
> ### Deployment Risk Assessment
> 
> **Risk Score: 46/100** | **Decision: APPROVED**
> 
> ---
> 
> #### Code Analysis Summary
> 
> - Core execution scheduling logic updated with database query logic changes
> - New error handling added for database server time retrieval
> - WaitTracker service enhanced to use DB server time and detect clock skew
> - Wait node code removal could affect workflow timing behavior
> - Extensive new and enhanced tests improve reliability and coverage
> 
> #### Recommendation
> 
> Monitor execution sc

### :mag: Key Risk Factors
- **Unknown**
- **Unknown**
- **Unknown**
- **Unknown**
- **Unknown**

---
:robot: *Analyzed by [DriftWatch](https://github.com) — Multi-Agent Pre-Deployment Risk Intelligence*

SlugVortex:
## :white_check_mark: DriftWatch Risk Analysis — APPROVED

| Metric | Value |
|--------|-------|
| **Risk Score** | **40/100** (Medium) |
| **Blast Radius** | 13 files, 3 services (score: 30) |
| **Deployment Weather** | :sunny: 0/100 |
| **Pipeline Duration** | 27.9s |

### :dart: Bottom Line
Moderate risk — review recommended before deploying.

### :speech_balloon: Negotiator Says
> ### Deployment Risk Assessment
> 
> **Risk Score: 40/100** | **Decision: APPROVED**
> 
> ---
> 
> #### Code Analysis Summary
> 
> - Core DB repository logic changed with new DB server time handling
> - WaitTracker service timing logic changed to use DB server time with caching and skew detection
> - Test coverage increased extensively but some test data removed
> - No historical incidents matched files or services in this PR
> 
> #### Recommendation
> 
> Monitor closely in staging and early production deployments due to chan

### :mag: Key Risk Factors
- **Unknown**
- **Unknown**
- **Unknown**
- **Unknown**

---
:robot: *Analyzed by [DriftWatch](https://github.com) — Multi-Agent Pre-Deployment Risk Intelligence*

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -586,28 +586,38 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+
+		const lookaheadCondition =
+			dbType === 'postgresdb'
+				? "e.waitTill <= NOW() + INTERVAL '15 seconds'"
+				: "e.waitTill <= datetime('now', '+15 seconds')";
+
+		return await this.createQueryBuilder('e')
+			.select(['e.id', 'e.waitTill'])
+			.where(lookaheadCondition)
+			.andWhere('e.status = :status', { status: 'waiting' })
```
tomi, line 617: The behaviour is different now. Before it was `NOT IN ('crashed')`

packages/@n8n/config/src/configs/endpoints.config.ts
```diff
@@ -79,6 +79,10 @@ class PrometheusMetricsConfig {
 	/** How often (in seconds) to update workflow statistics metrics. */
 	@Env('N8N_METRICS_WORKFLOW_STATISTICS_INTERVAL')
 	workflowStatisticsInterval: number = 300;
+
+	/** Whether to expose a gauge for the clock difference between this instance and the database server. */
+	@Env('N8N_METRICS_INCLUDE_DB_CLOCK_SKEW_METRIC')
+	includeDbClockSkewMetric: boolean = false;
```
tomi: I don't think this is really needed for this feature? Can we move the gauge addition to a separate PR?

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -586,28 +586,38 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
+			.orderBy('e.waitTill', 'ASC')
+			.getMany();
+	}
 
+	async getServerTime(): Promise<Date> {
```
tomi: I don't think this file is the correct place for this. I would move it to `time.repository.ts` or `clock.repository.ts`

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -586,28 +586,38 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
...
-			where,
-			order: {
-				waitTill: 'ASC',
-			},
-		});
+		// SQLite returns a string in 'YYYY-MM-DD HH:MM:SS.SSS' format
+		const [{ now }] = (await this.query("SELECT STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW') AS now")) as [
+			{ now: string },
+		];
+		const date = new Date(now.replace(' ', 'T') + 'Z');
```
tomi: Since we control the format in `STRFTIME` we can add these there so we don't need to format them on the JS side

packages/cli/src/wait-tracker.ts
```diff
@@ -46,20 +49,49 @@ export class WaitTracker {
...
+	/**
+	 * Returns an approximation of the DB server's current time.
+	 * Fetches from the DB at most once every 60s and approximates intermediate
+	 * values by adding elapsed local wall-clock time since the last fetch.
+	 *
+	 * The clock skew warning reflects the skew at the last cache refresh rather
+	 * than live skew — intentional, to detect structural drift (e.g. broken NTP)
+	 * rather than transient noise.
+	 */
+	private async getApproximateServerTime(): Promise<Date> {
```
tomi: This should be its own class instead here

packages/cli/src/wait-tracker.ts
```diff
@@ -46,20 +49,49 @@ export class WaitTracker {
...
+		if (this.serverTimeCache !== null && nowMs - this.serverTimeCache.localTimeAtFetch < 60_000) {
+			const elapsed = nowMs - this.serverTimeCache.localTimeAtFetch;
+			return new Date(this.serverTimeCache.serverTime.getTime() + elapsed);
+		}
+		const serverTime = await this.executionRepository.getServerTime();
+		this.serverTimeCache = { serverTime, localTimeAtFetch: nowMs };
+		return serverTime;
+	}
+
 	async getWaitingExecutions() {
```
tomi, line 64: This runs on an interval so there can be two different instance of this function execution at the same time. It would much easier to reason about this if we didn't allow concurrent executions of this function

packages/cli/src/wait-tracker.ts
```diff
@@ -46,20 +49,49 @@ export class WaitTracker {
 
 	@OnLeaderTakeover()
 	private startTracking() {
-		// Poll every 60 seconds a list of upcoming executions
 		this.mainTimer = setInterval(() => {
 			void this.getWaitingExecutions();
-		}, 60000);
+		}, 5000);
```
tomi, line 57: Now that we poll more frequently, there is a much higher likelihood that this can lead to duplicate execution continuations. The main protection we have is checking if `this.waitingExecutions[executionId] !== undefined`. Since `startExecution` removes the item from that map at the beginning before querying it from DB, there's a chance that we end up scheduling the same execution twice. We need to protect against taht

packages/cli/src/wait-tracker.ts
```diff
@@ -70,19 +102,21 @@ export class WaitTracker {
...
 			const executionId = execution.id;
-			if (this.waitingExecutions[executionId] === undefined) {
-				const triggerTime = execution.waitTill!.getTime() - new Date().getTime();
-				this.waitingExecutions[executionId] = {
-					executionId,
-					timer: setTimeout(() => {
+			if (this.waitingExecutions[executionId] !== undefined) continue;
+			if (execution.waitTill === null) continue; // row without a waitTill should never be returned, skip defensively
+
+			const triggerTime = execution.waitTill.getTime() - serverTime.getTime();
```
tomi: I'm not sure how accurate this actually is. `this.executionRepository.getWaitingExecutions()` can take a long time, for example couple seconds. The DB time is fetched concurrently and if it is X, when we reach this line the actual server time is X+2s. This means we would continue the execution 2s late. Is it a big deal? Probably not. However would be nice to use a more robust mechanism for this

tomi (changes requested):

packages/@n8n/config/src/configs/endpoints.config.ts
shortstacked: Done!

packages/@n8n/db/src/repositories/execution.repository.ts
shortstacked: Intentional narrowing. determineFinalExecutionStatus() in shared-hook-functions.ts always sets status to 'waiting' when waitTill is set — there’s no code path that writes an execution to the DB with a waitTill date and a different status. The old NOT IN ('crashed') was overly broad and could theoretically match stale rows with status error/running that still had a waitTill set, which would be incorrect to resume.

packages/cli/src/wait-tracker.ts
shortstacked: Fixed — added isPolling guard with try/finally. Skips the poll if the previous one hasn’t resolved yet. Only the leader polls (multi-main is handled by @OnLeaderTakeover/@OnLeaderStepdown), so this is a single-instance concern.

packages/cli/src/wait-tracker.ts
shortstacked: Mitigated by DbClock — server time is fetched independently via Promise.all alongside the execution query, with 60s cache + wall-clock interpolation. Also fixed localTimeAtFetch to capture after the DB round-trip so the approximation doesn’t run ahead by query latency.

packages/@n8n/db/src/repositories/clock.repository.ts
```diff
@@ -0,0 +1,31 @@
+import { GlobalConfig } from '@n8n/config';
+import { Service } from '@n8n/di';
+import { DataSource } from '@n8n/typeorm';
+import { UnexpectedError } from 'n8n-workflow';
+
+@Service()
```
tomi, line 7: Can we a concise JSDoc description of this repository

packages/@n8n/db/src/repositories/clock.repository.ts
```diff
@@ -0,0 +1,31 @@
...
+export class ClockRepository {
+	constructor(
+		private readonly dataSource: DataSource,
+		private readonly globalConfig: GlobalConfig,
```
tomi: We can use DatabaseConfig directly, DI is able inject also subconfigs

packages/@n8n/db/src/repositories/clock.repository.ts
```diff
@@ -0,0 +1,31 @@
...
+	) {}
+
+	async getServerTime(): Promise<Date> {
```
tomi: Since it gets the _DB_ server time, maybe getDbTime would be more approriate? Server can be ambiguous

packages/@n8n/db/src/repositories/clock.repository.ts
```diff
@@ -0,0 +1,31 @@
...
+		if (this.globalConfig.database.type === 'postgresdb') {
+			const [{ now }] = await this.dataSource.query<[{ now: Date }]>(
+				'SELECT CURRENT_TIMESTAMP(3) AS now',
+			);
+			return new Date(now.getTime());
```
tomi: Why this double wrapping if `now` is already a `Date`?

packages/@n8n/db/src/repositories/__tests__/execution.repository.test.ts
```diff
@@ -406,4 +408,59 @@ describe('ExecutionRepository', () => {
 			expect(result.data.resultData.error?.message).toBe('The execution was cancelled manually');
 		});
 	});
+
+	describe('getWaitingExecutions()', () => {
```
tomi, line 412: I feel an integration test would be more meaningful way to test this method

packages/cli/src/services/db-clock.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Provides an approximation of the DB server's current time.
+ *
+ * Fetches from the DB at most once every 60s and approximates intermediate
+ * values by adding elapsed local wall-clock time since the last fetch.
+ */
+@Service()
+export class DbClock {
+	private cache: { serverTime: Date; localTimeAtFetch: number } | null = null;
```
tomi: nit: Fetch is often linked with HTTP request. Maybe this could be named `localTimeAtQuery`?

packages/cli/src/services/db-clock.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+	constructor(private readonly clockRepository: ClockRepository) {}
+
+	async getApproximateServerTime(): Promise<Date> {
```
tomi: Same here, server can be ambiguous. Maybe `getApproximateDbTime`?

packages/cli/src/services/db-clock.service.ts
```diff
@@ -0,0 +1,30 @@
...
+		const nowMs = Date.now();
+		if (this.cache !== null && nowMs - this.cache.localTimeAtFetch < 60_000) {
+			const elapsed = nowMs - this.cache.localTimeAtFetch;
+			return new Date(this.cache.serverTime.getTime() + elapsed);
+		}
+		const serverTime = await this.clockRepository.getServerTime();
+		this.cache = { serverTime, localTimeAtFetch: Date.now() };
```
tomi: nit: could also be `setCache(serverTime)`

packages/cli/src/services/db-clock.service.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+@Service()
+export class DbClock {
+	private cache: { serverTime: Date; localTimeAtFetch: number } | null = null;
+
+	constructor(private readonly clockRepository: ClockRepository) {}
+
+	async getApproximateServerTime(): Promise<Date> {
+		const nowMs = Date.now();
+		if (this.cache !== null && nowMs - this.cache.localTimeAtFetch < 60_000) {
```
tomi: nit: Could also be `if (!isCacheStale(nowMs))` to be a bit more readable

packages/cli/src/services/db-clock.service.ts
tomi: This doesn't now take network latency into account. If there's significant lag for the communication with the DB, let's say 3s, the cached time is 3s off. Now it's not the _that_ bad, the timers only fire later than they should. It does however cause the skew logging, which is not really correct since it's due to latency and not due to clocks having different values

We could try to mitigate this by:
account for round-trip time by stamping Date.now() both before and after the query, then offsetting serverTime by half the round-trip.

It's still not perfect, but I don't think there is a better way when your using a remote clock

packages/cli/src/wait-tracker.ts
```diff
@@ -46,43 +51,64 @@ export class WaitTracker {
...
+
+			const executionIds = executions.map((execution) => execution.id).join(', ');
+			this.logger.debug(
+				`Found ${executions.length} executions. Setting timer for IDs: ${executionIds}`,
+			);
+
+			for (const execution of executions) {
+				const executionId = execution.id;
+				if (this.waitingExecutions[executionId] !== undefined) continue;
+				if (execution.waitTill === null) continue; // row without a waitTill should never be returned, skip defensively
```
tomi: Can we add error reporting for this and report it as `critical`? This is a state that should never happen

packages/cli/src/wait-tracker.ts
```diff
@@ -96,22 +122,23 @@ export class WaitTracker {
 
 	async startExecution(executionId: string) {
 		this.logger.debug(`Resuming execution ${executionId}`, { executionId });
-		delete this.waitingExecutions[executionId];
 
-		// Get the data to execute
 		const fullExecutionData = await this.executionRepository.findSingleExecution(executionId, {
```
tomi: What happens if this now fails? There is an entry in `this.waitingExecutions` so it will never be picked up again by the polling. The timer also never fires since its oneshot only. Basically the execution is now stuck forever.

One option would be to wrap this whole thing in a try..finally

packages/cli/src/wait-tracker.ts
```diff
@@ -175,6 +199,8 @@ export class WaitTracker {
 		Object.keys(this.waitingExecutions).forEach((executionId) => {
 			clearTimeout(this.waitingExecutions[executionId].timer);
 		});
+		this.dbClock.resetCache();
+		this.isPolling = false;
```
tomi: I feel both of these are unnecessary. The DBClock cache should be an internal implementation detail that outsiders shouldnt need to be concerned with.

Resetting the internal poll state also feels wrong since it should only managed by the poll implementation. If we reset it here and for whatever reason it repolls, it can cause overlapping polls. Better to have the poll method manage its state itself

tomi (changes requested):

packages/@n8n/db/src/repositories/__tests__/execution.repository.test.ts
shortstacked: Added one

packages/cli/src/wait-tracker.ts
```diff
@@ -46,43 +52,73 @@ export class WaitTracker {
...
+			}
+
+			if (executions.length === 0) {
+				return;
+			}
 
-		// Add timers for each waiting execution that they get started at the correct time
+			const executionIds = executions.map((execution) => execution.id).join(', ');
+			this.logger.debug(
+				`Found ${executions.length} executions. Setting timer for IDs: ${executionIds}`,
```
tomi: This log can now feel a bit spammy cause we poll more often and don't ignore already schedule IDs:

```ts
12:07:19.587   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:07:19.589   debug   Found 1 executions. Setting timer for IDs: 520433 { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:07:24.588   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:07:24.594   debug   Found 1 executions. Setting timer for IDs: 520433 { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:07:29.589   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:07:29.591   debug   Found 1 executions. Setting timer for IDs: 520433 { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:07:32.102   debug   Resuming execution 520433 { "scopes": ["waiting-executions"], "executionId": "520433", "file": "wait-tracker.js", "function": "startExecution" }
```

I think we should filter out those IDs what have already been scheduled before logging and processing them

tomi (commented):
Thank you for addressing all the feedback 🫶 One final comment.

Also the implications of this are that short waits (1s) can now take anything between 1 and 5 seconds. I wouldn't it's necessarily the biggest of deals, but might come as a surprise to users.

packages/cli/src/wait-tracker.ts
tomi: Actually the whole polling is quite spammy now that it happens so often. We should remove this log

```
12:10:44.611   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:10:49.611   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:10:54.611   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:10:59.612   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:04.611   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:09.612   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:14.614   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:19.615   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:24.615   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:29.615   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
12:11:34.616   debug   Querying database for waiting executions { "scopes": ["waiting-executions"], "file": "wait-tracker.js", "function": "getWaitingExecutions" }
```

tomi (approved):
👏 🚀

PR #27089, opened 2026-03-16
fix(core): Handle invalid JSON responses from OAuth2 token endpoints

afitzek:
## Summary

Wraps `JSON.parse()` in the shared OAuth2 client library (`@n8n/client-oauth2`) with a try-catch so that when a token endpoint returns HTML (e.g. an error page, WAF block, or redirect) with a `Content-Type: application/json` header, a descriptive `ResponseError` is thrown instead of an uncaught `SyntaxError`.

Previously this surfaced as `"Unexpected token '<', "<!DOCTYPE "... is not valid JSON"`, which was confusing and provided no actionable context. The error message now includes a truncated preview of the unexpected response body.

https://linear.app/n8n/issue/IAM-379

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/client-oauth2/src/client-oauth2.ts
+++ b/packages/@n8n/client-oauth2/src/client-oauth2.ts
@@ -127,7 +127,17 @@ export class ClientOAuth2 {
 		const body = response.data as string;
 
 		if (contentType.startsWith('application/json')) {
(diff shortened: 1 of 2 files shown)


guillaumejacquart (approved):
good catch !

PR #27093, opened 2026-03-16
feat(editor): Use server-side search for project sharing dropdowns

r00gm:
## Summary

Converts all `ProjectSharing` callers to use the paginated `GET /projects` endpoint (added in #27036) with server-side search, replacing client-side filtering that loaded all projects upfront. This prevents UI freezes when large numbers of users (~10k) exist.

**Changes:**
- Add `searchProjects` API function and Pinia store method using `getFullApiResponse` with `take`/`search`/`activated` params
- Refactor `ProjectSharing` to use `useAsyncState` with `remote` + `remote-method` for debounced server-side search
- Show a "more results" indicator when the server has more results than displayed
- Refactor `ProjectMoveResourceModal` to use `ProjectSharing` instead of raw `N8nSelect` + `getAvailableProjects` (which loaded ALL projects)
- Update all other callers (`DeleteUserModal`, `CredentialSharing`, `InsightsDashboard`, `ResourceFiltersDropdown`, `MoveToFolderModal`, `SourceControlPushModal`, `ProjectDeleteDialog`, `ProjectSettings`, `WorkflowShareModal`) to remove client-side project lists and use server-side search

image

https://linear.app/n8n/issue/IAM-382

Depends on #27036 (merged)

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/WorkflowShareModal.ee.vue
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowShareModal.ee.vue
@@ -111,10 +111,6 @@ const workflowOwnerName = computed(() =>
 	workflowsEEStore.getWorkflowOwnerName(`${workflow.value.id}`),
 );
 
-const projects = computed(() =>
-	projectsStore.personalProjects.filter((project) => project.id !== workflow.value.homeProject?.id),
-);
-
 const numberOfMembersInHomeTeamProject = computed(() => teamProject.value?.relations.length ?? 0);
 
 const workflowRoleTranslations = computed(() => ({
@@ -225,8 +221,6 @@ const goToUpgrade = () => {
 
 const initialize = async () => {
 	if (isSharingEnabled.value) {
-		await projectsStore.getAllProjects();
-
 		// Fetch workflow if it exists and is not new
 		if (workflowsStore.isWorkflowSaved[workflow.value.id]) {
 			await workflowsListStore.fetchWorkflow(workflow.value.id);
@@ -290,7 +284,6 @@ watch(
 						<ProjectSharing
 							v-model="sharedWithProjects"
 							:home-project="workflow.homeProject"
-							:projects="projects"
 							:roles="workflowRoles"
 							:readonly="!workflowPermissions.share"
 							:static="isHomeTeamProject || !workflowPermissions.share"
--- a/packages/frontend/editor-ui/src/app/components/WorkflowShareModal.ee.vue
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowShareModal.ee.vue
@@ -19,7 +19,7 @@ import ProjectSharing from '@/features/collaboration/projects/components/Project
 import { useProjectsStore } from '@/features/collaboration/projects/projects.store';
 import type { ProjectSharingData, Project } from '@/features/collaboration/projects/projects.types';
 import { ProjectTypes } from '@/features/collaboration/projects/projects.types';
-import { createRemoteProjectSearch } from '@/features/collaboration/projects/projects.utils';
+import { useRemoteProjectSearch } from '@/features/collaboration/projects/projects.utils';
 import type { ProjectListItem } from '@/features/collaboration/projects/projects.types';
 import { useRolesStore } from '@/app/stores/roles.store';
 import { usePageRedirectionHelper } from '@/app/composables/usePageRedirectionHelper';
@@ -113,7 +113,7 @@ const workflowOwnerName = computed(() =>
 	workflowsEEStore.getWorkflowOwnerName(`${workflow.value.id}`),
 );
 
-const searchFn = createRemoteProjectSearch(projectsStore);
+const searchFn = useRemoteProjectSearch();
 const filterFn = (project: ProjectListItem) =>
 	project.type === 'personal' && project.id !== workflow.value.homeProject?.id;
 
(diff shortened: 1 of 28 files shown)


packages/frontend/editor-ui/src/app/components/WorkflowShareModal.ee.vue
```diff
@@ -111,9 +113,9 @@ const workflowOwnerName = computed(() =>
 	workflowsEEStore.getWorkflowOwnerName(`${workflow.value.id}`),
 );
 
-const projects = computed(() =>
-	projectsStore.personalProjects.filter((project) => project.id !== workflow.value.homeProject?.id),
-);
+const searchFn = createRemoteProjectSearch(projectsStore);
+const filterFn = (project: ProjectListItem) =>
+	project.type === 'personal' && project.id !== workflow.value.homeProject?.id;
```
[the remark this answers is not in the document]
r00gm: Valid finding, but out of scope for this PR. WorkflowShareModal is partially migrated to workflowDocumentStore (already uses it for scopes) but homeProject still reads from the workflow ref in ~10 places. The homeProject composable exists and is ready, completing the migration for this component is a follow-up task.

keep bringing those up.

packages/frontend/editor-ui/src/features/execution/insights/components/InsightsDashboard.vue
```diff
@@ -191,18 +192,17 @@ watch(
...
-		(project) => project.name && !emailPattern.test(project.name.trim()),
-	),
-);
+const searchFn = createAvailableProjectSearch(projectsStore);
+const filterFn = (project: ProjectListItem) =>
+	!!project.name && !emailPattern.test(project.name.trim());
+
+onBeforeMount(async () => {
+	// Load myProjects for members so createAvailableProjectSearch can filter locally
+	await projectsStore.getAvailableProjects();
```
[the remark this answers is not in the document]
r00gm: fixed: https://github.com/n8n-io/n8n/pull/27093/changes/98f08f210ab35754ad784dc0b7c757ba1adceb1b

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectSharing.vue
```diff
@@ -107,17 +130,37 @@ const projectIcon = computed<IconOrEmoji>(() => {
 	return defaultIcon;
 });
 
+// ── Search logic ──
+const executeSearch = async (query: string) => {
+	try {
+		const result = await props.searchFn(query);
```
afitzek, line 139: Could a race condition happen here? If we start multiple executeSearch runs we await the result and once it is back, we write the result of this query. If a different newer query was quicker, we would overwrite the results from the older query. Maybe `useDebounceFn` takes care of this, but even if it does, the onMounted call, would be in depended from this? Should we "store" the latest query id, and only write results of the current scope is still the latest id?

afitzek (commented):
I think this looks good, just one small comment, not sure if this could be an issue though.

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectSharing.vue
r00gm: agree, with the race condition and the fix. i'm implementing it now.

the right way of doing this would be with `AbortSignal` but i'll have to plum it through 4 layers... going with the quick fix

r00gm: done https://github.com/n8n-io/n8n/pull/27093/changes/0965cf42416b447a91510b17b0ca2d33b270cf6a

cstuncsik: Yes, classic situation
Need to cancel the in-progress request when firing a new

cstuncsik: (how easy would this be with observables ;) )

packages/frontend/editor-ui/src/features/collaboration/projects/projects.utils.ts
```diff
@@ -1,4 +1,50 @@
...
+export const DEFAULT_PROJECT_SEARCH_PAGE_SIZE = 50;
+
+export type ProjectSearchResult = { count: number; data: ProjectListItem[] };
+export type ProjectSearchFn = (query: string) => Promise<ProjectSearchResult>;
+
+/**
+ * Remote search for Group 1 consumers (sharing/transfer modals).
+ * Always searches via GET /projects?search=&take= for ALL roles.
+ */
+export function createRemoteProjectSearch(
```
cstuncsik: Just a nitpick: passing the store as an argument is unconventional for Vue + Pinia.

Consider making these composables (`useRemoteProjectSearch`, `useAvailableProjectSearch`) that call `useProjectsStore()` internally, which is more idiomatic and easier to test.

r00gm: done https://github.com/n8n-io/n8n/pull/27093/commits/a512981de82d93ab39aad9a94d24716551e2a790 https://github.com/n8n-io/n8n/pull/27093/commits/3137ea80afaab4ea6c680ee78ec22d3ca64682c8

cstuncsik (approved):

PR #27102, opened 2026-03-16
feat(core): Apply execution redaction to real-time push events (no-changelog)

afitzek:
## Summary

Applies the existing `ExecutionRedactionService` pipeline to WebSocket/SSE push events
(`nodeExecuteAfterData`, `executionStarted`) so that redaction policies are enforced in
real-time, not just on `GET /executions/:id`.

**Key design decisions:**

- **Copy-on-write redaction** — Adds a `copyOnWrite` option to `ExecutionRedactionOptions`.
  The service checks each strategy's new `wouldModify()` method *before* cloning. For the
  common case (`policy=none`, no sensitive output fields), zero copies are made.
- **Reuses existing pipeline** — Constructs a lightweight `RedactableExecution` from hook
  context and feeds it through the same `ExecutionRedactionService` strategies. No separate
  redaction logic for push events.
- **`wouldModify()` on strategy interface** — New required method on
  `IExecutionRedactionStrategy`. `FullItemRedactionStrategy` always returns `true` (if in
  pipeline, it mutates). `NodeDefinedFieldRedactionStrategy` checks `buildSensitiveFieldsMap`
  and returns `true` only when there are sensitive fields or unresolvable node types.
- **Lazy user resolution** — The `User` (with `role` relation) is resolved once per execution
  and cached across all node hooks in that execution.

https://linear.app/n8n/issue/IAM-364

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/cli/src/execution-lifecycle/__tests__/execution-lifecycle-hooks.test.ts
+++ b/packages/cli/src/execution-lifecycle/__tests__/execution-lifecycle-hooks.test.ts
@@ -1,6 +1,6 @@
 import { Logger } from '@n8n/backend-common';
 import { mockInstance } from '@n8n/backend-test-utils';
-import type { Project } from '@n8n/db';
+import type { Project, User } from '@n8n/db';
 import { ExecutionRepository, UserRepository } from '@n8n/db';
 import { mock } from 'jest-mock-extended';
 import {
@@ -10,6 +10,7 @@ import {
 	ExecutionLifecycleHooks,
 	BinaryDataConfig,
 } from 'n8n-core';
+import { stringify } from 'flatted';
 import { createRunExecutionData, ExpressionError } from 'n8n-workflow';
 import type {
 	IRunExecutionData,
@@ -56,8 +57,8 @@ describe('Execution Lifecycle Hooks', () => {
 	const binaryDataService = mockInstance(BinaryDataService);
 	const ownershipService = mockInstance(OwnershipService);
 	const workflowExecutionService = mockInstance(WorkflowExecutionService);
-	mockInstance(UserRepository);
-	mockInstance(ExecutionRedactionServiceProxy);
+	const userRepository = mockInstance(UserRepository);
+	const redactionProxy = mockInstance(ExecutionRedactionServiceProxy);
 
 	const nodeName = 'Test Node';
 	const nodeType = 'n8n-nodes-base.testNode';
@@ -142,6 +143,8 @@ describe('Execution Lifecycle Hooks', () => {
 
 	beforeEach(() => {
 		jest.clearAllMocks();
+		userRepository.findOne.mockResolvedValue(mock<User>());
+		redactionProxy.processExecution.mockImplementation(async (execution) => execution);
 		workflowData.settings = {};
 		successfulRun.data = createRunExecutionData({
 			resultData: {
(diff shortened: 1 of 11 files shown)


packages/cli/src/modules/redaction/executions/strategies/full-item-redaction.strategy.ts
```diff
@@ -17,6 +17,11 @@ import type {
 export class FullItemRedactionStrategy implements IExecutionRedactionStrategy {
 	readonly name = 'full-item-redaction';
 
+	wouldModify(_execution: RedactableExecution, _context: RedactionContext): boolean {
```
phyllis-noester: imo wouldModify is a confusing name. wouldModify what?

packages/cli/src/modules/redaction/executions/execution-redaction.interfaces.ts
```diff
@@ -12,6 +12,10 @@ export interface RedactionContext {
 	readonly redactExecutionData: boolean | undefined;
 	/** Pre-resolved by the orchestrator — true if the user can see unredacted data. */
 	readonly userCanReveal: boolean;
+	/** Generic memo store — strategies may cache intermediate results here
+	 *  to avoid redundant computation across wouldModify/apply calls.
+	 *  Keyed by strategy name. */
+	readonly memo: Map<string, unknown>;
```
phyllis-noester, line 18: shouldn't we use cache service for consistency?

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
```diff
@@ -48,13 +48,19 @@ export class ExecutionRedactionService implements ExecutionRedaction {
 
 	/**
 	 * Thin wrapper around `processExecutions` for single-execution callers.
+	 *
+	 * With `copyOnWrite: true`, the original execution is never mutated. Returns
```
phyllis-noester: I think copyOnWrite is confusing because it;s not intuitive that you copy something on write just for redaction. can we either make this more explicit in the naming or determine ad hoc based on redaction policy rather than this boolean?

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
```diff
@@ -106,17 +112,29 @@ export class ExecutionRedactionService implements ExecutionRedaction {
...
 				user: options.user,
 				redactExecutionData: options.redactExecutionData,
 				userCanReveal,
+				memo: new Map(),
 			};
 			const pipeline = this.buildPipeline(execution, context, policyAllowsReveal);
+
+			let target = execution;
+			if (options.copyOnWrite) {
+				const needsClone = pipeline.some((s) => s.wouldModify(execution, context));
```
phyllis-noester: same here I think wouldModify / needsclone is a bit confusing we should make it more explicit from naming or separate the clone related code out

packages/cli/src/modules/redaction/executions/execution-redaction.interfaces.ts
afitzek: No this is just a cache for the duration of the redaction execution it should only life for this duration as well. Also we never need to share it into a different service or instance. So I don't think the cache service makes sense here.

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
afitzek: Can you help me making this more intuitive with a different name? 

In my mind it makes sense as it is, but maybe I'm too deep in it. 😅 This is an argument for the the calling function, so the calling function can tell the redaction system, if the redaction system should create a copy if it needs to redact something. Basically this is a memory optimization, because the redaction system determines if a redaction is necessary, if it is not necessary we can safe the deep clone in this new use case that this PR is adding.

There are basically two cases (which are independent of the redaction policy):

1. We want to mutate this in place, this happens when we fetch an execution from the db to send it to the user. To safe memory, we don't want to copyOnWrite, it should mutate the data in place.

2. We are sending the execution to a user, but the execution data is still being used (because it is an active execution), so we don't want to mutate the execution data, because this is still in use. In this case we only want to create a copy in memory, if there are changes to the data.

packages/cli/src/modules/redaction/executions/strategies/full-item-redaction.strategy.ts
afitzek: This implements the interface function here: https://github.com/n8n-io/n8n/pull/27102/changes/BASE..9ccae32820c9a90e1b69afc063816ac4db3ac4c6#diff-8b5a5db5fa2bc80de6a617708aa4221a028f94822c5c1b6ea3cd0c87c18fb2d9R32

Does the comment there make it more clear or can you suggest a better name for it in this case? 😄 

`/** Returns true if apply() would mutate the execution data. Used by the
	 *  copy-on-write path to avoid unnecessary cloning. */`

packages/cli/src/modules/redaction/executions/execution-redaction.service.ts
afitzek: Can you be more specific here as well? I'm not sure I understand the confusion tbh. 😅

packages/cli/src/execution-lifecycle/__tests__/execution-lifecycle-hooks.test.ts
```diff
@@ -10,6 +10,7 @@ import {
 	ExecutionLifecycleHooks,
 	BinaryDataConfig,
 } from 'n8n-core';
+import { stringify } from 'flatted';
```
BGZStephen, line 13: What is this 👀, curious what the reasoning is for this method of serialisation / deserialisation?

afitzek: Historically all our workflow data and execution data is using this json format in the database.

BGZStephen (commented):
I'm not entirely sure how to test this, is this something you can provide guidance on testing to make sure it works?

phyllis-noester (approved):

PR #27107, opened 2026-03-16
feat: Enable secure invite links

BGZStephen:
## Summary

Signup and invitation acceptance now use a single **token** in the URL (e.g. `/signup?token=...`) and `POST /rest/invitations/accept` with `{ token, firstName, lastName, password }` instead of the legacy `inviterId`/`inviteeId` query params and `POST /rest/invitations/:inviteeId/accept`. This PR updates tests and E2E helpers to match that behavior so CI and local E2E runs pass.

**Changes:**

- **`packages/frontend/editor-ui/src/features/core/auth/views/SignupView.test.ts`**
  - Expect `route.query.token` and call `validateSignupToken({ token })` / `acceptInvitation` with token-based payloads.
  - Replace legacy inviterId/inviteeId tests with assertions that missing token shows error and redirects; add test for “inviterId but no token”.
  - Adjust “form unfilled” test to only assert `acceptInvitation` is not called (form validation may block submit).

- **`packages/testing/playwright/services/public-api-helper.ts`**
  - Read `token` from invite URL; POST to `/rest/invitations/accept` with `{ token, firstName, lastName, password }`.
  - Throw a clear error if the invite URL has no token.

- **`packages/testing/playwright/services/user-api-helper.ts`**
  - Same: parse `token` from invite URL and POST to `/rest/invitations/accept` with token and user details.

**How to test:**

- Frontend: `cd packages/frontend/editor-ui && pnpm test src/features/core/auth/views/SignupView.test.ts`
- Backend: `cd packages/cli && pnpm test test/integration/controllers/invitation/invitation.controller.integration.test.ts`
- E2E: `pnpm --filter=n8n-playwright test:local tests/e2e/sharing/access-control.spec.ts` (and e.g. `tests/e2e/building-blocks/user-service.spec.ts`)

- https://linear.app/n8n/issue/IAM-403

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/controllers/__tests__/invitation.controller.test.ts
```diff
@@ -225,229 +226,6 @@ describe('InvitationController', () => {
 		});
 	});
 
-	describe('acceptInvitation', () => {
-		it('throws a BadRequestError if SSO is enabled', async () => {
```
phyllis-noester, line 229: is this no longer the case?

phyllis-noester (dismissed):
🥳

packages/cli/src/controllers/__tests__/invitation.controller.test.ts
BGZStephen: This test was essentially duplicated while we have both login systems side by side, there is still a test in this file for `throws a BadRequestError if SSO is enabled` for the token based path which is now the default.

afitzek (approved):
🚀 Tested it also in the provided test instance

PR #27111, opened 2026-03-16
ci: Unify QA metrics pipeline to single webhook, format, and BigQuery table (no-changelog)

shortstacked:
## Summary

Collapses 4 separate webhook endpoints and 8 CI secrets into a single unified metrics pipeline. All telemetry — Playwright perf/benchmark metrics, Turbo build stats, Docker image stats, and container startup times — now uses the same payload shape and lands in one BigQuery table with JSON dimensions for slicing.

**What changed:**
- Add `.github/scripts/send-metrics.mjs` — shared helper for CI scripts (`metric()`, `buildContext()`, `sendMetrics()`)
- Rewrite `send-build-stats.mjs` — per-package metric rows with `{package, cache, task}` dimensions instead of raw Turbo blob
- Rewrite `send-docker-stats.mjs` — `{image, platform}` dimensions
- Update container telemetry env var: `CONTAINER_TELEMETRY_WEBHOOK` → `QA_METRICS_WEBHOOK_URL`
- Update Playwright metrics reporter: renamed env vars to `QA_METRICS_WEBHOOK_*`, aligned metric shape (`benchmark_name`/`value`/`unit`) and context fields (camelCase, added `runUrl`/`attempt`/`timestamp`)
- Normalize benchmark harness: `LoadProfile` and `TriggerType` moved to `types.ts`, dimensions replace label strings, `trigger` is now required (no default)
- Update all 6 affected workflow YAML files to use `QA_METRICS_WEBHOOK_*` secrets
- Rewrite `CI-TELEMETRY.md` to document the unified pipeline

**Verification:** Trigger any E2E workflow + CI build on this branch — rows should appear in `qa_performance_metrics` with `dimensions` populated and `benchmark_name` set.

**After merge — remove these GitHub secrets:**
`BUILD_STATS_WEBHOOK_URL/USER/PASSWORD`, `CONTAINER_TELEMETRY_WEBHOOK`, `DOCKER_STATS_WEBHOOK_URL`

## Review / Merge checklist

- [x] Docs updated (`CI-TELEMETRY.md`, `reporters/USAGE.md`, `containers/README.md`)
- [x] Tests included.


.github/scripts/send-build-stats.mjs
```diff
@@ -1,104 +1,67 @@
...
-const summaryPath = files.length > 0 ? join(runsDir, files[0]) : null;
-
-const webhookUrl = process.env.BUILD_STATS_WEBHOOK_URL;
-const webhookUser = process.env.BUILD_STATS_WEBHOOK_USER;
-const webhookPassword = process.env.BUILD_STATS_WEBHOOK_PASSWORD;
 
-if (!summaryPath) {
+const files = readdirSync(runsDir)
+	.filter((f) => f.endsWith('.json'))
+	.sort();
```
Matsuuu, line 30: nit: Not sure what the sort here does as it only sorts the string aphabetically and the strings are just jumbled alphanumerics (e.g. 3B3lBWzWUPQKI8JS0GQeUFK4h3V.json)

Most of these runs are assumably run in a vacuum so there is only one json file but if we actually wanted the "Latest" file we would need to introduce something like 

(picked up from StackOverflow)
```javascript
import { readdirSync, lstatSync } from "fs";

const orderReccentFiles = (dir: string) =>
  readdirSync(dir)
    .filter(f => lstatSync(f).isFile())
    .map(file => ({ file, mtime: lstatSync(file).mtime }))
    .sort((a, b) => b.mtime.getTime() - a.mtime.getTime());

const getMostRecentFile = (dir: string) => {
  const files = orderReccentFiles(dir);
  return files.length ? files[0] : undefined;
};
```

.github/scripts/send-metrics.mjs
```diff
@@ -0,0 +1,94 @@
...
+		git: {
+			sha: process.env.GITHUB_SHA?.slice(0, 8) ?? null,
+			branch: process.env.GITHUB_HEAD_REF ?? process.env.GITHUB_REF_NAME ?? null,
+			pr: prMatch ? parseInt(prMatch[1], 10) : null,
+		},
+		ci: {
+			runId,
+			runUrl:
+				runId && process.env.GITHUB_REPOSITORY
+					? `https://github.com/${process.env.GITHUB_REPOSITORY}/actions/runs/${runId}`
```
Matsuuu, line 38: nit:

We could also use `${process.env.GITHUB_SERVER_URL}`, though I don't think we're moving away from github anytime soon.

Matsuuu (approved):
Looks really good. Nicely unified! A couple of nits but nothing blocking

PR #27113, opened 2026-03-16
feat(Wordpress Node): Add support for OAuth2

Joffcom:
## Summary

Adds support for OAuth2 to the Wordpress node, OAuth2 is only available "out of the box" on wordpress.com hosted blogs.

This is the first step for CE-418

https://linear.app/n8n/issue/NODE-4658/add-oauth-support-to-wordpress-node

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
+++ b/packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
@@ -1,0 +1,112 @@
+import type { ICredentialType, INodeProperties } from 'n8n-workflow';
+
+export class WordpressOAuth2Api implements ICredentialType {
+	name = 'wordpressOAuth2Api';
+
+	displayName = 'WordPress OAuth2 API';
+
+	extends = ['oAuth2Api'];
+
+	documentationUrl = 'wordpress';
+
+	properties: INodeProperties[] = [
+		{
+			displayName:
+				'OAuth2 authentication works with WordPress.com-hosted sites only. For self-hosted WordPress, use the WordPress API (Basic Auth) credential instead.',
+			name: 'wordpressComNotice',
+			type: 'notice',
+			default: '',
+		},
+		{
+			displayName: 'Grant Type',
+			name: 'grantType',
+			type: 'hidden',
+			default: 'authorizationCode',
+		},
+		{
+			displayName: 'Authorization URL',
+			name: 'authUrl',
+			type: 'hidden',
+			default: 'https://public-api.wordpress.com/oauth2/authorize',
+			required: true,
+		},
+		{
+			displayName: 'Access Token URL',
+			name: 'accessTokenUrl',
+			type: 'hidden',
+			default: 'https://public-api.wordpress.com/oauth2/token',
+			required: true,
+		},
+		{
+			displayName: 'Auth URI Query Parameters',
+			name: 'authQueryParameters',
+			type: 'hidden',
+			default: '',
+		},
+		{
+			displayName: 'Authentication',
+			name: 'authentication',
+			type: 'hidden',
+			default: 'header',
+		},
+		{
+			displayName: 'Use Custom Domain',
+			name: 'customDomain',
+			type: 'boolean',
+			default: false,
+			description:
+				'Whether your WordPress.com site uses a custom domain instead of a .wordpress.com subdomain',
+		},
+		{
+			displayName: 'Custom Domain',
+			name: 'customDomainUrl',
+			type: 'string',
+			displayOptions: {
+				show: {
+					customDomain: [true],
+				},
+			},
+			default: '',
+			placeholder: 'myblog.com',
+			description:
+				"Your WordPress.com site's custom domain. Used as the site identifier in API requests — calls still route through public-api.wordpress.com.",
+		},
+		{
+			displayName: 'Custom Scopes',
+			name: 'customScopes',
+			type: 'boolean',
+			default: false,
+			description: 'Whether to define custom OAuth2 scopes instead of the defaults',
+		},
+		{
+			displayName:
+				'The default scopes needed for the node to work are already set. If you change these the node may not function correctly.',
+			name: 'customScopesNotice',
+			type: 'notice',
+			default: '',
+			displayOptions: {
+				show: {
+					customScopes: [true],
+				},
+			},
+		},
+		{
+			displayName: 'Enabled Scopes',
+			name: 'enabledScopes',
+			type: 'string',
+			displayOptions: {
+				show: {
+					customScopes: [true],
+				},
+			},
+			default: 'global',
+			description: 'Space-separated list of OAuth2 scopes to request',
+		},
+		{
+			displayName: 'Scope',
+			name: 'scope',
+			type: 'hidden',
+			default: '={{$self["customScopes"] ? $self["enabledScopes"] : "global"}}',
+		},
+	];
+}
--- a/packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
+++ b/packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
@@ -67,6 +67,7 @@ export class WordpressOAuth2Api implements ICredentialType {
 				},
 			},
 			default: '',
+			required: true,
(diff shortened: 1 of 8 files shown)


packages/nodes-base/nodes/Wordpress/__tests__/GenericFunctions.test.ts
```diff
@@ -34,6 +54,61 @@ describe('Wordpress > GenericFunctions', () => {
...
+			);
+		});
+
+		it('should use public-api.wordpress.com base URL without custom domain', async () => {
+			await wordpressApiRequest.call(mockFunctions, 'GET', '/posts', {}, {});
+
+			expect(mockFunctions.helpers.requestWithAuthentication).toHaveBeenCalledWith(
+				'wordpressOAuth2Api',
+				expect.objectContaining({
+					uri: 'https://public-api.wordpress.com/wp/v2/posts',
```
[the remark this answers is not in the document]
Joffcom: The wordpress API still uses public-api.wordpress.com for wordpress.com hosted site

packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
```diff
@@ -0,0 +1,112 @@
+import type { ICredentialType, INodeProperties } from 'n8n-workflow';
```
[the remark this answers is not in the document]
Joffcom: Credential tests are not needed for OAuth credential types as we know they work when the connect.

packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
```diff
@@ -0,0 +1,113 @@
...
+			displayName: 'Custom Domain',
+			name: 'customDomainUrl',
+			type: 'string',
+			displayOptions: {
+				show: {
+					customDomain: [true],
+				},
+			},
+			default: '',
+			required: true,
```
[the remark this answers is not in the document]
Joffcom: For user-authorization OAuth2 flows like WordPress.com's, the auth flow itself is the test: the credential is only usable after the user completes the OAuth2 redirect dance and receives an access token. Adding a test request here would be adding something that 65 out of 66 similar credentials don't have, and it would require the token to already exist before it can be tested — which is circular.

packages/nodes-base/nodes/Wordpress/GenericFunctions.ts
```diff
@@ -18,7 +18,33 @@ export async function wordpressApiRequest(
...
+		if (credentials.customDomain && credentials.customDomainUrl) {
+			let domain = credentials.customDomainUrl as string;
+			try {
+				domain = new URL(domain).hostname;
+			} catch {
+				domain = domain.replace(/\/+$/, '');
+			}
+			sitePrefix = `/sites/${domain}`;
+		}
+		baseUri = `https://public-api.wordpress.com/wp/v2${sitePrefix}`;
```
DawidMyslak: When `customDomain` is false, `sitePrefix` is empty and the URL becomes `.../wp/v2/posts` — but the WordPress.com API always requires `/sites/{site}/` in the path. This would fail for regular `.wordpress.com` subdomain users too.

Would it make sense to always require a site identifier for OAuth2, rather than gating it behind the custom domain toggle? Something like a single "WordPress.com Site" field that accepts either `myblog.wordpress.com` or `myblog.com`.

packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
```diff
@@ -0,0 +1,113 @@
+import type { ICredentialType, INodeProperties } from 'n8n-workflow';
+
+export class WordpressOAuth2Api implements ICredentialType {
```
DawidMyslak, line 3: The basic auth credential has a `test` request so users can validate their connection from the UI. Would be great to add one here too — e.g. a GET to `/users/me` on the WordPress.com API.

packages/nodes-base/nodes/Wordpress/GenericFunctions.ts
```diff
@@ -18,7 +18,33 @@ export async function wordpressApiRequest(
...
+	if (isOAuth2) {
+		const credentials = await this.getCredentials('wordpressOAuth2Api');
+		credentialType = 'wordpressOAuth2Api';
+		let sitePrefix = '';
+		if (credentials.customDomain && credentials.customDomainUrl) {
+			let domain = credentials.customDomainUrl as string;
+			try {
+				domain = new URL(domain).hostname;
+			} catch {
+				domain = domain.replace(/\/+$/, '');
```
DawidMyslak: Nit: the catch branch only strips trailing slashes, so `myblog.com/path` (no protocol) would keep the `/path`. A `.split('/')[0]` would handle that edge case.

packages/nodes-base/nodes/Wordpress/__tests__/GenericFunctions.test.ts
```diff
@@ -26,6 +32,20 @@ describe('Wordpress > GenericFunctions', () => {
...
+		it('should use wordpressApi credential for basic auth', async () => {
+			mockFunctions.helpers.requestWithAuthentication.mockResolvedValue({ data: 'testData' });
+			await wordpressApiRequest.call(mockFunctions, 'GET', '/posts', {}, {});
+
+			expect(mockFunctions.getCredentials).toHaveBeenCalledWith('wordpressApi');
+			expect(mockFunctions.helpers.requestWithAuthentication).toHaveBeenCalledWith(
+				'wordpressApi',
+				expect.objectContaining({
+					uri: 'http://example.com/wp-json/wp/v2/posts',
+					rejectUnauthorized: true,
```
DawidMyslak, line 44: This test locks in a URL without the `/sites/{site}` segment — would need updating once the base URL construction is fixed.

packages/nodes-base/credentials/test/WordpressOAuth2Api.credentials.test.ts
```diff
@@ -0,0 +1,159 @@
+import { ClientOAuth2 } from '@n8n/client-oauth2';
```
DawidMyslak: The OAuth2 flow tests here are mostly exercising `@n8n/client-oauth2` rather than the credential itself. The property assertions are great — maybe trim the flow tests in favor of testing the scope expression logic?

DawidMyslak (commented):
I've tried a new way of reviewing PR, directly from Cursor, I hope the comments make sense. Non of them are blockers, let me know if the comments are valuable. I will run a second scan soon.

packages/nodes-base/credentials/WordpressOAuth2Api.credentials.ts
Joffcom: We don't typically tests for OAuth credentials as the connection working is normally enough of a test, If it wasn't working you wouldn't be able to connect. This does sound like a possible future check to also add in credential tests and to update the UI to not look odd.

packages/nodes-base/nodes/Wordpress/GenericFunctions.ts
Joffcom: Good catch!

packages/nodes-base/nodes/Wordpress/GenericFunctions.ts
Joffcom: Fixed and updated the test

packages/nodes-base/nodes/Wordpress/__tests__/GenericFunctions.test.ts
Joffcom: Should be resolved.

packages/nodes-base/credentials/test/WordpressOAuth2Api.credentials.test.ts
Joffcom: I am mixed on this, We do want to make sure the credential does load custom scopes and works as expected as well to prevent anything unexpected breaking in client-oauth2 although we could add more tests to that package it felt right to have it in the credential as well to cover any parameter renames in the credential file.

Will reduce them anyway though.

DawidMyslak (approved):

PR #27126, opened 2026-03-17
fix(core): Emit `leader-takeover` on leadership mismatch in `checkLeader`

ivov:
## Summary

Scenario:

1. A main is the leader, has `WaitTracker` running
2. This main terminates, and the key still has this main's name, with a few seconds of TTL left
3. This main restarts and tries to claim leadership via `MultiMainSetup.init` during startup. This fails (as expected) because the key is still set, left over from before the restart. This main marks itself as follower
4. `WaitTracker.init` runs during startup. This main is a follower so `WaitTracker` does not start (as expected)
5. The first `MultiMainSetup.checkLeader` runs. This main reads the key and finds its own name left over from before the restart, and marks itself as leader. But importantly, this main does not fire the `leader-takeover` event, because we're assuming that when a main finds its own name in Redis, this is just a routine leadership check
6. Main is now the leader with no `WaitTracker` so waiting executions do not resume

Likely a `StatefulSet` rolling restart on cloud is faster than a restart in the old docker compose setup on internal, and also the hostname is fixed in k8s vs. dynamic in docker compose, where the hostname defaults to the container ID which changes on restart. 

This fix moves `markAsLeader()` and adds `emit('leader-takeover')` inside the existing guard, so the event fires exactly once on the follower-to-leader transition.

https://linear.app/n8n/issue/CAT-2622

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/scaling/__tests__/multi-main-setup.ee.test.ts
+++ b/packages/cli/src/scaling/__tests__/multi-main-setup.ee.test.ts
@@ -1,0 +1,110 @@
+import { mockLogger } from '@n8n/backend-test-utils';
+import type { GlobalConfig } from '@n8n/config';
+import { MultiMainMetadata } from '@n8n/decorators';
+import { mock } from 'jest-mock-extended';
+import type { ErrorReporter, InstanceSettings } from 'n8n-core';
+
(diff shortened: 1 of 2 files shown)


tomi (approved):
🚀

PR #27132, opened 2026-03-17
fix(editor): Remove toast bottom offset when AI chat panel is open

scdekov:
## Summary

When both the logs panel and the AI builder chat panel are open on the canvas, toast notifications were rendered with an unnecessary bottom offset. 

**Before** (AI builder + logs panel open — toast has unnecessary bottom offset):
image

**After** (toast positioned correctly):

image
image

## Review / Merge checklist

- [x] Tests included.


MiloradFilipovic (approved):
Works well. The only thing I am concerned about is how maintainable this is. If we add another panel that pushes toasts up, we would need to keep this `computed` in sync. So wondering is there a generic way (maybe just css?) to keep these on the bottom

PR #27151, opened 2026-03-17
feat(core): Introduce CredentialDependency entity to track credential dependencies

alielkhateeb:
## Summary

Introduces a new `CredentialDependency` entity and supporting infrastructure to track which external secret providers a credential depends on.

https://www.loom.com/share/27e084b98d7945aeb4308a1de505de0a

### Changes

- **New `CredentialDependency` entity** (`packages/@n8n/db`): Stores `credentialId`, `dependencyType` (`externalSecretProvider`), and `dependencyId` with a unique constraint and cascade delete on credential removal.
- **DB migration** (`1773000000000-CreateCredentialDependencyTable`): Creates the `credential_dependency` table for both SQLite and PostgreSQL.
- **New `CredentialDependencyRepository`**: Provides `findCredentialIdsByDependencyId`, `upsertDependenciesForCredential`, and `syncDependenciesForCredential` methods.
- **Updated `CredentialsService`**: Parses credential data on create/update to extract and persist external secret provider dependencies.
- **Updated `validation.ts`**: Adds helpers to extract dependency IDs from credential values.
- **Updated repositories** for `CredentialsRepository` and `SecretsProviderConnectionRepository` to support dependency-aware queries.
- **Tests**: Updated and extended tests for `CredentialsService`, `CredentialsController`, and the public API credentials service.

### How to test

**Create flow**

1. Create a credential that references an external secret provider value (e.g. `={{ $secrets.myVault.myKey }}`).
2. Verify a row is created in `credential_dependency` with `dependencyType = 'externalSecretProvider'`.

**Update flow**

3. Update the credential to add a new external secret provider reference — verify a new dependency row is inserted.
4. Update the credential to remove an existing external secret provider reference — verify the corresponding dependency row is deleted.
5. Update the credential to swap one external secret provider reference for another — verify the old row is removed and the new one is inserted.
6. Update the credential without changing any references — verify the dependency rows are unchanged.

**Delete flow**

7. Delete the credential and verify all associated dependency rows are cascade-deleted.

https://linear.app/n8n/issue/LIGO-230/be-switch-to-index-based-query-for-getting-all-credentials-referencing

## Review / Merge checklist

- [ ] Tests included.


packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
```diff
@@ -0,0 +1,203 @@
+import { Container } from '@n8n/di';
+import { Cipher } from 'n8n-core';
+
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const credentialDependencyTable = 'credential_dependency';
+const externalSecretProviderDependencyType = 'externalSecretProvider';
+const providerKeyPattern = '[A-Za-z0-9]+(?:-[A-Za-z0-9]+)*';
```
alielkhateeb, line 8: Sort of a duplicate of https://github.com/n8n-io/n8n/blob/7a73194d1099e48b4cfbc44dcd1f91207b00b2d7/packages/cli/src/credentials/validation.ts#L16 but since migrations should be self-contained I created a new one here

packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
alielkhateeb, line 1: The utilities in this file might give the impression of duplicate code, that's because it is duplicate code 😂 somehow

Usually migrations are meant to be self contained with minimal dependencies to avoid breaking when business logic change over time, if this is not the right way here at n8n I'm happy to amend and reuse the existing utility methods :)

packages/cli/src/credentials/credential-dependency.service.ts
```diff
@@ -0,0 +1,42 @@
...
+
+		if (providerId === null) return undefined;
+
+		return {
+			dependencyType: EXTERNAL_SECRET_PROVIDER_DEPENDENCY_TYPE,
+			dependencyId: providerId,
+		};
+	}
+
+	async resolveProviderIdsFromCredentialData(
```
alielkhateeb: At first this might look like it should be part of `Credentials Service` but actually this is solely used for resolving dependencies, so it fits better in the credential dependency service IMO

packages/@n8n/db/src/repositories/credentials.repository.ts
```diff
@@ -278,9 +289,19 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
...
 	): SelectQueryBuilder<CredentialsEntity> {
 		const qb = this.createQueryBuilder('credential');
 
+		if (credentialIds) {
+			if (credentialIds.length === 0) {
+				qb.andWhere('1 = 0');
+				return qb;
+			}
+
+			qb.andWhere('credential.id IN (:...credentialIds)', { credentialIds });
```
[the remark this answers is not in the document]
alielkhateeb: That was a really good observation @cubic-dev-ai

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -655,22 +656,43 @@ export class CredentialsService {
...
+		decryptedCredentialData?: ICredentialDataDecryptedObject,
+	) {
 		await this.externalHooks.run('credentials.update', [newCredentialData]);
+		const nextProviderIds = decryptedCredentialData
+			? await this.credentialDependencyService.resolveProviderIdsFromCredentialData(
+					decryptedCredentialData,
+				)
+			: undefined;
+
+		return await this.credentialsRepository.manager.transaction(async (transactionManager) => {
```
alielkhateeb, line 689: wrapping update in transaction now so the dependency sync is atomic with the credential update

packages/cli/src/credentials/validation.ts
```diff
@@ -77,6 +77,23 @@ function containsExternalSecrets(data: ICredentialDataDecryptedObject): boolean
 	return secretPaths.length > 0;
 }
 
+export function extractExternalSecretProviderKeys(
```
ireneea: Should this function belong to the `validation.ts` file?

alielkhateeb: No, that does not belong here 😂 I will create a new utility and send you the commit hash here

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -715,6 +737,19 @@ export class CredentialsService {
...
+					await this.credentialDependencyService.resolveProviderIdsFromCredentialData(
+						decryptedData,
+					);
+				await this.credentialDependencyRepository.upsertDependenciesForCredential({
+					credentialId: savedCredential.id,
+					dependencyType: EXTERNAL_SECRET_PROVIDER_DEPENDENCY_TYPE,
+					dependencyIds,
+					entityManager: transactionManager,
+				});
+			}
```
ireneea: [nit] is it worth wrapping this logic directly in a method inside the CredentialDependencyService

packages/cli/src/credentials/validation.ts
alielkhateeb: @ireneea 4646f534f12b84d13dcb5c8b9e79b9f6313f92ed does this make sense? Also identified some similar use cases which are refactored to centralize the parsing/utils logic

packages/cli/src/credentials/credentials.service.ts
alielkhateeb: @ireneea that's a really good idea, having this here makes credentials know too much about the credential dependency, I'll move it and adjust 🚀

packages/cli/src/credentials/external-secrets.utils.ts
ireneea, line 1: ❤️ Love this refactoring

packages/cli/src/credentials/validation.ts
ireneea: The refactoring looks much better, thanks for making the change 👍

packages/cli/src/credentials/credentials.service.ts
alielkhateeb: @ireneea eefaa5c943d658569ecce7649d0b671e7afa2fb2 love this, led to not needing the credential dependency repository in credentials service at all and this makes so much more sense 👏

packages/cli/src/credentials/external-secrets.utils.ts
```diff
@@ -0,0 +1,76 @@
...
+	const providerKeys = new Set<string>();
+
+	const expressionBlocks = expression.matchAll(/\{\{(.*?)\}\}/gs);
+
+	for (const expression of expressionBlocks) {
+		const expressionContent = expression[1]; // Content inside {{ }}
+
+		// Match all dot notation occurrences: $secrets.providerKey
+		const dotMatches = expressionContent.matchAll(
+			new RegExp(`\\$secrets\\.(${PROVIDER_KEY_PATTERN})`, 'g'),
```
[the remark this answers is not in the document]
ireneea: Can you give an example of an expression where there would cause an incorrect dependency mapping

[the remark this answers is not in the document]
alielkhateeb: @ireneea aee6b9fde641f8dc5092b29bad4d29dddc6dc722 I think this covers it :) with the test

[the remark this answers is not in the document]
alielkhateeb: Does this https://github.com/n8n-io/n8n/pull/27151#discussion_r2959607542 cover the case you mentioned? Any regressions in doing that?

packages/cli/src/services/credentials-tester.service.ts
```diff
@@ -211,9 +211,7 @@ export class CredentialsTester {
 				});
 
 				// Keep all credentials data keys which have a secret value
-				credentialsDataSecretKeys = getAllKeyPaths(credentialsDecrypted.data, '', [], (value) =>
-					value.includes('$secrets.'),
-				);
+				credentialsDataSecretKeys = getExternalSecretExpressionPaths(credentialsDecrypted.data);
```
konstantintieber, line 214: Nice upgrade of logic from `decryptedCredentialData.includes('$secrets")` to the newly exported util 👍

packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
konstantintieber: I think it's good that migrations don't import code but rather duplicate it 👍 
I wouldn't want the behaviour of the migration to change if we change the util implementation in the future (or at least I wouldn't expect as a dev that changing a function in our code will affect a migration that was built in the past).

alielkhateeb: @konstantintieber Awesome, thanks for reassuring me 🫶

packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
```diff
@@ -0,0 +1,203 @@
...
+				column('dependencyType').varchar(64).notNull,
+				column('dependencyId').varchar(255).notNull,
+			)
+			.withForeignKey('credentialId', {
+				tableName: 'credentials_entity',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withIndexOn(['credentialId'])
+			.withIndexOn(['dependencyType', 'dependencyId']).withCreatedAt;
```
konstantintieber, line 39: Good set of indexes you added here 👍 
I think this one here should also be added as decorator on the entity: `@Index(['dependencyType', 'dependencyId'])`

packages/@n8n/db/src/repositories/credentials.repository.ts
```diff
@@ -185,9 +188,44 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
...
+			];
+
+			qb.select(defaultSelect.map((k) => `credential.${k}`));
+			qb.leftJoinAndSelect('credential.shared', 'shared');
+			qb.leftJoinAndSelect('shared.project', 'project');
+			qb.leftJoinAndSelect('project.projectRelations', 'projectRelations');
+
+			if (includeData) qb.addSelect('credential.data');
 
+			return await qb.getMany();
```
konstantintieber: Seeing this early return here makes me think: does that mean that if I call `findAllGlobalCredentials` with `includeData: true` and dependency filters, that the `includeData:  true` will be ignored, since that one is handled only if dependency filters are not defined?

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -145,6 +152,15 @@ export class CredentialsService {
...
 		const returnAll = hasGlobalScope(user, 'credential:list');
 		const isDefaultSelect = !listQueryOptions.select;
+		const dependencyFilter = externalSecretsStore
+			? await this.credentialDependencyService.resolveExternalSecretsStoreDependencyFilter(
+					externalSecretsStore,
+				)
+			: undefined;
+
+		if (externalSecretsStore && !dependencyFilter) {
+			return [];
```
konstantintieber, line 180: Should we throw an error here telling the user that we couldn't query by the provided external secret provider key, because no secret provider with that key exists?

packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
alielkhateeb: @konstantintieber True 98c88f4fcc8351fdb0b10618217eaa7557bb79b3
I wondered - for my own knowledge - if we have any sort of "sync" process between entities and database state, basically if we ever update the entities and "generate" the migrations?

konstantintieber: No automatic sync process that I'm aware of. All human based 😅

packages/cli/src/credentials/credentials.service.ts
alielkhateeb: @konstantintieber I would see this as a filter, so if the secret provider doesn't exist we return an empty array, just like searching a list of items by a keyword that matches nothing 

But I definitely see a potential refactoring here, basically the params aren't clear, I'll amend and send the commit hash here for another round and to get your opinion

konstantintieber: Yeah you're right. a "get many credentials" endpoint is likely not the place to validate that the external secret provider key that was passed in has a match in our db.

We're essentially saying: if you're looking for credentials that use a vault that doesn't exist, then you also won't find any credentials 👍

packages/@n8n/db/src/entities/credential-dependency-entity.ts
```diff
@@ -0,0 +1,37 @@
...
+
+	@Column({ length: 36 })
+	@Index()
+	credentialId: string;
+
+	@Column({ length: 64 })
+	dependencyType: CredentialDependencyType;
+
+	@Column({ length: 255 })
+	dependencyId: string;
```
konstantintieber, line 32: Nit: While I'm fine with the `dependencyId`, the existing `WorkflowDependency` entity calls this `dependencyKey`

konstantintieber: That might be specific to the fact that a workflow has more varied kinds of dependencies though (e.g. node type).

packages/cli/src/credentials/credentials.service.ts
alielkhateeb: 59176edacc41fbceed56ea13b8bc507f0c4c6c74 I think this makes the filter more explicit, IMO, what do u think?

packages/@n8n/db/src/entities/credential-dependency-entity.ts
alielkhateeb: Initially I had it as `dependencyKey` to follow the same as `workflowDepdency` and I've given this a thought, and I didn't want people to mistakenly insert `providerKey` instead of `secretConnectionId`, this is the only reason why I went with `dependencyId` rather than `dependencyKey`

konstantintieber: Good call 👍

packages/@n8n/db/src/repositories/credentials.repository.ts
alielkhateeb: Hmm, yeah I see how it is easily missed since we send `includeData` to `toFindManyOptions` in case there is no filter, I see the confusion

I have extracted the branch where we want the filtered results into a private method, that way it is a bit less confusing and easier to trace the code. 83698dbed6bec57d0ea303b2d6517085785f98d2

packages/cli/src/credentials/credentials.service.ts
konstantintieber: Hmm I'm indecisive on that one. Is `includeGlobal` not a filter? Classic naming things...

packages/cli/src/modules/external-secrets.ee/secrets-providers-connections.service.ee.ts
```diff
@@ -201,9 +206,17 @@ export class SecretsProvidersConnectionsService {
 	async deleteConnection(providerKey: string, userId: string): Promise<SecretsProviderConnection> {
 		const connection = await this.findConnectionOrFail(providerKey);
 		const projectInfo = this.extractProjectInfo(connection);
+		const dependencyId = connection.id.toString();
 
-		await this.projectAccessRepository.deleteByConnectionId(connection.id);
-		await this.repository.remove(connection);
+		await this.repository.manager.transaction(async (entityManager) => {
+			await this.projectAccessRepository.deleteByConnectionId(connection.id, entityManager);
+			await this.credentialDependencyService.deleteDependencyById({
```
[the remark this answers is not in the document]
alielkhateeb: How did deleteConnectionForProject lose transactional atomicity? In this PR this is unchanged compared to master isn't it?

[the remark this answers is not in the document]
konstantintieber: I think what this bot means is that these three statements should be in one transaction:
```
await this.credentialDependencyService.deleteDependencyById({
	dependencyType: EXTERNAL_SECRET_PROVIDER_DEPENDENCY_TYPE,
	dependencyId: connectionId.toString(),
});
await this.projectAccessRepository.deleteByConnectionId(connectionId);
await this.repository.delete({ id: connectionId });
```

Because the first two depend on the successful deletion in the third statement.

So even on master right now, the fact that `await this.projectAccessRepository.deleteByConnectionId` and `await this.repository.delete` are not in a transaction is not our usual way of implementing this.
I also wonder why we're not using `onDelete: 'cascade'` on the `ManyToOne project` relation of `ProjectSecretsProviderAccess`.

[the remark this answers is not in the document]
konstantintieber: Weird, but the bot is talking about a different piece of code. The one I mentioned is in `SecretsProvidersConnectionsService.deleteConnectionForProject`, while the bot commented about `SecretsProvidersConnectionsService.deleteConnection` 🤔

packages/@n8n/db/src/repositories/credentials.repository.ts
```diff
@@ -185,14 +188,63 @@ export class CredentialsRepository extends Repository<CredentialsEntity> {
...
-		const findManyOptions = this.toFindManyOptions({ includeData });
+	async findAllGlobalCredentials(
+		options: {
+			includeData?: boolean;
+			filters?: {
+				dependency?: CredentialDependencyFilter;
+			};
+		} = {},
+	): Promise<CredentialsEntity[]> {
+		const { includeData = false, filters } = options;
```
konstantintieber, line 199: TIL how default values in JS's destructuring syntax works 👍

konstantintieber (approved):
LGTM

packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
```diff
@@ -0,0 +1,203 @@
...
+
+		await createIndex(
```
guillaumejacquart, line 41: Why the separate createIndex ? why not another `.withIndexOn(` ?

guillaumejacquart (approved):
one nit comment, but looks good, thanks 🚀

packages/@n8n/db/src/migrations/common/1773000000000-CreateCredentialDependencyTable.ts
alielkhateeb: That’s an absolutely valid question ✅

The main reason is that I found it hard to clearly spot unique indexes when they’re defined alongside non-unique ones using `withIndexOn`. For example:

```ts
.withIndexOn(['credentialId'])
.withIndexOn(['dependencyType', 'dependencyId'])
.withIndexOn(['credentialId', 'dependencyType', 'dependencyId'], true);
```

Here, the `true` is quite subtle and easy to miss when scanning the code.

Since this is a migration file (and not something we expect to revisit or refactor), I preferred to make the unique index explicit. This makes it easier for reviewers and future readers to spot which index enforces uniqueness vs which ones are just for performance.

Also… the fact that you pointed it out probably means my plan worked 😂 haha

PR #27155, opened 2026-03-17
build: Add `import-x/no-extraneous-dependencies` lint rule to `n8n-workflow`

ivov:
## Summary

Enables the `import-x/no-extraneous-dependencies` ESLint rule for `packages/workflow`.

This ensures every import has a corresponding entry in `package.json`, catching dependencies that only work by accident via hoisting. Would have prevented the issue fixed in #27129. (Enabling this rule for _all_ packages is out of scope.)

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/workflow/eslint.config.mjs
+++ b/packages/workflow/eslint.config.mjs
@@ -5,6 +5,7 @@ export default defineConfig(
 	baseConfig,
 	{
 		rules: {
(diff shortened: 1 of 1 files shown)


tomi (commented):
Are there some violations of this in other packages or why is it out of scope?

tomi (approved):
🚀

PR #22790, opened 2025-12-05
fix(cli): register webhooks when activating workflow via REST API

DmitryVDV:
## Summary

When activating a workflow via REST API (POST /rest/workflows/{id}/activate), webhooks are not being registered because `shouldAddWebhooks()` in `ActiveWorkflowManager` only returns `true` for `init` and `leadershipChange` activation modes, not for `activate`.

## Fix

Added `activate` to the list of activation modes that trigger webhook registration in `shouldAddWebhooks()` method.

**Before:**
```typescript
if (['init', 'leadershipChange'].includes(activationMode)) return true;
```

**After:**
```typescript
if (['init', 'leadershipChange', 'activate'].includes(activationMode)) return true;
```

## Testing

Tested manually on a development build:
1. Created a workflow with a webhook node
2. Activated it via REST API: `POST /rest/workflows/{id}/activate`
3. Verified the webhook responds with 200 OK and "Workflow was started"

## Related Issue

Fixes #21614

Issue #21614, opened 2025-11-06
Deployment + Activation via API does not register webhook path


rileypx:
### Bug Description

When deploying and activating a workflow using the REST API, no webhook paths are being registered, which causes this error when attempting to execute such workflows:

>  HTTP error executing workflow 'workflow name': status=404, response={"code":404,"message":"The requested webhook \"POST 4c60e797-b8bf-4955-9daa-a1de775c4df6\" is not registered.","hint":"The workflow must be active for a production URL to run successfully. You can activate the workflow using the toggle in the top-right of the editor. Note that unlike test URL calls, production URL calls aren't shown on the canvas (only in the executions list)"}

However, the workflow is both deployed and active. The only solution I've found is to make a change and save via the UI, which is not possible at scale.

I notice the frontend makes an internal call to /rest/webhooks/find with the path whenever it is saved - this seems to do the registering:

<img width="533" height="199" alt="Image" src="https://github.com/user-attachments/assets/2695d16d-053e-444d-bd8a-7ad476fedc52" />

<img width="533" height="199" alt="Image" src="https://github.com/user-attachments/assets/e381c88e-ab85-48cb-a0db-8bf8408c308d" />

This behavior is not included in the API, and I am unable to mimic the request using my existing API key with any combination of headers. I am making a POST request to a webhook configured for POST method. From my perspective, this makes the API unusable headlessly.

### To Reproduce

Deploy a workflow using API
Activate said workflow using API
Make request to production webhook URL

### Expected behavior

Workflow executes

### Debug Info

# Debug info

## core

- n8nVersion: 1.111.1
- platform: docker (cloud)
- nodeJsVersion: 22.19.0
- database: sqlite
- executionMode: regular
- concurrency: 20
- license: community
- consumerId: 00000000-0000-0000-0000-000000000000

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 720 hours
- maxCount: 25000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/141.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2025-11-06T18:22:42.956Z

### Operating System

n8n cloud

### n8n Version

1.111.1

### Node.js Version

n8n cloud

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

[...]

PR #27161, opened 2026-03-17
fix(core): Assign webhook ID to API-created webhook nodes

ivov:
## Summary

When a workflow with a webhook node is created via the public API, we never assign a `webhookId` to those nodes. Without `webhookId`, `getNodeWebhookPath()` constructs the wrong path format, so the registered webhook path can't match incoming requests.

PR #22790 adds `'activate'` to `shouldAddWebhooks()`, but webhooks are already registered during activation on single-instance setups (`isLeader` being `true`). The problem is not that registration is skipped, but rather that registration uses the wrong path because `webhookId` is missing from the node.

Four spots:

- **Public API create** -> Only entry point for API-created workflows.
- **`WorkflowService.update`** -> Assigns `webhookId` to any webhook node still missing one (e.g. workflows created before this fix). Existing IDs are never overwritten.
- **MCP create tool** -> MCP-generated workflows skip the UI. Refactored to use the shared helper.
- **MCP update tool** -> Technically redundant since `WorkflowService.update` now also calls it, but harmless (idempotent). Refactored to use the shared helper.

https://linear.app/n8n/issue/CAT-2624
#21614

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/workflow-helpers.ts
+++ b/packages/cli/src/workflow-helpers.ts
@@ -4,12 +4,14 @@ import { Container } from '@n8n/di';
 import type {
 	IDataObject,
 	INodeCredentialsDetails,
+	INodeTypes,
 	IRun,
 	ITaskData,
 	IWorkflowBase,
 	IWorkflowSettings,
 	RelatedExecution,
 } from 'n8n-workflow';
+import { resolveNodeWebhookId } from 'n8n-workflow';
 import { v4 as uuid } from 'uuid';
 
 import { VariablesService } from '@/environments.ee/variables/variables.service.ee';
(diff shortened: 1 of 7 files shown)


packages/cli/src/workflow-helpers.ts
```diff
@@ -68,6 +70,25 @@ export function addNodeIds(workflow: IWorkflowBase) {
 	});
 }
 
+/**
+ * Assign webhookId to any webhook node that is missing one.
+ * The UI does this on the frontend when adding nodes to the canvas,
+ * but workflows created via the API skip that step.
+ */
+export function resolveNodeWebhookIds(workflow: IWorkflowBase, nodeTypes: INodeTypes) {
```
tomi, line 78: 👍 for unifying

tomi (approved):
🚀

PR #27169, opened 2026-03-17
feat(core): Add project context to execution log metadata

krider2010:
## Summary
- Adds `projectId` and `projectName` as optional metadata fields to structured execution logs (worker start/finish/status, post-execution errors)
- Propagates project info through `IWorkflowExecutionDataProcess` → `JobData` so scaling workers have it available
- Also adds `workflowName` to log calls where it was previously missing (job-processor, workflow-runner)
- Fields are purely additive — omitted when `undefined`, no impact on existing log parsers

## Review / Merge checklist
- [x] PR title follows [conventions](https://github.com/n8n-io/n8n/blob/master/.github/pull_request_title_conventions.md)
- [x] Tests included

## Test plan
- [ ] Run `pnpm test src/scaling/__tests__/job-processor.service.test.ts` from `packages/cli` — 19 tests pass, including 2 new tests for project metadata presence/absence
- [ ] Run `pnpm test src/logging/__tests__/logger.test.ts` from `packages/@n8n/backend-common` — 22 tests pass, including 2 new tests for optional metadata field behavior
- [x] Manual: Start n8n with `N8N_LOG_FORMAT=json N8N_LOG_LEVEL=debug`, trigger a workflow in a team project, confirm `projectId`/`projectName` appear in JSON log output
- [x] Manual: Trigger an execution outside project context (e.g., credential test) and confirm no `projectId`/`projectName` fields appear


packages/cli/src/workflow-runner.ts
```diff
@@ -146,7 +146,17 @@ export class WorkflowRunner {
 		// Register a new execution
 		const executionId = await this.activeExecutions.add(data, restartExecutionId);
 
-		const { id: workflowId, nodes } = data.workflowData;
+		const { id: workflowId, name: workflowName, nodes } = data.workflowData;
+
+		this.logger.info(`Execution ${executionId} started for workflow "${workflowId}"`, {
```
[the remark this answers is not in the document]
krider2010: @MarcL would love your input here

packages/@n8n/backend-common/src/logging/__tests__/logger.test.ts
```diff
@@ -231,6 +231,70 @@ describe('Logger', () => {
...
+				workflowId: 'wf-1',
+				projectId: 'proj-1',
+				projectName: 'Test Project',
+			});
+
+			expect(stdoutSpy).toHaveBeenCalledTimes(1);
+			const output = stdoutSpy.mock.lastCall?.[0];
+			if (typeof output !== 'string') {
+				fail(`expected 'output' to be of type 'string', got ${typeof output}`);
+			}
```
MarcL, line 261: Avoid logic in tests unless this is a pattern we already use?

packages/@n8n/backend-common/src/logging/__tests__/logger.test.ts
```diff
@@ -231,6 +231,70 @@ describe('Logger', () => {
...
+			});
+			const logger = new Logger(globalConfig, mock<InstanceSettingsConfig>());
+
+			logger.info('Workflow execution started', { workflowId: 'wf-1' });
+
+			expect(stdoutSpy).toHaveBeenCalledTimes(1);
+			const output = stdoutSpy.mock.lastCall?.[0];
+			if (typeof output !== 'string') {
+				fail(`expected 'output' to be of type 'string', got ${typeof output}`);
+			}
```
MarcL, line 289: Avoid logic in tests unless this is a pattern we already use?

packages/cli/src/scaling/__tests__/job-processor.service.test.ts
```diff
@@ -919,4 +919,124 @@ describe('JobProcessor', () => {
...
+				expect.objectContaining({
+					workflowId: 'wf-1',
+					workflowName: 'Test Workflow',
+					projectId: 'proj-123',
+					projectName: 'My Project',
+				}),
+			);
+		});
+
+		it('should omit project info from log metadata when absent from job data', async () => {
```
MarcL: [nit] Test doesn't test for "omit" but for them being `undefined`. Change the expectation to remove `projectId` and `projectName` or clarify the test name?

packages/cli/src/workflows/workflow-execution.service.ts
```diff
@@ -210,6 +212,10 @@ export class WorkflowExecutionService {
 		}
 
 		if (data) {
+			const project = await this.ownershipService.getWorkflowProjectCached(workflowData.id);
```
MarcL, line 215: Is it OK to make this extra DB call in all manual executions?
Sounds like it's cached but wanted to double check.

packages/cli/src/workflows/workflow-execution.service.ts
```diff
@@ -262,6 +268,8 @@ export class WorkflowExecutionService {
 		executionMode: WorkflowExecuteMode = 'chat',
 		pushRef?: string,
 	) {
+		const project = await this.ownershipService.getWorkflowProjectCached(workflowData.id);
```
MarcL, line 271: Is it OK to make this extra DB call in all chat workflows?
Does this happen for every chat message?
Sounds like it's cached but wanted to double check.

packages/cli/src/workflow-runner.ts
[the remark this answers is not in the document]
MarcL: Yes - I think we're adding a `started` log before the credentials have been checked so it could `throw` on line 162.

Maybe do it at the end of the function?
Also, should we make it a `.debug` rather than `.info` to avoid spamming the logs?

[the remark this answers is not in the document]
MarcL: Ah - is this just duplicating the same message inside `job-processor.ts`?
It logs the start of the execution (line 95).

MarcL (commented):
A few comments but looks good otherwise.

One thing to note is that someone could change the project name before someone reads the logs. Probably unlikely but do they also need the project name or would the project ID, which shouldn't change, be enough?

packages/@n8n/backend-common/src/logging/__tests__/logger.test.ts
krider2010: It's an existing convention

packages/@n8n/backend-common/src/logging/__tests__/logger.test.ts
krider2010: It's an existing convention

packages/cli/src/workflows/workflow-execution.service.ts
krider2010: Yep it's cached

packages/cli/src/scaling/__tests__/job-processor.service.test.ts
krider2010: Fixed

packages/cli/src/workflows/workflow-execution.service.ts
krider2010: It's cached

MarcL (dismissed):

MarcL (approved):

PR #27178, opened 2026-03-17
fix(core): VM test Group G — RCE prevention & expression fixture fixes

despairblue:
## Summary

Fixes failing VM expression engine tests (Group G) by addressing security and compatibility gaps in the isolated-vm bridge.

- **Unsafe property guard in extend()**: Added `UNSAFE_PROPERTY_NAMES` blocklist to VM-side `findExtendedFunction()` — blocks access to `__proto__`, `constructor`, `prototype`, etc. via expression extensions (20 tests)
- **`__sanitize` override protection**: Replaced direct property assignment with `Object.defineProperty` getter/setter that throws on override attempts (1 test)
- **`$item()` legacy syntax**: Implemented `$item(itemIndex)` in the VM by creating lazy proxies rooted at `['$item', index, '$json']` and special-casing `$item` paths in the bridge's `getValueAtPath`/`getArrayElement` callbacks (1 test)
- **Array proxy `has` trap**: Added `has` trap returning true for numeric indices within bounds, fixing `[].concat(proxyArray)` which uses `in` to check element existence (1 test)

**Test results**: 25 failing → 2 remaining (both deferred — Int8Array cross-isolate transfer and ExpressionError propagation)

## Related

https://linear.app/n8n/issue/CAT-2543

## Review / Merge checklist

- [ ] Tests pass with `N8N_EXPRESSION_ENGINE=vm pnpm test test/expression.test.ts` in `packages/workflow`
- [ ] No changes to host-side expression evaluation (current engine unaffected)

### DIFF
--- a/packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
+++ b/packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
@@ -316,9 +316,20 @@ export class IsolatedVmBridge implements RuntimeBridge {
 		// Used by createDeepLazyProxy when accessing properties
 		const getValueAtPath = new (getIvm().Reference)((path: string[]) => {
 			// Navigate to value
+			// Special-case: paths starting with ['$item', index] call data.$item(index)
+			// to get the sub-proxy for that item, then continue navigating the rest.
 			let value: unknown = data;
-			for (const key of path) {
-				value = (value as Record<string, unknown>)?.[key];
+			let startIndex = 0;
+			const itemFn = (data as Record<string, unknown>).$item;
+			if (path.length >= 2 && path[0] === '$item' && typeof itemFn === 'function') {
+				const itemIndex = parseInt(path[1], 10);
+				if (!isNaN(itemIndex)) {
+					value = (itemFn as (i: number) => unknown)(itemIndex);
+					startIndex = 2;
+				}
+			}
+			for (let i = startIndex; i < path.length; i++) {
+				value = (value as Record<string, unknown>)?.[path[i]];
 				if (value === undefined || value === null) {
 					return value;
 				}
@@ -359,9 +370,19 @@ export class IsolatedVmBridge implements RuntimeBridge {
 		// Used by array proxy when accessing numeric indices
 		const getArrayElement = new (getIvm().Reference)((path: string[], index: number) => {
 			// Navigate to array
+			// Special-case: paths starting with ['$item', index] call data.$item(index)
 			let arr: unknown = data;
-			for (const key of path) {
-				arr = (arr as Record<string, unknown>)?.[key];
+			let startIndex = 0;
+			const itemFn = (data as Record<string, unknown>).$item;
+			if (path.length >= 2 && path[0] === '$item' && typeof itemFn === 'function') {
+				const itemIndex = parseInt(path[1], 10);
+				if (!isNaN(itemIndex)) {
+					arr = (itemFn as (i: number) => unknown)(itemIndex);
+					startIndex = 2;
+				}
+			}
+			for (let i = startIndex; i < path.length; i++) {
+				arr = (arr as Record<string, unknown>)?.[path[i]];
 				if (arr === undefined || arr === null) {
 					return undefined;
 				}
--- a/packages/@n8n/expression-runtime/src/extensions/extend.ts
+++ b/packages/@n8n/expression-runtime/src/extensions/extend.ts
@@ -50,7 +50,40 @@ interface FoundFunction {
 	function: Function;
 }
 
+/**
+ * Property names that must never be accessed via expression extensions.
+ * Matches the host-side blocklist in packages/workflow/src/utils.ts.
+ */
+const UNSAFE_PROPERTY_NAMES = new Set([
+	'__proto__',
+	'prototype',
+	'constructor',
+	'getPrototypeOf',
+	'mainModule',
+	'binding',
+	'_linkedBinding',
+	'_load',
+	'prepareStackTrace',
+	'__lookupGetter__',
+	'__lookupSetter__',
+	'__defineGetter__',
+	'__defineSetter__',
+	'caller',
+	'arguments',
+	'getBuiltinModule',
+	'dlopen',
+	'execve',
+	'loadEnvFile',
+]);
+
 function findExtendedFunction(input: unknown, functionName: string): FoundFunction | undefined {
+	const name = typeof functionName === 'string' ? functionName : String(functionName);
+	if (UNSAFE_PROPERTY_NAMES.has(name)) {
+		throw new ExpressionExtensionError(
+			`Cannot access "${name}" via expression extension due to security concerns`,
+		);
+	}
+
 	// eslint-disable-next-line @typescript-eslint/no-restricted-types
 	let foundFunction: Function | undefined;
 	if (Array.isArray(input)) {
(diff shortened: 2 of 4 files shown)


packages/@n8n/expression-runtime/src/extensions/extend.ts
```diff
@@ -50,7 +50,40 @@ interface FoundFunction {
 	function: Function;
 }
 
+/**
```
despairblue, line 53: What it does: Adds a UNSAFE_PROPERTY_NAMES blocklist (20 entries matching the host-side utils.ts) to findExtendedFunction() in extend.ts. When a property name like constructor, __proto__, or prototype is passed to extend() or extendOptional(), it throws ExpressionExtensionError("Cannot access X due to security concerns") instead of returning
  undefined.                                                                                                                                                                                                                                                                                                                                                
                                                                  
  Why: The current (non-VM) engine blocks these via the host-side sanitizePropertyName() check. The VM isolate inherently prevents RCE from these attacks (no host access), but the tests expect the same explicit throw behavior. This also provides defense-in-depth.                                                                                     
                                                                  
  Tests fixed (20):                                                                                                                                                                                                                                                                                                                                         
  - should block extend() constructor access on arrow functions (x2 — extend + extendOptional)
  - should block extend() constructor access on extend itself                                                                                                                                                                                                                                                                                               
  - should block extend() constructor access on extendOptional    
  - should block extend() constructor access on isNaN                                                                                                                                                                                                                                                                                                       
  - should block extend() constructor access on parseFloat        
  - should block extend() __proto__ access                                                                                                                                                                                                                                                                                                                  
  - should block extend() prototype access                                                                                                                                                                                                                                                                                                                  
  - should block extend() with custom toString() returning constructor                                                                                                                                                                                                                                                                                      
  - should block extend() with custom toString() returning __proto__                                                                                                                                                                                                                                                                                        
  - All of the above duplicated a second time (lines 634-748 duplicate lines 506-690)

packages/@n8n/expression-runtime/src/runtime/reset.ts
```diff
@@ -43,7 +43,15 @@ export function resetDataProxies(timezone?: string): void {
 
 	// __sanitize must be on __data because PrototypeSanitizer generates:
 	// obj[this.__sanitize(expr)] where 'this' is __data (via .call(__data) wrapping)
-	globalThis.__data.__sanitize = __sanitize;
+	// Use a non-writable property descriptor so override attempts throw instead of silently succeeding.
```
despairblue, line 46: What it does: Replaces globalThis.__data.__sanitize = __sanitize (a plain property assignment) with Object.defineProperty using a getter that returns __sanitize and a setter that throws ExpressionError('Cannot override "__sanitize" due to security concerns').                                                                                       
                                                                                                                                                                                                                                                                                                                                                            
  Why: The current engine's __sanitize can't be overridden because tournament evaluates expressions in a sandboxed scope. In the VM, __data is a plain object — a direct assignment is writable by default, so an expression like this.__sanitize = () => "constructor" could bypass the sanitizer. The getter/setter makes the property non-writable and   
  throws visibly on override attempts.                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                            
  Test fixed (1):                                                 
  - should block __sanitize override attempt

packages/@n8n/expression-runtime/src/runtime/reset.ts
```diff
@@ -178,7 +178,14 @@ export function resetDataProxies(timezone?: string): void {
 	// Wire builtins so tournament's VariablePolyfill resolves them from __data
 	initializeBuiltins(globalThis.__data as Record<string, unknown>);
 
-	// TODO: Add other function properties as needed ($item, $vars, etc.)
+	// $item(itemIndex) returns a sub-proxy for the specified item (legacy syntax)
```
despairblue, line 181: What it does: Two changes:
  - Isolate side (reset.ts): Adds $item(itemIndex) function on __data that returns { $json: createDeepLazyProxy(['$item', index, '$json']), $binary: createDeepLazyProxy(['$item', index, '$binary']) }                                                                                                                                                     
  - Bridge side (isolated-vm-bridge.ts): getValueAtPath and getArrayElement callbacks special-case paths starting with ['$item', index] — they call data.$item(index) on the host-side WorkflowDataProxy to get the sub-proxy for that item, then continue navigating the remaining path segments                                                           
                                                                                                                                                                                                                                                                                                                                                            
  Why: $item() is legacy syntax (undocumented) that the current engine supports via WorkflowDataProxy. The VM had no implementation — tournament transforms $item(0).$json["key"] to ("$item" in this ? this : global).$item(0).$json["key"], which needs a callable $item on __data.                                                                       
                                                                                                                                                                                                                                                                                                                                                            
  Test fixed (1):                                                                                                                                                                                                                                                                                                                                           
  - ={{$item(0).$json["Set URL"].json["base_domain"]}}{{$json["link"]}} — expected "leftright", was getting "right"

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
```diff
@@ -95,6 +95,14 @@ export function createDeepLazyProxy(basePath: string[] = []): any {
 			// Handle arrays - metadata: { __isArray: true, __length: number }
 			if (value && typeof value === 'object' && value.__isArray) {
 				const arrayProxy = new Proxy([] as any[], {
+					has(arrTarget: any, arrProp: string | symbol): boolean {
```
despairblue, line 98: What it does: Adds a has trap to the array proxy in lazy-proxy.ts that returns true for numeric indices within 0..length-1.
                                                                                                                                                                                                                                                                                                                                                            
  Why: Array.prototype.concat uses the in operator (index in array) to check whether elements exist before copying them. Without a has trap, the proxy's backing array is [] (empty), so 0 in proxy returns false — concat treats every slot as a hole and produces [undefined, undefined, undefined].                                                      
                                                                                                                                                                                                                                                                                                                                                            
  Test fixed (1):                                                                                                                                                                                                                                                                                                                                           
  - ={{ [].concat($json["Create or update"].json["vid"]) }} — expected [1, 2, 3], was getting [undefined, undefined, undefined]

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -316,9 +316,20 @@ export class IsolatedVmBridge implements RuntimeBridge {
...
 		const getValueAtPath = new (getIvm().Reference)((path: string[]) => {
 			// Navigate to value
+			// Special-case: paths starting with ['$item', index] call data.$item(index)
+			// to get the sub-proxy for that item, then continue navigating the rest.
 			let value: unknown = data;
-			for (const key of path) {
-				value = (value as Record<string, unknown>)?.[key];
+			let startIndex = 0;
+			const itemFn = (data as Record<string, unknown>).$item;
+			if (path.length >= 2 && path[0] === '$item' && typeof itemFn === 'function') {
```
ivov, line 324: Do we need this special case also for `callFunctionAtPath`?

packages/@n8n/expression-runtime/src/runtime/reset.ts
```diff
@@ -170,7 +178,14 @@ export function resetDataProxies(timezone?: string): void {
...
+	globalThis.__data.$item = function (itemIndex: number) {
```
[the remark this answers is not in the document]
ivov: I was also wondering this

ivov (approved):
No deep understanding of all this yet, only minor comments, non-blocking.

PR #27184, opened 2026-03-17
feat(editor): Add canvas-only mode

RicardoE105:
## Summary
- Introduce `N8N_CANVAS_ONLY` environment variable that enables a canvas-only mode
- When enabled, hides the header, sidebar, and overlay elements (assistant button, chat overlay) in the workflow layout
- Adds `canvasOnly` setting to `FrontendSettings` and exposes it via `useSettingsStore`

## Test plan
- [ ] Set `N8N_CANVAS_ONLY=true` and verify header, sidebar, and overlays are hidden
- [ ] Verify default behavior (env var unset) remains unchanged
- [ ] Verify canvas and workflow functionality works correctly in canvas-only mode

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
@@ -7,9 +7,11 @@ withDefaults(
 	defineProps<{
 		items: ITabBarItem[];
 		modelValue?: string;
+		floating?: boolean;
 	}>(),
 	{
 		modelValue: MAIN_HEADER_TABS.WORKFLOW,
+		floating: false,
 	},
 );
 
@@ -27,6 +29,7 @@ function onUpdateModelValue(tab: string, event: MouseEvent): void {
 		v-if="items"
 		:class="{
 			[$style.container]: true,
+			[$style.floating]: floating,
 			['tab-bar-container']: true,
 		}"
 	>
@@ -43,7 +46,7 @@ function onUpdateModelValue(tab: string, event: MouseEvent): void {
 	position: absolute;
 	bottom: 0;
 	left: 50%;
-	transform: translateX(-50%) translateY(150%);
+	transform: translateX(-50%) translateY(50%);
 	min-height: 30px;
 	display: flex;
 	padding: var(--spacing--5xs);
@@ -51,6 +54,10 @@ function onUpdateModelValue(tab: string, event: MouseEvent): void {
 	border-radius: var(--radius);
 	transition: all 150ms ease-in-out;
 	z-index: 100; // Should float above other layout components in any page
+}
+
+.floating {
+	transform: translateX(-50%) translateY(150%);
 }
 
 @media screen and (max-width: 430px) {
--- a/packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
@@ -43,7 +43,7 @@ function onUpdateModelValue(tab: string, event: MouseEvent): void {
 	position: absolute;
 	bottom: 0;
 	left: 50%;
-	transform: translateX(-50%) translateY(50%);
+	transform: translateX(-50%) translateY(150%);
 	min-height: 30px;
 	display: flex;
 	padding: var(--spacing--5xs);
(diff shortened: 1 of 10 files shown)


krider2010 (commented):
A couple of things that might still be visible in canvas-only mode that could look odd in an iframe embed:
- Dynamic banners / notification banners at the top of the page
- Canvas control buttons (zoom in/out/fit/reset) — there's an existing `hideControls` mechanism that could be reused if these aren't wanted

Are these conscious decisions to leave them visible?

packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
```diff
@@ -43,7 +43,7 @@ function onUpdateModelValue(tab: string, event: MouseEvent): void {
 	position: absolute;
 	bottom: 0;
 	left: 50%;
-	transform: translateX(-50%) translateY(50%);
+	transform: translateX(-50%) translateY(150%);
```
krider2010: This change from `translateY(50%)` to `translateY(150%)` affects all users, not just canvas-only mode. Is this intentional? If so, could you call it out in the PR description? It seems like it could be a separate change.

packages/frontend/editor-ui/src/app/layouts/WorkflowLayout.vue
```diff
@@ -86,15 +89,15 @@ onBeforeUnmount(() => {
 		<template #header>
 			<AppHeader />
 		</template>
-		<template #sidebar>
+		<template v-if="!isCanvasOnly" #sidebar>
```
krider2010, line 92: The `#footer` slot (LogsPanel) isn't hidden in canvas-only mode — is that intentional for the iframe embedding use case?

nik8n: Yes. This should be part of the canvas

packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
nik8n: Ay. Good catch, will fix.

packages/frontend/editor-ui/src/app/components/MainHeader/TabBar.vue
```diff
@@ -53,6 +56,10 @@ function onUpdateModelValue(tab: string, event: MouseEvent): void {
 	z-index: 100; // Should float above other layout components in any page
 }
 
+.floating {
```
MiloradFilipovic, line 60: I feel like using `transform` with percentages here could end up being finicky. We should use a fixed top offset, something like this could work:


```suggestion
.floating {
    top: var(--spacing--4xs);
    align-items: center; // <-- We can add this to the main .container class also
}
```

packages/frontend/editor-ui/src/app/layouts/WorkflowLayout.vue
```diff
@@ -86,15 +89,15 @@ onBeforeUnmount(() => {
...
 			<AppSidebar />
 		</template>
 		<LoadingView v-if="isLoading" />
 		<RouterView v-else />
 		<template v-if="layoutProps.logs" #footer>
 			<LogsPanel />
 		</template>
-		<template #overlays>
+		<template v-if="!isCanvasOnly" #overlays>
 			<AskAssistantFloatingButton v-if="assistantStore.isFloatingButtonShown" />
```
MiloradFilipovic, line 101: This will hide floating elements, like assistant button. Is this intentional?

Image

MiloradFilipovic (commented):
Looks good. I left a couple of minor comments so take a look if they make sense.
My major concern is that all of the routes are still accessible in the canvas-only mode, so you can easily reach other parts of the APP (like workflow list, settings, evals, ...) by going straight to their URL. If this is intended behavior, then it' good to go from my end.

packages/frontend/editor-ui/src/app/layouts/WorkflowLayout.vue
krider2010: In response to the routes, I sidestepped that for now. I think we should look at it via a role that triggers when canvas only is used in the config. @fabian8n has some ideas here but it shouldn't block this work.

MiloradFilipovic (approved):
Looks good to me, thanks for addressing the feedback.
If other things (like still being able to access non-cavas pages in canvas-only mode) look good from the product side, it's good to go.

PR #27188, opened 2026-03-17
fix(core): Improve performance of the push/pull modal getStatus

geemanjs:
## Summary
Moves the expensive query of looking up project relations multiple times in the `getStatus` check into a "temporary cache" and attaching it as "context" through the `getStatus` flow

This comes with a slight trade-off that it loads all projects into memory - but this should be relatively small even with 10,000's of projects. 

LIGO-262

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/source-control.ee/source-control-scoped.service.ts
```diff
@@ -24,75 +25,53 @@ export class SourceControlScopedService {
 		private readonly workflowRepository: WorkflowRepository,
 	) {}
 
+	async createContext(user: User): Promise<SourceControlContext> {
+		if (hasGlobalScope(user, 'project:update')) {
```
geemanjs: This behaviour is the same as before - but I wonder if we should check they have all permissions?

packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
```diff
@@ -90,54 +90,46 @@ import { SourceControlContext } from './types/source-control-context';
...
+		}
+		return this.byTeamId.get(owner.teamId);
 	}
-	return owner;
+}
+
+const toStatusOwner = (project: Project | undefined): StatusResourceOwner | undefined => {
+	if (project?.type === 'team' || project?.type === 'personal') {
+		return { type: project.type, projectId: project.id, projectName: project.name };
+	}
```
ireneea:
```suggestion
	if (project?.type) {
		return { type: project.type, projectId: project.id, projectName: project.name };
	}
```

[nit] It doesn't really change the outcome: It's just that if there is a type it's always 'team' or 'personal'

packages/cli/src/modules/source-control.ee/source-control-scoped.service.ts
ireneea: I see what you mean. This is quite an implicit check. It is true now that users with global 'project:update' scope also have 'project:list', 'project:delete' and so on. 

But that might not always be the case especially if we introduce custom system roles. I think it might be worth updating this

packages/cli/src/modules/source-control.ee/source-control-scoped.service.ts
```diff
@@ -24,75 +25,53 @@ export class SourceControlScopedService {
 		private readonly workflowRepository: WorkflowRepository,
 	) {}
 
+	async createContext(user: User): Promise<SourceControlContext> {
```
ireneea: Could this be the logic inside the `SourceControlContext` constructor? This would also mean moving all the custom project fetching inside the `SourceControlContext` as well. I think `fetchAllProjects`,  `fetchAuthorizedProjects` and `fetchAccessibleWorkflowIds` are only used to create this context.

packages/cli/src/modules/source-control.ee/source-control-scoped.service.ts
```diff
@@ -24,75 +25,53 @@ export class SourceControlScopedService {
...
-			where,
-		});
+		return context.accessibleWorkflowIds.map((wfId) => ({ id: wfId }) as WorkflowEntity);
 	}
 
 	getProjectsWithPushScopeByContextFilter(
 		context: SourceControlContext,
 	): FindOptionsWhere<Project> | undefined {
 		if (context.hasAccessToAllProjects()) {
-			// In case the user is a global admin or owner, we don't need a filter
```
ireneea, line 95: ❤️

packages/cli/src/modules/source-control.ee/source-control-status.service.ee.ts
```diff
@@ -147,67 +149,46 @@ export class SourceControlStatusService {
...
-		);
+		const [
+			workflowsResult,
+			credentialsResult,
+			variablesResult,
+			dataTablesResult,
+			tagsMappingsResult,
+			foldersMappingResult,
+			projectsResult,
+		] = await Promise.all([
```
ireneea, line 172: I can see why this is a massive plus for the speed. Really wonder how this translate with large data volume on memory usage 😅

packages/cli/src/modules/source-control.ee/source-control-scoped.service.ts
geemanjs: I did this initially - but then I thought "context" should just be a "thing" that gets passed around rather than needing "service" level DI things and the second best place to put it seemed to be the service 😆

packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
alielkhateeb: I don't even think it's a nit, if we change `StatusResourceOwner['type']` to be  `'personal' | 'team' | 'user';` this part would return undefined instead of actual `StatusResourceOwner` when the type is `user`, so I think IMO Irénée's suggestion is much safer for future changes to not cause hidden runtime issues

packages/cli/src/modules/source-control.ee/source-control-scoped.service.ts
alielkhateeb: I think both perspectives are valid 😂 here’s how I reasoned about it and why I reached the suggestion below:

- It feels right to keep `SourceControlContext` “dumb” (as James suggested)  
  - No dependencies, not a service, and lightweight → easier to pass around  

- It doesn’t feel right to have `createContext` inside `SourceControlScopedService` (as Irénée suggested)  
  - That gives it two responsibilities: constructing the context *and* consuming it

I couldn’t even add a TLDR 😂 My suggestions were too long for it, but here we go:

- Create a `SourceControlContextFactory` service to isolate context creation. This would move the `fetch*` private methods there and keep `SourceControlContext` lightweight and easy to pass around.  
- This would also allow `SourceControlScopedService` to become into something more like a `SourceControlScopeFilterService` (or at least could have a better name IMO).  
- [Optionally] We could cache `hasAccessToAllProjects` and move it to the factory when creating the context. That would avoid recomputing `hasGlobalAccess` on every call and would be stored as a simple `readonly hasAccessToAllProjects: boolean` on SourceControlContext class

packages/cli/src/modules/source-control.ee/source-control-status.service.ee.ts
geemanjs: When testing locally - I found that `master` spiked to `~1GB` of memory - whereas after this PR it spiked to `~750MB` so 🤞 this translates on other instances

packages/cli/src/modules/source-control.ee/source-control-export.service.ee.ts
```diff
@@ -403,19 +403,15 @@ export class SourceControlExportService {
...
-				: existingFolders.folders.filter((folder) => {
-						return !allowedProjects.some((project) => project.id === folder.homeProjectId);
-					});
+			let foldersToKeepUnchanged: ExportableFolder[] = [];
+			if (!context.hasAccessToAllProjects()) {
+				const existingFolders = await readFoldersFromSourceControlFile(fileName);
+				foldersToKeepUnchanged = existingFolders.folders.filter(
+					(folder) => !context.canAccessProject(folder.homeProjectId),
+				);
+			}
```
alielkhateeb, line 414: Changing the logic here allowed us to conditionally call `readFoldersFromSourceControlFile` only when context does NOT have access to all projects, which improves performance further

alielkhateeb (approved):
🔥 🚀 I like how clean this is 😍

PR #27192, opened 2026-03-17
feat(core): Add filtering parameters to get_execution MCP tool

RicardoE105:
## Summary
- Add three optional parameters (`includeData`, `nodeNames`, `truncateData`) to the `get_execution` MCP tool to control response payload size
- By default, only execution metadata is returned (no `data` field), reducing context window consumption for large workflows
- Callers can opt in to full data with `includeData: true`, filter to specific nodes with `nodeNames`, and cap item arrays with `truncateData`
- `nodeNames = []` correctly returns empty data (rather than all data), matching the parameter description
- `resultData.pinData` is now also filtered by `nodeNames`, consistent with `runData` filtering

https://linear.app/n8n/issue/ADO-4942

## Review / Merge checklist
- [x] Tests included.


packages/cli/src/modules/mcp/tools/get-execution.tool.ts
```diff
@@ -54,20 +79,28 @@ export const createGetExecutionTool = (
...
 
 		try {
 			await getMcpWorkflow(workflowId, user, ['workflow:read'], workflowFinderService);
 
-			// Retrieve the execution with full data
-			const execution = await executionRepository.findWithUnflattenedData(executionId, [
-				workflowId,
-			]);
+			// Use lightweight metadata query when data isn't needed
+			const fullExecution = includeData
```
MiloradFilipovic: Nit: Can this be simplified to something like this:
```ts
 const execution = includeData
  	? await executionRepository.findWithUnflattenedData(executionId, [workflowId])
  	: await executionRepository.findIfAccessible(executionId, [workflowId]);
```

packages/cli/src/modules/mcp/tools/get-execution.tool.ts
```diff
@@ -149,3 +192,44 @@ export const createGetExecutionTool = (
...
+	data: IRunExecutionData,
+	nodeNames?: string[],
+	truncateData?: number,
+): IRunExecutionData {
+	// Shallow clone is sufficient — the result is serialized immediately and never mutated
+	const filtered = { ...data, resultData: { ...data.resultData } };
+
+	let runData = filtered.resultData.runData ?? {};
+
+	if (nodeNames?.length) {
```
MiloradFilipovic: If the tool is called with `nodeNames = []` this will skip the filtering and return all node data. I don't think this is expected and is not in line with the parameter description.
Maybe we should distinguish “omitted” from “provided but empty” values here.

packages/cli/src/modules/mcp/tools/get-execution.tool.ts
```diff
@@ -149,3 +192,44 @@ export const createGetExecutionTool = (
...
+		for (const name of nodeNames) {
+			if (name in runData) {
+				filteredRunData[name] = runData[name];
+			}
+		}
+		runData = filteredRunData;
+	}
+
+	if (truncateData) {
+		const truncated: IRunData = {};
```
MiloradFilipovic, line 236: This only trims `runData` but looking at the `IRunExecutionData` type, there are a few more node-scoped fields, like `startData`. Do we want to filter these also?

MiloradFilipovic (commented):
Works nicely 🙌
I left a few comments so take a look if they make sense

MiloradFilipovic (approved):
Thanks for addressing 🙌

PR #27196, opened 2026-03-18
fix(core): Use published version for error workflow execution

### LINKED ISSUE #26692: Button publish bug
### Bug Description

I encountered a situation where unpublished changes in a workflow don’t take effect during production execution if a cron job is used. However, if an error trigger node is used, the production execution runs with the latest version of the workflow, even with unpublished changes.

To clarify:
When using a cron job trigger, the changes that I make in the workflow, but haven’t published, do not reflect in the production execution.
On the other hand, when using an error trigger node, even though the changes are not published, they are still used in the production execution.
This seems like an inconsistency in how the workflow versions are applied, depending on the trigger used. I’ve attached images showing both cases for further clarification.

Image

Image

Image

### To Reproduce

1. Create a new workflow or open an existing one.
2. Make changes to a node (for example, add or modify a node) but do not publish the workflow.
3. Use a Cron Trigger node in the workflow and activate the workflow (e.g., set it to run automatically at scheduled intervals).
4. Observe that the changes you made (but did not publish) are not reflected in the production execution. The workflow runs with the previous (published) version.
5. Next, use an Error Trigger node instead of the Cron Trigger node in the same workflow.
6. Activate the workflow again and observe that the unpublished changes are applied in the production execution, even though the workflow was not published.

### Expected behavior

https://docs.n8n.io/workflows/publish/

### Debug Info

# Debug info

## core

- n8nVersion: 2.9.4
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: postgres
- concurrency: -1
- consumerId: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX

## storage

- success: all
- error: all
- progress: true
- manual: true
- binaryMode: database

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 150000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/145.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-03-06T15:36:29.044Z

### Operating System

docker (self-hosted)

### n8n Version

2.9.4

### Node.js Version

24.13.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

RicardoE105:
## Summary

Error workflows were incorrectly executing with **draft (unpublished) nodes** instead of the published version. When a workflow failed and triggered its configured error workflow, the `executeErrorWorkflow` method loaded the `activeVersion` relation but passed the raw `workflowData` entity (containing draft nodes/connections) to `WorkflowRunner.run()`. The runner then rebuilt the `Workflow` from the draft data, ignoring the published version entirely.

**Root cause:** In `WorkflowExecutionService.executeErrorWorkflow()`, the `workflowInstance` was correctly built from `activeVersion.nodes`, but `workflowData` (with draft nodes) was passed to the runner at line 433. The runner creates a new `Workflow` from `data.workflowData.nodes` (draft), not from the already-validated `workflowInstance`.

**Fix:** Override `workflowData.nodes` and `workflowData.connections` with the `activeVersion` values before passing to the runner, ensuring the published version is used throughout.

### How to reproduce

1. Create an error workflow with an Error Trigger node and publish it
2. Create a main workflow that intentionally fails, set the error workflow in settings, and publish it
3. Add a new node to the error workflow **without publishing**
4. Trigger the main workflow in production mode
5. **Before fix:** The error workflow execution includes the unpublished node
6. **After fix:** The error workflow execution only includes the published nodes

Fixes https://github.com/n8n-io/n8n/issues/26692

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/workflows/workflow-execution.service.ts
+++ b/packages/cli/src/workflows/workflow-execution.service.ts
@@ -318,12 +318,20 @@ export class WorkflowExecutionService {
 			}
 
 			const executionMode = 'error';
+
+			// Use published nodes/connections for execution, not the draft.
+			// The workflowData entity contains draft nodes, but the activeVersion
+			// relation holds the published state. We must override the entity's
+			// nodes/connections so the WorkflowRunner uses the published version.
+			workflowData.nodes = workflowData.activeVersion.nodes;
+			workflowData.connections = workflowData.activeVersion.connections;
+
 			const workflowInstance = new Workflow({
 				id: workflowId,
 				name: workflowData.name,
 				nodeTypes: this.nodeTypes,
-				nodes: workflowData.activeVersion.nodes,
-				connections: workflowData.activeVersion.connections,
+				nodes: workflowData.nodes,
+				connections: workflowData.connections,
 				active: true,
 				staticData: workflowData.staticData,
 				settings: workflowData.settings,
--- a/packages/cli/src/workflows/workflow-execution.service.ts
+++ b/packages/cli/src/workflows/workflow-execution.service.ts
@@ -320,9 +320,6 @@ export class WorkflowExecutionService {
 			const executionMode = 'error';
 
 			// Use published nodes/connections for execution, not the draft.
-			// The workflowData entity contains draft nodes, but the activeVersion
-			// relation holds the published state. We must override the entity's
-			// nodes/connections so the WorkflowRunner uses the published version.
 			workflowData.nodes = workflowData.activeVersion.nodes;
 			workflowData.connections = workflowData.activeVersion.connections;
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/workflows/workflow-execution.service.ts
```diff
@@ -318,12 +318,20 @@ export class WorkflowExecutionService {
 			}
 
 			const executionMode = 'error';
+
+			// Use published nodes/connections for execution, not the draft.
```
dariacodes, line 322: Could you reduce these comments to one line? I think the first one is enough

dariacodes (dismissed):
Thanks for fixing this!

dariacodes (approved):

PR #27206, opened 2026-03-18
feat: Instance AI and local gateway modules (no-changelog)

OlegIvaniv:
## Summary

Implements n8n AI, an assistant that lives inside n8n and can build workflows, manage credentials, run executions, query data tables, browse the web, and interact with the user's local machine.

The system is structured as an orchestrator agent that delegates to specialized sub-agents (workflow builder, web researcher, etc.) and streams progress back to the frontend over SSE. Users interact through a chat interface with human-in-the-loop confirmations for sensitive operations like domain access, setup, and file system changes.

### New packages

`@n8n/instance-ai` is the core agent runtime. It contains the orchestrator and sub-agent factory, all tool definitions (~70 tools across workflows, executions, credentials, nodes, data tables, templates, web research, filesystem, and workspace management), the compaction layer for staying within context window limits, domain access gating, working memory, LangSmith tracing, and the prompt construction pipeline.

`@n8n/fs-proxy` is a local daemon that bridges a remote n8n instance with the user's machine. It exposes filesystem, shell, screenshot, mouse/keyboard, and browser automation as MCP tools over SSE. Designed to run in the system tray or as `npx @n8n/fs-proxy serve`.

`@n8n/local-gateway` is the native Electron tray app wrapper around fs-proxy. It handles connection approval prompts, per-capability permissions (filesystem, screenshots, input simulation, browser, shell), and platform-specific packaging for macOS/Windows.

`@n8n/mcp-browser` is an MCP server for browser automation. It connects to Chrome via the browser extension using CDP, exposes navigation/click/type/snapshot tools, and returns accessibility snapshots with every response so the agent can interact in a single roundtrip.

`@n8n/mcp-browser-extension` is the Chrome extension companion that bridges n8n AI to real browser tabs via CDP, using the user's actual profile, cookies, and sessions.

### Backend (`packages/cli`)

The `instance-ai` module follows controller-service-repository pattern. Controller handles SSE streaming, thread management, confirmations, settings, and memory endpoints. The adapter service connects to an external AI service proxy (or uses a direct LLM connection) and routes all LLM/Brave search/LangSmith calls through it. Thread ownership is enforced at the endpoint level. A compaction service manages context window limits. DB entities cover threads, messages, run snapshots, iteration logs, observational memory, resources, and workflow snapshots, all behind a single consolidated migration.
Background task management supports parallel agent runs with per-task trace isolation.

### Frontend (`packages/frontend/editor-ui`)

The `instanceAi` feature module provides the chat experience: `InstanceAiView` is the main view, backed by a Pinia store and a shared event reducer (also used on the backend for snapshot building). Around 50 Vue components handle the timeline, tool call visualization, artifact previews (workflow canvas, data tables, execution results), domain access approval with severity-aware split buttons, credential setup wizards, memory panels, thread management, and markdown rendering. Composables manage SSE event relay, canvas preview state, execution push events, and resource registration.

The chat integrates into the command bar via `useInstanceAiCommands` and the overall AI feature area alongside the existing assistant and chat hub.

### Shared types (`@n8n/api-types`)

Branded ID types (RunId, AgentId, ThreadId, ToolCallId) prevent accidental swaps. The `instance-ai.schema.ts` defines the full SSE event vocabulary and the `agent-run-reducer.ts` provides a serializable state machine consumed by both frontend and backend. DTOs for confirm requests and thread renames live here too.

## Review / Merge checklist

- [ ] Tests included. 


aalises (dismissed):

pnpm-workspace.yaml
```diff
@@ -20,6 +20,15 @@ catalog:
   '@codemirror/search': 6.5.11
   '@codemirror/state': 6.5.3
   '@codemirror/view': 6.39.8
+  '@mastra/core': 1.9.0
```
Cadiac, line 23: Current version is up at 1.17.0, 1.9.0 is a month stale...
Most of these new packages are a bit stale.

pnpm-workspace.yaml
```diff
@@ -131,7 +143,9 @@ catalogs:
 minimumReleaseAge: 4320
 
 minimumReleaseAgeExclude:
```
Cadiac, line 150: Do we need these new excludes (anymore), not having them would be nice.

packages/@n8n/api-types/src/schemas/agent-run-reducer.ts
```diff
@@ -0,0 +1,442 @@
+/**
+ * Shared event reducer for Instance AI agent runs.
+ *
+ * Used by both the frontend (live SSE updates) and the backend (snapshot building).
+ * All state is plain objects/arrays — no Map/Set — so it's serializable, Pinia-safe,
```
Cadiac, line 5: I had a look and it seems like we're using the `isSafeObjectKey` properly to guard against prototype pollution, but this is something that we need to remember to consider in future too.

packages/@n8n/api-types/src/utils/allowed-domains.ts
```diff
@@ -0,0 +1,231 @@
+/**
+ * Pre-approved documentation domains for web fetch tools.
+ * Shared between ai-workflow-builder and instance-ai packages.
+ */
```
Cadiac, line 4: these being hardcoded and not something configurable seems like something we should address, perhaps not in the first release but later.

Cadiac: We also now have both 

n8n/packages/@n8n/ai-workflow-builder.ee/src/tools/utils/allowed-domains.ts and
n8n/packages/@n8n/api-types/src/utils/allowed-domains.ts

with pretty much the same contents

packages/@n8n/fs-proxy/src/tools/filesystem/fs-utils.ts
```diff
@@ -0,0 +1,210 @@
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+
+import type { AffectedResource } from '../types';
+
+const MAX_ENTRIES = 10_000;
+const DEFAULT_MAX_DEPTH = 8;
+
+export const EXCLUDED_DIRS = new Set([
```
Cadiac, line 9: This being a naive hardcoded list like this also feels pretty unpolished / surprising, even if helpful

packages/@n8n/instance-ai/src/runtime/background-task-manager.ts
```diff
@@ -0,0 +1,188 @@
...
+			task.outcome = typeof raw === 'string' ? undefined : raw.outcome;
+			await options.onCompleted?.(task);
+		} catch (error) {
+			if (task.abortController.signal.aborted) return;
+			task.status = 'failed';
+			task.error = error instanceof Error ? error.message : String(error);
+			await options.onFailed?.(task);
+		}
+
+		await options.onSettled?.(task);
```
Cadiac: Should this await be wrapped in `try ... finally` to clean it up even if something throws?

Cadiac: Addressed in 0853750d31

packages/@n8n/instance-ai/src/runtime/resumable-stream-executor.ts
```diff
@@ -0,0 +1,2146 @@
...
+			const parsedSuspension = parseSuspension(chunk);
+			if (parsedSuspension) {
+				if (!suspension) {
+					suspension = parsedSuspension;
+					if (options.control.mode === 'auto') {
+						options.control.onSuspension?.(parsedSuspension);
+						pendingConfirmation = options.control.waitForConfirmation(parsedSuspension.requestId);
+					}
+				} else if (!isSameSuspension(parsedSuspension, suspension)) {
+					console.warn('[HITL] additional suspension encountered before resume; deferring', {
```
Cadiac, line 1978: We should use a logger at instance-ai package rather than console logs ignoring our usual verbosity config

packages/@n8n/instance-ai/src/runtime/run-state-registry.ts
```diff
@@ -0,0 +1,360 @@
...
+	user: TUser;
+	researchMode?: boolean;
+	messageGroupId?: string;
+}
+
+export interface StartedRunState extends ActiveRunState {
+	messageGroupId?: string;
+}
+
+export class RunStateRegistry<TUser = unknown> {
```
Cadiac, line 71: This will need some major rework for multi-main support, but that's out of scope now..

packages/@n8n/config/src/configs/instance-ai.config.ts
```diff
@@ -0,0 +1,104 @@
...
+	/** Brave Search API key for web search. No key = search + research agent disabled. */
+	@Env('INSTANCE_AI_BRAVE_SEARCH_API_KEY')
+	braveSearchApiKey: string = '';
+
+	/** SearXNG instance URL for web search (e.g. "http://searxng:8080"). Empty = disabled. No API key needed. */
+	@Env('N8N_INSTANCE_AI_SEARXNG_URL')
+	searxngUrl: string = '';
+
+	/** Restrict filesystem access to this directory path. Empty = full filesystem access (reads any path the n8n process can access). */
+	@Env('N8N_INSTANCE_AI_FILESYSTEM_PATH')
```
Cadiac, line 94: I think this comment isn't true. This is used at n8n/packages/cli/src/modules/instance-ai/instance-ai.service.ts,

```
	/** Local filesystem is only available when an explicit base path is configured. */
	isLocalFilesystemAvailable(): boolean {
		return !!this.instanceAiConfig.filesystemPath?.trim();
	}
```

which would return false with the default empty value?

packages/@n8n/instance-ai/src/tools/templates/types.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export interface TemplateWorkflow {
+	name?: string;
+	nodes: TemplateNode[];
+	connections: TemplateConnections;
+}
+
+// ── Template API request/response shapes ────────────────────────────────────
+
+// Retrieved from https://api.n8n.io/api/templates/categories
```
Cadiac, line 42: These can change / new ones can be added.. but I guess that's fine if in future older n8n versions might not find from all available templates

packages/@n8n/instance-ai/src/tools/workflows/build-workflow.tool.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const patchSchema = z.object({
+	old_str: z.string().describe('Exact string to find in the code'),
+	new_str: z.string().describe('Replacement string'),
+});
+
+export function createBuildWorkflowTool(context: InstanceAiContext) {
+	// Keeps the last code submitted (or patched) so patches work even before save,
+	// and always match the LLM's own code — not a roundtripped version.
+	let lastCode: string | null = null;
```
Cadiac, line 20: Can this be from another user of a previous tool call?

packages/@n8n/instance-ai/src/tools/workflows/materialize-node-type.tool.ts
```diff
@@ -0,0 +1,142 @@
...
+						return {
+							nodeId,
+							path: '',
+							content: '',
+							error: result?.error ?? `No type definition found for '${nodeId}'.`,
+						};
+					}
+
+					const version = result.version ?? 'latest';
+					const filePath = `${root}/node-types/${nodeIdToPath(nodeId)}/${version}.ts`;
```
Cadiac, line 102: This should use proper filesystem's resolve (or some safe variant of it) instead of manually constructing the path and assuming linux style path separators - this might also be vulnerable to path traversal through version or nodeId

packages/@n8n/instance-ai/src/workflow-builder/extract-code.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Comprehensive import statement with all available SDK functions.
+ * This is prepended to workflow code so the LLM knows what's available.
+ */
+export const SDK_IMPORT_STATEMENT =
+	"import { workflow, node, trigger, sticky, placeholder, newCredential, ifElse, switchCase, merge, splitInBatches, nextBatch, languageModel, memory, tool, outputParser, embedding, embeddings, vectorStore, retriever, documentLoader, textSplitter, fromAi, expr } from '@n8n/workflow-sdk';";
+
+/** Matches any import statement (single-line, multi-line, side-effect, default, namespace) */
+const IMPORT_REGEX = /^\s*import\s+(?:[\s\S]*?from\s+)?['"]([^'"]+)['"];?\s*$/gm;
```
Cadiac, line 17: doesn't this miss dynamic imports, do we care?

```
const foobar = await import('foobar');
import('foobar').then(m => ...);
```

packages/@n8n/instance-ai/src/workspace/snapshot-manager.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * Builder Image Manager
+ *
+ * Builds and caches a pre-warmed Daytona Image with config files and node_modules
```
Cadiac, line 4: I think this comment isn't accurate - last week when looking into this file it just basically prepares a dockerfile, and doesn't actually build anything yet - the `Image` class doesn't do any real operations yet, it just prepares the image to be created later.

This makes this file to only basically do string operations, constructing the "dockerfile" with little gain in caching, if claude is to be trusted here.

packages/@n8n/instance-ai/src/workspace/snapshot-manager.ts
```diff
@@ -0,0 +1,55 @@
...
+				'mkdir -p /home/daytona/workspace/src /home/daytona/workspace/chunks /home/daytona/workspace/node-types',
+			)
+			.runCommands(
+				`echo '${b64(PACKAGE_JSON)}' | base64 -d > /home/daytona/workspace/package.json`,
+				`echo '${b64(TSCONFIG_JSON)}' | base64 -d > /home/daytona/workspace/tsconfig.json`,
+				`echo '${b64(BUILD_MJS)}' | base64 -d > /home/daytona/workspace/build.mjs`,
+			)
+			.runCommands('cd /home/daytona/workspace && npm install --ignore-scripts');
+
+		console.log(
```
Cadiac, line 44: This should use our logger

packages/@n8n/local-gateway/src/main/daemon-controller.test.ts
```diff
@@ -0,0 +1,5 @@
+describe('DaemonController', () => {
+	it('should be importable', () => {
+		expect(true).toBeDefined();
```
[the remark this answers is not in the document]
Cadiac: +1, this needs real tests

packages/@n8n/local-gateway/src/main/daemon-controller.ts
```diff
@@ -0,0 +1,123 @@
...
+			this.server = null;
+			this.setStatus('stopped');
+		});
+	}
+
+	async disconnectClient(): Promise<void> {
+		logger.debug('Disconnecting client');
+		if (!this.server || this._port === null) return;
+		try {
+			await fetch(`http://localhost:${this._port}/disconnect`, { method: 'POST' });
```
Cadiac, line 84: The host should perhaps also be configurable and not always localhost?
(I'm not that familiar with electron conventions.)

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -0,0 +1,642 @@
...
+					},
+					[userId],
+				);
+			});
+		};
+		req.once('close', cleanup);
+		res.once('finish', cleanup);
+	}
+
+	@Post('/gateway/init', { skipAuth: true })
```
Cadiac, line 545: There should probably be rate limiting on this + the other gateway endpoints, given that they have `skipAuth: true` on them - especially if using `N8N_INSTANCE_AI_GATEWAY_API_KEY` set by the user.

packages/cli/src/modules/instance-ai/filesystem/local-gateway-registry.ts
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * Manages per-user Local Gateway connections.
+ * Each user has their own LocalGateway instance, pairing token, and session key.
+ * Provides a reverse lookup from API key to userId for routing daemon requests.
+ */
+export class LocalGatewayRegistry {
+	private readonly userGateways = new Map<string, UserGatewayState>();
+
+	/** Reverse lookup: pairing token or session key → userId. Used to route daemon requests. */
+	private readonly apiKeyToUserId = new Map<string, string>();
```
Cadiac, line 32: so if n8n restarts all these registered sessions are gone, and the users need to go through the pairing process again? Later on I assume we have plans to store these on the db?

packages/cli/src/modules/instance-ai/filesystem/local-gateway-registry.ts
```diff
@@ -0,0 +1,208 @@
...
+	startDisconnectTimer(userId: string, onDisconnect: () => void): void {
+		const state = this.getOrCreate(userId);
+		this.clearDisconnectTimer(userId);
+		const graceMs = Math.min(INITIAL_GRACE_MS * Math.pow(2, state.reconnectCount), MAX_GRACE_MS);
+		state.reconnectCount++;
+		state.disconnectTimer = setTimeout(() => {
+			state.disconnectTimer = null;
+			this.disconnectGateway(userId);
+			// Session key is kept alive so the daemon can re-authenticate on reconnect.
+			// It is only cleared on explicit /gateway/disconnect.
```
Cadiac, line 186: These session keys never expiring / rotating might be risky, I would love to see some kind of a refresh token mechanism with expiring short lived session tokens at least (later when this is more mature)

packages/cli/src/modules/instance-ai/web-research/summarize-content.ts
```diff
@@ -0,0 +1,48 @@
...
+	const prompt = `Summarize the following web page content. Preserve:
+- API endpoints, URLs, and authentication details
+- Code examples and configuration snippets
+- Technical specifications and requirements
+- Step-by-step instructions
+
+Remove navigation, ads, footers, and repetitive content. Output markdown.
+
+---
+${page.content}`;
```
Cadiac, line 38: `generateFn` is never passed (this is only called from instance-ai.adapter.service.ts) so this is essentially dead code for now, but this is one of the spots where we should attempt to guard against prompt injection in the future,

packages/cli/src/modules/instance-ai/instance-ai-memory.service.ts
```diff
@@ -0,0 +1,356 @@
...
+		// Fetch agent tree snapshots: DB table (new) + legacy metadata fallback
+		const dbSnapshots = await this.dbSnapshotStorage.getAll(threadId).catch((error) => {
+			this.logger.warn('Failed to load DB snapshots, falling back to metadata', {
+				threadId,
+				error: error instanceof Error ? error.message : String(error),
+			});
+			return [];
+		});
+
+		// Legacy metadata snapshots for threads created before the table migration
```
Cadiac: I think we don't need this, no need to support threads from early development time

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -0,0 +1,674 @@
...
+		private readonly instanceAiService: InstanceAiService,
+		private readonly memoryService: InstanceAiMemoryService,
+		private readonly settingsService: InstanceAiSettingsService,
+		private readonly eventBus: InProcessEventBus,
+		private readonly moduleRegistry: ModuleRegistry,
+		private readonly push: Push,
+		globalConfig: GlobalConfig,
+	) {
+		this.gatewayApiKey = globalConfig.instanceAi.gatewayApiKey;
+		this.instanceBaseUrl = globalConfig.editorBaseUrl || `http://localhost:${globalConfig.port}`;
```
Cadiac, line 66: the fallback should also use `globalConfig.host`

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -0,0 +1,674 @@
...
+	@Get('/settings/service-credentials')
+	@GlobalScope('instanceAi:manage')
+	async listServiceCredentials(req: AuthenticatedRequest) {
+		return await this.settingsService.listServiceCredentials(req.user);
+	}
+
+	@Get('/memory/:threadId')
+	@GlobalScope('instanceAi:message')
+	async getMemory(req: AuthenticatedRequest, _res: Response, @Param('threadId') threadId: string) {
+		return await this.memoryService.getWorkingMemory(req.user.id, threadId);
```
Cadiac, line 379: isn't this missing `this.memoryService.checkThreadOwnership` check?

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -0,0 +1,674 @@
...
+
+	@Put('/memory/:threadId')
+	@GlobalScope('instanceAi:message')
+	async updateMemory(
+		req: AuthenticatedRequest,
+		_res: Response,
+		@Param('threadId') threadId: string,
+		@Body payload: InstanceAiUpdateMemoryRequest,
+	) {
+		await this.memoryService.updateWorkingMemory(req.user.id, threadId, payload.content);
```
Cadiac, line 391: This is also missing `this.memoryService.checkThreadOwnership` check?

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	@Get('/threads/:threadId/context')
+	@GlobalScope('instanceAi:message')
+	async getThreadContext(
+		req: AuthenticatedRequest,
+		_res: Response,
+		@Param('threadId') threadId: string,
+	) {
+		return await this.memoryService.getThreadContext(req.user.id, threadId);
```
Cadiac, line 488: This also appears to be missing `this.memoryService.checkThreadOwnership` check of the thread?

Cadiac: Addressed in 739b2ac1948b34caff6e221c54d9ed07786fc802

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
Cadiac: addressed in 739b2ac1948b34caff6e221c54d9ed07786fc802

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
Cadiac: addressed in 739b2ac1948b34caff6e221c54d9ed07786fc802

packages/nodes-base/nodes/Set/test/v2/raw.test.ts
```diff
@@ -109,15 +111,13 @@ describe('test Set2, rawMode/json Mode', () => {
 		};
 
 		it('should parse json with resolved expression data and compose a return item', async () => {
-			jest.spyOn(utils, 'parseJsonParameter');
-			jest.spyOn(utils, 'composeReturnItem');
```
Cadiac, line 113: what happened to `composeReturnItem`, seems unexpected?

packages/cli/src/modules/instance-ai/internal-messages.ts
```diff
@@ -0,0 +1,27 @@
...
+ * are LLM-facing only — they must never reach the UI.
+ *
+ * The service writes this format,
+ * the parser reads it (cleanStoredUserMessage).
+ */
+
+/** Legacy sentinel used for old auto-follow-up runs — kept for parsing historical messages. */
+export const AUTO_FOLLOW_UP_MESSAGE = '(continue)';
+
+/** Matches the legacy `<background-tasks>`, current `<running-tasks>`, planned-task follow-up, or background-task-completed prefix. */
```
Cadiac, line 15: Removed <background-tasks> from here in https://github.com/n8n-io/n8n/pull/27860

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -0,0 +1,2141 @@
...
+	type PlannedTaskGraph,
+	type PlannedTaskRecord,
+	type SandboxConfig,
+	type SpawnBackgroundTaskOptions,
+	type ServiceProxyConfig,
+	type StreamableAgent,
+	WorkflowTaskCoordinator,
+	WorkflowLoopStorage,
+} from '@n8n/instance-ai';
+import { createAnthropic } from '@ai-sdk/anthropic';
```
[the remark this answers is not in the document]
Cadiac: https://github.com/n8n-io/n8n/pull/27862 addressed here

packages/cli/src/modules/instance-ai/node-definition-resolver.ts
Cadiac, line 1: Refactored this to use `safeJoinPath` from n8n's core instead of implementing its own path traversal protection at https://github.com/n8n-io/n8n/pull/27878, will merge that later on.

packages/cli/src/modules/mcp/tools/workflow-builder/workflow-builder-tools.service.ts
```diff
@@ -19,6 +36,20 @@ export class WorkflowBuilderToolsService {
...
 	private initPromise: Promise<void> | undefined;
 
+	// Cached tool instances — created once, reused across invocations
+	private searchTool: InvokableTool<{ queries: string[] }> | undefined;
+
+	private getTool: InvokableTool<{ nodeIds: NodeRequest[] }> | undefined;
+
+	private suggestTool: InvokableTool<{ categories: string[] }> | undefined;
+
+	// Result caches — keyed by normalized input, cleared on node type refresh
```
Cadiac, line 46: I wonder how much memory these caches end up taking under use as they only get cleared on node types refresh - something to investigate later

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+	workflowSnapshots: 'instance_ai_workflow_snapshots',
+	runSnapshots: 'instance_ai_run_snapshots',
+	iterationLogs: 'instance_ai_iteration_logs',
+} as const;
+
+export class CreateInstanceAiTables1775000000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(table.threads)
+			.withColumns(
+				column('id').varchar().primary.notNull,
```
tomi: Are this and/or the other IDs in these tables uuids? If so, we could use the `.uuid` data type which on PG takes half the space compared to storing it as varchar

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+				column('resourceId').varchar().notNull,
+				column('title').text.default("''").notNull,
+				column('metadata').json,
+			)
+			.withIndexOn('resourceId').withTimestamps;
```
tomi, line 22: Is this a FK to resources table?

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+
+		await createTable(table.messages)
+			.withColumns(
+				column('id').varchar().primary.notNull,
+				column('threadId').varchar().notNull,
+				column('content').text.notNull,
+				column('role').varchar(16).notNull,
+				column('type').varchar(32),
+				column('resourceId').varchar(),
```
tomi: Is this a FK to resources table?

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+		await createTable(table.resources).withColumns(
+			column('id').varchar().primary.notNull,
+			column('workingMemory').text,
+			column('metadata').json,
+		).withTimestamps;
+
+		await createTable(table.observationalMemory)
+			.withColumns(
+				column('id').varchar().primary.notNull,
+				column('lookupKey').varchar().notNull,
```
tomi: How is this generated? Should it have some upper bound?

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+				column('scope').varchar(16).notNull,
+				column('threadId').varchar(),
```
tomi: Should this be a FK to threads table?

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+				column('resourceId').varchar().notNull,
```
tomi: Should this be a FK to resources table?

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
```diff
@@ -0,0 +1,134 @@
...
+
+		await createTable(table.workflowSnapshots)
+			.withColumns(
+				column('runId').varchar().primary.notNull,
+				column('workflowName').varchar().primary.notNull,
+				column('resourceId').varchar(),
+				column('status').varchar(),
+				column('snapshot').text.notNull,
+			)
+			.withIndexOn('workflowName')
```
tomi: This is unnecessary index, the other covers this one as well

tomi (commented):
Couple comments regarding the migrations

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: some of them are, but for instance on `instance_ai_resources` the IDs can be like `e2b24ef6-9023-4e00-9ab2-bdea9c549833:data-table-manager`. I'll study these and see if some of them can be replaced with UUID.

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: No, in the case of threads `resourceId` gets value from the user ID of the actor. We use that to find threads later (and to verify their ownership).

Cadiac: Or well, technically yes - the `instance_ai_resources` gets its ID values from the same user ID references - but since those are actually referring to user I don't think FK is fully correct here, and also the ID column on `instance_ai_resources` might also get these kinds of `uuid:role` composite keys where just the prefix refers to the user and no matching rows exist under resources:

Image

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: see comment above, yes kind of, but no

Image

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: It gets values like 
`${scope}:${threadId ?? 'null'}:${resourceId}`, making it varchar(255)

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: adding FK with cascade set null

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: leaving these resourceId fk's out

packages/@n8n/db/src/migrations/common/1775000000000-CreateInstanceAiTables.ts
Cadiac: Seems like most of these are nanoIds with length of 16, some with some prefixes - gave them constraint of varchar(36) and made the threadId uuid since it actually was UUID, but left the rest as varchars

tomi (dismissed):
Migrations LGTM 👍

packages/cli/src/modules/instance-ai/entities/instance-ai-message.entity.ts
```diff
@@ -0,0 +1,25 @@
+import { WithTimestamps } from '@n8n/db';
+import { Column, Entity, Index, PrimaryColumn } from '@n8n/typeorm';
+
+@Entity({ name: 'instance_ai_messages' })
+export class InstanceAiMessage extends WithTimestamps {
+	@PrimaryColumn({ type: 'varchar', length: 36 })
+	id: string;
+
+	@Index()
+	@Column({ type: 'uuid' })
```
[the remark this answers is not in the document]
mutdmour: @Cadiac worth addressing this?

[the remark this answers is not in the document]
Cadiac: kind of valid but not really, we had some defensive fallback to empty string.
made it throw instead in b7993b38fbad1029d2e83030fe398b8098cc25bf. I don't think we ever hit that path in real life but mastra types it optional

tomi (approved):

tomi (dismissed):

tomi (dismissed):

tomi (dismissed):

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -4965,6 +4966,11 @@
 	"commandBar.chat.deleteSession": "Delete \"{title}\"",
 	"commandBar.chat.open.searchPlaceholder": "Search by f title...",
 	"commandBar.chat.stop": "Stop message generation",
+	"commandBar.sections.instanceAi": "Instance AI",
```
Cadiac, line 4969: These should get renamed

tomi (dismissed):

tomi (dismissed):

PR #27211, opened 2026-03-18
fix(core): Fix race condition when stopping jobs in queue mode

ivov:
## Summary

Fix the race condition on stopping an execution in queue mode where the main process calls `job.moveToFailed()` (which deletes the job key from Redis) while the worker's Bull processor is still running, causing `Missing key for job updateProgress` errors.

To reproduce:

1. Start main and worker in queue mode with `OFFLOAD_MANUAL_EXECUTIONS_TO_WORKERS=true`
2. Create a workflow: Webhook → Code node with `await new Promise(r => setTimeout(r, 120_000))`
3. Publish the workflow and trigger the webhook
4. In the UI, go to Executions → find the running execution → click Stop
5. **Before fix**: Worker logs `Missing key for job <id> updateProgress` and `Missing key for job <id> failed`, execution appears stuck in UI
6. **After fix**: Worker throws `ManualExecutionCancelledError`, Bull moves the job to failed cleanly, no missing key errors

https://linear.app/n8n/issue/CAT-2616

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/scaling/scaling.service.ts
+++ b/packages/cli/src/scaling/scaling.service.ts
@@ -5,14 +5,7 @@ import { ExecutionRepository } from '@n8n/db';
 import { OnLeaderStepdown, OnLeaderTakeover, OnShutdown } from '@n8n/decorators';
 import { Container, Service } from '@n8n/di';
 import { ErrorReporter, InstanceSettings } from 'n8n-core';
-import {
-	BINARY_ENCODING,
-	sleep,
-	jsonStringify,
-	ensureError,
-	UnexpectedError,
-	ManualExecutionCancelledError,
-} from 'n8n-workflow';
+import { BINARY_ENCODING, sleep, jsonStringify, ensureError, UnexpectedError } from 'n8n-workflow';
 import type { IExecuteResponsePromiseData, IRun } from 'n8n-workflow';
 import assert, { strict } from 'node:assert';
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/scaling/scaling.service.ts
```diff
@@ -264,8 +257,7 @@ export class ScalingService {
 		try {
 			if (await job.isActive()) {
 				await job.progress({ kind: 'abort-job' }); // being processed by worker
-				await job.discard(); // prevent retries
-				await job.moveToFailed(new ManualExecutionCancelledError(job.data.executionId), true); // remove from queue
+				this.logger.debug('Sent abort signal to worker', props);
```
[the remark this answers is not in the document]
ivov: As explained, this PR is only for the race condition.

packages/cli/src/scaling/scaling.service.ts
```diff
@@ -264,8 +257,7 @@ export class ScalingService {
 		try {
 			if (await job.isActive()) {
 				await job.progress({ kind: 'abort-job' }); // being processed by worker
```
tomi, line 259: What happens if this signal never reaches the worker? E.g. if worker has crashed or due to network issues. What happens to the job, will it still be cancelled?

Also where does the execution now get moved to canceled state?

ivov: > What happens if this signal never reaches the worker? E.g. if worker has crashed or due to network issues. What happens to the job, will it still be cancelled?

Signal delivery failure is a limitation of Redis pubsub being fire and forget. This PR [scopes out that failure mode](https://linear.app/n8n/issue/CAT-2616/stopped-executions-continue-running-on-workers-in-queue-mode#comment-71393165) but, if the worker is alive but signal is lost, the job finishes normally, whereas if the worker crashed even though the signal was sent correctly, the job is marked as failed.

> Also where does the execution now get moved to canceled state?

 In thw worker, `workflowRun` resolves with `run.status === 'canceled'` so the worker throws `ManualExecutionCancelledError`. Bull catches that error from the processor callback and moves the job to failed state, i.e. Bull's own lifecycle handles the Redis key cleanup.

tomi (approved):

PR #27215, opened 2026-03-18
feat(Microsoft Agent 365 Trigger Node): Mcp tools logs

michael-radency:
## Summary

adds logs from ms mcp tools to the agent output

https://linear.app/n8n/issue/NODE-4604/check-can-we-add-microsoft-mcp-tool-calls-to-the-n8n-logs

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/nodes-langchain/nodes/vendors/Microsoft/microsoft-utils.ts
```diff
@@ -203,12 +216,31 @@ export async function getMicrosoftMcpTools(
...
-			const callToolFunc = createCallTool(tool.name, client, timeout, (errorMessage) => {
-				console.error(`Tool "${tool.name}" execution error:`, errorMessage);
-			});
-			return mcpToolToDynamicTool({ ...tool, name: prefixedName }, callToolFunc);
+			const prefixedName = buildMcpToolName(serverName, tool.name);
+
+			const callToolWithLogging = async (args: IDataObject) => {
+				let isError = false;
+				const callTool = createCallTool(tool.name, client, timeout, (errorMessage) => {
+					console.error(`Tool "${tool.name}" execution error:`, errorMessage);
```
RomanDavydchuk, line 238: Is `console.error` fine here or should we use a logger?

packages/@n8n/nodes-langchain/nodes/vendors/Microsoft/microsoft-utils.ts
```diff
@@ -203,12 +216,31 @@ export async function getMicrosoftMcpTools(
...
+					isError = true;
+				});
+				const start = Date.now();
+				const result = await callTool(args);
```
RomanDavydchuk, line 242: Will this throw in a case of an error? I assume no, since `createCallTool` accepts on error callback, but want to double check, since if it does throw, the result won't be added to `mcpToolCallLogs`

RomanDavydchuk (changes requested):
Looks good mostly, a couple of questions:

packages/@n8n/nodes-langchain/nodes/vendors/Microsoft/microsoft-utils.ts
michael-radency: it's fine until in the beta state

packages/@n8n/nodes-langchain/nodes/vendors/Microsoft/microsoft-utils.ts
michael-radency: it would not, `createCallTool` does not throws
```
export const createCallTool =
	(name: string, client: Client, timeout: number, onError: (error: string) => void) =>
	async (args: IDataObject) => {
		let result: Awaited<ReturnType<Client['callTool']>>;

		function handleError(error: unknown) {
			const errorDescription =
				getErrorDescriptionFromToolCall(error) ?? `Failed to execute tool "${name}"`;
			onError(errorDescription);
			return errorDescription;
		}

		try {
			result = await client.callTool({ name, arguments: args }, CompatibilityCallToolResultSchema, {
				timeout,
			});
		} catch (error) {
			return handleError(error);
		}

		if (result.isError) {
			return handleError(result);
		}

		if (result.toolResult !== undefined) {
			return result.toolResult;
		}

		if (result.content !== undefined) {
			return result.content;
		}

		return result;
	};
```

RomanDavydchuk (approved):
LGTM

PR #27222, opened 2026-03-18
feat(instance-ai): add rolling compaction with centralized model registry

aalises:
## Summary

Closes https://linear.app/n8n/issue/AI-2233

- Adds automatic rolling compaction of older thread messages into a structured summary to maintain long-range context in Instance AI conversations
- New `InstanceAiCompactionService` (CLI) + `generateCompactionSummary` helper (`@n8n/instance-ai`) handle the full lifecycle
- Compaction state stored in thread metadata (`instanceAiConversationSummary`) — no DB migration needed
- Best-effort, non-blocking: failures log a warning and proceed without summary
- Subtle "Recalling conversation..." status indicator via new `status` streaming event
- Context window lookup uses the **Chat Hub model registry** (`context-limits.ts`) — single source of truth covering 100+ models across all providers

## How it works

1. Before each agent stream, `prepareCompactedContext()` estimates the total token usage across all thread messages
2. If total tokens exceed **80% of the model's context window**, compaction triggers on the older message prefix (before the `lastMessages` tail)
3. A lightweight Mastra Agent (no tools, no memory) generates a rolling summary with five fixed sections: Goal, Important facts, Workflow state, Open issues, Likely next step
4. The summary is prepended to the user message as a `<conversation-summary>` XML block
5. System prompt instructs the agent to use the summary for long-range continuity and recent raw messages for exact wording

## Compaction trigger

The trigger is **token-based**, matching the pattern used by Claude Code and OpenAI's auto-compaction:

- **Token estimation**: `ceil(chars / 4)` across all thread messages + 8k fixed overhead (system prompt, tools, working memory)
- **Context window**: looked up from the **Chat Hub model registry** (`maxContextWindowTokens` in `context-limits.ts`), which covers all supported providers (OpenAI, Anthropic, Google, Groq, Mistral, xAI, DeepSeek, etc.). Uses exact match first, then prefix match for short aliases (e.g. `claude-sonnet-4-5` → `claude-sonnet-4-5-20250929`), with a conservative 128k fallback for unknown models.
- **Threshold**: compacts when total estimated tokens exceed **80%** of the model's context window (configurable via `compactionThreshold` parameter, 0–1)
- **Minimum delta**: at least 500 unsummarized tokens required to avoid trivial compaction calls
- **Incremental**: each compaction only processes the delta since the last `upToMessageId`

Example: with Claude Sonnet (200k window), compaction triggers when thread messages total ~160k+ estimated tokens.

## Key design decisions

- **Non-destructive**: raw messages stay in storage/UI — only model input is compacted
- **Token-based trigger**: context percentage, not message count — a few large tool-result messages trigger earlier than many short ones
- **Centralized model registry**: context window sizes are looked up from the Chat Hub's `maxContextWindowTokens` map rather than hardcoded per model family — new models are picked up automatically when the registry is updated
- **High-signal only**: user/assistant text is preserved; tool payloads, binary data, and execution outputs are dropped from the summary input
- **Thread-scoped**: summary is per-thread, not user-global
- **Input order**: `<conversation-summary>` → `<background-tasks>` → user message
- **Status UX**: pulsing dot + "Recalling conversation..." label appears before first token, via new `status` streaming event type

### DIFF
--- a/packages/cli/src/modules/instance-ai/compaction.service.ts
+++ b/packages/cli/src/modules/instance-ai/compaction.service.ts
@@ -1,0 +1,229 @@
+import { Logger } from '@n8n/backend-common';
+import { Service } from '@n8n/di';
+import { generateCompactionSummary } from '@n8n/instance-ai';
+import type { ModelConfig } from '@n8n/instance-ai';
+import type { Memory } from '@mastra/memory';
+import type { MastraMessageContentV2 } from '@mastra/core/agent';
+import type { MastraDBMessage } from '@mastra/core/storage';
+
+import { TypeORMMemoryStorage } from './storage/typeorm-memory-storage';
+
+const METADATA_KEY = 'instanceAiConversationSummary';
+
+interface ConversationSummaryMetadata {
+	version: number;
+	upToMessageId: string;
+	summary: string;
+	updatedAt: string;
+}
+
+/**
+ * Manages rolling compaction of older thread messages into a summary.
+ * Stores compaction state in thread metadata — no DB migration needed.
+ *
+ * Design:
+ *   - recent tail (lastMessages) is never compacted
+ *   - only the older prefix before the tail gets summarized
+ *   - raw messages stay in storage for debugging/UI — only model input is compacted
+ *   - compaction is best-effort: failures log a warning and return null
+ */
+@Service()
+export class InstanceAiCompactionService {
+	constructor(
+		private readonly logger: Logger,
+		private readonly memoryStorage: TypeORMMemoryStorage,
+	) {}
+
+	/**
+	 * If compaction is needed, generate a summary and return a formatted
+	 * `<conversation-summary>` block for prompt injection. Returns null if
+	 * compaction is not needed or if a cached summary is still valid.
+	 */
+	async prepareCompactedContext(
+		threadId: string,
+		memory: Memory,
+		modelId: ModelConfig,
+		lastMessages: number,
+	): Promise<string | null> {
+		try {
+			const recentTail = lastMessages;
+			const compactBatch = Math.max(6, Math.floor(recentTail / 2));
+
+			// Load all messages for the thread, ordered chronologically
+			const { messages: allMessages } = await this.memoryStorage.listMessages({
+				threadId,
+				perPage: false,
+				orderBy: { field: 'createdAt', direction: 'ASC' },
+			});
+
+			if (allMessages.length <= recentTail) {
+				// Thread is short enough — no compaction needed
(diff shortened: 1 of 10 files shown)


packages/cli/src/modules/instance-ai/compaction.service.ts
```diff
@@ -0,0 +1,307 @@
+import type { ChatHubLLMProvider } from '@n8n/api-types';
+import { Logger } from '@n8n/backend-common';
+import { Service } from '@n8n/di';
+import { generateCompactionSummary } from '@n8n/instance-ai';
+import type { ModelConfig } from '@n8n/instance-ai';
+import type { Memory } from '@mastra/memory';
+import type { MastraMessageContentV2 } from '@mastra/core/agent';
+import type { MastraDBMessage } from '@mastra/core/memory';
+
+import { maxContextWindowTokens } from '@/modules/chat-hub/context-limits';
```
Cadiac, line 10: the accuracy of these might not be perfect, we should actually add the new openAI models here for instance 😅 these were mostly built by hand / AI last year

packages/cli/src/modules/instance-ai/compaction.service.ts
```diff
@@ -0,0 +1,307 @@
...
+const METADATA_KEY = 'instanceAiConversationSummary';
+
+const DEFAULT_CONTEXT_WINDOW = 128_000;
+
+/**
+ * Rough token estimate: ~4 chars per token for English text.
+ * Good enough for triggering decisions — not used for billing.
+ */
+function estimateTokens(text: string): number {
+	return Math.ceil(text.length / 4);
```
Cadiac, line 23: Langchain shipped with `trimMessages` (that required knowing the model) to do trimming and we overcomplicated things there on chat hub side, this is probably better approach here 👍

packages/cli/src/modules/instance-ai/compaction.service.ts
aalises: Yeah... figured out for a first version we could be good with this. Feel like depending on the model we have to maintain a list of tokens unless we can somehow get it from the provider itself which I don't think

Cadiac (approved):
👍 seems like a pretty clean way to do this

packages/cli/src/modules/instance-ai/compaction.service.ts
Cadiac: we can probably ship with this for now, maybe update the list at some point

Cadiac: and at least with langchain it didn't seem to be easy to get hte limits by provider in a generic way at least, which is why we ended up with this list.

PR #27240, opened 2026-03-18
feat(core): Introduce expression-based role resolver for SSO claim mapping (no-changelog)

afitzek:
## Summary

Adds the foundational role resolver service for expression-based SSO role mapping (IAM feature, behind feature flag — not yet wired).

- **`RoleResolverService`** — evaluates admin-defined mapping rules against SSO claims using n8n's sandboxed expression engine (`Expression.resolveWithoutWorkflow`). Instance role rules use first-match-wins; project role rules use first-match-wins per project. Failing expressions are safely treated as `false`.
- **`ClaimsContextBuilder`** — builds the `$claims`, `$oidc`, `$saml`, and `$project` context objects that expressions can reference. OIDC merges ID token + userInfo; SAML uses raw attributes. `$project` context (id, name, type, description) is injected per-rule during project role evaluation.
- **Types** — `RoleMappingRule`, `RoleMappingConfig`, `ResolvedRoles`, `RoleResolverContext`, `ProjectInfo`
- Project info is fetched in a single batch query before expression evaluation, keeping DB transactions short and isolated from potentially slow expressions.
- 74 unit tests covering both protocols, ordering, fallback, disabled rules, missing projects, and error handling.

https://linear.app/n8n/issue/IAM-394/role-resolver-expression-engine

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/provisioning.ee/claims-context.builder.ts
+++ b/packages/cli/src/modules/provisioning.ee/claims-context.builder.ts
@@ -1,0 +1,37 @@
+import type { ProjectInfo, RoleResolverContext } from './role-resolver-types';
+
+export function buildOidcClaimsContext(
+	idTokenClaims: Record<string, unknown>,
+	userInfo: Record<string, unknown>,
+): RoleResolverContext {
+	return {
+		$claims: structuredClone({ ...idTokenClaims, ...userInfo }),
+		$oidc: {
+			idToken: structuredClone(idTokenClaims),
+			userInfo: structuredClone(userInfo),
+		},
+		$provider: 'oidc',
+	};
+}
+
+export function buildSamlClaimsContext(
+	rawAttributes: Record<string, unknown>,
+): RoleResolverContext {
+	return {
+		$claims: structuredClone(rawAttributes),
+		$saml: {
+			attributes: structuredClone(rawAttributes),
+		},
+		$provider: 'saml',
+	};
+}
+
+export function withProjectContext(
+	context: RoleResolverContext,
+	project: ProjectInfo,
+): RoleResolverContext {
+	return {
+		...context,
+		$project: { ...project },
+	};
+}
--- a/packages/cli/src/modules/provisioning.ee/role-resolver-types.ts
+++ b/packages/cli/src/modules/provisioning.ee/role-resolver-types.ts
@@ -1,0 +1,50 @@
+export interface RoleResolverContext {
+	/** Protocol-agnostic merged claims */
+	$claims: Record<string, unknown>;
+
+	/** OIDC-specific data */
+	$oidc?: {
+		idToken: Record<string, unknown>;
+		userInfo: Record<string, unknown>;
+	};
+
+	/** SAML-specific data */
+	$saml?: {
+		attributes: Record<string, unknown>;
+	};
+
+	// Future: $ldap?: { attributes: Record<string, unknown> }
+
+	$provider: 'oidc' | 'saml' | 'ldap';
+
+	/** Populated per-rule during project role evaluation */
+	$project?: ProjectInfo;
+}
+
+export interface RoleMappingRule {
+	id: string;
+	expression: string;
+	role: string;
+	/** undefined = instance role rule; set = project role rule */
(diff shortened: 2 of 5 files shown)


packages/cli/src/modules/provisioning.ee/role-resolver.service.ee.ts
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		return result;
+	}
+
+	private evaluateExpression(expression: string, context: RoleResolverContext): boolean {
+		try {
+			const result = Expression.resolveWithoutWorkflow(
+				expression,
+				context as unknown as IDataObject,
```
BGZStephen, line 115: Why do you need to cast here?

packages/cli/src/modules/provisioning.ee/role-resolver-types.ts
```diff
@@ -0,0 +1,50 @@
+export interface RoleResolverContext {
```
BGZStephen, line 1: Does one generic interface make sense? or a discriminated interface based on type to allow stricter enforcement for a `attributes`, `idToken` and `userInfo`?

packages/cli/src/modules/provisioning.ee/role-resolver.service.ee.ts
```diff
@@ -0,0 +1,126 @@
...
+			);
+			return String(result) === 'true';
+		} catch (error) {
+			this.logger.warn('Role resolver expression evaluation failed, treating as false', {
+				expression,
+				error: error instanceof Error ? error.message : String(error),
```
BGZStephen, line 121: Can this error ever include the idToken or userInfo for the request? May not want to log PII

packages/cli/src/modules/provisioning.ee/claims-context.builder.ts
```diff
@@ -0,0 +1,37 @@
...
+		$claims: structuredClone({ ...idTokenClaims, ...userInfo }),
+		$oidc: {
+			idToken: structuredClone(idTokenClaims),
+			userInfo: structuredClone(userInfo),
+		},
+		$provider: 'oidc',
+	};
+}
+
+export function buildSamlClaimsContext(
```
phyllis-noester, line 17: just for my understanding: we would prob call this from getAttributesFromLoginResponse or in a following step, correct?

packages/cli/src/modules/provisioning.ee/role-resolver.service.ee.ts
```diff
@@ -0,0 +1,126 @@
...
+	constructor(
+		private readonly logger: Logger,
+		private readonly projectRepository: ProjectRepository,
+	) {}
+
+	async resolveRoles(
+		config: RoleMappingConfig,
+		context: RoleResolverContext,
+	): Promise<ResolvedRoles> {
+		const projects = await this.collectProjectInfoForRules(config.projectRoleRules);
```
phyllis-noester, line 27: do we need to do that? given that we will let users select a project from a drop down I think we could also just try assigning the roles and then fail if role/project doesn't exist

phyllis-noester (approved):
some questions but nothing urgent. nice :)

packages/cli/src/modules/provisioning.ee/role-resolver-types.ts
afitzek: This will be handed to the expression engine and just builds the context that might be available for the expression, I think there is not really value in a discriminated interface.

packages/cli/src/modules/provisioning.ee/role-resolver.service.ee.ts
afitzek: This is to build up the project information to be available inside an expression. So that in an expression we can use `$project....` to access project information and make a decision based on it.

packages/cli/src/modules/provisioning.ee/claims-context.builder.ts
afitzek: Yes, these are like the factory methods.

PR #27241, opened 2026-03-18
docs: Add .env.local.example template and dotenvx for local dev setup (no-changelog)

JoseBra:
## Summary

- Adds `.env.local.example` with a minimal set of environment variables for local development (license, AI, LangSmith tracing, data folder isolation)
- Installs `@dotenvx/dotenvx` as a devDependency so contributors can load the file with variable expansion support (e.g. `$HOME`)
- Updates `CONTRIBUTING.md` with copy-paste commands for loading env vars during development

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -1,0 +1,59 @@
+# =============================================================================
+# n8n local development — minimal environment variables
+# =============================================================================
+# This is a minimal example covering some local environment variables for development.
+# Many more variables exist — search for @Env() decorators in the codebase.
+# Most of them already have default values, so you only need to fill in the ones you need to change.
+#
+# Usage (run from the repo root):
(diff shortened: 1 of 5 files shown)


Matsuuu (changes requested):
Dotenv is only run from the Cli. Not on the project, right? Can we just instruct developers to use `npx` or an alternative instead of adding a dependency to the whole project?

JoseBra:
> Dotenv is only run from the Cli. Not on the project, right? Can we just instruct developers to use `npx` or an alternative instead of adding a dependency to the whole project?

Yes, we could use a global instalation if you don't want to explicitly add it to the project

Matsuuu (approved):
🚀

PR #27248, opened 2026-03-18
feat(core): Add search_projects, search_folders MCP tools and folderId to create_workflow

RicardoE105:
## Summary
- Add `search_projects` MCP tool to find accessible projects by name/type
- Add `search_folders` MCP tool to find folders within a project (with `folder:list` scope authorization)
- Add `folderId` parameter to `create_workflow_from_code` tool, with validation that `projectId` is required when `folderId` is set
- Register both new tools in the MCP builder tools section

## Related Linear ticket
https://linear.app/n8n/issue/ADO-4942

## Test plan
- [x] Unit tests for `search_projects` tool (5 tests)
- [x] Unit tests for `search_folders` tool (5 tests, including access control)
- [x] Unit test for `folderId` without `projectId` validation
- [x] Existing `mcp.service.test.ts` tests updated and passing
- [x] Typecheck passes
- [x] Lint passes


MiloradFilipovic (approved):
Perfect, thanks 🚀
The only problem I have with it is that it requires `projectId` when searching for folders, so if I specify just a folder name it needs to go trough all projects. This works well though, I am just worried about token consumption but we can optimize later if that turns out to be a problem.

PR #27251, opened 2026-03-18
feat(Google Sheets Node): Add multi-sheet read support

### LINKED ISSUE #35877: Google Sheets Node: Get Row(s) cannot read every sheet in a document
### Bug Description

The Google Sheets node's Get Row(s) operation requires selecting exactly one sheet. There is no way to read every sheet of a spreadsheet in a single node.

This is a common shape for spreadsheets partitioned by month, region, or client, where every tab shares the same header row and the data is only split across tabs for human convenience.

Working around it today means knowing every sheet name up front and wiring one branch (or one loop iteration) per sheet. That is brittle: the node offers no way to enumerate a document's sheets at runtime, so the workflow silently misses data the moment somebody adds a tab, and errors when somebody renames one.

I've opened PR #27251 implementing this.

### To Reproduce

1. Create a Google Sheets document with several tabs sharing the same header row, e.g. Jan, Feb, Mar
2. Add a Google Sheets node and set Resource: Sheet, Operation: Get Row(s)
3. Select the document in the Document field
4. Open the Sheet field — it accepts a single sheet only, with no "all sheets" choice
5. Execute the node

Rows are returned for the one selected sheet only. To get all three tabs you must add three Google Sheets nodes (or a loop driven by a hard-coded list of sheet names), and revisit the workflow whenever the tabs change.

### Expected behavior

An option on Get Row(s) to read every sheet in the document in one execution, returning the union of their rows and indicating which sheet each row came from, so downstream nodes can filter or group by it.

Adding or renaming a tab should not require editing the workflow.

### Debug Info

Reported against the current master branch of the n8n repository (n8n 2.34.0), reviewed in source rather than a running instance.

### Operating System

Windows 11

### n8n Version

2.34.0

### Node.js Version

22.18.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

ramygamal231:
Related to #35877

## Summary
This PR introduces the ability to read all sheets within a Google Spreadsheet using the Google Sheets node's **Get Row(s)** operation. Previously, users had to specify a single sheet or iterate through them manually.

## Key Changes

- **New Parameter**  
  Added `sheetSelectionMode` allowing users to choose between:
  - **Single Sheet** (current behavior)
  - **All Sheets**

- **Dynamic Fetching**  
  In "All Sheets" mode, the node first fetches the list of all sheet names from the spreadsheet using the Google Sheets API.

- **Improved Output**  
  Each row returned in "All Sheets" mode includes a `_sheetName` field to indicate its source tab.

- **UI Logic**  
  The "Sheet" selector is automatically hidden when "All Sheets" mode is selected to simplify the user interface.

- **Filters**  
  Filters are intentionally disabled in "All Sheets" mode for this initial version to ensure stability (users can use the **Filter** node downstream).

## Related Linear Tickets, Github Issues, and Community Forum Posts
Resolves the need for bulk reading multiple spreadsheet tabs in one go.

## Review / Merge Checklist

- [ ] Docs updated (or follow-up ticket created)  
- [x] Tests included  
- [ ] PR labeled  

## How to Test

1. Select **Resource:** Sheet and **Operation:** Get Row(s)  
2. Change **Sheet Selection Mode** to **All Sheets**  
3. Run the node  
4. Verify that rows from all worksheet tabs are correctly flattened and returned with the `_sheetName` property


ramygamal231:
Thanks for the pointer — I've hit a loop here and could use a steer from a maintainer.

Following the request above, I opened #35877 describing the problem this PR solves. It was closed automatically as `closed:enhancement/feature`, with a note that GitHub issues are used exclusively for bugs and that feature requests belong on the forum.

So the two automations are asking for opposite things: this PR can't leave triage without a linked GitHub issue, and a GitHub issue describing a feature gets closed on sight. I don't think there's a way for me to satisfy both.

I've posted the request on the forum as the issue bot suggested: [FORUM_LINK](https://community.n8n.io/t/google-sheets-node-read-all-sheets-in-a-document-with-one-get-row-s-node/306706)

Happy to resolve this whichever way suits your process — reopening #35877, pointing `Closes` at something else, or waiving the requirement for this PR. Just let me know which you'd prefer.

For context on where the PR stands, the test coverage requested earlier has been pushed:

- unit tests for fetching the sheet list, looping through sheets, injecting `_sheetName`, and per-sheet fetch errors
- a router test covering the routing decision, including that other operations still resolve their sheet
- a `NodeTestHarness` workflow test running the node through the execution engine against nock-mocked Google endpoints

Writing them also surfaced four issues in my original implementation, fixed in the same push: a crash when the spreadsheet response carried no `sheets`, chart-only sheets being read (the values API rejects their range, failing the whole node), a column genuinely named `_sheetName` being overwritten, and the Options collection losing its `untilSheetSelected` guard. With Continue On Fail set, one unreadable sheet no longer discards rows already read.

The branch is also rebased onto current master.

ramygamal231:
Hi @Joffcom, could this be reopened? I think it got auto-closed by mistake.

The PR links to #35877 with a Closes line, which is what got it through triage after the bot asked for a linked issue a few days back. But #35877 was itself auto-closed as a feature request, since the tracker is bug-only, and it looks like the automation then closed this PR because its linked issue was "resolved." It wasn't really resolved, just moved off the issue tracker as out of scope.

You'd already assigned @RomanDavydchuk to review, so the change is still live and ready from my end. 

To stop the two automations closing each other, maybe I could change "Closes #35877" for a plain reference to #35877 like "Related to #35877".

Joffcom:
Hey @ramygamal231 

I will run it past the team, our guidelines were recently tweaked to be more explicit which meant some PRs have been closed which possibly should have been closed sooner.

I will let you know in the morning what comes back.

ramygamal231:
Ok, thanks

RomanDavydchuk:
Hey @ramygamal231, thanks for letting us know, I'll reopen the PR and take a look some time later

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/Sheet.resource.ts
```diff
@@ -202,6 +229,7 @@ export const descriptions: INodeProperties[] = [
 			show: {
 				resource: ['sheet'],
 				operation: ['append', 'appendOrUpdate', 'clear', 'delete', 'read', 'remove', 'update'],
+				sheetSelectionMode: ['single'],
```
RomanDavydchuk: This makes it so that the `sheetName` is shown only for `read` operation with `sheetSelectionMode` `single`. In other words, it's not shown for other operations where it's needed

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -101,77 +157,95 @@ export const description: SheetProperties = [
...
+			}
+		} catch (error) {
+			const message = `Failed to read rows from sheet "${currentSheetName}": ${
+				error instanceof Error ? error.message : String(error)
+			}`;
+
+			// Without this a single unreadable sheet would discard the rows already read
+			if (this.continueOnFail()) {
+				returnData.push({
+					json: { [SHEET_NAME_FIELD]: currentSheetName, error: message },
```
RomanDavydchuk: This top-level property causes the engine to route the output item into the wrong branch, when using "Continue (using error output)", see here where I simulated an error during `readSheet` call:

Image

The reason is that the engine considers an item as an "error item" only if it has these top-level properties: `error`, `message`, `details`. So if you want to pass the sheet name, I suggest doing it under the `details` object

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -101,77 +157,95 @@ export const description: SheetProperties = [
...
+		},
 	},
 ];
 
+/**
+ * Reads every sheet of the spreadsheet in turn, tagging each row with its source
+ * sheet. Filters are not offered in this mode, so one pass over the spreadsheet
+ * is enough regardless of how many input items there are.
+ */
+async function readAllSheets(
```
RomanDavydchuk, line 238: Looks like this function always runs once, regardless of the number of input items. This is not how most nodes work and also it's not how the read operation but for the single sheet works.

See, how for two input items I get 4 output items, even though there are only 2 rows in the sheet:

Image

When I run the read operation for all sheets, each row is returned only once (in this example I have two sheets with two rows each):

Image

packages/nodes-base/nodes/Google/Sheet/test/v2/node/read.test.ts
```diff
@@ -78,4 +79,246 @@ describe('Google Sheet - Read', () => {
...
+				pairedItem: { item: 0 },
+			},
+			{
+				json: { _sheetName: 'Q2', name: 'Alan', amount: '30', row_number: 3 },
+				pairedItem: { item: 0 },
+			},
+		]);
+	});
+
+	test('should read the spreadsheet once regardless of how many input items there are', async () => {
```
RomanDavydchuk: The expectation is incorrect here (see one of my comments above)

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -1,11 +1,15 @@
...
 
 import { dataLocationOnSheet, outputFormatting } from './commonDescription';
 import type { GoogleSheet } from '../../helpers/GoogleSheet';
 import type { SheetProperties } from '../../helpers/GoogleSheets.types';
-import { untilSheetSelected } from '../../helpers/GoogleSheets.utils';
+import { getGridSheetNames, untilSheetSelected } from '../../helpers/GoogleSheets.utils';
 import { readSheet } from '../utils/readOperation';
 
+/** Column added to every row in "All Sheets" mode, naming the sheet it came from */
+const SHEET_NAME_FIELD = '_sheetName';
```
RomanDavydchuk: I think we should change this name. I assume you chose `_sheetName` to minimize the chance of collisions with column names, but we already have `row_number` for the row number, so let's stick to the convention
```suggestion
const SHEET_NAME_FIELD = 'sheet_name';
```

RomanDavydchuk (changes requested):
Thanks for your contribution! I looked through the code and did some manual testing. The new mode for the read operation seems to be working well, beside some edge cases I covered in the code comments. Also I found an issue with some node parameters being broken for other operations

Besides that, I'm not too sure if it makes sense to leave out the filters for the new read mode. In my mind that would be quite easy: just need to merge together the column names from all sheets and provide them all. If there is a hidden reason as to why it's hard or if I'm missing something - please let me know 🙏

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
ramygamal231: ok noted. I renamed to sheet_name to match row_number.

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/Sheet.resource.ts
ramygamal231: Ok noted. The show rule lists all operations again with no sheetSelectionMode condition, so the field shows everywhere it did before. It's now hidden only for read + All Sheets (via a hide rule), where there's no single sheet to pick.

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
ramygamal231: Fixed. The error now goes on the item's top-level error property, so the engine routes it to the error output, and the sheet name stays in json.

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
ramygamal231: Ok noted. I changed it to read every sheet once per input item, matching single-sheet read.

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/Sheet.resource.ts
```diff
@@ -203,6 +230,12 @@ export const descriptions: INodeProperties[] = [
 				resource: ['sheet'],
 				operation: ['append', 'appendOrUpdate', 'clear', 'delete', 'read', 'remove', 'update'],
 			},
+			// Only the read operation's "All Sheets" mode has no single sheet to pick;
+			// every other operation still needs this field
+			hide: {
+				operation: ['read'],
+				sheetSelectionMode: ['all'],
```
RomanDavydchuk, line 237: This causes the RLC to be hidden for `operation: 'read'` regardless of the `sheetSelectionMode`. Here you can see I picked "Single", but it's still not shown: 

Image

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/Sheet.resource.ts
```diff
@@ -135,6 +135,33 @@ export const descriptions: INodeProperties[] = [
...
+				description:
+					'Read rows from every sheet in the document, adding a "sheet_name" field to each row',
+			},
+		],
+		description:
+			'Whether to read from one sheet or from every sheet in the document. "All Sheets" makes one request per sheet, so it can use up a lot of API quota on large documents.',
+		displayOptions: {
+			show: {
+				resource: ['sheet'],
+				operation: ['read'],
```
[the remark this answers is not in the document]
RomanDavydchuk: Valid comment, pointed this out too

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -101,77 +174,139 @@ export const description: SheetProperties = [
...
+					});
+				}
+			} catch (error) {
+				const message = `Failed to read rows from sheet "${currentSheetName}": ${
+					error instanceof Error ? error.message : String(error)
+				}`;
+
+				// Without this a single unreadable sheet would discard the rows already read.
+				// The error goes on the top-level `error` property so the engine routes it
+				// to the error output; the sheet name stays readable in `json`.
```
RomanDavydchuk: nit: unnecessary comment

packages/nodes-base/nodes/Google/Sheet/v2/methods/loadOptions.ts
```diff
@@ -116,3 +116,55 @@ export async function getSheetHeaderRowAndSkipEmpty(
...
+
+	if (responseData === undefined) {
+		throw new NodeOperationError(this.getNode(), 'No data got returned');
+	}
+
+	const seen = new Set<string>();
+	const returnData: INodePropertyOptions[] = [];
+
+	for (const entry of responseData.sheets ?? []) {
+		if (entry.properties?.sheetType !== 'GRID' || !entry.properties?.title) {
```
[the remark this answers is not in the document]
RomanDavydchuk: Valid comment by the bot, should address it

packages/nodes-base/nodes/Google/Sheet/v2/methods/loadOptions.ts
```diff
@@ -116,3 +116,55 @@ export async function getSheetHeaderRowAndSkipEmpty(
...
+			continue;
+		}
+
+		const title = entry.properties.title;
+		const sheetData = await sheet.getData(`${title}!1:1`, 'FORMATTED_VALUE');
```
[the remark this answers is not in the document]
RomanDavydchuk: Doesn't seem to be the case

packages/nodes-base/nodes/Google/Sheet/test/v2/node/read.test.ts
```diff
@@ -78,4 +79,300 @@ describe('Google Sheet - Read', () => {
...
+		sheet.getData = vi.fn(async (range: string) => {
+			if (range === 'Q1') throw new Error('Quota exceeded');
+			return rowsBySheet[range];
+		});
+
+		const result = await execute.call(mockExecuteFunctions as IExecuteFunctions, sheet, '');
+
+		// The failing sheet becomes an error item: the sheet name stays in json, and
+		// the error sits on the top-level `error` property so the engine can route it
+		// to the error output instead of the success output
```
RomanDavydchuk: Same unnecessary comment

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -1,11 +1,20 @@
...
 
 import { dataLocationOnSheet, outputFormatting } from './commonDescription';
 import type { GoogleSheet } from '../../helpers/GoogleSheet';
 import type { SheetProperties } from '../../helpers/GoogleSheets.types';
-import { untilSheetSelected } from '../../helpers/GoogleSheets.utils';
+import { getGridSheetNames, untilSheetSelected } from '../../helpers/GoogleSheets.utils';
 import { readSheet } from '../utils/readOperation';
 
+/** Column added to every row in "All Sheets" mode, naming the sheet it came from */
+const SHEET_NAME_FIELD = 'sheet_name';
```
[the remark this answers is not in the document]
RomanDavydchuk: Not sure where it's advertised as `_sheetName`

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -101,77 +174,139 @@ export const description: SheetProperties = [
...
+				const message = `Failed to read rows from sheet "${currentSheetName}": ${
+					error instanceof Error ? error.message : String(error)
+				}`;
+
+				// Without this a single unreadable sheet would discard the rows already read.
+				// The error goes on the top-level `error` property so the engine routes it
+				// to the error output; the sheet name stays readable in `json`.
+				if (this.continueOnFail()) {
+					returnData.push({
+						json: { [SHEET_NAME_FIELD]: currentSheetName },
```
RomanDavydchuk: Even though the items are now correctly routed to the error branch, this doesn't appear in the output item:

Image

This is because if `item.error` is present the engine discards `item.json`. If you want to keep `sheet_name` field, I would suggest you follow my advice and put it under `details` key (and use `item.json.error` instead of `item.error`). But make sure to merge in `master` before that, looks like your branch doesn't have this code yet

RomanDavydchuk (changes requested):
Thanks for addressing my comments 🙏 I did find some more issues, the most important of which, in my opinion, is the fact that the "Sheet Name" selector is not shown for the "Single" sheet selection mode. The bot also pointed out some valid issues, although not all of it's comments are true 😄

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
```diff
@@ -27,42 +36,104 @@ const combineFiltersOptions: INodeProperties = {
...
+export const readFilter: INodeProperties = createReadFilter({
+	loadOptionsDependsOn: ['sheetName.value'],
+	loadOptionsMethod: 'getSheetHeaderRowWithGeneratedColumnNames',
+});
+
+// "All Sheets" has no single sheet to read headers from, so the column list is
+// the union of every sheet's header row
+const readFilterAllSheets: INodeProperties = createReadFilter({
+	loadOptionsDependsOn: ['documentId.value'],
+	loadOptionsMethod: 'getSheetHeaderRowWithGeneratedColumnNamesForAllSheets',
```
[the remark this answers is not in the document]
RomanDavydchuk: Not sure what this means

ramygamal231:
Thanks for the thorough review 🙏. I merged latest master first, then addressed the issues raised by the bot and the notes you provided. For example:

- Sheet selector in Single mode is fixed. It now shows for single-sheet reads (and all other operations) and only hides for All Sheets.

- Error output: The sheet name is now kept on error items (under details), so it shows up on the error branch.

- sheet_name is kept as-is per your call, matching row_number.

- Removed the comments you flagged.

Let me know if there is anything else.

ramygamal231:
Reopening 🙏

The pull request got auto-closed again by the linked-issue automation (it fires because the PR is linked to #35877, which is already marked resolved). I've changed the description from Closes #35877 to Related to #35877 so it won't keep getting auto-closed.

ramygamal231:
@RomanDavydchuk Could you reopen it? I've switched the Closes #35877 link to Related to #35877, so I hope the automation doesn't auto-close it again. Thanks 🙏

packages/nodes-base/nodes/Google/Sheet/v2/actions/sheet/read.operation.ts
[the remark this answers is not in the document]
RomanDavydchuk: Turned out to be not relevant due to `row_number` being prepended

RomanDavydchuk (changes requested):
Looks much better, thanks for addressing my comments 🙏 

Currently the CI is not passing due to typecheck, so you'd need to fix that first:
```
Error: nodes/Google/Sheet/test/v2/actions/sheet/Sheet.resource.test.ts(10,20): error TS2345: Argument of type 'Record<string, unknown>' is not assignable to parameter of type 'ICredentialDataDecryptedObject | INodeParameters'
```

Also, during my manual testing I noticed this: "Return only First Matching Row" settings is applied per sheet instead of globally.

I have an example, here's `Sheet1`:
Image

`Sheet2`:
Image

If I run the node with filter `col_2=42`, it'll match 3 rows in total (the first one in `Sheet1` and both in `Sheet2`). If I turn on "Return only First Matching Row" - it returns *two* rows: the first one from `Sheet1` and the first one from `Sheet2`:
Image

But IMO it would make more sense if it returned only the first match across all sheets. That would fit the use-case better, what do you think?

PR #27278, opened 2026-03-19
fix(core): Fix unhandled rejection in task broker on runner disconnect

ivov:
## Summary

When a runner disconnects mid-task for any reason, `TaskBroker.deregisterRunner` runs and deletes both the runner and its tasks from the broker's maps. But the requester may still be processing an in-flight RPC call from that runner that was disconnected and deregistered. So when the requester finishes and sends back the response, `TaskBroker.handleRequesterRpcResponse` calls `TaskBroker.getRunnerOrFailTask` for a task that no longer exists, so it throws an error that becomes an unhandled promise rejection.

This fix skips the `getRunnerOrFailTask` call when we already know the task is gone so we know there will be no runner for it, and `TaskRequester.sendMessage` catches any remaining unexpected errors. Every path that removes a task from the map notifies the requester first, so the requester already knows the runner is gone and the RPC response has nowhere to go.

https://linear.app/n8n/issue/CAT-2555

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/task-runners/task-broker/__tests__/task-broker.service.test.ts
+++ b/packages/cli/src/task-runners/task-broker/__tests__/task-broker.service.test.ts
@@ -714,6 +714,28 @@ describe('TaskBroker', () => {
 				nodeTypes,
 			});
 		});
+
+		it('should discard `requester:rpcresponse` for an already-cleaned-up task', async () => {
+			await expect(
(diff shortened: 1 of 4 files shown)


tomi (approved):
Thanks for fixing 👌

PR #27283, opened 2026-03-19
fix(core): Fix MCP OAuth discovery URL construction and grant type selection

jjnyn:
## Summary

Fixes two issues in the MCP OAuth 2.0 credential flow that caused failures with
authorization servers that follow the spec strictly (e.g. Common Room):

**1. `selectGrantTypeAndAuthenticationMethod` throws when `refresh_token` is absent**

The grant type selector required both `authorization_code` AND `refresh_token` to
be present in the server's `grant_types_supported` metadata before entering the
authorization code branch. The OAuth 2.0 spec (and many real-world providers) only
advertise `authorization_code` — `refresh_token` is a token type, not a grant type
that servers are required to list. This caused n8n to fall through all branches and
throw "No supported grant type and authentication method found" even when the server
fully supports the authorization code flow.

**2. Discovery URL construction produces double `/.well-known/` paths**

When the input `authorizationServerUrl` already has a `/.well-known/` path (e.g.
the server URL itself is a discovery document URL, or the fallback path from
`discoverProtectedResourceMetadata` resolves to one), the code was appending
`/.well-known/openid-configuration` onto it, producing a broken URL like:
`https://example.com/.well-known/openid-configuration/.well-known/openid-configuration`

The fix detects when `pathComponent` already starts with `/.well-known` and skips
path-insertion variants in that case, falling back directly to root-level discovery.

## Review / Merge checklist
- [x] Tests included. 


yehorkardash (dismissed):

jjnyn:
> @jjnyn, thanks for your contribution! Your PR looks is good to go, however there are some formatting errors. Could you resolve them, please?

Hi @yehorkardash I've fixed the formatting error. Would you mind taking a look again?

TimBakerCircle:
@jjnyn  @yehorkardash  wanted to advise I tested this code on my local deployment and it solved the mcp oauth problem i was facing.  Thanks.

yehorkardash (approved):

PR #25813, opened 2026-02-16
fix(core): Use pg driver native statement timeout instead of command line

tomi:
## Summary

Bump @n8n/typeorm and pass in the statement timeout via driver config instead of command line options

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/CAT-2350/use-driver-native-statement-timeout-instead-of-command-line-options

## Review / Merge checklist

- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #27295, opened 2026-03-19
fix(core): Apply Postgres statement_timeout via SET, not startup param

### LINKED ISSUE #25705: Unsupported startup parameter in options: statement_timeout
### Bug Description

Hi,

i use n8n with external PostgreSQL database. I just upgraded from 2.6.3 to 2.7.4 and now i get the following error during startup: 

```bash
Initializing n8n process                                                                                                                                                                                                                                      
There was an error initializing DB                                                                                                                                                                                                                            
unsupported startup parameter in options: statement_timeout  
```

Looks like it is related to that PR: https://github.com/n8n-io/n8n/pull/25041

### To Reproduce

start n8n 2.7.4 with postgres database

### Expected behavior

should work

### Debug Info

-

### Operating System

docker

### n8n Version

2.7.4

### Node.js Version

docker

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

ivov:
## Summary

Fixes Postgres `statement_timeout` breaking startup on AWS RDS Proxy and other providers that reject startup parameters The previous fix #25813 moved from `-c statement_timeout=n` to typeorm's native `statementTimeout` property, but both paths still end up in being passed via the `pg` driver.

- https://linear.app/n8n/issue/CAT-2645
- Closes #25705

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/connection/__tests__/db-connection-options.test.ts
+++ b/packages/@n8n/db/src/connection/__tests__/db-connection-options.test.ts
@@ -112,7 +112,6 @@ describe('DbConnectionOptions', () => {
 					poolSize: 2,
 					migrations: postgresMigrations,
 					connectTimeoutMS: 20000,
-					statementTimeout: 300_000,
 					ssl: false,
 					extra: {
 						idleTimeoutMillis: 30000,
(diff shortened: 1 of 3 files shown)


tomi (commented):
Since we have the config option on the typeorm library, it would make more sense to let the driver handle that instead of doing it in the app side

PR #27296, opened 2026-03-19
fix(instance-ai): persist thread delete and rename to backend

aalises:
## Summary
Kapture 2026-03-19 at 18 07 46

- Thread deletion and renaming in Instance AI were purely client-side (in-memory Pinia state only), so changes were lost on page refresh
- Added `DELETE` and `PATCH /instance-ai/threads/:threadId` backend endpoints
- Wired frontend store actions to call the API before updating local state

## Linear ticket
https://linear.app/n8n/issue/AI-2240/instance-ai-thread-delete-and-rename-not-persisted-to-backend

## Test plan
- [ ] Delete a thread → refresh page → thread should stay deleted
- [ ] Rename a thread → refresh page → new title should persist
- [ ] Delete the last/active thread → a new empty thread should be created
- [ ] Deleting a thread that hasn't been persisted yet (no messages sent) should not call the API

### DIFF
--- a/packages/@n8n/api-types/src/dto/instance-ai/instance-ai-rename-thread-request.dto.ts
+++ b/packages/@n8n/api-types/src/dto/instance-ai/instance-ai-rename-thread-request.dto.ts
@@ -3,5 +3,5 @@ import { z } from 'zod';
 import { Z } from '../../zod-class';
 
 export class InstanceAiRenameThreadRequestDto extends Z.class({
-	title: z.string().trim().min(1),
+	title: z.string().trim().min(1).max(255),
 }) {}
--- a/packages/@n8n/api-types/src/dto/instance-ai/instance-ai-rename-thread-request.dto.ts
+++ b/packages/@n8n/api-types/src/dto/instance-ai/instance-ai-rename-thread-request.dto.ts
@@ -1,0 +1,7 @@
+import { z } from 'zod';
+
+import { Z } from '../../zod-class';
+
+export class InstanceAiRenameThreadRequestDto extends Z.class({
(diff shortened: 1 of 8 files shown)


packages/@n8n/api-types/src/dto/instance-ai/instance-ai-rename-thread-request.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod';
+
+import { Z } from '../../zod-class';
+
+export class InstanceAiRenameThreadRequestDto extends Z.class({
+	title: z.string().trim().min(1),
```
Cadiac: Lets also enforce some sane max length for these given the db column is text

packages/cli/src/modules/instance-ai/instance-ai-memory.service.ts
```diff
@@ -223,6 +223,29 @@ export class InstanceAiMemoryService {
...
+
+	async renameThread(
+		_userId: string,
+		threadId: string,
+		title: string,
+	): Promise<InstanceAiThreadInfo> {
+		const memory = this.createMemoryInstance();
+		const thread = await memory.getThreadById({ threadId });
+		if (!thread) {
+			throw new Error(`Thread ${threadId} not found`);
```
Cadiac: Prefer `NotFoundError` to not have these pollute our sentry error logs (and to make this return 404 automatically)

aalises: both fixed

Cadiac (approved):

PR #27297, opened 2026-03-19
fix: Preserve `structuredContent` in MCP nodes to prevent AI Agent reasoning loops

### LINKED ISSUE #26963: MCP Client Tool drops structuredContent from MCP tool results, causing AI Agent tool loops
### Bug Description

When using AI Agent + OpenAI Chat Model + MCP Client Tool, MCP tool results appear to lose structuredContent before they are passed back into the agent loop.

The MCP server returns a valid tools/call response containing both:
- content
- structuredContent

But inside n8n, the tool output exposed to the agent only contains a text-like response array, where the useful payload is embedded as a JSON string inside response[0].text.

Because the agent no longer receives the structured tool result, it repeatedly calls the same MCP tool instead of producing a final answer.

### To Reproduce

1. Connect an AI Agent node to:
  -   an OpenAI Chat Model node with Use Responses API enabled
  -   an MCP Client Tool node exposing a tool named get-site-info
4. Ask the agent to call the tool and return fields from the tool result.
5. Observe that:
  - the MCP tool is called successfully
  - the agent calls it repeatedly
  - the model emits no final text output

### Expected behavior

The AI Agent should receive the MCP tool result including structuredContent, use that structured object, and return a final response such as:

```
{
  "id": "69b347da1c8db5d19d1fa1d6",
  "workspaceId": "69b347971c8db5d19d1f9f24",
  "displayName": "Mark's Fabulous Thing",
  "shortName": "this is cool"
}
```

### Debug Info

The MCP server returns valid structuredContent on the wire, but n8n appears to flatten the tool result into something like:

```
{
  "response": [
    {
      "type": "text",
      "text": "{\"id\":\"69b347da1c8db5d19d1fa1d6\",\"workspaceId\":\"69b347971c8db5d19d1f9f24\",\"displayName\":\"Mark's Fabulous Site\",\"shortName\":\"marks-fabulous-site\", ... }"
    }
  ]
}
```

The agent then repeatedly calls the same MCP tool instead of finalizing.

The final model node output is empty:
```
[
  {
    "response": {
      "generations": [
        [
          {
            "text": ""
          }
        ]
      ]
    },
    "tokenUsageEstimate": {
      "completionTokens": 0,
      "promptTokens": 843,
      "totalTokens": 843
    }
  }
]
```

**Raw MCP response from direct curl**

Calling the MCP server directly shows that structuredContent is present and correct:
```
curl -k -X POST "https://wfdev.io:8443/api/mcp" \
  -H "Authorization: Bearer $SERVICE_SECRET" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -H "X-User-Id: 69b347961c8db5d19d1f9f01" \
  -H "X-Site-Id: 69b347da1c8db5d19d1fa1d6" \
  -d '{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "tools/call",
    "params": {
      "name": "get-site-info",
      "arguments": {
        "siteId": "69b347da1c8db5d19d1fa1d6"
      }
    }
  }' | jq
```

Response:
```
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{\"id\":\"69b347da1c8db5d19d1fa1d6\",\"workspaceId\":\"69b347971c8db5d19d1f9f24\",\"displayName\":\"Mark's Fabulous Thing\",\"shortName\":\"marks-fabulous-thing\",\"previewUrl\":null,\"timeZone\":\"America/Los_Angeles\",\"createdOn\":\"2026-03-12T23:10:18.699Z\",\"lastUpdated\":\"2026-03-12T23:10:30.623Z\",\"lastPublished\":null,\"parentFolderId\":null,\"customDomains\":[],\"locales\":null,\"dataCollectionEnabled\":false,\"dataCollectionType\":\"always\",\"googleTagIds\":[]}"
      }
    ],
    "structuredContent": {
      "id": "69b347da1c8db5d19d1fa1d6",
      "workspaceId": "69b347971c8db5d19d1f9f24",
      "displayName": "Mark's Fabulous Thing",
      "shortName": "marks-fabulous-thing",
      "previewUrl": null,
      "timeZone": "America/Los_Angeles",
      "createdOn": "2026-03-12T23:10:18.699Z",
      "lastUpdated": "2026-03-12T23:10:30.623Z",
      "lastPublished": null,
      "parentFolderId": null,
      "customDomains": [],
      "locales": null,
      "dataCollectionEnabled": false,
      "dataCollectionType": "always",
      "googleTagIds": []
    }
  },
  "jsonrpc": "2.0",
  "id": 2
}
```

**Why this seems like an n8n bug**
The external MCP server is returning valid structured tool output.

The issue seems to happen inside the MCP Client Tool / agent integration path, where structuredContent is not preserved and only text content survives. That makes the AI Agent treat the tool result as opaque text instead of structured data, which likely causes the repeated tool call loop.

**Workaround**
Current workaround is to avoid giving the raw MCP tool directly to the agent, and instead wrap it in a workflow that parses the JSON string from the text block and returns a flat JSON object. This works around the issue, but it should not be necessary when the MCP server already returns structuredContent.

**Question**
Can MCP Client Tool preserve and pass structuredContent from MCP tools/call results into the AI Agent tool loop?

If helpful, I can also rewrite this into a shorter GitHub-issue version with less context and tighter formatting.

### Operating System

alpine:3.19 aws k8

### n8n Version

2.3.5

### Node.js Version

20.10.x

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

sinehypernova-0718:
# Summary
This PR fixes a data-loss issue in the MCP Client where `structuredContent` from tool calls was being discarded, forcing AI Agents into infinite retry loops.

---

## The Problem
The MCP specification allows tools to return both `content` (text/images) and `structuredContent` (machine-readable JSON). 
Previously, n8n's implementation:
1. **In the AI Agent Tool:** Explicitly ignored `structuredContent`, returning only the `content` array. This forced LLMs (especially OpenAI via the Responses API) to try and parse stringified JSON from text blocks. When the model couldn't find a clean JSON object matching its tool schema, it assumed the call failed and re-invoked the tool indefinitely.
2. **In the Standalone Node:** Only exposed the `content` array, making it difficult for users to access specific data fields without using a `JSON.parse()` expression node.

## The Fix
Implemented a high-performance, type-safe bridge for structured data:
- **Agent Tool (`utils.ts`):** Added priority logic ($O(1)$) to return `structuredContent` directly if it exists. This provides the LLM with the exact JSON object it requires to fulfill its internal schema validation.
- **Standalone Node (`McpClient.node.ts`):** Implemented a merge strategy ($O(K)$) that flattens `structuredContent` into the node's top-level JSON output while preserving the `content` array for human-readable logs.

## Technical Hardening
- **Type Safety:** Added a 4-point guard (`undefined`, `null`, `object`, and `!Array`) to ensure malformed MCP responses cannot trigger runtime TypeErrors.
- **Key Protection:** In the standalone node, the logic ensures that the authoritative MCP `content` array is never overwritten by keys inside the structured object.
- **Performance:** Used shallow copies and reference returns to maintain optimal execution time and memory footprint.

## Tests

```
Test Suites: 18 passed, 18 total
Tests:       331 passed, 331 total
Snapshots:   0 total
Time:        7.076 s
```

---

# Related Linear tickets, Github issues, and Community forum posts

Fixes: #26963
Linear Reference: [GHC-7250](https://linear.app/n8n/issue/GHC-7250)

---

# Review / Merge checklist

- [x] Tests included. 
- [x] Linter/Typecheck: `pnpm typecheck` passed with Exit Code 0.

---

## Verification Table

| Input Scenario | AI Agent Result | Standalone Node JSON |
| :--- | :--- | :--- |
| **Only `content`** | `[{type: "text"}]` | `{ "content": [...] }` |
| **Only `structuredContent`** | `{ "id": 123 }` | `{ "id": 123 }` |
| **Mixed (Both)** | `{ "id": 123 }` (Priority) | `{ "id": 123, "content": [...] }` |
| **Malformed (String)** | Falls back to `content` | Falls back to `content` |


packages/@n8n/nodes-langchain/nodes/mcp/McpClient/McpClient.node.ts
```diff
@@ -289,10 +289,26 @@ export class McpClient implements INodeType {
 					content.push(contentItem as IDataObject);
 				}
 
+				const outputJson: IDataObject = {};
+				if (
+					result.structuredContent &&
+					typeof result.structuredContent === 'object' &&
```
BerniWittmann: This guard is repeated across both files, maybe we can extract that as a small helper: 

```
isStructuredContent(value: unknown): value is IDataObject
```

packages/@n8n/nodes-langchain/nodes/mcp/McpClient/McpClient.node.ts
```diff
@@ -289,10 +289,26 @@ export class McpClient implements INodeType {
...
+
+				if (content.length > 0) {
+					outputJson.content = content;
+				}
+
 				returnData.push({
-					json: {
-						content: content.length > 0 ? content : undefined,
-					},
+					json: outputJson,
```
BerniWittmann: I think i would try to keep the output-shape stable for existing users. Instead i would propose: 

```
json: {
  content: content.length > 0 ? content : undefined,
  structuredContent: (isStructuredContent(result.structuredContent) ? result.structuredContent : undefined),
}
```

(my example is using the extracted helper from the other comment)

packages/@n8n/nodes-langchain/nodes/mcp/McpClient/McpClient.node.ts
```diff
@@ -289,10 +289,26 @@ export class McpClient implements INodeType {
...
 				}
 
+				const outputJson: IDataObject = {};
+				if (
+					result.structuredContent &&
+					typeof result.structuredContent === 'object' &&
+					!Array.isArray(result.structuredContent)
+				) {
+					for (const key of Object.keys(result.structuredContent)) {
+						if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
```
BerniWittmann: a cleaner alternative could be to use object-spread, so we dont have to maintain this "deny-list"

BerniWittmann (changes requested):
Thanks for your contribution 🙌 

I think `McpClientTool.node.ts` in its `execute`path is still dropping structured Content, so it should have the same preservation there

sinehypernova-0718:
Hey! @BerniWittmann

Thanks for the review. I’ll address all the requested changes and push the updates as soon as possible.

Thank you!

packages/@n8n/nodes-langchain/nodes/mcp/McpClient/McpClient.node.ts
sinehypernova-0718: Quick question, do you prefer I add the `isStructuredContent` helper once (e.g. in McpClientTool's utils and cross-import from McpClient), or keep it localized in each node's own utils? Either works, just want to match what you'd prefer here.

BerniWittmann: I think there is a `packages/@n8n/nodes-langchain/nodes/mcp/shared/utils.ts` file where it could fit nicely.

sinehypernova-0718: Thanks! That actually makes sense, I will use the shared `utils.ts` for the helper.

packages/@n8n/nodes-langchain/nodes/mcp/McpClient/McpClient.node.ts
sinehypernova-0718: For the deny-list block in `McpClient.node.ts`, I'm planning to replace it with:

```ts
const outputJson: IDataObject = {};
if (isStructuredContent(result.structuredContent)) {
    outputJson.structuredContent = { ...result.structuredContent };
}
if (content.length > 0) {
    outputJson.content = content;
}
```

So structuredContent stays as its own property (stable shape) and the spread avoids needing the deny-list. That what you had in mind?


Just confirming if you like this approach before committing

BerniWittmann: sounds good :+1:

sinehypernova-0718: Perfect! thanks for confirming. I’ll push the changes shortly, helper in the shared `utils.ts`, object spread replacing the deny-list, and `structuredContent` kept as its own property for stable output shape. Appreciate the review! 🙏

Thank you!

sinehypernova-0718:
@BerniWittmann all feedback addressed:

`isStructuredContent` helper added to shared `utils.ts`, no more repeated guard

Deny-list replaced with object spread `{ ...structuredContent }` in `McpClient.node.ts`

Output shape kept stable, `structuredContent` as own property, not flattened

`structuredContent` now preserved in `McpClientTool.node.ts` execute path so the AI agent receives it

Ready for re-review whenever you get a chance. 

Let me know any further changes are required

Thanks for your patience! 🙏

sinehypernova-0718:
@BerniWittmann all the main changes are pushed. I'm updating the tests now to match the new `structuredContent` output shape, will commit that shortly.

sinehypernova-0718:
@BerniWittmann  all changes are in place shared helper, object spread, stable output shape, structuredContent preserved in both nodes, and tests are updated.

Let me know if anything else needs a tweak. 

Thanks! 🙏

BerniWittmann:
@sinehypernova-0718 I think there are some merge conflicts with master now that need resolving? Will then do a review towards the afternoon

sinehypernova-0718:
@BerniWittmann sounds good, I’ll resolve the merge conflicts and push the fix. 
Thanks for the heads up, looking forward to the review! 🙏

sinehypernova-0718:
@BerniWittmann merge conflicts resolved, all feedback addressed. ✅

Ready for final review! 🙏

packages/@n8n/nodes-langchain/nodes/mcp/McpClientTool/McpClientTool.node.ts
```diff
@@ -440,55 +411,53 @@ export class McpClientTool implements INodeType {
...
+								structuredContent: result.structuredContent,
+							}),
+						},
+						pairedItem: {
+							item: itemIndex,
+						},
+					});
 				}
-			} finally {
-				await client.close();
```
BerniWittmann, line 491: why was this removed?

packages/@n8n/nodes-langchain/nodes/mcp/McpClientTool/McpClientTool.node.ts
```diff
@@ -426,11 +402,6 @@ export class McpClientTool implements INodeType {
 		const returnData: INodeExecutionData[] = [];
 
 		for (let itemIndex = 0; itemIndex < items.length; itemIndex++) {
-			const signal = this.getExecutionCancelSignal();
-			if (signal?.aborted) {
-				throw new NodeOperationError(node, 'Execution was cancelled', { itemIndex });
```
BerniWittmann, line 431: why was this removed?

packages/@n8n/nodes-langchain/nodes/mcp/McpClientTool/McpClientTool.node.ts
```diff
@@ -385,39 +375,25 @@ export class McpClientTool implements INodeType {
...
-				);
-			});
-
-			this.logger.debug(`McpClientTool: Connected to MCP Server with ${tools.length} tools`);
-
-			const toolkit = new StructuredToolkit(tools);
-
-			return { response: toolkit, closeFunction: async () => await client.close() };
-		} catch (e) {
-			await client.close();
```
BerniWittmann, line 418: why was this removed?

packages/@n8n/nodes-langchain/nodes/mcp/McpClientTool/McpClientTool.node.ts
```diff
@@ -360,11 +356,6 @@ export class McpClientTool implements INodeType {
 			throw error;
 		};
 
-		const signal = this.getExecutionCancelSignal();
-		if (signal?.aborted) {
-			return setError(new NodeOperationError(node, 'Execution was cancelled', { itemIndex }));
```
BerniWittmann, line 365: why was this remvoed?

packages/@n8n/nodes-langchain/nodes/mcp/McpClientTool/McpClientTool.node.ts
```diff
@@ -100,20 +101,15 @@ async function connectAndGetTools(
...
-		const mcpTools = getSelectedTools({
-			tools: allTools,
-			mode: config.mode,
-			includeTools: config.includeTools,
-			excludeTools: config.excludeTools,
-		});
-
-		return { client: client.result, mcpTools, error: null };
-	} catch (error) {
-		await client.result.close();
```
BerniWittmann, line 114: the try catch got removed as well

packages/@n8n/nodes-langchain/nodes/mcp/McpClient/McpClient.node.ts
```diff
@@ -238,94 +239,96 @@ export class McpClient implements INodeType {
...
-					throw new NodeOperationError(node, errorMessage, {
-						itemIndex,
-					});
+					continue;
 				}
-			}
 
-			return [returnData];
-		} finally {
-			await client.result.close();
```
BerniWittmann, line 328: this got removed as well

BerniWittmann (changes requested):
Hey @sinehypernova-0718 
Thanks for doing all these changes, but i think resolving the merge conflicts introduced some unintended changes

sinehypernova-0718:
@BerniWittmann  really sorry about that. Since, I am currently away from my main setup, I am bound to use Github from my phone and the GitHub conflict resolver on mobile stripped out a bunch of blocks it shouldn't have. Completely my fault for not catching it before pushing.


Really appreciate your patience, I know this has taken way longer than it should've. I'll have a clean update ready shortly. 🙏

BerniWittmann:
@sinehypernova-0718 No worries 👍

sinehypernova-0718:
@BerniWittmann thank you so much! Really appreciate your patience and understanding. This is exactly why I love the n8n community, people like you make open source worth contributing to. 🙏 Fixes coming shortly!

sinehypernova-0718:
@BerniWittmann All the missing blocks that were accidentally removed during the merge conflict resolution have now been fully restored in the latest commit.

Really appreciate your patience and thorough review, everything should be clean now. Let me know if you spot anything else. 

Thank you! 🙏

sinehypernova-0718:
@BerniWittmann  I've pushed some more fixes, the missing blocks should all be back, the import is sorted in the latest commit, and I've double-checked everything I can from here. I believe this should finally be good to go.

I know how valuable your time is and I'm genuinely sorry for dragging this out and wasting it with back-and-forth mistakes. Lesson learned, won't let a merge conflict get the better of me again. 

Really appreciate you sticking with me through this. 🙏

BerniWittmann:
@sinehypernova-0718 No worries at all, looking good already :+1: 

Onyl some automated linters/checks fail (See CI failures). Looks like just some violations regarding indentation and "formal" checks.

sinehypernova-0718:
@BerniWittmann Thank you! Glad it's looking good now. I'll check the CI failures and fix those formatting issues right away. Appreciate it! 🙏

sinehypernova-0718:
@BerniWittmann formatting issues and indentation should both be fixed now. Let me know if there's anything else needed. 

Thanks!

BerniWittmann:
@sinehypernova-0718 Changes look good, but as you can see backend unit tests (https://github.com/n8n-io/n8n/actions/runs/24961004396/job/73089476037?pr=27297) are failiing and they seem related to your changes. 
We would need to fix them first. 
Let me know if you need help with that :)

sinehypernova-0718:
@BerniWittmann  Ah, I missed that, thanks for catching it. I'll dig into the unit test logs now and figure out what's going wrong. Might take you up on the offer for help once I see what's failing. Appreciate it!

sinehypernova-0718:
Hey! @BerniWittmann I found it,  three `supplyData` tests were missing the name property on the mock node, causing `buildMcpToolName` to fail. Fixed and pushed. Should be green now. 🤞 Will tag you once the CI has finished running. 

Thank You!

sinehypernova-0718:
@BerniWittmann fix is pushed but it seems like the CI checks still hasn't triggered yet automatically. Could you trigger it when you get a chance?

Thanks! 🙏

sinehypernova-0718:
@BerniWittmann the CI checks finally triggered and that check if now green ✅ 

Thanks again for all the guidance! 🙏

BerniWittmann (approved):
🙌 Thanks for your persistence and contribution. Congrats 🎉 . Will be merging this as soon as I can. 

(CI did not run initially, because i have to approve workflow runs manually for Community PRs)

sinehypernova-0718:
@BerniWittmann Thanks a lot, I really appreciate you sticking through with me. Learned a lot from it.
Thanks again for the guidance and encouragement and for merging this! 🙏 🚀

PR #27334, opened 2026-03-20
fix(core): Confirm messages immediately when no destination is listening

guillaumejacquart:
## Summary

Restores inline message confirmation for unlicensed instances where the log-streaming module is not initialized. Previously, messages would accumulate indefinitely across restarts since confirmations only occurred through the licensed module, causing OOM errors when reading unfinished executions on startup.

The fix adds a check in `send()` to confirm messages immediately when no `'message'` listener exists, while preserving the existing behavior for licensed instances where the destination service handles confirmation.

Fixes https://linear.app/n8n/issue/IAM-404

## Review / Merge checklist

- [x] Tests included (two new integration tests validating both paths)
- [x] Code follows existing patterns from prior implementation

### DIFF
--- a/packages/cli/src/eventbus/message-event-bus/message-event-bus.ts
+++ b/packages/cli/src/eventbus/message-event-bus/message-event-bus.ts
@@ -206,7 +206,7 @@ export class MessageEventBus extends EventEmitter {
 			this.logger.debug(`Found unsent event messages: ${unsentMessages.length}`);
 			for (const unsentMsg of unsentMessages) {
 				this.logger.debug(`Retrying: ${unsentMsg.id} ${unsentMsg.__type}`);
-				await this.emitMessage(unsentMsg);
+				this.emitMessageWithCallback('message', unsentMsg);
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


packages/cli/src/eventbus/message-event-bus/message-event-bus.ts
```diff
@@ -224,6 +224,11 @@ export class MessageEventBus extends EventEmitter {
 		}
 		for (const msg of msgs) {
 			this.logWriter?.putMessage(msg);
+			// If no destination is listening (e.g. log-streaming module not licensed),
+			// confirm the message immediately to prevent unbounded accumulation across restarts.
+			if (this.listenerCount('message') === 0) {
```
BGZStephen: This feels weird, surely we should just not call `send` if the feature is not licensed?

guillaumejacquart: the confirmSent method only do this under the hood:
```ts
this.logWriter?.confirmMessageSent(msg.id, source);
```

it's needed for the log writer to consider the event as been "handled", and thus discarded on restart (because on restart, we browse the logs to know which one are not completely sent (meaning written in logs + potential external destination + sent  confirmed)).

BGZStephen: Sure, but I guess what I mean, is it's an odd pattern that we have an exclusion pattern here, instead of an inclusion.

i.e. If there are listeners, we should opt in to send to that destination, as opposed to we check that there are specifically no destinations to choose not to send it, if that makes sense?

guillaumejacquart: I think the code and naming might be misleading, but that `if` is not to prevent sending messages to the destinations, it's to consider the message as "handled" by the log writer. if there is log streaming license, and log streaming destinations setup, we call the confirm sent asynchronously on the log streaming side after the event is sent. 

The "opt-in" mechanism is already there for all destinations (they register themselves to listen and handle specific events, if they have log streaming license). 

I think another pattern we could do is register a default listener in the event bus, and override it from the log streaming if the license is featured:
```ts
// In MessageEventBus.initialize()
this.on('message', (msg, confirmCallback) => confirmCallback());

// In LogStreamingDestinationService.onModuleInit()
this.eventBus.removeAllListeners('message');
this.eventBus.on('message', this.handleMessage.bind(this));
```

Or moving the destination orchestration layer back to the event bus, so that the if / else logic 

```ts
// MessageEventBus
private messageHandlers: MessageHandler[] = [];

registerHandler(handler: MessageHandler) {
    this.messageHandlers.push(handler);
}

async send(msgs) {
    for (const msg of msgs) {
        this.logWriter?.putMessage(msg);
        
        if (this.messageHandlers.length === 0) {
            this.confirmSent(msg, { id: '0', name: 'eventBus' });
        } else {
            for (const handler of this.messageHandlers) {
                await handler.handleMessage(msg, () => this.confirmSent(msg, ...));
            }
        }
        
        await this.emitMessage(msg); // metrics, etc.
    }
}
```

That might be part of a later tech debt ticket if we agree it's an improvement point. What do you think ?

BGZStephen: Yeah, I think this code is quite confusing, for example, why do we call `await this.emitMessage(msg);` if there are no destinations?

I'm happy to approve this as is, it just feels like building confusion on top of confusion

guillaumejacquart: Yeah, i see what you mean. I think the confusion comes from the fact that we mix 3 concerns into 1:
- log writing
- log destination handling
- prometheus metrics handling (that's why we have the `await this.emitMessage(msg);` nonetheless)

Let me see what I can do about it

BGZStephen: Exactly this, it felt like there are already side effects in this function which might be legacy

guillaumejacquart: Hey ! I see no fast way into refactoring this. What I did for now is to make the flow more explicit in the comments and function names. Let me know what you think. 
I'll create a tech debt ticket for better split and perhaps centralized orchestration flow with clearer interface contracts.

BGZStephen (approved):

PR #27339, opened 2026-03-20
feat(fs-proxy): Permission management

dlavrenuek:
## Summary

Defines and Implements permission management in n8n Gateway.
- [Specification](https://github.com/n8n-io/n8n/blob/bfb796f571a0a56565a16f449b118cb285442b37/packages/%40n8n/fs-proxy/spec/local-gateway.md#permission-management)
- Implemented Deny/Ask/Allow permission modes and user confirmation for all tool groups within the Gateway
- Implemented settings initialization and persistence
- Additional security measures:
  - User must confirm the settings (permission modes) before the connection can be established and can change these settings
- Implemented settings templates for easier setup (recommended, allow all, custom)
- Added new flags for non-interactive mode: `--non-interactive` (no user prompts, will decline per default), `--auto-confirm` (auto confirm all user prompts)
- Refactored configuration a bit, removed unused or outdated configuration options

https://linear.app/n8n/issue/NODE-4655/implement-permission-framework-with-confirmation-options

## Running locally

To run the n8n Gateway locally:

1. check out the branch `feature/instance-ai-gateway-permissions`
2. install dependencies and build n8n
```shell
$ pnpm i
$ pnpm build
```
3. Start n8n and then start n8n Gateway
```shell
$ cd packages/@n8n/fs-proxy
$ pnpm start
```

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/fs-proxy/spec/local-gateway.md
+++ b/packages/@n8n/fs-proxy/spec/local-gateway.md
@@ -224,12 +224,13 @@ the matching resources.
 When a tool group operates in `Ask` mode, confirmation is scoped to a
 **resource**. The resource is defined by the tool itself. For Browser
 Automation the resource is the **domain** (e.g. `github.com`). For Shell
-Execution the resource is the **command** (e.g. `npm`). For other tool groups
-the resource is determined by the respective tool.
-
-A single tool call may involve multiple resources (e.g. a chained shell
-command `npm install && git push` involves two resources: `npm` and `git`).
-Each resource must be confirmed independently before the tool executes.
+Execution the resource is the **normalized command**: wrapper commands
+(`sudo`, `env`, etc.) and environment variable assignments are stripped, and
+the executable basename replaces an absolute path (e.g. `sudo apt install foo`
+→ `apt install foo`). Compound or otherwise unrecognizable commands (chained
+operators, command substitution, variable-indirect execution, relative paths)
+are returned as-is so the full command is visible in the confirmation prompt.
+For other tool groups the resource is determined by the respective tool.
 
 Resource-level `always deny` rules take precedence over the tool group
 permission mode. A resource with a stored `always deny` rule is blocked
--- a/packages/@n8n/fs-proxy/spec/local-gateway.md
+++ b/packages/@n8n/fs-proxy/spec/local-gateway.md
@@ -179,9 +179,152 @@ are independent — read-only mode remains the default.
 
 ---
 
-## Human-in-the-Loop (HITL)
+## Permission Management
 
-> **Work in progress.** This section will define how the AI requests user
-> confirmation before performing sensitive or destructive local operations
-> (e.g. executing a shell command, clicking a UI element). Details to be
-> specified.
+The Local Gateway uses a two-tier permission model: **tool group permission
+modes** (coarse-grained, configured at startup) and **resource-level rules**
+(fine-grained, confirmed at runtime during tool execution).
+
+### Tool Group Permission Modes
+
+Each tool group has an independent permission mode, configured before the
+gateway connects and stored in the gateway configuration file.
+
+| Tool Group | Available Modes |
+|---|---|
+| Filesystem Access | Deny / Ask / Allow |
+| Filesystem Write Access | Deny / Ask / Allow |
+| Shell Execution | Deny / Ask / Allow |
+| Computer Control | Deny / Ask / Allow |
+| Browser Automation | Deny / Ask / Allow |
+
+**Deny** — The tool group is disabled. Its tools are not registered with the
+n8n instance; the AI has no knowledge of them.
+
+**Ask** — The tool group is enabled. Before each tool execution the user is
+prompted to confirm. Confirmation is scoped to a resource (see below).
+Existing resource-level rules are applied automatically without prompting.
+
+**Allow** — The tool group is enabled. All tool calls execute without user
+confirmation. Resource-level `always allow` rules have no effect in this mode.
+Permanently stored `always deny` rules still take precedence and will block
+the matching resources.
+
+**Constraints:**
+
+- The gateway cannot start unless at least one tool group is set to `Ask` or
+  `Allow`.
+- If Filesystem Access is set to `Deny`, Filesystem Write Access is also
+  disabled regardless of its own mode.
+
+---
+
+### Resource-Level Rules
+
+When a tool group operates in `Ask` mode, confirmation is scoped to a
+**resource**. The resource is defined by the tool itself. For Browser
+Automation the resource is the **domain** (e.g. `github.com`). For Shell
+Execution the resource is the **command** (e.g. `npm`). For other tool groups
+the resource is determined by the respective tool.
+
+A single tool call may involve multiple resources (e.g. a chained shell
+command `npm install && git push` involves two resources: `npm` and `git`).
+Each resource must be confirmed independently before the tool executes.
+
+Resource-level `always deny` rules take precedence over the tool group
+permission mode. A resource with a stored `always deny` rule is blocked
+regardless of whether the tool group is set to `Ask` or `Allow`. All
+other resource-level rules (`allow once`, `allow for session`, `always allow`)
+apply only when the tool group is in `Ask` mode.
+
+#### Rule Types
+
+| Rule | Effect | Persistence |
+|---|---|---|
+| Allow once | Execute this specific invocation only | Not stored |
+| Allow for session | Execute all invocations of this resource until the session ends | In-memory, cleared on session end |
+| Always allow | Execute all future invocations of this resource | Stored permanently in config |
+| Deny once | Block this specific invocation only | Not stored |
+| Always deny | Block all future invocations of this resource | Stored permanently in config |
+
+Permanently stored resource-level rules (`always allow`, `always deny`) are
+stored in the gateway configuration file, separately from the tool group
+permission modes.
+
+---
+
+### Runtime Confirmation Prompt
+
+When a tool group is in `Ask` mode and no stored rule applies to the resource,
+the user is presented with a confirmation prompt. The prompt shows:
+
+- The tool group being used
+- The resource being accessed (domain, command, path, etc.)
+- A description of the action the AI intends to perform
+- The confirmation options: `Allow once`, `Allow for session`, `Always allow`,
+  `Deny once`, `Always deny`
+
+---
+
+### Session
+
+A session is defined as a single active connection between the Local Gateway
+and the n8n instance. A session ends when the user explicitly disconnects or
+the n8n instance terminates the connection. A temporary network interruption
+followed by automatic reconnection is considered part of the same session.
+
+`Allow for session` rules persist across such re-connections and are cleared
+only when the session ends.
+
+---
+
+## Startup Configuration
+
+### Permission Setup
+
+Before the gateway connects, the user must configure the permission mode for
+each tool group. The gateway will not start unless at least one tool group is
+enabled (`Ask` or `Allow`).
+
+**CLI** — An interactive prompt lists each tool group with its current mode.
+If a valid configuration already exists the user can confirm it with `y` or
+edit individual modes before proceeding.
+
+**Native application** — The user sees an equivalent configuration UI.
+
+### Filesystem Root Directory
+
+When any filesystem tool group (Filesystem Access or Filesystem Write Access)
+is enabled, the user must specify a root directory. The AI can only access
+paths within this directory — all operations on paths outside are rejected.
+This applies to both read and write operations.
+
+### Configuration Templates
+
+To simplify first-time setup, three templates are available. When no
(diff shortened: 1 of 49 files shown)


packages/@n8n/fs-proxy/spec/local-gateway.md
```diff
@@ -118,9 +118,152 @@ the root is denied.
...
+(fine-grained, confirmed at runtime during tool execution).
+
+### Tool Group Permission Modes
+
+Each tool group has an independent permission mode, configured before the
+gateway connects and stored in the gateway configuration file.
+
+| Tool Group | Available Modes |
+|---|---|
+| Filesystem Access | Deny / Ask / Allow |
```
pdwarf, line 195: This is read-only access I presume?

packages/@n8n/fs-proxy/spec/local-gateway.md
```diff
@@ -118,9 +118,152 @@ the root is denied.
...
+### Resource-Level Rules
+
+When a tool group operates in `Ask` mode, confirmation is scoped to a
+**resource**. The resource is defined by the tool itself. For Browser
+Automation the resource is the **domain** (e.g. `github.com`). For Shell
+Execution the resource is the **command** (e.g. `npm`). For other tool groups
+the resource is determined by the respective tool.
+
+A single tool call may involve multiple resources (e.g. a chained shell
+command `npm install && git push` involves two resources: `npm` and `git`).
```
pdwarf: This breaks with the patterns users may expect from eg. using Claude Code, where you approve this request as one, not as two. It sounds like you'd plan to have two "Approval" dialogs in this scenario, which IMO is overkill/bad UX.

Why not treat this as one resource?

packages/@n8n/fs-proxy/spec/local-gateway.md
```diff
@@ -118,9 +118,152 @@ the root is denied.
...
+regardless of whether the tool group is set to `Ask` or `Allow`. All
+other resource-level rules (`allow once`, `allow for session`, `always allow`)
+apply only when the tool group is in `Ask` mode.
+
+#### Rule Types
+
+| Rule | Effect | Persistence |
+|---|---|---|
+| Allow once | Execute this specific invocation only | Not stored |
+| Allow for session | Execute all invocations of this resource until the session ends | In-memory, cleared on session end |
```
pdwarf, line 246: What's a session in this context?

packages/@n8n/fs-proxy/spec/local-gateway.md
```diff
@@ -118,9 +118,152 @@ the root is denied.
...
+
+### Session
+
+A session is defined as a single active connection between the Local Gateway
+and the n8n instance. A session ends when the user explicitly disconnects or
+the n8n instance terminates the connection. A temporary network interruption
+followed by automatic reconnection is considered part of the same session.
+
+`Allow for session` rules persist across such re-connections and are cleared
+only when the session ends.
```
pdwarf, line 278: What would be a typical scenario where a session ends? User shuts down computer at end-of-day?

packages/@n8n/fs-proxy/spec/local-gateway.md
```diff
@@ -118,9 +118,152 @@ the root is denied.
...
+### Configuration Templates
+
+To simplify first-time setup, three templates are available. When no
+configuration file exists the user selects a template before editing
+individual modes.
+
+| Template | Filesystem Access | Filesystem Write Access | Shell Execution | Computer Control | Browser Automation |
+|---|---|---|---|---|---|
+| **Recommended** (default) | Allow | Ask | Deny | Deny | Ask |
+| **Yolo** | Allow | Allow | Allow | Allow | Allow |
```
pdwarf, line 312: xD

packages/@n8n/fs-proxy/spec/local-gateway.md
dlavrenuek: yes. for write access there is a separate `Filesystem Write Access` permission

packages/@n8n/fs-proxy/spec/local-gateway.md
dlavrenuek: Right now I would implement all command execution as being treated as `any command`. So the user would either need to approve all command tool calls or allow (per session/permanently) all of them. The reason is simple: we need a reliable way of determining which commands are executed and I don't think that regex is reliable enough. Then for one dialog/vs two dialogs - we can think what would be a better UX. From security perspective: let's say you always want to allow npm commands but always want to be asked about git commands, how can you do it with a single dialog? Displaying different form than just 4 buttons would break UI/UX consistency. Applying the same rules to both commands takes away flexibility. Just checking the first command (without a reliable way of knowing all commands that will be executed) will compromise on security

packages/@n8n/fs-proxy/spec/local-gateway.md
dlavrenuek: For CLI tool:
user starts cli tool -> connection to n8n instance is established -> session is started
user stops the cli tool or n8n instance disconnects (for example user disables local gateway in instance UI settings) or stop n8n instance -> session ended

For Native application
Similar but there the user additionally has a button "Disconnect" and "Stop Daemon" -> both will reset the session

packages/@n8n/fs-proxy/spec/local-gateway.md
dlavrenuek: Sadly I'm not that creative or funny, I took the yolo mode idea from your research document on a similar tool 😔

packages/@n8n/fs-proxy/spec/local-gateway.md
elsmr: OpenCode (and I suspect Claude code too) have this concept of arity to decide where to `*` commands: https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/permission/arity.ts

It's not perfect though, I think it makes sense to keep it simple and start with just exact commands

packages/@n8n/fs-proxy/spec/local-gateway.md
```diff
@@ -118,9 +118,152 @@ the root is denied.
...
+whether the process has OS-level write access to that file — this is
+independent of the permission model for tools. If write access is unavailable
+the configuration is active only for the lifetime of the current process.
+
+The configuration file stores:
+
+- Permission mode per tool group
+- Filesystem root directory (required when any filesystem capability is
+  enabled)
+- Permanently stored resource-level rules (`always allow` / `always deny`)
```
elsmr, line 331: Claude settings is good inspiration here:

```json
{
  "$schema": "https://json.schemastore.org/claude-code-settings.json",
  "permissions": {
    "allow": [
      "Bash(npm run lint)",
      "Bash(npm run test *)",
      "Read(~/.zshrc)"
    ],
    "deny": [
      "Bash(curl *)",
      "Read(./.env)",
      "Read(./.env.*)",
      "Read(./secrets/**)"
    ]
  },
  "env": {
    "CLAUDE_CODE_ENABLE_TELEMETRY": "1",
    "OTEL_METRICS_EXPORTER": "otlp"
  }
}
```

packages/@n8n/fs-proxy/spec/local-gateway.md
dlavrenuek: Good examples. I did implement the command extraction in a more advanced way. Updated the spec. A good example of how it works can be seen in the tests: https://github.com/n8n-io/n8n/pull/27339/changes#diff-a6e7d000574e3a550f21e2997b6839a5c29a9fc65d1d494ba466320216cf5d99

packages/@n8n/fs-proxy/spec/local-gateway.md
dlavrenuek: I did implement it in a similar way but with separation by tool group. This should be easier to read for humans and have a better performance

```json
  "resourcePermissions": {
    "filesystemRead": {
      "allow": [],
      "deny": [
        "/Users/dimitrilavrenuk/Documents/Work/n8n/packages/@n8n/fs-proxy/tsconfig.json"
      ]
    }
  },
```

packages/@n8n/fs-proxy/src/settings-store.ts
```diff
@@ -0,0 +1,268 @@
...
+	allowForSession(toolGroup: ToolGroup, resource: string): void {
+		let set = this.sessionAllows.get(toolGroup);
+		if (!set) {
+			set = new Set();
+			this.sessionAllows.set(toolGroup, set);
+		}
+		set.add(resource);
+	}
+
+	alwaysAllow(toolGroup: ToolGroup, resource: string): void {
```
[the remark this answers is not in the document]
dlavrenuek: This is not valid. Because deny is evaluated first the user will never be prompted to grant access to a resource which is already denied.

packages/@n8n/fs-proxy/src/settings-store.ts
```diff
@@ -0,0 +1,271 @@
...
+		}
+	}
+
+	private async persist(): Promise<void> {
+		const dir = path.dirname(this.filePath);
+		await fs.mkdir(dir, { recursive: true, mode: 0o700 });
+		await fs.writeFile(this.filePath, JSON.stringify(this.persistent, null, 2), {
+			encoding: 'utf-8',
+			mode: 0o600,
+		});
```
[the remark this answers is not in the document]
dlavrenuek: if the settings file already exists then it must have been created by the user and we don't want to change it's permissions

packages/@n8n/fs-proxy/src/tools/filesystem/copy-file.ts
```diff
@@ -17,6 +17,22 @@ export const copyFileTool: ToolDefinition<typeof inputSchema> = {
 		'Copy a file to a new path. Overwrites the destination if it already exists. Parent directories at the destination are created automatically.',
 	inputSchema,
 	annotations: { defaultPermission: 'confirm' },
+	async getAffectedResources({ sourcePath, destinationPath }, { dir }) {
+		return [
+			await buildFilesystemResource(
```
dlavrenuek, line 22: For move and copy there will be two promts asking the user if they allow reading original file and write to new destination

dlavrenuek: All other operations only have one resource confirmation prompt

packages/@n8n/fs-proxy/src/gateway-client.ts
```diff
@@ -334,12 +343,49 @@ export class GatewayClient {
...
+					break;
+				case 'allowForSession':
+					settingsStore.allowForSession(resource.toolGroup, resource.resource);
+					break;
+				case 'alwaysAllow':
+					settingsStore.alwaysAllow(resource.toolGroup, resource.resource);
+					break;
+				case 'alwaysDeny':
+					settingsStore.alwaysDeny(resource.toolGroup, resource.resource);
+					throw new Error(`User denied access to ${resource.toolGroup}: ${resource.resource}`);
```
elsmr: Should we have another message when it's always denied? so the agent does not retry

packages/@n8n/fs-proxy/src/settings-store.ts
```diff
@@ -0,0 +1,271 @@
...
+interface PersistentSettings {
+	logLevel?: string;
+	port?: number;
+	permissions: Partial<Record<ToolGroup, PermissionMode>>;
+	filesystemDir?: string;
+	resourcePermissions: Partial<Record<ToolGroup, ResourcePermissions>>;
+}
+
+function isValidPersistentSettings(raw: unknown): raw is PersistentSettings {
+	if (typeof raw !== 'object' || raw === null) return false;
```
elsmr: Can we use a Zod schema here?

elsmr (commented):
Love the interactive prompt when you first launch with simple presets 💯 

For the browser tools I found that it doesn't always identify the domain correctly:


```
✔ Grant permission — browser: duckduckgo.com Allow for 
session
  ✓ browser_navigate (16887ms) {"tool":"browser_navigate","durationMs":16887}
  ▸ browser_snapshot sessionId=sess_1x8Uc-msFJj1 {"tool":"browser_snapshot","args":{"sessionId":"sess_1x8Uc-msFJj1"}}
✔ Grant permission — browser: browser Allow for session
  ✓ browser_snapshot (15144ms) {"tool":"browser_snapshot","durationMs":15144}
```

The snapshot was also on `duckduckgo.com` but it identified it as "browser" and stored it in config like that

packages/@n8n/fs-proxy/src/settings-store.ts
dlavrenuek: Good idea, fixed.

packages/@n8n/fs-proxy/src/gateway-client.ts
dlavrenuek: good idea, changed it to `User permanently denied access...`

elsmr (approved):

PR #27354, opened 2026-03-20
fix(core): Allow expressions in OAuth credential URL fields

guillaumejacquart:
## Summary

This fix allows expression-prefixed values (like `=https://...` and `={{ $vars.url }}`) in OAuth credential URL fields, enabling users to use n8n variables for multi-environment OAuth configurations. The validation that was rejecting these expressions now skips them at save time, since they're properly validated at runtime after expression resolution.

- https://linear.app/n8n/issue/IAM-386
- Fixes https://github.com/n8n-io/n8n/issues/26940
- Fixes https://github.com/n8n-io/n8n/issues/26131

## Review / Merge checklist

- [x] Tests included (added 4 test cases covering valid/invalid URLs and expression-prefixed values)
- [ ] PR Labeled with `release/backport` (if urgent)

### DIFF
--- a/packages/cli/src/credentials/credentials.service.ts
+++ b/packages/cli/src/credentials/credentials.service.ts
@@ -32,6 +32,7 @@ import {
 	CREDENTIAL_EMPTY_VALUE,
 	deepCopy,
 	displayParameter,
+	isExpression,
 	isINodePropertyCollection,
 	NodeHelpers,
 } from 'n8n-workflow';
@@ -1073,7 +1074,7 @@ export class CredentialsService {
 			const oauthUrlFields = ['authUrl', 'accessTokenUrl', 'serverUrl'] as const;
 			for (const field of oauthUrlFields) {
 				const value = data[field];
-				if (typeof value === 'string' && value.trim() !== '') {
+				if (typeof value === 'string' && value.trim() !== '' && !isExpression(value)) {
 					validateOAuthUrl(value);
 				}
 			}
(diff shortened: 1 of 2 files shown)


packages/cli/src/credentials/credentials.service.ts
```diff
@@ -1073,7 +1074,7 @@ export class CredentialsService {
 			const oauthUrlFields = ['authUrl', 'accessTokenUrl', 'serverUrl'] as const;
 			for (const field of oauthUrlFields) {
 				const value = data[field];
-				if (typeof value === 'string' && value.trim() !== '') {
+				if (typeof value === 'string' && value.trim() !== '' && !isExpression(value)) {
```
BGZStephen, line 1077: Should we be evaluating the expression to validate it on the back-end? or are we happy this isn't validated?

BGZStephen (approved):

packages/cli/src/credentials/credentials.service.ts
guillaumejacquart: I don't think we can, as the evaluation could require run context data that is not here on configuration time. I've seen this pattern of discarding specific field validation if it contains expressions (e.g https://github.com/n8n-io/n8n/blob/master/packages/@n8n/ai-workflow-builder.ee/src/validation/checks/parameters.ts#L215)

BGZStephen: Ah, fair, I imagine this is going to be tricky if we begin implementing allow / deny listing domains instance wide, if we don't evaluate this before saving

guillaumejacquart: good point. I guess we would need to re-introduce some kind of validation context, meaning we would need to know if the expression can be validated in globally of it requires runtime data

PR #27356, opened 2026-03-20
fix: Validate workflow size for workflows with pinned data on backend (no-changelog)

dariacodes:
## Summary

It turned out we're validating pinned data only on frontend. Not validating it on backend might lead to large workflows being created by AI agents. This PR adds this validation to backend.

Closes [ADO-4995](https://linear.app/n8n/issue/ADO-4995/bug-no-backend-size-validation-for-workflows-with-pinned-data)

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/workflows/workflow.service.ts
```diff
@@ -406,6 +406,11 @@ export class WorkflowService {
 			await validateEntity(workflowUpdateData);
 		}
 
+		// Validate pinData size after all mutations are applied
+		if ('pinData' in workflowUpdateData) {
+			WorkflowHelpers.validatePinDataSize({ ...workflow, ...workflowUpdateData });
+		}
```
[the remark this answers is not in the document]
dariacodes: We do not want the existing users to have a bad experience, so they should get this error when trying to update pinned data

packages/cli/src/workflows/workflow-creation.service.ts
```diff
@@ -72,17 +72,12 @@ export class WorkflowCreationService {
...
-			const personalProject = await this.projectRepository.getPersonalProjectForUserOrFail(user.id);
-			effectiveProjectId = personalProject.id;
-		}
-
-		await WorkflowHelpers.replaceInvalidCredentials(newWorkflow, effectiveProjectId);
+		await WorkflowHelpers.replaceInvalidCredentials(newWorkflow);
 
 		WorkflowHelpers.addNodeIds(newWorkflow);
 
+		WorkflowHelpers.validatePinDataSize(newWorkflow);
```
scdekov: Should we do this right before/after `validateEntity` in order to not call the external hook (and potentially other side effects) if the pin data is too big?

packages/cli/src/public-api/v1/handlers/workflows/workflows.handler.ts
```diff
@@ -38,16 +36,15 @@ export = {
...
-			);
-
-			await replaceInvalidCredentials(workflow, project.id);
+			await replaceInvalidCredentials(workflow);
 
 			addNodeIds(workflow);
-			resolveNodeWebhookIds(workflow, Container.get(NodeTypes));
 
-			addNodeIds(workflow);
+			validatePinDataSize(workflow);
```
scdekov: Move at the beginning of that function

packages/cli/src/workflow-helpers.ts
```diff
@@ -1,23 +1,50 @@
...
+			`Pinned data exceeds the maximum allowed size of ${MAX_PINNED_DATA_SIZE / (1024 * 1024)} MB`,
+		);
+	}
+
+	const { pinData: _, ...workflowWithoutPinData } = workflow;
+	const workflowSize =
+		Buffer.byteLength(JSON.stringify(workflowWithoutPinData), 'utf8') + pinDataSize;
+	const limit = MAX_WORKFLOW_SIZE - MAX_EXPECTED_REQUEST_SIZE;
+	if (workflowSize > limit) {
+		throw new BadRequestError('Workflow with pinned data exceeds the maximum allowed size');
```
scdekov: Should we mention the limit here too?

scdekov (dismissed):

scdekov (approved):

PR #27357, opened 2026-03-20
fix(core): Fix execution history when flow includes wait node

DawidMyslak:
## Summary

Re-applies the fix from #23146 (which was reverted in #25610 due to PostgreSQL timezone issues) with a corrected approach.

**Original bug:** When a workflow with a Wait node resumes, `setRunning()` overwrites the original `startedAt` timestamp with the current time, causing wrong execution duration and sort order in execution history.

**Why the original fix broke PostgreSQL:** PR #23146 used raw SQL `COALESCE(startedAt, :startedAt)` with `DateUtils.mixedDateToUtcDatetimeString()`, which produces timezone-ambiguous date strings. On PostgreSQL with `timestamptz` columns, these strings were interpreted using the session timezone instead of UTC, causing 1-hour offsets (or even negative durations) for non-UTC users.

**This fix:** Uses a transaction with standard TypeORM `findOneBy` + `update` methods instead of raw SQL. TypeORM handles date serialization correctly per database driver (SQLite and PostgreSQL), eliminating the timezone mismatch entirely.

### How to test

1. Create a workflow: Webhook → Wait (resume by webhook) → Edit Fields
2. Trigger the webhook to start execution
3. Execution enters "waiting" state — note the `startedAt` time
4. Resume the execution via the wait webhook
5. Verify `startedAt` remains unchanged (not overwritten with resume time)

Sample workflow is also provided in the [Linear ticket](https://linear.app/n8n/issue/CAT-1854/execution-history-incorrect-when-flow-includes-wait-node).

- https://linear.app/n8n/issue/CAT-1854/execution-history-incorrect-when-flow-includes-wait-node
- https://linear.app/n8n/issue/CAT-2306/community-issue-wrong-execution-time-after-upgrade-to-264
- Original fix: https://github.com/n8n-io/n8n/pull/23146
- Revert: https://github.com/n8n-io/n8n/pull/25610

## Review / Merge checklist

- [x] Tests included.


tomi (dismissed):
🚀

tomi (approved):

PR #27361, opened 2026-03-20
fix(Google Drive): Download file node does not trigger error output when continueOnFail is enabled

danieljayasurya-e:
##  Bug

The Google Drive "Download File" node does not correctly handle `continueOnFail()` when "error output" is enabled.

When a file download fails (for example, بسبب insufficient permissions), the node silently returns an empty array instead of routing execution to the error output branch. This causes the workflow to stop successfully without triggering the error path.

---

## Fix

Updated the error handling in the download operation:

- When `continueOnFail()` is enabled:
  - Return a structured error using `createExecutionError`
  - Ensure the error is passed to the error output branch
- When disabled:
  - Preserve existing behavior (throw error)

---

## How to test

1. Create a workflow with:
   - Google Drive → Download File node
2. Use a file without access permission
3. Enable:
   - `Continue On Fail → using error output`
4. Connect error output to another node (e.g., Set node)

### Expected
- Error branch executes
- Workflow does NOT stop silently

---

## 🔗 Related issue

fixes #27312

---

## Notes

Previously, the node returned an empty array on failure, which prevented error output execution. This change aligns behavior with other n8n nodes that properly support `continueOnFail()`.

---

## Checklist

- [ ] Docs updated (not required for this fix)
- [ ] Tests included (can be added in follow-up if needed)
- [ ] No backport required

### DIFF
--- a/packages/nodes-base/nodes/Google/Drive/v1/GoogleDriveV1.node.ts
+++ b/packages/nodes-base/nodes/Google/Drive/v1/GoogleDriveV1.node.ts
@@ -2722,9 +2722,17 @@ export class GoogleDriveV1 implements INodeType {
 			} catch (error) {
 				if (this.continueOnFail()) {
 					if (resource === 'file' && operation === 'download') {
-						items[i].json = { error: error.message };
+						// Return a dedicated error item so the workflow engine can move it to the
+						// error output instead of treating the download as a silent empty result.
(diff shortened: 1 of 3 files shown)


danieljayasurya-e (commented):
fix(Google Drive): Download file node does not trigger error output when continueOnFail is enabled

PR #27364, opened 2026-03-20
fix(editor): Remove unused options from NDV settings for agent model nodes

scdekov:
## Summary

The reported issue highlighted 2 problems:
1. We're showing retry options for model nodes in the NDV settings but we don't use those and in that case the user should either set retry on the agent node or do 2.
2. If retry is set on the model using the langchain mechanism we're not retrying 429 errors

Since the user that reported the issue was trying to setup the retry using the option in the NDV settings I've hidden those options (solving problem 1. here). Problem 2 has been tackled in a separate branch

https://linear.app/n8n/issue/AI-2164/community-issue-openai-chat-model-node-does-not-retry-429-rate-limit

## Review / Merge checklist

- [ ] Tests included. 


JoseBra (approved):
LGTM

PR #27371, opened 2026-03-20
fix(editor): Node references in expressions not updated when the renamed node has quotes

RomanDavydchuk:
## Summary

Whenever a node that has quotes in the name (for example, `Call 'Some workflow'`) was renamed, expressions referencing the node weren't updated. This was due to the fact that the expression contained literal backslashes (`$('Call \'Some workflow\'')`) that weren't taken into account. This PR fixes that, also makes sure that backslashes and quotes are properly escaped when replacing the old name with a new one

https://linear.app/n8n/issue/IAM-378/node-renaming-breaks-when-there-are-quotes-in-the-node-name

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/workflow/test/node-reference-parser-utils.test.ts
+++ b/packages/workflow/test/node-reference-parser-utils.test.ts
@@ -126,10 +126,50 @@ describe('NodeReferenceParserUtils', () => {
 				expected: 'someRandomExpression("oldName")',
 			},
 			{
-				expression: '$("old\\"Name")',
-				previousName: 'old\\"Name',
+				expression: '$("someone\'s node")',
+				previousName: "someone's node",
+				newName: 'other node',
+				expected: '$("other node")',
+			},
+			{
+				expression: '$(\'some "node"\')',
+				previousName: 'some "node"',
+				newName: 'other node',
+				expected: "$('other node')",
+			},
+			{
+				expression: '$("test\\\\some")',
+				previousName: 'test\\some',
+				newName: 'other node',
+				expected: '$("other node")',
+			},
+			{
+				expression: '$("old name")',
+				previousName: 'old name',
+				newName: 'new "name"',
+				expected: '$("new \\"name\\"")',
+			},
+			{
+				expression: '$("old name")',
+				previousName: 'old name',
+				newName: "new 'name'",
+				expected: '$("new \\\'name\\\'")',
+			},
+			{
+				expression: '$("test some")',
+				previousName: 'test some',
+				newName: 'other\\node',
+				expected: '$("other\\\\node")',
+			},
+			{
+				// $("old\"Na\\me\'") -> old"Na\me'
+				expression: '$("old\\"Na\\\\me\\\'")',
+				// old"Na\me'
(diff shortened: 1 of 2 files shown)


packages/workflow/test/node-reference-parser-utils.test.ts
```diff
@@ -126,10 +126,50 @@ describe('NodeReferenceParserUtils', () => {
...
+				expression: '$("test\\\\some")',
+				previousName: 'test\\some',
+				newName: 'other node',
+				expected: '$("other node")',
+			},
+			{
+				expression: '$("old name")',
+				previousName: 'old name',
+				newName: 'new "name"',
+				expected: '$("new \\"name\\"")',
```
BGZStephen, line 150: I'm a little confused why:

`new "name"` ends up as `'$("new \\"name\\"")'` (double back slash)

When: 

`"new 'name'"` returns `'$("new \\\'name\\\'")'` (triple back slash)

packages/workflow/test/node-reference-parser-utils.test.ts
```diff
@@ -126,10 +126,50 @@ describe('NodeReferenceParserUtils', () => {
...
+			{
+				// $("old\"Na\\me\'") -> old"Na\me'
+				expression: '$("old\\"Na\\\\me\\\'")',
+				// old"Na\me'
+				previousName: 'old"Na\\me\'',
+				// n\'ew\"Name
 				newName: 'n\\\'ew\\"Name',
-				expected: '$("n\\\'ew\\"Name")',
+				// $("n\\\'ew\\\"Name") -> n\'ew\"Name
+				expected: '$("n\\\\\\\'ew\\\\\\"Name")',
```
BGZStephen, line 172: This one doesn't quite seem right, `ew\\"Name` should evaluate to `ew\\\\\"Name`, no? 5 `\` not 6

packages/workflow/test/node-reference-parser-utils.test.ts
RomanDavydchuk: In the case with single quotes the third slash is needed to escape the quotes around `name`, since the `expected` string is itself is wrapped in single quotes

packages/workflow/test/node-reference-parser-utils.test.ts
RomanDavydchuk: 6 `\` in `ew\\\\\\"Name` would evaluate to a string `ew\\\"Name`, so 3 `\` in the result. Then, since it'll be evaluated again, `\\\"` will become literal `\"` characters (`\\` + `\"`) -> `ew\"Name`

In a case with 5 `\` we'd have `\\\\\"` -> `\\"` (`\\` + `\\` + `\"`). But keep in mind that `\\"` is inside of a string in `$("...")`, so the `"` should be escaped by a 3rd `\`

BGZStephen: What a mess :cry:

BGZStephen (approved):

PR #27372, opened 2026-03-20
fix(Xero Node): Make OAuth2 scopes configurable to support write operations

### LINKED ISSUE #27363: Xero OAuth2 node does not support granular scopes / fails to authorize or perform write actions
The native n8n Xero integration appears to be outdated and does not properly support Xero’s newer OAuth2 + granular scope model. As a result, we are unable to reliably authenticate or perform write operations (e.g. updating contacts), and have had to switch to HTTP nodes as a workaround.

**What we’re trying to do**
We are building an automation to:
- Trigger from Airtable
- Search for a contact in Xero using a unique ID (AccountNumber)
- Update the contact fields

**Expected behavior**
- Xero OAuth2 connection should allow selecting or handling required scopes (e.g. accounting.contacts)
- Successful authentication should allow both:
- Reading contacts (GET)
- Updating contacts (POST)

**Actual behavior**
- The native Xero node does not expose or allow configuration of scopes
- Authentication flow completes, but:
- Write requests fail (401 / 403 errors depending on setup)
- Behavior is inconsistent compared to direct HTTP calls
- The connector does not appear to support Xero’s newer granular scope requirements

**What we observed**
- GET requests (e.g. retrieving contacts) may work
- POST requests (e.g. updating contacts) fail using the native node
- Using HTTP Request nodes with the same OAuth credentials works correctly when:
- Proper scopes are configured
- xero-tenant-id header is included

**Workaround currently in use**
We are using HTTP Request nodes instead of the native Xero node:
- OAuth2 credential configured manually
- Custom headers:
- xero-tenant-id
- Accept: application/json
- Content-Type: application/json
- API endpoint:
- https://api.xero.com/api.xro/2.0/Contacts

This works as expected, but is more complex and harder to maintain.

**Impact**
- Prevents use of native Xero node for production workflows
- Increases complexity of automation setup
- Makes maintenance and onboarding more difficult for the team

**Request**
- Update the Xero node to support:
- Granular OAuth2 scopes (e.g. accounting.contacts)
- Proper handling of write operations (POST/PUT)

**Additional context**
We were able to successfully complete the same workflow using HTTP nodes, which suggests the issue is specific to the native Xero integration rather than our credentials or Xero setup.

anmolxlight:
Fixes #27363 by allowing users to specify custom OAuth2 scopes (like `accounting.contacts`) required for write operations.

### DIFF
--- a/packages/nodes-base/credentials/XeroOAuth2Api.credentials.ts
+++ b/packages/nodes-base/credentials/XeroOAuth2Api.credentials.ts
@@ -1,6 +1,6 @@
 import type { ICredentialType, INodeProperties } from 'n8n-workflow';
 
-const scopes = [
+const defaultScopes = [
 	'offline_access',
 	'accounting.transactions',
 	'accounting.settings',
@@ -36,12 +36,6 @@ export class XeroOAuth2Api implements ICredentialType {
 			default: 'https://identity.xero.com/connect/token',
 		},
 		{
-			displayName: 'Scope',
-			name: 'scope',
-			type: 'hidden',
-			default: scopes.join(' '),
-		},
-		{
 			displayName: 'Auth URI Query Parameters',
 			name: 'authQueryParameters',
 			type: 'hidden',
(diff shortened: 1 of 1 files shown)


Joffcom:
@anmolxlight As this has been in a draft state for a while and has been already fixed I am going to move this to closed, Thanks for taking the time to open a PR though.

PR #27397, opened 2026-03-22
fix(nodes-base): implement S3 bucket delete operation and fix related issues

ybucci:
## Summary

Fixes #27396

Three issues were found and fixed in the S3 node:

### 1. `bucket > delete` operation not implemented

The **Delete** operation was defined in `BucketDescription.ts` and visible in the UI, but the execution block was **never implemented** in `S3.node.ts`. When triggered, it silently returned `success` with 0 output items in ~2ms — without making any HTTP request.

**Fix:** Added the missing `if (operation === 'delete')` block inside `if (resource === 'bucket')`.

```typescript
//https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteBucket.html
if (operation === 'delete') {
  const name = this.getNodeParameter('name', i) as string;
  await s3ApiRequestSOAP.call(this, `${name}`, 'DELETE', '');
  const executionData = this.helpers.constructExecutionMetaData(
    this.helpers.returnJsonArray({ success: true }),
    { itemData: { item: i } },
  );
  returnData.push(...executionData);
}
```

### 2. `s3ApiRequestSOAP` silently swallows errors on empty responses

In `GenericFunctions.ts`, the `s3ApiRequestSOAP` function had a `catch (error) { return error }` pattern. Any S3 operation returning a non-XML response (e.g. `HTTP 204 No Content`) would cause `parseString` to throw, which was silently caught and **returned as data** instead of being re-thrown. This masked real failures.

**Fix:** Empty responses (`HTTP 204`) now return `undefined` early. Parse errors are re-thrown correctly.

### 3. `destinationPath` in file copy requires undocumented leading slash

The `file > copy` operation parses `destinationPath` assuming it starts with `/` (using `split('/')[1]` to extract the bucket name). However, the UI field description does not mention this requirement, unlike `sourcePath` which explicitly states *"should start with (/)"*. Without the leading slash, the bucket name resolves to the file key, causing a `"resource not found"` error.

**Fix:** Normalize `destinationPath` to always start with `/` before parsing, making it resilient to both `bucket/key` and `/bucket/key` formats.

## Test plan

- [ ] Create a bucket via S3 node (`bucket > create`) ✅
- [ ] Delete a bucket via S3 node (`bucket > delete`) — previously broken, now fixed
- [ ] Copy a file with `destinationPath` without leading slash — previously broken, now fixed
- [ ] Verify existing operations (upload, download, getAll, search, folder ops) are unaffected

## Notes

- Verified against MinIO (S3-compatible). `DELETE /{bucket}` returns `HTTP 204 No Content`, which was the root cause of the silent failure in `s3ApiRequestSOAP`.
- The same `BucketDescription.ts` is shared with the AWS S3 node — that node should be checked for the same missing `delete` implementation.

### DIFF
--- a/packages/nodes-base/nodes/S3/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/S3/GenericFunctions.ts
@@ -144,6 +144,10 @@ export async function s3ApiRequestSOAP(
 		option,
 		region,
 	);
+	// Empty responses (e.g. HTTP 204 No Content) are valid and should return undefined,
+	// not swallow a parse error silently.
+	if (!response) return undefined;
+
 	try {
 		return await new Promise((resolve, reject) => {
 			parseString(response as string, { explicitArray: false }, (err, data) => {
@@ -154,7 +158,7 @@ export async function s3ApiRequestSOAP(
 			});
 		});
 	} catch (error) {
-		return error;
+		throw error;
 	}
 }
 
--- a/packages/nodes-base/nodes/S3/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/S3/GenericFunctions.ts
@@ -144,9 +144,11 @@ export async function s3ApiRequestSOAP(
 		option,
 		region,
 	);
-	// Empty responses (e.g. HTTP 204 No Content) are valid and should return undefined,
+	const nodeVersion = this.getNode().typeVersion ?? 1;
+
+	// v1.1+: Empty responses (e.g. HTTP 204 No Content) are valid and should return undefined,
 	// not swallow a parse error silently.
-	if (!response) return undefined;
+	if (nodeVersion >= 1.1 && !response) return undefined;
 
 	try {
 		return await new Promise((resolve, reject) => {
@@ -158,7 +160,9 @@ export async function s3ApiRequestSOAP(
 			});
 		});
 	} catch (error) {
-		throw error;
+		// v1.1+: Re-throw parse errors instead of silently returning them as data.
+		if (nodeVersion >= 1.1) throw error;
+		return error;
 	}
 }
 
--- a/packages/nodes-base/nodes/S3/S3.node.ts
+++ b/packages/nodes-base/nodes/S3/S3.node.ts
@@ -23,7 +23,7 @@ export class S3 implements INodeType {
 		// eslint-disable-next-line n8n-nodes-base/node-class-description-icon-not-svg
 		icon: 'file:s3.png',
 		group: ['output'],
-		version: 1,
+		version: [1, 1.1],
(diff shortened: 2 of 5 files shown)


packages/nodes-base/nodes/S3/GenericFunctions.ts
```diff
@@ -144,6 +144,10 @@ export async function s3ApiRequestSOAP(
 		option,
 		region,
 	);
+	// Empty responses (e.g. HTTP 204 No Content) are valid and should return undefined,
+	// not swallow a parse error silently.
+	if (!response) return undefined;
```
Joffcom: This may change functionality that could be considered a breaking change. It will need a light version if this is going to be needed

packages/nodes-base/nodes/S3/S3.node.ts
```diff
@@ -548,7 +558,11 @@ export class S3 implements INodeType {
 							).toUpperCase();
 						}
 
-						const destinationParts = destinationPath.split('/');
+						// Normalize: ensure destinationPath starts with '/' (e.g. '/bucket/key')
+						const normalizedDestination = destinationPath.startsWith('/')
```
Joffcom: What happens if it doesn't start with a slash? This may also need a light version due to default behaviour change.

Joffcom (commented):
Added a couple of notes, this will also need tests to either be updated or added.

packages/nodes-base/nodes/S3/GenericFunctions.ts
ybucci: Thanks for the feedback! Addressed in the latest commit by introducing light versioning (v1.1).

The empty-response guard (`if (!response) return undefined`) and the parse-error re-throw are now gated behind `typeVersion >= 1.1`. Existing nodes on v1 keep the original behaviour unchanged.

The node version array was bumped to `[1, 1.1]` so new instances automatically get the fixed behaviour.

packages/nodes-base/nodes/S3/S3.node.ts
ybucci: Good catch! Also addressed via light versioning (v1.1).

If `destinationPath` doesn't start with `/` on v1, the bug is preserved as-is (legacy behaviour): `split('/')` on e.g. `'bucket/key.txt'` gives `['bucket', 'key.txt']`, so `[1]` resolves to the file key instead of the bucket name, causing a `resource not found` error — exactly what was happening before.

On v1.1, the path is normalised before splitting, so both `'bucket/key.txt'` and `'/bucket/key.txt'` are handled correctly.

Joffcom:
@ybucci I was just checking some of the PRs in my queue, It looks like there are now conflicts on this one and the CLA might not be signed which is a requirement.

PR #27399, opened 2026-03-23
fix(googleFirebaseCloudFirestore Node): Preserve field types for numeric strings and falsy values

### LINKED ISSUE #11413: FireStore node saving STRING as NUMBER
### Bug Description

I have created this workflow where I use the **SET node** to create new values (variables) and then use **Google Cloud Firestore** node to store data in a collection. In the SET node I made few variables which is declared as STRING but has a value of number like 9999. When I execute this workflow, n8n creates field in the document but changes the datatype in firestore to number. When I use double quotes to enclose variable value in SET node, then the workflow add additional quotes in the firestore document field.

### To Reproduce

1. Create a Manual Trigger, then SET node and then Google firestore node
2. Connect Manual trigger to SET node
3. Edit the SET node and use Manual Mapping to create a STRING field.
4. Set the value of this STRING field to any numeric data like 9999
5. Now set up the Google firestore as follows (or similar)
     Set **Resource** to **Document**
     **Operation** to **Create or Update**
     **Project** and **Database** to anything suitable
     **Collection** to any suitable collection name
     **Update Key** to any document ID (in firebase)
     **Columns/Attributes** to the names of fields used in **SET** node
6. Now execute the n8n workflow and fields would be generated in Firestore collection

step1
Step2
Step3
Step4

### Expected behavior

Since I have set the datatype as STRING in the SET node so the datatype in Google firestore should also remain the STRING but somehow it gets converted to NUMBER which causes the leading zeros to disappear.

### Operating System

Ubuntu 24.04.1 LTS

### n8n Version

Version 1.64.3

### Node.js Version

v20.18.0

### Database

SQLite (default)

### Execution mode

main (default)

xndvaz:
## Summary

- Preserve input types in Cloud Firestore `jsonToDocument` by converting to numeric Firestore values only when the input is actually a JavaScript `number`.
- Prevent accidental timestamp coercion of year-like numeric strings (for example `"9999"`) by only parsing strings as timestamps when they look date-like (`-`, `T`, or `:` present).
- Fix `create` operation field handling to preserve falsy values (`0`, `false`, `""`) by checking field presence with `hasOwnProperty` instead of truthiness.
- Keep existing behavior for missing columns: they are still written as `nullValue`.
- Add regression coverage:
  - update `GenericFunctions` unit tests for numeric string handling
  - add a new Cloud Firestore `create` workflow test that validates request payload and output for `0`, `false`, empty string, numeric string, and a missing field

## How to test

```bash
cd packages/nodes-base
N8N_USER_FOLDER=/tmp/n8n-test-cloudfirestore pnpm test nodes/Google/Firebase/CloudFirestore/__test__
```

## Related

Fixes #11413


Qodo-Free-For-OSS:
Hi, In document create/upsert, a field explicitly present with value `undefined` will pass the `hasOwnProperty` check and be converted by `jsonToDocument(undefined)` into `{}`, producing an invalid Firestore field value instead of `nullValue`.

**How to fix:** Treat undefined as nullValue

> ### Issue description
> `jsonToDocument(undefined)` currently returns `{}`, which becomes an invalid Firestore field value when create/upsert sees an explicitly-present `undefined` column.
> 
> ### Issue Context
> Create/upsert now uses `hasOwnProperty` so falsy values are preserved, but this also includes `undefined`.
> 
> ### Fix Focus Areas
> - packages/nodes-base/nodes/Google/Firebase/CloudFirestore/GenericFunctions.ts[88-117]
>   - Add an explicit `value === undefined` branch (likely return `{ nullValue: null }`, or alternatively treat `undefined` as "missing" consistently).
> - packages/nodes-base/nodes/Google/Firebase/CloudFirestore/__test__/GenericFunctions.test.ts[345-353]
>   - Update the edge-case expectation for `undefined` accordingly.
> - (Optional) packages/nodes-base/nodes/Google/Firebase/CloudFirestore/GoogleFirebaseCloudFirestore.node.ts[187-203]
>   - If you prefer presence logic: treat `undefined` same as missing and write `nullValue`.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

xndvaz:
Implemented a follow-up fix for the explicit undefined-field case in create/upsert.

Changes included:
- jsonToDocument(undefined) now maps to { nullValue: null } instead of {}
- Updated Cloud Firestore GenericFunctions unit test expectation for undefined accordingly

Validation:
- nodes/Google/Firebase/CloudFirestore/__test__/GenericFunctions.test.ts passes with the new behavior (34 passed).

xndvaz:
Follow-up fix for the explicit `undefined` field case has been added and pushed.

Validation:
- `nodes/Google/Firebase/CloudFirestore/__test__/GenericFunctions.test.ts` passes with the updated expectation (`34/34`).

If there are no other concerns, this should be ready for another look.

xndvaz:
@pdwarf, could you please take another look at this PR? The follow-up fix for the explicit `undefined` field case has been added and validated.

PR #27415, opened 2026-03-23
fix(Box Node): Fix issue where Box trigger node was not paginating correctly

### LINKED ISSUE #25013: Box trigger node logic is broken
### Bug Description

Due to recent changes in the Box API for [listing a user's webhooks](https://developer.box.com/reference/get-webhooks), the Box trigger logic appears to be broken in two places:

- The helper function [`boxApiRequestAllItems`](https://github.com/n8n-io/n8n/blob/df8bafc24fa13dad25d498928533d64b2acf128d/packages/nodes-base/nodes/Box/GenericFunctions.ts#L50) uses `limit` and `offset` to increment pagination; however, the API now only supports scroll markers
- The trigger webhook method [`checkExists`](https://github.com/n8n-io/n8n/blob/n8n%402.6.1/packages/nodes-base/nodes/Box/BoxTrigger.node.ts#L285) is relying on fields that no longer exist (at least from the initial API call). The API for listing webhooks no longer appears to include the `address` or `triggers` properties of a webhook, breaking the logic in the comparisons to identify an existing webhook. It seems that a secondary call to the [get webhook by ID](https://developer.box.com/reference/get-webhooks-id) endpoint will now be necessary

**Note:** I have a patch that fixes these issues and would be happy to submit a PR

### To Reproduce

1. Create valid `Box OAuth2 API` credentials and connect it to a user account
2. Create a workflow
3. Add the Box trigger node and configure it with the credentials from step 1
4. Specify valid parameters for the events, target type, and target ID
5. Activate/publish the workflow to cause the webhook creation
6. Observe that the UI enters a failed hanging state. Upon refreshing, the workflow is activated/published but the webhook does not exist and does not work

If this is attempted on an instance where Box trigger webhooks already exist (were created before this bug), then you will notice that no new ones can be added

### Expected behavior

I should be able to add Box triggers without issue

### Debug Info

# Debug info

## core

- n8nVersion: 2.4.6
- nodeJsVersion: 22.21.1
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/143.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-01-28T23:03:32.827Z

### Operating System

macOS Tahoe 26.2

### n8n Version

2.4.6

### Node.js Version

22.21.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

Joffcom:
## Summary
The Box trigger node was using the wrong pagination type, This adds a second pagination type and updates the node to use it along with fetching the details for the webhook if needed due to the "mini webhook" option.

https://linear.app/n8n/issue/NODE-4656/community-issue-box-trigger-node-logic-is-broken#comment-54ad0262
Closes #25013

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Box/BoxTrigger.node.ts
```diff
@@ -262,27 +262,41 @@ export class BoxTrigger implements INodeType {
 			async checkExists(this: IHookFunctions): Promise<boolean> {
 				const webhookUrl = this.getNodeWebhookUrl('default');
 				const webhookData = this.getWorkflowStaticData('node');
-				const events = this.getNodeParameter('events') as string;
+				const events = this.getNodeParameter('events') as string[];
```
ShireenMissi, line 265: should line 308 also be updated?
`const events = this.getNodeParameter('events') as string;`

ShireenMissi (commented):
The change is solid 👏 I left one comment to address before I can approve

packages/nodes-base/nodes/Box/BoxTrigger.node.ts
Joffcom: @ShireenMissi nice spot 🙏🏻

ShireenMissi (approved):
LGTM ✅

PR #26907, opened 2026-03-11
feat(editor): Add Workflow Publish Timeline tab to Workflow History 

CharlieKolb:
## Summary

Introduce a new tab on the workflow history for displaying a timeline of published state, version and dates.
Only appears for workflows which were published at any point.

image

Also add a new item to the `Publish` dropdown to highlight this functionality.

image

Currently shows an empty screen if there is no publish history to be explicit about the fact.

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/ADO-4993

## Review / Merge checklist

- [x] I have seen this code, I have run this code, and I take responsibility for this code.
- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #27428, opened 2026-03-23
feat(core): Track instance version history (no-changelog)

CharlieKolb:
## Summary

Add a new db table which tracks version updates on launch. Exposes some unused functions on the controller - I mostly need `getDateSinceContinuouslyAtLeastVersion` but I figured the others may demonstrate other useful access patterns.

Not sure if we want any additional data on this, open to suggestions.

Also not sure if the major, minor, patch ints are flexible enough for long term storage, but it lets us avoid parsing and handling edge cases on read. Maybe we should add a `affix` columns which stores an optional affix parsed from `N8N_VERSION`?

_____

Concrete use case is https://github.com/n8n-io/n8n/pull/26907, where we previously deleted past publish history entries when the version itself was deleted, and we want to tell apart complete histories from partial ones

https://linear.app/n8n/issue/ADO-5002

## Review / Merge checklist

- [ ] Tests included. 


RicardoE105 (commented):
Good stuff 🚀 

issue (blocking): init() can crash the instance on a non-critical feature.

If the DB is temporarily unreachable loadCache() will throw and bubble up through module init. Version tracking shouldn't be able to take down the instance:

```
  async init(): Promise<void> {
      try {
          await this.loadCache();
          await this.checkAndRecordCurrentVersion();
      } catch (error) {
          this.logger.warn('Failed to initialize version history', { error });
      }
  }
```

  ---
  suggestion (non-blocking): Consider handling the race in multi-main setups.

  If two main instances start simultaneously after an upgrade, both will check the cache, find no entry for the new
  version, and both will insert. There's no unique constraint on (major, minor, patch). The duplicate entries are
  functionally harmless (the scan algorithms still return correct results), but an upsert or unique constraint would keep
   the data clean.

  ---
  thought: Pre-release versions are silently collapsed.

  parseVersion('2.5.0-beta.1') stores {2, 5, 0}, so a beta-to-stable transition on the same major.minor.patch won't be
  recorded. If the debugging use case (finding cloud instances that ran a bad beta) matters later, this is a gap. If not,
   a comment in parseVersion acknowledging the intentional loss would help future readers.

packages/cli/src/modules/instance-version-history/instance-version-history.service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Returns the smallest (minimum) version since the given date.
+	 */
+	async getMinVersionSince(since: Date): Promise<SemVer | undefined> {
+		const cache = await this.getCache();
+		let min: VersionEntry | undefined;
+		for (const entry of cache) {
+			if (entry.createdAt >= since) {
+				if (!min || compareVersions(entry, min) < 0) {
```
scdekov: nit: use `versionGte`

scdekov (dismissed):

scdekov (approved):

PR #27440, opened 2026-03-23
feat(core): Add RoleMappingRule entity and database tables

BGZStephen:
## Summary

Adds the database layer for **role mapping rules**: rules map an expression to a target role, scoped by **type** and ordered by **`order`** (lower = higher priority; first match wins). A rule can apply to **multiple projects** via a join table.

**Schema**
- `role_mapping_rule`: `id` (UUID, PK), `expression` (text), `role`, `type`, `order` (integer); indexes on `role` and on `(type, order)`.
- `role_mapping_rule_project`: composite PK `(roleMappingRuleId, projectId)`; FKs to `role_mapping_rule` and `project` with `ON DELETE CASCADE`; index on `projectId`.

**Application**
- New TypeORM entity `RoleMappingRule` with `@ManyToMany` to `Project` and `@JoinTable` on `role_mapping_rule_project`.
- Inverse `@ManyToMany` on `Project.roleMappingRules`.
- UUIDs for rule IDs are generated in `@BeforeInsert` with `randomUUID()` (no DB default), consistent with `@n8n/db` guidance.

**Migrations**
- Common migration `1772800000000-CreateRoleMappingRuleTable` registered for **PostgreSQL and SQLite**.

**How to test**
1. Run migrations against a dev DB (`n8n start` or your usual migration path).
2. Confirm tables exist: `role_mapping_rule`, `role_mapping_rule_project`.
3. Optional: persist a `RoleMappingRule` with `projects` loaded/saved via TypeORM and verify join rows.

https://linear.app/n8n/issue/IAM-396

## Review / Merge checklist

- [ ] Tests included.


packages/@n8n/db/src/entities/role-mapping-rule.ts
```diff
@@ -0,0 +1,54 @@
...
+@Index(['role'])
+@Index(['type', 'order'])
+export class RoleMappingRule {
+	@PrimaryColumn('uuid')
+	id: string;
+
+	@Column({ type: 'text' })
+	expression: string;
+
+	@Column({ type: String, length: 255 })
```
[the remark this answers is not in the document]
BGZStephen: No, string is fine

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
```diff
@@ -0,0 +1,46 @@
...
+ * via join table `role_mapping_rule_project`.
+ */
+export class CreateRoleMappingRuleTable1772800000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(ruleTable)
+			.withColumns(
+				column('id').uuid.primary.notNull,
+				column('expression').text.notNull,
+				column('role').varchar(128).notNull, // matches slug length of role table
+				column('type').varchar(64).notNull,
```
guillaumejacquart: Maybe we could add a comment here about the possible values ?

packages/@n8n/db/src/entities/role-mapping-rule.ts
```diff
@@ -0,0 +1,54 @@
...
+	type Relation,
+} from '@n8n/typeorm';
+import { randomUUID } from 'node:crypto';
+
+import { Project } from './project';
+
+@Entity()
+@Index(['role'])
+@Index(['type', 'order'])
+export class RoleMappingRule {
```
guillaumejacquart: Just curious: why not extending the WithTimestampsAndStringId as it seems is an implicit 'standard' in the codebase (I think ?)
Why uuid instead of nanoid ?

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
guillaumejacquart: Maybe we could add a comment for the expected values here ? I mean a sql comment (`.comment`)

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
```diff
@@ -0,0 +1,46 @@
...
+ * evaluation (first match wins) within a rule type. Rules apply to many projects
+ * via join table `role_mapping_rule_project`.
+ */
+export class CreateRoleMappingRuleTable1772800000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(ruleTable)
+			.withColumns(
+				column('id').uuid.primary.notNull,
+				column('expression').text.notNull,
+				column('role').varchar(128).notNull, // matches slug length of role table
```
guillaumejacquart, line 17: I see no foreign key constraint on the role table (slug column), is that by design ?

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
```diff
@@ -0,0 +1,46 @@
...
+				column('type').varchar(64).notNull,
+				column('order').int.notNull,
+			)
+			.withIndexOn('role')
+			.withIndexOn(['type', 'order']);
+
+		await createTable(joinTable)
```
guillaumejacquart, line 34: One concern i have is around db consistency: right now with the suggested solution, you can have a role_mapping_rule row with type = 'instance' but associated to projects. 
Are we fine with that ?
The alternatives would be to:
- create 2 distinct tables for instance and project level rules
- or maybe dropping the type, and considering the rule without any project mapping (left join ...null) as global

guillaumejacquart (commented):
Good stuff ! some comments

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
BGZStephen: Yep, that's fair, let me add it and handle cascading updates and deletes

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
BGZStephen: I see what you're saying, but I don't see it as that much of a concern, the repository and service layers should be responsible for validation of data prior to insertion, which should provide adequate consistency, while retaining flexibility to introduce other types like we have for roles (workflow, credential)

packages/@n8n/db/src/entities/role-mapping-rule.ts
BGZStephen: Good catch, I think I was just unlucky with the repository I referenced when writing this that didn't extend :D have added

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
BGZStephen: Sure, never done that before but happy to give it a go

packages/@n8n/db/src/migrations/common/1772800000000-CreateRoleMappingRuleTable.ts
guillaumejacquart: That makes sense, fine for me

packages/@n8n/db/src/entities/role-mapping-rule.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Role } from './role';
+
+@Entity()
+@Index(['role'])
+@Index(['type', 'order'])
+export class RoleMappingRule extends WithTimestampsAndStringId {
+	@Column({ type: 'text' })
+	expression: string;
+
+	@ManyToOne('Role', 'roleMappingRules')
```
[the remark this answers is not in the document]
BGZStephen: No, this isn't the pattern elsewhere

guillaumejacquart (approved):
Thanks for addressing the comments !

PR #27448, opened 2026-03-23
chore: Add create-skill agent skill

alexander-gekov:
## Summary

Adds a concise **create-skill** agent skill under `.claude/skills/create-skill/` so contributors have a shared pattern for writing skills: short `SKILL.md`, frontmatter rules, progressive disclosure, and where to put skills in this repo (`.claude/skills/`, including Cursor picking up project skills from there).

**How to review:** Read `.claude/skills/create-skill/SKILL.md` and check it matches how you want the team to author skills.

https://linear.app/n8n/issue/NODE-4674

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/.claude/skills/create-skill/SKILL.md
+++ b/.claude/skills/create-skill/SKILL.md
@@ -68,6 +68,20 @@ description: >-         # max 1024 chars, non-empty — see below
 
 **Rough size:** aim for **well under ~200 lines** in `SKILL.md`; if it grows, split detail out.
 
+### Scope: one job per skill (and parent skills)
+
+- **Single responsibility** — one primary workflow or decision tree per skill. If triggers and steps diverge a lot (e.g. “create issue” vs “create PR” vs “full ticket → PR flow”), split into **smaller dedicated skills**.
+- **Prefer small + compose** — two or three focused skills keep irrelevant detail out of context until needed. A **parent** (orchestrator) skill can say *when* to follow each child workflow and link to their `SKILL.md`; avoid pasting full child content into the parent.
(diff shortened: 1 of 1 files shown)


.claude/skills/create-skill/SKILL.md
```diff
@@ -0,0 +1,111 @@
+---
```
DawidMyslak, line 1: This all looks good, some minor thoughts from me.

---

Could we include some guidelines on how to properly reference (if available) MCPs, tools or even other skills?

For example, some people (not all) might have a Linear MCP installed which is required for the SKILL to operate. Should we come up with some conventions? (cc @Joffcom).

Also, I think there is a consensus on preferring command-line tools over MCPs if available, usually they're perfectly understood by the clanker and don't destroy the context window.

---

Another one is about single responsibility principle. I think people struggle to understand how broad the skill should be. Is it better to have 2/3 smaller, dedicated skills for precise tasks and then later, have a bigger skill invoking those 2/3 skills, or just make a single "large" skill?

alexander-gekov: All good points for discussions 👍

Personally, I think smaller skills are better and we can have parent skills that can reference them, main benefit being not polluting the context when not necessary.

As for the reference of MCP/tools I'll have a look at what the best practice is rn

alexander-gekov: Added some guidelines for when referencing MCPs/tools/etc and some notes on large skills v small specific skills

DawidMyslak (approved):

PR #27484, opened 2026-03-24
feat(core): Add Slack signature identifier for dynamic credentials (no-changelog)

phyllis-noester:
## Summary
  - Adds SlackSignatureIdentifier for HMAC-SHA256 signature verification with constant-time comparison
  - Implements replay attack prevention (5-min window)
  - Supports user_id and team_user subject claim modes

  Part 1/3 of [IAM-211](https://linear.app/n8n/issue/IAM-211)

  ## Test plan
  - [x] Signature verification (valid/invalid/expired)
  - [x] Both subject claim modes
  - [x] Options validation and NaN timestamp handling


packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/identifiers/slack-signature-identifier.ts
```diff
@@ -0,0 +1,143 @@
...
+		if (
+			!context.metadata?.rawBody ||
+			!context.metadata?.signature ||
+			!context.metadata?.timestamp
+		) {
+			throw new IdentifierValidationError('Missing Slack signature verification data');
+		}
+		this.verifySlackSignature(
+			options.signingSecret,
+			context.metadata.timestamp as string,
```
BGZStephen: A shame to cast here, is there no way to safely parse these as string?

BGZStephen (dismissed):

BGZStephen (approved):

PR #27485, opened 2026-03-24
feat(core): Add Slack signature extractor hook for dynamic credentials (no-changelog)

phyllis-noester:
## Summary
  - Adds SlackSignatureExtractor context establishment hook
  - Extracts Slack user identity from slash commands, interactive messages, and events
  - Carries raw signature data in metadata for downstream verification
  - Masks sensitive headers in trigger items

  Part 2/3 of [IAM-211](https://linear.app/n8n/issue/IAM-211)

  ## Test plan
  - [x] Identity extraction from all 3 payload formats
  - [x] Enterprise ID handling
  - [x] Header masking and raw body fallbacks
  - [x] Error handling for missing data


packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/slack-signature-extractor.ts
```diff
@@ -0,0 +1,190 @@
...
+		// Extract identity from the Slack payload
+		const body = triggerItem.json['body'];
+		const { userId, teamId, enterpriseId } = this.extractSlackIdentity(body);
+
+		if (!userId) {
+			throw new Error('Could not extract user_id from Slack payload');
+		}
+
+		// Mask Slack signature headers in trigger items to avoid leaking them
+		headers['x-slack-signature'] = '**********';
```
BGZStephen, line 96: Should this use the blanking values constant?

phyllis-noester: I don't think so, we also don't use it in packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/http-header-extractor.ts and IMO it makes sense to set to a different value since we're actually blanking this out on write rather than read

packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/slack-signature-extractor.ts
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	/**
+	 * Extracts Slack user identity from various payload formats.
+	 * Slack sends different payload structures for:
+	 * - Slash commands (form-encoded: user_id, team_id)
+	 * - Interactive messages (JSON: user.id, team.id)
+	 * - Events (JSON: event.user, team_id)
+	 */
+	private extractSlackIdentity(body: unknown): {
```
BGZStephen, line 160: Would be nice to extract these to their own functions to shorten the code here:

`if (typeof payload['user_id'] === 'string') {` -> extractFlatFieldBody()

`if (user && typeof user === 'object' && 'id' in user) {` -> extractInteractivePayloadBody

`if (event && typeof event === 'object') {` -> extractEventPayloadBody

packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/slack-signature-extractor.ts
```diff
@@ -0,0 +1,190 @@
...
+		userId: string | undefined;
+		teamId: string | undefined;
+		enterpriseId: string | undefined;
+	} {
+		if (!body || typeof body !== 'object') {
+			return { userId: undefined, teamId: undefined, enterpriseId: undefined };
+		}
+
+		const payload = body as Record<string, unknown>;
```
BGZStephen: We shouldn't be casting here, it's unsafe moving from unknown to record should ideally be handled by zod

BGZStephen (dismissed):
The Zod comment is probably the most pressing, the other comments asre mainly coding preference

packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/slack-signature-extractor.ts
```diff
@@ -0,0 +1,218 @@
...
+	constructor(private readonly logger: Logger) {}
+
+	hookDescription: HookDescription = {
+		name: 'SlackSignatureExtractor',
+		displayName: 'Slack Identity Extractor',
+		options: [],
+	};
+
+	isApplicableToTriggerNode(nodeType: string): boolean {
+		return nodeType === 'n8n-nodes-base.webhook' || nodeType === 'webhook';
```
afitzek, line 58: Did we try to set this up in a slack trigger node as well?

afitzek (approved):
nice 👍

packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/slack-signature-extractor.ts
phyllis-noester: I did but I generally can't get the slack trigger to work (even without dynamic creds, some 429 error stuff) so I suggested to move it into a follow-up ticket

PR #27486, opened 2026-03-24
feat(core): Add Slack credential resolver for dynamic credentials (no-changelog)

phyllis-noester:
## Summary
  - Adds SlackCredentialResolver using Slack signing for per-user credential storage
  - Registers Slack components in the dynamic-credentials module
  - Handles encrypt/decrypt with safe error handling (no credential leaks in logs)

## How to test:
- Add the dynamic-creds-test Slack app to a channel
- Call the /test-get-emails slash command (app configured at https://api.slack.com/apps/A0AGW5T570C/general)
- The related n8n workflow should execute and return results — no "Raw Body" setting needed on the webhook node https://test-iam211-slack-cred-rs.stage-app.n8n.cloud/workflow/p4dN8dHH35Jm6FmJ

  ## Test plan
  - [x] Get/set/delete credential lifecycle
  - [x] Decrypt failure handling (ERR_OSSL_BAD_DECRYPT)
  - [x] User isolation
  - [x] Options validation error paths


packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/slack-credential-resolver.ts
```diff
@@ -0,0 +1,179 @@
...
+		const key = await this.slackSignatureIdentifier.resolve(context, parsedOptions);
+		await this.storage.deleteCredentialData(credentialId, key, handle.resolverId, parsedOptions);
+	}
+
+	async deleteAllSecrets(handle: CredentialResolverHandle): Promise<void> {
+		await this.storage.deleteAllCredentialData(handle);
+	}
+
+	async validateOptions(options: CredentialResolverConfiguration): Promise<void> {
+		const result = SlackCredentialResolverOptionsSchema.safeParse(options);
```
guillaumejacquart: Why not call the parseOptions here ? in order to have a single place where we parse the options

guillaumejacquart (commented):
Just a small comment

packages/cli/src/modules/dynamic-credentials.ee/credential-resolvers/slack-credential-resolver.ts
phyllis-noester: good point! addressed now :)

guillaumejacquart (approved):
🚀

PR #27492, opened 2026-03-24
ci: Create stable release on GitHub promotion

Matsuuu:
## Summary

When promoting a GitHub release to latest, also create (or recreate) a floating `stable` release tag that always points to the most recently promoted version. This makes it easy to reference the latest stable release by a fixed tag name.

Changes:
- After promoting the versioned release, look up and delete any existing `stable` release
- Create a new `stable` release pointing to the same commit and with the same release notes
- Output the stable release URL as a GitHub Actions output (`stable_release_url`)

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/.github/scripts/promote-github-release.mjs
+++ b/.github/scripts/promote-github-release.mjs
@@ -1,4 +1,5 @@
 import {
+	deleteRelease,
 	ensureEnvVar,
 	getExistingRelease,
 	initGithub,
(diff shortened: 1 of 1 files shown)


.github/scripts/promote-github-release.mjs
```diff
@@ -34,8 +35,29 @@ async function promoteGitHubRelease() {
 
 	console.log(`Successfully updated release ${releaseResponse.data.html_url}`);
 
+	const existingStableRelease = await getExistingRelease('stable');
```
tomi, line 38: Does this operate on the tag or the GH release?

Matsuuu: It finds a GitHub Release by the tag name. This will remove the release but not the **Git** tag.

We do delete-then-insert here to prevent artifacts from being dragged from older releases to new ones.

.github/scripts/promote-github-release.mjs
```diff
@@ -34,8 +35,29 @@ async function promoteGitHubRelease() {
...
+	if (existingStableRelease) {
+		await deleteRelease(existingStableRelease.id);
```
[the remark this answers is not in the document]
Matsuuu: Yep, this is by design.

tomi (approved):
🚀

PR #27516, opened 2026-03-24
fix(instance-ai): Scope project listing to only return projects the user can access (no-changelog)

aalises:
## Summary

Fixes a bug where the AI assistant (and MCP `search_projects` tool) listed **all users' personal projects** instead of only the current user's, causing confusing prompts like:

> "I see you have multiple personal projects. Which personal project is the MTG Card Bot currently in?"
> 1. "Jaakko Husso" 2. "Chat User" 3. "Member User"

### Root cause

`ProjectRepository.getAccessibleProjects()` and `getAccessibleProjectsAndCount()` used:

```sql
WHERE type = 'personal' OR projectRelations.userId = :userId
```

The first branch (`type = 'personal'`) matched **every** personal project in the instance regardless of ownership. Since every user already has a `projectRelation` to their own personal project (as `project:personalOwner`), the second branch already captures the user's own personal project — the first branch just adds everyone else's.

### Who uses these methods (and why it's a bug for all of them)

Three production callers rely on the fixed repository methods:

1. **`ProjectController.getAllProjects`** (`project.controller.ts`) — the `GET /projects` REST endpoint used by the UI sidebar and project pickers. Was returning other users' personal projects in the list even though the current user has no scopes on them and can't do anything with them.

2. **`InstanceAiAdapterService`** (`instance-ai.adapter.service.ts`) — calls `getAccessibleProjects` to determine which projects the user can work with. This was the main trigger for the bug — the AI saw all personal projects and asked "Which personal project?" on multi-user instances.

3. **MCP `search-projects` tool** (`search-projects.tool.ts`) — calls `getAccessibleProjectsAndCount` to search for projects the user can access. The tool's own description states its intent: *"Search for projects accessible to the current user. Use this to find a project ID before creating a workflow in a specific project."* It is explicitly scoped to `user.id` and meant to return only projects the user can work with — the buggy repository method was violating this contract by returning every personal project in the instance.

### Why this does NOT affect workflow transfers

Workflow transfer uses a **completely separate code path**. `WorkflowService.transferWorkflow()` validates the destination project via `projectService.getProjectWithScope()`, which checks whether the user has the `workflow:create` scope on that specific project — it never calls `getAccessibleProjects` or `getAccessibleProjectsAndCount`. Users don't have `workflow:create` on other users' personal projects, so transfers to foreign personal projects were never possible regardless of this bug.

### Fix

Remove the blanket `{ type: 'personal' }` clause from both repository methods. The user's own personal project is already returned via the `projectRelations.userId` join. Admin users are unaffected — they go through a separate `projectRepository.find()` code path that returns everything.

**Files changed:**
- `packages/@n8n/db/src/repositories/project.repository.ts` — Remove the `type = 'personal'` OR clause; switch from `LEFT JOIN` + `DISTINCT` to a simpler `INNER JOIN` (no duplicates possible now)
- `packages/cli/test/integration/database/repositories/project.repository.test.ts` — Update tests to assert the corrected behavior (member only sees own personal project)

## Related

- Linear: https://linear.app/n8n/issue/AI-2264
- Notion: https://www.notion.so/n8n/AI-sees-multiple-personal-projects-32d5b6e0c94f80eea0faf0f60dd58d69

## Test plan

- [x] Integration tests updated and passing — member now sees only own personal project + team projects they belong to
- [x] Project service unit tests passing
- [x] MCP search-projects tool unit tests passing
- [x] Project controller unit tests passing
- [x] Typecheck passing

### DIFF
--- a/packages/@n8n/db/src/repositories/project.repository.ts
+++ b/packages/@n8n/db/src/repositories/project.repository.ts
@@ -33,17 +33,13 @@ export class ProjectRepository extends Repository<Project> {
 		});
 	}
 
-	// This returns personal projects of ALL users OR shared projects of the user
 	async getAccessibleProjects(userId: string) {
 		return await this.find({
-			where: [
-				{ type: 'personal' },
-				{
-					projectRelations: {
-						userId,
-					},
+			where: {
+				projectRelations: {
+					userId,
 				},
-			],
+			},
 		});
 	}
 
@@ -61,20 +57,10 @@ export class ProjectRepository extends Repository<Project> {
 		userId: string,
 		options: ProjectListOptions,
 	): Promise<[Project[], number]> {
-		// Build a subquery that finds the IDs of accessible projects, avoiding
-		// duplicate rows from the LEFT JOIN on projectRelations which would
-		// produce incorrect counts with getManyAndCount().
 		const idsQuery = this.createQueryBuilder('p')
-			.select('DISTINCT p.id', 'id')
-			.leftJoin('p.projectRelations', 'pr')
-			.where(
-				new Brackets((qb) => {
-					qb.where('p.type = :personalType', { personalType: 'personal' }).orWhere(
-						'pr.userId = :userId',
-						{ userId },
-					);
-				}),
-			);
+			.select('p.id', 'id')
+			.innerJoin('p.projectRelations', 'pr')
+			.where('pr.userId = :userId', { userId });
 
 		if (options.search) {
 			idsQuery.andWhere('LOWER(p.name) LIKE LOWER(:search)', {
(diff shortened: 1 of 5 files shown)


packages/@n8n/db/src/repositories/project.repository.ts
```diff
@@ -33,17 +33,13 @@ export class ProjectRepository extends Repository<Project> {
 		});
 	}
 
-	// This returns personal projects of ALL users OR shared projects of the user
 	async getAccessibleProjects(userId: string) {
```
Cadiac, line 36: This doesn't fully guarantee that the user is able to do anything within this project - `project:chatUser` role is a weak role who is only able to chat on chat hub with workflows within that project - but I think we can ignore that and do this assumption that every project the user has some project role in is "accessible".

Cadiac (approved):
🚀

PR #27522, opened 2026-03-24
feat(editor): Replace restore/changes UI with inline version cards

DeveloperTheExplorer:
## Summary

Replaces the old inline "Restore version" links on user messages with dedicated **Version Card** components (`ChatVersionCard`) that appear between AI generations, showing node-level diffs.

Adds **non-destructive version restore**: when a user restores to an older version, messages after the restored version collapse into a `CollapsedMessagesGroup` instead of being deleted. The LLM thread is still truncated (backend), but the UI preserves the full history.

Key changes:

- **ChatVersionCard**: expandable card showing per-node diffs (added/modified/deleted), with restore, diff view, and history menu actions
- **CollapsedMessagesGroup**: collapsed region for messages after a restore point, with expand/collapse and "not visible to AI" note
- **Non-destructive restore flow**: `activeVersionCardId` and `resumeAfterRestoreMessageId` persisted to DB via new migration columns on `workflow_builder_session`
- **Post-modification versioning**: version cards use the post-modification workflow version (not the pre-modification revertVersionId), ensuring accurate diffs
- **Session reload parity**: version card IDs use `version-card-${userMessageId}` format consistently between live streaming and session reload
- **Collapsed-aware diffing**: `useReviewChanges` skips collapsed cards when finding predecessors for non-collapsed versions

### How to test

1. Open a workflow with the AI builder
2. Send 2-3 prompts that modify the workflow
3. Verify version cards appear between generations with correct node diffs
4. Click the `...` menu on an older version card → Restore
5. Confirm messages after the restored version collapse (with "X versions hidden" divider)
6. Send a new message — verify collapsed region stays, new messages appear after
7. Refresh the page — verify collapse state persists
8. Restore to the latest version — verify no collapse occurs (current conversation stays visible)

https://linear.app/n8n/issue/AI-2162

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/features/ai/assistant/composables/useReviewChanges.ts
```diff
@@ -221,14 +307,35 @@ export function useReviewChanges() {
...
 
 		const sourceWorkflow = {
 			...workflowsStore.workflow,
-			nodes: cachedVersionNodes.value,
-			connections: cachedVersionConnections.value,
+			nodes: sourceNodes,
+			connections: sourceConnections,
+		};
+		const targetWorkflow = {
+			...workflowsStore.workflow,
```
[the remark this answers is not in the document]
DeveloperTheExplorer: The issue is not valid — or at least, not actionable right now. Here's why:

The openDiffView function at [useReviewChanges.ts:336-345](vscode-webview://1143kj7dnjk9ueaigtvlgu6rilou7mbs6uum4scb7g7u215sagsr/packages/frontend/editor-ui/src/features/ai/assistant/composables/useReviewChanges.ts#L336-L345) spreads workflowsStore.workflow to provide the IWorkflowDb shape required by the diff modal, then overrides nodes and connections with version-specific data.

The comment suggests using workflowDocumentStore instead, but workflowDocumentStore doesn't expose name, description, or versionId — fields required by IWorkflowDb and actively read by the diff view (e.g., useWorkflowDiffUI reads name at line 64-65, settings at 41-42, tags at 75-78). There's no useWorkflowDocumentName composable — those fields haven't been migrated yet.

Additionally, there's no functional bug here: both sourceWorkflow and targetWorkflow use the same ...workflowsStore.workflow base, so name/settings/tags are identical on both sides and produce zero false diffs. The only fields that differ between source and target are nodes and connections, which come from the version cache — and that's correct behavior.

packages/frontend/editor-ui/src/app/composables/useWorkflowSaving.ts
```diff
@@ -221,6 +221,14 @@ export function useWorkflowSaving({
 				workflowDataRequest.expectedChecksum = workflowDocumentStore.checksum;
 				workflowDataRequest.autosaved = autosaved;
 
+				console.debug('[saveCurrentWorkflow] sending update', {
```
schrothbn: Shouldn't be here

packages/frontend/editor-ui/src/features/ai/assistant/composables/useReviewChanges.ts
```diff
@@ -221,14 +314,35 @@ export function useReviewChanges() {
...
+
+		const latestCard = cards[cards.length - 1];
+		const latestCached = versionDataCache.value.get(latestCard.data.versionId);
+		if (!latestCached) return;
+
+		// Source: previous version's data (or empty workflow for the first card)
+		let sourceNodes: INode[] = [];
+		let sourceConnections: IConnections = {};
+		if (cards.length > 1) {
+			const prevCard = cards[cards.length - 2];
```
schrothbn: This ignores the collapsed state of the cards so we always compare the current workflow to the last version even if it's hidden. This seems very confusing and not intented?

schrothbn (commented):
Generally looks okay but ignoring the collapsed state when opening the diffView seems at least confusing to the user.

schrothbn (approved):

ivov (approved):
Migrations LGTM, haven't looked at the rest

PR #27527, opened 2026-03-24
feat(core): Add Redis storage backend for instance registry (no-changelog)

afitzek:
## Summary

Add Redis-backed storage for the instance registry module, enabling distributed n8n deployments to track instances via Redis.

- Define 4 Lua scripts (`REGISTER`, `READ_ALL`, `CLEANUP`, `UNREGISTER`) as TypeScript constants for atomic multi-key Redis operations using `{instance:}` hash tags for cluster slot co-location
- Implement `RedisInstanceStorage` with dedicated Redis client, Zod validation, non-blocking error handling, and 5s command timeout
- Add unit tests (25 cases) covering all methods, error paths, and schema validation
- Add `'instance-registry'` log scope and `'registry(n8n)'` Redis client type

- https://linear.app/n8n/issue/IAM-326
- https://linear.app/n8n/issue/IAM-327

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
```diff
@@ -0,0 +1,209 @@
...
+
+			const record = jsonParse<Record<string, unknown>>(json);
+			const state = new Map<string, InstanceRegistration>();
+
+			for (const [key, value] of Object.entries(record)) {
+				const parsed = instanceRegistrationSchema.safeParse(value);
+				if (parsed.success) {
+					state.set(key, parsed.data);
+				} else {
+					this.logger.warn('Skipping invalid state entry', {
```
guillaumejacquart, line 128: Just wondering whether it would be worth having some kind of "invalid state" state for this kind of errors, so that the instance is not completely invisible from the registry.
Same for the AI reviewer comment in getAllRegistrations

guillaumejacquart (commented):
Just one comment, otherwise looks good !

packages/@n8n/api-types/src/instance-registry-types.ts
```diff
@@ -4,7 +4,9 @@ import { z } from 'zod';
 const InstanceRegistrationSchemaV1 = z
 	.object({
 		schemaVersion: z.literal(1),
-		instanceKey: z.string(),
+		instanceKey: z
```
BGZStephen, line 7: On the face of it this seems like a possibly breaking change?

BGZStephen, line 7: On the face of things, this seems like a potentially breaking change as we're restricting a schema which isn't backwards compatible?

packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
```diff
@@ -0,0 +1,213 @@
...
+			type: 'registry(n8n)',
+			extraOptions: { commandTimeout: REGISTRY_CONSTANTS.OPERATION_TIMEOUT_MS },
+		});
+	}
+
+	async register(registration: InstanceRegistration): Promise<void> {
+		await this.upsertRegistration(registration, 'register');
+	}
+
+	async heartbeat(registration: InstanceRegistration): Promise<void> {
```
BGZStephen, line 37: Can a heartbeat and an unregister ever encounter a race condition that would cause issues?

packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
```diff
@@ -0,0 +1,213 @@
...
+					try {
+						const parsed = instanceRegistrationSchema.safeParse(jsonParse(json));
+						if (!parsed.success) {
+							this.logger.warn('Skipping invalid registration entry', {
+								error: parsed.error.message,
+							});
+							return null;
+						}
+						return parsed.data;
+					} catch (error) {
```
BGZStephen, line 76: This catch block is redundant, safeParse won't throw an error here

If you wanted a try / catch, it's actually perfectly fine to just 

```
try {
  return instanceRegistrationSchema.parse(jsonParse(json));
} catch (err) {
  this.logger.warn('Skipping malformed registration entry', {
	error: ensureError(error).message,
  });
  return null
}

packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
BGZStephen: Makes sense, you'd then be able to see any instances in a weird state in the UI

packages/@n8n/api-types/src/instance-registry-types.ts
afitzek: Yes, but this feature is still in development, so that should be fine.

afitzek: Yes, but this feature is still in development, so that should be fine.

packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
afitzek: `jsonParse` could throw an error before we reach safeParse.

BGZStephen: Ah yes, must have missed that!

packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
afitzek: Yes I think they could, but it wouldn't matter, because the entries would in that case just expire, after the TTL.

BGZStephen: so the worst case is that an instance could show as being registered for an extended time (TTL) at which point it would automatically become unregistered? Doesn't seem so bad if the TTL is low

afitzek: Exactly, I think the TTL is 60 seconds

BGZStephen (approved):

PR #27528, opened 2026-03-24
feat(core): Workflow level otel

geemanjs:
## Summary
Adds a workflow execute span:
image

Individual nodes are coming next

LIGO-386 

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/otel/otel.service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	init() {
+		if (!this.config.enabled) return;
+
+		this.sdk = new NodeSDK({
+			resource: resourceFromAttributes({
+				[ATTR_SERVICE_NAME]: 'n8n',
+				[ATTR_SERVICE_VERSION]: N8N_VERSION,
+				'n8n.instance.id': this.instanceSettings.instanceId,
+				'n8n.instance.role': this.instanceSettings.instanceType,
```
alielkhateeb: @geemanjs just curious if there is a reason why those aren't constants like the above keys?

geemanjs: Claude code 😆 - I'll change them - and write it in the README so it does that next time

packages/cli/src/modules/otel/handlers/workflow-start.handler.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { ATTR } from '../otel.constants';
+import type { SpanHandler } from './interfaces';
+import type { SpanRegistry } from '../span-registry';
+import type { Tracer } from '@opentelemetry/api';
+
+export class WorkflowStartHandler implements SpanHandler<WorkflowExecuteBeforeContext> {
+	handle(ctx: WorkflowExecuteBeforeContext, spans: SpanRegistry, tracer: Tracer) {
+		const span = tracer.startSpan('workflow.execute', {
+			attributes: {
```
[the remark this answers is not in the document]
geemanjs: Tags are not attached to the workflow execute

alielkhateeb (approved):
🔥

packages/cli/src/modules/otel/__tests__/otel-workflow-tracing.integration.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+let otel: OtelTestProvider;
+let workflowRunner: WorkflowRunner;
+let executionRepository: ExecutionRepository;
+let project: Project;
+
+beforeAll(async () => {
+	otel = OtelTestProvider.create();
+
+	process.env.N8N_OTEL_ENABLED = 'true';
```
MarcL, line 30: Better practice to remember the old values and reset in `afterAll` in case the test runner shares the process across multiple test suites?

packages/cli/src/modules/otel/__tests__/otel-workflow-tracing.integration.test.ts
```diff
@@ -0,0 +1,208 @@
...
+});
+
+beforeEach(async () => {
+	otel.reset();
+	await testDb.truncate(['ExecutionEntity', 'WorkflowEntity', 'Project']);
+	project = await createTeamProject('OTel Test Project');
+});
+
+afterAll(async () => {
+	delete process.env.N8N_OTEL_ENABLED;
```
MarcL, line 54: Reset this env var (as mentioned in above comment)?

packages/cli/src/modules/otel/__tests__/otel-workflow-tracing.integration.test.ts
geemanjs: I'll do this in the next PR - looks like this was i the merge queue when you wrote the comment 😄

MarcL: Doh!

alielkhateeb: I have thought about this while reviewing and skipped since we don't use a single process, but you're right it's still a risk even when we don't use `runInBand`, I've addressed it here d994fadbdb74879735886b90df87a703a60d061c since this PR is merged

packages/cli/src/modules/otel/__tests__/otel-workflow-tracing.integration.test.ts
alielkhateeb: Addressed in d994fadbdb74879735886b90df87a703a60d061c

PR #26949, opened 2026-03-12
ci: Update GitHub Actions to latest versions for Node.js 24 compatibility

shortstacked:
## Summary

Updates 26 GitHub Actions across 59 files to their latest versions ahead of the [Node.js 20 deprecation deadline (June 2, 2026)](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

**13 major version bumps** (node20 → node24):
- `actions/checkout` v4 → v6.0.2
- `actions/setup-node` v4/v6.1 → v6.3.0
- `actions/github-script` v7 → v8
- `actions/upload-artifact` v4/v6 → v7.0.0
- `actions/attest-sbom` v3 → v4.0.0
- `peter-evans/create-pull-request` v6/v7 → v8.1.0
- `docker/login-action` v3 → v4.0.0
- `docker/setup-buildx-action` v3 → v4.0.0
- `docker/setup-qemu-action` v3 → v4.0.0
- `docker/build-push-action` v6 → v7.0.0
- `hashicorp/setup-terraform` v3 → v4.0.0
- `astral-sh/setup-uv` v6 → v7.4.0
- `fjogeleit/http-request-action` v1 → v2.0.0

**13 minor/patch bumps** to latest:
- `pnpm/action-setup` v4.2.0 → v4.3.0
- `rharkor/caching-for-turbo` v2.3.5 → v2.3.11
- `actions/create-github-app-token` v2.0.6 → v2.2.1
- `codecov/codecov-action` v5.4.3 → v5.5.2
- `codecov/test-results-action` v1.1.1 → v1.2.1
- `CodSpeedHQ/action` v4.11.0 → v4.11.1
- `anchore/sbom-action` v0.20.6 → v0.23.1
- `isbang/compose-action` v2.0.0 → v2.5.0
- `useblacksmith/setup-docker-builder` v1.2.0 → v1.4.0
- `azure/login` v2.1.1 → v2.3.0
- `azure/setup-helm` v4 → v4.3.1
- `azure/setup-kubectl` v4 → v4.0.1
- `ncipollo/release-action` v1 → v1.20.0

**Other changes:**
- Replaced `dorny/paths-filter` (node20, no upgrade path) with built-in `ci-filter` composite in `util-data-tooling.yml`
- Fixed pre-existing actionlint warning: untrusted `github.head_ref` in `ci-restrict-private-merges.yml`
- Standardized all `actions/checkout` refs to a single SHA (was 6 different SHAs)

[...]

PR #27535, opened 2026-03-25
ci: Replace vars context with runner.name in composite action

Matsuuu:
## Summary

The `vars` context is not supported in composite actions (only in workflow files). Upgrading to `actions/setup-node@v6.3.0` in #26949 introduced a stricter template validator that throws `TemplateValidationException` at action load time — even when the `if:` condition would evaluate to `false`.

Replaces `vars.RUNNER_PROVIDER` checks with `contains(runner.name, 'blacksmith')`, which uses the `runner` context that is fully supported in composite actions. The logic is equivalent since all Blacksmith runners follow the `blacksmith-*` naming convention.

No caller changes needed.

https://github.com/n8n-io/n8n/actions/runs/23531866039

## Review / Merge checklist

- [x] Docs updated or follow-up ticket created. (N/A — CI config only)
- [x] Tests included. (N/A — CI config only)


tomi (approved):
🚀

PR #27541, opened 2026-03-25
fix(core): Remaining VM test fixes — error propagation, proxy traps, and cross-realm assertions

despairblue:
## Summary

Fixes the remaining VM expression engine test failures and adds dual-engine CI coverage. Brings the full `packages/workflow` test suite to **0 failures** with `N8N_EXPRESSION_ENGINE=vm` and ensures both engines run on every test invocation.

### Fixes
- **WHATWG percent-encoding**: VM-side `urlEncode` now encodes `!'()~` to match `URLSearchParams` output
- **Lazy proxy `ownKeys`/`getOwnPropertyDescriptor` traps**: `Object.keys()`, `Object.entries()`, `for...in`, and `JSON.stringify()` now work on proxy objects, including root proxies
- **Cross-realm test assertions**: Relaxed assertions for typed arrays, `WeakMap`/`WeakSet`, `Intl`, `Math`, `Atomics`, `ArrayBuffer`, `SharedArrayBuffer`, `DataView` to test accessibility and behavior instead of cross-realm `toEqual`
- **Process properties**: Added `process` as a lazy proxy in the VM isolate
- **`renderExpression` complexity**: Extracted `mapVmError()` to bring method complexity below eslint limit

### Error propagation across isolate boundary

Errors thrown inside the isolate or in host-side callbacks lose their class, name, and custom properties when crossing the V8 boundary. This PR adds a serialization pipeline that preserves them:

```mermaid
flowchart TD
    subgraph Host["Host - Node.js"]
        CB["Bridge callback<br/>(getValueAtPath, etc.)"]
        SE["serializeError()"]
        RE["reconstructError()"]
        MAP["mapVmError()<br/>in expression.ts"]
    end

    subgraph Isolate["V8 Isolate"]
        EXPR["Expression code"]
        E["E() handler<br/>(swallows TypeError)"]
        PROXY["Lazy proxy<br/>throwIfErrorSentinel()"]
        CATCH["Outer try-catch"]
    end

    CB -->|throws| SE
    SE -->|returns sentinel| PROXY
    PROXY -->|throws sentinel| E
    E -->|re-throws non-TypeError| CATCH

    EXPR -->|throws| E

    CATCH -->|returns sentinel via copy:true| RE
    RE -->|restores name + extra| MAP
    MAP -->|maps to ExpressionError| MAP
```

Key design decisions:
- **Sentinels, not callbacks**: Errors are serialized into `{ __isError, name, message, stack, extra }` sentinel objects and returned as values. No callback round-trip needed.
- **Single reconstruction point**: Errors are only reconstructed on the host in `reconstructError()`. The isolate throws the sentinel directly.
- **`copy:true` for extra properties**: Custom error properties (context, functionality, etc.) are transferred via V8 structured clone, not JSON.stringify.
- **Name-based mapping in `expression.ts`**: The bridge can't import host error classes, so `mapVmError()` maps by `.name` to `ExpressionError`/`ExpressionExtensionError`. A TODO tracks moving this to a shared package.

### Dual-engine CI
- Workflow package vitest config now uses workspace with two projects (`current-engine` and `vm-engine`)
- Both engines run on every `pnpm test` — locally and in CI, no workflow file changes needed
- Extracted `createBaseInlineConfig()` from `@n8n/vitest-config` for workspace project sharing

## Related

- https://linear.app/n8n/issue/CAT-2543 (Group G — parent ticket, PR #27178)
- https://linear.app/n8n/issue/CAT-2627 (Int8Array cross-isolate transfer — resolved)
- https://linear.app/n8n/issue/CAT-2628 (ExpressionError propagation — resolved)
- https://linear.app/n8n/issue/CAT-2527 (Dual-engine CI — resolved)

## Test plan

- [x] `pnpm --filter n8n-workflow test` — 4120 tests pass (2060 per engine)
- [x] Current engine unaffected
- [x] `pnpm --filter @n8n/expression-runtime test` — 107 tests pass

### DIFF
--- a/packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
+++ b/packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
@@ -21,6 +21,31 @@ function getIvm(): IsolatedVm {
 const BUNDLE_RELATIVE_PATH = path.join('dist', 'bundle', 'runtime.iife.js');
 
 /**
+ * Serialize an error into a transferable metadata object.
+ *
+ * Host-side callbacks (getValueAtPath, etc.) catch errors and return this
+ * sentinel instead of letting the error cross the isolate boundary (which
+ * strips custom class identity and properties). The isolate-side proxy
+ * detects __isError and reconstructs a proper Error to throw.
+ */
+function serializeError(err: unknown): Record<string, unknown> {
+	if (err instanceof Error) {
+		const extra: Record<string, unknown> = {};
+		for (const key of Object.keys(err)) {
+			extra[key] = (err as unknown as Record<string, unknown>)[key];
+		}
+		return {
+			__isError: true,
+			name: err.name,
+			message: err.message,
+			stack: err.stack,
+			extra,
+		};
+	}
+	return { __isError: true, name: 'Error', message: String(err), extra: {} };
+}
+
+/**
  * Read the runtime IIFE bundle by walking up from `__dirname` until
  * `dist/bundle/runtime.iife.js` is found.
  *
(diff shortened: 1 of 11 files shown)


packages/workflow/test/expression.test.ts
```diff
@@ -100,22 +100,29 @@ describe('Expression', () => {
 			expect(evaluate('={{new Object()}}')).toEqual(new Object());
 
 			expect(evaluate('={{new Array()}}')).toEqual([]);
-			expect(evaluate('={{new Int8Array()}}')).toEqual(new Int8Array());
```
despairblue, line 103: Can't use instance checks, so instead checking differently if these are available in the expression engine.

packages/@n8n/expression-runtime/src/runtime/reset.ts
```diff
@@ -72,6 +72,7 @@ export function resetDataProxies(timezone?: string): void {
 	globalThis.__data.$prevNode = createDeepLazyProxy(['$prevNode']);
 	globalThis.__data.$data = createDeepLazyProxy(['$data']);
 	globalThis.__data.$env = createDeepLazyProxy(['$env']);
+	globalThis.__data.process = createDeepLazyProxy(['process']);
```
despairblue, line 75: Only allows this subset:
https://github.com/n8n-io/n8n/blob/455b64dc1a0b717233629799cc28e6c33a1c0d06/packages/workflow/src/expression.ts#L522-L535

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
ivov, line 1: Curious why we aren't adding `throwIfErrorSentinel` to the `has` trap? As in, the host callback might throw in an `in` check.

ivov, line 1: `__getArrayElement` calls are now wrapped with `serializeError` so theymight return sentinels and we aren't checking for sentinels on the isolate side? We'll end up with a sentinel getting assigned?

```ts
const element = globalThis.__getArrayElement.applySync(null, [path, index], { ... });
if (element && typeof element === 'object' && element.__isArray) { ... }
else if (element && typeof element === 'object' && element.__isObject) { ... }
else {
    arrTarget[arrProp] = element;
}
```

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
```diff
@@ -36,8 +82,36 @@ export function getProxyPath(obj: object): string[] | undefined {
  * @param basePath - Current path in object tree (e.g., ['$json', 'user'])
  * @returns Proxy object with lazy loading behavior
  */
-export function createDeepLazyProxy(basePath: string[] = []): any {
+export function createDeepLazyProxy(basePath: string[] = [], knownKeys?: string[]): any {
+	// Cache for keys fetched from the bridge (root proxies without knownKeys).
+	// Shared between ownKeys and getOwnPropertyDescriptor for consistency.
+	let fetchedKeys: string[] | undefined;
+
+	function resolveKeys(): string[] {
```
ivov, line 90: `resolveKeys` silently swallows is the value happens to be a sentinel?

packages/workflow/src/expression.ts
```diff
@@ -46,6 +46,61 @@ setErrorHandler((error: Error) => {
...
+		const err = error as unknown as Record<string, unknown>;
+		const context = err.context;
+		const options: ExpressionErrorOptions =
+			typeof context === 'object' && context !== null
+				? { ...(context as ExpressionErrorOptions) }
+				: {};
+		if (err.functionality === 'pairedItem') {
+			options.functionality = 'pairedItem';
+		}
+		return new ExpressionError(error.message, options);
```
ivov: We're missing `description` here, it's implicitly spread into options but missing as a regular instance field.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -488,8 +534,38 @@ export class IsolatedVmBridge implements RuntimeBridge {
...
+  try {
+    var __result = (function() {
+      ${code}
+    }).call(__data);
+    return __prepareForTransfer(__result);
+  } catch(e) {
+    if (e && e.__isError) return e;
+    var extra = {};
+    for (var k in e) {
+      if (e.hasOwnProperty(k)) extra[k] = e[k];
```
ivov: Curious why not filter name, msg, stack like on the isolate side?

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
```diff
@@ -151,9 +232,9 @@ export function createDeepLazyProxy(basePath: string[] = []): any {
 			}
 
 			// Handle objects - metadata: { __isObject: true, __keys: string[] }
-			if (value && typeof value === 'object' && value.__isObject) {
-				// Create nested proxy for recursive lazy loading
-				target[prop] = createDeepLazyProxy(path);
+			if (isObjectMetadata(value)) {
+				// Create nested proxy for recursive lazy loading, passing known keys
+				target[prop] = createDeepLazyProxy(path, value.__keys);
```
ivov, line 240: For consistency we should pass the keys when getting array element objects like we're doing here?

ivov (commented):
Couple comments!

packages/@n8n/vitest-config/node.ts
```diff
@@ -1,29 +1,30 @@
...
+/**
+ * Shared test options without the outer defineConfig wrapper.
+ * Use this when you need to spread the config into workspace projects.
+ */
+export const createBaseInlineConfig = (options: InlineConfig = {}): InlineConfig => ({
+	silent: true,
+	globals: true,
+	environment: 'node',
+	reporters: process.env.CI === 'true' ? ['default', 'junit'] : ['default'],
+	outputFile: { junit: './junit.xml' },
```
ivov, line 13: Not super familiar with these test artifacts, but if the output file name is the same for both test runs, the last one overwrites?

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -315,126 +349,138 @@ export class IsolatedVmBridge implements RuntimeBridge {
...
+						__keys: Object.keys(value),
+					};
+				}
 
-			// Primitive value
-			return value;
+				// Primitive value
+				return value;
+			} catch (err) {
+				return serializeError(err);
```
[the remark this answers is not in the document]
despairblue: Fixed in 43d949088a.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: resolved in 43d949088a.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: fixed in 43d949088a.

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: nice catch: 19accc704bcc62bcc4643c70d18feb94e70d8a84

packages/@n8n/vitest-config/node.ts
despairblue: wow, nice catch, those need to stay top level, fixed in ea87a3bb2392c74fceae380f2a56f26f60f5a458

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: no reason, fixed in 62c50918acf3b8a14df1841da29bb5fe7e828ec2

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
despairblue: done in c18c980f2f324316b46051d317da7d7b3c129bbe

packages/workflow/src/expression.ts
despairblue: done in e10655dc1270b2128f55dbb23a546b6b571614dc

packages/workflow/src/expression.ts
```diff
@@ -46,6 +46,54 @@ setErrorHandler((error: Error) => {
...
+
+	// Name-based reconstruction for errors that crossed the isolate boundary
+	if (error instanceof Error && error.name === 'ExpressionExtensionError') {
+		return new ExpressionExtensionError(error.message);
+	}
+	if (error instanceof Error && error.name === 'ExpressionError') {
+		const reconstructed = new ExpressionError(error.message);
+		Object.assign(reconstructed, error);
+		return reconstructed;
+	}
```
ivov, line 91: In the `ExpressionExtensionError` case we don't need to reconstruct? `ExpressionExtensionError` has `description`.

packages/@n8n/expression-runtime/src/runtime/reset.ts
ivov, line 1: `globalThis.$items` is missing the sentinel check

ivov (commented):
Thanks for addressing everything! Two more and we're done. I'm reviewing with a fine-tooth comb as I fear any issues here will be tough to debug. 

CI issue looks unrelated.

packages/workflow/src/expression.ts
despairblue: fixed 7307a5277f46036af87d5a6fa3a012344215892d

packages/@n8n/expression-runtime/src/runtime/reset.ts
despairblue: fixed a4333a37a2a139e78241209f23229e3accb6f85b

ivov (approved):
🚀

PR #27547, opened 2026-03-25
fix(core): Prioritize HTTP status codes for retry logic in `httpRequest` nodes

sinehypernova-0718:
## Summary

Fixed a critical bug in the workflow execution engine where nodes would incorrectly trigger a `Retry on Fail` loop when receiving a successful response if the JSON body contained an `"error": null` field.

The previous logic used a strict `!== undefined` check on the `json.error` property. In JavaScript, `null !== undefined` evaluates to `true`, causing the system to interpret a successful "no error" response as a node failure.

### Technical Changes

The fix updates the failure detection logic in the workflow engine to use an explicit error value check. It ensures that only values that are not `null`, `undefined`, or `false` are interpreted as errors.

### Location: 
`packages/core/src/execution-engine/workflow-execute.ts`

### Logic Change:
**Before:**
```typescript
let nodeFailed = !isEngineRequest(runNodeData) && runNodeData.data?.[0]?.[0]?.json?.error !== undefined;
```
**After:**

```typeScript
const isErrorValue = (v: unknown) => v !== undefined && v !== null && v !== false;
const checkFailure = (data: IRunNodeResponse | EngineRequest) =>
    !isEngineRequest(data) && isErrorValue(data.data?.[0]?.[0]?.json?.error);
```

The updated check ensures that `null`, `undefined`, and `false` are no longer treated as failures, while actual error strings and objects still trigger retries as intended.

---

### Before the Fix (Buggy Behavior)
The following screenshots demonstrate that when the field is renamed to status_info, only one request occurs (Control). However, as soon as the field is named error, the engine triggers multiple unintended retries.

---

**In this test, I changed the field name to `error: null`. Even though the server still sent a `200 OK`, n8n triggered 3 extra retries. This proves the bug is specifically tied to the name of the "error" field.**

Screenshot 2026-03-24 175209

---

**In this test, I renamed the field to `status_info: null`. The node worked perfectly and only sent one request. This proves that n8n handles `null` values correctly when the field name is different.**

Screenshot 2026-03-24 175102

### After the fix

**With the new check `!!` applied, the engine now correctly ignores the `null` value. As shown in the logs, even though the response still contains "error": null, n8n now treats it as a success and only sends one request. The unintended retries are completely resolved.**

Screenshot 2026-03-25 152514

**I have verified that the logic now correctly handles "errors" where the API returns a `200 OK` but includes an error message in the JSON body (e.g., `{ "error": "value" }`).**

Screenshot 2026-03-25 213519

---

### Test Results
```
Test Suites: 1 passed, 1 total
Tests:       117 passed, 117 total
Snapshots:   0 total
Time:        4.077 s, estimated 7 s
```

fixes: #27439
Linear Reference: [GHC-7421](https://linear.app/n8n/issue/GHC-7421)

## Review / Merge checklist

- [x] Tests included. 


packages/core/src/execution-engine/workflow-execute.ts
```diff
@@ -1626,6 +1626,9 @@ export class WorkflowExecute {
 						await hooks.runHook('nodeExecuteBefore', [executionNode.name, taskStartedData]);
 					}
 					let maxTries = 1;
+					const isTransportNode = executionData.node.type.toLowerCase().includes('httprequest');
+					const checkFailure = (data: IRunNodeResponse | EngineRequest) =>
+						!isEngineRequest(data) && !!data.data?.[0]?.[0]?.json?.error && !isTransportNode;
```
Flexicon: Thanks for digging into this! Would you be able to simplify to just the truthiness check and drop the `isTransportNode` logic entirely?

The `!!` coercion already fixes the reported bug - `!!null` is `false`, so `{ "error": null }` won't trigger a retry. Excluding HTTP request nodes from body-error checking is a separate behaviour change (it would silently suppress retries even for truthy errors on those nodes), which feels like it
warrants its own discussion. 🙂 


```suggestion
					const checkFailure = (data: IRunNodeResponse | EngineRequest) =>
						!isEngineRequest(data) && !!data.data?.[0]?.[0]?.json?.error;
```

packages/core/src/execution-engine/__tests__/workflow-execute.test.ts
```diff
@@ -3403,4 +3408,89 @@ describe('WorkflowExecute', () => {
 			expect(result.finished).toBe(true);
 		});
 	});
+
+	describe('Node Retry Logic (Error Coercion)', () => {
```
Flexicon: The tests here reimplement the logic locally rather than testing the production code, so if `workflow-execute.ts` changes down the line, these tests won't catch any regressions.

Suggestion: Would it be possible to add an integration test to the existing WorkflowExecute describe block? The meaningful regression test is: a node with `retryOnFail: true` that returns `{ "error": null }` should only call `runNode` once. The existing test setup in this file should provide a good starting point and pattern to follow.

Flexicon (changes requested):
Thank you for contributing! The changeset is a good start here, but it just needs a bit of tightening up.

sinehypernova-0718:
Hi @Flexicon,

First of all, thank you very much for the review and the helpful feedback!

I’ve updated the PR in commit 0b1eb7c93e86e57997103464f1bd94d534bd1d58 to address your comments:

**Logic Simplification**: I removed the `isTransportNode` logic entirely. The check now relies strictly on the `!!` truthiness coercion of the error field. As you noted, this fixes the reported bug while keeping the scope tight and avoiding side effects for HTTP nodes.

**Integration Test**: I’ve replaced the previous unit test with a proper integration test within the `WorkflowExecute` block in `workflow-execute.test.ts`. It verifies that a node with `retryOnFail: true` that returns `{ "error": null }` is called exactly once.

Edit: I just realized I had left the old unit test block in my previous commit by mistake 😅 Now, it has been removed in the latest commit 3d33f8aae83c704b8135d0c1383e80a0e9cdb7bd

Will shortly update the PR description too :)

Everything should be as expected now. Ready for another look whenever you have a moment! 🚀

packages/core/src/execution-engine/__tests__/workflow-execute.test.ts
```diff
@@ -46,9 +46,11 @@ import {
...
 import assert from 'node:assert';
 
-import * as Helpers from '@test/helpers';
-import { legacyWorkflowExecuteTests, v1WorkflowExecuteTests } from '@test/helpers/constants';
-
+import * as Helpers from '../../../test/helpers';
+import {
+	legacyWorkflowExecuteTests,
+	v1WorkflowExecuteTests,
+} from '../../../test/helpers/constants';
```
Flexicon: Could we also revert these changes? I don't think they are necessary or needed here 🙏🏻

packages/core/src/execution-engine/__tests__/workflow-execute.test.ts
```diff
@@ -80,6 +82,34 @@ describe('WorkflowExecute', () => {
 			.mockImplementation(() => {});
 	});
 
+	it('should not trigger retry when error field is null', async () => {
```
Flexicon: Could we wrap this in a `describe('retryOnFail')` to create a logical place for any future tests around this property logic?

sinehypernova-0718:
Thanks for the suggestion, @Flexicon! 
I’m currently away from my desk, but I’ll jump back in as soon as I’m home to wrap that test in the `describe('retryOnFail')` block and push the update within 2 hours. Talk soon!

Thank You! for understanding 😊

sinehypernova-0718:
Hey @Flexicon! Apologies for the delay! But, now the imports are reverted and the test is wrapped in a `describe('retryOnFail')`. 

Let me know if you need anything else 

Thank You for your patience! 🙏

sinehypernova-0718:
Hi @Flexicon, just a gentle ping on this one if it got missed!
All your requests are addressed, whenever you have a moment during your current review cycle, I'd appreciate a final look. Thanks!

packages/core/src/execution-engine/workflow-execute.ts
```diff
@@ -1626,6 +1626,8 @@ export class WorkflowExecute {
 						await hooks.runHook('nodeExecuteBefore', [executionNode.name, taskStartedData]);
 					}
 					let maxTries = 1;
+					const checkFailure = (data: IRunNodeResponse | EngineRequest) =>
+						!isEngineRequest(data) && !!data.data?.[0]?.[0]?.json?.error;
```
Flexicon: Final request @sinehypernova-0718 sorry for the confusion 🙏🏻 

I've talked with the team and we believe the safest approach here would be to actually go for lowest change possible to ensure we don't introduce unexpected regressions for existing workflows. To that end it would make the most sense if we start by treating anything other than: `null`, `undefined` and `false` as error values.

So in this case something like this would suffice:


```suggestion
                    const isErrorValue = (v: any) => v !== undefined && v !== null && v !== false;
					const checkFailure = (data: IRunNodeResponse | EngineRequest) =>
						!isEngineRequest(data) && isErrorValue(data.data?.[0]?.[0]?.json?.error);
```

sinehypernova-0718: No worries at all, @Flexicon! I completely understand wanting to minimize the regression footprint. Safety first!
I am pushing the changes now...

sinehypernova-0718: I've updated the logic as suggested! I used `unknown` for the helper type to ensure we pass the `'no-explicit-any'` linting rule while keeping the logic exactly as the team requested. 🚀

Flexicon (approved):
Thank you for your contribution! 🎉

sinehypernova-0718:
> Thank you for your contribution! 🎉

Awesome, thanks @Flexicon! I really appreciate it! 🎉🎊

PR #27562, opened 2026-03-25
feat(core): Improve public API middleware to handle different error types

sandra0503:
## Summary

Replaces the inline error handler in the public API router with a structured middleware that handles:

1. **n8n domain errors** (`ResponseError` subclasses like `NotFoundError`, `ForbiddenError`, `BadRequestError`) — responds with the error's own HTTP status code and message
2. **OpenAPI validation errors** (from `express-openapi-validator`) — responds with the validator's status code
3. **Unknown errors** — responds with 500 without leaking internals

This means public API handlers can simply `throw` n8n errors and they'll be automatically mapped to the correct HTTP response, instead of needing per-handler try/catch with manual `res.status().json()` calls.

### What has changed?

The middleware runs **after** the handlers. Handlers that catch errors and call `res.status(...).json(...)` themselves do not go through this middleware, so those endpoints are **not affected** by this change (majority of our operations).

This only affects "unhandled" errors — `ResponseError` subclasses now return the actual status code, and `UnexpectedError`/`OperationalError` become 500 instead of 400.

| Thrown value | Before (master) | After (this branch) |
|---|---|---|
| `ResponseError` (e.g. `NotFoundError`) | 400 | Uses `httpStatusCode` (e.g. 404) |
| `ResponseError` with `httpStatusCode === 400` (e.g. `BadRequestError`) | 400 | 400 (unchanged) |
| `UserError` | 400 | 400 (unchanged) |
| `UnexpectedError` | 400 | 500 + generic "Internal server error" |
| `OperationalError` | 400 | 500 + generic "Internal server error" |
| `HttpError` (express-openapi-validator) | `error.status` or 400 | `error.status` or 400 (unchanged) |
| Plain `Error` / unknown | 400 | 500 + generic "Internal server error" |

### How to test

Any public API call that triggers a domain error (e.g. `GET /api/v1/workflows/nonexistent-id`) should return the correct status code and `{ message }` body.

https://linear.app/n8n/issue/LIGO-384

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/executions/execution.service.ts
```diff
@@ -223,7 +224,7 @@ export class ExecutionService {
 		if (!execution.data.executionData) throw new AbortedExecutionRetryError();
 
 		if (execution.finished) {
-			throw new UnexpectedError('The execution succeeded, so it cannot be retried.');
+			throw new BadRequestError('The execution succeeded, so it cannot be retried.');
```
MarcL: Do we want to do this here? I think we're mixing the HTTP error in the service and we should catch this in the controller/handler instead and rethrow. The service could be called from non-HTTP code.

packages/cli/src/executions/execution.service.ts
```diff
@@ -295,7 +296,7 @@ export class ExecutionService {
 						executionId,
 						nodeName: stack.node.name,
 					});
-					throw new WorkflowOperationError(
+					throw new BadRequestError(
```
MarcL: Same here - I don't think we should introduce a HTTP error response.

packages/cli/src/public-api/public-api-error-response.ts
```diff
@@ -0,0 +1,43 @@
...
+import { HttpError } from 'express-openapi-validator/dist/framework/types';
+import { UnexpectedError, UserError } from 'n8n-workflow';
+
+import { ResponseError } from '@/errors/response-errors/abstract/response.error';
+
+/**
+ * Maps errors from the public API stack to HTTP responses. Used by the
+ * Express error middleware in `index.ts`.
+ */
+export function sendPublicApiErrorResponse(res: Response, error: unknown): void {
```
MarcL: Should `error` be type `Error` instead of `unknown`?

packages/cli/src/public-api/public-api-error-response.ts
```diff
@@ -0,0 +1,43 @@
...
+	if (error instanceof UnexpectedError) {
+		res.status(500).json({
+			message: 'Internal server error',
+		});
+		return;
+	}
+
+	if (error instanceof HttpError) {
+		res.status(error.status || 400).json({
+			message: error instanceof Error ? error.message : 'Bad request',
```
MarcL: Do we need this check here? We already know it's a `HttpError` which derives from `Error` so it will always have a `message` won't it?

packages/cli/src/public-api/public-api-error-response.ts
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if (error instanceof UserError) {
+		res.status(400).json({
+			message: error.message,
+		});
+		return;
+	}
+
+	if (error instanceof UnexpectedError) {
```
MarcL: Can this just fall through to the final `500` response if we've not matched any other errors?

packages/cli/test/integration/public-api/executions.test.ts
```diff
@@ -322,7 +323,7 @@ describe('POST /executions/:id/retry', () => {
 	test('should return 400 when trying to retry a finished execution', async () => {
 		const executionServiceSpy = jest
 			.spyOn(Container.get(ExecutionService), 'retry')
-			.mockRejectedValue(new UnexpectedError('The execution succeeded, so it cannot be retried.'));
+			.mockRejectedValue(new BadRequestError('The execution succeeded, so it cannot be retried.'));
```
MarcL: From my comment in the service code, let's not mix HTTP errors in there and leave this test as is.

packages/cli/src/public-api/index.ts
```diff
@@ -155,18 +156,11 @@ function createApiRouter(
...
-			res.status(error.status || 400).json({
-				message: error.message,
-			});
-		},
-	);
+	const publicApiErrorHandler: ErrorRequestHandler = (error, _req, res, _next) => {
+		sendPublicApiErrorResponse(res, error);
+	};
+
+	apiController.use(publicApiErrorHandler);
```
MarcL, line 168: The middleware you've added doesn't match the expected signature for Express does it?
It should be:

```typescript
function middleware(error: Error, request: express.Request, response: express.Response, next:express.NextFunction)
```

We now have:

```typescript
function sendPublicApiErrorResponse(res: Response, error: unknown)
```

packages/cli/src/public-api/public-api-error-response.ts
MarcL: Just seen the calling code and I think you'll still need the signature to match Express error middleware expectations:

```typescript
function sendPublicApiErrorResponse(error: Error, _request: express.Request, response: express.Response, _next:express.NextFunction) : void
```

packages/cli/src/executions/execution.service.ts
sandra0503: Good point, will handle this in the handler

packages/cli/test/integration/public-api/executions.test.ts
sandra0503: Agree

packages/cli/src/public-api/public-api-error-response.ts
sandra0503: Not needed indeed!

packages/cli/src/public-api/public-api-error-response.ts
sandra0503: Agree, Error instead of unknown is better

packages/cli/src/public-api/v1/public-api-error-response.ts
```diff
@@ -0,0 +1,36 @@
...
+	if (error instanceof HttpError) {
+		res.status(error.status || 400).json({
+			message: error.message,
+		});
+		return;
+	}
+
+	res.status(500).json({
+		message: 'Internal server error',
+	});
```
tomi: Do we log or report these somewhere, e.g. in a separate middleware? I'd expect anything 500 is a bug that should be addressed?

packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
```diff
@@ -217,6 +218,8 @@ export = {
 					return res.status(409).json({ message: error.message });
 				} else if (error instanceof NotFoundError) {
 					return res.status(404).json({ message: error.message });
+				} else if (error instanceof UnexpectedError) {
+					throw new BadRequestError(error.message);
```
tomi: Not sure 400 Bad request is the correct status in this case. Unexpected error should occur only if there's a bug in the implementation. In that case 500 would be more appropriate response, with the error being logged/reported. Or is there a known case where this might happen?

packages/cli/src/public-api/v1/public-api-error-response.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	if (error instanceof UserError) {
+		res.status(400).json({
+			message: error.message,
+		});
+		return;
+	}
+
+	if (error instanceof HttpError) {
+		res.status(error.status || 400).json({
```
alielkhateeb: HttpError has a required status property, so I don't see the case of `400` happening, however, I understand the reason of guarding against no status during runtime, but is `400` the right fallback here?

alielkhateeb (commented):
One suggestion/question that keeps coming up for me:

Could we centralize error **classification** (status + normalized fields) in a shared mapper, while keeping separate serializers so each surface keeps its current behavior?

That would let us:
- keep internal responses enriched in `ResponseHelper.sendErrorResponse` (`code/hint/meta`)
- keep Public API responses minimal in its `ErrorRequestHandler` (`message`)
- remove duplicated status/mapping logic without changing current contracts

I can already see overlap between `sendErrorResponse` and `sendPublicApiErrorResponse`.

Possible shape:
- services throw domain/application errors (not transport-specific response payloads)
- shared mapper converts them into a normalized HTTP error descriptor
- each API surface serializes that descriptor differently:
  - internal API serializes an enriched HTTP error
  - public API serlializes a sanitized/simplified HTTP error

What do you think?

packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts
sandra0503: I would agree, however there is a test expecting 400 with this error message: "The execution succeeded, so it cannot be retried" in this case. I adjusted the test case now

packages/cli/test/integration/executions.controller.test.ts
```diff
@@ -139,7 +139,7 @@ describe('POST /executions/stop', () => {
 		await testServer
 			.authAgentFor(owner)
 			.post(`/executions/${incorrectExecutionId}/stop`)
-			.expect(500);
+			.expect(400);
```
alielkhateeb, line 142: Since status codes are changing (And I would question this in a mostly refactoring PR) A quick question on the status code 400 here: 
Since the execution ID is part of the path, a non-existent ID means the URL itself resolves to nothing. 
Wouldn't that make this a 404 Not Found rather than 400?

I think `incorrectExecutionId` is actually a `nonExistentExecutionId` and I think we should split this test into:
- Test non existent ID => 404
- Test no access => 403?

packages/cli/test/integration/public-api/executions.test.ts
```diff
@@ -336,8 +336,8 @@ describe('POST /executions/:id/retry', () => {
 
 		const response = await authUser1Agent.post(`/executions/${execution.id}/retry`);
 
-		expect(response.statusCode).toBe(400);
-		expect(response.body.message).toBe('The execution succeeded, so it cannot be retried.');
+		expect(response.statusCode).toBe(500);
+		expect(response.body.message).toBe('Internal server error');
```
alielkhateeb: I know this might not be within the scope of this PR but this change led me to question: 
When retrying an already-finished execution, why do we throw an unexpected error? This feels like a well-understood, predictable business case, we do expect (and allow) users to "retry" already finished executions, so I don't see this as an unexpected error, neither do I see it a Bad request 😄 My mental model mapped this to a 409 Conflict, as we're trying to retry something whose state doesn't allow it

alielkhateeb (commented):
I really like those changes, what do you think as the author of these? Do you feel something is off? They look great? 

The couple of comments I have are things to discuss overall as I think our status codes need a revisit, my only worry is that we're changing status codes, is this a "breaking change"? 

Happy to hop on a call and discuss!

packages/cli/test/integration/executions.controller.test.ts
sandra0503: True, should be 404 or 403 as you describe. I will look into this, not sure why it's 400 actually

packages/cli/src/errors/http-error-classifier.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export const HttpErrorKind = {
+	ResponseError: 'responseError',
+	UserError: 'userError',
+	UnexpectedError: 'unexpectedError',
+	HttpError: 'httpError',
+	ServerError: 'serverError',
+} as const;
+
+export type HttpErrorKind = (typeof HttpErrorKind)[keyof typeof HttpErrorKind];
```
alielkhateeb: Which of these is the correct type?

packages/cli/test/integration/executions.controller.test.ts
sandra0503: 400 is now the case because the status code of `UserError` is mapped correctly 

Why we're using 400 in this case is another question but I think for the sake of this PR I would leave it. This would be a refactoring in the controller level that should get addressed separately?

alielkhateeb: I'm just worried of the changed behaviour, if the API returns a different status code, I remember we show some notification on the UI, will it still show up? I once saw "Cannot retry execution because it was successful" or something, I suppose this will break now, if it wouldn't and we're confident this response status code change is harmless then let's ship this 🥳

packages/cli/src/errors/http-error-classifier.ts
sandra0503: The type `HttpErrorKind` becomes "responseError" | "userError" | "unexpectedError" | "httpError" | "serverError"

alielkhateeb: I find it confusing to have a const and a type with the same name and case, shouldn't this be an enum? It will provide us with type-safety AND runtime object to iterate on

packages/cli/test/integration/executions.controller.test.ts
sandra0503: This shouldn't make a difference in the UI! I simulated this case and it shows as expected
Screenshot 2026-03-30 at 14 31 44

packages/cli/test/integration/public-api/executions.test.ts
sandra0503: I changed this to return 409 `ConflictError`, makes more sense now!

packages/cli/src/errors/http-error-classifier.ts
sandra0503: It's one option to write it like this (like in `WsStatusCodes `). But I'm also more used to enum, so happy to change it to an enum

packages/cli/src/public-api/v1/public-api-error-response.ts
sandra0503: I suggest to address this in a follow-up ticket

alielkhateeb (approved):
Amazing work 👏 🔥 
I really like the clear separation between the classifier and the serializers 🚀

PR #27564, opened 2026-03-25
feat(instance-ai): add browser extension for local browser automation

elsmr:
## Summary

Adds a Chrome browser extension and CDP relay infrastructure for local browser automation via the instance AI agent.

### Key changes

- **Browser extension** (`@n8n/mcp-browser-extension`): Chrome extension that connects to n8n via WebSocket, relays CDP commands to browser tabs using lazy-attach (debugger attaches on first command, not on connect)
- **CDP relay protocol** (`@n8n/mcp-browser`): Replaces Playwright session management with a CDP relay that routes commands through the extension using CDP `targetId` 
- **Auto-snapshot & waitForCompletion**: Action tools automatically capture page state after mutations and wait for network/navigation to settle
- **Agent can see new tabs and pop up windows**: `newTabs` property is returned by action tools when relevant.
- **Markdown content extraction**: `browser_content` tool replaces `browser_text` with structured markdown output
- **Gateway status in settings**: Local gateway section shows connection status, host identifier, and tool categories
- **Pending connection approval flow**: Chat UI shows pending gateway approval requests
- **Onboarding instructions**: Setup wizard guidance for gateway and browser extension installation
- **Lazy-attach model**: Debugger only attaches when first CDP command targets a tab, reducing permission prompts and overhead
- **Unified logging**: Consistent log format across `mcp-browser`, `mcp-browser-extension`, and `fs-proxy` with dev debug lines
- **Cleanup**: Removed unused adapters (SafariDriver, WebDriver), old docs, and session manager

### Screenshots

image

image

image

image

image

image

## Related Linear ticket

https://linear.app/n8n/issue/NODE-4728


packages/@n8n/api-types/src/schemas/instance-ai.schema.ts
```diff
@@ -325,10 +325,27 @@ const mcpOneOfInputSchema = z.object({ oneOf: z.array(mcpObjectInputSchema) });
 
 const mcpInputSchema = z.union([mcpObjectInputSchema, mcpAnyOfInputSchema, mcpOneOfInputSchema]);
 
+export const mcpToolAnnotationsSchema = z.object({
+	/** Tool category — used to route tools to the correct sub-agent (e.g. 'browser', 'filesystem') */
+	category: z.string().optional(),
+	/** Default permission level for this tool (n8n-specific) */
+	defaultPermission: z.enum(['allow', 'confirm', 'block']).optional(),
```
dlavrenuek: This will probably be not needed any more when we have permission confirmation logic in n8n instance itself?

packages/@n8n/fs-proxy/src/daemon.ts
```diff
@@ -130,8 +143,13 @@ async function handleConnect(req: http.IncomingMessage, res: http.ServerResponse
...
 
 	if (!isAllowed) {
-		const confirm = daemonOptions.confirmConnect ?? cliConfirmConnect;
-		const approved = await confirm(url);
+		const confirmFn = daemonOptions.confirmConnect ?? cliConfirmConnect;
+		broadcastSSE('pending_approval', {
+			method: daemonOptions.approvalMethod ?? 'cli',
+			url,
+		});
+		const approved = await confirmFn(url);
```
dlavrenuek: This looks like an additional logic for permission confirmation. Is it needed for the browser use functionality here?

packages/@n8n/mcp-browser-extension/package.json
```diff
@@ -0,0 +1,25 @@
...
+		"clean": "rimraf dist .turbo",
+		"build": "vite build --config vite.sw.config.mts && vite build --config vite.ui.config.mts",
+		"typecheck": "tsc --noEmit",
+		"lint": "eslint . --quiet",
+		"lint:fix": "eslint . --fix",
+		"test": "jest",
+		"test:unit": "jest"
+	},
+	"dependencies": {
+		"vue": "catalog:frontend"
```
dlavrenuek, line 18: It looks like currently the extension bundles the whole vue package via node_modules. It should be worth looking into "bundling" as a standalone/single js file later

packages/@n8n/mcp-browser/src/adapters/playwright.ts
```diff
@@ -80,51 +105,72 @@ export class PlaywrightAdapter implements BrowserAdapter {
...
+
+		// Launch the browser and detect early spawn failures (ENOENT, EACCES, etc.)
+		await new Promise<void>((resolve, reject) => {
+			const child = execFile(chromePath, [connectUrl]);
+			const earlyFailTimer = setTimeout(() => resolve(), 2_000);
+			child.on('error', (spawnError: Error) => {
+				clearTimeout(earlyFailTimer);
+				log.error('browser spawn error:', spawnError.message);
+				reject(new BrowserExecutableNotFoundError(`${config.browser} (${spawnError.message})`));
+			});
```
dlavrenuek, line 132: This could be made a bit more reliable because if there is a failure after 2 seconds both resolve and reject will be executed.

packages/@n8n/mcp-browser-extension/src/ui/App.vue
```diff
@@ -0,0 +1,740 @@
...
+		await loadTabs();
+	} else {
+		await loadTabs();
+	}
+});
+
+const controlledTabs = computed(() => {
+	// When connected, show all tabs — the extension only tracks controlled ones.
+	// controlledTabIds are CDP targetIds, not Chrome tab IDs, so we just use the count.
+	return tabs.value;
```
dlavrenuek: This will always show all tabs as "controlled tabs", which is not correct. It should show the assigned tabs and new ones that where opened

packages/@n8n/mcp-browser-extension/src/ui/App.vue
```diff
@@ -0,0 +1,740 @@
...
+}
+
+async function refreshStatus(): Promise<void> {
+	const currentStatus = await chrome.runtime.sendMessage({ type: 'getStatus' });
+	log.debug('refreshStatus response:', currentStatus);
+	const statusObj = (currentStatus && typeof currentStatus === 'object' ? currentStatus : {}) as {
+		connected?: boolean;
+		tabIds?: string[];
+	};
+	controlledTabIds.value = statusObj.tabIds ?? [];
```
dlavrenuek, line 130: The `controlledTabIds` are not updated when a new tab is opened, only when you reload the extension view

packages/@n8n/mcp-browser/src/tools/interaction.ts
```diff
@@ -86,20 +86,20 @@ const browserTypeSchema = z
 	})
 	.describe('Type text into an element');
```
dlavrenuek, line 87: I've noticed that the ai agent always does click element (tool) then type (tool). It should not be required to click the element because type can auto focus it (if it doesn't already), right? May be we can just adjust the tool description accordingly

packages/@n8n/mcp-browser/src/tools/session.ts
```diff
@@ -49,23 +34,23 @@ const browserOpenOutputSchema = z.object({
 	),
 });
 
-function browserOpen(
-	sessionManager: SessionManager,
-	toolGroupId: string,
-): ToolDefinition<typeof browserOpenSchema> {
+function browserConnect(
```
dlavrenuek, line 37: Right now the `browser_connect` tool must be called explicitly and the agent does not get notified when the browser disconnected (for example extension or tab was closed). Should we make it simpler and less error prone to use by just allowing the tab/navigation tools to be called directly? These would then connect to the browser internally if it's not already the case.

dlavrenuek: If we would change the behavior, let's do it in a separate PR, because this one is already pretty big

packages/@n8n/mcp-browser/src/server-config.ts
```diff
@@ -70,34 +49,10 @@ export function parseServerOptions(argv = process.argv.slice(2)): ServerOptions
...
 	if (args.browser) cliConfig.defaultBrowser = args.browser as Config['defaultBrowser'];
-	if (args.mode) cliConfig.defaultMode = args.mode as Config['defaultMode'];
-	if (args.headless !== undefined) cliConfig.headless = args.headless;
-	if (args.viewport) cliConfig.viewport = parseViewport(args.viewport);
-	if (args['session-ttl-ms'] !== undefined) cliConfig.sessionTtlMs = args['session-ttl-ms'];
-	if (args['max-sessions'] !== undefined) cliConfig.maxConcurrentSessions = args['max-sessions'];
-	if (args['profiles-dir']) cliConfig.profilesDir = args['profiles-dir'];
 
 	// Merge: env ← cli (CLI wins)
 	const config: Partial<Config> = { ...envConfig, ...cliConfig };
```
dlavrenuek, line 58: I'm wondering why we have a separate config handling in the `mcp-browser` package. Imo the config should be already handled in fs-proxy. Let's not change it in this PR tho

packages/cli/src/modules/instance-ai/filesystem/local-gateway-registry.ts
```diff
@@ -188,6 +196,11 @@ export class LocalGatewayRegistry {
 		state.disconnectTimer = null;
 	}
 
+	/** Update the pending-approval flag on a user's gateway. */
+	setGatewayPendingApproval(userId: string, pending: boolean, method?: 'cli' | 'app'): void {
+		this.getOrCreate(userId).gateway.setPendingApproval(pending, method);
```
dlavrenuek: There is a lot of logic to pass pendingApproval either as `cli` or `app` - if I see it correctly it is only used to display in the chat where the user should approve, and it's not 100% clear if we would have the cli tool and the app or only one option

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAi.api.ts
```diff
@@ -129,6 +127,23 @@ export async function createGatewayLink(
 	);
 }
 
+/**
+ * POST /instance-ai/gateway/pending-approval -> { ok }
+ * Notify the backend that the daemon is waiting for user approval.
+ */
+export async function setGatewayPendingApproval(
```
dlavrenuek: I don't think that we need these api endpoints

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAiSettings.store.ts
```diff
@@ -244,6 +256,31 @@ export const useInstanceAiSettingsStore = defineStore('instanceAiSettings', () =
...
+				data = {};
+			}
+			isPendingApproval.value = true;
+			approvalMethod.value = data.method ?? 'cli';
+			if (justApprovedTimer) {
+				clearTimeout(justApprovedTimer);
+				justApprovedTimer = null;
+			}
+			isJustApproved.value = false;
+			void setGatewayPendingApproval(rootStore.restApiContext, true, approvalMethod.value);
```
dlavrenuek: The approval (notification) flow looks strange to me. It uses the public (unprotected) /events endpoint in the daemon. The n8n frontend then listens to this endpoint to make a request to the n8n backend via rest calls - but the daemon is already connected to the backend and can communicate these events safely over the backend channel without exposing any new api endpoints or leaking data on a public channel.

packages/cli/src/modules/instance-ai/instance-ai-settings.service.ts
```diff
@@ -201,7 +200,8 @@ export class InstanceAiSettingsService {
 			credentialType,
 			credentialName,
 			modelName: prefs.modelName || this.extractModelName(this.config.model),
-			filesystemDisabled: prefs.filesystemDisabled ?? false,
+			localGatewayDisabled:
```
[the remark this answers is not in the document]
elsmr: No need to preserve existing `filesystemDisabled` settings, feature is unreleased

dlavrenuek (commented):
The browser functionality works. In some cases the agent did stop the work or did run into 200k context window overflow error, but this should be handled in instance-ai and not in the gateway anyway. The approval logic looks very strange to me, I would not include it in the current form if not really required for any other functionality

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAiSettings.store.ts
dlavrenuek: Also the communication of daemon -> backend does not require the instance ai to be open, so it works more reliably

packages/@n8n/fs-proxy/src/tools/filesystem/write-file.ts
```diff
@@ -17,7 +17,7 @@ export const writeFileTool: ToolDefinition<typeof inputSchema> = {
 	description:
 		'Create a new file with the given content. Overwrites if the file already exists. Content must not exceed 512 KB.',
 	inputSchema,
-	annotations: { defaultPermission: 'confirm' },
+	annotations: {},
```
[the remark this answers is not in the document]
elsmr: `defaultPermission` annotation was unused, we have a new permission system

dlavrenuek (approved):
Nice 🚀

PR #27565, opened 2026-03-25
feat(core): Add instance setting to disable public chat in Chat Trigger

patdt:
Hi everyone,

we noticed that any authenticated n8n user can currently enable public chat for a workflow. That can lead to workflow data being shared publicly when it shouldn’t be.

This PR adds an instance-level setting to disable the public chat mode of the built-in **Chat Trigger**, following a "better safe than sorry" approach.

When `N8N_DISABLE_PUBLIC_CHAT_TRIGGER=true` is set:

- the **Make Chat Publicly Available** option is no longer shown
- fields that only apply to public chat are hidden as well
- **Make Available in n8n Chat Hub** still remains available
- at runtime, `public = false` is still enforced, even if an older workflow JSON still contains that field

This is now implemented in the following places:

- `packages/@n8n/config/src/configs/chat-trigger.config.ts`
  - defines the instance-level config for `N8N_DISABLE_PUBLIC_CHAT_TRIGGER`
- `packages/@n8n/api-types/src/frontend-settings.ts`
  - exposes `chatTrigger.disablePublicChat` to the frontend
- `packages/frontend/editor-ui/src/app/stores/settings.store.ts`
  - stores the setting in frontend state
- `packages/frontend/editor-ui/src/features/ndv/settings/composables/useNodeSettingsParameters.ts`
  - hides public-chat-only Chat Trigger settings in the editor UI
- `packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts`
  - enforces `public = false` at runtime when the setting is enabled

This replaces the earlier CLI-side metadata filtering approach.

## How to test

1. Build and run n8n normally without the env var.
   - Confirm the Chat Trigger still shows **Make Chat Publicly Available**.
2. Run n8n with:
   - `N8N_DISABLE_PUBLIC_CHAT_TRIGGER=true`
3. Create or edit a workflow with a Chat Trigger.
4. Confirm:
   - **Make Chat Publicly Available** is hidden
   - public-only fields are hidden
   - **Make Available in n8n Chat Hub** is still visible
5. Confirm runtime behavior still blocks public chat when the env var is enabled.

## Checks run

- `pnpm build`
- `pnpm build:n8n`
- `pnpm test` in `packages/cli`
- `pnpm test` in `packages/@n8n/nodes-langchain`
- manual Docker smoke test with `N8N_DISABLE_PUBLIC_CHAT_TRIGGER=true`

## Review / Merge checklist

- [x] Tests included.
- [ ] PR labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if this is considered an urgent fix that needs to be backported)

_No docs update is included in this PR. A docs follow-up can be added separately if maintainers want `N8N_DISABLE_PUBLIC_CHAT_TRIGGER` documented._


packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
```diff
@@ -26,6 +26,8 @@ import { createPage } from './templates';
 import { assertValidLoadPreviousSessionOption } from './types';
 
 const CHAT_TRIGGER_PATH_IDENTIFIER = 'chat';
+
+const isPublicChatTriggerDisabled = () => process.env.N8N_DISABLE_PUBLIC_CHAT_TRIGGER === 'true';
```
jeanpaul: Would be better to keep the environment variables in the config (`@n8n/config`), which makes it easier to see which variables are there. That will also make your test easier, because you don't have to mess around with environment variables in the test.
Also, this function seems duplicated (although simple enough) from `chat-trigger-policy.util.ts`.

packages/cli/src/node-types.ts
```diff
@@ -183,7 +184,7 @@ export class NodeTypes implements INodeTypes {
 				? descriptionCopy.name
 				: `n8n-nodes-base.${descriptionCopy.name}`; // nodes-base nodes are unprefixed
 
-			return descriptionCopy;
+			return applyPublicChatTriggerPolicy(descriptionCopy);
```
jeanpaul: I would probably add support for this in `useNodeSettingsParameters.ts:175` (`shouldDisplayNodeParameter`), or one of the helpers of that. This feels not the right place, as it's impacting things at a very low level

jeanpaul (changes requested):
Hi! Thanks for the PR. Haven't ran the code yet, but had a couple of remarks. Main things:
- Let's keep checking for env-variables in the config
- I'm wondering if the current way of removing the settings that have to deal with `public` should go in `cli`, or should actually belong in the way we normally do display options. I would go with the latter. Please have a look to see if that's feasible.

packages/frontend/editor-ui/src/features/ndv/settings/composables/useNodeSettingsParameters.ts
```diff
@@ -378,12 +429,18 @@ export function useNodeSettingsParameters() {
 			if (path) {
 				const resolvedValues = deepCopy(nodeParameters);
 				set(resolvedValues, path, nodeParams);
-				return nodeHelpers.displayParameter(resolvedValues, parameter, path, node, displayKey);
+				return nodeHelpers.displayParameter(
```
aalises, line 430: This line still passes the original `parameter` instead of `effectiveParameter`. If a Chat Trigger parameter had its `public: [false]` display condition stripped earlier in the function, this code path bypasses that transformation.

```suggestion
			return nodeHelpers.displayParameter(nodeParams, effectiveParameter, '', node, displayKey);
```

patdt: good catch! thanks

aalises (dismissed):
LGTM. If @jeanpaul is alright with it we can merge. There are a couple conflicts to address first though, once rebased you can re-request the approval.

patdt:
@jeanpaul  addressed all review feedback and resolved the merge conflicts

jeanpaul (dismissed):

jeanpaul (dismissed):

patdt:
fixed the config test failure; was missing the chatTrigger parameter in the GlobalConfig test defaults. sorry for the extra round @jeanpaul

jeanpaul:
> fixed the config test failure; was missing the chatTrigger parameter in the GlobalConfig test defaults. sorry for the extra round @jeanpaul

No worries, let's try again :)

jeanpaul (approved):

patdt:
@jeanpaul is this PR still on track to be merged or has it been deprioritized? Just asking so I know whether to keep an eye on it. Thank you very much though!

jeanpaul:
@patdt some tests can be a bit flaky, and I was OOO for a bit. Restarted the failed tests again, so hopefully everything goes in today

PR #27568, opened 2026-03-25
feat(core): Add OTEL unhappy path handling and safe trace exporter

alielkhateeb:
## Summary

https://www.loom.com/share/62b74630914a47d792735c1367f3fe0d

Adds robust error handling to the OpenTelemetry module for cases where span export fails:

- Introduces `SafeTraceExporter`, a wrapper around any `SpanExporter` that catches exceptions during `export`, `shutdown`, and `forceFlush`, logging failures without crashing the process
- Deduplicates export failure logs (only logs the first failure to avoid log spam)
- Parses error types properly in the workflow-end handler
- Adds comprehensive tests for the new exporter and updated module behaviour

## Review / Merge checklist

- [ ] Tests included.

## How to test

You can use - just like in the video - something like Jaeger easily and locally.

1. Create a docker-compose.yml
```yml
services:
  jaeger:
    image: jaegertracing/jaeger:latest
    ports:
      - "16686:16686" # UI
      - "4317:4317"   # OTLP gRPC
      - "4318:4318"   # OTLP HTTP
```

2. Run `docker-compose up -d`
3. Add env variables before running n8n
```
    N8N_OTEL_ENABLED=true \
    N8N_OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318 \
    N8N_OTEL_EXPORTER_OTLP_TRACING_PATH=/v1/traces \
    N8N_OTEL_EXPORTER_SERVICE_NAME=n8n \
```
4. Run n8n normally and visit `http://localhost:16686/` to access Jaeger UI and see the spans
5. For unhappy paths, run `docker-compose down` and start n8n again / execute workflows and check the console


packages/cli/src/modules/otel/otel.service.ts
```diff
@@ -71,4 +81,31 @@ export class OtelService {
...
+	}
+
+	private async checkEndpointReachability(url: string): Promise<void> {
+		try {
+			// HEAD is used for a cheap connectivity check (no request/response body).
+			// OTLP endpoints are POST-only, so this will often return 4xx, but any
+			// HTTP response means the server is reachable. We only catch network errors.
+			await fetch(url, {
+				method: 'HEAD',
+				signal: AbortSignal.timeout(2_000),
```
geemanjs: We should perhaps make this timeout configurable? Its super annoying that the otel sdk doesn't allow you to check whether the exporter is listening. I would assume that even if this "fails" we might still end up "connected" and sending stuff 🤯

packages/cli/src/modules/otel/__tests__/workflow-end.handler.test.ts
```diff
@@ -0,0 +1,59 @@
+import type { WorkflowExecuteAfterContext } from '@n8n/decorators';
+import type { Span } from '@opentelemetry/api';
+import { mock } from 'jest-mock-extended';
+
+import { WorkflowEndHandler } from '../handlers/workflow-end.handler';
+import { ATTR } from '../otel.constants';
+import { SpanRegistry } from '../span-registry';
+
+describe('WorkflowEndHandler', () => {
```
geemanjs: I personally think we should do this test with integration testing - mostly because the "behaviour of the engine" is a part of if this is successful or not. 

To me these tests validate mapping from a->b but I don't gain the confidence that when a workflow is actually retried I get a span for it.

packages/cli/src/modules/otel/safe-trace-exporter.ts
```diff
@@ -0,0 +1,67 @@
+import type { Logger } from '@n8n/backend-common';
+import { ExportResultCode } from '@opentelemetry/core';
+import type { ReadableSpan, SpanExporter } from '@opentelemetry/sdk-trace-base';
+
+type LogFailureAction = 'export' | 'shutdown' | 'forceFlush';
+
+export class SafeTraceExporter implements SpanExporter {
```
geemanjs: I wonder if we need this class at all. This implements the requirements in the ticket but I wonder if we should be doing them exactly like the ticket 😆 having done some more research.. 

OTEL sort of has a "thing" about "never propagating up and breaking your app" - but right now it is "silent" about it

We can plug into the diagnostics at WARN level and I think we will get the results we want without us having to write/maintain any code 👀 
https://opentelemetry.io/docs/languages/js/getting-started/nodejs/#troubleshooting

```
import { diag, DiagConsoleLogger, DiagLogLevel } from '@opentelemetry/api';                                                                                                                                                                              
diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.WARN);
// We might need to implement a custom logger to plug it into the n8n logger.
```

packages/cli/src/modules/otel/handlers/workflow-end.handler.ts
```diff
@@ -1,27 +1,33 @@
-import { SpanStatusCode } from '@opentelemetry/api';
 import type { WorkflowExecuteAfterContext } from '@n8n/decorators';
+import { Service } from '@n8n/di';
+import { SpanStatusCode } from '@opentelemetry/api';
 
 import { ATTR } from '../otel.constants';
 import type { SpanHandler } from './interfaces';
 import type { SpanRegistry } from '../span-registry';
 
+@Service()
```
geemanjs, line 9: I personally don't think these should be services and injected - I think they should be "hard coded" in the OTEL Service itself "for now". 

For me Service's should be used where:
1. We might want to change the implementation for something different
2. There is a case for the class being reused elsewhere - which I don't think there is here

Happy to be challenged on this thinking though 👀

packages/cli/src/modules/otel/n8n-instrumentation.ts
```diff
@@ -1,43 +1,65 @@
...
-			ctx,
-			this.spans,
-			trace.getTracer(TRACER_NAME),
+		this.executeLifecycleHandler(
+			() => this.lifecycleHandlers.workflowExecuteAfter.handle(ctx, this.spans, this.tracer),
+			'workflowExecuteAfter',
 		);
 	}
+
+	private executeLifecycleHandler(handler: () => void, event: keyof OtelLifecycleHandlers): void {
```
geemanjs: I'm not a big fan of pushing the events into a set here as we have no cleanup mechanism for it. 

This feels like a good case for an abstract class maybe?

```jsx
export abstract class SpanHandler<T> {
	handle(ctx: T, spans: SpanRegistry, tracer: Tracer) {
		try {
			this.handleAndThrow()
		}  catch (err) {
			logger.log(err)
		}

	}

	abstract handleAndThrow(ctx: T, spans: SpanRegistry, tracer: Tracer): void;
}
```

packages/cli/src/modules/otel/__tests__/workflow-end.handler.test.ts
alielkhateeb: Valid. I think this whole file can be deleted and we can just confirm the behavior in integration tests. modified here d2570cc6faa385dc9998b9c521b8c78a72b3b4ab

packages/cli/src/modules/otel/otel.service.ts
alielkhateeb: Yes unfortunately because the whole purpose here is to "notify" via the console log, but we don't want to prevent later successful calls.

Extracted the timeout configuration to `startupConnectivityTimeoutMs` in config, here 71cc6b105cea458857115c0f5358b5a84e508c75

packages/cli/src/modules/otel/safe-trace-exporter.ts
alielkhateeb: Oh nice, I will check if we can implement the one-time logging behaviour through that and remove the class if we can do that without it.

alielkhateeb: 5f1b0bb7dcd2ea2a185d1f17100f15b6341e7429

packages/cli/src/modules/otel/handlers/workflow-end.handler.ts
alielkhateeb: I see where you’re coming from, but yes, I’d slightly push back on this 😂

Since `N8nInstrumentation` is already DI-managed, creating these dependencies manually with `new` would introduce a mixed pattern of DI and manual dependencies. We’re already injecting things like the logger as well, so keeping this in DI feels more consistent to me, and avoids falling into a "We use DI, but sometimes not" scenario 😄 

However, given that consistency is the only benefit from my side, I don’t think either approach is objectively wrong, I’d lean toward DI unless there’s a strong reason not to. If you feel strongly about keeping this hard-coded for now, I have no major concerns with that either

packages/cli/src/modules/otel/n8n-instrumentation.ts
alielkhateeb: I will think about this now, but the set is quite bounded assuming you're talking about loggedFailureEvents, this will not grow more than 2 elements in size so it's quite safe

alielkhateeb: I think an abstract class is a bit of an overkill now, I simplified the signature e7c7ca5e0dd1a335f9cb65fe8624febb405e85d6 .. happy to amend of course

Can you elaborate what the benefit of introducing an abstract class here? I believe your 2 points (pushing to a set & introducing an abstract class) are not related to each other, correct?

geemanjs (approved):

packages/cli/src/modules/otel/handlers/workflow-end.handler.ts
geemanjs: Where my head was going in this case was we'd eventually "outsource" the creation of the `SpanHandlers` to a factory which would look at config and create the 'SpanHandler's you wanted / didn't want based on the config. 

But YAGNI rules probably apply here - I have an opinion but like you I don't hold it strongly here 😄

packages/cli/src/modules/otel/n8n-instrumentation.ts
geemanjs: I did totally mix concerns here 😆 . 

Abstract class..
From my side I find the "wrapping" of the "function" call makes the code look ugly. Is that a good reason to change it? Probably not.

I come from a Java background and given a lot of the code is written in a Object oriented type model here I personally think the "responsibility" here for capturing and doing nothing with the exception is the SpanHandler. Nothing else needs to know about that given we are happy for this to silently fail.  

Again in the java world - if we threw the exception we'd have to declare that on the function signature which feels like leaking implementation detail. If the span fails we don't want anything to care about that... 

Saying this 🤣 I could make an argument for the other way saying the "handling of errors" should always be done by "the caller" - as it's their choice if they want to do something with the error or ignore it. 

Again I don't hold a strong enough view to make us change this 😆

packages/cli/src/modules/otel/handlers/workflow-end.handler.ts
alielkhateeb: Indeed, let's keep it as is now, it's such an easy piece of code to change/adapt to future needs

packages/cli/src/modules/otel/n8n-instrumentation.ts
alielkhateeb: I agree 100% that it was ugly sending a method in the params, this is why I changed it in my latest commit, and implementing the abstract class is a very stright forward process that I don't foresee ever getting complicated to do, it's hard to make a clear decision tbh when we only have 2 type of events, let's proceed and when implementing more features around this I think it will be easier to take sides 😂

PR #27573, opened 2026-03-25
feat(core): Add isolate pooling for VM expression engine

ivov:
## Summary

This PR adds isolate pooling to the VM expression engine, following an **isolate-per-workflow** approach: each workflow execution acquires one isolate at start, holds it for the execution's lifetime, and disposes it on completion. The pool pre-warms replacements in the background to eliminate cold-start latency.

- Pool size matches the execution concurrency limit, so a warm bridge is always available when an execution starts.
- Credential resolution (outside execution context) uses the same acquire/evaluate/release path; if the pool is exhausted it falls back to a cold-start bridge.
- The `Expression` instance is the owner key — `acquireIsolate()`/`releaseIsolate()` are instance methods that use `this` as the WeakMap key, replacing the previous `executionId`-based Map.

## Flow

```mermaid
sequenceDiagram
    participant WE as WorkflowExecute
    participant Expr as Expression (instance)
    participant Eval as ExpressionEvaluator
    participant Pool as IsolatePool
    participant Bridge as IsolatedVmBridge
    participant Isolate as V8 Isolate

    Note over WE,Isolate: Execution start
    WE->>Expr: acquireIsolate()
    Expr->>Eval: acquire(this)
    Eval->>Pool: acquire()
    Note over Pool: Pool size = concurrency limit,<br/>bridge always available
    Pool-->>Eval: bridge
    Pool->>Pool: replenish() in background
    Eval->>Eval: bridgesByOwner.set(expr, bridge)

    Note over WE,Isolate: Expression evaluation (sync, repeats per expression)
    WE->>Expr: renderExpression(expression, data)
    Expr->>Eval: evaluate(expression, data, this, { timezone })
    Eval->>Eval: getBridge(this) → bridgesByOwner.get(expr)
    Eval->>Eval: getTransformedCode(expression) → code cache or Tournament
    Eval->>Bridge: execute(transformedCode, data, { timezone })
    Bridge->>Isolate: script.runSync(wrappedCode)
    Isolate-->>Bridge: result
    Bridge-->>Eval: result
    Eval-->>Expr: result
    Expr-->>WE: result

    Note over WE,Isolate: Execution end (in .finally())
    WE->>Expr: releaseIsolate()
    Expr->>Eval: release(this)
    Eval->>Eval: bridgesByOwner.delete(expr)
    Eval->>Pool: release(bridge)
    Pool->>Bridge: dispose()
    Bridge->>Isolate: isolate.dispose()
    Note over Isolate: V8 heap freed
```

https://linear.app/n8n/issue/CAT-2279

## Review / Merge checklist

- [x] Tests included.


packages/workflow/src/expression.ts
```diff
@@ -535,8 +544,11 @@ export class Expression {
 			}
 
 			try {
+				const rawExecutionId = (data as { $execution?: { id: string } }).$execution?.id;
```
ivov: Tried fixing the legacy type here for accuracy, but changing it ripples through everywhere, so had to settle on asserting.

packages/core/src/execution-engine/workflow-execute.ts
```diff
@@ -1440,6 +1441,10 @@ export class WorkflowExecute {
 			// eslint-disable-next-line complexity
 			const returnPromise = (async () => {
 				try {
+					if (this.additionalData.executionId) {
```
ivov: I suspect this optionality is stale and `executionId` is always present, but too big to change here.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -561,6 +561,6 @@ export class IsolatedVmBridge implements RuntimeBridge {
 	 * @returns true if disposed, false otherwise
 	 */
 	isDisposed(): boolean {
-		return this.disposed;
+		return this.disposed || this.isolate.isDisposed;
```
despairblue, line 678: Does this mean `dispose` can fail on lines 544-547?

packages/@n8n/expression-runtime/src/types/evaluator.ts
```diff
@@ -35,6 +33,13 @@ export interface EvaluatorConfig {
 	 * Maximum number of tournament-transformed expressions to cache (LRU).
 	 */
 	maxCodeCacheSize: number;
+
+	/**
+	 * Number of bridges to pre-warm in the pool.
+	 * Should match the execution concurrency limit so each concurrent execution
```
despairblue: I don't think it must match it, in low through put environments it may be a waste of memory to keep all those isolates around.

packages/workflow/src/expression.ts
```diff
@@ -210,13 +212,11 @@ export class Expression {
 		if (!this.vmEvaluator) {
 			// Dynamic import to avoid loading expression-runtime in browser environments
 			const { ExpressionEvaluator, IsolatedVmBridge } = await import('@n8n/expression-runtime');
-			const bridge = new IsolatedVmBridge({ timeout: options?.timeout ?? 5000 });
-			const DEFAULT_MAX_CODE_CACHE_SIZE = 1024;
-			const parsed = parseInt(process.env.N8N_EXPRESSION_ENGINE_MAX_CODE_CACHE_SIZE ?? '', 10);
-			const maxCodeCacheSize = parsed || DEFAULT_MAX_CODE_CACHE_SIZE;
+			const concurrencyLimit = parseInt(process.env.N8N_CONCURRENCY_PRODUCTION_LIMIT ?? '', 10);
```
despairblue: Should the amount of warm isolated-vm keep around be separately configurable from the concurrency? Because otherwise with a concurrency of five in cloud we would keep five warm isolates around all the time, right?

packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
```diff
@@ -349,3 +363,74 @@ describe('Integration: IsolatedVmBridge error handling', () => {
...
+		await evaluator.initialize();
+	});
+
+	afterAll(async () => {
+		await evaluator.dispose();
+	});
+
+	beforeEach(async () => {
+		// Let any in-flight replenishment from the previous test complete
+		await new Promise((resolve) => setTimeout(resolve, 200));
```
despairblue: Can we expose this on the bridge so we don't have to rely on an arbitrary timeout?

packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
```diff
@@ -349,3 +363,74 @@ describe('Integration: IsolatedVmBridge error handling', () => {
...
+		await evaluator.release(caller);
+	});
+
+	it('should replenish after release so pool recovers', async () => {
+		const caller1 = {};
+		await evaluator.acquire(caller1);
+		await evaluator.release(caller1);
+
+		// Wait for replenishment
+		await new Promise((resolve) => setTimeout(resolve, 100));
```
despairblue: Can we expose this on the bridge so we don't have to rely on an arbitrary timeout?

packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
```diff
@@ -349,3 +363,74 @@ describe('Integration: IsolatedVmBridge error handling', () => {
...
+		const result1 = evaluator.evaluate('{{ $json.a }}', { $json: { a: 'first' } }, caller);
+		const result2 = evaluator.evaluate('{{ $json.b }}', { $json: { b: 'second' } }, caller);
+
+		expect(result1).toBe('first');
+		expect(result2).toBe('second');
+
+		await evaluator.release(caller);
+	});
+
+	it('should replenish after release so pool recovers', async () => {
```
despairblue: The pool also replenishes after the `acquire` right?

packages/cli/src/webhooks/webhook-helpers.ts
```diff
@@ -741,14 +742,8 @@ export async function executeWebhook(
...
-			const evaluatedResponseData = workflow.expression.getComplexParameterValue(
-				workflowStartNode,
-				webhookData.webhookDescription.responseData,
-				executionMode,
-				additionalKeys,
-				undefined,
-				'firstEntryJson',
-			) as string | undefined;
+			const evaluatedResponseData =
+				await context.evaluateComplexWebhookDescriptionExpression<string>('responseData');
```
despairblue:
```suggestion
				await context.evaluateComplexWebhookDescriptionExpression<string>('responseData', undefined, 'firstEntryJson');
```

right?

packages/@n8n/expression-runtime/src/pool/__tests__/isolate-pool.test.ts
```diff
@@ -0,0 +1,122 @@
...
+	it('should throw when pool is exhausted', async () => {
+		const factory = createFactory();
+		const pool = new IsolatePool(factory, 1);
+		await pool.initialize();
+		pool.acquire();
+		expect(() => pool.acquire()).toThrow(PoolExhaustedError);
+		await pool.dispose();
+	});
+
+	it('should dispose bridge on release and replenish', async () => {
```
despairblue:
```suggestion
	it('should should not reuse bridges', async () => {
```

?

packages/@n8n/expression-runtime/src/pool/__tests__/isolate-pool.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+	it('should dispose all bridges on pool disposal', async () => {
+		const factory = createFactory();
+		const pool = new IsolatePool(factory, 3);
+		await pool.initialize();
+		await pool.dispose();
+		const bridges = await Promise.all(
+			factory.mock.results.map((r) => r.value as Promise<RuntimeBridge>),
+		);
+		for (const bridge of bridges) {
```
despairblue, line 72: If the factory was not called at all this test passes and doesn't test anything.

```suggestion
    expect(factory).toHaveBeenCalledTimes(3);
		for (const bridge of bridges) {
```

packages/cli/src/credentials-helper.ts
despairblue, line 225: I think we need to acquire here as well.

despairblue (commented):
Regarding the changes in the packages/cli/src/credentials-helper.ts:

Are these functions called individually? I think (hope) they are mostly called from one entry point, `executeWebhook`, and that entry point is called from the waiting, live and test webhooks module. 
That means we can acquire and release the isolate in the `executeWebhook` function (if we want to minimize changes) to avoid needing to churn through X isolates to handle a single webhook.

I mean this: 8a761ac9311576e8723a2fd27901b5d344270c30 (hide whitespace)

If possible the isolate acquisition should be done by whoever owns (e.g. instantiates) the workflow and it should then also be release there. What do you think?

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
ivov: I don't think so! `dispose()` can't fail on those lines because of this change. If anything this change makes dispose less likely to reach those lines, not more. Before this cahnge, an OOM-killed isolate would still reprot as not disposed, so `dispose` would be called and those `.release()` calls would run against a dead isolate. Now `isDisposed()` catches that case and the pool skips calling `dispose` entirely

packages/@n8n/expression-runtime/src/types/evaluator.ts
ivov: Updated phrasing

packages/workflow/src/expression.ts
ivov: Thanks! This was just a placeholder during WIP

packages/cli/src/webhooks/webhook-helpers.ts
ivov: Good catch! Restored

packages/@n8n/expression-runtime/src/pool/__tests__/isolate-pool.test.ts
ivov: Went with another version here

packages/cli/src/webhooks/webhook-helpers.ts
```diff
@@ -460,469 +460,481 @@ export async function executeWebhook(
...
-		// the default that people know as early as possible (probably already testing phase)
-		// that something does not resolve properly.
-		const errorMessage = `The response mode '${responseMode}' is not valid!`;
-		responseCallback(new UnexpectedError(errorMessage), {});
-		throw new InternalServerError(errorMessage);
-	}
+			responseCode,
+			responseData,
+			checkAllMainOutputs,
+			responsePropertyName,
```
ivov: Pre-evaluating `responsePropertyName`, `responseContentType`, `responseBinaryPropertyName` because `extractWebhookLastNodeResponse` gets called in a `.then()` callback that runs after `executeWebhook` returns and the isolate is released

packages/cli/src/credentials-helper.ts
ivov: Hmm `resolveValue` is only called from `authenticate` which is called from `request-helper-functions` during node execution, which happens inside the acquire/release section in `workflow-execute`

ivov: But, there is still a gap in `credentials-tester.service.ts` wrapping `routingNode.runNode()` 

Plugging that now

shortstacked (approved):

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: > dispose() can't fail on those lines because of this change

I meant do we need this change, because the `dispose` method and fail to set the `disposed` property to `true`. 

> Before this cahnge, an OOM-killed isolate would still reprot as not disposed, so dispose would be called and those .release() calls would run against a dead isolate.

1. I'm assuming calling `.release` would throw instead of being a no-op?
2. If the isolate died we don't have to release anything anymore?

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -140,45 +140,50 @@ export class LiveWebhooks implements IWebhookManager {
...
-		const workflowStartNode = workflow.getNode(webhookData.node);
-
-		if (workflowStartNode === null) {
-			throw new NotFoundError('Could not find node to process webhook.');
-		}
-
-		if (!authAllowlistedNodes.has(workflowStartNode.type)) {
-			sanitizeWebhookRequest(request);
+		await workflow.expression.acquireIsolate();
+		try {
```
despairblue, line 144: Doing this increases the diff, at least if you enable `ignore whitespace`. 
And disregarding that option, isn't it more correct to have this in the helpers instead? Just in case we add another webhook kind?

despairblue (dismissed):
Looks good, just one comment which is a bit opinionated.

packages/cli/src/webhooks/live-webhooks.ts
ivov: Started by centralizing in the helpers, but had to move it up to the live/waiting/test webhook callsites because we need to include `getNodeWebhooks` in the acquire-release bracket

shortstacked (dismissed):

packages/testing/playwright/fixtures/base.ts
```diff
@@ -84,7 +84,7 @@ export const test = base.extend<
 				try {
 					return JSON.parse(raw) as Record<string, string>;
 				} catch {
-					console.warn(`[base.ts] Failed to parse N8N_TEST_ENV: ${raw}`);
+					console.warn('[base.ts] Failed to parse N8N_TEST_ENV');
```
despairblue, line 87: Isn't is useful to have the raw value in the logs?

despairblue (dismissed):

packages/testing/playwright/fixtures/base.ts
ivov: Useful for debugging, but we might end up passing secrets via envs (e.g. license) and leaking them

shortstacked (approved):

PR #27585, opened 2026-03-25
feat(anthropicNode): Add structured output support via output_config.…

danielpodwysocki:
## Summary
This allows forcing the output format of the anthropic message API. eg. I can force it to be a single boolean value.

This helps users such as me who want to do some classification based tasks.

For example, with this I can force the Anthropic API to give me one of three categories of task, eg. "shipping", "production", "backoffice", classify it with a category like "isToil"  as a boolean value, etc. It lets me be sure of the response format so that I can craft better decisions for nodes ahead of this one.

I built&tested it practically and can see the right behavior.

Documenation on the functionality this is adding: https://platform.claude.com/docs/en/build-with-claude/structured-outputs

To test, you would insert into the field this adds a JSON such as:

```
{
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    },
    "coolness_score_percentage": {
      "type": "integer"
    }
  },
  "required": [
    "name",
    "email"
  ],
  "additionalProperties": false
}
```

It would look like this:

image

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/vendors/Anthropic/Anthropic.node.test.ts
+++ b/packages/@n8n/nodes-langchain/nodes/vendors/Anthropic/Anthropic.node.test.ts
@@ -272,6 +272,72 @@ describe('Anthropic Node', () => {
 				enableAnthropicBetas: {},
 			});
 		});
+
+		it('should send output_config and return structured_output when structured output is enabled', async () => {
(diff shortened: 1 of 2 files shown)


Bra-Mo:
This seems incredibly useful. Any way to get this into beta soon?

PR #27589, opened 2026-03-26
chore: Add n8n Claude Code plugin with setup-mcps skill

cstuncsik:
## Summary

Add a project-level Claude Code plugin (`n8n`) with an initial `setup-mcps` skill that helps engineers configure commonly used MCP servers (Linear, Notion, Context7, Figma).

### Key Decisions

**Why a plugin instead of a standalone skill?**
To get the `n8n:` namespace prefix for all n8n-specific skills, avoiding name collisions with built-in or personal skills. Claude Code only supports colon-namespaced skills (`n8n:setup-mcps`) through the plugin system — standalone `.claude/skills/` entries cannot be namespaced. This also provides a home for future n8n skills, commands, and agents under the same `n8n:` prefix.

**Why only one skill in this PR?**
Starting with `setup-mcps` to validate the plugin structure and registration. Other existing skills (e.g., `create-issue`, `create-pr`, `content-design`) will be migrated into the plugin in a follow-up PR.

**Why only user and local scope (no project scope)?**
Project scope writes MCP config to `.claude/settings.json`, which is tracked in git. Since MCP credentials are personal (OAuth tokens, API keys), they should not end up in version control. User scope makes MCPs available across all projects; local scope (`settings.local.json`) keeps them project-specific but gitignored.

**Why ask scope per MCP instead of once for all?**
Engineers may want different scopes for different MCPs. For example, Context7 and Figma are useful across all projects (user scope), while Linear or Notion might only be needed for this project (local scope).

### What the skill does
- Checks which MCPs are already configured (matches by URL, not name)
- Presents a multi-select menu of available MCPs
- Asks scope per MCP (user or local — not project, since `.claude/settings.json` is tracked)
- Installs using official recommended commands (`--transport http` with `/mcp` endpoints)

### Other changes
- Removed `.claude/mcp.json` which contained an outdated Linear MCP config
- Registered the plugin in `.claude/settings.json` via `extraKnownMarketplaces` and `enabledPlugins`

### Note
Plugin skill namespacing requires omitting the `name` field from SKILL.md frontmatter due to a [known Claude Code bug](https://github.com/anthropics/claude-code/issues/17271).

https://linear.app/n8n/issue/IAM-451

## Review / Merge checklist

- [ ] Tests included.


Matsuuu (dismissed):
This is great, Thanks Csaba! 🚀

.claude/plugins/n8n/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Plugin Structure
+
+```
+.claude/plugins/n8n/
+├── .claude-plugin/
+│   ├── marketplace.json    # Marketplace manifest
+│   └── plugin.json         # Plugin identity
+├── skills/
+│   └── setup-mcps/
```
DawidMyslak: Maybe just `sample-skill/` here? We probably don't want the list of folders in README to be in sync with the real state? Agents might start updating this file after adding each skill.

.claude/plugins/n8n/README.md
```diff
@@ -0,0 +1,67 @@
...
+│       └── SKILL.md
+└── README.md
+```
+
+## Known Issues
+
+- Plugin skill namespacing requires omitting the `name` field from SKILL.md
```
DawidMyslak, line 65: Ouch. Had one more comment that we have a missing `name` in the skill ;)

DawidMyslak (dismissed):

.claude/plugins/n8n/README.md
cstuncsik: Updated

DawidMyslak (approved):
🫶

PR #27593, opened 2026-03-26
feat: Add AI Gateway support for AI nodes

michael-radency:
## Summary

image

***

AI Gateway lets users run supported AI nodes (OpenAI, Anthropic, Google, etc.) through n8n's managed proxy — no personal API keys required. n8n authenticates to the provider on the user's behalf using a credit-based system, making it easy to get started with AI workflows without any credential setup.

## How it works

When a user enables the gateway on a node credential field, n8n stores a marker (`__aiGatewayManaged: true`) instead of a real credential. At execution time, the credential helper detects this marker and calls `AiGatewayService.getSyntheticCredential()`, which fetches a short-lived JWT from the gateway and returns a synthetic credential pointing to the gateway endpoint. The node code is completely unaware — it behaves as if it received a normal credential.

JWTs are cached per user (up to 500 entries) and proactively refreshed at 90% of their lifetime. Gateway configuration (supported nodes, provider routing, field mappings) is cached locally with a 1-hour TTL, so new providers can be added on the gateway side without n8n code changes.

## What's included

**Backend**
- `AiGatewayService`: synthetic credential generation, JWT lifecycle management, config + usage fetching
- `CredentialsHelper`: intercepts `__aiGatewayManaged` credentials and delegates to the service
- New `/ai/gateway/config` and `/ai/gateway/usage` endpoints

**Frontend**
- Per-node toggle in `NodeCredentials` — enables the gateway for that credential slot and restores the previous real credential when turned off; workflow is auto-saved after toggle
- Remaining credits badge shown inline when the gateway is active
- Node creator tags gateway-compatible nodes with a "Free Credits" badge (when experiment is active)
- `aiGateway` store + `useAiGateway` composable for config/credit state
- Settings page: paginated usage log per request — date, provider, model, input/output tokens, credits deducted
- `ai_gateway` PostHog experiment flag + `settings.aiGateway.enabled` backend flag (both must be active)
- Cleanup on disable: stale gateway-managed credentials are removed from nodes automatically

https://linear.app/n8n/issue/NODE-4692/ai-gateway-p0-n8n-frontendbackend

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
michael-radency, line 1: not sure why github shows so much diff here, it's very simple and clear in vscode

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+ * - `urlField`     — credential field the node reads for the base URL (`url` for most, `host` for googlePalmApi)
+ * - `apiKeyField`  — credential field the node reads for the API key (`apiKey` for most)
+ *
+ * To add a provider: add an entry here + add its type to `AI_GATEWAY_CREDENTIAL_TYPES` in `@n8n/constants`.
+ */
+const GATEWAY_PROVIDER_CONFIG: Record<
+	AiGatewayCredentialType,
+	{ gatewayPath: string; urlField: string; apiKeyField: string }
+> = {
+	googlePalmApi: { gatewayPath: '/v1/gateway/google', urlField: 'host', apiKeyField: 'apiKey' },
```
alexander-gekov: I have it as `/v1/ai-gateway` in my side, i can change mine later, no problems

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+		if (cached && cached.expiresAt > Date.now() + 60_000) {
+			return cached.token;
+		}
+
+		const baseUrl = this.globalConfig.aiAssistant.baseUrl;
+		const licenseCert = await this.license.loadCertStr();
+
+		const response = await fetch(`${baseUrl}/v1/gateway/credentials`, {
+			method: 'POST',
+			headers: { 'Content-Type': 'application/json' },
```
alexander-gekov: for p0 now we do 100 credits per instance, so licenseCert is enough as payload, otherwise we would need x-user-id header as well

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+import { Service } from '@n8n/di';
+import { InstanceSettings } from 'n8n-core';
+import type { ICredentialDataDecryptedObject } from 'n8n-workflow';
+import { UserError } from 'n8n-workflow';
+
+import { License } from '@/license';
+
+type AiGatewayCredentialType = (typeof AI_GATEWAY_CREDENTIAL_TYPES)[number];
+
+interface GatewayTokenResponse {
```
alexander-gekov, line 15: One thing that I haven't done is for jwt token to have TTL, do we need that for p0?

michael-radency: since credentials involved I would say yes

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
alexander-gekov: yeah, it happened to me as well on another PR, idk 🤷‍♂️

packages/@n8n/constants/src/index.ts
```diff
@@ -50,13 +50,17 @@ export const LICENSE_QUOTAS = {
...
 	INSIGHTS_RETENTION_MAX_AGE_DAYS: 'quota:insights:retention:maxAgeDays',
 	INSIGHTS_RETENTION_PRUNE_INTERVAL_DAYS: 'quota:insights:retention:pruneIntervalDays',
 	WORKFLOWS_WITH_EVALUATION_LIMIT: 'quota:evaluations:maxWorkflows',
 } as const;
 
 export const UNLIMITED_LICENSE_QUOTA = -1;
+
+export const AI_GATEWAY_CREDENTIAL_TYPES = ['googlePalmApi'] as const;
+export type AiGatewayCredentialType = (typeof AI_GATEWAY_CREDENTIAL_TYPES)[number];
 export const DEFAULT_WORKFLOW_HISTORY_PRUNE_LIMIT = 24;
```
DawidMyslak: I think we briefly discussed it, should that list come from backend response? Otherwise we will need n8n update to enable new models in the gateway.

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Per-provider routing config.
+ *
+ * - `gatewayPath`  — path appended to `aiAssistant.baseUrl` in the synthetic credential
+ * - `urlField`     — credential field the node reads for the base URL (`url` for most, `host` for googlePalmApi)
+ * - `apiKeyField`  — credential field the node reads for the API key (`apiKey` for most)
+ *
+ * To add a provider: add an entry here + add its type to `AI_GATEWAY_CREDENTIAL_TYPES` in `@n8n/constants`.
+ */
+const GATEWAY_PROVIDER_CONFIG: Record<
```
DawidMyslak: If we want to make it work fully flexibile, without a need for n8n update to add new models, this config could be stored like that on the backend side and returned to each n8n client. Potentially we can skip for P0, but worth noticing.

packages/cli/src/services/frontend.service.ts
```diff
@@ -521,6 +521,16 @@ export class FrontendService {
 			this.settings.aiCredits.setup = !!this.globalConfig.aiAssistant.baseUrl;
 		}
 
+		// TODO: restore `isAiCreditsEnabled &&` before shipping
```
DawidMyslak: Leaving a comment just to have an item to be resolved before merging.

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+	/**
+	 * Returns a synthetic credential for the given type, pointing the node at the gateway
+	 * instead of the real provider. Called from `CredentialsHelper.getDecrypted` when
+	 * `nodeCredentials.__gatewayProxy` is set.
+	 */
+	async getSyntheticCredential(
+		credentialType: string,
+		userId: string,
+	): Promise<ICredentialDataDecryptedObject> {
+		// TODO: enforce license before shipping — throw FeatureNotLicensedError when not licensed
```
DawidMyslak: Leaving a comment just to have an item to be resolved before merging.

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+		if (!this.licenseState.isAiCreditsLicensed()) {
+			console.warn('AI Gateway: license check failed (feat:aiCredits not licensed)');
```
DawidMyslak: Should we use Logger?

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,106 @@
...
+			method: 'POST',
+			headers: { 'Content-Type': 'application/json' },
+			body: JSON.stringify({ licenseCert }),
+		});
+
+		if (!response.ok) {
+			throw new UserError(`Failed to fetch AI Gateway token: HTTP ${response.status}`);
+		}
+
+		const { token, expiresIn } = (await response.json()) as GatewayTokenResponse;
```
DawidMyslak, line 269: Optional: Are they guaranteed to always exist?

We could have an extra check:
```
if (!token || typeof expiresIn !== 'number') {
    throw new UserError('AI Gateway returned an invalid token response.');
}
```

packages/workflow/src/interfaces.ts
```diff
@@ -1327,6 +1327,7 @@ export interface IWebhookFunctions extends FunctionsBaseWithRequiredKeys<'getMod
 export interface INodeCredentialsDetails {
 	id: string | null;
 	name: string;
+	__gatewayProxy?: boolean;
```
DawidMyslak: Just wanted to validate this naming pattern with `__` - was this inspired by something already existing? I guess yes, but let's double check.

On the other side, "gateway" and "proxy" describe the same architectural role here - an intermediary that forwards requests. Combining them in `__gatewayProxy` says the same thing twice.

Looking at how the codebase names things:
- The feature is consistently called "AI Gateway" (AiGatewayService, useAiGateway, isAiGatewayEnabled)
- The UI label says "n8n Managed Credentials"
- The flag is `__gatewayProxy`
- The helper functions say `isCredentialProxied`, `isProxied`

So there are actually multiple different terms floating around for the same concept 🤔. I think this will land in workflow JSON, so let's carefully pick the right thing.

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -38,6 +38,8 @@ import { isEmpty } from '@/app/utils/typesUtils';
 import { getResourcePermissions } from '@n8n/permissions';
```
DawidMyslak, line 38: The `isCredentialExisting`, `areAllCredentialsSet` and `getNodeCredentialIssues` functions potentially all need to recognize `__gatewayProxy` and treat it as "credentials are properly configured."

DawidMyslak (changes requested):

packages/@n8n/constants/src/index.ts
michael-radency: I think so, this is for n8n's consumption, in UI and for nodes, BE does not need to care about that

packages/cli/src/services/ai-gateway.service.ts
michael-radency: main knowledge here is n8n related and is needed inside credentials and nodes, if we change something in credentials, for example, we would have to update BE, if we move config there

packages/workflow/src/interfaces.ts
michael-radency: thanks @DawidMyslak , I did revisit naming to be consistent with our other ai related features(`aiGateway`), and added comment `__aiGatewayManaged` to indicate that it's not suppose to be manually modified

 "n8n Managed Credentials" was updated, maybe not final
image

packages/cli/src/services/ai-gateway.service.ts
DawidMyslak: Ah, when I said backend, I referenced to the ai-service. Imagine we are asked to add a new model node to the AI Gateway, with the current implementation this requires the entire n8n update, I think for some customers this might take a few months? Happy to clarify in the call.

michael-radency: got it, yes, but I would say let's postpone it, until system settle

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -602,163 +658,165 @@ async function onQuickConnectSignIn(credentialTypeName: string) {
...
-					/>
-					<span :class="$style.setupManuallyContainer">
-						<N8nText size="small" :class="$style.setupManuallyOr">
-							{{ i18n.baseText('nodeCredentials.quickConnect.or') }}
-						</N8nText>
-						<N8nLink
-							:class="$style.setupManuallyLink"
-							theme="secondary"
-							underline
+				<template v-else-if="!isAiGatewayManaged(type.name)">
```
DawidMyslak: Not visible in the diff, but above we have `<div v-if="readonly">` which I think is set to true when browsing execution history. We should probably address that, otherwise it'll be broken.

packages/cli/src/controllers/ai.controller.ts
```diff
@@ -256,6 +258,18 @@ export class AiController {
 		}
 	}
 
+	@Get('/gateway/credits')
```
DawidMyslak, line 274: I think we need `@Licensed` decorator. Not sure yet what name/param to use. Maybe @alexander-gekov can confirm.

packages/frontend/editor-ui/src/app/composables/useAiGateway.ts
```diff
@@ -0,0 +1,53 @@
...
+	const creditsQuota = ref<number | undefined>(undefined);
+
+	async function fetchCredits(): Promise<void> {
+		if (!isEnabled.value) return;
+		try {
+			const data = await getGatewayCredits(rootStore.restApiContext);
+			creditsRemaining.value = data.creditsRemaining;
+			creditsQuota.value = data.creditsQuota;
+		} catch {
+			// Keep previous values on error
```
DawidMyslak: Could we log at least so we can track this issue in Sentry? `console.error` should be tracked

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,143 @@
...
+		if (!baseUrl) {
+			throw new UserError('AI Gateway is not configured. Set the AI assistant base URL.');
+		}
+
+		const licenseCert = await this.license.loadCertStr();
+		const response = await fetch(`${baseUrl}/v1/gateway/credits`, {
+			method: 'POST',
+			headers: { 'Content-Type': 'application/json' },
+			body: JSON.stringify({ licenseCert, instanceId: this.instanceSettings.instanceId, userId }),
+		});
```
DawidMyslak, line 165: The licenseCert is sent in every POST call. Not a bug, but worth confirming the gateway server validates this securely with @alexander-gekov and that sending the full cert on every poll (including after each execution via the `workflowExecutionData` watcher) is intentional vs. using a cached/lighter auth mechanism?

DawidMyslak (changes requested):

packages/cli/src/services/ai-gateway.service.ts
alexander-gekov: hmmm, i see no need to send the licenseCert? we can use the existing jwt we get from /credentials ? that's what i expect

packages/cli/src/controllers/ai.controller.ts
alexander-gekov: I changed to use 'gateway' for now, to be on the same page, we can change later if we decide

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,143 @@
...
+		const data = (await response.json()) as GatewayCreditsResponse;
+		if (typeof data.creditsQuota !== 'number' || typeof data.creditsRemaining !== 'number') {
+			throw new UserError('AI Gateway returned an invalid credits response.');
+		}
+
+		return data;
+	}
+
+	/**
+	 * Returns a cached JWT for `instanceId:userId`, fetching a fresh one from the gateway
```
alexander-gekov, line 231: To keep in mind: from ai-assistant side, I also re-fresh jwt if it is at > 90% of its lifetime.

packages/cli/src/controllers/ai.controller.ts
michael-radency: added TODO for later to consider

packages/cli/src/services/ai-gateway.service.ts
michael-radency: updated to use jwt
```
	/**
	 * Returns the current credits quota and remaining credits for the given user.
	 */
	async getCreditsRemaining(userId: string): Promise<GatewayCreditsResponse> {
		const baseUrl = this.globalConfig.aiAssistant.baseUrl;
		if (!baseUrl) {
			throw new UserError('AI Gateway is not configured. Set the AI assistant base URL.');
		}

		const jwt = await this.getOrFetchToken(userId);
		const response = await fetch(`${baseUrl}/v1/gateway/credits`, {
			method: 'GET',
			headers: { Authorization: `Bearer ${jwt}` },
		});

		if (!response.ok) {
			throw new UserError(`Failed to fetch AI Gateway credits: HTTP ${response.status}`);
		}

		const data = (await response.json()) as GatewayCreditsResponse;
		if (typeof data.creditsQuota !== 'number' || typeof data.creditsRemaining !== 'number') {
			throw new UserError('AI Gateway returned an invalid credits response.');
		}

		return data;
	}
```

packages/cli/src/services/ai-gateway.service.ts
michael-radency: synced in code to refresh at 90%

packages/cli/src/services/ai-gateway.service.ts
alexander-gekov: In the body we can leave only `licenseCert`

the other stuff can be moved in headers, the ai assistant expects these for debugging and monitoring:
`x-user-id`, `x-consumer-id`,`x-sdk-version` (this is the ai-assistant package version, we can link it via the package.json),`x-n8n-version`,`x-instance-id`

packages/cli/src/services/ai-gateway.service.ts
```diff
@@ -0,0 +1,150 @@
...
+	 * Returns a synthetic credential for the given type, pointing the node at the gateway
+	 * instead of the real provider. Called from `CredentialsHelper.getDecrypted` when
+	 * `nodeCredentials.__aiGatewayManaged` is set.
+	 */
+	async getSyntheticCredential(
+		credentialType: string,
+		userId: string,
+	): Promise<ICredentialDataDecryptedObject> {
+		// TODO: enforce license before shipping — throw FeatureNotLicensedError when not licensed
+		if (!this.licenseState.isAiCreditsLicensed()) {
```
alexander-gekov: `isAiCreditsLicensed()` checks license for `feat:aiCredits`, maybe we need two new methods for checking license by `feat:aiGateway` and `quota:aiGatewayCredits` .

I also noticed `feat:aiGateway` is missing from `@n8n/constants/src/index.ts`

packages/cli/src/constants.ts
```diff
@@ -17,13 +17,19 @@ export const CUSTOM_API_CALL_NAME = 'Custom API Call';
...
+export const AI_ASSISTANT_SDK_DIR = dirname(dirname(require.resolve('@n8n_io/ai-assistant-sdk')));
 export const TEMPLATES_DIR = join(CLI_DIR, 'templates');
 export const NODES_BASE_DIR = dirname(require.resolve('n8n-nodes-base'));
 export const EDITOR_UI_DIST_DIR = join(dirname(require.resolve('n8n-editor-ui')), 'dist');
 
 const packageJsonPath = join(CLI_DIR, 'package.json');
+const aiAssistantPackageJsonPath = join(AI_ASSISTANT_SDK_DIR, 'package.json');
 const n8nPackageJson = jsonParse<n8n.PackageJson>(readFileSync(packageJsonPath, 'utf8'));
+const aiAssistantPackageJson = jsonParse<n8n.PackageJson>(
+	readFileSync(aiAssistantPackageJsonPath, 'utf8'),
```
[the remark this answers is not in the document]
michael-radency: this mirrors
`const n8nPackageJson = jsonParse<n8n.PackageJson>(readFileSync(packageJsonPath, 'utf8'));`

packages/cli/src/controllers/ai.controller.ts
DawidMyslak: Could we address it now, it's just one-liner? The `getBuilderCredits` endpoint right below has `@Licensed('feat:aiBuilder')`.

IMHO this endpoint should have `@Licensed('feat:aiGateway')`. Without it, any authenticated user can call the gateway credits endpoint and trigger outbound requests from the server to the external gateway service, even without a license.

packages/frontend/editor-ui/src/app/composables/useAiGateway.ts
```diff
@@ -0,0 +1,53 @@
...
+	const { saveCurrentWorkflow } = useWorkflowSaving({ router });
+
+	const isEnabled = computed(
+		() =>
+			postHogStore.getVariant(AI_GATEWAY_EXPERIMENT.name) === AI_GATEWAY_EXPERIMENT.variant &&
+			settingsStore.isAiGatewayEnabled,
+	);
+
+	const creditsRemaining = ref<number | undefined>(undefined);
+	const creditsQuota = ref<number | undefined>(undefined);
```
DawidMyslak: Just a note, these are created as new `ref()`s inside the function body. Every component that calls `useAiGateway()` gets its own independent copy. No shared memory.

I believe both `AiGatewayToggle.vue` and `NodeCredentials.vue` are accessing different data points. A bit fragile long-term, unless we want `fetchCredits` to be called by each component individually? Small Pinia store could solve it.

packages/cli/src/controllers/ai.controller.ts
michael-radency: @DawidMyslak 
we did add `aiGatewayDev` mode for local development and adding `@Licensed('feat:aiGateway')` would break it, I'll try to get feedback on this, but in general - feature is available only if `aiGateway` is in the license, and all calls gated either by token or license check, so it should not any pose security risks

packages/cli/src/credentials-helper.ts
```diff
@@ -346,6 +348,13 @@ export class CredentialsHelper extends ICredentialsHelper {
 		raw?: boolean,
 		expressionResolveValues?: ICredentialsExpressionResolveValues,
 	): Promise<ICredentialDataDecryptedObject> {
+		if (nodeCredentials.__aiGatewayManaged) {
+			return await this.aiGatewayService.getSyntheticCredential(
+				type,
+				additionalData.userId ?? additionalData.activatedByUserId ?? '',
```
alexander-gekov: I think there are more cases where userId can be undefined, seems like when trigger is through n8n chat it does not get proper userId?

michael-radency: @alexander-gekov 
do you have some concrete examples? If you mean models in chat hub then yes, it's not working right now, it would require separate UI and there are issues with streaming

but it works if you expose agent through Chat Trigger node

image
image

alexander-gekov: I have the issue when:
* Chat Trigger node
* Do NOT touch model select in OpenAI as that will get /credentials with userId
* Send message in chat -> (if first time) it will try to fetch credentials and pass userId: "" (empty string) and ai gateway service will return 400 error
image
image

packages/frontend/editor-ui/src/features/shared/nodeCreator/components/NodeCreator.vue
```diff
@@ -113,6 +114,10 @@ function onDrop(event: DragEvent) {
 	}
 }
 
+onMounted(() => {
+	void useAiGatewayStore().fetchConfig();
```
DawidMyslak: This calls the store's `fetchConfig()` directly, bypassing the `isEnabled` guard (present in the `useAiGateway` composable). I think unlicensed users will get a 403?

packages/cli/src/credentials-helper.ts
DawidMyslak: What if we have a brand new workflow, not activated yet, can `activatedByUserId` be empty?

packages/frontend/editor-ui/src/app/stores/aiGateway.store.ts
```diff
@@ -0,0 +1,96 @@
...
+	const fetchError = ref<Error | null>(null);
+
+	async function fetchConfig(): Promise<void> {
+		if (config.value !== null) return;
+		try {
+			config.value = await getGatewayConfig(rootStore.restApiContext);
+			fetchError.value = null;
+		} catch (error) {
+			fetchError.value = toError(error);
+			console.error('[aiGatewayStore] Failed to fetch gateway config:', error);
```
DawidMyslak: We have four `console.error` calls in this file, but `fetchError` is already being set for consumers of the store. In the rest of the codebase we typically surface errors through toasts or let components handle them via the stored error state. Do we need them for dev debugging?

packages/frontend/editor-ui/src/features/ai/gateway/views/SettingsAiGatewayView.vue
```diff
@@ -0,0 +1,218 @@
...
+
+	tbody tr:last-child td {
+		border-bottom: none;
+	}
+}
+
+.empty {
+	text-align: center;
+	color: var(--color--text--tint-2);
+	padding: var(--spacing--xl) !important;
```
DawidMyslak: Nit: could we avoid the `!important` here with a more specific selector?

packages/frontend/editor-ui/src/features/ai/gateway/views/SettingsAiGatewayView.vue
```diff
@@ -0,0 +1,218 @@
...
+}
+
+function providerLabel(provider: string): string {
+	const map: Record<string, string> = {
+		openai: 'OpenAI',
+		anthropic: 'Anthropic',
+		google: 'Google',
+	};
+	return map[provider.toLowerCase()] ?? provider;
+}
```
DawidMyslak: Hardcoded provider label map will need manual updates per new provider.

packages/frontend/editor-ui/src/features/ai/gateway/views/SettingsAiGatewayView.vue
```diff
@@ -0,0 +1,218 @@
...
+	} finally {
+		isLoading.value = false;
+	}
+}
+
+async function refresh(): Promise<void> {
+	await load();
+}
+
+async function loadMore(): Promise<void> {
```
DawidMyslak, line 48: Rapid "Load More" clicks could increment offset multiple times. A quick `if (isLoading.value) return` guard would prevent that 👍.

packages/frontend/editor-ui/src/features/ai/gateway/views/SettingsAiGatewayView.vue
```diff
@@ -0,0 +1,218 @@
...
+				<N8nButton
+					:label="i18n.baseText('settings.n8nGateway.usage.refresh')"
+					icon="refresh-cw"
+					type="secondary"
+					:loading="isLoading"
+					@click="refresh"
+				/>
+			</div>
+
+			<div :class="$style.tableWrapper">
```
DawidMyslak, line 86: Nit: I think in the future we might need alternative approach. With "load more" and with huge amount of data, the virtual list will be needed. Or we can do a trivial 1/2/3 (per-page) pagination and limited rendering. No need to address now.

packages/@n8n/api-types/src/dto/ai/ai-gateway-config-response.dto.ts
```diff
@@ -0,0 +1,11 @@
+export interface AiGatewayProviderConfigEntry {
+	gatewayPath: string;
+	urlField: string;
+	apiKeyField: string;
+}
+
+export interface AiGatewayConfigDto {
```
DawidMyslak: `AiGatewayConfigDto` is a plain interface while other DTOs use Zod - worth aligning for consistency?

packages/frontend/editor-ui/src/features/credentials/components/NodeCredentials.vue
```diff
@@ -297,6 +301,23 @@ onMounted(() => {
 	});
 
 	ndvEventBus.on('credential.createNew', onCreateAndAssignNewCredential);
+
+	void aiGateway.fetchConfig();
```
DawidMyslak, line 305: `NodeCredentials.vue` and `NodeCreator.vue` - Both call `fetchConfig()` on mount, but inconsistently, `NodeCredentials` goes through the composable (guarded), while `NodeCreator` hits the store directly (unguarded). Worth aligning on one approach, ideally a single init point for licensed users.

DawidMyslak (changes requested):
A few more comments from me @michael-radency - I think we're very close, maybe we have a mismatch between the store and the composable usage.

packages/cli/src/credentials-helper.ts
michael-radency: @DawidMyslak , @alexander-gekov 
resolved in https://github.com/n8n-io/n8n/pull/27593/commits/5ef9f7edd8d0d017508a93221c4314c0d3bf783d

DawidMyslak (dismissed):
LGTM 👍 👏

alexander-gekov (dismissed):
Apart from OpenAI node comment, lgtm! Cheers!

DawidMyslak (approved):
Confirming that https://github.com/n8n-io/n8n/pull/27593/changes/f51fb3b15b09796f9d72a2e7a0d69f0763955956 looks good 👌

PR #27598, opened 2026-03-26
ci: Use track-specific npm dist-tags on publish

Matsuuu:
## Summary

When publishing packages to npm, only set the `latest` dist-tag for `stable` track releases. Other tracks now publish under their own tag:

- `stable` → `latest`
- `beta` → `beta`
- `v1` (and other version-named tracks) → `release-v1` (prefixed to avoid npm rejecting semver-like tag names)

Previously, all non-`n8n` packages were published without an explicit tag, which defaulted to `latest` regardless of track.

## Review / Merge checklist

- [ ] Tests included.


Joffcom (approved):

tomi (approved):
🚀

PR #27609, opened 2026-03-26
feat: GET /role-mapping-rule endpoint

BGZStephen:
## Summary

This PR adds a list endpoint for role mapping rules.

IAM-396

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/provisioning.ee/role-mapping-rule.controller.ee.ts
```diff
@@ -14,6 +21,20 @@ export class RoleMappingRuleController {
...
 	) {}
 
+	@Get('/')
+	@GlobalScope('roleMappingRule:list')
+	async list(
+		_req: AuthenticatedRequest,
+		res: Response,
+		@Query query: ListRoleMappingRuleQueryDto,
+	): Promise<RoleMappingRuleListResponse | Response> {
+		if (!this.licenseState.isProvisioningLicensed()) {
```
afitzek, line 31: Do we have a decorator for the license check?

afitzek (dismissed):
Looks good to me, one comment, but I'm not sure if we have such a decorator.

packages/cli/src/modules/provisioning.ee/role-mapping-rule.controller.ee.ts
BGZStephen: I actually think we need to decide where this all happens! The modules have liacense and feature checks, so it does make me wonder if they are strictly needed in here, :shrug:

afitzek: You are right, the module would not be loaded, so we can probably skip it here.

BGZStephen: Yeah, assuming we clearly isolate this, and sub-code within a module should be available for the entire module, this would work

afitzek (approved):

PR #27624, opened 2026-03-26
fix: Ensure monotonic message timestamps for agent (no-changelog)

yehorkardash:
### Why message history looked shuffled

Several messages in one agent turn could end up with the same `createdAt` (the same `Date.now()` millisecond) or with timestamps that did not follow append order. Thread history loaded from storage is ordered by `createdAt` (Postgres/SQLite also use a `seq` tiebreaker). Duplicate or inconsistent times still produce unstable ordering for anything that sorts by time only. `InMemoryMemory` used the stored timestamp for filtering and replay; pagination with `before` / `limit` could slice the wrong boundary—messages appeared to reorder between turns or after suspend/resume.

### Fix: monotonic timestamps in `AgentMessageList`

- Track `lastCreatedAt` while building the list.
- For `input` and `response` messages: set `createdAt` to `max(hint, lastCreatedAt + 1)`, where `hint` is an existing message time or `Date.now()`.
- For `history` (DB-loaded): keep timestamps exact; advance `lastCreatedAt` to the max so new live messages always sort strictly later (handles clock skew and prior monotonic runs).
- `deserialize()` recomputes `lastCreatedAt` from restored messages so continuation after checkpoint stays ordered.

Persistence (`InMemoryMemory`, Postgres, SQLite) stores the message-owned `createdAt` so reloads match the list. Design notes: `packages/@n8n/agents/docs/agent-runtime-architecture.md` (Design decisions: Monotonic `createdAt` for persisted order).

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](https://github.com/n8n-io/n8n/blob/master/.github/pull_request_title_conventions.md))
- [x] Docs updated in-repo (`agent-runtime-architecture.md`) for this behavior
- [x] Tests included
- [ ] PR labeled for backport if urgent


packages/@n8n/agents/src/storage/sqlite-memory.ts
```diff
@@ -205,9 +204,13 @@ export class SqliteMemory implements BuiltMemory {
 
 		return result.rows
 			.map((row) => {
-				const msg = parseJsonSafe(row.content as string) as AgentMessage & { id?: string };
+				const msg = parseJsonSafe(row.content as string) as AgentMessage & {
+					id?: string;
+					createdAt?: Date;
+				};
```
bjorger: Any particular reason why `AgentDbMessage` is not used here?

yehorkardash: Good catch

bjorger (approved):

bjorger (approved):

PR #27636, opened 2026-03-26
refactor(core): Extract CommunityPackagesLifecycleService from controller

sandra0503:
## Summary
- Extracts shared install/update/uninstall logic from `CommunityPackagesController` into a new `CommunityPackagesLifecycleService`
- Prepares for reuse by the Public API in a follow-up PR

## How to test
Installing, updating and uninstalling of community packages should still work

## Related Linear ticket
https://linear.app/n8n/issue/LIGO-391

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
+++ b/packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
@@ -1,0 +1,345 @@
+import { Service } from '@n8n/di';
+import {
+	RESPONSE_ERROR_MESSAGES,
+	STARTER_TEMPLATE_NAME,
+	UNKNOWN_FAILURE_REASON,
+} from '@/constants';
+import type { UserLike } from '@/events/maps/relay.event-map';
+import { EventService } from '@/events/event.service';
+import { BadRequestError } from '@/errors/response-errors/bad-request.error';
+import { InternalServerError } from '@/errors/response-errors/internal-server.error';
+import { NotFoundError } from '@/errors/response-errors/not-found.error';
+import { Push } from '@/push';
+import { InstanceSettings } from 'n8n-core';
+import { valid } from 'semver';
+import type { PublicInstalledPackage } from 'n8n-workflow';
+
+import { CommunityNodeTypesService } from './community-node-types.service';
+import { CommunityPackagesService } from './community-packages.service';
+import type { CommunityPackages } from './community-packages.types';
+import type { InstalledPackages } from './installed-packages.entity';
+import { executeNpmCommand, isNpmExecErrorWithStdout } from './npm-utils';
+
+const {
+	PACKAGE_NOT_INSTALLED,
+	PACKAGE_NAME_NOT_PROVIDED,
+	PACKAGE_VERSION_NOT_FOUND,
+	PACKAGE_DOES_NOT_CONTAIN_NODES,
+	PACKAGE_NOT_FOUND,
+} = RESPONSE_ERROR_MESSAGES;
+
+const isCommunityPackageInstallClientError = (error: Error) =>
+	[PACKAGE_VERSION_NOT_FOUND, PACKAGE_DOES_NOT_CONTAIN_NODES, PACKAGE_NOT_FOUND].some(
+		(msg) => typeof error.message === 'string' && error.message.includes(msg),
+	);
+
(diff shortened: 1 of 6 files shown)


packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
```diff
@@ -0,0 +1,345 @@
...
+			if (error instanceof Error && error.cause instanceof Error) {
+				message += `\nCause: ${error.cause.message}`;
+			}
+
+			const clientError =
+				error instanceof Error ? isCommunityPackageInstallClientError(error) : false;
+			throw new (clientError ? BadRequestError : InternalServerError)(message);
+		}
+
+		if (!hasLoaded) this.communityPackagesService.removePackageFromMissingList(name);
```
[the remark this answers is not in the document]
sandra0503: Same as before

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
```diff
@@ -0,0 +1,345 @@
...
+export class CommunityPackagesLifecycleService {
+	constructor(
+		private readonly push: Push,
+		private readonly communityPackagesService: CommunityPackagesService,
+		private readonly eventService: EventService,
+		private readonly communityNodeTypesService: CommunityNodeTypesService,
+		private readonly instanceSettings: InstanceSettings,
+	) {}
+
+	async listInstalledPackagesHydrated(): Promise<PublicInstalledPackage[] | InstalledPackages[]> {
```
konstantintieber: Nit naming: the old controller method used to be not include the `Hydrated` in the method name. I gotta admit that I don't know what i tmeans.

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
```diff
@@ -0,0 +1,345 @@
...
+				hydratedPackages = this.communityPackagesService.matchMissingPackages(hydratedPackages);
+			}
+		} catch {
+			// Ignore errors when matching missing packages
+		}
+
+		return hydratedPackages;
+	}
+
+	async installWithSideEffects(
```
konstantintieber: What does the `withSideEffects` refer to?

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
```diff
@@ -0,0 +1,345 @@
...
+		let pendingUpdates: CommunityPackages.AvailableUpdates | undefined;
+
+		try {
+			await executeNpmCommand(['outdated', '--json'], {
+				doNotHandleError: true,
+				cwd: this.instanceSettings.nodesDownloadDir,
+			});
+		} catch (error) {
+			if (isNpmExecErrorWithStdout(error) && error.code === 1) {
+				pendingUpdates = JSON.parse(error.stdout) as CommunityPackages.AvailableUpdates;
```
geemanjs: Two things on this one:
1. I think we usually parse json with a `jsonParse` util somewhere. 
2. Should we wrap this in a try catch given the `stdout` is coming from the output of a failed exec command

geemanjs (dismissed):
I see the aim here is mostly just to extract what was already present in the controller - I wonder if we should make it better or raise tickets for it to be made better in future 🤔?

ireneea (dismissed):

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
sandra0503: Have to admit, this was Claude being extra precise. It means that the package gets enriched with additional data (e.g. `updateAvailable` coming from the npm registry).
I am fine with using a more simple name

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
sandra0503: Same here, we can use a more simple name (side effects only describes sending telemetry events after installing)

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
sandra0503: I can address this

konstantintieber (approved):

PR #27637, opened 2026-03-26
feat(core): Enable community package management via Public API

sandra0503:
## Summary
- Adds four Public API endpoints for community package management: POST (install), GET (list), PATCH (update), DELETE (uninstall)
- Permissions are checked via API key scopes  communityPackage:{install,list,update,uninstall}` 
- Added OpenAPI specs for all operations
- Delegates to shared community package service
- 
https://www.loom.com/share/223fcc44f6e9433eafe2d3829c48d9eb

## Related Linear ticket
https://linear.app/n8n/issue/LIGO-391

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/public-api/types.ts
+++ b/packages/cli/src/public-api/types.ts
@@ -279,6 +279,17 @@ export declare namespace DataTableRequest {
 }
 
 // ----------------------------------
+//           /community-packages
+// ----------------------------------
+
+export declare namespace CommunityPackageRequest {
+	type Install = AuthenticatedRequest<{}, {}, { name: string; version?: string }>;
+	type List = AuthenticatedRequest;
+	type Update = AuthenticatedRequest<{ packageName: string }, {}, { version?: string }>;
+	type Uninstall = AuthenticatedRequest<{ packageName: string }>;
+}
+
+// ----------------------------------
 //           /audit
 // ----------------------------------
 
--- a/packages/cli/src/public-api/types.ts
+++ b/packages/cli/src/public-api/types.ts
@@ -285,8 +285,8 @@ export declare namespace DataTableRequest {
 export declare namespace CommunityPackageRequest {
 	type Install = AuthenticatedRequest<{}, {}, { name: string; version?: string }>;
 	type List = AuthenticatedRequest;
-	type Update = AuthenticatedRequest<{ packageName: string }, {}, { version?: string }>;
-	type Uninstall = AuthenticatedRequest<{ packageName: string }>;
+	type Update = AuthenticatedRequest<{ name: string }, {}, { version?: string }>;
+	type Uninstall = AuthenticatedRequest<{ name: string }>;
 }
 
 // ----------------------------------
--- a/packages/cli/src/public-api/v1/handlers/community-packages/__tests__/community-packages.handler.test.ts
+++ b/packages/cli/src/public-api/v1/handlers/community-packages/__tests__/community-packages.handler.test.ts
@@ -5,9 +5,10 @@ import type { Response } from 'express';
 import { RESPONSE_ERROR_MESSAGES } from '@/constants';
 import { BadRequestError } from '@/errors/response-errors/bad-request.error';
 import { NotFoundError } from '@/errors/response-errors/not-found.error';
+import type { InstalledPackages } from '@/modules/community-packages/installed-packages.entity';
 import { CommunityPackagesLifecycleService } from '@/modules/community-packages/community-packages.lifecycle.service';
 import * as middlewares from '@/public-api/v1/shared/middlewares/global.middleware';
-import type { InstalledPackages } from '@/modules/community-packages/installed-packages.entity';
+import { mapToCommunityPackage, mapToCommunityPackageList } from '../community-packages.mapper';
 import { mock } from 'jest-mock-extended';
 
 const mockMiddleware = jest.fn(async (_req: unknown, _res: unknown, next: unknown) =>
@@ -64,7 +65,7 @@ describe('CommunityPackages Handler', () => {
 				mockUser,
 				'publicApi',
 			);
-			expect(mockResponse.json).toHaveBeenCalledWith(mockInstalledPackage);
+			expect(mockResponse.json).toHaveBeenCalledWith(mapToCommunityPackage(mockInstalledPackage));
 		});
 
 		it('should return 400 when name is missing', async () => {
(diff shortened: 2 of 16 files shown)


packages/cli/src/public-api/v1/handlers/community-packages/community-packages.handler.ts
```diff
@@ -0,0 +1,101 @@
...
+		apiKeyHasScope('communityPackage:install'),
+		async (
+			req: AuthenticatedRequest<Record<string, never>, unknown, { name: string; version?: string }>,
+			res: express.Response,
+		): Promise<express.Response> => {
+			const lifecycle = Container.get(CommunityPackagesLifecycleService);
+
+			try {
+				const installedPackage = await lifecycle.install(
+					{ name: req.body.name, version: req.body.version, verify: false },
```
geemanjs, line 26: What does this verify flag do?

packages/cli/src/public-api/v1/handlers/community-packages/community-packages.handler.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	getInstalledPackages: [
+		apiKeyHasScope('communityPackage:list'),
+		async (
+			_req: AuthenticatedRequest<Record<string, never>>,
+			res: express.Response,
+		): Promise<express.Response> => {
+			const lifecycle = Container.get(CommunityPackagesLifecycleService);
+			const packages = await lifecycle.listInstalledPackages();
+			return res.json(packages);
```
geemanjs: I'm unsure if we have a pattern here - but I feel like we should do explicit mapping from internal service -> public api response? (makes any changes deliberate) rather than automatically mapping whatever we return from the service? Means if someone changes the return of `lifecycle.listInstalledPackages` it will break on the typescript mapping and we'll know it's a breaking api change.

geemanjs: This applies to all handlers by the looks of it.

packages/cli/src/public-api/types.ts
```diff
@@ -278,6 +278,17 @@ export declare namespace DataTableRequest {
...
 
+// ----------------------------------
+//           /community-packages
+// ----------------------------------
+
+export declare namespace CommunityPackageRequest {
+	type Install = AuthenticatedRequest<{}, {}, { name: string; version?: string }>;
+	type List = AuthenticatedRequest;
+	type Update = AuthenticatedRequest<{ packageName: string }, {}, { version?: string }>;
+	type Uninstall = AuthenticatedRequest<{ packageName: string }>;
```
geemanjs: Given we use `name` in the `install` maybe we should do the same here?

packages/cli/test/integration/public-api/community-packages.test.ts
```diff
@@ -0,0 +1,267 @@
+jest.mock('@/modules/community-packages/npm-utils', () => ({
```
geemanjs, line 1: A question on this file - I'm usually a big advocate of integration like testing (which this is).. But, given the public api is a "thin slice" on top of the "lifecycle service" which then calls the `CommunityPackageService` - a part of me feels like we should mock out the `lifecycle service`

I could definitely make an argument for both sides here - but keen to hear your thoughts?

packages/cli/test/integration/public-api/community-packages.test.ts
```diff
@@ -0,0 +1,267 @@
...
+		const apiKey = await addApiKey(ownerUser, { scopes });
+		ownerUser.apiKeys = [apiKey];
+		owner = ownerUser;
+	});
+
+	describe('GET /community-packages', () => {
+		it('should return 401 without API key', async () => {
+			const response = await testServer.publicApiAgentWithoutApiKey().get('/community-packages');
+			expect(response.status).toBe(401);
+		});
```
geemanjs, line 68: I think it'd be good to add another test with an api key that doesn't have the right scope associated with it.

packages/cli/test/integration/public-api/community-packages.test.ts
sandra0503: I see what you mean but I think because it's an integration test, it's still good to use actual `lifecycle service` here. This way changes targeted for internal api wouldn't go unnoticed

packages/cli/src/public-api/v1/handlers/community-packages/community-packages.handler.ts
sandra0503: If enabled it only allows to install "verified" community nodes. We as n8n are curating the list of verified/vetted commnunity nodes
In the UI we also allow unverified community nodes when ticking the box

Screenshot 2026-03-31 at 13 26 02

packages/cli/src/public-api/v1/handlers/community-packages/community-packages.handler.ts
sandra0503: Sounds like a good idea, will add

packages/cli/src/public-api/v1/handlers/community-packages/community-packages.handler.ts
```diff
@@ -0,0 +1,98 @@
...
+		): Promise<express.Response> => {
+			const lifecycle = Container.get(CommunityPackagesLifecycleService);
+
+			try {
+				const installedPackage = await lifecycle.install(
+					{ name: req.body.name, version: req.body.version, verify: false },
+					req.user,
+					'publicApi',
+				);
+				return res.json(mapToCommunityPackage(installedPackage));
```
geemanjs, line 30: Can i suggest we do this with zod schemas - like in here?
https://www.notion.so/n8n/Request-Response-2265b6e0c94f8032a12ce711f5478a60#2265b6e0c94f804581f5c4d68e9d5773

sandra0503: I added a schema, which gets parsed in the mapper, let me know if that's how you would do it :)

geemanjs (approved):

packages/cli/src/public-api/v1/handlers/community-packages/__tests__/community-packages.handler.test.ts
```diff
@@ -0,0 +1,204 @@
...
+const handler = require('../community-packages.handler');
+
+describe('CommunityPackages Handler', () => {
+	let mockLifecycle: jest.Mocked<CommunityPackagesLifecycleService>;
+	let mockResponse: Partial<Response>;
+
+	const mockUser = { id: 'test-user-id', role: { slug: 'global:owner' } };
+
+	const mockInstalledPackage = mock<InstalledPackages>({
+		packageName: 'n8n-nodes-test',
```
geemanjs, line 28: You'll hate me for this one - should we call this `name`?

sandra0503: This is inline with internal APi's `InstalledPackages` type, but we could change it to `name` in `CommunityPackageResponse`?

geemanjs: Lets keep it like this! Merge it!

PR #27655, opened 2026-03-27
feat(core): Move id extraction logic into resolver (no-changelog)

phyllis-noester:
## Summary
This PR is part 3 of the Slack dynamic credentials implementation
  (IAM-211).

  - Removed rawBody dependency in the extractor:
  SlackSignatureExtractor no longer requires the "Raw Body" option to
  be enabled on the webhook node. Instead of reading from binary data,
   it reconstructs the URL-encoded form body from the already-parsed
  json.body object using URLSearchParams. This matches how the spike
  worked and avoids adding friction for users configuring the webhook.
  - Moved identity parsing into the resolver: Previously the extractor
   parsed user_id/team_id from the Slack payload and set the user ID
  as the context identity. This logic has been moved into
  SlackSignatureIdentifier (the resolver), centralizing all identity
  resolution and signature verification in one place. The extractor
  now passes the reconstructed raw body as the identity, and the
  resolver is responsible for verifying the Slack signature and
  deriving the storage key.

  ## How to test:

  1. Add the dynamic-creds-test Slack app to a channel
  2. Call the /test-get-emails slash command (app configured at
  https://api.slack.com/apps/A0AGW5T570C/general)
  3. The related n8n workflow should execute and return results — no
  "Raw Body" setting needed on the webhook node https://test-iam211-slack-cred-rs.stage-app.n8n.cloud/workflow/p4dN8dHH35Jm6FmJ

 https://linear.app/n8n/issue/IAM-211

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/__tests__/slack-signature-extractor.test.ts
+++ b/packages/cli/src/modules/dynamic-credentials.ee/context-establishment-hooks/__tests__/slack-signature-extractor.test.ts
@@ -15,7 +15,6 @@ function createSlackTriggerItem(
 			headers: {
 				'x-slack-request-timestamp': '1234567890',
 				'x-slack-signature': 'v0=somesignature',
-				'content-type': 'application/x-www-form-urlencoded',
 				...overrides.headers,
 			},
 			body: overrides.body ?? {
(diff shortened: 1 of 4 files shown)


afitzek (approved):
Nice. 🚀

PR #27665, opened 2026-03-27
fix(core): Fix `/healthz` endpoint when using `N8N_PATH`

### LINKED ISSUE #26930: Wrong healthz location and query when setting N8N_PATH
### Bug Description

When starting n8n >= 2.9.4 (see PR #26129) with `N8N_PATH=/n8n/` and a reverse-proxy setup, healthz is no longer at localhost/healthz and calls from the editor to healthz are made to /n8n//healthz which is a 404 page returning HTTP 200 OK.

Image

### To Reproduce

Example:
1. You want n8n at https://domain.com/n8n/ because you already have something else at the root of your domain, so you set up reverse proxy and deploy n8n 2.9.4 or greater via docker with N8N_PATH set to /n8n/ so that calls are made to /n8n/ instead of /.

### Expected behavior

`N8N_PATH` should not change backend paths and the frontend shouldn't make calls to healthz with double slash. When setting `N8N_PATH=/n8n/`, healthz should remain at `localhost/healthz` as it did until 2.9.4 and the editor should make calls to `/n8n/healthz` instead of /n8n//healthz.

### Debug Info

# Debug info

## core

- n8nVersion: 2.10.3
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/145.0.0.0 safari/537.36 edg/145.0.0.0
- isTouchDevice: false

Generated at: 2026-03-12T08:05:42.543Z

### Operating System

Debian 13 + Docker

### n8n Version

2.10.3

### Node.js Version

24.13.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

MiloradFilipovic:
## Summary
Fixes an issue introduced in #26129 that caused `/healtz` endpoint to point to wrong location when using `N8N_PATH` env variable and reverse proxy setup.|

This PR introduces two methods that resolve health endpoint location separately:
- `resolveBackendHealthEndpointPath()` - always returns the bare health endpoint (e.g. `/healthz`)
- `resolveFrontendHealthEndpointPath()` - prepends `N8N_PATH` with `path.posix.join()` for proper slash normalization

Fixes ADO-5016
Fixes [26930](https://github.com/n8n-io/n8n/issues/26930)

## Review / Merge checklist

- [ ] Tests included. 


despairblue (dismissed):
Approved, just one thing we could consider additionally:

In useBackendStatus should we check the json instead of the 404 page which also returns a 200 status code, which is what was masking this bug?

https://github.com/n8n-io/n8n/blob/e1221d6dfbf1d873d0847990d75014c155557af6/packages/frontend/editor-ui/src/app/composables/useBackendStatus.ts#L30-L30

Something like:

```ts
  const data = await response.json() as { status: string };                                                                                                                                                                                                                                                                                                                 
  return response.ok && data.status === 'ok';     
```

despairblue (approved):

JonnyPower:
this appears to have caused a regression in n8n - blocking ability to modify/save workflows - in my deployment my healthz endpoint still returns plain text `ok` - not a json object with a `status` key - https://github.com/n8n-io/n8n/issues/28850#issuecomment-4298787379

I don't see the change to the real service here to make it return `json`

packages/cli/src/abstract-server.ts
JonnyPower, line 150: ah it was already there.. wonder what's going on

EDIT: I had an nginx healthz endpoint shadowing n8n's. using the N8N_ENDPOINT_HEALTH resolved.

PR #27671, opened 2026-03-27
fix: NodeCredentials use correct credential type rather than first

phyllis-noester:
## Summary

When a node has multiple credential types (e.g. Alpha API and Beta API), clicking “Create New Credential” from the Beta dropdown would open the modal showing Alpha’s fields instead.

Root cause: In CredentialEdit.vue, the selectedCredentialType computed ignored the activeId prop (which carries the requested credential type) when requiredCredentials was true. It always fell back to the first credential type from the node definition.

Fix: Added a check for props.activeId before the fallback, so the modal honors whichever credential type was actually requested.

Test: Verifies that when activeId is set to the second credential type, getNewCredentialName is called with that type — not the first one.

https://github.com/user-attachments/assets/37859ae6-c9e8-4dd6-9b81-4bed71d9a924

https://linear.app/n8n/issue/IAM-413/community-issue-nodecredentials-have-internally-only-one

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -143,6 +143,13 @@ const selectedCredentialType = computed(() => {
 	if (selectedCredential.value !== '') {
 		return credentialsStore.getCredentialTypeByName(selectedCredential.value) ?? null;
 	} else if (requiredCredentials.value) {
+		// If a specific credential type was requested via activeId, honor it
+		if (props.activeId) {
+			const requestedType = credentialsStore.getCredentialTypeByName(props.activeId);
```
guillaumejacquart: I see in the `getNodeCredentialForSelectedAuthType` function (used below), we use the nodeType.credentials to find the credential type.
I'm wondering if we should also use the same instead of the whole credentials store. 
Maybe even update the getNodeCredentialForSelectedAuthType instead, to accept the activeId arg or something

guillaumejacquart (commented):
Some code organization comment

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
phyllis-noester: makes sense, refactored it to use activeNodeType

cstuncsik (dismissed):

guillaumejacquart (dismissed):

BGZStephen (dismissed):

packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialEdit.vue
```diff
@@ -143,14 +143,21 @@ const selectedCredentialType = computed(() => {
...
+		// No auth options — fall back to the explicitly requested type or the first credential
+		if (props.activeId) {
+			const nodeCredential = activeNodeType.value?.credentials?.find(
+				(c) => c.name === props.activeId,
+			);
+			if (nodeCredential) {
+				return nodeCredential;
+			}
+		}
+		return activeNodeType.value?.credentials ? activeNodeType.value.credentials[0] : null;
```
guillaumejacquart: how about:
`return activeNodeType.value?.credentials?.[0] ?? null;` for clearer nullish check on the first credentials (instead of truthy).
Also, this makes me think that maybe all that could (should?) be inside the getNodeCredentialForSelectedAuthType function ? (not sure though)

phyllis-noester: makes sense changed the return value! getNodeCredentialForSelectedAuthType is also used in other places and doesn't know about the specific props of this credential but just looks at the properties of the given node type so I'd keep it as is. 
having said that, I think this file could generally be refactored a bit. I didn't touch it now because I just wanted to ship this as a quick fix (that then exploded a bit because i lack some context on creds) 🙈

guillaumejacquart (approved):

PR #27673, opened 2026-03-27
refactor: Upgrade to TypeScript 6.0.2

ivov:
## Summary

What changed in TS 6 | Changes made to adapt
-- | --
`types` defaults to `[]` instead of auto-including all `@types/*` | Added `types: ["node", "jest"]` to codemirror-lang and stylelint-config packages (longer term, these FE packages should stop using Jest!). Switched `@n8n/json-schema-to-zod` to extend `tsconfig.backend.json` which already carries them. Excluded test helper dirs from nodes-base's build tsconfig. Added `tsconfig.scripts.json` to `@n8n/extension-sdk` to cover the orphaned `scripts/` directory that uses Node built-ins.
`rootDir` defaults to the tsconfig dir instead of being inferred from sources | Added explicit `rootDir: "src"` to the three standalone `@n8n/json-schema-to-zod` build configs. Added `rootDir: "."` to ts-jest's inline tsconfig in the root `jest.config.js` to fix TS5011 when ts-jest compiles a single file (e.g. `globalSetup.ts`) in isolation.
Cross-package path aliases fail when `outDir` is set without `rootDir` | Replaced `outDir: "dist"` with `noEmit: true` in `design-system`, `rest-api-client`, `editor-ui`, and `storybook` tsconfigs (all use Vite/tsdown for building, never tsc)
`noUncheckedSideEffectImports` enabled by default under `strict` | Set to `false` in both base `tsconfig.common.json` and `modern/tsconfig.json` (will need to be a separate dedicated effort)
`moduleResolution: node` and `module: umd` are deprecated | Added `ignoreDeprecations: "6.0"` to the two shared base configs and two standalone configs that use these settings (will need to be a separate dedicated effort)
DOM lib updated with new interface members | Edited a polyfill in `editor-ui/__tests__/setup.ts`
`Uint8Array` is now `Uint8Array<ArrayBufferLike>`; `WebSocket.send()` requires `Uint8Array<ArrayBuffer>` | Type cast at the call site in `@n8n/crdt` (will need to be a separate dedicated to propagate through Y.js call sites, best done by folks who know this package)
Property access on `never` is now a type error | Cast `body` in the `default` branch of an exhaustive switch in `log-streaming.controller.ts`

Separate dedicated efforts are tracked here:
- https://linear.app/n8n/issue/CAT-2713
- https://linear.app/n8n/issue/CAT-2714
- https://linear.app/n8n/issue/CAT-2715

## Review / Merge checklist

- [ ] Tests included. 


Matsuuu (dismissed):
🚀 📦

Matsuuu (dismissed):

Matsuuu (approved):

PR #27675, opened 2026-03-27
feat(core): Improve custom auth JSON masking, preserve shape, redact leaf values

guillaumejacquart:
## Summary

Improves the masking of `httpCustomAuth` credentials' JSON field. Previously, the entire JSON value was replaced with a single blanking sentinel, losing the structure. Now:

- **Backend (`redact`)**: parses the JSON and replaces every leaf value with `***`, preserving the key/object structure so users can see what keys exist
- **Backend (`unredact`)**: merges the edited JSON on save/test — leaves still set to `***` are restored from the saved credential, changed values are kept
- **Frontend**: detects the shaped-redacted JSON and displays it as-is in the editor, allowing direct editing without clearing the field first
- **Reload after save**: credential data is re-fetched after saving so the redacted shape is immediately visible

The redaction logic was also refactored for clarity: `redactValues` is now a thin wrapper over `redactSensitiveProperties` (schema-based blanking) and `redactJsonObject` (leaf-level JSON redaction), with extensive tests covering deep nesting, arrays, type changes, key deletion and addition.

### Demo
https://www.loom.com/share/a212bb7d491f4834a72d4146356ce154

https://linear.app/n8n/issue/IAM-408

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/credentials/__tests__/credentials.service.test.ts
+++ b/packages/cli/src/credentials/__tests__/credentials.service.test.ts
@@ -13,8 +13,10 @@ import { CREDENTIAL_ERRORS, CredentialDataError, Credentials, type ErrorReporter
 import {
 	CREDENTIAL_BLANKING_VALUE,
 	CREDENTIAL_EMPTY_VALUE,
+	jsonParse,
 	type ICredentialDataDecryptedObject,
 	type ICredentialType,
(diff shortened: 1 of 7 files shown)


afitzek (approved):
👍

PR #27681, opened 2026-03-27
feat(instance-ai): Add n8n sandbox service as workspace provider

tomi:
## Summary

Adds the n8n sandbox service as a new sandbox provider alongside Daytona and local. This introduces:

- **N8nSandboxClient**: HTTP client for the sandbox service API — handles sandbox lifecycle, streamed command execution, and file operations
- **Mastra adapters**: `N8nSandboxServiceSandbox` (command execution) and `N8nSandboxFilesystem` (file I/O) implement Mastra's workspace interfaces
- **DockerfileStepsBuilder**: Fluent builder for Dockerfile instructions sent at sandbox creation, replacing the previous image-caching approach
- **Configuration**: New `N8N_SANDBOX_SERVICE_URL` / `N8N_SANDBOX_SERVICE_API_KEY` env vars, `n8n-sandbox` provider option, and `httpHeaderAuth` credential support in admin settings
- **Admin UI**: Sandbox provider dropdown now includes "n8n Sandbox Service" with its own credential selector
- **SandboxConfig refactor**: Discriminated union type for provider-specific config fields with proper type narrowing via assert helpers

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/instance-ai/src/workspace/n8n-sandbox-client.ts
```diff
@@ -0,0 +1,604 @@
...
+
+	private async readExecResult(
+		response: Response,
+		request: Pick<N8nSandboxExecRequest, 'onStdout' | 'onStderr'>,
+	): Promise<N8nSandboxExecResult> {
+		if (!response.body) {
+			throw new Error('Sandbox exec response body is not readable');
+		}
+
+		const reader = response.body.getReader();
```
aalises: The stream reader is never released when `processLine` throws (e.g. on an `'error'` event). This can probably leak data

packages/@n8n/instance-ai/src/workspace/n8n-sandbox-image-manager.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class N8nSandboxImageManager {
+	private cachedDockerfile: DockerfileStepsBuilder | null = null;
+
+	getDockerfile(): DockerfileStepsBuilder {
+		if (this.cachedDockerfile) return this.cachedDockerfile;
+
+		this.cachedDockerfile = new DockerfileStepsBuilder()
+			.run(
+				'mkdir -p /home/user/workspace/src /home/user/workspace/chunks /home/user/workspace/node-types',
```
aalises: This path is hardcoded to `/home/user/workspace` but the builder prompt gets its root dynamically via `getWorkspaceRoot(workspace)`. 

I'd maybe extract this as a constant shared with wherever `getWorkspaceRoot` resolves it for it to be consistent (as now it would break if the sandbox service home directory changes one day for some reason)

packages/@n8n/instance-ai/src/workspace/n8n-sandbox-client.ts
tomi: Good catch — wrapped the read loop in a `try/finally` that calls `reader.releaseLock()` so the stream is always released, even when `processLine` throws on an error event. Fixed in b613b19.

packages/@n8n/instance-ai/src/workspace/n8n-sandbox-image-manager.ts
tomi: Agreed — extracted `N8N_SANDBOX_WORKSPACE_ROOT` (derived from `N8N_SANDBOX_HOME` + `WORKSPACE_DIR`) in `sandbox-setup.ts` and updated the image manager to use it. Now there's a single source of truth for the n8n sandbox workspace path. Fixed in b613b19.

packages/@n8n/instance-ai/src/workspace/create-workspace.ts
```diff
@@ -34,6 +66,14 @@ export function createSandbox(config: SandboxConfig): DaytonaSandbox | LocalSand
 		});
 	}
 
+	if (config.provider === 'n8n-sandbox') {
+		return new N8nSandboxServiceSandbox({
+			apiKey: config.apiKey,
+			serviceUrl: config.serviceUrl,
+			timeout: config.timeout ?? 300_000,
```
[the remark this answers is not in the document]
tomi: Valid — the timeout was passed into `N8nSandboxServiceSandbox` but `executeCommand` only used the per-call `options?.timeout`, ignoring the sandbox-level default. Fixed by falling back to `this.options.timeout` when no per-command timeout is provided: `timeoutMs: options?.timeout ?? this.options.timeout`. Fixed in b613b19.

packages/@n8n/instance-ai/src/workspace/n8n-sandbox-client.ts
tomi: I actually rewrote the stream processing to make it more readable

aalises (approved):

PR #27684, opened 2026-03-27
fix(instance-ai): Resolve entity names in tool confirmation messages (no-changelog)

r00gm:
## Summary
- Add optional name parameters (`workflowName`, `folderName`, `credentialName`) to all tools that show confirmation dialogs
- Use `name ?? id` fallback so confirmation messages show human-readable names instead of raw IDs (e.g. `Delete folder "My Reports"?` instead of `Delete folder "l9SUnmxhlarCXMSd"?`)
- Add system prompt instruction telling the agent to always include entity names when calling tools

## Test plan
- [ ] Trigger a destructive action (e.g. delete folder) and verify the confirmation dialog shows the entity name
- [ ] Verify fallback to ID still works if agent doesn't pass the name

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/workflows/delete-workflow.tool.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/delete-workflow.tool.ts
@@ -12,6 +12,10 @@ export function createDeleteWorkflowTool(context: InstanceAiContext) {
 			'Archive a workflow by ID. This is a soft delete that unpublishes the workflow if needed and can be undone later.',
 		inputSchema: z.object({
 			workflowId: z.string().describe('ID of the workflow to archive'),
+			workflowName: z
+				.string()
+				.optional()
+				.describe('Name of the workflow (for confirmation message)'),
 		}),
 		outputSchema: z.object({
 			success: z.boolean(),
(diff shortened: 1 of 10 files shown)


packages/@n8n/instance-ai/src/tools/workflows/delete-workflow.tool.ts
```diff
@@ -33,17 +37,9 @@ export function createDeleteWorkflowTool(context: InstanceAiContext) {
 
 			// State 1: First call — suspend for confirmation (unless always_allow)
 			if (needsApproval && (resumeData === undefined || resumeData === null)) {
-				const workflow = await Promise.resolve(context.workflowService.get(input.workflowId)).catch(
```
aalises, line 36: This would be more reliable but add some more latency/memory... The workflow name the LLM could hallucinate?

I'm leaning a bit more towards adding the deterministic fetch here, I feel like the more we can add something deterministically and don't leave it to the LLMs judgement, the better (unless the performance dip would be big, not sure what this WF entity has but it looks massive)

I'm fine with either way tbf, your call. If we get some reports that the LLM is making up workflow names which I doubt, we can always add that back easily, not a biggie

aalises (approved):

packages/@n8n/instance-ai/src/tools/workflows/delete-workflow.tool.ts
r00gm: in theory i should use the same entities that it uses to send the ids to the tools. But yes, if it hallucinates then we will have to request the entities

PR #27691, opened 2026-03-27
fix(core): Handle invalid percent sequences and equals signs in HTTP response headers

Joffcom:
## Summary

Fixes a `URI malformed` error thrown by the HTTP Request node when a server responds with a `Content-Disposition` header containing a filename with a bare `%` not followed by two hex digits (e.g. `attachment; filename="my_scan_144dpi_75%.pdf"`).

**Root cause:** `parseHeaderParameters()` unconditionally called `decodeURIComponent()` on every header parameter value. A bare `%` that isn't a valid percent-encoded sequence causes `decodeURIComponent` to throw. The error only appeared in **File** and **Autodetect** response format modes (not Text mode, which skips header parsing).

**Changes to `parse-incoming-message.ts`:**
- Strip quotes before attempting to percent-decode, so decoding operates on the actual content rather than the content-with-outer-quotes (more semantically correct)
- Fall back to the raw value when decoding fails, rather than throwing
- Use `indexOf('=')` instead of `split('=')` to correctly parse parameter values that contain `=` characters (e.g. `boundary=----WebKit...==`, `filename="report=final.pdf"`)
- Skip malformed parameters that have no `=` (previously silently injected `{'': 'undefined'}` into the parsed result via `decodeURIComponent(undefined)` coercion)

Backward-compatible: non-standard quoted+percent-encoded filenames (e.g. `filename="screenshot%20(1).png"`) continue to decode correctly.

https://linear.app/n8n/issue/NODE-4618
fixes #26031

## Review / Merge checklist

- [x] Tests included.


DawidMyslak (approved):
Looks good 👍.

One question: should this get release/backport tags? It's a hard crash for any user downloading a file with % in the name, so it seems like a candidate.

Minor note (not a blocker, pre-existing): `contentDisposition.split(';')` on line 90 will still break if a quoted filename contains a literal semicolon (e.g. `attachment; filename="file;name.txt"`). Not in scope here, but worth a follow-up if it ever surfaces.

PR #27698, opened 2026-03-27
feat(core): Add project-level insights permissions and filtering

fabian8n:
## Summary
- Project admins and personal project owners can now view insights scoped to their projects, not just instance owners/admins
- Insights dashboard filters by project via a dropdown, with data scoped to accessible projects
- Extracts `getUserProjectIdsWithScope` to de-duplicate the DB query used by both `userHasScopes` and `InsightsService`
- Moves authorization from controller-level `@GlobalScope` to service-layer `getEffectiveProjectIds`, supporting both global and project-scoped `insights:list` access
- Fixes double-fetch flicker in InsightsDashboard caused by async `onBeforeMount` + watch both firing
https://www.loom.com/share/9450d324ea4247afa414bbf933c0f236

## Changes
| Area | Detail |
|------|--------|
| `check-access.ts` | Extract `getUserProjectIdsWithScope`; `userHasScopes` de-duplicated to call it |
| `project-scopes.ee.ts` | Add `insights:list` to admin/owner scopes; intentionally excluded from editor scopes |
| `insights.service.ts` | `getEffectiveProjectIds` centralizes auth for all 3 service methods |
| `insights-by-period.repository.ts` | `projectId` → `projectIds` (array) for multi-project filtering |
| `insights.store.ts` | `canViewInsights` computed covers global + project-level access, gated on `isInsightsEnabled` |
| `module.descriptor.ts` | Route guard uses `custom` middleware → `canViewInsights` |
| `InsightsDashboard.vue` | Fix double-fetch with `isInitializing` flag; fix `filterFn` (permissions-only); remove debug artifact |
| `rbac.store.ts` | Defensive `?? []` for unloaded project scopes |
| Tests | 4 new `getUserProjectIdsWithScope` tests, 6 auth tests for `getEffectiveProjectIds`, updated controller test mocks |

## Test plan
- [ ] Instance owner sees all insights (no project filter applied)
- [ ] Instance owner can filter by specific project in dropdown
- [ ] Project admin sees insights scoped to their accessible projects
- [ ] Project admin filtering to a specific project works
- [ ] Project editor cannot access insights page (redirected)
- [ ] Personal project owner can see their personal project insights
- [ ] No double-fetch/flicker on initial page load
- [ ] `pnpm typecheck` passes for `packages/cli` and `packages/frontend/editor-ui`
- [ ] All unit tests pass (check-access, insights.service, insights.controller)
- [ ] Docs follow-up ticket created for [docs.n8n.io/insights](https://docs.n8n.io/insights/)

## Related
https://linear.app/n8n/issue/LIGO-358


packages/@n8n/permissions/src/roles/scopes/project-scopes.ee.ts
```diff
@@ -87,8 +88,13 @@ export const PERSONAL_PROJECT_OWNER_SCOPES: Scope[] = [
 	'dataTable:listProject',
 	'dataTable:readRow',
 	'dataTable:writeRow',
+	'insights:list',
```
geemanjs, line 91: Can they get insights for a personal project?

packages/@n8n/permissions/src/roles/scopes/project-scopes.ee.ts
```diff
@@ -87,8 +88,13 @@ export const PERSONAL_PROJECT_OWNER_SCOPES: Scope[] = [
...
 ];
 
+// Note: insights:list is intentionally NOT included here. Editors can read and
+// run workflows but do not have access to project-level analytics. Only project
+// admins (REGULAR_PROJECT_ADMIN_SCOPES / PERSONAL_PROJECT_OWNER_SCOPES) and
+// instance owners/admins (global scope) can view insights.
```
geemanjs, line 97:
```suggestion
```
I don't think this is needed

packages/cli/src/modules/insights/__tests__/insights.controller.test.ts
```diff
@@ -834,7 +850,7 @@ describe('InsightsController', () => {
 					startDate: expect.any(Date),
 					endDate: expect.any(Date),
 					periodUnit: 'day',
-					projectId: 'test-project',
+					projectIds: ['test-project'],
```
geemanjs, line 853: Just to confirm - This is on the internal api? otherwise this is a breaking change we need to think about.

packages/cli/src/modules/insights/__tests__/insights.service.integration.test.ts
```diff
@@ -7,10 +7,22 @@ import {
...
+jest.mock('@n8n/permissions', () => ({
+	...jest.requireActual('@n8n/permissions'),
+	hasGlobalScope: jest.fn().mockReturnValue(true),
+}));
+
+jest.mock('@/permissions.ee/check-access', () => ({
+	userHasScopes: jest.fn().mockResolvedValue(true),
+}));
+
+const anyUser = mock<User>();
```
geemanjs, line 25: Given integration tests are used to test a more "black box" flow - Using mocks here feels like an anti-pattern. 

We should use "correct users" here. 

I'd also expect to see the "global user" and "project access" user testing in here. This is the place we can build real confidence in those checks given the permissions are mocked in both the controller and service tests. 

I think we should have:
global - accessing global
global - accessing project
project - accessing global - denied
project - accessing project with no access - denied
project - accessing project with access - allowed

packages/cli/src/modules/insights/insights.service.ts
```diff
@@ -68,19 +73,56 @@ export class InsightsService {
...
 
+	/**
+	 * Returns the effective project ID filter for insights queries.
+	 *
+	 * - Global `insights:list` scope: returns undefined (no filter — all projects)
+	 * - Project-level `insights:list` scope with specific projectId: returns [projectId]
+	 * - Project-level `insights:list` scope without projectId: returns all accessible project IDs
+	 * - No scope: throws ForbiddenError
+	 */
+	private async getEffectiveProjectIds(
```
geemanjs, line 84: This function is a bit weird. 

It returns a list of project ids the user can see but returns `undefined` which means "all projects"

packages/cli/src/permissions.ee/__tests__/check-access.test.ts
```diff
@@ -14,7 +14,91 @@ import { CredentialsFinderService } from '@/credentials/credentials-finder.servi
...
+		return {
+			innerJoin: jest.fn().mockReturnThis(),
+			where: jest.fn().mockReturnThis(),
+			andWhere: jest.fn().mockReturnThis(),
+			groupBy: jest.fn().mockReturnThis(),
+			having: jest.fn().mockReturnThis(),
+			select: jest.fn().mockReturnThis(),
+			getRawMany: jest.fn().mockResolvedValue([{ id: 'project-1' }]),
+		};
+	}
```
geemanjs, line 32: The fact these tests mock the query builder and validates that the query builder has been called with "specific text" doesn't fill me with confidence. 

A better model would be to query against an actual database instance (sqlite I think we have elsewhere) as the behaviour of "the actual database" and "typeorm" will likely be different. 

Likewise they are incredibly brittle - if we change the order / way we query the database to optimise performance for example - these tests will fail even if the query input / outputs is correct. We want to validate inputs / outputs. Not "I tried calling the database with where project = 123"

packages/frontend/editor-ui/src/app/stores/rbac.store.ts
```diff
@@ -120,7 +120,9 @@ export const useRBACStore = defineStore(STORES.RBAC, () => {
 			scope,
 			{
 				global: globalScopes.value,
-				project: context?.projectId ? scopesByProjectId.value[context.projectId] : [],
+				// scopesByProjectId is only populated for resources in view, so default to []
+				// to avoid a TypeError when checking a project that hasn't been loaded yet.
+				project: context?.projectId ? (scopesByProjectId.value[context.projectId] ?? []) : [],
```
geemanjs, line 125: This does not read easily 😆

geemanjs (commented):
I'm not overly sure on this implementation 👀. I wonder if we should split the "project level" insights from the "global level" insights on the backend - would be keen to here @ireneea's view here 👍. I think this is the pattern we followed on secret providers. 

I also haven't really commented on the FE side - to me it looks a bit messy 👀.

PR #27708, opened 2026-03-28
fix(core): Add missing break statements in filter condition evaluation

### LINKED ISSUE #27693: If node forwards true statement to false branch
### Bug Description

I am using the if node in my workflow which received a statusCode from a website I checked with a previous node. The if node compares the receiving statusCode (200) to the number 200, validates it and follows through with the false branch instead of the true branch.

I already made sure I receive the statusCode as a number and that I compare the number to 200 as number and not as string. It still keeps going after the false branch instead of the true branch.

Image

### To Reproduce

See bug description.

### Expected behavior

The if node should follow the true branch if the statement is true.

### Debug Info

# Debug info

## core

- n8nVersion: 2.13.4
- platform: npm
- nodeJsVersion: 24.14.0
- nodeEnv: undefined
- database: sqlite
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2026-03-27T15:48:32.018Z

### Operating System

Windows VServer

### n8n Version

2.13.4

### Node.js Version

24.14.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

abhishekDeshmukh74:
## Summary

The `executeFilterCondition` function in `packages/workflow/src/node-parameters/filter-parameter.ts` was missing `break` statements in the outer `switch(operator.type)` block. Only the `string` case had a `break`; the `number`, `dateTime`, `boolean`, `array`, and `object` cases all lacked them, causing fall-through between type cases.

While current operations within each case use `return` (preventing fall-through in the happy path), this is a defect because:
- If the inner switch doesn't match an operation, execution falls through to the next type's case block
- For example, a number comparison falling through to the `dateTime` case would hit `if (!left || !right) return false`, incorrectly returning `false` for falsy number values like `0`

### How to test
1. Run the unit tests: `cd packages/workflow && pnpm test test/filter-parameter.test.ts`
2. All 230 tests pass including 8 new tests covering number equality with string coercion

- Fixes https://github.com/n8n-io/n8n/issues/27693
- https://linear.app/n8n/issue/GHC-7487
- https://linear.app/n8n/issue/NODE-4872/community-issue-if-node-forwards-true-statement-to-false-branch

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/workflow/src/node-parameters/filter-parameter.ts
+++ b/packages/workflow/src/node-parameters/filter-parameter.ts
@@ -310,6 +310,8 @@ export function executeFilterCondition(
 				case 'lte':
 					return left <= right;
 			}
+
+			break;
 		}
 		case 'dateTime': {
 			const left = leftValue as DateTime;
@@ -339,6 +341,8 @@ export function executeFilterCondition(
 				case 'beforeOrEquals':
 					return left.toMillis() <= right.toMillis();
 			}
+
+			break;
 		}
 		case 'boolean': {
 			const left = leftValue as boolean;
@@ -358,6 +362,8 @@ export function executeFilterCondition(
 				case 'notEquals':
 					return left !== right;
 			}
+
+			break;
 		}
 		case 'array': {
 			const left = (leftValue ?? []) as unknown[];
@@ -385,6 +391,8 @@ export function executeFilterCondition(
 				case 'notEmpty':
 					return left.length !== 0;
 			}
+
+			break;
 		}
 		case 'object': {
 			const left = leftValue;
@@ -395,6 +403,8 @@ export function executeFilterCondition(
 				case 'notEmpty':
 					return !!left && Object.keys(left).length !== 0;
 			}
+
+			break;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mahdirajaee (commented):
This is a necessary correctness fix. Without `break` statements at the end of each type case block (`number`, `dateTime`, `boolean`, `array`, `object`), execution falls through to the next case in the switch, which means an unrecognized operator on a `number` condition would silently evaluate the `dateTime` logic and so on. The fall-through was masked because most operator strings don't match across types, but it could produce incorrect results in edge cases — particularly with loose type validation where coerced values might accidentally satisfy a downstream comparison. The test coverage added is thorough: the parameterized `number:equals` tests with string coercion under loose validation directly exercise the code path that would have been affected. The strict-validation test confirming the throw is also a good regression guard. Clean fix with minimal risk since these `break` statements only affect the unreachable-under-normal-conditions tail of each inner switch.

BerniWittmann (approved):
Thank you for your contribution 🙌 I took the liberty of adding another test for testing the fall-through when an invalid comaprsion value (`equal` instead of `equals`) has been used

PR #27719, opened 2026-03-29
fix(GraphQL Node): handle non-array errors in response

corazzione:
## Summary

- GraphQL servers can return `errors` as a string instead of an array (e.g. `{\"errors\": \"An unexpected error occurred\"}`)
- The node crashed with `response.errors?.map is not a function` because it assumed `errors` is always an array
- Fix adds an `Array.isArray` check before calling `.map()`, falling back to string or object messages for non-array values

Fixes #27601

## Review / Merge checklist

- [x] Tests included.


mahdirajaee (commented):
Good defensive fix. The GraphQL spec technically requires `errors` to be an array, but real-world servers don't always comply, so guarding against non-array `response.errors` is the right thing to do here. The branching logic is clean — `Array.isArray(errors)` for the standard path, `String(errors)` fallback for the unexpected shape, and constructing a proper `JsonObject` with a `message` key for `NodeApiError` in the non-array case. One thing worth noting: if `errors` happens to be an object (e.g., `{ message: "something" }`) rather than a plain string, `String(errors)` would produce `[object Object]` — it might be worth handling that case with `JSON.stringify` or checking for `.message` directly. That said, this is a clear improvement over the current code which would throw a `TypeError` on `.map()` when `errors` isn't an array. The type cast through `unknown` to `JsonObject` for the array path is a reasonable workaround for the type mismatch.

corazzione:
Closing this PR as requested while reviewing my open pull requests: closing because the linked issue #27601 is already closed.

PR #27722, opened 2026-03-29
fix(Schedule Trigger Node): use deterministic cron second to prevent duplicate registrations

corazzione:
## Summary

- `toCronExpression()` used `randomInt(0, 60)` for the seconds component on every call
- When a workflow was saved while active, re-registration generated a new random second, producing a different cron expression
- `ScheduledTaskManager.registerCron()` builds the dedup key using `expression`, so the old and new keys differ — both crons end up registered, causing the workflow to execute twice per interval
- Fix derives the seconds offset from a SHA-256 hash of `nodeId-intervalIndex`, keeping executions spread across the minute while making expressions stable across saves and restarts
- Existing workflows will get a new (stable) second offset on next activation — this is a one-time shift, same as any first activation

Fixes #27103

## Review / Merge checklist

- [x] Tests included.


mahdirajaee (commented):
Solid fix. Replacing `randomInt(0, 60)` with a SHA-256-based deterministic second derived from the node ID plus interval index is the right approach to prevent duplicate cron registrations on every workflow save. The `deterministicSecond` function is clean — taking only 8 hex characters and modding by 60 gives a uniform-enough distribution for this use case. Good call making the `seed` parameter optional so that `toCronExpression` remains backward-compatible for callers that don't need determinism. One small note: the existing code uses `randomInt(0, 60)` for the `minute` fallback on lines below the change — if determinism matters for the second, it might eventually matter for the minute value too in similar re-registration scenarios, though that's a separate concern. The seed format `${nodeId}-${index}` is a sensible composite key. Clean, minimal change that directly addresses the root cause.

corazzione:
Closing this PR as requested while reviewing my open pull requests: closing because the linked issue #27103 is already closed.

PR #27731, opened 2026-03-30
feat(core): Add instance registry service (no-changelog)

afitzek:
## Summary

Add the core `InstanceRegistryService` for the instance registry module, enabling n8n processes to register themselves, maintain heartbeats, and clean up on shutdown.

- **Backend selection**: Automatically selects Redis storage in distributed mode (`isMultiMain` or queue mode) or in-memory storage for single-instance deployments, using lazy imports to avoid loading Redis dependencies when unnecessary
- **Instance lifecycle**: Generates a unique UUID instance key on startup, registers with the storage backend, and runs a 30-second heartbeat interval that refreshes `lastSeen` while preserving `registeredAt`
- **Graceful shutdown**: Stops the heartbeat timer and unregisters from storage via `@OnShutdown()` on the module, with error-resilient cleanup that never blocks shutdown
- **Public API**: Exposes `getAllInstances()`, `getLocalInstance()`, `getLastKnownState()`, `saveLastKnownState()`, and `storageBackend` for use by the proxy service and reconciliation cycle
- **Redis storage stub**: Adds a placeholder `RedisInstanceStorage` that will be replaced by #27527
- **Log scope**: Registers `instance-registry` in the logging config
- **Tests**: 20 unit tests covering backend selection, UUID format, registration, heartbeat resilience, shutdown safety, and all public API methods
- **REST Controller** (`GET /instance-registry`): Exposes cluster info (registered instances + version mismatch detection) behind `orchestration:read` scope.
- **Settings UI page**: New "Instance Registry" entry under Admin Settings that displays the raw cluster info JSON, accessible via `/settings/instance-registry`.
- **REST API client**: Frontend API helper for fetching cluster info.

Here is a preview of the settings page:

image

- https://linear.app/n8n/issue/IAM-329

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/instance-registry/instance-registry.service.ts
+++ b/packages/cli/src/modules/instance-registry/instance-registry.service.ts
@@ -1,0 +1,128 @@
+import type { InstanceRegistration } from '@n8n/api-types';
+import { Logger } from '@n8n/backend-common';
+import { ExecutionsConfig } from '@n8n/config';
+import { Service } from '@n8n/di';
+import { randomUUID } from 'node:crypto';
+import { InstanceSettings } from 'n8n-core';
+
+import { N8N_VERSION } from '@/constants';
+
+import { REGISTRY_CONSTANTS } from './instance-registry.types';
+import type { InstanceStorage } from './storage/instance-storage.interface';
+
+/**
+ * Core service for instance lifecycle management in the Instance Registry.
+ *
+ * Handles backend selection (Redis vs memory), instance registration,
+ * periodic heartbeat, and graceful shutdown/unregistration.
+ */
+@Service()
+export class InstanceRegistryService {
+	private storage!: InstanceStorage;
+
+	private heartbeatInterval: ReturnType<typeof setInterval> | null = null;
+
+	private readonly instanceKey = randomUUID();
+
+	private registeredAt = 0;
+
+	constructor(
+		private readonly instanceSettings: InstanceSettings,
+		private readonly executionsConfig: ExecutionsConfig,
+		private readonly logger: Logger,
+	) {
+		this.logger = this.logger.scoped('instance-registry');
+	}
+
+	async init() {
+		this.storage = await this.selectStorage();
+		this.registeredAt = Date.now();
+
+		const registration = this.buildRegistration();
+		await this.storage.register(registration);
+		this.startHeartbeat();
+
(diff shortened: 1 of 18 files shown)


packages/cli/src/modules/instance-registry/instance-registry.service.ts
```diff
@@ -0,0 +1,131 @@
...
+		this.startHeartbeat();
+
+		this.logger.info('Instance registered', {
+			instanceKey: this.instanceKey,
+			backend: this.storage.kind,
+			instanceType: this.instanceSettings.instanceType,
+		});
+	}
+
+	async shutdown() {
```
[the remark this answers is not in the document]
afitzek: The shutdown is happening via the module and the process already has a grace period for the shutdown implemented, so there is nothing to be done here.

packages/cli/src/modules/instance-registry/storage/redis-instance-storage.ts
```diff
@@ -31,11 +31,18 @@ export class RedisInstanceStorage implements InstanceStorage {
...
-		await this.upsertRegistration(registration, 'register');
+		await this.upsertRegistration(registration);
 	}
 
 	async heartbeat(registration: InstanceRegistration): Promise<void> {
-		await this.upsertRegistration(registration, 'heartbeat');
+		try {
+			await this.upsertRegistration(registration);
+		} catch (error) {
+			this.logger.warn('Failed to heartbeat instance', {
```
phyllis-noester, line 41: so if one instance stops responding we send warnings until it resolves itself or a human fixes it?

packages/cli/src/modules/instance-registry/instance-registry.service.ts
```diff
@@ -0,0 +1,139 @@
...
+ */
+@Service()
+export class InstanceRegistryService {
+	private storage!: InstanceStorage;
+
+	private heartbeatInterval: ReturnType<typeof setInterval> | null = null;
+
+	private readonly instanceKey = randomUUID();
+
+	private registeredAt = 0;
```
phyllis-noester, line 27: why set this to 0 instead keeping it empty?

phyllis-noester (dismissed):
nice! also tested locally (not in multimain/queue however) 🔥

phyllis-noester (approved):
🚀

PR #27746, opened 2026-03-30
fix(core): Rename data table columns during source control pull

ireneea:
## Summary

Source control pull now issues `ALTER TABLE RENAME COLUMN` when a data table column is renamed in git (same column ID, different name). Previously only metadata was updated, leaving the physical PostgreSQL column with the old name and causing query failures.

https://linear.app/n8n/issue/LIGO-404
closes #27143

## Review / Merge checklist

- [x] Tests included.

<details>
<summary>Implementation plan</summary>

## Root Cause

`source-control-import.service.ee.ts` lines 1406-1429: the column upsert loop handles deleting removed columns and adding new columns, but when an existing column (same ID) has a different name, it only updates metadata via `trx.save()` — never calls `dataTableDDLService.renameColumn()` to issue `ALTER TABLE RENAME COLUMN`.

## Fix

After `trx.save()` in the upsert loop, compare `existingColumnNameMap.get(column.id)` against `column.name`. On mismatch, call `dataTableDDLService.renameColumn()` within the same transaction.

</details>

### DIFF
--- a/packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
+++ b/packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
@@ -1422,6 +1422,20 @@ export class SourceControlImportService {
 						});
 						columnEntities.push(columnEntity);
 
+						// Rename columns whose name changed (same ID, different name)
+						if (!isNewTable && existingColumnIds.has(column.id)) {
+							const oldName = existingColumnNameMap.get(column.id);
+							if (oldName && oldName !== column.name) {
+								await this.dataTableDDLService.renameColumn(
+									dataTable.id,
+									oldName,
+									column.name,
+									dbType,
+									trx,
+								);
+							}
+						}
+
 						// Add new columns to existing physical table
 						if (!isNewTable && !existingColumnIds.has(column.id)) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
```diff
@@ -1422,6 +1422,20 @@ export class SourceControlImportService {
 						});
 						columnEntities.push(columnEntity);
 
+						// Rename columns whose name changed (same ID, different name)
+						if (!isNewTable && existingColumnIds.has(column.id)) {
+							const oldName = existingColumnNameMap.get(column.id);
+							if (oldName && oldName !== column.name) {
```
geemanjs, line 1428: should we be deleting a column if it doesn't exist? or does it do this somewhere else already?

geemanjs (approved):

packages/cli/src/modules/source-control.ee/source-control-import.service.ee.ts
ireneea: Yup that case is already handled 👍

PR #27748, opened 2026-03-30
feat(workflow-sdk): Dagre-based layout for AI-generated workflows

r00gm:
## Summary
- Replace the SDK's simple BFS layout with a Dagre-based algorithm matching the FE's `useCanvasLayout`.
- Apply layout in `submit-workflow` and `build-workflow` tools before saving to DB, since the sandbox runs its own SDK copy
- Export `layoutWorkflowJSON()` that builds a `GraphNode` map from serialized JSON and delegates to `calculateNodePositions`

# Prompt
Make one workflow. Add a Webhook node connected to an AI Agent with OpenAI, web search, calculator and memory tools. Separately in the same workflow, add a Schedule Trigger connected to Google Sheets, then a Switch with 4 outputs, each connected to a Slack node. Do NOT create two workflows.

| Before | After |
| ------------- | ------------- |
| image | image  |

## Test plan
- [ ] Generate AI workflow (agent + subnodes) — verify positions match FE tidy-up layout
- [ ] Generate simple linear workflow — verify LR positioning
- [ ] Generate disconnected subgraphs — verify vertical arrangement
- [ ] Existing `pnpm test` in workflow-sdk passes (11,411 tests)
- [ ] Full build passes

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
@@ -274,10 +274,8 @@ export function createSubmitWorkflowTool(
 			}
 
 			// Apply Dagre-based layout to produce positions matching the FE's tidy-up
-			layoutWorkflowJSON(buildOutput.workflow);
-
 			// Override name if provided
-			const json = buildOutput.workflow;
+			const json = layoutWorkflowJSON(buildOutput.workflow);
 			if (name) {
 				json.name = name;
 			} else if (!json.name && !workflowId) {
--- a/packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
@@ -9,7 +9,7 @@
 import { createTool } from '@mastra/core/tools';
 import type { Workspace } from '@mastra/core/workspace';
 import type { WorkflowJSON } from '@n8n/workflow-sdk';
-import { validateWorkflow, layoutWorkflowJSON } from '@n8n/workflow-sdk';
+import { validateWorkflow } from '@n8n/workflow-sdk';
 import { createHash, randomUUID } from 'node:crypto';
 import { z } from 'zod';
 
@@ -273,9 +273,8 @@ export function createSubmitWorkflowTool(
 				};
 			}
 
-			// Apply Dagre-based layout to produce positions matching the FE's tidy-up
 			// Override name if provided
-			const json = layoutWorkflowJSON(buildOutput.workflow);
+			const json = buildOutput.workflow;
 			if (name) {
 				json.name = name;
 			} else if (!json.name && !workflowId) {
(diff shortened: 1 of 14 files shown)


packages/@n8n/workflow-sdk/src/workflow-builder/layout-utils.test.ts
```diff
@@ -155,33 +259,51 @@ describe('calculateNodePositions', () => {
...
+				createGraphNode('trigger', 'n8n-nodes-base.manualTrigger', triggerConns),
+			);
+			nodes.set('set', createGraphNode('set', 'n8n-nodes-base.set'));
+			nodes.set('note', createGraphNode('note', STICKY_NODE_TYPE));
+
+			const positions = calculateNodePositions(nodes);
+
+			// Non-sticky nodes get positions
+			expect(positions.has('trigger')).toBe(true);
+			expect(positions.has('set')).toBe(true);
```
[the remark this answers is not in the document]
r00gm: fixed https://github.com/n8n-io/n8n/pull/27748/commits/d02d24b9c8ead0484ea75be7f81c6a5241d793a6

packages/@n8n/workflow-sdk/src/workflow-builder/layout-utils.ts
```diff
@@ -1,80 +1,676 @@
...
  * Layout Utility Functions
  *
- * Functions for calculating node positions in workflow layouts.
+ * Dagre-based layout that mirrors the frontend's useCanvasLayout algorithm
+ * to produce positions matching what the FE tidy-up would generate.
  */
 
-import { NODE_SPACING_X, DEFAULT_Y, START_X } from './constants';
-import type { GraphNode } from '../types/base';
+import dagre from '@dagrejs/dagre';
```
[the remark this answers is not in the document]
r00gm: The SDK’s `layout-utils` is imported by `json-serializer` which is part of the builder core, so dagre is already loaded whenever any workflow is serialized. A dynamic `await import() `would require making `calculateNodePositions` async, which cascades into the sync `serialize() `method in `json-serializer.ts`. Keeping the top-level import.

packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
```diff
@@ -273,8 +273,9 @@ export function createSubmitWorkflowTool(
 				};
 			}
 
+			// Apply Dagre-based layout to produce positions matching the FE's tidy-up
 			// Override name if provided
-			const json = buildOutput.workflow;
+			const json = layoutWorkflowJSON(buildOutput.workflow);
```
mutdmour, line 279: in the sandbox, we build the wrokflow and generate json like this `workflow().add(nodeA.to(nodeB)).toJSON()`

^ this is the workflow sdk. The natural place here for this layout is in the json serializer. `workflow().toJSON({ tidyUp: true })`
would be more fitting.

r00gm: i've been looking into this and i don't think this is optional. 
AFAIK, the builder produces nodes without positions and the previous implementation post processed the workflow to make sure every node had them.
This is probably due to the fact that the engine executes from the top left corner to the bottom right corner and capitalizes on this to find the trigger nodes that should be executed first.

agree that it should be  `.toJSON()` but idont think it should't be opt-(in/out) `.toJSON({ tidyUp: true })`

we could keep both approaches BFS left-to-right layout and Dagre but then we would have to maintain both when IMO doesn't make sense 

WDYT?

mutdmour: I would keep both. Because if a workflow already has positions, then we don't necessarily want to change that. The SDK is use

PR #27760, opened 2026-03-30
fix(Read/Write Files from Disk Node): Add search aliases for binary file nodes

BerniWittmann:
## Summary

Adds `Binary File`, `Read Binary File`, and `Write Binary File` as search aliases to the Read/Write Files from Disk node. This makes it easier for users to find this node when searching for the legacy binary file node names.

> Note: This only happens because the alias search only matches on the full phrase. (When using the Search Term `Write Binary File` the alias `Binary` does not trigger.) Decided against rewriting the alias matching functionality to also match on alias that appear in the search term as a sub-term, due to the small nature of the issue (XS + onboarding + low prio) and unclear side effects. 

https://linear.app/n8n/issue/NODE-4225

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/Files/ReadWriteFile/ReadWriteFile.node.json
+++ b/packages/nodes-base/nodes/Files/ReadWriteFile/ReadWriteFile.node.json
@@ -10,7 +10,20 @@
 			}
 		]
 	},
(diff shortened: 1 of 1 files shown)


dlavrenuek (approved):
🎉

PR #27789, opened 2026-03-30
feat(core): OTEL - workflow node tracing

geemanjs:
## Summary

Adds "node level" tracing.
image
image

Also exposes a `tracing` bag that node's can write to that will land the data into OTEL. 

ligo-394

## Review / Merge checklist

- [x] Tests included. 


packages/workflow/src/interfaces.ts
```diff
@@ -2774,6 +2774,11 @@ export interface ITaskMetadata {
 	 * Contains token for security validation.
 	 */
 	resumeUrl?: string;
+
+	/**
+	 * Key-value pairs that can set for tracing - they will be attached to the OTEL node span
```
alielkhateeb: Super duper nit haha:
```suggestion
	 * Key-value pairs that can be set for tracing - they will be attached to the OTEL node span
```

packages/cli/src/modules/otel/otel.constants.ts
```diff
@@ -17,4 +17,11 @@ export const ATTR = {
 	EXECUTION_IS_RETRY: 'n8n.execution.is_retry',
 	EXECUTION_RETRY_OF: 'n8n.execution.retry_of',
 	EXECUTION_ERROR_TYPE: 'n8n.execution.error_type',
+
+	NODE_ID: 'n8n.node.id',
+	NODE_NAME: 'n8n.node.name',
+	NODE_TYPE: 'n8n.node.type',
+	NODE_TYPE_VERSION: 'n8n.node.type_version',
+	NODE_ITEMS_INPUT: 'n8n.node.items.input',
+	NODE_ITEMS_OUTPUT: 'n8n.node.items.output',
```
alielkhateeb, line 26: What do you think of having a `NODE_TERMINATION_REASON`?
I feel like it's a case that is missing in this PR, what happens with dangling nodes? E.g. nodes started but then I cancel the workflow so the node will never actually "end" from observability perspective and will be cleaned up.

Wouldn't it be beneficial to have visibility on such use case?

packages/cli/src/modules/otel/handlers/node-end.handler.ts
```diff
@@ -0,0 +1,72 @@
...
+
+		const outputItems = this.countItems(ctx.taskData.data);
+		span.setAttribute(ATTR.NODE_ITEMS_OUTPUT, outputItems);
+
+		const inputItems = this.countInputItems(ctx);
+		span.setAttribute(ATTR.NODE_ITEMS_INPUT, inputItems);
+
+		if (ctx.taskData.metadata?.tracing) {
+			for (const [key, value] of Object.entries(ctx.taskData.metadata.tracing)) {
+				span.setAttribute(`n8n.node.custom.${key}`, value);
```
alielkhateeb, line 31: Super Nit: I'm a big fan of private methods, only because it names a block rather than my dumb self trying to trace the code and understand what are we actually doing here 😂

alielkhateeb (approved):
Looks good 🚀 
What I like the most about this PR is actually all the noise removed from workflow-fixtures 😂🔥

PR #27811, opened 2026-03-31
feat(core): Add data table MCP tool

CharlieKolb:
## Summary

Expose the following data table functionality for MCP:

- Search/List data tables
- Create and rename data tables (remarkably no deletion for now)
- Add/delete/rename columns
- Insert rows (remarkably no reading for now)

https://linear.app/n8n/issue/ADO-4938

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/mcp/tools/data-table.tool.ts
```diff
@@ -0,0 +1,491 @@
...
+		const telemetryPayload: UserCalledMCPToolEventPayload = {
+			user_id: user.id,
+			tool_name: 'search_data_tables',
+			parameters: { query, projectId, limit },
+		};
+
+		try {
+			const safeLimit = Math.min(Math.max(1, limit), SEARCH_MAX_RESULTS);
+
+			const result = await dataTableAggregateService.getManyAndCount(user, {
```
CharlieKolb: Do we want to support pagination in these? Doesn't look like the other resources do, right?

packages/cli/src/modules/mcp/tools/data-table.tool.ts
```diff
@@ -0,0 +1,491 @@
...
+		}
+	},
+});
+
+// #endregion
+
+// #region Create Data Table
+
+const columnSchema = z.object({
+	name: z
```
CharlieKolb: Should we use [the data table schemas](https://github.com/n8n-io/n8n/blob/1bacba3d8b3a16eec34f510495b4807cdc4cd2ae/packages/@n8n/api-types/src/schemas/data-table.schema.ts) we already have or is it best to keep these separate?

packages/cli/src/modules/mcp/tools/data-table.tool.ts
```diff
@@ -0,0 +1,491 @@
+import type { ListDataTableQueryDto } from '@n8n/api-types';
```
MiloradFilipovic: Can we define each tool from here in it's separate file?
This way it follow convention we used for existing tools and makes it easier to reason about the whole module when it's modularized a bit more

packages/cli/src/modules/mcp/tools/data-table.tool.ts
MiloradFilipovic: Would be ideal, but we don't do it in other places as said, so ok to skip here

packages/cli/src/modules/mcp/tools/data-table.tool.ts
MiloradFilipovic: If possible, yes

packages/cli/src/modules/mcp/tools/data-table.tool.ts
```diff
@@ -0,0 +1,491 @@
...
+				})),
+			});
+
+			const tableColumns = await dataTableService.getColumns(result.id, projectId);
+
+			const output = {
+				id: result.id,
+				name: result.name,
+				projectId: result.projectId,
+				columns: tableColumns.map((col) => ({
```
MiloradFilipovic: Nit: If this just mirrors the input in case of success, we can omit it from the output to save a few tokens

packages/cli/src/modules/mcp/tools/data-table.tool.ts
```diff
@@ -0,0 +1,491 @@
...
+			throw error;
+		}
+	},
+});
+
+// #endregion
+
+// #region Modify Data Table
+
+const modifyInputSchema = {
```
MiloradFilipovic: I would also love to break this up into separate tools for each operation

packages/cli/src/modules/mcp/tools/data-table.tool.ts
```diff
@@ -0,0 +1,491 @@
...
+				structuredContent: output,
+			};
+		} catch (error) {
+			const errorMessage = error instanceof Error ? error.message : String(error);
+			telemetryPayload.results = {
+				success: false,
+				error: errorMessage,
+			};
+			telemetry.track(USER_CALLED_MCP_TOOL_EVENT, telemetryPayload);
+			throw error;
```
MiloradFilipovic: `search_data_tables` catches errors and returns `isError: true` with a graceful response, while `create_data_table`, `modify_data_table`, `and `add_data_table_rows` all throw error after tracking telemetry. Is this intentional?

MiloradFilipovic (commented):
Very nice work 👌
I left a few comments and please check if cubic comment makes sense (we need to make sure to respect user and project-scopes when accessing resources like data tables).

packages/cli/src/modules/mcp/tools/data-table.tool.ts
CharlieKolb: This is just for the columns, right?

MiloradFilipovic: Yes

packages/cli/src/modules/mcp/tools/data-table.tool.ts
CharlieKolb: I guess the difference is that read-only operations fail gracefully whereas the write operations fail with an error. I wouldn't really know the nuances for MCP, so do let me know which is preferred

MiloradFilipovic (dismissed):
Nice, thanks 🙌

MiloradFilipovic (approved):
🚢

PR #27813, opened 2026-03-31
fix(Data Table Node): Change copy for Limit to indicate it applies per input row

CharlieKolb:
## Summary

Let's be a bit more explicit here since this has repeatedly caused confusion and user intention is usually to execute this node once anyway.

https://linear.app/n8n/issue/ADO-4976

## Review / Merge checklist

- [ ] Tests included. 


RobTF9 (changes requested):
Change copy to: "Limit per input row"

RobTF9 (approved):

PR #27816, opened 2026-03-31
test(core): Replace mock-based VM error tests with end-to-end integration tests

despairblue:
## Summary
- Replaces mock-based ExpressionError and ExpressionExtensionError description preservation tests with real end-to-end integration tests
- The old tests used `setVmEvaluator()` to mock the evaluator, so errors never crossed the isolate boundary — they passed trivially via `isExpressionError()` without exercising `mapVmError`'s reconstruction
- New tests throw from a getter on `connectionInputData`, so the error crosses the bridge boundary, gets serialized/reconstructed, and goes through `mapVmError`
- Uses `toEqual` to assert the entire error object matches, catching any field dropped during reconstruction

Linear ticket: https://linear.app/n8n/issue/CAT-2543


ivov (dismissed):

ivov (approved):

sanchezmaldonadojesusadrian14-coder (commented):
Espero que esté cambio sea funcional

PR #27824, opened 2026-03-31
feat(core): Add OIDC test login endpoint with frontend Test button

afitzek:
## Summary

- Add `POST /sso/oidc/config/test` endpoint that initiates an OIDC authorization flow in test mode — captures IdP claims and renders a result page without creating a session or modifying user data
- Encode a `testMode` flag in the OIDC state JWT so the callback handler can distinguish test flows from real logins
- Add a "Test" button to the OIDC settings form in the frontend (enabled when discovery endpoint, client ID, and client secret are configured)
- Render success/failure HTML pages showing returned OIDC attributes (email, name, subject) with proper HTML escaping

https://linear.app/n8n/issue/IAM-502/oidc-add-test-login

## Review / Merge checklist

- [x] Tests included.

## How to test

1. Configure an OIDC provider in **Settings → SSO → OIDC** (discovery endpoint, client ID, client secret)
2. Click the **Test** button — a new browser tab opens with the IdP login page
3. Authenticate with the IdP — the tab should render a success page showing the returned attributes (email, first name, last name, subject)
4. If authentication fails or the provider returns an error, the tab should render a failure page with the error message
5. Verify that no session cookie is set and no user record is created/modified after the test flow

### DIFF
--- a/packages/cli/src/modules/sso-oidc/oidc.service.ee.ts
+++ b/packages/cli/src/modules/sso-oidc/oidc.service.ee.ts
@@ -102,19 +102,25 @@ export class OidcService {
 		};
 	}
 
-	generateState() {
+	generateState(testMode = false) {
 		const state = `n8n_state:${randomUUID()}`;
+		const payload: Record<string, unknown> = { state };
+		if (testMode) {
+			payload.testMode = true;
+		}
 		return {
-			signed: this.jwtService.sign({ state }, { expiresIn: '15m' }),
+			signed: this.jwtService.sign(payload, { expiresIn: '15m' }),
 			plaintext: state,
 		};
 	}
 
-	verifyState(signedState: string) {
+	verifyState(signedState: string): { state: string; testMode?: boolean } {
 		let state: string;
+		let testMode: boolean | undefined;
 		try {
 			const decodedState = this.jwtService.verify(signedState);
 			state = decodedState?.state;
+			testMode = decodedState?.testMode;
 		} catch (error) {
 			this.logger.error('Failed to verify state', { error });
 			throw new BadRequestError('Invalid state');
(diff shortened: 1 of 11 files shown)


BGZStephen (approved):

packages/cli/src/modules/sso-oidc/oidc.service.ee.ts
```diff
@@ -335,6 +341,101 @@ export class OidcService {
 		return user;
 	}
 
+	async generateTestLoginUrl(): Promise<{ url: URL; state: string; nonce: string }> {
+		await this.loadOpenIdClient();
+		const config = await this.loadConfig(true);
```
guillaumejacquart, line 346: why not use getOidcConfiguration like on the `generateLoginUrl` method ?

packages/cli/src/modules/sso-oidc/oidc.service.ee.ts
```diff
@@ -335,6 +341,101 @@ export class OidcService {
 		return user;
 	}
 
+	async generateTestLoginUrl(): Promise<{ url: URL; state: string; nonce: string }> {
```
guillaumejacquart, line 344: This method looks a lot like the generateLoginUrl. I'm wondering if we shouldn't "just" add a parameter to it to accept the test mode

packages/cli/src/modules/sso-oidc/oidc.service.ee.ts
```diff
@@ -335,6 +341,101 @@ export class OidcService {
...
+			nonce: nonce.plaintext,
+			...(config.authenticationContextClassReference.length > 0 && {
+				acr_values: config.authenticationContextClassReference.join(' '),
+			}),
+		});
+
+		return { url: authorizationURL, state: state.signed, nonce: nonce.signed };
+	}
+
+	async processTestCallback(
```
guillaumejacquart, line 381: Also looks a lot like the loginUser method. Maybe both could call the same one that process the state and nonce verification, as well as fetching the token and user info ?

guillaumejacquart (commented):
some comments

PR #27837, opened 2026-03-31
perf(core): Resolve VM proxy results on host side to avoid cross-bridge walk

despairblue:
## Summary

When the VM expression engine evaluates an expression that returns an object
(e.g. `{{ $json }}`, `{{ $json.items }}`), `__prepareForTransfer` previously
walked the lazy proxy tree inside the isolate, triggering a cross-bridge
`applySync` callback for every leaf node. For large data (10k items with
nested objects) this caused ~60k+ boundary crossings per evaluation —
~3,000x slower than the other VM benchmarks.

**Fix:** When `__prepareForTransfer` encounters a lazy proxy, it now returns a
lightweight sentinel containing the proxy's path. The host detects the
sentinel after `runSync` and resolves the value directly from its in-memory
data — zero boundary crossings.

All results are wrapped in a typed sentinel (`DataResultSentinel` or
`ProxyResultSentinel`) so that user code cannot forge a proxy sentinel to
influence host-side path resolution.

### Benchmark results (10k-item array with nested objects)

| Benchmark | Master (ops/sec) | Branch (ops/sec) | Improvement |
|-----------|------------------|-------------------|-------------|
| VM: `{{ $json }}` | 1.45 | 175 | **~121x faster** |
| VM: `{{ $json.items }}` | 1.42 | 188 | **~132x faster** |

No regressions in any existing VM benchmarks (simple property, nested property,
array access, conditionals, extension calls — all within noise).

### Changes

- Extract shared `prepareForTransfer` + sentinel types to `src/shared/serialize.ts`
- Add proxy detection in isolate-side `__prepareForTransfer` (returns path sentinel)
- Wrap all results in typed sentinels (`DataResultSentinel` / `ProxyResultSentinel`)
  to prevent sentinel forgery from user code
- Add host-side sentinel resolution in `IsolatedVmBridge.execute()`
- Register array proxies in `proxyPaths` WeakMap (were previously untracked)
- Deduplicate path navigation in bridge callbacks via `resolvePathInData`
- Validate `__path` is `string[]` in proxy sentinel type guard
- Add `IsolateResult` union type covering all `runSync` return values
- Add integration tests for object-return expressions, sentinel forgery, and
  prototype consistency
- Add Tier 1 benchmarks comparing both engines on object-return pattern

https://linear.app/n8n/issue/CAT-2595

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
+++ b/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
@@ -284,6 +284,18 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 		});
 	});
 
+	it('should not allow __proto__ key to pollute the returned object prototype', () => {
+		const result = evaluator.evaluate(
+			'{{ JSON.parse(\'{"__proto__": {"isAdmin": true}, "name": "test"}\') }}',
+			{ $json: {} },
+		) as Record<string, unknown>;
+
+		expect(result.name).toBe('test');
+		// __proto__ should be a plain data property, not a prototype setter
+		expect(Object.prototype.hasOwnProperty.call(result, '__proto__')).toBe(true);
+		expect(result.isAdmin).toBeUndefined();
+	});
+
 	it('should throw on invalid timezone', async () => {
 		const data = { $json: { x: 1 } };
 
--- a/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
+++ b/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
@@ -288,12 +288,29 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 		const result = evaluator.evaluate(
 			'{{ JSON.parse(\'{"__proto__": {"isAdmin": true}, "name": "test"}\') }}',
 			{ $json: {} },
+			caller,
 		) as Record<string, unknown>;
 
 		expect(result.name).toBe('test');
 		// __proto__ should be a plain data property, not a prototype setter
 		expect(Object.prototype.hasOwnProperty.call(result, '__proto__')).toBe(true);
 		expect(result.isAdmin).toBeUndefined();
+	});
+
+	it('should not leak host-side functions via forged proxy sentinel', () => {
+		const data = {
+			$json: {},
+			$items: () => 'should not be reachable',
+		};
+
+		// A forged sentinel that tries to resolve $items (a function) on the host
+		const result = evaluator.evaluate(
+			'{{ JSON.parse(\'{"__isProxyResult": true, "__path": ["$items"]}\') }}',
+			data,
+			caller,
+		);
+
+		expect(typeof result).not.toBe('function');
 	});
 
 	it('should throw on invalid timezone', async () => {
@@ -362,7 +379,7 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 			},
 		};
 
-		const result = evaluator.evaluate('{{ $json.user }}', data);
+		const result = evaluator.evaluate('{{ $json.user }}', data, caller);
 
 		expect(result).toEqual({
 			name: 'Alice',
@@ -380,7 +397,7 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 			},
 		};
 
-		const result = evaluator.evaluate('{{ $json }}', data);
+		const result = evaluator.evaluate('{{ $json }}', data, caller);
 
 		expect(result).toEqual({
 			name: 'Alice',
@@ -399,7 +416,7 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 			},
 		};
 
-		const result = evaluator.evaluate('{{ $json.items }}', data);
+		const result = evaluator.evaluate('{{ $json.items }}', data, caller);
 
 		expect(result).toEqual([
 			{ id: 1, name: 'one' },
--- a/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
+++ b/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
@@ -339,6 +339,62 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 		expect(result).toBe(systemOffset);
 	});
 
+	it('should return an entire object (not just a leaf primitive)', () => {
+		const data = {
+			$json: {
+				user: {
+					name: 'Alice',
+					age: 30,
+					address: { city: 'Berlin', zip: '10115' },
+				},
+			},
+		};
+
+		const result = evaluator.evaluate('{{ $json.user }}', data);
+
+		expect(result).toEqual({
+			name: 'Alice',
+			age: 30,
+			address: { city: 'Berlin', zip: '10115' },
+		});
+	});
+
+	it('should return the root $json object', () => {
+		const data = {
+			$json: {
+				name: 'Alice',
+				active: true,
+				tags: ['admin', 'user'],
+			},
+		};
+
+		const result = evaluator.evaluate('{{ $json }}', data);
+
+		expect(result).toEqual({
+			name: 'Alice',
+			active: true,
+			tags: ['admin', 'user'],
+		});
+	});
+
+	it('should return an array of objects', () => {
+		const data = {
+			$json: {
+				items: [
+					{ id: 1, name: 'one' },
+					{ id: 2, name: 'two' },
+				],
+			},
+		};
+
+		const result = evaluator.evaluate('{{ $json.items }}', data);
+
+		expect(result).toEqual([
+			{ id: 1, name: 'one' },
+			{ id: 2, name: 'two' },
+		]);
+	});
+
 	it('should support Object.keys() on root proxy data', () => {
 		const data = {
 			$json: { name: 'Alice', age: 30, city: 'Berlin' },
--- a/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
+++ b/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
@@ -284,19 +284,6 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 		});
 	});
 
-	it('should not allow __proto__ key to pollute the returned object prototype', () => {
-		const result = evaluator.evaluate(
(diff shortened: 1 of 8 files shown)


packages/@n8n/expression-runtime/src/runtime/serialize.ts
```diff
@@ -1,53 +1,22 @@
...
  *
- * Note: JS Date objects survive structured clone with prototype intact
- * (Date is a standard structured-cloneable type) and are not converted.
- *
- * Circular references are not handled — expression results should not
- * contain cycles.
+ * For non-proxy values, delegates to the shared prepareForTransfer.
  */
 export function __prepareForTransfer(value: unknown): unknown {
-	if (value === null || value === undefined) return value;
```
despairblue, line 27: extracted to shared module

packages/@n8n/expression-runtime/src/runtime/lazy-proxy.ts
```diff
@@ -230,6 +230,7 @@ export function createDeepLazyProxy(basePath: string[] = [], knownKeys?: string[
 					},
 				});
 
+				proxyPaths.set(arrayProxy, path);
```
despairblue, line 235: This is here so that `isLazyProxy` can correctly identify array proxies.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -588,6 +590,15 @@ export class IsolatedVmBridge implements RuntimeBridge {
 				throw this.reconstructError(result);
 			}
 
+			// Step 7: If the result is a proxy-result sentinel, resolve from
+			// host-side data. This avoids the cross-bridge callback storm that
+			// would occur if __prepareForTransfer walked the proxy inside the
+			// isolate (one callback per leaf node).
+			if (isProxyResultSentinel(result)) {
+				const resolved = resolvePathInData(data, result.__path);
+				return prepareForTransfer(resolved);
```
despairblue, line 485: Maybe that name should be changed now 🤔

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -588,6 +541,15 @@ export class IsolatedVmBridge implements RuntimeBridge {
...
+			}
+
```
despairblue: This is the main part, everything else in this file is a refactor (extracting path resolution logic to `resolvePathInData` and reusing this here).

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -588,6 +541,15 @@ export class IsolatedVmBridge implements RuntimeBridge {
 				throw this.reconstructError(result);
 			}
 
+			// Step 7: If the result is a proxy-result sentinel, resolve from
+			// host-side data. This avoids the cross-bridge callback storm that
+			// would occur if __prepareForTransfer walked the proxy inside the
+			// isolate (one callback per leaf node).
+			if (isProxyResultSentinel(result)) {
```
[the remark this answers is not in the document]
despairblue: Addressed in df421b5905 — __prepareForTransfer now always wraps results in a DataResultSentinel or ProxyResultSentinel. A forged `{ __isProxyResult: true, __path: [...] }` from user code gets wrapped in a DataResultSentinel, so the host treats it as data. Sentinel forgery is structurally impossible.

packages/@n8n/expression-runtime/src/shared/serialize.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	// Array — walk elements
+	if (Array.isArray(value)) return value.map(prepareForTransfer);
+
+	// Only walk plain objects — structured-cloneable types like Date, Map, Set,
+	// RegExp, Error, typed arrays survive copy:true with prototypes intact.
+	if (!isPlainObject(value)) return value;
+
+	// Plain object — walk values
+	const result: Record<string, unknown> = {};
```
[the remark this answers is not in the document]
despairblue: Addressed in a326da06c9 — changed to `Object.create(null)` so `__proto__` is treated as a plain data property.

packages/@n8n/expression-runtime/src/shared/serialize.ts
```diff
@@ -0,0 +1,64 @@
...
+		const iso = value.toISO();
+		return iso === 'Invalid Interval' ? null : iso;
+	}
+
+	// Array — walk elements
+	if (Array.isArray(value)) return value.map(prepareForTransfer);
+
+	// Only walk plain objects — structured-cloneable types like Date, Map, Set,
+	// RegExp, Error, typed arrays survive copy:true with prototypes intact.
+	if (!isPlainObject(value)) return value;
```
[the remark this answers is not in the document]
despairblue: Pre-existing behavior, not introduced by this PR. Out of scope here.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -350,25 +356,7 @@ export class IsolatedVmBridge implements RuntimeBridge {
...
-						startIndex = 2;
-					}
-				}
-				for (let i = startIndex; i < path.length; i++) {
-					value = (value as Record<string, unknown>)?.[path[i]];
-					if (value === undefined || value === null) {
-						return value;
-					}
-				}
+				const value = resolvePathInData(data, path);
```
ivov, line 296: If we merge this after #27895 we'll need to use `resolvePathInData` for `$()` as well

packages/@n8n/expression-runtime/src/shared/serialize.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/** All possible return types from script.runSync in the isolate. */
+export type IsolateResult = ErrorSentinel | ProxyResultSentinel | DataResultSentinel;
+
+export function isProxyResultSentinel(value: unknown): value is ProxyResultSentinel {
+	return (
+		typeof value === 'object' &&
+		value !== null &&
+		(value as Record<string, unknown>).__isProxyResult === true
```
ivov: We could also check if `__path` is a `string[]`

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
ivov, line 1: We should align the `execute` docline with these changes, and fix the numbering if steps (4 is skippd)

packages/@n8n/expression-runtime/src/shared/serialize.ts
```diff
@@ -0,0 +1,95 @@
...
+	// Only walk plain objects — structured-cloneable types like Date, Map, Set,
+	// RegExp, Error, typed arrays survive copy:true with prototypes intact.
+	if (!isPlainObject(value)) return value;
+
+	// Plain object — walk values
+	const result: Record<string, unknown> = Object.create(null);
+	for (const key of Object.keys(value)) {
+		result[key] = prepareForTransfer(value[key]);
+	}
+	return result;
```
ivov: Data path returns an `Object.prototype` object but a proxy returns a `null` prototype object, I don't think this breaks anything in how we consume expression results downstream but we might want to make them consistent just in case? `hasOwnPropery` would behave differently?

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -592,7 +549,22 @@ export class IsolatedVmBridge implements RuntimeBridge {
 				console.log('[IsolatedVmBridge] Expression executed successfully');
 			}
 
-			return result;
+			// Step 7: Unwrap the transfer sentinel.
+			// __prepareForTransfer always wraps results so that user code
+			// cannot forge a ProxyResultSentinel.
+			if (isProxyResultSentinel(result)) {
```
ivov, line 483: If it's not directly a proxy we don't benefit? 

```
{{ { data: $json } }}
```

Or maybe I'm misunderstanding

packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
```diff
@@ -284,6 +284,35 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
...
+		};
+
+		// A forged sentinel that tries to resolve $items (a function) on the host
+		const result = evaluator.evaluate(
+			'{{ JSON.parse(\'{"__isProxyResult": true, "__path": ["$items"]}\') }}',
+			data,
+			caller,
+		);
+
+		expect(typeof result).not.toBe('function');
```
ivov: `expect(result).toEqual({ __isProxyResult: true, __path: ['$items'] });`

ivov (commented):
Clever! But also scary

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: fixed a63ef4072549db515f928c1e1a584d637cb0044e

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: already done

packages/@n8n/expression-runtime/src/shared/serialize.ts
despairblue: done 8010406b88111815405c3a69f7fde5177a2037c9

packages/@n8n/expression-runtime/src/shared/serialize.ts
despairblue: Reverted the `Object.create(null)` thing. It's not really an attack vector.
1b7d6a7a34ab899b6d2f0fb8e14667a076111d57

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: That is true, I've been putting this off for now. Will look into this though in a bit.

packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
despairblue: done c3a7d542caa1fc1ba01272817467135e3a934867

despairblue:
not necessary anymore and too outdated

PR #27853, opened 2026-03-31
fix(core): Add streaming keepalive to prevent proxy timeout during long agent executions

DeveloperTheExplorer:
## Summary

When a long-running AI Agent tool call (e.g. Browserbase browser agent ~3 min) produces no streaming chunks, reverse proxies like Cloudflare time out the idle connection (~100s) and inject a raw 524 error HTML page into the chat stream. The user sees jumbled HTML markup in the chat panel even though the agent eventually completes successfully.

This PR fixes the issue with three layers:

1. **Socket configuration** (ChatTrigger.node.ts) — Apply the same socket settings the SSE push path uses (`setTimeout(0)`, `setNoDelay(true)`, `setKeepAlive(true)`) to prevent TCP-level idle timeouts.

2. **Execution-bound heartbeat** (workflow-runner.ts) — Write an empty `\n` every 30s to the chunked response while the execution is active. The heartbeat is tied to the execution lifecycle via `workflowExecuteAfter` hook, so it does not outlive the execution or mask real timeouts (`EXECUTIONS_TIMEOUT` still fires independently). The frontend line parser already skips empty lines.

3. **Frontend HTML guard** (chat message.ts) — In `createLineParser()`, skip non-JSON lines containing `<!DOCTYPE` or `<html` instead of rendering them as chat text. Defense-in-depth if a proxy still injects HTML despite layers 1 and 2.

### How to test

1. Set up a Chat Trigger workflow with a long-running AI Agent tool (e.g. Browserbase agent that takes >100s)
2. Deploy behind Cloudflare or a reverse proxy with an idle timeout
3. Send a chat message and wait for the tool to complete
4. Verify no HTML error appears in the chat — only the agent's actual response

https://linear.app/n8n/issue/AI-2319

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
```diff
@@ -894,9 +894,19 @@ export class ChatTrigger extends Node {
...
+			req.socket.setNoDelay(true);
+			req.socket.setKeepAlive(true);
+
+			// Set up streaming response headers.
+			// Content-Encoding: identity prevents the compression middleware from
+			// wrapping the response in zlib, which can buffer small writes (like
+			// keepalive heartbeats) and prevent them from reaching the network.
 			res.writeHead(200, {
 				'Content-Type': 'application/json; charset=utf-8',
+				'Content-Encoding': 'identity',
```
lucamattiazzi: [according to the specs](https://httpwg.org/specs/rfc9110.html#field.content-encoding) `identity` should not be used in the `content-encoding` header, since it's reserved for the `accepted-encoding` header.

I've seen in a [different issue on a different repo](https://github.com/honojs/hono/issues/3449) that it was still used in combination with Cloudflare, is this the reason we are using it here?

lucamattiazzi (dismissed):

lucamattiazzi (commented):
sorry for re-requesting my own review, I did a successful test of the new code but could not yet replicate the bug, testing it now!

packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
DeveloperTheExplorer: That's a great point! It was part of the attempt to fix the bug but it isn't necessary. Good find!

lucamattiazzi (approved):

PR #27857, opened 2026-03-31
feat(instance-ai) : Enable feature toggling setting for Instance AI and disabling of Chathub when enabled

Joco-95:
## Summary

As part of the opt-in flow, users will be opted into Instance AI by default - with the possibility of toggling the feature on and off later via settings. This PR implements the toggle functionality :
* Hides all Instance AI-related UI except for settings (Sidebar, commands ...)
* Disables Instance AI-related routes
* Disables Chathub (icon hidden on sidebar, Chathub settings hidden) when Instance AI is enabled

https://github.com/user-attachments/assets/cbac7226-55ea-4139-84e8-8490ab34d21c

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -103,6 +110,7 @@ export class InstanceAiController {
 		@Param('threadId') threadId: string,
 		@Query query: InstanceAiEventsQuery,
 	) {
+		this.requireInstanceAiEnabled();
```
mutdmour, line 113: nit - this could be a guard instead applied to the the entire controller. (except maybe the settings endpoints that could be moved out)

packages/frontend/editor-ui/src/app/moduleInitializer/moduleInitializer.ts
```diff
@@ -73,7 +74,23 @@ const checkModuleAvailability = (options: any) => {
...
+
+	// Special case for instance-ai settings route : always accessible even whe instanceAiEnabled admin setting is false
+	// Other instance-ai related routes are disabled
+	if (options.to.meta.moduleName === 'instance-ai') {
+		const routeName = options.to.name;
+		if (routeName !== INSTANCE_AI_SETTINGS_VIEW) {
+			const enabled = settingsStore.moduleSettings['instance-ai']?.instanceAiEnabled;
+			if (enabled === false) {
+				return false;
+			}
```
mutdmour, line 90: nit - `return enabled;`

packages/@n8n/api-types/src/frontend-settings.ts
```diff
@@ -273,6 +273,7 @@ export type FrontendModuleSettings = {
 	 */
 	'instance-ai'?: {
 		enabled: boolean;
+		instanceAiEnabled: boolean; // Admin on/off switch : used to toggle Instance AI on and off on settings
```
mutdmour, line 276: What's the difference between `enabled` and `instanceAiEnabled`? If different, can we rename `instanceAiEnabled` to be more clear?

packages/@n8n/config/src/configs/instance-ai.config.ts
```diff
@@ -42,6 +42,12 @@ export class InstanceAiConfig {
 	@Env('N8N_INSTANCE_AI_SUB_AGENT_MAX_STEPS')
 	subAgentMaxSteps: number = 100;
 
+	/**
+	 * When false, Instance AI chat and related user-facing features are off; admins can
+	 * still open Instance AI settings. Persisted via admin settings API.
+	 */
+	instanceAiEnabled: boolean = true;
```
mutdmour: how does this play with the instance ai module flag? that needs to enable the feature in cloud to 1% of new users?

Joco-95: The module is always enabled - however `instanceAiEnabled` is independent and the setting we persist on settings to toggle the feature on and off at the UI level / disable the routes, without impacting the overall module load (since we do need it for things like the settings page which is needed to toggle it on and off).

packages/frontend/editor-ui/src/app/moduleInitializer/moduleInitializer.ts
Joco-95: This was on purpose - the default state of the Instance AI feature on instance creation is enabled. This ensures that on initial startup before the settings are populated on the store (via `getModuleSettings()`), we only consider the setting disabled when explicitly set via settings (requires user interaction)

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
Joco-95: I'd prefer leaving this as-is to avoid creating a separate controller for settings

packages/@n8n/api-types/src/frontend-settings.ts
Joco-95: `instanceAiEnabled` indicates "has the admin turned the feature on" (which can be toggled on or off), `enabled` is related to the module itself being enabled and available on the instance (including the Instance AI settings page which is available even when `instanceAiEnabled` is toggled off)

mutdmour: Seems like they can be just one. If module is available, always show settings page.

PR #27858, opened 2026-03-31
fix(core): Preserve manual scopes for inherited generic OAuth2 credentials

corazzione:
## Summary
- preserve saved `scope` values for credential types that inherit from editable generic OAuth2 credentials
- keep the old cleanup behavior for OAuth2 credentials whose scope is meant to stay hidden
- add a regression test covering a custom credential type extending `oAuth2Api`

## Testing
- added a focused unit test in `packages/cli/src/oauth/__tests__/oauth.service.test.ts`

Fixes #27780


guillaumejacquart:
Hey @corazzione 
Thanks for your contribution. We're aligned with your proposed solution. 
would you mind fixing the failing tests as pointed out by the CI. I think especially the backend ones are relevant: https://github.com/n8n-io/n8n/actions/runs/24657288850/job/72106505018?pr=27858

corazzione:
Addressed the backend unit test failure on this branch.

The new `getParentTypes()` call in `OauthService.getOAuthCredentials()` required a default mock in the existing `oauth.service` test setup, so I added a `[]` fallback in `beforeEach()` and kept the inherited-scope case covered by the dedicated test.

Validation run locally:
- `pnpm --dir packages/cli exec cross-env N8N_LOG_LEVEL=silent DB_SQLITE_POOL_SIZE=4 DB_TYPE=sqlite jest --config=jest.config.unit.js --runInBand src/oauth/__tests__/oauth.service.test.ts`
- `git diff --check`

guillaumejacquart (approved):

PR #27859, opened 2026-03-31
feat(core): Enable instance owner setup via environment variables

ireneea:
## Summary

- Allow configuring the instance owner (email, name, password hash) via environment variables during bootstrap
- Add `InstanceSettingsLoaderService` with a step-based architecture for startup configuration
- UI is disabled when the instance owner is configured via env vars:
image
 

https://linear.app/n8n/issue/LIGO-422

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/config/src/configs/logging.config.ts
```diff
@@ -28,7 +28,6 @@ export const LOG_SCOPES = [
 	'cron',
 	'community-nodes',
 	'chat-hub',
-	'instance-ai',
```
geemanjs, line 31: instance-ai was in here twice

packages/@n8n/config/src/configs/instance-settings-loader.config.ts
```diff
@@ -0,0 +1,25 @@
+import { Config, Env } from '../decorators';
+
+@Config
+export class InstanceSettingsLoaderConfig {
+	/** When true, update the instance owner with the provided values on every startup. */
+	@Env('INSTANCE_OWNER_OVERRIDE')
+	ownerOverride: boolean = false;
```
alielkhateeb: Do we really need an explicit `ownerOverride` boolean here? In practice, it has no effect unless both `ownerEmail` and `ownerPasswordHash` are provided. I’d suggest removing `ownerOverride` and deriving behavior from the presence of those two fields instead. This also avoids a confusing misconfiguration where users set `ownerEmail` and `ownerPasswordHash` but forget to set `ownerOverride: true`.

alielkhateeb: Ok responding to myself, I see the condition is ANDed here https://github.com/n8n-io/n8n/blob/b3b81b92da786b3e555cdc54b4c3fa2f682339f4/packages/cli/src/instance-settings-loader/loaders/owner.instance-settings-loader.ts#L56
So I'm a bit confused now, we will skip only if all three values are NOT provided? This made my brain raise so many questions sorry haha

geemanjs: You're in the process that I went through which makes me wonder if we "should" follow through with this behaviour or change it to something different. 

IMO if you set an env var with username / password we should ban you from updating it in the UI 🙄. This flag exists because the UI is still enabled and the database is the "source of the truth"

It exists purely for this case:
1. Set instance owner up via env vars `test / 1234`
2. Update it in the UI  `test / 4321`
3. When we "restart" what do we do? 

IMO we should not allow you to do no 2 if no 1 is set.. but me and @ireneea did have a long discussion on this with lots of back and forth and both ways satisfy different types of users.

alielkhateeb: I agree 100%. I see the options now:

1. **Owner credentials seeded on startup**
   - This is the current behaviour when `INSTANCE_OWNER_OVERRIDE=true`
   - Env vars are used to initialize/update the owner on every startup, but the UI remains editable afterward

2. **Locked mode**
   - Something like `N8N_OWNER_MANAGED_BY_ENV`
   - This would make the owner credentials read-only in the UI

I’m fine with either approach. That said, if we go with *locked mode*, I’d prefer renaming the env var to reflect **where the source of truth is** (i.e. managed by env), rather than using a more generic term like `override`.

More importantly, regardless of which option we choose, we should show this on the UI. For example: “_This instance owner is managed by environment configuration_”

If we go with locked mode, this becomes essential to avoid confusion.

We should also log some information during startup what is happening with owner credentials: e.g. "Owner credentials were overridden" or something 😄

alielkhateeb (approved):
Amazing work 🥳

PR #27876, opened 2026-04-01
chore(core): Enable TypeScript strict mode in packages/cli (no-changelog)

Matsuuu:
## Summary

> [!IMPORTANT]
> This PR is 90% generated by Claude with heavy instructions, goals and guard rails from me. This is to function as an initial run / PoC of porting the `cli` package to strict mode and instead of a gradual approach, uses a "throw everything but the kitchen sink" approach to getting the initial version out. 

> [!NOTE]
> After the review, changes and code corrections were done mostly by human review

Enables TypeScript `strict` mode in `packages/cli` by adding `"strict": true` to its `tsconfig.json` and resolving all resulting type errors across the package.

### Approach

Rather than papering over errors with broad casts, the fixes target root causes:

- **Express module augmentation** — Added `src/express-request-augmentation.d.ts` which augments the global `Express.Request` interface with `user?`, `authInfo?`, `browserId?`, and `rangeQuery?`. This eliminates the TypeScript contravariance issue that required every auth/range-query middleware to be cast `as RequestHandler`.
- **Definite assignment assertions (`!`)** — Added to TypeORM entity column properties (correct per TypeORM docs: ORM assigns after construction) and DI-initialized properties set in `init()` methods.
- **Type guards** — Replaced `as INodePropertyOptions[]` with a proper `isNodePropertyOptions()` type guard in the public API credentials service.
- **Interface alignment** — Fixed `IDeferredPromise<IRun | undefined>` vs `IDeferredPromise<IRun>` mismatch in `active-workflow-manager.ts`.
- **Miscellaneous strict fixes** — Narrower callback types, `as const` on string literals, proper `this` parameter typing, missing event map entries.

### What was removed

- All `as RequestHandler` / `as express.RequestHandler` casts at middleware call sites (controller registry, server, push handler, executions controller, credentials/workflow/mcp/test-runs controllers)
- `useUnknownInCatchVariables: false` override stays in place (a separate cleanup)

https://linear.app/n8n/issue/N8N-2847/enable-strict-in-tsconfig-in-cli

## Review / Merge checklist

- [ ] Tests included.


packages/cli/src/auth/auth.service.ts
ivov, line 1: ⚠️ Why are we broadening the type here and in many other spots? `AuthenticatedRequest` → `Request`

We may want to consider disabling `strictFunctionTypes` (this'd also allow removing unsafe assertions added in this PR) until we can tackle this properly.

packages/cli/src/executions/parse-range-query.middleware.ts
```diff
@@ -17,12 +17,8 @@ const isValid = (arg: JsonObject) => validate(arg, SCHEMA).valid;
...
  * Middleware to parse the query string in a request to retrieve a range of execution summaries.
  */
-export const parseRangeQuery = (
-	req: ExecutionRequest.GetMany,
-	res: Response,
-	next: NextFunction,
-) => {
-	const { limit, firstId, lastId } = req.query;
+export const parseRangeQuery = (req: Request, res: Response, next: NextFunction) => {
+	const { limit, firstId, lastId } = req.query as Partial<ExecutionRequest.QueryParams.GetMany>;
```
ivov: We're replacing typing with asserting here - likely papering over a mismatch. Same elsewhere, e.g. `filterListQueryMiddleware`, `paginationListQueryMiddleware`, etc.

packages/cli/src/modules/insights/database/entities/insights-shared.ts
```diff
@@ -15,8 +15,10 @@ export const PeriodUnitToNumber = {
 export type PeriodUnit = keyof typeof PeriodUnitToNumber;
 
 export type PeriodUnitNumber = (typeof PeriodUnitToNumber)[PeriodUnit];
-export const NumberToPeriodUnit = Object.entries(PeriodUnitToNumber).reduce(
-	(acc, [key, value]: [PeriodUnit, PeriodUnitNumber]) => {
+export const NumberToPeriodUnit = (
+	Object.entries(PeriodUnitToNumber) as Array<[PeriodUnit, PeriodUnitNumber]>
+).reduce(
```
ivov: `reduce` accepts a generic, or that doesn't typecheck? Same below.

packages/cli/src/workflows/workflow.service.ts
```diff
@@ -263,7 +263,9 @@ export class WorkflowService {
 			though. So to avoid leaking the information we just delete it.
 		*/
 		workflows.forEach((workflow) => {
-			delete workflow.shared;
+			if ('shared' in workflow) {
+				delete (workflow as ListQueryDb.Workflow.WithSharing).shared;
+			}
```
ivov: There must be a cleaner way?

packages/cli/src/express-request-augmentation.d.ts
ivov, line 1: Please double check if this is needed, we already have `APIRequest`, `AuthenticatedRequest`, etc.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -349,7 +349,7 @@ export class ActiveWorkflowManager {
 			const emit = (
 				data: INodeExecutionData[][],
 				responsePromise?: IDeferredPromise<IExecuteResponsePromiseData>,
-				donePromise?: IDeferredPromise<IRun | undefined>,
+				donePromise?: IDeferredPromise<IRun>,
```
ivov: ⚠️ Please double check this, are we sure this can't be `undefined`?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -315,7 +315,7 @@ export class ActiveWorkflowManager {
 					void executePromise.then((executionId) => {
 						this.activeExecutions
 							.getPostExecutePromise(executionId)
-							.then(donePromise.resolve)
+							.then((run) => run !== undefined && donePromise.resolve(run))
```
ivov: ⚠️ Please double check this, possible behavior change.

packages/cli/src/services/dynamic-node-parameters.service.ts
```diff
@@ -266,8 +262,9 @@ export class DynamicNodeParametersService {
 		const method = this.getMethod('actionHandler', handler, nodeType);
 		const workflow = this.getWorkflow(nodeTypeAndVersion, currentNodeParameters, credentials);
 		const thisArgs = this.getThisArg(path, additionalData, workflow);
+		const stringPayload = typeof payload === 'string' ? payload : undefined;
```
ivov: ⚠️ What about `IDataObject`? Silently becomes `undefined`?

packages/cli/src/scaling/pubsub/publisher.service.ts
ivov, line 1: ⚠️ Here and in lots of other spots we're suppressing errors with `!`. Should we consider relaxing `strictPropertyInitialization` instead until we can tackle this?

packages/cli/src/public-api/v1/shared/middlewares/global.middleware.ts
```diff
@@ -103,14 +103,19 @@ function tagMiddleware(
 
 export const apiKeyHasScope = (apiKeyScope: ApiKeyScope) => {
 	const middleware = Container.get(PublicApiKeyService).getApiKeyScopeMiddleware(apiKeyScope);
-	return tagMiddleware(middleware, apiKeyScope);
+	return tagMiddleware(middleware as (...args: unknown[]) => unknown, apiKeyScope);
```
ivov: Looks like we need to fix the typing instead of asserting here.

ivov (commented):
Thanks for starting on this! Couple comments

Matsuuu:
According to review did the next actions:

- Kept the express augmentation to help with handling middlewared requests.
- Kept AuthenticatedRequest and other in-house Request interfaces where applicable and makes sense
- Replace Type assertions with sane type guards
- Created a `invert` helper function te replace generic Reducers in e.g. `insights-shared.ts`. 
    - Typescript needs a little extra help here to understand the types and not create recursive types so a helper function was needed
- Relaxed strict mode a bit with setting `strictPropertyInitialization ` and `strictFunctionTypes` to false for the time being. Followups on these at a later date

packages/cli/src/express-request-augmentation.d.ts
ivov, line 1: Didn't we land on disabling `strictFunctionTypes`? The contravariance problem this augmentation was created for should be gone? Same for the `AuthenticatedRequest` → `Request` sweep?

packages/cli/src/commands/db/revert.ts
```diff
@@ -64,7 +64,7 @@ export async function main(
 	description: 'Revert last database migration',
 })
 export class DbRevertMigrationCommand {
-	private connection: Connection;
+	private connection!: Connection;
```
ivov: Didn't we land on `strictPropertyInitialization: false`? Are all these `!` still needed?

packages/cli/src/middlewares/list-query/filter.ts
```diff
@@ -1,21 +1,18 @@
...
 import { WorkflowFilter } from './dtos/workflow.filter.dto';
 
-export const filterListQueryMiddleware = async (
-	req: ListQuery.Request,
-	res: Response,
-	next: NextFunction,
-) => {
-	const { filter: rawFilter } = req.query;
+export const filterListQueryMiddleware: RequestHandler = async (req, res, next) => {
+	const listQueryReq = req;
```
ivov: TS requires these no-op renames or it's LLM noise?

packages/cli/src/webhooks/webhook-helpers.ts
```diff
@@ -78,6 +83,13 @@ import * as WorkflowHelpers from '@/workflow-helpers';
 import { WorkflowRunner } from '@/workflow-runner';
 import merge from 'lodash/merge';
 
+export function isWebhookOrWebhookOptionsRequest(
+	req: express.Request,
+): req is WebhookRequest | WebhookOptionsRequest {
+	// Mostly a dumb type guard to keep TS happy
+	return req.method === 'OPTIONS' || typeof req.path === 'string';
+}
+
```
ivov: Can we either make it a real guard or drop it and use an explicit cast at the single callsite? Leaving a misleading guard in the codebase is worse than either.

packages/cli/src/requests.ts
```diff
@@ -50,6 +51,15 @@ export namespace ListQuery {
...
 
+export function appendListQueryOptions(req: Request, options: ListQuery.Options) {
+	const listQueryReq = req as unknown as ListQuery.Request;
+	listQueryReq.listQueryOptions = { ...listQueryReq.listQueryOptions, ...options };
+}
+
+export function isListQueryRequest(req: Request): req is ListQuery.Request {
+	return 'listQueryOptions' in req;
+}
+
```
ivov, line 58: `isListQueryRequest` has no callers, can we drop it? It also sits next to `appendListQueryOptions`, which handles the same narrowing via `as unknown as ListQuery.Request`. 

Either the guard is the right tool (in which case `appendListQueryOptions` should use it instead of asserting) or it isn't, but having both in the same file is a mixed signal?

packages/cli/src/workflow-execute-additional-data.ts
```diff
@@ -519,7 +525,7 @@ export async function getBase({
 			envProviderState: EnvProviderState,
 			executeData?: IExecuteData,
 		) {
-			return await Container.get(TaskRequester).startTask(
+			return await Container.get(TaskRequester as ServiceIdentifier<TaskRequester>).startTask(
```
ivov: This suggests a DI typing gap.

packages/workflow/src/interfaces.ts
```diff
@@ -424,9 +424,11 @@ export type GenericValue = string | object | number | boolean | undefined | null
 export type CloseFunction = () => Promise<void>;
 
 export interface IDataObject {
-	[key: string]: GenericValue | IDataObject | GenericValue[] | IDataObject[];
+	[key: string]: IDataObjectValue;
 }
 
+export type IDataObjectValue = GenericValue | IDataObject | GenericValue[] | IDataObject[];
```
ivov: A bit ambivalent here. `IDataObject` is a scourge wreaking havoc throughout our types and we should aim to root it out, rather than make its internals public. Any chance we can prevent this?

ivov (commented):
Thanks for taking this on, getting closer!

packages/cli/src/requests.ts
Matsuuu: Yeah the check was called on call sites more, but I abstracted it to `appendListQueryOptions`. Just forgot to actually utilize the guard there.

packages/cli/src/webhooks/webhook-helpers.ts
Matsuuu: Removed and replaced with an explicit cast.

packages/cli/src/middlewares/list-query/filter.ts
Matsuuu: LLM Noise, resolved.

packages/cli/src/commands/db/revert.ts
Matsuuu: Missed this one in the sweep.

packages/cli/src/workflow-execute-additional-data.ts
Matsuuu: This was an unnecessary cast. Removed.

packages/cli/src/modules/instance-ai/web-research/cache.ts
```diff
@@ -39,7 +39,7 @@ export class LRUCache<T> {
 
 		// Evict oldest if at capacity
 		if (this.map.size >= this.maxEntries) {
-			const oldest = this.map.keys().next().value as string | undefined;
+			const oldest = this.map.keys().next().value;
```
ivov, line 42: Totally unrelated, but really nice LRU cache going on here! No doubly linked list 👌🏻

packages/cli/src/services/rate-limit.service.ts
```diff
@@ -93,6 +93,6 @@ export class RateLimitService {
 	}
 
 	private extractUserIdentifier(req: AuthenticatedRequest): string {
-		return `user:${req.user.id}`;
+		return `user:${req.user?.id}`;
```
ivov: Shouldn't an authenticated request always have a user ID?

packages/cli/src/task-runners/task-runner-module.ts
```diff
@@ -96,7 +97,7 @@ export class TaskRunnerModule {
 			'@/task-runners/task-managers/local-task-requester'
 		);
 		this.taskRequester = Container.get(LocalTaskRequester);
-		Container.set(TaskRequester, this.taskRequester);
+		Container.set(TaskRequester as ServiceIdentifier<TaskRequester>, this.taskRequester);
```
ivov: Is the assertion needed? Same question for any other new DI containers assertions.

packages/cli/src/controller.registry.ts
```diff
@@ -219,11 +219,8 @@ export class ControllerRegistry {
 	}
 
 	private createScopedMiddleware(accessScope: AccessScope): RequestHandler {
-		return async (
-			req: AuthenticatedRequest<{ credentialId?: string; workflowId?: string; projectId?: string }>,
-			res,
-			next,
-		) => {
+		return async (req, res, next) => {
+			if (!isAuthenticatedRequest(req)) throw new UnauthenticatedError();
```
ivov, line 223: Is this redundant with L224?

packages/@n8n/db/src/entities/types-db.ts
```diff
@@ -439,6 +439,10 @@ export type AuthenticatedRequest<
 	tokenGrant?: TokenGrant;
 };
 
+export function isAuthenticatedRequest(req: express.Request): req is AuthenticatedRequest {
+	return 'user' in req && req.user !== null;
+}
+
```
ivov, line 453: 1. Too permissive?
2. Misplaced? Why in `@n8n/db`?

packages/cli/src/push/push-helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import type { PushRequest, PushResponse, SSEPushRequest, WebSocketPushRequest } from './types';
+
+export function isPushRequest(req: Request): req is PushRequest {
+	return 'pushRef' in req.query && typeof req.query.pushRef === 'string';
+}
+
+export function isSSEPushRequest(req: Request): req is SSEPushRequest {
+	const hasWs = 'ws' in req;
+	return isPushRequest(req) && (!hasWs || (hasWs && req.ws === undefined));
```
ivov: Does `!hasWs || (hasWs && req.ws === undefined)` reduce to `!hasWs || req.ws === undefined`?

packages/frontend/editor-ui/src/app/utils/aiUtils.ts
```diff
@@ -203,6 +209,26 @@ const outputTypeParsers: {
...
+		return false;
+	}
+	if (
+		'id' in message &&
+		(!Array.isArray(message.id) ||
+			!message.id.every((item): item is string => typeof item === 'string'))
+	)
+		return false;
+
+	return true;
```
ivov: Is this correct? A message missing `lc`, `type` and `id` will pass straight through?

packages/cli/src/execution-lifecycle/execution-lifecycle-hooks.ts
```diff
@@ -115,10 +115,10 @@ class ModulesHooksRegistry {
 				case 'workflowExecuteResume':
 					hooks.addHandler(eventName, async function (workflowInstance, executionData) {
 						const context = {
-							type: 'workflowExecuteResume',
+							type: 'workflowExecuteResume' as const,
 							workflow: this.workflowData,
 							workflowInstance,
-							executionData,
+							executionData: executionData!,
```
ivov: Silencing with `!` hides the case where it genuinely is `undefined`, unless the type is inaccurate here.

packages/cli/src/requests.ts
```diff
@@ -50,6 +51,17 @@ export namespace ListQuery {
 	};
 }
 
+export function appendListQueryOptions(req: Request, options: ListQuery.Options) {
+	const listQueryReq = req as unknown as ListQuery.Request;
```
ivov: `as unknown as` undermines the point of the helper? Using the type predicate `isListQueryRequest` to narrow, then assign, might be cleaner.

packages/frontend/editor-ui/src/app/utils/aiUtils.ts
```diff
@@ -1,13 +1,19 @@
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
ivov: Hoisting `isMemoryMessage` above its use would avoid the disable.

packages/cli/src/push/index.ts
```diff
@@ -95,8 +101,15 @@ export class Push extends TypedEmitter<PushEvents> {
...
 			this.authService.createAuthMiddleware({ allowSkipMFA: false }),
-			(req: SSEPushRequest | WebSocketPushRequest, res: PushResponse) =>
-				this.handleRequest(req, res),
+			(req, res) => {
+				if (!isWebSocketPushRequest(req) && !isSSEPushRequest(req)) {
+					throw new BadRequestError('Request is not a PushRequest');
+				}
+				if (!isPushResponse(res)) {
+					throw new BadRequestError('Malformed response object');
+				}
```
ivov, line 111: If it's a malformed response it's not a bad request? Also isn't this covered by the prior check?

packages/workflow/src/interfaces.ts
ivov: Extracting the alias adds a hop without clarifying anything, and once exported from `packages/workflow` it becomes part of the public API surface that downstream packages and community nodes can depend on.

From a quick look the two callsites appear to be `aiUtils.ts` where we can use `unkonwn` and `job-processor.ts` where we should be able to use `IDataObject` or `IDataObject[]`.

ivov (commented):
Nearly there! 💪🏻

packages/cli/src/task-runners/task-runner-module.ts
Matsuuu: I swear I get blind to these after scanning through the code so many times. Good catch.

packages/cli/src/controller.registry.ts
Matsuuu: It acts as a type guard as `isAuthenticateRequest` uses `req is AuthenticatedRequest` as it's return value. It's a bit verbose but needed for type guarding.

packages/@n8n/db/src/entities/types-db.ts
Matsuuu: I feel that type guards should be where the type is defined. Do you have a recommendation on where this should be put?

As for permissive, We don't really have any other clear way to check when a request is a AuthenticatedRequest than this.

Open for suggestions 👍🏼

packages/cli/src/push/push-helpers.ts
Matsuuu: Silly me

packages/cli/src/push/index.ts
Matsuuu: Changed to InternalServerError as this is mostly to type guard the response, but it should never be anything else than that. So it's mostly functioning as an assertion

Matsuuu:
Merged master to this branch and seems that we have some fialing tests. I'll try and fix those too

packages/@n8n/db/src/entities/types-db.ts
ivov: Let's bring this in and tackle it separately 👍🏻

ivov (approved):

PR #27877, opened 2026-04-01
fix(core): Debounce per event type in pubsub

ivov:
## Summary

Fix shared debounce in pubsub subscriber that could cause one debounced command to silently drop another of a different type arriving within 300ms

n/a

## Review / Merge checklist

- [x] Tests included. 


guillaumejacquart (approved):
Thanks ! looks good 🚀

PR #27884, opened 2026-04-01
feat(core): Support projectId when creating workflow via public API

sandra0503:
## Summary

The public API `POST /workflows` endpoint now accepts an optional `projectId` in the payload, allowing callers to create workflows directly in a specific project instead of always defaulting to the user's personal project.

**Changes:**
- Added optional `projectId` to request body of `POST /workflows`
- Delegates to WorkflowCreationService
- Fix order in WorkflowCreationService: project authorization now runs before credential validation so callers receive the correct HTTP status when the target project is missing or inaccessible
- Returns 404 if the project doesn't exist, 403 if the user lacks access
- Falls back to the user's personal project when `projectId` is omitted

https://www.loom.com/share/0a0a796d89b446f6a0e4577531e6b7b6

**How to test:**
```bash
# Create workflow in a specific project
curl -X POST 'http://localhost:5678/api/v1/workflows?projectId=<PROJECT_ID>' \
  -H 'X-N8N-API-KEY: <API_KEY>' \
  -H 'Content-Type: application/json' \
  -d '{"name": "test", "nodes": [...], "connections": {}, "settings": {"executionOrder": "v1"}}'

# Without projectId — still creates in personal project
curl -X POST 'http://localhost:5678/api/v1/workflows' ...
```

https://linear.app/n8n/issue/LIGO-430

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/public-api/v1/handlers/workflows/workflows.handler.ts
```diff
@@ -28,26 +29,37 @@ import {
...
+			const project = projectId
+				? await projectService.getProjectWithScope(req.user, projectId, ['workflow:create'])
+				: await projectRepository.getPersonalProjectForUserOrFail(req.user.id);
+
+			if (!project) {
+				if (projectId && !(await projectRepository.exists({ where: { id: projectId } }))) {
+					throw new NotFoundError('Project not found');
+				}
+				throw new ForbiddenError('Forbidden');
+			}
```
geemanjs: I feel like this logic should sit in the service really - Whilst we're not here to do a full tidy up - For me a big anti-pattern is when a "controller" is doing a "repository" call.

geemanjs: I also wonder if the `projectId` should be in the body rather than in a query parameter 🤔 . But thats probably more up for debate!

sandra0503: Yes, `projectId` in body is actually how we do it when creating other resources too. Will change

sandra0503: I am currently working on reusing the service from internal API instead of the service in here

packages/cli/src/public-api/v1/handlers/workflows/spec/schemas/workflow.yml
```diff
@@ -13,6 +13,10 @@ properties:
   name:
     type: string
     example: Workflow 1
+  projectId:
```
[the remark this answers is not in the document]
sandra0503: we decided to move it to request body

packages/cli/src/public-api/v1/handlers/workflows/workflows.handler.ts
```diff
@@ -28,43 +24,11 @@ import {
...
 import { encodeNextCursor } from '../../shared/services/pagination.service';
-
 export = {
 	createWorkflow: [
 		apiKeyHasScope('workflow:create'),
 		async (req: WorkflowRequest.Create, res: express.Response): Promise<express.Response> => {
-			const workflow = req.body;
-
-			workflow.active = false;
-			workflow.versionId = uuid();
```
geemanjs, line 39: I wonder if we should add these back in when doing:
`const workflow = Object.assign(new WorkflowEntity(), rest);`

sandra0503: These fields are created the same way in the WorkflowCreationService, see my comment further down

packages/cli/src/workflows/workflow-creation.service.ts
sandra0503, line 77: See here

geemanjs (approved):

PR #27889, opened 2026-04-01
feat(instance-ai): Add planner sub-agent with progressive plan rendering (no-changelog)

OlegIvaniv:
## Summary

Adds a planner sub-agent that designs solution architectures for multi-step requests. Instead of the orchestrator building the task graph directly, it spawns an inline planner that discovers available resources (credentials, data tables, best practices), designs the blueprint, and shows it to the user for approval before any work begins.

The previous flow had two problems. First, the user stared at a loading planner for ~60s with zero visibility into what was being planned, everything appeared at once when `submit-blueprint` was called at the very end. Second, after the planner finished there was a ~33s orchestrator round-trip before the plan review UI appeared.

This PR fixes both. The planner now calls `add-plan-item` once per task as it designs each one, so the PlanReviewPanel populates progressively during planning. Plan approval happens inside the planner's own stream via Mastra HITL `suspend()`, eliminating the orchestrator round-trip entirely. When the user requests changes, feedback goes back to the same planner agent, it makes surgical edits with `remove-plan-item` / `add-plan-item` and re-submits, no restart needed.

The flow looks like:

```
orchestrator calls plan-with-agent
  planner discovers (best-practices, credentials, tables)
  planner calls add-plan-item(data-table-1)   -> PlanReviewPanel: [task 1]
  planner calls add-plan-item(workflow-1)      -> PlanReviewPanel: [task 1, task 2]
  planner calls add-plan-item(workflow-2)      -> PlanReviewPanel: [task 1, task 2, task 3]
  planner calls submit-plan                    -> suspend() -> approve/reject buttons appear
  user approves                                -> resume -> tasks dispatched
```

### New files

- `blueprint-accumulator.ts`: incremental task accumulation with upsert, dep reconciliation, removal, approval tracking
- `add-plan-item.tool.ts`: planner tool for progressive task emission + removal for revision
- `submit-plan.tool.ts`: Mastra HITL tool that suspends for user approval and returns the decision for the revision loop
- `plan-with-agent.tool.ts`: orchestration tool that spawns the planner sub-agent with discovery + planning + approval tools
- `plan-agent-prompt.ts`: planner system prompt (infer-don't-ask, incremental emission, revision handling)
- `blueprint.schema.ts`: zod schemas for blueprint items (workflow, data-table, research, delegate)

Beyond the new files, changes touch three areas: the `planItems` transport pipeline (api-types schemas, reducer, map-chunk validation, SSE events) so full task details flow from the planner's suspend payload to the frontend; the HITL plumbing (`consume-with-hitl` passes `maxSteps` to `resumeStream`, without this Mastra defaults to ~5 steps on resume and the planner stops early); and the frontend rendering (`AgentTimeline` does three-state planner display: loading/interactive/read-only, `PlanReviewPanel` gets a `loading` prop for the progressive preview).

## Review / Merge checklist

- [ ] Tests included. 


aalises (commented):
`findPlannerConfirmation()` is called quite some times in the Vue file (v-if, :key, :planned-tasks, :loading, :read-only, event handlers). Worth extracting to a single `computed` ref so it's not recomputing everything

packages/cli/src/modules/instance-ai/message-parser.ts
```diff
@@ -268,6 +268,19 @@ export function parseStoredMessages(
 		// in the assistant message's content
 	}
 
+	// Deduplicate user messages by content.
+	// The user message may be pre-saved for HITL refresh recovery and then
+	// saved again by Mastra's incremental save queue — remove the duplicate.
+	for (let i = messages.length - 1; i > 0; i--) {
+		if (
```
aalises: This only catches adjacent duplicates with identical content. If Mastra inserts an AI message between the pre-saved and stream-saved user messages, the duplicate would be there. 

Since the pre-save already uses a deterministic ID (`user-${messageId}`), deduplicating by ID would be more reliable maybe.

aalises (approved):

PR #27890, opened 2026-04-01
feat(core): Add lint rule to flag string literals in node inputs/outputs

garritfra:
## Summary

Adds a new ESLint rule `node-connection-type-literal` to `@n8n/eslint-plugin-community-nodes` that flags string literals (e.g. `'main'`) in node description `inputs` and `outputs` properties and requires using the `NodeConnectionTypes` enum from `n8n-workflow` instead. The rule provides an auto-fix for all known connection type values. It is enabled as `error` in both `recommended` configs.

The reverse mapping from string value to enum key is derived directly from `NodeConnectionTypes` at runtime, so it stays in sync automatically as new connection types are added.

https://linear.app/n8n/issue/NODE-4776/add-lint-rule-flag-string-literals-in-inputsoutputs-instead-of

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/eslint-plugin-community-nodes/src/rules/node-connection-type-literal.ts
```diff
@@ -0,0 +1,98 @@
...
+	findClassProperty,
+	findObjectProperty,
+	createRule,
+} from '../utils/index.js';
+
+// n8n-workflow's ESM dist uses bare module specifiers that Node's native ESM
+// loader cannot resolve. Loading via CJS (createRequire) sidesteps this.
+const { NodeConnectionTypes } = createRequire(import.meta.url)('n8n-workflow') as {
+	NodeConnectionTypes: Record<string, string>;
+};
```
garritfra, line 16: This one feels very weird to me, but Claude repeatedly assured me that this is the best way to access this CJS dependency 👀

garritfra: I'll research if we can drop the `"type": "module"` from the plugin package.

garritfra: The blast radius is significant. The key findings:

- All internal consumers (`nodes-base`, `nodes-langchain`, `node-cli`, `scan-community-package`) use ESM imports — named imports from `.mjs` files or TypeScript
- ESLint 9 flat config is ESM-first by design; CJS is non-standard there
- External community node developers get the ESM-based config via the `node-cli` scaffold template
- `import.meta.url` is already in our rule (`createRequire(import.meta.url)`) — that would need to change if we went CJS

So **switching the plugin to CJS is not viable** without a cascade of changes across multiple packages and external users.

That leaves the options as:
1. **Keep `createRequire`** (current state) — isolated, correct, no blast radius
2. **Fix `n8n-workflow`'s ESM build** with a post-build extension script — ~25 lines, fixes the root cause, no blast radius on the plugin side

Given all of this, the `createRequire` approach is clearly the right call for this PR. If someone wants to fix the ESM build of `n8n-workflow` properly, that's a separate self-contained change.

elsmr (approved):

PR #27910, opened 2026-04-01
feat: Computer use HITL confirmations in Instance AI

dlavrenuek:
## Summary

Implements `Computer Use` tool call confirmation in Instance AI. Because the `Computer Use` (old name `Local Gateway`) has own permission checks with configurable allow/deny logic stored on the computer the instance itself does not necessary know which tool calls must be confirmed. When `Computer Use` operates in `confirmationMode: instance` tool call execution which require user confirmation will fail with and the confirmation options will be included in the error response. These options contain randomly generated `decision id` which is stored locally. Once users confirms the request the tool call is executed again with passed `decision id`. This flow adds a level of protection to prevent the agent from executing tools while bypassing HITL logic (for example by passing "allow" on it's own)

Following diagram shows how the confirmation flow works:

```mermaid
sequenceDiagram
    participant FE as Browser (Frontend)
    participant SRV as n8n Server
    participant DB as Database
    participant D as fs-proxy Daemon

    Note over SRV: First invocation — tool execute() called by Mastra
    SRV->>D: callTool({ name, args }) via LocalGateway
    D-->>SRV: { isError: true, content: ["GATEWAY_CONFIRMATION_REQUIRED::..."] }
    SRV->>SRV: parse GatewayConfirmationRequiredPayload
    SRV->>DB: suspend() — persist agent snapshot + confirmation payload
    SRV-->>FE: SSE confirmation-request event<br/>{ inputType: "resource-decision", resourceDecision: { resource, description, options } }

    FE->>FE: show GatewayResourceDecision panel
    Note over FE: User clicks a decision button (e.g. Allow for session)
    FE->>SRV: POST /confirm/:requestId { approved: true, resourceDecisionToken: "<token>" }
    SRV->>DB: load agent snapshot, resume with resumeData

    Note over SRV: Second invocation — tool execute() called with resumeData
    SRV->>D: callTool({ name, args, _confirmation: "<token>" }) via LocalGateway
    D->>D: validate token, apply decision, execute tool
    D-->>SRV: { content: [...], isError: false }
    SRV-->>FE: SSE tool-result / text-delta events
```

https://linear.app/n8n/issue/NODE-4757/implement-n8n-gateway-tool-call-confirmation-in-n8n-instance-ai

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/fs-proxy/src/gateway-client.ts
```diff
@@ -395,28 +409,84 @@ export class GatewayClient {
...
+				);
 			}
 
 			if (rule === 'allow') continue;
 
-			const decision = await confirmResourceAccess(resource);
+			let decision: ResourceDecision;
+
+			if (resolvedDecision) {
+				decision = resolvedDecision;
```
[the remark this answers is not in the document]
dlavrenuek: in multi-resource tools like copy or move file the confirmation is asked for all resources at once, so also the decision is valid for all of these resources

packages/@n8n/fs-proxy/src/tools/types.ts
```diff
@@ -47,6 +47,18 @@ export type ResourceDecision =
...
+export const RESOURCE_DECISION_KEYS: ResourceDecision[] = [
+	'allowOnce',
+	'allowForSession',
+	'alwaysAllow',
+	'denyOnce',
+	'alwaysDeny',
+];
+
+/** Prefix used to signal a gateway confirmation is required (instance mode). */
+export const GATEWAY_CONFIRMATION_REQUIRED_PREFIX = 'GATEWAY_CONFIRMATION_REQUIRED::';
```
elsmr, line 60: Can we import the same one from `api-types`?

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -123,6 +124,7 @@ export class InstanceAiService {
 		string,
 		{ threadId: string; messageGroupId?: string; tracing: InstanceAiTraceContext }
 	>();
+	private readonly pendingSubAgentConfirmations = new Map<string, PendingConfirmation>();
```
elsmr: Is anything ever added to this map?

elsmr (commented):
Works very well in my testing 🎉 

I'm not sure if the `decision id` tokens add security. The LLM does not control what is sent to `POST /confirm/:requestId`, we make that call from the frontend code, and the frontend has access to the decision IDs anyway

packages/@n8n/fs-proxy/src/tools/types.ts
dlavrenuek: right not the `fs-proxy` does not have dependency on `api-types`, it only include `@n8n/mcp-browser` from the packages (which itself does not include any other n8n packages). I would spend some time first checking what impact importing this would have on the package size (not sure if we have f.e. tree shaking). Is it ok to do it in a separate PR?

packages/cli/src/modules/instance-ai/instance-ai.service.ts
dlavrenuek: that's a leftover from my first implementation, removed it.

elsmr (approved):

PR #27915, opened 2026-04-01
fix: Use different runtime for each agent call, refactor state management

yehorkardash:
## Summary

Previously, the Agent class reused a single AgentRuntime instance across all generate(), stream(), and resume() calls. This caused state from one run to leak into subsequent runs, and made concurrent calls unsafe.
This PR fixes the issue by:

- Splitting build output into a reusable AgentRuntimeConfig (built once, cached) and a per-run AgentRuntime (created fresh on each call)
- Tracking all active event buses in activeEventBuses so abort() correctly cancels every concurrent run
- Moving event handler registration to the Agent level and injecting them into each per-run event bus at creation time
- Sharing a single RunStateManager across per-run runtimes so resume() can find state from a prior stream()/generate() call
- Pre-fetching model cost once during build() and passing it through config instead of fetching per-run

## Review / Merge checklist
- [x] Tests included. 


bjorger (approved):
LGTM

PR #27882, opened 2026-04-01
feat(Alibaba Cloud Chat Model Node): Add new node

DawidMyslak:
## Summary

Add Alibaba Cloud Chat Model as a LangChain sub-node (`lmChatAlibabaCloud`) that connects to the AI Agent via DashScope's OpenAI-compatible API.

**[Loom Recording](https://www.loom.com/share/7d83f48e979c48c3a93f71ef747ed645)**

### What it does

- New credential (`alibabaCloudApi`) with API key and hidden DashScope international base URL
- New chat model node in `@n8n/nodes-langchain` using `ChatOpenAI` with custom base URL (same pattern as DeepSeek, xAI Grok, OpenRouter)
- Dynamic model loading via `GET /models` endpoint, default model: `qwen-plus`
- Configurable options: temperature, top-p, frequency/presence penalty, max tokens, response format (text/JSON), timeout, retries
- Alibaba Cloud SVG icon

### How to test

1. Get a DashScope API key from [Alibaba Cloud Model Studio](https://www.alibabacloud.com/help/en/model-studio/get-api-key)
2. Create a workflow with AI Agent node
3. Connect the new "Alibaba Cloud Chat Model" sub-node to the Agent's Chat Model input
4. Configure credentials with the API key
5. Select a Qwen model (e.g. `qwen-plus`, `qwen-turbo`, `qwen3-max`)
6. Run the workflow and verify the Agent can use the model for chat completions

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/NODE-4774

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md))
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [x] Tests included.
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

<details>
<summary>Implementation plan</summary>

# Alibaba Cloud Chat Model Node

## Context

Linear ticket [NODE-4774](https://linear.app/n8n/issue/NODE-4774) requests an AI Agent model node for Alibaba Cloud/Qwen. This is separate from [PR #27814](https://github.com/n8n-io/n8n/pull/27814), which adds a standalone node in `nodes-base`. Our node lives in `@n8n/nodes-langchain` and implements `supplyData` to connect with the AI Agent.

[...]

PR #27814, opened 2026-03-31
feat(Model Studio Node): New Node

dmolenaars:
## Summary

Add the initial version of the Alibaba Cloud Model Studio Node with the "Text", "Image", and "Video" resources which utilize the Qwen, Z-Image, and Wan models from Alibaba Cloud Model Studio.

### Features
**Text Operations**
Message: Chat completions with Qwen models (Qwen3 Max, Qwen3.5 Plus/Flash, and various MoE variants). Supports multi-turn conversations, web search, and configurable parameters (temperature, top-p, top-k, repetition penalty, etc.)
**Image Operations**
Analyze: Vision-language analysis using Qwen-VL models to answer questions about images
Generate: Text-to-image generation with Z-Image Turbo, Wan 2.6 T2I, and Qwen Image models. Supports multiple resolutions and prompt enhancement
Download: Save generated images as binary files
**Video Operations**
Text to Video: Generate videos from text prompts using Wan 2.6 T2V (720P/1080P, 2-15 seconds, single/multi-shot)
Image to Video: Animate still images with optional motion prompts using Wan 2.6 I2V models
Download: Save generated videos as MP4 binary files

## Review / Merge checklist

- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if the PR is an urgent fix that needs to be backported)

PR #27928, opened 2026-04-01
feat(Alibaba Cloud Model Studio Node): Add new node

DawidMyslak:
## Summary

Add the Alibaba Cloud Model Studio standalone node, based on community PR #27814 by @dmolenaars, integrated into core with review feedback applied.

**[Loom recording](https://www.loom.com/share/bfb7eddbe55d4a27a08120ac3970771e)**

### Features
- **Text**: Chat completions with Qwen models (Qwen3 Max, Qwen3.5 Plus/Flash, MoE variants). Supports multi-turn conversations, web search, and configurable parameters. Includes tool connector for AI Agent integration.
- **Image**: Vision-language analysis (Qwen-VL), text-to-image generation with auto-download (Z-Image Turbo, Wan 2.6 T2I, Qwen Image).
- **Video**: Text-to-video and image-to-video generation (Wan 2.6 T2V/I2V) with async polling, plus video download.

### Architecture notes
- **Model lists are currently static**, matching the original community PR. The DashScope native API does not expose a unified `/models` endpoint filtered by capability. However, the OpenAI-compatible API (`/compatible-mode/v1/models`) is available and can be used in a follow-up to switch to dynamic model loading — this is fully backward compatible since the parameter name (`modelId`) and type (`options`) won't change.
- **Telemetry**: The text model parameter uses `modelId` (matching OpenAI/Gemini/Anthropic convention) and the node is registered in `AI_VENDOR_NODE_TYPES` for analytics tracking.

### Changes from community PR
- Relocated node from `nodes-base` to `nodes-langchain/nodes/vendors/` (matching OpenAI, Gemini, Anthropic pattern)
- Unified credential to shared `AlibabaCloudApi` in `nodes-langchain/credentials/` (matching PR #27882's pattern) with region selector and workspace ID for Frankfurt
- Fixed codex identifier, removed unused `requestDefaults` and `routing` blocks
- Renamed snake_case parameter names to camelCase per n8n conventions
- Replaced `throw new Error()` with `NodeOperationError`
- Removed `extends IDataObject` from request body interface
- Set `usableAsTool: true`, human-readable model display names
- Extracted `versionDescription.ts` following vendor node pattern
- Added dynamic `inputs` for Tools connector on text/message operation
- Auto-download generated images as binary data (matching Gemini behavior)
- Removed redundant Download Image operation (generate now handles it)
- Renamed text model parameter to `modelId` for telemetry compatibility
- Added Frankfurt workspace ID validation with `UserError`
- Unified `simplifyOutput` and `pairedItem` formats across all operations
- Added node type constant to `AI_VENDOR_NODE_TYPES` for telemetry

https://linear.app/n8n/issue/NODE-4759

Based on community contribution: https://github.com/n8n-io/n8n/pull/27814

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/AlicloudModelStudio.node.ts
```diff
@@ -0,0 +1,12 @@
+import type { IExecuteFunctions, INodeExecutionData, INodeType } from 'n8n-workflow';
+
+import { router } from './actions/router';
+import { versionDescription } from './actions/versionDescription';
+
+export class AlicloudModelStudio implements INodeType {
```
RomanDavydchuk: The chat model node is named `LmChatAlibabaCloud`, i.e. with a `AlibabaCloud` and not `Alickoud`, should we keep this consistent?

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/transport/index.ts
```diff
@@ -0,0 +1,124 @@
...
+};
+
+const REGION_BASE_URLS: Record<string, string> = {
+	'ap-southeast-1': 'https://dashscope-intl.aliyuncs.com',
+	'us-east-1': 'https://dashscope-us.aliyuncs.com',
+	'cn-beijing': 'https://dashscope.aliyuncs.com',
+	'cn-hongkong': 'https://cn-hongkong.dashscope.aliyuncs.com',
+};
+
+export function getBaseUrl(credentials: IDataObject): string {
```
RomanDavydchuk, line 18: Should we reuse the helper introduced in the chat model node PR?

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/transport/index.ts
```diff
@@ -0,0 +1,124 @@
...
+		this,
+		'alibabaCloudApi',
+		options,
+	);
+
+	// DashScope returns `error: null` on success; normalize to `undefined` so downstream
+	// checks like `if (response.error)` don't see a key that looks like an error field.
+	if (response && response.error === null) {
+		response.error = undefined;
+	}
```
RomanDavydchuk: If the goal is to remove the `error` key entirely from the object, I think it's better to use `delete`

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/router.ts
RomanDavydchuk, line 1: While this is not something that should be changed, I think it's worth pointing out that the nodes that use this "router pattern" usually have a bit simpler logic for routing to different operations, instead of just `if/else`s, example: https://github.com/n8n-io/n8n/blob/e016b552106ece56d56aa4f82833acebdbb2a5ee/packages/%40n8n/nodes-langchain/nodes/vendors/OpenAi/v2/actions/router.ts#L30

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/text/message.operation.ts
```diff
@@ -0,0 +1,325 @@
...
+		operation: ['message'],
+		resource: ['text'],
+	},
+};
+
+export const description = updateDisplayOptions(displayOptions, properties);
+
+// All current dropdown models are multimodal. This list exists so that non-multimodal models
+// added later (e.g. via dynamic loading) are automatically routed to the text-generation endpoint.
+const MULTIMODAL_MODELS = [
```
RomanDavydchuk, line 251: I don't like hardcoded models here just like in the params, but I'm not sure if there is a good way to get around this, since this list is defined in the code and not a node parameter. We might need to just accept it and deal with it

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/text/message.operation.ts
```diff
@@ -0,0 +1,325 @@
...
+						displayName: 'Role',
+						name: 'role',
+						type: 'options',
+						options: [
+							{
+								name: 'User',
+								value: 'user',
+							},
+							{
+								name: 'System',
```
RomanDavydchuk: I'm not sure if it's a good idea to have `system` as one of the options here, given system prompts can only be specified once and not a part of the conversation, but rather exist separately. Other nodes usually provide a `System Message` as a separate option, example: https://github.com/n8n-io/n8n/blob/e016b552106ece56d56aa4f82833acebdbb2a5ee/packages/%40n8n/nodes-langchain/nodes/vendors/Anthropic/actions/text/message.operation.ts#L163

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/text/message.operation.ts
```diff
@@ -0,0 +1,325 @@
...
+	const output = isMultimodal
+		? (response.output?.choices?.[0]?.message?.content?.[0]?.text as string) || ''
+		: response.output?.text || response.output?.choices?.[0]?.message?.content || '';
+
+	return {
+		json: simplifyOutput
+			? { text: output }
+			: {
+					model,
+					response: output,
```
RomanDavydchuk: Let's maybe keep the field names the same in both simplified and "full" output?

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/versionDescription.ts
```diff
@@ -0,0 +1,69 @@
...
+	codex: {
+		alias: ['qwen', 'dashscope', 'alibaba', 'model studio', 'video', 'image'],
+		categories: ['AI'],
+		subcategories: {
+			AI: ['Agents', 'Miscellaneous', 'Root Nodes'],
+		},
+		resources: {
+			primaryDocumentation: [
+				{
+					url: 'https://www.alibabacloud.com/help/en/model-studio',
```
RomanDavydchuk: This is meant to be a link to n8n docs

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/text/message.operation.ts
RomanDavydchuk, line 1: One question regarding this operation as a whole: should we add tool support? Currently all standalone AI nodes have tool support for `message` operation, so I think it won't hurt, but will increase scope, obviously

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/image/generate.operation.ts
```diff
@@ -0,0 +1,228 @@
...
+import { updateDisplayOptions } from 'n8n-workflow';
+import { apiRequest } from '../../transport';
+import type { IImageOptions, IModelStudioRequestBody } from '../../helpers/interfaces';
+
+const properties: INodeProperties[] = [
+	{
+		displayName: 'Model',
+		name: 'imageModel',
+		type: 'options',
+		options: [
```
RomanDavydchuk: Same here, can this be not hardcoded?

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/image/generate.operation.ts
```diff
@@ -0,0 +1,228 @@
+import type { INodeProperties, IExecuteFunctions, INodeExecutionData } from 'n8n-workflow';
+import { updateDisplayOptions } from 'n8n-workflow';
+import { apiRequest } from '../../transport';
+import type { IImageOptions, IModelStudioRequestBody } from '../../helpers/interfaces';
+
+const properties: INodeProperties[] = [
+	{
+		displayName: 'Model',
+		name: 'imageModel',
```
RomanDavydchuk: Is it intentional that the name is not `modelId`? Because we rely on the model field to be named exactly `modelId` for telemetry: https://github.com/n8n-io/n8n/blob/e016b552106ece56d56aa4f82833acebdbb2a5ee/packages/workflow/src/telemetry-helpers.ts#L647

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/image/analyze.operation.ts
```diff
@@ -0,0 +1,159 @@
+import type { INodeProperties, IExecuteFunctions, INodeExecutionData } from 'n8n-workflow';
+import { updateDisplayOptions } from 'n8n-workflow';
+import { apiRequest } from '../../transport';
+import type { IModelStudioRequestBody } from '../../helpers/interfaces';
+
+const properties: INodeProperties[] = [
+	{
+		displayName: 'Model',
+		name: 'visionModel',
```
RomanDavydchuk: Same question as with `imageModel`

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/image/analyze.operation.ts
```diff
@@ -0,0 +1,159 @@
...
+				value: 'qwen3-vl-plus',
+				description: 'Enhanced vision-language model',
+			},
+		],
+		default: 'qwen3-vl-flash',
+		description: 'The model to use for image analysis',
+	},
+	{
+		displayName: 'Image URL',
+		name: 'imageUrl',
```
RomanDavydchuk, line 45: Two questions regarding this param:
1. Is it only possible to pass in the URL? Or can we pass base64 data too? If so, would be nice to support binary input here
2. Can multiple images be attached? If so, we might want to allow users to specify multiple URLs separated by commas

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/image/analyze.operation.ts
```diff
@@ -0,0 +1,159 @@
...
+	);
+
+	const output = (response.output?.choices?.[0]?.message?.content?.[0]?.text as string) || '';
+
+	return {
+		json: simplifyOutput
+			? { text: output }
+			: {
+					model: visionModel,
+					response: output,
```
RomanDavydchuk: Same as for `message` operation regarding field names

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.t2v.operation.ts
```diff
@@ -0,0 +1,207 @@
...
+	INodeProperties,
+} from 'n8n-workflow';
+import { NodeOperationError, updateDisplayOptions } from 'n8n-workflow';
+import { apiRequest, pollTaskResult } from '../../transport';
+import type { IModelStudioRequestBody } from '../../helpers/interfaces';
+
+export const properties: INodeProperties[] = [
+	{
+		displayName: 'Model',
+		name: 'videoModel',
```
RomanDavydchuk: Same question regarding model field name

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.t2v.operation.ts
```diff
@@ -0,0 +1,207 @@
...
+			},
+		],
+		default: 'single',
+		description: 'Whether to generate a single-shot or multi-shot narrative video',
+	},
+	{
+		displayName: 'Simplify Output',
+		name: 'simplifyOutput',
+		type: 'boolean',
+		default: false,
```
RomanDavydchuk: Usually we have `simpify` with `default: true`, any reason why it's `false` here?

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.t2v.operation.ts
```diff
@@ -0,0 +1,207 @@
...
+				default: true,
+				description: 'Whether to generate audio for the video',
+			},
+			{
+				displayName: 'Audio URL',
+				name: 'audioUrl',
+				type: 'string',
+				default: '',
+				placeholder: 'https://example.com/audio.mp3',
+				description: 'Public URL to an audio file to use for the video',
```
RomanDavydchuk: Can we support binary data here as well?

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.i2v.operation.ts
```diff
@@ -0,0 +1,223 @@
...
+	INodeProperties,
+} from 'n8n-workflow';
+import { NodeOperationError, updateDisplayOptions } from 'n8n-workflow';
+import { apiRequest, pollTaskResult } from '../../transport';
+import type { IModelStudioRequestBody } from '../../helpers/interfaces';
+
+export const properties: INodeProperties[] = [
+	{
+		displayName: 'Model',
+		name: 'i2vModel',
```
RomanDavydchuk: Same as other places

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/video/generate.i2v.operation.ts
```diff
@@ -0,0 +1,223 @@
...
+		description: 'The model to use for image-to-video generation',
+	},
+	{
+		displayName: 'Image URL',
+		name: 'imgUrl',
+		type: 'string',
+		required: true,
+		default: '',
+		placeholder: 'https://example.com/image.png',
+		description: 'The URL of the first-frame image to generate video from',
```
RomanDavydchuk, line 54: I assume here we can only support 1 image, given the description says `The URL of the first-frame image to generate video from`, but can we support binary data here?

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.i2v.operation.ts
```diff
@@ -0,0 +1,223 @@
...
+			},
+		],
+		default: 'single',
+		description: 'Whether to generate a single-shot or multi-shot narrative video',
+	},
+	{
+		displayName: 'Simplify Output',
+		name: 'simplifyOutput',
+		type: 'boolean',
+		default: false,
```
RomanDavydchuk: Same question as in text to video: why `false` and not `true`

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.i2v.operation.ts
```diff
@@ -0,0 +1,223 @@
...
+				default: true,
+				description: 'Whether to generate audio for the video',
+			},
+			{
+				displayName: 'Audio URL',
+				name: 'audioUrl',
+				type: 'string',
+				default: '',
+				placeholder: 'https://example.com/audio.mp3',
+				description: 'Public URL to an audio file to use for the video',
```
RomanDavydchuk: Same question as in text to video operation

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/download.operation.ts
RomanDavydchuk, line 1: Code here looks good, but I do wonder if it would be better to do something similar to the generate image operation and add a toggle like "Download Video" to I2V and T2V operations? I know that the Gemini node follows the same pattern with a separate operation, but a toggle might be a better UX. Could be something to ask product/design

RomanDavydchuk (changes requested):
Looks good mostly, don't see any issues that would stop the node from working, but I did leave quite a few questions and comments regarding UX and other small implementation details. Also, I still wasn't able to create an account to manually test, so I'll trust the Loom video you prodided

Regarding the model selection being hardcoded: I'm not a fan since today new models tend to come out rather frequently, so it's a pain to maintain it. Maybe we should use the compatible API, like you mentioned, since it's already in use in the chat model node? And maybe then determine if a model is for text, image or video generation based on the name? Lastly, there is an array of multimodal models in the code and it would also be hard to maintain, maybe even harder, since it's buried in the code and not a node parameter. I'm not sure whether there is a good way to get rid of it, the only idea I have is to "flip the script" and have a list of models that are _not multimodal_, since going new models are multimodal mostly, I assume

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/text/message.operation.ts
DawidMyslak: I think I saw OpenAI doing the exact same thing, but generally agree with you, will change.
Screenshot 2026-04-07 at 15 05 07

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/image/analyze.operation.ts
DawidMyslak: Good question, I trusted the initial PR to only support a single URL. Let me explore more.

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/generate.t2v.operation.ts
DawidMyslak: Good catch, leftover from the original PR, let me tweak.

packages/@n8n/nodes-langchain/nodes/vendors/AlicloudModelStudio/actions/video/download.operation.ts
DawidMyslak: I see OpenAI node is doing something similar around binary vs url output. Will try to follow.

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/video/generate.t2v.operation.ts
```diff
@@ -0,0 +1,278 @@
...
+				type: 'string',
+				default: 'audio',
+				placeholder: 'e.g. audio',
+				hint: 'The name of the input field containing the binary audio data',
+				displayOptions: {
+					show: {
+						audioInputType: ['binary'],
+					},
+				},
+			},
```
RomanDavydchuk, line 163: We have [this](https://github.com/n8n-io/n8n/blob/4b3b40e2385b9e8be68a8f7f05e7aae4dedaf847/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInputWrapper.vue#L132) check on the frontend to display this tooltip:

Image

This won't work now because the name does not have `binaryPropertyName` (the first letter is upper case), but you can set `typeOptions.binaryDataProperty = true` to achieve the same effect

Also, we usually use `data` as the default value for binary property name, but it's not a problem to have it something else

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/actions/video/generate.i2v.operation.ts
```diff
@@ -0,0 +1,340 @@
...
+				type: 'string',
+				default: 'audio',
+				placeholder: 'e.g. audio',
+				hint: 'The name of the input field containing the binary audio data',
+				displayOptions: {
+					show: {
+						audioInputType: ['binary'],
+					},
+				},
+			},
```
RomanDavydchuk, line 212: Same here regarding `audioBinaryPropertyName`, although `audio` as a default binary property name makes sense here, because we also have an image that can also come from binary, and it does already have `data` as default binary property name. So perhaps for the sake of consistency we can leave `default: 'audio'` for this and for the previous occurrence 🤷

packages/@n8n/nodes-langchain/nodes/llms/LmChatAlibabaCloud/alibaba-cloud-base-url.ts
```diff
@@ -1,14 +1,28 @@
...
 }
 
-export const BASE_URL_EXPRESSION = `={{ (${getBaseUrl.toString()})($credentials.region, $credentials.workspaceId) }}`;
+export const BASE_URL_EXPRESSION = `={{ (() => {
+	const hosts = ${JSON.stringify(REGION_BASE_HOSTS)};
+	const region = $credentials.region;
+	if (region === 'eu-central-1') {
+		return 'https://' + $credentials.workspaceId + '.eu-central-1.maas.aliyuncs.com/compatible-mode/v1';
+	}
+	return (hosts[region] || hosts['ap-southeast-1']) + '/compatible-mode/v1';
```
RomanDavydchuk: nit: can interpolate `COMPATIBLE_MODE_SUFFIX` here, since stringified `REGION_BASE_HOSTS` is already used in this expression

RomanDavydchuk (changes requested):
Looks really good, thanks for addressing the feedback! Left two comments regarding binary properties that are quick to fix and one small nit 😄

RomanDavydchuk (approved):
Tested locally and works nicely 🎉

PR #27929, opened 2026-04-01
fix(editor): Skip only current step when clicking Later in workflow setup

aalises:
## Summary

Fixes a P0 bug in Instance AI workflow setup where clicking "Later" on a credential setup step
skips **all** remaining steps instead of just the current one.

**Root cause:** `handleLater()` called `confirmAction(requestId, false)` unconditionally, which
sends `approved: false` to the backend. The backend treats this as a terminal exit and returns
`{ deferred: true }` without re-suspending for remaining steps.

**Fix:** In wizard mode, "Later" now:
1. Clears the selection for the current card (explicit skip)
2. Advances to the next step if there are more
3. Auto-applies the partial set if it was the last step and some cards were completed
4. Only defers the entire setup when no cards have been completed at all

No backend changes needed — the backend already supports partial applies (`partial: true` with
`skippedNodes`) via State 4 in `setup-workflow.tool.ts`.

https://www.notion.so/n8n/3355b6e0c94f8016ad04e72e8b3b1f05

## Review / Merge checklist

- [ ] Tests included.


mutdmour (commented):
looks good. can we add some tests?

mutdmour (dismissed):

mutdmour (dismissed):

mutdmour (approved):

PR #27939, opened 2026-04-02
fix(core): Fixing issues across instance ai (no-changelog)

mutdmour:
## Summary
- When the AI service proxy is enabled (license + N8N_AI_ASSISTANT_BASE_URL),
the Model, Search, and Sandbox settings sections are hidden from the UI and
their fields are stripped from update requests. When the proxy is not
configured, all settings remain visible and functional.
- Fix bug where preview does not load if workflow name not returned in tool call. Fall back to 'workflowId' when workflowId is present without a name. Then enrich from the workflow store via a workflowNameLookup callback so all consumers see the real name
Also fixed the inline artifact card in chat messages (AgentTimeline) to resolve names from the enriched registry
- The execution adapter was passing `accessibleWorkflowIds` to
`findManyByRangeQuery`, but that field is not part of
`ExecutionSummaries.RangeQuery`. Without `user`/`sharingOptions`, the
query builder fell into `WHERE 1 = 0`, returning zero results every
time — causing the agent to loop on "Checking executions" without
finding anything.

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/instance-ai/instance-ai-settings.service.ts
```diff
@@ -143,6 +150,9 @@ export class InstanceAiSettingsService {
 	async updateAdminSettings(
 		update: InstanceAiAdminSettingsUpdateRequest,
 	): Promise<InstanceAiAdminSettingsResponse> {
+		if (this.aiService.isProxyEnabled()) {
+			this.stripProxyManagedAdminFields(update);
+		}
```
aalises, line 156: When the proxy is enabled, we would silently drop fields from the update request and returns 200 OK. An API consumer (or future UI code) sending `credentialId` or `sandboxEnabled` will get a success response with no indication that their changes were ignored. We could maybe return some warning in the response payload or rejecting with a 4xx

Would avoid confusing "update succeeded but nothing changed" scenario.

mutdmour: Sure. Will address

aalises (approved):

PR #27944, opened 2026-04-02
feat(core): Add JTI store with atomic consume and cleanup job for token exchange

afitzek:
## Summary

Replaces the skeleton `JtiStoreService` with a database-backed implementation that prevents token replay attacks in the token exchange flow.

**Key changes:**

- **`TokenExchangeJti` entity + migration** — New `token_exchange_jti` table with columns `jti` (PK), `expiresAt`, `createdAt`. Supports both PostgreSQL and SQLite.
- **`TokenExchangeJtiRepository`** — Atomic `atomicConsume(jti, expiresAt)` method using `INSERT ... ON CONFLICT DO NOTHING` (Postgres CTE) / `INSERT OR IGNORE` + `changes()` (SQLite transaction). Returns `true` on first use, `false` on replay. Also provides `deleteExpiredBatch(batchSize)` for bounded cleanup.
- **`JtiStoreService`** — Delegates to the repository with a 60-second grace period on expiry to account for clock skew.
- **`JtiCleanupService`** — Periodic background job that deletes expired JTI rows in configurable batches. Loops until fewer than `batchSize` rows are deleted per iteration. Graceful shutdown via `@OnShutdown()`.
- **`EmbedAuthController`** — Added IP-based rate limiting (20 req/min) to GET and POST login endpoints.
- **Config** — Two new env vars: `N8N_TOKEN_EXCHANGE_JTI_CLEANUP_INTERVAL_SECONDS` (default: 60) and `N8N_TOKEN_EXCHANGE_JTI_CLEANUP_BATCH_SIZE` (default: 1000).

**Test coverage:**
- Unit tests for `JtiStoreService`, `JtiCleanupService`, and `TokenExchangeJtiRepository` (mock-based, both Postgres and SQLite paths)
- Integration test (`token-exchange-jti.repository.integration.test.ts`) validating `atomicConsume` and `deleteExpiredBatch` against a real database

https://linear.app/n8n/issue/IAM-461

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/db/src/migrations/common/1775116241000-CreateTokenExchangeJtiTable.ts
```diff
@@ -0,0 +1,17 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const tableName = 'token_exchange_jti';
+
+export class CreateTokenExchangeJtiTable1775116241000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(tableName).withColumns(
+			column('jti').varchar(255).primary,
+			column('expiresAt').timestamp().notNull,
+			column('createdAt').timestamp().notNull,
```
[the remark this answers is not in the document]
afitzek: atomicConsume inserts the createdAt date in its query.

packages/cli/src/modules/token-exchange/token-exchange.module.ts
```diff
@@ -23,5 +28,8 @@ export class TokenExchangeModule implements ModuleInterface {
 
 		await import('./token-exchange.controller');
 		await import('./controllers/embed-auth.controller');
+
+		const { JtiCleanupService } = await import('./services/jti-cleanup.service');
+		Container.get(JtiCleanupService).startCleanup();
```
guillaumejacquart: I'm wondering if we should do this only for the leader. the cleanup seems safe, but rather to prevent waste of db resources

packages/cli/test/integration/token-exchange/token-exchange-jti.repository.integration.test.ts
```diff
@@ -0,0 +1,101 @@
...
+			expect(row!.jti).toBe('jti-persisted');
+		});
+
+		it('should allow consuming different JTIs independently', async () => {
+			const result1 = await repository.atomicConsume('jti-a', futureDate);
+			const result2 = await repository.atomicConsume('jti-b', futureDate);
+
+			expect(result1).toBe(true);
+			expect(result2).toBe(true);
+		});
```
guillaumejacquart, line 55: Maybe we should add a test that checks atomiticy under concurrency situation
Like
```ts
it('should allow only one consumer when called concurrently with the same JTI', async () => {
  const results = await Promise.all([
    repository.atomicConsume('jti-race', futureDate),
    repository.atomicConsume('jti-race', futureDate),
  ]);

  const consumed = results.filter(Boolean);
  expect(consumed).toHaveLength(1);
});
```

packages/cli/src/modules/token-exchange/database/repositories/token-exchange-jti.repository.ts
```diff
@@ -0,0 +1,126 @@
...
+	 * changes()) because the sqlite-pooled driver dispatches WITH-prefixed
+	 * queries via `.all()` instead of `.run()`, which breaks DML execution.
+	 */
+	async atomicConsume(jti: string, expiresAt: Date): Promise<boolean> {
+		const esc = (name: string) => this.manager.connection.driver.escape(name);
+		const tableName = this.escapedTableName;
+
+		if (this.dbType === 'postgresdb') {
+			const [row] = (await this.query(
+				`WITH ins AS (
```
guillaumejacquart, line 37: This would not have worked ?
```ts
const result = await this.createQueryBuilder()
  .insert()
  .into(TokenExchangeJti)
  .values({ jti, expiresAt, createdAt: new Date() })
  .orIgnore()
  .execute();

return result.identifiers.length > 0;
```

guillaumejacquart (commented):
minor comments

packages/cli/src/modules/token-exchange/token-exchange.module.ts
afitzek: I thought about this as well. My thinking was, that since it doesn't really matter, it might be easier to just have every instance run a cleanup. But I'm happy to make this a leader job.

guillaumejacquart: I guess that's fine, keeps the code simpler, and the table is supposed to be quite small scale anyway. I was also hesitating with suggesting the leader check or not. 👍

packages/cli/src/modules/token-exchange/database/repositories/token-exchange-jti.repository.ts
afitzek: I think in sqlite this is not reliable, to determine if a value was inserted or not. Thats why I wanted to make sure and opted for the raw SQL queries.

guillaumejacquart: AH yes, that does ring a bell for sqlite. Then i case no much point in have 2 different patterns for postgres and sqlite. ok for me

packages/cli/test/integration/token-exchange/token-exchange-jti.repository.integration.test.ts
afitzek: Done — added the concurrency test:

```ts
it('should allow only one consumer when called concurrently with the same JTI', async () => {
  const results = await Promise.all([
    repository.atomicConsume('jti-race', futureDate),
    repository.atomicConsume('jti-race', futureDate),
  ]);

  const consumed = results.filter(Boolean);
  expect(consumed).toHaveLength(1);
});
```

Passes on SQLite. The unique PK constraint on `jti` guarantees exactly one of the concurrent inserts wins.

packages/cli/src/modules/token-exchange/token-exchange.module.ts
afitzek: Implemented the leader-only pattern following the established convention (same as `ExecutionsPruningService`, `WorkflowHistoryCompactionService`, etc.):

- Added `@OnLeaderTakeover()` on `startCleanup()`
- Added `@OnLeaderStepdown()` on new `stopCleanup()` method
- Added `init()` that checks `instanceSettings.isLeader` before starting
- Module now calls `init()` instead of `startCleanup()` directly
- Added unit tests for the leader/non-leader behavior

guillaumejacquart (dismissed):

guillaumejacquart (approved):

PR #27951, opened 2026-04-02
fix(Anthropic Node): Use models endpoint for credential test

guillaumejacquart:
## Summary

The Anthropic credential test was failing with "The resource you are requesting could not be found" even with a valid API key. The root cause is that the test used `POST /v1/messages` with a hardcoded model (`claude-haiku-4-5-20251001`) which returns 404 for users whose API plan doesn't include that specific model.

This switches the credential test to `GET /v1/models`, a read-only endpoint that only requires a valid API key — no model access needed. This is consistent with how OpenAI and Mistral credentials are tested. The fix is zero-cost (no tokens consumed) and future-proof (no hardcoded model to go stale).

- https://linear.app/n8n/issue/IAM-445
- Fixes https://github.com/n8n-io/n8n/issues/27491

## Review / Merge checklist

- [x] Tests included.


phyllis-noester (approved):
nice 👍

PR #27957, opened 2026-04-02
fix: Add credential auth and test for PostHog, NASA, Peekalink, Clearbit, Uptime Robot

BerniWittmann:
## Summary

Add `authenticate` and `test` properties to 5 credential files that were missing them. Each credential's auth pattern was verified against the node's GenericFunctions to ensure the `authenticate` property matches the actual API auth mechanism.

| Credential | Linear Issue | Auth Method | Test Endpoint | Custom Auth? | Auth Source | API Docs |
|------------|-------------|-------------|---------------|-------------|------------|----------|
| PostHog API | [NODE-2747](https://linear.app/n8n/issue/NODE-2747) | Body `api_key` (project token) | `POST /decide/` | No | [GenericFunctions.ts](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/PostHog/GenericFunctions.ts) | [PostHog API](https://posthog.com/docs/api) |
| NASA API | [NODE-2804](https://linear.app/n8n/issue/NODE-2804) | Query param `api_key` | `GET /planetary/apod` | No | [GenericFunctions.ts](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Nasa/GenericFunctions.ts) | [NASA API](https://api.nasa.gov/) |
| Peekalink API | [NODE-2698](https://linear.app/n8n/issue/NODE-2698) | Bearer header | `POST /` | No | [Peekalink.node.ts](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Peekalink/Peekalink.node.ts) | [Peekalink API](https://www.peekalink.io/) |
| Clearbit API | [NODE-2683](https://linear.app/n8n/issue/NODE-2683) | Bearer header | `GET /v2/companies/find` | No | [GenericFunctions.ts](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Clearbit/GenericFunctions.ts) | [Clearbit API](https://dashboard.clearbit.com/docs) |
| Uptime Robot API | [NODE-2732](https://linear.app/n8n/issue/NODE-2732) | Form body `api_key` | `POST /v2/getAccountDetails` | No | [GenericFunctions.ts](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/UptimeRobot/GenericFunctions.ts) | [Uptime Robot API](https://uptimerobot.com/api/) |

https://linear.app/n8n/issue/NODE-4780
https://linear.app/n8n/issue/NODE-2621
Closes https://linear.app/n8n/issue/NODE-3093

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/credentials/PeekalinkApi.credentials.ts
+++ b/packages/nodes-base/credentials/PeekalinkApi.credentials.ts
@@ -1,4 +1,9 @@
-import type { ICredentialType, INodeProperties } from 'n8n-workflow';
+import type {
+	IAuthenticateGeneric,
+	ICredentialTestRequest,
+	ICredentialType,
+	INodeProperties,
+} from 'n8n-workflow';
 
 export class PeekalinkApi implements ICredentialType {
 	name = 'peekalinkApi';
@@ -16,4 +21,24 @@ export class PeekalinkApi implements ICredentialType {
 			default: '',
 		},
 	];
+
+	authenticate: IAuthenticateGeneric = {
+		type: 'generic',
+		properties: {
+			headers: {
+				'X-API-Key': '={{$credentials.apiKey}}',
+			},
+		},
+	};
+
+	test: ICredentialTestRequest = {
+		request: {
+			baseURL: 'https://api.peekalink.io',
+			url: '/',
+			method: 'POST',
+			body: {
+				link: 'https://www.example.com',
+			},
+		},
+	};
 }
--- a/packages/nodes-base/credentials/PeekalinkApi.credentials.ts
+++ b/packages/nodes-base/credentials/PeekalinkApi.credentials.ts
@@ -26,7 +26,7 @@ export class PeekalinkApi implements ICredentialType {
 		type: 'generic',
 		properties: {
 			headers: {
-				'X-API-Key': '={{$credentials.apiKey}}',
+				Authorization: '=Bearer {{$credentials.apiKey}}',
 			},
 		},
 	};
--- a/packages/nodes-base/credentials/ClearbitApi.credentials.ts
+++ b/packages/nodes-base/credentials/ClearbitApi.credentials.ts
@@ -1,4 +1,9 @@
-import type { ICredentialType, INodeProperties } from 'n8n-workflow';
+import type {
+	IAuthenticateGeneric,
+	ICredentialTestRequest,
+	ICredentialType,
+	INodeProperties,
+} from 'n8n-workflow';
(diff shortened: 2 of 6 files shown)


packages/nodes-base/credentials/PeekalinkApi.credentials.ts
```diff
@@ -16,4 +21,24 @@ export class PeekalinkApi implements ICredentialType {
 			default: '',
 		},
 	];
+
+	authenticate: IAuthenticateGeneric = {
+		type: 'generic',
+		properties: {
+			headers: {
+				'X-API-Key': '={{$credentials.apiKey}}',
```
elsmr: According to the curl example I got from them it should be `Authorization: Bearer`

elsmr (changes requested):
Tested and most of them work great but:

- Peekalink didn't work, I added a new API key to Bitwarden that works with curl, but not in n8n
- Clearbit I couldn't test because they stopped offering free accounts

The Peekalink one should be fixed, for Clearbit let's just double check that our credential matches the API docs

packages/nodes-base/credentials/PeekalinkApi.credentials.ts
BerniWittmann: The problem is, that the node Implementation is outdated as well (using X-API-KEY header). 

So the change matches the nodes actual (but wrong) behavior. The question now is, how do we want to handle such cases. I propose to adjust the Batching Logic to flag these inconsistencies/outdated nodes to then decide how to handle these (e.g. leave out of batching process). Does that sound good, or do you have other ideas? 

For the current PR i will push a quick fix

elsmr: Yeah sounds good, I think for Cloud tools like Peekalink we should adopt the new behavior because the old version is broken for everyone.

But if it's a self hosted tool, multiple API versions can be out there, so we should keep the old behavior and add optional support for the new behavior.

elsmr (approved):

packages/nodes-base/credentials/PeekalinkApi.credentials.ts
BerniWittmann: Sounds good :+1: 
Generally I'd also want to avoid making too many direct changes to nodes and rather focus on adding missing  credential testings and avoid scope creep :)
Thanks for your help

PR #27959, opened 2026-04-02
fix(editor): Await confirmAction in handleLater to prevent stuck sub-agents (no-changelog)

aalises:
## Summary
- **Fix fire-and-forget `handleLater`** in `InstanceAiCredentialSetup` and `InstanceAiWorkflowSetup`: the `confirmAction` API call was not awaited (`void store.confirmAction(...)`), so the UI showed "Skipped" even if the backend never received the confirmation — leaving the sub-agent permanently stuck at `waitForConfirmation`
- Now `handleLater` awaits the API call and rolls back UI state on failure, matching `handleContinue`/`handleApply`
- Added backend debug/warn logging to `resolveConfirmation` for diagnosability


Cadiac (approved):
👍

PR #27967, opened 2026-04-02
fix(instance-ai): reduce memory footprint (no-changelog)

OlegIvaniv:
## Summary

Fixes several memory and stability issues in instance-ai discovered during heap profiling.

### Memory fixes

- The cached Mastra singleton in `registerWithMastra()` and `ensureMastraRegistered()` was passing agents to the constructor, which stored the first agent's closures (tools, prompts, thread context) permanently in Mastra's internal `#agents` dict. In a multi-user environment this means User A's agent gets pinned in the singleton shared by all subsequent users. Fixed by creating Mastra with only storage. Agents get the back-reference for suspend/resume via `__registerMastra()` without being added to `#agents`.

- `InstanceAiAdapterService` was calling `collectTypes()` which triggers `postProcessLoaders()`, an expensive rebuild of the full node type registry (~31 MB). Switched to reading from the static JSON file that `FrontendService` writes at startup.

- Brotli compressors from Express response compression were costing ~8.6 MB of native memory each, retained via keep-alive connections. Added a controller middleware that strips `br` from `Accept-Encoding` to downgrade to gzip, and disabled compression entirely on instance-ai SSE streams.

- Upgraded Mastra packages to get lazy tiktoken loading (avoids ~15 MB WASM allocation on import) and a token growth fix. Also patched the module loader to prevent Mastra from eagerly importing tiktoken and tokenizer (~60 MB combined).

- LangSmith tracing now uses a `gzipFetch` wrapper that forces gzip encoding for both proxy and direct client paths, avoiding the brotli decompressor overhead on the tracing side.

### Other fixes

- LangSmith `patchRun` was logging noisy 409 "payloads already received" errors on retries. The `gzipFetch` wrapper now returns a synthetic 200 for 409 responses since the data already landed.

- Mastra's `wrapWithReadTracker` calls `stat()` before file writes and catches `FileNotFoundError` for new files. But `DaytonaFilesystem.stat()` was throwing Daytona SDK's `DaytonaNotFoundError` which Mastra didn't recognize, causing `write_file` to fail on new files that don't exist yet. Now translates Daytona 404s to Mastra's `FileNotFoundError`.

- The regex-based JSON syntax highlighter in `ToolResultJson.vue` was vulnerable to catastrophic backtracking on large tool outputs. Replaced with a structural character walker.

## Review / Merge checklist

- [ ] Tests included. 


aalises (commented):
[Minor] `browserCredentialSetupInputSchema` and `browserCredentialSetupResumeSchema` are actually the schemas for `createPauseForUserTool()`, not for the browser-credential-setup tool itself (which uses `browserCredentialSetupToolInputSchema`). 

Maybe the naming will mislead maintainers into thinking these schemas belong to `createBrowserCredentialSetupTool()`.

I would maybe rename to `pauseForUserInputSchema` / `pauseForUserResumeSchema` to match the tool they serve.

aalises (approved):

PR #27969, opened 2026-04-02
feat(core): Add static key loading and startup validation for TrustedKeyService (no-changelog)

afitzek:
## Summary

- Implement `TrustedKeyService.initialize()` to parse, validate, and store trusted public keys from `N8N_TOKEN_EXCHANGE_TRUSTED_KEYS` config at module startup
- Add comprehensive key validation: algorithm family consistency checks, PEM parsing, key-type/algorithm compatibility, duplicate kid detection, and HMAC/none algorithm rejection
- JWKS key sources are recognised but not yet supported — they log a warning and are skipped
- Add `trustedKeys` config field to `TokenExchangeConfig` (supports `_FILE` variant for secret management)
- Wire `TrustedKeyService.initialize()` into `TokenExchangeModule.initialize()` so keys are loaded before the controller starts

https://linear.app/n8n/issue/IAM-460

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/token-exchange/services/trusted-key.service.ts
+++ b/packages/cli/src/modules/token-exchange/services/trusted-key.service.ts
@@ -62,11 +62,19 @@ export class TrustedKeyService {
 		let parsed: unknown;
 		try {
 			parsed = JSON.parse(raw);
-		} catch {
+		} catch (error) {
+			this.logger.error('Failed to parse trusted keys JSON', { error });
 			throw new UnexpectedError('Failed to parse trusted keys JSON');
 		}
 
-		const sources = z.array(TrustedKeySourceSchema).parse(parsed);
+		const sourcesResult = z.array(TrustedKeySourceSchema).safeParse(parsed);
+
+		if (!sourcesResult.success) {
+			this.logger.error('Trusted keys JSON has invalid format', { error: sourcesResult.error });
+			throw new UnexpectedError('Trusted keys JSON has invalid format');
+		}
+
+		const sources = sourcesResult.data;
 
 		for (const source of sources) {
 			if (source.type === 'jwks') {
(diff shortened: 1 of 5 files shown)


packages/cli/src/modules/token-exchange/services/trusted-key.service.ts
```diff
@@ -1,23 +1,155 @@
...
- * when the trusted-key persistence layer is introduced.
+ * Keys are loaded from the `N8N_TOKEN_EXCHANGE_TRUSTED_KEYS` config
+ * (or `_FILE` variant) and stored in-process. JWKS sources are
+ * recognised but not yet supported — they log a warning and are skipped.
  */
 @Service()
 export class TrustedKeyService {
+	private readonly logger: Logger;
+
+	private readonly keys = new Map<string, ResolvedTrustedKey>();
```
BGZStephen, line 40: In memory doesn't feel ideal, especially for multi-main setups?

packages/cli/src/modules/token-exchange/services/trusted-key.service.ts
```diff
@@ -1,23 +1,155 @@
...
+		};
+
+		if (!expectedTypes[family].includes(keyType ?? '')) {
+			throw new UnexpectedError(
+				`Trusted key "${kid}": key type "${keyType}" does not match algorithm family "${family}"`,
+			);
+		}
+
+		// 5. Reject duplicate kid
+		if (this.keys.has(kid)) {
```
BGZStephen: a bit nitpicky, but it feels like this should be the first check?

packages/cli/src/modules/token-exchange/services/trusted-key.service.ts
afitzek: This will change in a follow up, when we support jwls and use the cache service. 2a.3 or 4 I think

BGZStephen: Great, thanks!

packages/cli/src/modules/token-exchange/services/trusted-key.service.ts
```diff
@@ -1,23 +1,155 @@
...
+		}
+
+		let parsed: unknown;
+		try {
+			parsed = JSON.parse(raw);
+		} catch {
+			throw new UnexpectedError('Failed to parse trusted keys JSON');
+		}
+
+		const sources = z.array(TrustedKeySourceSchema).parse(parsed);
```
guillaumejacquart: Should we wrap this (same as the json parse) and throw a cleaner error if zod parse fails ?

guillaumejacquart (approved):
🚀

PR #27972, opened 2026-04-02
test(instance-ai): add memory benchmark harness with heap snapshot analysis (no-changelog)

OlegIvaniv:
## Summary

Adds a Playwright-based memory benchmark suite for instance-ai. The harness drives the system through parallel workflow builds, measures server-side heap after each round (via GC trigger + VictoriaMetrics polling), and runs memlab heap snapshot analysis to identify leaked objects.

### What's included

The benchmark driver opens multiple browser tabs, sends prompts to instance-ai, auto-approves HITL confirmations, waits for completion, then deletes threads and resets the database between rounds. This simulates realistic multi-user churn while isolating per-round memory growth.

The memory harness wraps this in a structured loop: capture baseline heap snapshot, run N phases (each a round of parallel builds), capture target and final snapshots, then run memlab's three-way analysis (leak detection, object growth, shape growth). Results are attached as Playwright test artifacts alongside container logs.

Test surfaces:
- `thread-churn.spec.ts` — parallel workflow builds with heap regression assertion
- `thread-churn-datatables.spec.ts` — builds involving data table operations
- `thread-churn-delegation.spec.ts` — builds with sub-agent delegation
- `single-thread-lifecycle.spec.ts` — create/message/delete lifecycle
- `cancel-abort.spec.ts` — mid-stream cancellation
- `sse-reconnection.spec.ts` — SSE disconnect/reconnect

The e2e controller gets two new endpoints: `POST /rest/e2e/gc` to trigger garbage collection and `GET /rest/e2e/heap-snapshot` to download a V8 heap snapshot for offline analysis.

## Review / Merge checklist
- [ ] Tests included. 


packages/testing/playwright/utils/benchmark/instance-ai-driver.ts
```diff
@@ -0,0 +1,304 @@
...
+		const threadIds = [...this.createdThreadIds];
+		for (const threadId of threadIds) {
+			await this.deleteThread(threadId);
+		}
+	}
+
+	/** Full cleanup: close tabs + delete threads + reset DB + re-authenticate. */
+	async cleanup(): Promise<void> {
+		await this.closeAllTabs();
+		await this.deleteAllThreads();
```
mutdmour: Question: is `resetDatabase()` necessary here? Instance-AI only loads artifacts associated with the active thread, so leftover workflows/credentials from round 1 should never get loaded into memory during round 2 (which operates on a fresh thread with no relationship to them).

More importantly, the DB reset could mask a real signal: if thread deletion is *supposed* to cascade-delete associated workflows/credentials but doesn't (orphaned rows), the reset silently cleans them up and you'd never notice the bug.

Consider replacing `resetDatabase()` with a more targeted check — e.g., after `deleteAllThreads()`, assert that orphaned artifacts *don't* exist in the DB. That turns a potential blind spot into an actual test. The re-auth could be avoided entirely since thread deletion shouldn't invalidate sessions.

If the reset is mainly for determinism (preventing the AI from behaving differently when it sees existing workflows), that's worth a comment explaining the intent.

mutdmour (commented):
Review comment on cleanup strategy

packages/testing/playwright/utils/benchmark/instance-ai-driver.ts
mutdmour: Follow-up on snapshot timing — the `target` heap snapshot in the churn tests may not be giving memlab what it needs.

**How memlab's three-snapshot model works:**
1. **Baseline** — snapshot of clean state
2. **Action** — do the thing (create threads, build workflows)
3. **Target snapshot** — taken immediately after, while data is still live in memory
4. **Cleanup** — undo the action (delete threads, free resources)
5. **Final snapshot** — taken after cleanup

Leak formula: `(target \ baseline) ∩ final` — objects that appeared in target and survived into final.

**What the churn tests currently do:**
1. Baseline snapshot ✅
2. Round 1: build → `cleanup()` (close tabs, delete threads, reset DB)
3. Round 2: build → `cleanup()`
4. Round 3: build → `cleanup()`
5. Target snapshot — but everything has already been cleaned up
6. Final GC + stabilize
7. Final snapshot
8. `analyzeHeapLeaks(baseline, target, final)`

Since target is taken after all cleanups, target ≈ final ≈ baseline (all "empty"). The formula `(target \ baseline)` is nearly empty, so memlab has very little to work with. The retainer chain analysis loses its power because it never saw the loaded state.

**Best practice — take target mid-action:**
1. Baseline snapshot
2. Rounds 1–2: build → cleanup (amplify small leaks)
3. Round 3: build → **target snapshot here, with threads/workflows still live**
4. Cleanup
5. Final snapshot
6. `analyzeHeapLeaks(baseline, target, final)`

This gives memlab the contrast it needs: target has both legitimate live objects and leaked objects, final has only leaked objects.

`single-thread-lifecycle.spec.ts` has the same issue — target is post-harness, not mid-action.

Ref: [How memlab Works](https://facebook.github.io/memlab/docs/how-memlab-works/)

packages/testing/playwright/tests/infrastructure/benchmarks-local/instance-ai/thread-churn.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+				thresholdMB: 5,
+				stableReadingsRequired: 3,
+			};
+
+			const result = await runMemoryBenchmark(
+				{
+					testInfo,
+					baseUrl: backendUrl,
+					metrics: services.observability.metrics,
+					dimensions: {
```
mutdmour, line 35: **Missing warmup phase:** The first round after server start pays a one-time cost — V8 JIT compilation of Instance-AI code paths, lazy module loading (LangChain, model clients), connection pool initialization, cache population. This allocates memory that isn't a leak but inflates the round-1 delta against baseline.

For example, if baseline is 50MB, round 1 might jump to 80MB (30MB of init overhead), but round 2 only goes to 82MB (2MB of real per-round cost). Without a warmup, that 30MB initialization spike pollutes the leak signal.

Suggestion: run one throwaway build before taking the baseline snapshot, so baseline already includes the JIT/module-loading cost and rounds only show incremental allocation:

```typescript
// Warmup: pay one-time init costs before baseline
await driver.runParallel([BENCHMARK_PROMPTS[0]]);
await driver.cleanup();
// NOW take baseline and start measured rounds
```

This applies to all 6 spec files.

packages/testing/playwright/tests/infrastructure/benchmarks-local/harness/memory-harness.ts
```diff
@@ -0,0 +1,185 @@
...
+		snapshotPaths.target = await takeHeapSnapshot(baseUrl, testInfo, 'target');
+	}
+
+	// --- Post-cleanup ---
+	let postCleanup = emptyHeap;
+	let leakMB = 0;
+
+	if (!dryRun) {
+		postCleanup = await getStableHeap(baseUrl, metrics, heapOptions);
+		leakMB = postCleanup.heapUsedMB - baseline.heapUsedMB;
```
mutdmour, line 218: **No assertions on `leakMB`:** The harness computes `leakMB` (post-cleanup minus baseline) and attaches it as a metric, but no spec file asserts on it. These benchmarks produce data but can never fail on a regression — they're observability without a gate.

Consider adding a configurable threshold to the harness config (e.g., `maxLeakMB: 20`) and having `runMemoryBenchmark` throw or return a `passed` boolean when the delta exceeds it. Even a generous threshold would catch catastrophic leaks automatically.

Same gap exists for RSS/non-heap overhead — `nonHeapOverheadMB` is computed and logged but never asserted on. Native memory leaks (unclosed streams, SSE buffers, socket handles) would only show up in RSS, not heap. A secondary threshold like `maxRssGrowthMB` would surface those.

mutdmour (commented):
A few additional observations on the benchmark harness

pnpm-lock.yaml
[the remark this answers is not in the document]
OlegIvaniv: `basic-ftp` is a transitive dep of `puppeteer` (via its proxy-agent stack). It's pulled in by `@langchain/community`, `@memlab/*`, and never used directly

packages/testing/playwright/utils/benchmark/instance-ai-driver.ts
OlegIvaniv: Fair point with db cleanup, I wanted to make sure there are no side-effects from different scenarios that would make the test run unpredictable, but even the workflow building scenario seems to run fine without full db cleanup. Also added `captureTargetAfterPhase` option to the harness config, specs now capture the target snapshot mid-action (after the one before last round, while threads/workflows are still live) before cleanup runs.

packages/testing/playwright/tests/infrastructure/benchmarks-local/harness/memory-harness.ts
OlegIvaniv: Added `maxLeakMB` and `maxRssGrowthMB` to the harness config. When exceeded, `result.passed` is `false` and specs assert on it

packages/testing/playwright/tests/infrastructure/benchmarks-local/instance-ai/thread-churn.spec.ts
OlegIvaniv: Added an optional `warmup` callback to `MemoryBenchmarkConfig`. All specs now run a lightweight warmup prompt ("List my current credentials") before baseline

mutdmour:
**Follow-up on cleanup strategy (extends [earlier comment on resetDatabase](https://github.com/n8n-io/n8n/pull/27972#discussion_r2086685498))**

I think `deleteAllThreads()` can go too, for the same reason as `resetDatabase()`.

What we're really measuring is: does the server give memory back after the user closes the tab? At that point SSE disconnects and the server should release all in-memory state for that thread. The thread row stays in the DB, but DB rows don't consume heap.

If thread deletion actually makes a difference to heap measurements, that's exactly the kind of leak this harness should be surfacing -- not silently cleaning up before we measure.

I'd just close tabs, GC, and measure. Simpler flow, stronger signal.

packages/testing/playwright/tests/infrastructure/benchmarks/harness/memory-harness.ts
```diff
@@ -0,0 +1,305 @@
...
+		const warmupElapsed = ((Date.now() - warmupStart) / 1000).toFixed(1);
+		console.log(`[MEMORY] Warmup complete [${warmupElapsed}s]`);
+	}
+
+	// --- Baseline ---
+	let baseline = emptyHeap;
+	let baselineRss: RssBreakdown | null = null;
+	if (!dryRun) {
+		baseline = await getStableHeap(baseUrl, metrics, heapOptions);
+		await attachMetric(testInfo, 'heap-baseline', baseline.heapUsedMB, 'MB', dimensions);
```
shortstacked: Can you change the 'heap-baseline' name here. This metric name is used for populating our nightly memory checks on idle instances.

shortstacked (approved):

PR #27984, opened 2026-04-02
feat(editor): Add Instance AI prompt suggestions

romeobalta:
## Summary

- Add empty-state prompt suggestions to the Instance AI composer.
- Add a `Quick examples` menu with predefined prompts.
- Preview prompt copy on hover and submit selected prompts directly.
- Hide suggestions once the composer is no longer empty, while a send is pending, or while an existing thread is still hydrating.
- Extract the suggestions UI into a dedicated component and extend focused frontend tests.

image

image

## Context

This PR includes a small amount of store and view work in addition to the new pills UI.

That extra work is there to support the new submission behavior safely:

- Prompt suggestions now submit directly, so rapid repeated clicks on a brand-new thread can otherwise create duplicate first messages before thread persistence finishes.
- Existing threads temporarily clear local messages before historical messages hydrate, which can otherwise make empty-state suggestions flash briefly in the wrong context.

## How to test

1. Open Instance AI on a new thread and verify the prompt suggestions render under the empty composer.
2. Hover a prompt pill and verify the input shows preview text without inserting anything.
3. Open `Quick examples`, hover an example, and verify the input preview updates.
4. Click a prompt pill and verify it submits immediately.
5. Click a quick example and verify it submits immediately.
6. Repeatedly click a prompt on a brand-new thread and verify only one first message / run is created.
7. Open an existing thread and verify empty-state suggestions do not appear while history is still loading.

## Review / Merge checklist

- [x] Tests included.


packages/frontend/@n8n/i18n/src/locales/en.json
Cadiac, line 1: Are these sample prompts something that we've seen the agent succeed with?

(we can iterate on these later with product)

romeobalta: Yes, @paulomation iterated on them by verifying they produce good results.

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiPromptSuggestions.vue
```diff
@@ -0,0 +1,442 @@
...
+
+.suggestionRow {
+	display: flex;
+	flex-wrap: wrap;
+	justify-content: center;
+	gap: var(--spacing--2xs);
+	width: 100%;
+}
+
+.suggestionButton {
```
Cadiac, line 264: I suppose we had nothing suitable for these kinds of buttons on our design system?

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiPromptSuggestions.vue
```diff
@@ -0,0 +1,442 @@
...
+				<span>{{ i18n.baseText(suggestion.labelKey) }}</span>
+				<N8nIcon
+					v-if="isMenuSuggestion(suggestion)"
+					:icon="isQuickExamplesOpen ? 'chevron-up' : 'chevron-down'"
+					:size="12"
+					:class="$style.suggestionChevron"
+				/>
+			</button>
+		</div>
+		<Transition name="quick-examples-fade">
```
Cadiac, line 197: Also I guess there was no suitable existing menu/dropdown component to use here, and making a new one for this was a good idea?

This probably works, I'm just raising as in the past we've been rather careful at not inventing new UI components outside our design system

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiPromptSuggestions.vue
romeobalta: The only things that could fit would be button, but they would be buttons instead of "pills" if that makes sense.

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAi.store.ts
```diff
@@ -632,11 +638,23 @@ export const useInstanceAiStore = defineStore('instanceAi', () => {
 		});
 	}
 
-	async function loadHistoricalMessages(threadId: string): Promise<void> {
+	async function loadHistoricalMessages(
+		threadId: string,
+		hydrationRequestToken?: number,
```
Cadiac, line 643: I'm honestly having a bit difficult time understanding this new `hydrationRequestToken` mechanism that was added - this was probably added to fix some bug, which one?

When did we run into these issues that this code is fixing, they seem a bit unrelated to adding suggested prompts?

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiPromptSuggestions.vue
romeobalta: We do have a dropdown component but it works quite differently than the look and feel we were going for here.

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAi.store.ts
romeobalta: Good question. When changing between sessions, the screen would flicker for a bit to the empty session layout (the one with the input in the middle) before loading the new session. This fix ensures that it keeps the layout (the input doesn't jump from the bottom of the screen to middle of the screen and then back down).

With the pills added it looked even worse.

Cadiac: fair enough

Cadiac (approved):
🚀

PR #27985, opened 2026-04-02
fix(core): Improve audit queries to avoid PostgreSQL bind parameter limits

MarcL:
## Summary

The audit endpoint's credentials risk reporter fetched all execution IDs then passed them into a single `WHERE IN (...)` clause. On instances with 250k+ executions, this exceeded PostgreSQL's 65,535 bind parameter limit (error code `08P01`), causing a 500 error.

Replaced the two-step fetch-then-IN pattern with a single `INNER JOIN` query that uses one bind parameter regardless of execution count. Also batched `deleteMany` as defense-in-depth and removed the now-unused `findByExecutionIds` method.

Existing integrtation tests in `redentials-risk-reporter.test.ts`.

https://linear.app/n8n/issue/LIGO-282

## Review / Merge checklist

- [ ] ~[Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.~
- [ ] ~Tests included.~
- [ ] ~PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if the PR is an urgent fix that needs to be backported)~


packages/@n8n/db/src/repositories/execution-data.repository.ts
```diff
@@ -18,18 +22,23 @@ export class ExecutionDataRepository extends Repository<ExecutionData> {
...
+				date: DateUtils.mixedDateToUtcDatetimeString(date),
+			})
+			.getMany()
+			.then((results) => results.map(({ workflowData }) => workflowData));
 	}
 
 	async deleteMany(executionIds: string[]) {
 		if (executionIds.length === 0) return;
 
-		await this.delete({ executionId: In(executionIds) });
```
geemanjs, line 33: This should probably be wrapped in a transaction? so it rollsback if one of the batches fail?

packages/@n8n/db/src/repositories/execution-data.repository.ts
```diff
@@ -18,18 +22,23 @@ export class ExecutionDataRepository extends Repository<ExecutionData> {
...
-		}).then((executionData) => executionData.map(({ workflowData }) => workflowData));
+	async getWorkflowsExecutedSince(date: Date) {
+		return await this.createQueryBuilder('executionData')
+			.select('executionData.workflowData')
+			.innerJoin('executionData.execution', 'execution')
+			.where('execution.startedAt >= :date', {
+				date: DateUtils.mixedDateToUtcDatetimeString(date),
+			})
+			.getMany()
+			.then((results) => results.map(({ workflowData }) => workflowData));
```
geemanjs, line 33: I wonder if there is a way to fetch here without having to iterate around all the results 🤔

packages/@n8n/db/src/repositories/execution-data.repository.ts
```diff
@@ -18,18 +22,23 @@ export class ExecutionDataRepository extends Repository<ExecutionData> {
...
-			where: {
-				executionId: In(executionIds),
-			},
-		}).then((executionData) => executionData.map(({ workflowData }) => workflowData));
+	async getWorkflowsExecutedSince(date: Date) {
+		return await this.createQueryBuilder('executionData')
+			.select('executionData.workflowData')
+			.innerJoin('executionData.execution', 'execution')
+			.where('execution.startedAt >= :date', {
+				date: DateUtils.mixedDateToUtcDatetimeString(date),
```
geemanjs, line 30: Do we have an index on this field?

packages/@n8n/db/src/repositories/execution-data.repository.ts
MarcL: Good shout. I'll do this.

packages/@n8n/db/src/repositories/execution-data.repository.ts
MarcL: Is this a big overhead to iterate through all of the results? It's not doing another DB call.

packages/@n8n/db/src/repositories/execution-data.repository.ts
MarcL: There is a composite index of `(workflowId, startedAt)` - https://github.com/n8n-io/n8n/blob/7399af34c975253917d06ff9f64b61b2fedf1101/packages/%40n8n/db/src/migrations/common/1723796243146-RefactorExecutionIndices.ts#L42

The code I replaced (`getIdsSince`) was using a similar query so we could add a separate `startedAt` if needed but I'd assumed this wasn't an existing performance issue. What do you think?

geemanjs (approved):

PR #27992, opened 2026-04-02
feat(core): Add input/output token tracking to node graph telemetry

DawidMyslak:
## Summary

Add `ai_input_tokens` and `ai_output_tokens` fields to `node_graph_string` telemetry, alongside the existing `ai_model` field. This enables tracking AI model token consumption per node in Metabase.

**[Loom recording](https://www.loom.com/share/df6201e965c64d25bbd14b53bf8bc92d)**

**Two categories of AI nodes are covered:**

1. **LM sub-nodes** (e.g. `lmChatOpenAi`, connected to Agents/Chains) — token data is already standardized via `N8nLlmTracing`, extracted directly from `runData` and summed across agent multi-call scenarios.

2. **Standalone vendor nodes** (defined in `AI_VENDOR_NODE_TYPES`: OpenAI, Anthropic, Ollama, Google Gemini) — token data comes from raw API responses but is normally stripped by the "Simplify output" default. We capture it via `setMetadata()` before simplification, preserving it for telemetry without affecting user-visible output.

A shared `accumulateTokenUsage()` utility in `n8n-workflow` handles accumulation across multi-item executions and keeps vendor node changes to a one-liner.

**How to verify:** Run a workflow with any AI node, check that `node_graph_string` in the telemetry event includes `ai_input_tokens` and `ai_output_tokens` for the AI node entries.

https://linear.app/n8n/issue/NODE-4781

## Review / Merge checklist

- [x] Tests included.


RomanDavydchuk (approved):
LGTM

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/actions/text/message.operation.ts
```diff
@@ -510,6 +511,21 @@ export async function execute(this: IExecuteFunctions, i: number): Promise<INode
 		body,
 	})) as GenerateContentResponse;
 
+	const captureUsage = () => {
+		const usageMetadata = (response as unknown as Record<string, unknown>).usageMetadata as
+			| { promptTokenCount: number; candidatesTokenCount: number }
+			| undefined;
```
alexander-gekov, line 517: Gemini also has `thoughtsTokenCount` I think

DawidMyslak: Good point!

We're planning a follow-up PR to extend token tracking to the other operations (analyze, classify, etc.). We'll consider `thoughtsTokenCount` for Gemini there as well (maybe same for OpenAI's `reasoning_tokens` if reasoning models are used).

Worth noting that thinking tokens are billed by the API but aren't part of the user-visible output, so we'll need to decide whether they go into `ai_output_tokens` (for cost accuracy) or get tracked as a separate field. For now shipping with message operations only.

PR #28008, opened 2026-04-03
refactor(core): Extract API key auth into AuthStrategy pattern (no-changelog)

guillaumejacquart:
## Summary

Extract the existing API key authentication logic from `PublicApiKeyService.getAuthMiddleware()` into a dedicated `ApiKeyAuthStrategy` class registered via an `AuthStrategyRegistry`. This introduces a strategy pattern so additional auth strategies (e.g. scoped JWT for token exchange) can be registered without modifying the core auth handler.

- New `ApiKeyAuthStrategy` class with the same auth logic (header check, key validation, user loading, JWT expiry, legacy key support)
- New `AuthStrategyRegistry` that evaluates strategies sequentially — first non-null result wins
- Telemetry (`public-api-invoked`) moved to the OpenAPI validator handler wrapper, where `version` is directly available (no `req.baseUrl` parsing hack)
- Constants (`API_KEY_AUDIENCE`, `API_KEY_ISSUER`, `PREFIX_LEGACY_API_KEY`) exported from `PublicApiKeyService` for reuse
- Integration tests for the new strategy; all 492 existing public API tests pass unchanged

https://linear.app/n8n/issue/IAM-463

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/public-api/index.ts
```diff
@@ -104,7 +107,21 @@ function createLazyValidatorMiddleware(
...
 						validateSecurity: {
 							handlers: {
-								ApiKeyAuth: Container.get(PublicApiKeyService).getAuthMiddleware(version),
+								ApiKeyAuth: async (req: AuthenticatedRequest) => {
+									const authenticated = await Container.get(AuthStrategyRegistry).authenticate(req);
+
+									if (authenticated) {
+										Container.get(EventService).emit('public-api-invoked', {
+											userId: req.user.id,
+											path: req.path,
```
phyllis-noester, line 126: might be interesting to log what auth strategy as well

packages/cli/src/services/api-key-auth.strategy.ts
```diff
@@ -0,0 +1,58 @@
...
+				});
+			} catch (e) {
+				if (e instanceof TokenExpiredError) return false;
+				throw e;
+			}
+		}
+
+		req.user = user;
+
+		void this.lastActiveAtService.updateLastActiveIfStale(user.id);
```
phyllis-noester: why do we do that in the api key strategy? aren't we interested in last active regardless of auth strategy?

packages/cli/src/services/api-key-auth.strategy.ts
```diff
@@ -0,0 +1,58 @@
...
+	constructor(
+		private readonly userRepository: UserRepository,
+		private readonly jwtService: JwtService,
+		private readonly lastActiveAtService: LastActiveAtService,
+	) {}
+
+	async authenticate(req: AuthenticatedRequest): Promise<boolean | null> {
+		const providedApiKey = req.headers[API_KEY_HEADER] as string | undefined;
+
+		if (!providedApiKey) return null;
```
phyllis-noester: we could also check whether it's not string or doesn;t exist instead of casting

phyllis-noester (dismissed):

phyllis-noester (approved):

PR #28009, opened 2026-04-03
feat(core): Add identity resolution for token exchange (no-changelog)

afitzek:
## Summary

Implements `IdentityResolutionService` that resolves verified external token claims (`sub`) to an n8n user. This replaces the skeleton stub that previously threw "not implemented".

**Part of the OAuth 2.0 Token Exchange initiative** (Phase 2a — Embed flow). External systems identify users by their own IDs (`sub` claim). This service maps those to internal n8n users — linking on first encounter and keeping profiles in sync on subsequent exchanges.

### Resolution algorithm

1. **Known sub** — `AuthIdentity` lookup by `sub` + `token-exchange` provider → return linked user
2. **Email fallback** — no identity match but email matches existing user → create `AuthIdentity` link, return user
3. **JIT provision** — no match at all → create user + personal project + `AuthIdentity` in a single transaction

### Role handling

- Role claim validated against the key's `allowedRoles` whitelist
- `global:owner` can never be assigned via token exchange (existing owners keep their role, new users are rejected)
- Existing users keep their current role when the claimed role is invalid or not allowed — login is never blocked due to role mismatch
- New users get `global:member` by default when no role claim is present

### Other changes

- `TokenExchangeService.verifyToken` now returns `{ claims, resolvedKey }` so `embedLogin` can pass `allowedRoles` to identity resolution
- Three new audit events: `token-exchange-identity-linked`, `token-exchange-user-provisioned`, `token-exchange-role-updated`

### How to test manually

1. Configure a trusted key with `allowedRoles: ['global:member', 'global:admin']`
2. Generate a JWT with `sub`, `email`, and optional `role` claim signed by the trusted key
3. Call `POST /rest/auth/embed` with `{ "token": "<jwt>" }`
4. Verify:
   - First call with a new `sub` + known email links the identity and returns the existing user
   - First call with a new `sub` + unknown email creates a new user with personal project
   - Subsequent calls with the same `sub` resolve via identity lookup (Path 1)
   - Role updates are applied when the claim is in `allowedRoles`
   - `global:owner` role claim is silently ignored for existing users and rejected for new users
5. Check audit events are emitted for linking, provisioning, and role changes

## Related tickets

- https://linear.app/n8n/issue/IAM-462

## Review checklist

- [X] Tests included.

### DIFF
--- a/packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
+++ b/packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
@@ -1,22 +1,236 @@
-import type { User } from '@n8n/db';
+import { Logger } from '@n8n/backend-common';
+import {
+	AuthIdentity,
+	AuthIdentityRepository,
+	GLOBAL_MEMBER_ROLE,
+	GLOBAL_ROLES,
+	UserRepository,
+	type User,
+} from '@n8n/db';
 import { Service } from '@n8n/di';
 
-import { UserError } from 'n8n-workflow';
+import { AuthError } from '@/errors/response-errors/auth.error';
 
 import type { ExternalTokenClaims } from '../token-exchange.schemas';
 
 /**
- * Resolves validated external token claims to an n8n User.
- *
- * This is a skeleton service — the real implementation with user
- * lookup/creation logic will be provided in a follow-up ticket.
+ * Password placeholder for JIT-provisioned users. This is not a valid bcrypt
+ * hash, so it can never match any input — the user can only authenticate
+ * through token exchange.
  */
+const INVALID_PASSWORD_PLACEHOLDER = '!token-exchange-no-password';
+
+type GlobalRoleKey = keyof typeof GLOBAL_ROLES;
+
+function isGlobalRole(role: string): role is GlobalRoleKey {
+	return role in GLOBAL_ROLES;
+}
+
 @Service()
 export class IdentityResolutionService {
+	private readonly logger: Logger;
+
+	constructor(
+		logger: Logger,
+		private readonly userRepository: UserRepository,
+		private readonly authIdentityRepository: AuthIdentityRepository,
+	) {
+		this.logger = logger.scoped('token-exchange');
+	}
+
 	/**
 	 * Map external identity claims to a local n8n user, creating one if necessary.
+	 *
+	 * Resolution order:
+	 * 1. AuthIdentity lookup by sub + token-exchange provider
+	 * 2. Email fallback — link existing user to this sub
+	 * 3. JIT provision — create user + personal project + identity in a transaction
+	 *
+	 * Role handling: the role claim is only applied when it is both valid and
+	 * permitted by the key's allowedRoles. For existing users the current role
+	 * is preserved when the claimed role is disallowed or absent — login is
+	 * never blocked because of a role mismatch.
 	 */
-	async resolve(_claims: ExternalTokenClaims): Promise<User> {
-		throw new UserError('Identity resolution is not yet implemented');
+	async resolve(claims: ExternalTokenClaims, allowedRoles?: string[]): Promise<User> {
+		// Path 1: known sub
+		const identity = await this.authIdentityRepository.findOne({
+			where: { providerId: claims.sub, providerType: 'token-exchange' },
+			relations: { user: { role: true } },
+		});
+
+		if (identity) {
+			this.logger.debug('Resolved user by auth identity', { sub: claims.sub });
+			const resolvedRole = this.resolveRoleForExistingUser(
+				claims.role,
+				allowedRoles,
+				identity.user.role?.slug,
+			);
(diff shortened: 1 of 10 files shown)


packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
```diff
@@ -1,22 +1,306 @@
...
 /**
- * Resolves validated external token claims to an n8n User.
- *
- * This is a skeleton service — the real implementation with user
- * lookup/creation logic will be provided in a follow-up ticket.
+ * Password placeholder for JIT-provisioned users. This is not a valid bcrypt
+ * hash, so it can never match any input — the user can only authenticate
+ * through token exchange.
  */
+const INVALID_PASSWORD_PLACEHOLDER = '!token-exchange-no-password';
```
BGZStephen, line 23: This feels odd, we currently support oauth users with no passwords, the owner user is bootstrapped with no password, meaning it's not a DB constraint, right? It feels like we should just leave the password null for JIT provisioned users

packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
```diff
@@ -1,22 +1,306 @@
...
+	 * 3. JIT provision — create user + personal project + identity in a transaction
+	 *
+	 * Role handling: the role claim is only applied when it is both valid and
+	 * permitted by the key's allowedRoles. For existing users the current role
+	 * is preserved when the claimed role is disallowed or absent — login is
+	 * never blocked because of a role mismatch.
 	 */
-	async resolve(_claims: ExternalTokenClaims): Promise<User> {
-		throw new UserError('Identity resolution is not yet implemented');
+	async resolve(
```
BGZStephen, line 64: Each of the three branches would benefit from being split out to reduce lines of code within this function and to make unit testing easier

packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
```diff
@@ -1,22 +1,306 @@
...
+	/**
+	 * Resolve the role claim for an existing user.
+	 *
+	 * Returns the role slug to sync to, or `undefined` to keep the current
+	 * role unchanged. Existing users are never blocked from logging in — if
+	 * the claimed role is invalid or not allowed, we simply skip the role
+	 * update and let them keep their current role.
+	 */
+	private resolveRoleForExistingUser(
+		roleClaim: ExternalTokenClaims['role'],
```
BGZStephen, line 203: It feels like we need to nail down whether this is a single role, or multiple roles? If it's an array, we should rename this to roles, not role, and just accept an array

packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
afitzek: I think if the password column is null, the use is deemed not active if I remember correctly. We also set an invalid password when we provision users in OIDC: https://github.com/n8n-io/n8n/blob/9b94862dc782c087f4e63e59fb81614d440e7afe/packages/cli/src/modules/sso-oidc/oidc.service.ee.ts#L332 

and in SAML:

https://github.com/n8n-io/n8n/blob/9b94862dc782c087f4e63e59fb81614d440e7afe/packages/cli/src/modules/sso-saml/saml-helpers.ts#L64

packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
```diff
@@ -1,22 +1,306 @@
...
+			return undefined;
+		}
+
+		if (!isGlobalRole(role)) {
+			this.logger.warn('Unknown role claim ignored', { role });
+			return undefined;
+		}
+
+		// If the claimed role is not in the allowed list, skip the update
+		if (allowedRoles && allowedRoles.length > 0 && !allowedRoles.includes(role)) {
```
BGZStephen, line 228: I think it's more natural to throw an error here and not fall back to the current role? Oauth flows tend to be very strict and throw errors often, letting this fall back could lead people to think their flow is OK but then not work consistently accross users?

packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
BGZStephen: oof, that feels wrong too, but that's another problem

packages/cli/src/modules/token-exchange/services/identity-resolution.service.ts
afitzek: This should be a single role, never multiple roles, this is my bad, it was wrong in the design document. I'll change it to be a single role.

BGZStephen: If it's easier we can do this in a one off follow up PR?

BGZStephen (approved):

PR #28011, opened 2026-04-03
fix(editor): Fix issue with credential based grouping in setup cards

scdekov:
## Summary
The credential type based grouping wasn't working well in the setup panel/builder setup in the following cases:
* 2 nodes sharing a credential and both having RLC param. If the user jumps straight to the second node card without filling the credential in the first they wouldn't be able to use the RLC (because credential is set in the first node)
* 2 nodes sharing a credential and only the second having a parameter to be filled. In that case with the old logic we'd show only credential picker for the second node (no card for the first) which means that if there is a node in between those two the execution would break if no credential is picked yet for the two nodes that are sharing

With the new grouping:
* First node(execution order) using a credential type always shows credential picker
* Following nodes having the same credential type
  * hidden if no other params than the credential
  * shows both independent credential picker and param inputs if there other params

## Test plan
Prompt to test - "Add two agents, one generates a joke, the second rates it. Both agents use openrouter with placeholders for the model values. Send the joke to telegram before it is being rated" 


MiloradFilipovic (commented):
Code looks good and the reported edge-cases work fine now 🙌
I was trying to test if HTTP nodes with same URLs work fine and run into two things that I would like to check:

I was using this prompt to create workflow with 2 HTTP nodes to notion API:

```
Build a workflow with two HTTP Request nodes for the same Notion API endpoint: https://api.notion.com/v1/pages

Requirements:

Workflow is triggered manually
Same URL on both HTTP Request nodes
Header auth type on both
Send notion DB id as X-Notion-DB header and leave editable placeholder for the value
```

Both nodes ended up looking ok but I was expecting to have credentials picker in setup cards and they were not there:

Image

Also, if I don't specify that workflow is triggered manually, builder creates just HTTP nodes and then it gets stuck in `Checking credentials...` step (possibly not related to this PR.

scdekov:
> Code looks good and the reported edge-cases work fine now 🙌 I was trying to test if HTTP nodes with same URLs work fine and run into two things that I would like to check:
> 
> I was using this prompt to create workflow with 2 HTTP nodes to notion API:
> 
> ```
> Build a workflow with two HTTP Request nodes for the same Notion API endpoint: https://api.notion.com/v1/pages
> 
> Requirements:
> 
> Workflow is triggered manually
> Same URL on both HTTP Request nodes
> Header auth type on both
> Send notion DB id as X-Notion-DB header and leave editable placeholder for the value
> ```
> 
> Both nodes ended up looking ok but I was expecting to have credentials picker in setup cards and they were not there:
> 
> Image
> Also, if I don't specify that workflow is triggered manually, builder creates just HTTP nodes and then it gets stuck in `Checking credentials...` step (possibly not related to this PR.

Interesting. I tried that prompt 3 times and got the right setup every time
image

@MiloradFilipovic is it possible that you had any followups or manual updates on the wf to get in that state? Or if you're able to reproduce it still could you drop the json of that workflow? May help me debug it

scdekov:
> > Code looks good and the reported edge-cases work fine now 🙌 I was trying to test if HTTP nodes with same URLs work fine and run into two things that I would like to check:
> > I was using this prompt to create workflow with 2 HTTP nodes to notion API:
> > ```
> > Build a workflow with two HTTP Request nodes for the same Notion API endpoint: https://api.notion.com/v1/pages
> > 
> > Requirements:
> > 
> > Workflow is triggered manually
> > Same URL on both HTTP Request nodes
> > Header auth type on both
> > Send notion DB id as X-Notion-DB header and leave editable placeholder for the value
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Both nodes ended up looking ok but I was expecting to have credentials picker in setup cards and they were not there:
> > Image
> > Also, if I don't specify that workflow is triggered manually, builder creates just HTTP nodes and then it gets stuck in `Checking credentials...` step (possibly not related to this PR.
> 
> Interesting. I tried that prompt 3 times and got the right setup every time image
> 
> @MiloradFilipovic is it possible that you had any followups or manual updates on the wf to get in that state? Or if you're able to reproduce it still could you drop the json of that workflow? May help me debug it

Ah I think I didn't paste the whole prompt that you shared. Let me try it again

scdekov:
> Build a workflow with two HTTP Request nodes for the same Notion API endpoint: https://api.notion.com/v1/pages
> 
> Requirements:
> 
> Workflow is triggered manually
> Same URL on both HTTP Request nodes
> Header auth type on both
> Send notion DB id as X-Notion-DB header and leave editable placeholder for the value

Still got the correct setup..
image

PR #28013, opened 2026-04-03
feat: support context caching in Google Vertex model

vViktorPL:
## Summary

Add optional [Vertex AI context caching](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/context-cache/context-cache-use) for the Google Vertex Chat Model: users can disable caching, use automatic cache creation with Redis-backed metadata, or reference a predefined `cachedContents` resource. Align request shaping with [Vertex expectations](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/context-cache/context-cache-use#context_cache_use_restrictions) (strip system messages and tool config when a cache applies), surface execution hints and UI notices where behavior diverges from the agent workflow, and avoid Redis races by using a short-lived connection per metadata operation.

Auto-caching strategy setup
Predefined cache strategy setup

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/llms/LmChatGoogleVertex/LmChatGoogleVertex.node.ts
+++ b/packages/@n8n/nodes-langchain/nodes/llms/LmChatGoogleVertex/LmChatGoogleVertex.node.ts
@@ -1,9 +1,15 @@
 import { ProjectsClient } from '@google-cloud/resource-manager';
 import type { GoogleAISafetySetting } from '@langchain/google-common';
 import { ChatVertexAI, type ChatVertexAIInput } from '@langchain/google-vertexai';
+import {
+	makeN8nLlmFailedAttemptHandler,
+	N8nLlmTracing,
+	getConnectionHintNoticeField,
+} from '@n8n/ai-utilities';
 import { formatPrivateKey } from 'n8n-nodes-base/dist/utils/utilities';
 import {
 	NodeConnectionTypes,
+	type INode,
 	type INodeType,
 	type INodeTypeDescription,
 	type ISupplyDataFunctions,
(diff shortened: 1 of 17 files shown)


packages/@n8n/nodes-langchain/nodes/llms/LmChatGoogleVertex/LmChatGoogleVertex.node.ts
lucamattiazzi, line 39: I think version should be bumped to 1.1

packages/@n8n/nodes-langchain/nodes/llms/LmChatGoogleVertex/context-caching/auto/auto-cache.ts
```diff
@@ -0,0 +1,553 @@
...
+			retryNotBeforeIso,
+		},
+		{ ttlSeconds: CACHE_TOO_SHORT_ERROR_TTL },
+	);
+}
+
+/**
+ * Fire-and-forget: completes `cachedContents.create` and updates Redis metadata (or deletes on hard failure).
+ */
+function scheduleContextCacheCreate(
```
lucamattiazzi, line 372: since this function isn't awaited, and google cache creation might not be fast enough, both the `ctx` and `logger` references might be used after the workflow has been completed, which might cause either UI degradation (errors not shown) or even errors. would it be possible to catch those errors, and highlight this possibility in the comments?

packages/@n8n/nodes-langchain/nodes/llms/LmChatGoogleVertex/context-caching/auto/auto-cache.ts
```diff
@@ -0,0 +1,553 @@
...
+				type: 'cold_no_create',
+				executionHints: [vertexContextCacheTooShortFromStoreHint()],
+			};
+		}
+	} else if (isExistingContextCache(meta)) {
+		if (parseExpireMs(meta.expireTime) <= nowMs) {
+			await deps.metadataStorage.delete(hash);
+			meta = null;
+			deps.logger.debug('VertexContextCache: cache_miss (expired)', metaBase);
+		} else if (meta.location !== deps.location) {
```
lucamattiazzi, line 234: is this actually something that can happen? isn't the location (and the model name a few lines later) used in computing the hash, and thus need to have the same value for the cache value to even be found?

lucamattiazzi (changes requested):

PR #28014, opened 2026-04-03
fix(If Node): Patches IF node when fields are missing

lucamattiazzi:
## Summary

This PR fixes a bug in AI generated workflows where the AI needs to generate an `n8n-nodes-base.if` node.
Due to missing informations to the AI, the node is generated with a broken structure, but will still run, e.g.:
```
{
  "parameters": {
    "conditions": {
      "conditions": [
        {
          "id": "rain-check",
          "leftValue": "={{ $json.hasRain }}",
          "operator": {
            "type": "boolean",
            "operation": "true"
          },
          "rightValue": ""
        }
      ]
    },
    "options": {}
  },
  "id": "aa3f0c98-2f23-478d-981d-4dbb7e6693e4",
  "name": "Is Rain Predicted?",
  "type": "n8n-nodes-base.if",
  "typeVersion": 2.3,
  "position": [
    912,
    208
  ]
}
```

When firing, though, it will always fire `true`, since the function `extractValueFilter` tries to extract the value of the `conditions` filter. since this filter is missing the field `combinator`, the function `isFilterValue` will return `false`, and the extracted value of this filter will be its own raw value, that will then evaluate as `true`ish

The way the fix is implemented is twofold: first, it applies the same patching that the UI does when the user opens such a misformed node, applying the default values where missing. Also, adds hints for the AI to create correctly formed nodes.

[related linear ticket](https://linear.app/n8n/issue/AI-2334/bug-ai-generated-n8n-nodes-baseif-nodes-always-resolve-to-true)

## Review / Merge checklist

- [ ] Tests included. 


burivuhster (dismissed):
Looks good!

packages/workflow/src/node-helpers.ts
```diff
@@ -775,6 +775,25 @@ export function getNodeParameters(
...
+					nodeProperties.type === 'filter' &&
+					nodeValues[nodeProperties.name] !== undefined &&
+					typeof nodeValues[nodeProperties.name] === 'object' &&
+					nodeValues[nodeProperties.name] !== null
+				) {
+					const defaultOptions = {
+						caseSensitive: true,
+						leftValue: '',
+						typeValidation: 'loose',
+						version: 1,
```
michael-radency, line 791: should type validation be `strict` there? I believe this is default

michael-radency (commented):
Overall looks good, just want to confirm on `typeValidation` mode

packages/workflow/src/node-helpers.ts
lucamattiazzi: that's right! just pushed the correction, thank you!

michael-radency (dismissed):
LGTM

burivuhster (approved):
🚀

PR #28015, opened 2026-04-03
feat: Add agent schema introspection (no-changelog)

yehorkardash:
## Summary

PR changes:
- added agent config introspection that allows dehydrating agent into a single json config and then reconstructing it from the config
- added credential provider that allows using credential names instead of raw secret keys

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/agents/src/types/sdk/schema.ts
```diff
@@ -0,0 +1,112 @@
...
+		thinking: ThinkingSchema | null;
+		toolCallConcurrency: number | null;
+		requireToolApproval: boolean;
+	};
+}
+
+export interface ToolSchema {
+	name: string;
+	description: string;
+	type: 'custom' | 'workflow' | 'provider' | 'mcp';
```
burivuhster, line 37: We may want to extract MCPs into a separate Agent schema prop in the future, as it might be more than a simple tool or tool set. But it's OK for now to keep things simple

packages/@n8n/agents/src/sdk/tool.ts
```diff
@@ -61,6 +64,10 @@ export function wrapToolForApproval(tool: BuiltTool, config: ApprovalConfig): Bu
 			} as ToolContext);
 		},
 	};
+	// Tag with approval symbol — Object.assign avoids TS index-signature issues
+	// on the BuiltTool interface while still being detectable via `in` operator.
+	Object.assign(wrapped, { [APPROVAL_WRAPPED]: true });
+	return wrapped;
```
burivuhster: Can it be simplified to a boolean flag in the `BuiltTool` itself?

packages/@n8n/agents/src/sdk/provider-capabilities.ts
```diff
@@ -0,0 +1,35 @@
...
+	label: string;
+	default: number | string;
+	options?: string[];
+}
+
+export const providerCapabilities: Record<
+	string,
+	{
+		thinking?: ProviderCapability;
+	}
```
burivuhster, line 13: Should it be `Record<string, ProviderCapability>`?

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+	for (const ts of schema.tools) {
+		if (!ts.editable) continue;
+		const schemas: { suspend?: ZodType; resume?: ZodType } = {};
+		if (ts.suspendSchemaSource) {
+			schemas.suspend = await handlerExecutor.evaluateSchema(ts.suspendSchemaSource);
+		}
+		if (ts.resumeSchemaSource) {
+			schemas.resume = await handlerExecutor.evaluateSchema(ts.resumeSchemaSource);
+		}
+		toolSchemas.set(ts.name, schemas);
```
burivuhster: Silent override is possible here if there are name duplicates. Maybe we can add a runtime check here and fail explicitly?

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	for (const ts of schema.tools) {
```
burivuhster: These two passes over `schema.tools` is a bit hard to follow for me. Is it possible to re-organize into one loop + code branches for all possible tool types?

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+
+	// --- Provider tools ---
+	for (const pt of schema.providerTools) {
+		if (pt.source) {
+			const evaluated = (await handlerExecutor.evaluateExpression(pt.source)) as {
+				name: string;
+				args?: Record<string, unknown>;
+			};
+			agent.providerTool({
+				name: evaluated.name as `${string}.${string}`,
```
burivuhster: This type assertion can be moved into the one above:
name: `${string}.${string}`

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	// --- Memory ---
+	if (schema.memory !== null) {
+		const memory = new Memory();
+		if (schema.memory.lastMessages !== null) {
+			memory.lastMessages(schema.memory.lastMessages);
+		}
+		agent.memory(memory);
+	}
```
burivuhster, line 187: Specific memory settings seem to be lost here – we use only `lastMessages` property from entire `MemorySchema` object

packages/@n8n/agents/src/types/sdk/schema.ts
```diff
@@ -0,0 +1,112 @@
...
+	guardrails: GuardrailSchema[];
+
+	mcp: McpServerSchema[] | null;
+	telemetry: TelemetrySchema | null;
+	checkpoint: 'memory' | null;
+
+	config: {
+		structuredOutput: {
+			enabled: boolean;
+			schemaSource: string | null; // original Zod source string
```
burivuhster, line 26: Do we need a dedicated `enabled` property? Can we simply check if the schemaSource is null or not, same as for other props like thinking?

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+ * fully-configured Agent instance with proxy handlers that delegate tool
+ * execution to the provided `HandlerExecutor`.
+ *
+ * All source expressions in the schema (provider tools, MCP configs,
+ * telemetry, structured output, suspend/resume schemas) are evaluated
+ * via `HandlerExecutor.evaluateExpression()` / `evaluateSchema()`.
+ *
+ * The `agent` parameter is the Agent instance to configure (avoids circular import).
+ */
+export async function fromSchema(
```
burivuhster, line 61: Would it make sense to refactor every section into a separate function? Like `fromModelSchema(agent, schema.model)`, `fromCredentialSchema(agent, schema.credential)`, etc.
They are small, but there are quite a few of them, so this function is a bit too long

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+		const outputSchema = await handlerExecutor.evaluateSchema(
+			schema.config.structuredOutput.schemaSource,
+		);
+		agent.structuredOutput(outputSchema);
+	}
+
+	// --- Credential provider ---
+	if (options.credentialProvider) {
+		agent.credentialProvider(options.credentialProvider);
+	}
```
burivuhster: This can be moved closer to handling `schema.credential`

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+ * Reconstruct a live Agent from an AgentSchema JSON.
+ *
+ * This is the inverse of `Agent.describe()` — it takes a serialised schema
+ * (produced by `describe()` or stored in the database) and rebuilds a
+ * fully-configured Agent instance with proxy handlers that delegate tool
+ * execution to the provided `HandlerExecutor`.
+ *
+ * All source expressions in the schema (provider tools, MCP configs,
+ * telemetry, structured output, suspend/resume schemas) are evaluated
+ * via `HandlerExecutor.evaluateExpression()` / `evaluateSchema()`.
```
burivuhster, line 57: Not sure about this, but: would it be more clear to evaluate all expressions first as a separate step? Then `fromSchema` will not have to worry about calling `await handlerExecutor.evaluateSchema` every time

packages/@n8n/agents/src/sdk/from-schema.ts
```diff
@@ -0,0 +1,318 @@
...
+	let toMessage: ((output: unknown) => AgentMessage | undefined) | undefined;
+	if (toolSchema.hasToMessage) {
+		if (executor.executeToMessageSync) {
+			const syncExecutor = executor.executeToMessageSync.bind(executor);
+			toMessage = (output: unknown): AgentMessage | undefined => {
+				return syncExecutor(toolSchema.name, output);
+			};
+		} else {
+			// Async executor cannot produce a synchronous toMessage result.
+			// Return undefined and log that executeToMessageSync is needed.
```
burivuhster: The comment here mentions "log that executeToMessageSync is needed", but nothing happens inside the toMessage function

packages/@n8n/agents/src/codegen/generate-agent-code.ts
```diff
@@ -0,0 +1,204 @@
...
+	assertAllHandled(rest);
+
+	// No manual indentation — Prettier formats at the end.
+	const parts: string[] = [];
+	let needsWorkflowTool = false;
+	parts.push(`export default new Agent('${escapeSingleQuote(agentName)}')`);
+
+	if (model.provider && model.name) {
+		parts.push(
+			`.model('${escapeSingleQuote(model.provider)}', '${escapeSingleQuote(model.name)}')`,
```
burivuhster: What about other ways arbitrary string can break this? E.g. line breaks or other special characters. As an additional precautious measure we may pass strings through something like
```
const escapedStr = JSON.stringify(rawString).slice(1, -1)
```

packages/@n8n/agents/src/codegen/generate-agent-code.ts
```diff
@@ -0,0 +1,204 @@
...
+		if (tool.outputSchemaSource) toolParts.push(`.output(${tool.outputSchemaSource})`);
+		if (tool.suspendSchemaSource) toolParts.push(`.suspend(${tool.suspendSchemaSource})`);
+		if (tool.resumeSchemaSource) toolParts.push(`.resume(${tool.resumeSchemaSource})`);
+		if (tool.handlerSource) toolParts.push(`.handler(${tool.handlerSource})`);
+		if (tool.toMessageSource) toolParts.push(`.toMessage(${tool.toMessageSource})`);
+		if (tool.requireApproval) toolParts.push('.requireApproval()');
+		if (tool.needsApprovalFnSource)
+			toolParts.push(`.needsApprovalFn(${tool.needsApprovalFnSource})`);
+		parts.push(`.tool(${toolParts.join('')})`);
+	}
```
burivuhster, line 184: Would it make sense to extract this into separate function?

packages/@n8n/agents/src/codegen/generate-agent-code.ts
```diff
@@ -0,0 +1,204 @@
...
+ */
+function assertAllHandled(_: Record<string, never>): void {
+	// intentionally empty — this is a compile-time-only check
+}
+
+// ---------------------------------------------------------------------------
+// Public API
+// ---------------------------------------------------------------------------
+
+export async function generateAgentCode(schema: AgentSchema, agentName: string): Promise<string> {
```
burivuhster, line 141: Please consider splitting this into multiple smaller functions, at least for larger sections (like tools, evaluations, imports)

burivuhster (commented):
Some comments/suggestions

packages/@n8n/agents/src/sdk/provider-capabilities.ts
yehorkardash: The current type provides better suggestions

packages/@n8n/agents/src/sdk/from-schema.ts
yehorkardash: Memory schema handling is lackluster in this PR, working on addressing it in another PR

packages/@n8n/agents/src/types/sdk/schema.ts
yehorkardash: I think we will end up with `schema: JSONSchema7` later on, will keep it that way for now

packages/@n8n/agents/src/sdk/from-schema.ts
yehorkardash: I think it's easier to do it in-place, otherwise we need separate structure or type to hold all evaluated schemas

burivuhster (dismissed):
Thanks for addressing my comments! Looks good, just take a look at cubic's comments

burivuhster (approved):
🚀

mike12345567 (approved):

Issue #28005, opened 2026-04-03
Code node issue


amold9167-art:
### Bug Description

Since Yesterday i am not able to process my code node. in n8n work flow.
I am getting this error for each code node.

Problem in node ‘Code in JavaScript‘
Task request timed out after 60 seconds

![Image](https://github.com/user-attachments/assets/b73d5ad8-4481-4e49-b888-1cdcca4cc494)

this is code used in code node.

for (const item of $input.all()) {
  const location = item.json["Location"];

  if (location && typeof location === "string") {
    item.json["Location"] = location
      .trim()
      .toLowerCase()
      .split(/\s+/)
      .map(word => word.charAt(0).toUpperCase() + word.slice(1))
      .join(" ");
  }
}

return $input.all();

### To Reproduce

Since Yesterday i am not able to process my code node. in n8n work flow.
I am getting this error for each code node.

Problem in node ‘Code in JavaScript‘
Task request timed out after 60 seconds

![Image](https://github.com/user-attachments/assets/b73d5ad8-4481-4e49-b888-1cdcca4cc494)

this is code used in code node.

for (const item of $input.all()) {
  const location = item.json["Location"];

  if (location && typeof location === "string") {
    item.json["Location"] = location
      .trim()
      .toLowerCase()
      .split(/\s+/)
      .map(word => word.charAt(0).toUpperCase() + word.slice(1))
      .join(" ");
  }
}

return $input.all();

### Expected behavior

India

### Debug Info

India

### Operating System

window

### n8n Version

10 pro

### Node.js Version

updated

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

PR #28016, opened 2026-04-03
fix(core): Fix duplicate pending request entry after task deferred in task broker

ssam18:
When a runner sends `runner:taskdeferred`, the request was being pushed back into `pendingTaskRequests` even though it was never removed to begin with `settleTasks` only sets `acceptInProgress = true` before calling `acceptOffer`. This left a duplicate entry that could consume future runner offers and starve legitimate requests from ever matching, eventually hitting the 60-second request timeout.

Removed the erroneous push since resetting `acceptInProgress = false` is sufficient and request is already in the queue waiting for the next offer. Also replaced the existing test that was papering over this bug (it manually called `setPendingTaskRequests` to avoid the duplicate) with a proper integration test using a real runner callback.

Fixes the issue #28005


ssam18:
@alexgrozav - Can you please review this PR?

alexgrozav:
@ssam18 Thank you for the PR! Unfortunately, it's not my area of expertise. I've reassigned to @tomi

tomi:
Hey @ssam18 and thanks for your PR. Is this essential the same fix as was done in #28315? Also your PR only has changes to test code, not in the actual implementation.

tomi:
@ssam18 I'll close this PR as I haven't heard back from you. Feel free to reopen if you think it's still needed.

PR #28021, opened 2026-04-03
fix(core): MCP tools called after workflow execution failure

luka-zivkovic:
## Summary

When a workflow execution fails or is cancelled, MCP tool nodes continued to execute and consume tokens. This was reported by a customer observing unnecessary token spend from MCP tool calls that fired after the parent agent had already errored out.

**Root cause:** The abort signal from the execution engine was not propagated to MCP tool calls, so in-flight and pending calls had no way to know the execution was already cancelled.

## Related Linear ticket

https://linear.app/n8n/issue/NODE-4546


yehorkardash:
@claude /review

packages/@n8n/nodes-langchain/nodes/agents/Agent/agents/ToolsAgent/V3/helpers/runAgent.ts
```diff
@@ -116,10 +116,13 @@ export async function runAgent(
...
-		const modelResponse = await executorWithTracing.invoke({
-			...invokeParams,
-			chat_history: chatHistory,
-		});
+		const modelResponse = await executorWithTracing.invoke(
+			{
+				...invokeParams,
+				chat_history: chatHistory,
+			},
+			executeOptions,
```
yehorkardash, line 124: is this passed to stop LLM execution?

luka-zivkovic: Yes, yes, correct. Added it because the streaming path already had this, but the non-streaming path was missing it

yehorkardash (approved):

PR #28036, opened 2026-04-04
feat(Postgres Node): Sort PostgreSQL schemas and tables alphabetically when fetching

GauthierPLM:
## Summary

When fetching the schema and tables from Psotgres' `information_schema`, order the results alphabetically to make it easier to find the needed table.

Currently, the order is random (depends on Postgres' internal data ordering), which is inconsistent and hard to use. I personally lose time each time I config. a PG node as I have to scroll in the tables / schema list and check each item to find the table I want. Alphabetic sorting would drastically simplify the search.

## Review / Merge checklist

- [ ] Tests included. 


garritfra (approved):
Thanks!

PR #28042, opened 2026-04-04
fix(MongoDB Chat Memory Node): Add connection pool limit

### LINKED ISSUE #27980: MongoDB Chat Memory node leaks connections — never closes MongoClient instances
### Bug Description

Hi n8n team,

We’re experiencing a critical MongoDB connection leak caused by the MongoDB Chat Memory node ([@n8n](https://community.n8n.io/u/n8n)/n8n-nodes-langchain.memoryMongoDbChat).

The problem

The MemoryMongoDbChat.node.js supplyData() method creates a new MongoClient on every workflow execution with no pool size limit (defaults to 100 connections per client). While a closeFunction is returned, it is not reliably called by the execution engine — particularly when used inside AI agent workflows with
multiple tool calls or when errors occur.

This results in leaked MongoClient instances that maintain heartbeat connections indefinitely, accumulating with every execution. In our case, we reached 355 leaked connections from n8n alone (out of 409 total), nearly exhausting MongoDB’s connection limit. MongoDB was only up for 4 minutes before reaching this
state.

Reproduction

Create a workflow with a webhook trigger → AI Agent → MongoDB Chat Memory node
Send repeated requests to the webhook (we hit ~1,000 executions in 6 hours)
Monitor MongoDB connections: db.serverStatus().connections
Connections grow linearly with executions and never drop

Root cause in code

[@n8n](https://community.n8n.io/u/n8n)/n8n-nodes-langchain/dist/nodes/memory/MemoryMongoDbChat/MemoryMongoDbChat.node.js

// Line 97 — no pool limits, no idle timeout
const client = new mongodb_2.MongoClient(connectionString);
await client.connect();

The closeFunction is defined but depends on the execution engine calling it. For high-frequency AI agent workflows, this doesn’t happen consistently.

Suggested fix

At minimum, add pool constraints to the MongoClient constructor:

const client = new mongodb_2.MongoClient(connectionString, {
minPoolSize: 0,
maxPoolSize: 1,
maxIdleTimeMS: 30000
});

Ideally, the node should also use a shared/cached MongoClient per credential rather than creating a new one per execution, similar to how other database nodes handle connection reuse.

Our workaround

We patched the file inside the container with the above options and restarted n8n. Connections dropped from 359 to 8 immediately.

Environment

n8n version: latest (Docker, n8nio/n8n:latest)
MongoDB: 8.2.6
Node type: [@n8n](https://community.n8n.io/u/n8n)/n8n-nodes-langchain.memoryMongoDbChat
Trigger frequency: ~3 executions/minute via webhook
Note: The regular MongoDB node (n8n-nodes-base.mongoDb) properly closes connections in a finally block, but also lacks maxPoolSize limits — worth adding as a safeguard there too.

Thanks for looking into this.

### To Reproduce

Create a workflow with a webhook trigger → AI Agent → MongoDB Chat Memory node
Send repeated requests to the webhook (we hit ~1,000 executions in 6 hours)
Monitor MongoDB connections: db.serverStatus().connections
Connections grow linearly with executions and never drop

### Expected behavior

To close connection after workflow completes processing

### Debug Info

Exists in the description

### Operating System

Ubuntu n8n in docker

### n8n Version

latest

### Node.js Version

latest

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

Vitalini:
## Summary

Fixes #27980 — MongoDB Chat Memory node leaks connections by creating a new `MongoClient` per workflow execution without pool constraints, and not closing the client when setup errors occur.

## Problem

The `MemoryMongoDbChat` node creates a new `MongoClient` on every `supplyData()` call with **no connection pool limits** (MongoDB driver defaults to 100 connections per client). While a `closeFunction` is returned, it is not reliably called by the execution engine — particularly in AI agent workflows with multiple tool calls or when errors occur.

This causes leaked `MongoClient` instances that maintain heartbeat connections indefinitely. Users have reported reaching **355 leaked connections in just 4 minutes** of uptime (~1,000 executions in 6 hours).

## Changes

1. **Added connection pool constraints** to `MongoClient`:
   - `minPoolSize: 0` — no idle connections kept alive
   - `maxPoolSize: 1` — each execution uses at most 1 connection (this node only needs a single collection operation)
   - `maxIdleTimeMS: 30_000` — idle connections are closed after 30 seconds as a safety net

2. **Added `client.close()` in the catch block** — if `client.connect()` or any subsequent setup fails, the client is now properly closed. Previously, the `closeFunction` was only registered on success, so errors during setup would leak the client entirely.

## Why this approach

- **Minimal and safe**: Does not change the node's architecture or execution flow
- **Matches the issue reporter's verified workaround**: They patched the same options in production and saw connections drop from 359 to 8 immediately
- **Defense in depth**: Even if `closeFunction` is not called, connections will self-close after 30s of idle time, and each client only holds 1 connection instead of up to 100
- **Consistent with existing patterns**: The regular MongoDB node (`nodes-base/MongoDb`) already closes the client in a `finally` block

## Testing

- Verified the fix compiles with no type errors
- The change is backward-compatible — no new parameters or behavior changes for users
- The connection pool options are standard MongoDB driver configuration

## Review / Merge checklist


burivuhster (approved):
Thanks for the contribution, @Vitalini 
Looks good to me

burivuhster:
@Vitalini We recently introduced a new requirement for all PRs to have an acknowledgment:

Can you please add it to the PR description?

alexbevi:
This will be really helpful for folks running into this issue. Would love to see this merged! @Vitalini did you see the [previous comment](https://github.com/n8n-io/n8n/pull/28042#issuecomment-4224001399)? Seems the PR description just needs an update.

burivuhster:
@Vitalini I've gone ahead and added the reviewer acknowledgment to the PR description myself. Hope that's all good with you, and thanks for the contribution!

Vitalini:
@burivuhster thank you! 🙏

PR #28054, opened 2026-04-06
fix(nodes-base): reload SQL sandbox context after SELECT failure to prevent "Isolate is disposed" error

octo-patch:
Fixes #27861

## Summary

When a SELECT query in the Merge node's **Combine By SQL** mode fails (e.g. because the 5-second sandbox timeout is exceeded on large datasets like 25k–100k rows), `isolated-vm` disposes the isolate. The existing code catches that error and **silently falls through** to the non-SELECT execution path — but that path reuses the now-disposed `context`, causing the confusing `"Isolate is disposed"` error users see instead of a meaningful timeout message.

The fix is a one-line addition: after the SELECT catch block, reload the sandbox context via `loadAlaSqlSandbox()`. Since `loadAlaSqlSandbox` already checks `sandboxIsolate.isDisposed` and recreates the isolate when needed, calling it again after a failure ensures the fallback path gets a valid, fresh context.

## Changes

- `packages/nodes-base/nodes/Merge/v3/actions/mode/combineBySql.ts`
  - Changed `const context` → `let context` so the variable can be reassigned
  - Added `context = await loadAlaSqlSandbox()` in the SELECT catch block

## Testing

The fix can be verified by:
1. Creating a Merge node (v3.2) with **Combine By SQL** mode
2. Using a `SELECT` query on a large dataset that triggers the 5-second timeout
3. Before the fix: error is "Issue while executing query | Isolate is disposed"
4. After the fix: the fallback path runs cleanly with a valid context (and produces a proper timeout error if the dataset is still too large)

## Review / Merge checklist

- [ ] Tests included.
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1`

### DIFF
--- a/packages/nodes-base/nodes/Jenkins/Jenkins.node.ts
+++ b/packages/nodes-base/nodes/Jenkins/Jenkins.node.ts
@@ -452,10 +452,13 @@ export class Jenkins implements INodeType {
 			async getJobParameters(this: ILoadOptionsFunctions): Promise<INodePropertyOptions[]> {
 				const job = this.getCurrentNodeParameter('job') as string;
 				const returnData: INodePropertyOptions[] = [];
-				const endpoint = `/job/${job}/api/json?tree=actions[parameterDefinitions[*]]`;
-				const { actions } = await jenkinsApiRequest.call(this, 'GET', endpoint);
-				for (const { _class, parameterDefinitions } of actions) {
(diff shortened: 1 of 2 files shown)


tannerwoodrum:
Would love if this got some attention! one of our critical workflows is depended on this being resolved any way to escalate?

PR #28065, opened 2026-04-06
feat: Add ToolFromNode implementation

yehorkardash:
## Summary

Adds `ToolFromNode` — a new tool type for SDK agents that executes a single n8n node as a callable tool, without requiring a full workflow. This complements the existing `WorkflowTool` (multi-step workflows) with a lightweight, single-node alternative.

### New execution mode: `ephemeral`

A new `ephemeral` workflow execution mode drives the node execution:
- **PinData support** — LLM input is pinned on a disposable Manual Trigger node so n8n expressions like `={{ $json.field }}` resolve against it
- **Unlimited concurrency** — bypasses the production queue (like `manual`/`integrated`)
- **Synchronous result** — the caller awaits the result inline (needed for tool calls)
- **No statistics / insights** — ephemeral workflows are filtered out of listings, execution history, and telemetry
Each tool call creates a transient `ManualTrigger → TargetNode` workflow, executes it, and deletes it immediately via `EphemeralNodeExecutor`.

**Important** This PR does not update agent builder LLM with instructions and tools to fetch node definitions. This will be added later on

https://linear.app/n8n/issue/AGENT-5

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/entities/types-db.ts
```diff
@@ -189,6 +189,7 @@ export namespace ExecutionSummaries {
 		id: string;
 		finished: boolean;
 		mode: WorkflowExecuteMode;
+		excludeModes: WorkflowExecuteMode[];
```
burivuhster:
```suggestion
```

burivuhster: We don't need this anymore, right?

packages/cli/src/modules/agents/types/node-tool.ts
```diff
@@ -0,0 +1,113 @@
...
+		}
+		return this;
+	}
+
+	/** Produce a marker BuiltTool. Resolved at runtime by the platform via tool.metadata. */
+	build(): BuiltTool {
+		const toolName = this._name ?? `node-tool-${this._nodeType}`;
+		return {
+			name: toolName,
+			description: this._description ?? `Execute the ${this._nodeType} node`,
```
burivuhster, line 101: Would `NodeHelpers.getToolDescriptionForNode` be useful here?

packages/cli/src/modules/agents/agents-credential-provider.ts
```diff
@@ -65,32 +48,24 @@ export class AgentsCredentialProvider implements CredentialProvider {
...
-			.map((c) => ({
-				id: c.id,
-				name: c.name,
-				type: c.type,
-			}));
+		return allCredentials.map((c) => ({
+			id: c.id,
+			name: c.name,
+			type: c.type,
+		}));
```
burivuhster, line 69: Don't we need it anymore?

packages/cli/src/modules/agents/node-tool-factory.ts
```diff
@@ -0,0 +1,106 @@
...
+			});
+		})
+		.build();
+}
+
+/**
+ * Build a Zod schema from a JSON Schema object stored in the tool metadata.
+ * Falls back to a passthrough object schema if no schema is stored.
+ */
+function buildZodSchema(
```
burivuhster: Can we utilize `@n8n/json-schema-to-zod` here?

burivuhster (commented):
Thanks for addressing the workflow persistence concern! Looks good, just a few minor suggestions

packages/cli/src/modules/agents/agents-credential-provider.ts
yehorkardash: Do you mean credentials provider or map method? Credentials provider is still used and will be used more

packages/cli/src/modules/agents/types/node-tool.ts
yehorkardash: Not really, it can't use imports from n8n-workflow, because this class runs in isolate. I will be changing in in JSON config PR

packages/cli/src/modules/agents/agents-credential-provider.ts
burivuhster: Sorry, I meant the filter by only supported credentials that is being removed here

packages/cli/src/modules/agents/node-tool-factory.ts
yehorkardash: @n8n/agents has support for json schema, but it wasn't properly exposed. Updated the libary to accept json schema in `input()`

packages/cli/src/modules/agents/agents-credential-provider.ts
yehorkardash: Agent builder uses it to search for credentials, but this provider was returning only a subset of credentials. I want to reuse search_credentials tool from workflow-sdk, but later on

burivuhster (approved):
🚀

PR #28066, opened 2026-04-06
feat: Add memory schema introspection and default memory

yehorkardash:
## Summary

**1. `MemorySchema` — richer structured fields**
Replaces `storage: 'memory' | 'custom'` with `name`, `constructorName`, and `connectionParams` fields sourced from `BuiltMemory.describe()`. Adds `semanticRecall.scope`, `workingMemory.scope`, and `titleGeneration` fields. `source` is kept temporarily during migration.

**2. No source code for memory reconstruction**
`fromSchema` no longer eval-reconstructs backends from the `source` string. Instead, `FromSchemaOptions.memoryRegistry` maps backend names to async factory functions. All memory options (`semanticRecall`, `workingMemory`, `titleGeneration`) are reconstructed from schema fields directly. Code generation in `generate-agent-code.ts` follows the same approach via `buildMemoryChain()` and `MemoryImportMap`.

**3. `BaseMemory` abstract class**
New `storage/base-memory.ts` provides a default `describe()` implementation. `SqliteMemory` and `PostgresMemory` both extend it. `PostgresMemory` gains a credential-based constructor path that resolves connection config at runtime without embedding secrets in `connectionParams`.

**4. `N8nMemory` — default persistent backend**
New TypeORM-backed `N8nMemory` service persists threads, messages, and working memory in the n8n database via a new migration (`agents_resources`, `agents_threads`, `agents_messages`). A sandbox-side `N8nMemoryMarker` stub is exposed as `N8nMemory` in `@n8n/agents-utils` — it only implements `describe()` so schemas serialise correctly; the host maps `'n8n'` in `memoryRegistry` to the real DI service. `N8nMemory` is now the default recommended storage in all code generation prompts.

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/db/src/migrations/common/1777000000000-CreateAgentMemoryTables.ts
```diff
@@ -0,0 +1,42 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class CreateAgentMemoryTables1777000000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
```
mike12345567, line 4: One thing I ended up implementing for the memory on the workflow builder, having a column with a summary or a plan for how to handle a summary (like its always the first message after) after a compactions has occurred.

mike12345567 (approved):
LGTM - had a play with it, works nicely!

packages/cli/src/modules/agents/integrations/n8n-memory.ts
```diff
@@ -0,0 +1,224 @@
...
+		threadId: string,
+		opts?: { limit?: number; before?: Date },
+	): Promise<AgentDbMessage[]> {
+		const where: FindOptionsWhere<AgentMessageEntity> = opts?.before
+			? { threadId, createdAt: LessThan(opts.before) }
+			: { threadId };
+
+		const entities = await this.messageRepository.find({
+			where,
+			order: { createdAt: 'ASC' },
```
mike12345567, line 87: Should this be sorted by oldest? Newer messages should likely come first.

packages/cli/src/modules/agents/agents.service.ts
```diff
@@ -325,6 +329,16 @@ export class AgentsService {
 		return { code, schema: freshSchema, updatedAt: saved.updatedAt.toISOString() };
 	}
 
+	private getMemoryRegistry(): Record<string, agents.MemoryFactory> {
+		return {
+			n8n: () => this.n8nMemory,
+			sqlite: (params) => new SqliteMemory(SqliteMemoryConfigSchema.parse(params)),
+			postgres: () => {
+				throw new Error('Not implemented');
```
mike12345567, line 337: Should this be not implemented, seems like there is quite a bit of Postgres implementation already.

packages/cli/src/modules/agents/integrations/n8n-memory.ts
yehorkardash: In agent framework messages are sorted in ASC order too, which represent a message history, where the most recent message is the last in the list

packages/cli/src/modules/agents/agents.service.ts
yehorkardash: I need to add more validation and refine generate-agent-code for it to work properly, so I decided to not do the implementation for now, until we switch to json config

PR #28067, opened 2026-04-06
feat(core): Support npm dist-tags in community node installation

### LINKED ISSUE #27401: Community Node installation fails with "Invalid version: beta" when using npm dist-tag instead of semver version
### Bug Description

When installing a Community Node via Settings → Community Nodes in n8n version 2.11.2 (docker, Linux), using an npm dist-tag (e.g. `@beta`) as the version fails.

Example input:
```
n8n-nodes-imap@beta
```
Resulting error:
```
Error installing new package
Invalid version: beta
```

It appears n8n treats everything after `@` as a semver version and fails for plain dist-tags, which are standard in the npm ecosystem.

### To Reproduce

1. In n8n 2.11.2 (docker, Linux), go to Settings > Community Nodes.
2. Attempt to install: `n8n-nodes-imap@beta`
3. Observe error: `Error installing new package` → `Invalid version: beta`

### Expected behavior

It should be possible to install Community Nodes using valid npm dist-tags (e.g. `@beta`, `@next`) as the version OR the application should provide a clearer message explaining that only semver versions are supported.

### Debug Info

n8n 2.11.2 (docker), Community Nodes via UI
n8n-nodes-imap npm package: https://www.npmjs.com/package/n8n-nodes-imap
Error: `Error installing new package` → `Invalid version: beta`

### Operating System

Linux (Docker)

### n8n Version

2.11.2

### Node.js Version

 24.13.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

krisn0x:
## Summary

Community node installation failed with `Invalid version: beta` when users specified npm dist-tags (e.g. `n8n-nodes-imap@beta`). The validation only accepted semver versions, but dist-tags like `beta`, `next`, and `latest` are valid npm version specifiers.

Unsafe specs (git URLs, tarball URLs, semver ranges) remain blocked by the upstream `INVALID_OR_SUSPICIOUS_PACKAGE_NAME` regex.

https://linear.app/n8n/issue/NODE-4653
fixes #27401

## Review / Merge checklist

- [x] Tests included.

Co-authored by Claude Opus 4.6


.gitignore
```diff
@@ -59,3 +59,6 @@ packages/cli/src/commands/export/outputs
 .claude/plans/
 lefthook-local.yml
 .playwright-mcp
+.n8n
+.superset
+.conductor
```
DawidMyslak: Seems to be unrelated to this PR?

packages/cli/src/modules/community-packages/npm-utils.ts
```diff
@@ -185,8 +186,14 @@ export async function checkIfVersionExistsOrThrow(
...
-			const versionInfo = jsonParse(stdout);
-			if (versionInfo === version) {
+			const resolvedVersion = jsonParse(stdout);
+			// npm resolves dist-tags to concrete versions, so we can only match on exact semver
+			const isDistTag = valid(version) === null;
+			if (
+				typeof resolvedVersion === 'string' &&
+				resolvedVersion.length > 0 &&
+				(isDistTag || resolvedVersion === version)
+			) {
```
DawidMyslak: When `isDistTag` is true, any non-empty `resolvedVersion` string passes here. Could `stdout` ever contain something unexpected (e.g. an error message) that would slip through? Would adding a `valid(resolvedVersion)` check be worth it?

DawidMyslak (changes requested):

.gitignore
krisn0x: separated

packages/cli/src/modules/community-packages/npm-utils.ts
krisn0x: valid, also updated the isDistTag variable name to be more precise and refactored that block to be a bit easier to read

packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
```diff
@@ -106,12 +106,20 @@ describe('CommunityPackagesService', () => {
...
 		});
 
-		test.each(['invalid', '1.a.b'])('should fail with invalid version', (version) => {
+		test.each(['1.a.b', '1invalid'])('should fail with invalid version', (version) => {
 			expect(() =>
 				communityPackagesService.parseNpmPackageName(`n8n-nodes-test@${version}`),
 			).toThrow(`Invalid version: ${version}`);
 		});
 
+		test.each(['beta', 'next', 'latest', 'invalid'])(
```
DawidMyslak: So `invalid` is valid 😄?

DawidMyslak (dismissed):

packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
DawidMyslak: @krisn0x this is not a blocker, just raising this before we add it to the merge queue.

krisn0x: Np, also did another pass on the tests and made some more changes:
 - Replaced 'invalid' with 'canary' and 'rc-1'
  - Added '-starts-with-dash' to invalid version tests — boundary case for the dist-tag regex
  - Added test that dist-tag resolution rejects non-semver CLI output (e.g. beta → 'not-a-version')

DawidMyslak (approved):

PR #28078, opened 2026-04-07
refactor(editor): Normalize sharedWithProjects field in workflow document store (no-changelog)

alexgrozav:
## Summary

Moves the `sharedWithProjects` field from the workflows EE store into the `workflowDocumentStore` composable pattern, consistent with how other workflow document fields (homeProject, scopes, meta, etc.) are managed. The share modal and workflow role detection now read from the document store with a fallback to the legacy workflow object.

https://linear.app/n8n/issue/CAT-2741

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/app/components/WorkflowShareModal.ee.vue
```diff
@@ -203,6 +204,7 @@ const onSave = async () => {
 			workflowId,
 			sharedWithProjects: sharedWithProjects.value,
 		});
+		workflowDocumentStore.value?.setSharedWithProjects(sharedWithProjects.value);
```
autologie: `workflowDocumentStore` is instantiated with `props.data.id` and looking at the logic above, `workflowId` here might be different from it (if the workflow isn't saved yet). A safer implementation would be

```ts
useWorkflowDocumentStore(createWorkflowDocumentId(workflowId).setSharedWithProjects(sharedWithProjects.value);
```

There's also a chance though, that it will never happen because of auto saving and we can simplify by cleaning up the saving logic itself.

alexgrozav: Great catch! I've updated the code.

autologie (approved):
I don't want to increase the scope of this PR but I feel we need a follow up to cleanup or deprecate `setWorkflowSharedWith` in `workflows.ee.store`, because the value is set but we don't read the set value anymore.

Anyway the changes look good 👍

PR #28093, opened 2026-04-07
fix(AWS DynamoDB Node): Add option to disable auto-parsing of numeric strings

BerniWittmann:
## Summary

Adds an **Automatically Parse Numbers** toggle (default: `true`) to the PutItem/upsert operation, addressing the issue where numeric strings (e.g. `"34"`) were always stored as DynamoDB Number (`N`) type.

- **When enabled (default):** preserves existing behavior — numeric-looking strings are stored as `N`
- **When disabled:** only actual JavaScript numbers are classified as `N` — strings stay as `S` regardless of content

This is a non-breaking change: existing workflows continue to work identically.

Based on the approach proposed by @umut-polat in commit `be16a1a845`, adapted for the current codebase.

https://linear.app/n8n/issue/NODE-4576

## Review / Merge checklist

- [x] Tests included.


packages/nodes-base/nodes/Aws/DynamoDB/__test__/utils.test.ts
```diff
@@ -0,0 +1,64 @@
...
+	});
+
+	it('should classify booleans as DynamoDB BOOL type', () => {
+		const result = adjustPutItem({ active: true } as unknown as PutItemUi);
+		expect(result).toEqual({ active: { BOOL: 'true' } });
+	});
+
+	it('should classify plain objects as DynamoDB M type', () => {
+		const result = adjustPutItem({ meta: { foo: 'bar' } } as unknown as PutItemUi);
+		expect(result).toEqual({ meta: { M: '[object Object]' } });
```
RomanDavydchuk: This feels wrong, I would expect `M` to be either stringified JSON or a regular object, i.e.
```javascript
		expect(result).toEqual({ meta: { M: '{ "foo": "bar" }' } });
```
or something like this
```javascript
		expect(result).toEqual({ meta: { M: { foo: { S: 'bar' } } } });
```
I'm not too familiar with DynamoDB's API to be sure right off the bat

Even though the PR does not touch any logic related to object handling, if this test passes it probably means that there is a bug that has been there for a while

RomanDavydchuk (changes requested):
This PR does fix a bug where it's pretty much impossible to insert a string with a numeric value as an actual string into DynamoDB, but there are two issues I see:
1. This is a breaking change. Before, every "number-looking" string would be treated as an actual number, while in some cases it's undesirable, some users might be relying on this behavior, either knowingly or not. We need to avoid such changes because they might break users' production workflows when they update to a new n8n version
2. There is "Define Below for Each Column" input mode, i.e. you can specify the fields in the UI instead of using the input items to the node:

Image

All of the values are actually strings by default, but the current behavior parses strings into numbers, if possible, so the `num` field in the screenshot above would be treated as a number (`N`). After the change it would be treated as a string (which is also a breaking change). It is still possible to make sure it's treated as a number, you'd need to use `Expression` mode:

Image

So it would be still possible to specify number through the UI, but more tricky and less intuitive

I see a couple of ways to fix this:
1. Introduce a new flag, called something like `Automatically parse numbers` with a description clearly explaining what it does. We'd leave this to `true` by default, so that it won't break existing workflows
2. Introduce a new node version and make a change that would affect only that version. We have [this field](https://github.com/n8n-io/n8n/blob/a6b051bfe310362b7b7882cc12af4f17ece6ffa9/packages/nodes-base/nodes/Aws/DynamoDB/AwsDynamoDB.node.ts#L38) that holds the node's version(s). You'd update it to `version: [1, 1.1]` and use `this.getNode().typeVersion` in the `execute` function to get what is the node's version during execution. Then, if it's an old version (`1`) - keep old behavior, otherwise - new behavior. Nodes that were already added to the canvas would keep the old version, while newly added nodes would have version `1.1`. P.S. This approach seems more "correct", but as I mentioned it would make it less intuitive to specify numeric values through the UI, so maybe we can add a dropdown to specify the field type, and not just `Field ID` and `Field Value`? But that would increase the scope for this fix

Lastly, I left a comment in code regarding one test case, looks like it uncovered another unrelated bug

Feel free to reach out if you want to chat more about how to go about this fix 🙌

RomanDavydchuk (approved):
LGTM and works as expected

PR #28097, opened 2026-04-07
feat(core): Add DB infrastructure for trusted keys and key sources (no-changelog)

afitzek:
## Summary

This PR adds the database layer for persisting trusted keys and their sources in the token exchange module. Currently, trusted keys are loaded from environment variables at startup and held only in memory. This change introduces two new tables (`trusted_key_source` and `trusted_key`) with TypeORM entities, repositories, and a reversible migration, laying the groundwork for dynamic key management (e.g., JWKS rotation and UI-managed keys).

**What's included:**
- **Migration** (`1776000000000-CreateTrustedKeyTables`): Creates `trusted_key_source` (with status tracking, config, and refresh metadata) and `trusted_key` (composite PK on `sourceId` + `kid`, FK cascade to source) for both PostgreSQL and SQLite.
- **Entities**: `TrustedKeySourceEntity` and `TrustedKeyEntity` with proper TypeORM decorators and relations.
- **Repositories**: `TrustedKeySourceRepository` and `TrustedKeyRepository` (with `findBySourceAndKid` and `findAllByKid` query methods).
- **Schema additions**: New `ui` source type in the Zod discriminated union, `TrustedKeyData` interface for the serialized JSON column, and exported type aliases (`TrustedKeySourceType`, `TrustedKeySourceStatus`).
- **Config**: `keyRefreshIntervalSeconds` env var for future leader-only refresh scheduling.
- **Module registration**: Entities lazy-loaded in `TokenExchangeModule.entities()`.

**Key design decisions:**
- Composite primary key (`sourceId`, `kid`) on `trusted_key` allows the same `kid` across different sources while enforcing uniqueness within a source.
- `trusted_key.data` is a JSON text column storing `TrustedKeyData` (algorithms, PEM, issuer, audience, roles, expiry) — keeps the schema stable as key metadata evolves.
- `trusted_key_source.status` tracks source health (`pending` → `healthy` | `error`) for observability of JWKS refresh cycles.
- `ui` source type is recognized in the schema but explicitly skipped during initialization (not yet supported).

### How to test manually

1. Start n8n with the token exchange feature flag enabled
2. Verify the migration runs without errors on a fresh database (check startup logs)
3. Confirm both tables exist: `SELECT * FROM trusted_key_source; SELECT * FROM trusted_key;`

### Related tickets

closes https://linear.app/n8n/issue/IAM-523/2a31-trustedkeystore-db-tables-entities-and-repositories

## Review checklist

- [x] Tests included.
   - Integration tests cover: CRUD operations, FK cascade deletes, composite PK uniqueness, cross-source kid sharing, `findAllByKid` queries, optional field round-tripping, and orphan key rejection.

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1776000000000-CreateTrustedKeyTables.ts
+++ b/packages/@n8n/db/src/migrations/common/1776000000000-CreateTrustedKeyTables.ts
@@ -1,0 +1,34 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+const sourceTable = 'trusted_key_source';
+const keyTable = 'trusted_key';
+
+export class CreateTrustedKeyTables1776000000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable(sourceTable).withColumns(
+			column('id').varchar(36).primary.notNull,
+			column('type').varchar(32).notNull,
+			column('config').text.notNull,
+			column('status').varchar(32).notNull.default("'pending'"),
+			column('lastError').text,
+			column('lastRefreshedAt').timestamp(),
+		).withTimestamps;
+
+		await createTable(keyTable)
+			.withColumns(
+				column('sourceId').varchar(36).primary.notNull,
+				column('kid').varchar(255).primary.notNull,
+				column('data').text.notNull,
+			)
+			.withForeignKey('sourceId', {
+				tableName: sourceTable,
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			}).withCreatedAt;
+	}
+
+	async down({ schemaBuilder: { dropTable } }: MigrationContext) {
+		await dropTable(keyTable);
+		await dropTable(sourceTable);
+	}
+}
--- a/packages/cli/src/modules/token-exchange/database/entities/trusted-key-source.entity.ts
+++ b/packages/cli/src/modules/token-exchange/database/entities/trusted-key-source.entity.ts
@@ -1,0 +1,25 @@
+import { DateTimeColumn, WithTimestamps } from '@n8n/db';
+import { Column, Entity, PrimaryColumn } from '@n8n/typeorm';
+
+import type { TrustedKeySourceStatus, TrustedKeySourceType } from '../../token-exchange.schemas';
+
+@Entity('trusted_key_source')
+export class TrustedKeySourceEntity extends WithTimestamps {
+	@PrimaryColumn('varchar', { length: 36 })
+	id: string;
+
+	@Column({ type: 'varchar', length: 32 })
+	type: TrustedKeySourceType;
+
+	@Column('text')
+	config: string;
+
+	@Column({ type: 'varchar', length: 32, default: 'pending' })
+	status: TrustedKeySourceStatus;
+
+	@Column({ type: 'text', nullable: true })
+	lastError: string | null;
+
+	@DateTimeColumn({ nullable: true })
+	lastRefreshedAt: Date | null;
+}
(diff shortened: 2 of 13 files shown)


packages/@n8n/db/src/migrations/common/1776000000000-CreateTrustedKeyTables.ts
BGZStephen, line 1: What do you think about implementing the database table for key rotations? If we add a generic `instance_key` table, and simply have this as a `type` of `trusted_key:token_exchange` then it would fit in to the proposed key rotation work, and the API's would also be re-usable (list, create, update, delete)

afitzek: I don't think we should mix these concerns, and I think the tables are going to be a bit different. I would suggest we keep it like this. We can still migrate to a shared table during the key rotation work, if it makes sense.

BGZStephen (approved):

packages/@n8n/db/src/migrations/common/1776000000000-CreateTrustedKeyTables.ts
```diff
@@ -0,0 +1,34 @@
...
+			column('config').text.notNull,
+			column('status').varchar(32).notNull.default("'pending'"),
+			column('lastError').text,
+			column('lastRefreshedAt').timestamp(),
+		).withTimestamps;
+
+		await createTable(keyTable)
+			.withColumns(
+				column('sourceId').varchar(36).primary.notNull,
+				column('kid').varchar(255).primary.notNull,
```
RicardoE105, line 20: nit: should we create an index here? do you see a pattern where we query only by kid? or is it an overkill?

packages/cli/src/modules/token-exchange/token-exchange.config.ts
```diff
@@ -20,6 +20,10 @@ export class TokenExchangeConfig {
 	@Env('N8N_TOKEN_EXCHANGE_TRUSTED_KEYS')
 	trustedKeys: string = '';
 
+	/** Interval in seconds between trusted key refresh runs (leader only). */
+	@Env('N8N_TOKEN_EXCHANGE_KEY_REFRESH_INTERVAL_SECONDS')
```
RicardoE105, line 24: nit: This is not used anywhere?

packages/cli/src/modules/token-exchange/database/entities/trusted-key-source.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+	id: string;
+
+	@Column({ type: 'varchar', length: 32 })
+	type: TrustedKeySourceType;
+
+	@Column('text')
+	config: string;
+
+	@Column({ type: 'varchar', length: 32, default: 'pending' })
+	status: TrustedKeySourceStatus;
```
RicardoE105, line 18: nit: should we add a comment to the column to document the possible values in the db?  See the example below from the insights column

```
const typeComment = '0: time_saved_minutes, 1: runtime_milliseconds, 2: success, 3: failure';

await createTable(names.t.insightsRaw)
	.withColumns(
		column('id').int.primary.autoGenerate2,
		column(names.c.insightsRaw.metaId).int.notNull,
		column('type').int.notNull.comment(typeComment),
		column('value').int.notNull,
		column('timestamp').timestampTimezone(0).default('CURRENT_TIMESTAMP').notNull,
	)
	.withForeignKey(names.c.insightsRaw.metaId, {
		tableName: names.t.insightsMetadata,
		columnName: names.c.insightsMetadata.metaId,
		onDelete: 'CASCADE',
	});
```

RicardoE105 (approved):
Looking good 🚀 . Left some comments but are not blocking

PR #28100, opened 2026-04-07
fix(ai-builder): Expose credential account context to prevent prompt/credential mismatch

aalises:
## Summary
- The AI agent had no way to know which account a credential belongs to ... `list-credentials` only returned `{id, name, type}`. This caused prompts (e.g. text classifier) to be tailored to the wrong context when the browser showed one email but the credential connected to a different one.
- `encryptAndSaveData` now extracts and persists an `accountIdentifier` (email/username) at OAuth connection time by checking direct token fields, decoding JWT `id_token` payloads, and inspecting nested user objects
- `list-credentials` and `get-credential` tools return `accountIdentifier` so the orchestrator can match credentials to the correct context
- For credentials connected before this change, `getAccountContext()` falls back to runtime extraction from stored token data and credential fields

## Related
https://www.notion.so/n8n/Context-of-credential-not-considered-when-creating-prompt-in-text-classifier-step-3365b6e0c94f80c19869cddc07c324fb


packages/cli/src/oauth/oauth.service.ts
```diff
@@ -175,6 +175,15 @@ export class OauthService {
 		toUpdate: ICredentialDataDecryptedObject,
 		toDelete: string[] = [],
 	) {
+		if (toUpdate.oauthTokenData && typeof toUpdate.oauthTokenData === 'object') {
```
Cadiac, line 178: I wonder is this additiona a good idea? Should we get some feedback on this addition from the IAM / payday team perhaps, and see what they think of us extending credentials like this?

aalises: Good call, I've tightened this up based on a security review (see latest commit):

1. **The `accountIdentifier` is stored inside the encrypted credential blob** (via `toUpdate` before `credentials.updateData()`), not as a separate column. It's encrypted at rest alongside the token data using the existing credential encryption, so it doesn't change the credential schema or database structure but an additional key in the encrypted JSON payload.

2. **The value is masked before it reaches the LLM**: `alice@gmail.com` becomes `al***@gmail.com`. The full email is never sent to the external AI provider.

3. **No new decryption surface**: for credentials that already have the pre-stored `accountIdentifier`, we use redacted decryption (secrets are blanked). Unredacted decryption is only used as a fallback for legacy credentials that need `oauthTokenData` scanning.

I think looping in the IAM team would be good either way to:
- Adding a derived field to the encrypted credential payload at OAuth callback time
- The masked identifier being visible to the AI orchestrator

packages/@n8n/instance-ai/src/tools/credentials/list-credentials.tool.ts
```diff
@@ -10,7 +10,9 @@ export const listCredentialsInputSchema = z.object({
 export function createListCredentialsTool(context: InstanceAiContext) {
 	return createTool({
 		id: 'list-credentials',
-		description: 'List credentials accessible to the current user. Never exposes secret data.',
+		description:
+			'List credentials accessible to the current user. Never exposes secret data. ' +
+			'Returns a masked accountIdentifier (e.g. "al***@gmail.com") when available, so you know which account each credential is connected to.',
```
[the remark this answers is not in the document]
aalises: This is incorrect. The masking is implemented in the adapter layer... `getAccountContext()` in `instance-ai.adapter.service.ts` applies a `mask()` function to every identifier before returning it. The tool calls `context.credentialService.getAccountContext()`, which returns the already-masked value. The description is accurate.

packages/@n8n/instance-ai/src/tools/credentials/list-credentials.tool.ts
```diff
@@ -51,10 +53,26 @@ export function createListCredentialsTool(context: InstanceAiContext) {
...
-			};
+			if (!context.credentialService.getAccountContext) {
+				return {
+					credentials: page.map(({ id, name, type }) => ({ id, name, type })),
+					total,
+				};
+			}
+
+			const enriched = await Promise.all(
+				page.map(async (cred) => {
```
Cadiac: Given our max limit is 200 can 200 concurrent `getAccountContext` calls be too much?

packages/cli/src/oauth/oauth.service.ts
Cadiac: What I was also thinking is do we want to store this derived `accountIdentifier` at all if we can just anyways derive it from the data we have, like we do on `// Fallback for legacy credentials` block at instance ai service anyways for currently saved credentials - was the main concern being able to avoid decrypting the `oauthTokenData` normally with new credentials that have this field, and this was the reason to introduce the new field to be able to avoid decrypting them on new credentials? I'm just trying to understand _why_

aalises: Yes, exactly...  the main motivation is **minimizing the decryption scope at read time**:

- **With the stored field**: redacted decryption is sufficient: `accountIdentifier` is not marked as a password field, so it survives redaction. The full secret payload (tokens, keys) is never materialized in memory 💪 
- **Without it**: every `getAccountContext` call has to do unredacted decryption to scan inside `oauthTokenData`, which exposes the full token payload in memory.

The fallback path in the adapter (the `// Fallback for legacy credentials` block) shows the cost: it needs `decrypt(credential, true)` (unredacted) and a dynamic import of `OauthService` ... which is the legacy path we want to avoid for new credentials.

packages/@n8n/instance-ai/src/tools/credentials/list-credentials.tool.ts
aalises: Have added `p-limit` to reduce the concurrency

Cadiac (dismissed):

Cadiac (approved):

PR #28104, opened 2026-04-07
feat(editor): Update built-in node icons to custom SVGs

Tuukkaa:
## Summary

Replace Font Awesome and file-based SVG icons for built-in nodes with new custom SVG node icons rendered inline via the design system.

**Key changes:**
- Add 86 new SVG node icons to `@n8n/design-system` under `N8nIcon/nodes/`
- Add `nodeIconSet` to the icon registry alongside existing `updatedIconSet`
- Update ~80 node definitions across `nodes-base` and `nodes-langchain` to use `node:` icon references
- Simplify `IconRef` type from `` `node:${string}.${string}` `` to `` `node:${string}` ``
- Add 11 new icon color tokens (`amber`, `teal`, `lavender`, `violet`, `lime`, `magenta`, `sky-blue`, `emerald`, `forest-green`, `rust`, `gray`) with light/dark mode values matching the exact colors from the original SVGs
- Add `'neutral'` to `ThemeIconColor` type
- New node icons render at larger sizes (48px canvas, 36px configuration nodes, 24px panels) while old icons keep existing sizes (40px, 30px, 20px)

**Lint rule note:** The `node-class-description-icon-not-svg` rule from `eslint-plugin-n8n-nodes-base` has been temporarily disabled in both `nodes-base` and `nodes-langchain` eslint configs. The external plugin (v1.16.5) only recognizes `fa:` and `.svg` icon formats — it doesn't know about the new `node:` prefix. Agreed with @ivov to turn it off for now; a follow-up PR will update `eslint-plugin-n8n-nodes-base` to support `node:` icons, after which the rule will be re-enabled.

https://linear.app/n8n/issue/DS-515

## Preview
image
image
image

## Review / Merge checklist

- [ ] Tests included.


packages/frontend/editor-ui/src/features/shared/nodeCreator/components/ItemTypes/NodeItem.vue
```diff
@@ -98,6 +99,12 @@ const nodeActions = computed(() => {
 	return actions[props.nodeType.name] || [];
 });
 
+const nodeListIconSize = computed(() => {
+	const icon = props.nodeType.icon;
+	const iconName = typeof icon === 'string' ? icon : undefined;
```
heymynameisrob, line 104: why wouldn't this always be a string?

packages/frontend/@n8n/design-system/src/components/N8nIcon/icons.ts
heymynameisrob, line 1: Worry about this file for two reasons:
1. We're importing this entire file, which will likely end up in the base bundle. So we'll import a ton of SVGs even if we don't use them.
2. It's huge. Maybe we should split this out into `node-icons.ts` and then maybe we can lazy import it on `Icon.vue` if the icon type is `node:*`

@cstuncsik wdyt?

heymynameisrob: @claude What do you think about the above?

heymynameisrob (commented):
Looking nice 🔥 

A nits and a comment just to clarify if we want to be leaner with our import bundle.

packages/frontend/@n8n/design-system/src/components/N8nIcon/icons.ts
cstuncsik: Absolutely agree
A third option could be going super lazy by letting the icon component load the icon via async import in one of the earliest component lifecycle hooks (maybe `beforeCreate`)

cstuncsik: @claude compare the above solution with the third option (the icon component loads the icon via async import in one of the earliest component lifecycle hooks (maybe beforeCreate)

)

packages/frontend/editor-ui/src/features/shared/nodeCreator/components/ItemTypes/NodeItem.vue
Tuukkaa: The Icon type can be either a string or a { light: string; dark: string } object for nodes that have different icons per theme (basically some rare nodes which don't have a new icon yet)

packages/frontend/@n8n/design-system/src/components/N8nIcon/icons.ts
Tuukkaa: Makes sense, I split the bundle and added lazy loading

packages/frontend/@n8n/design-system/src/components/N8nIcon/icons.ts
```diff
@@ -716,9 +716,19 @@ export const updatedIconSet = {
 
 export type IconName = keyof typeof updatedIconSet; // only new icon names should be used moving forward
 
+export { type NodeIconName } from './node-icons';
+
+const NODE_ICON_PREFIX = 'node:';
+
+export function isNodeIcon(iconName?: string): iconName is IconName {
```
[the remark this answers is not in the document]
Tuukkaa: Good catch — fixed in 771ac3d. `isNodeIcon` now narrows to `NodeIconName` instead of `IconName`.

packages/frontend/@n8n/design-system/src/components/N8nIcon/icons.ts
```diff
@@ -716,9 +716,19 @@ export const updatedIconSet = {
...
+}
+
 export function isSupportedIconName(iconName?: string): iconName is IconName {
-	// support both deprecated and updated icon names
 	return (
-		typeof iconName === 'string' && (iconName in updatedIconSet || iconName in deprecatedIconSet)
+		typeof iconName === 'string' &&
+		(iconName in updatedIconSet ||
+			iconName in deprecatedIconSet ||
+			iconName.startsWith(NODE_ICON_PREFIX))
```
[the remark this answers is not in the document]
Tuukkaa: This is intentional. Since `nodeIconSet` is now lazy-loaded (separate chunk), we can't synchronously check `iconName in nodeIconSet` — the module may not be loaded yet. The `node:*` prefix check is a reasonable trade-off: only the codebase itself constructs these strings from node type definitions, so `node:garbage` can't happen at runtime.

heymynameisrob (approved):
LGTM. You might struggle getting this merged because of the new CI rules of max 100 file changes FYI.

PR #28107, opened 2026-04-07
feat(agent-sdk): tool search process for node-tools

bjorger:
## Summary

This PR implements a tool-chain for the agent to discover nodes, that are available as tools.
The chain goes as follows:
- Discovery `search_nodes`: Provides the agent with a tool to search n8n nodes
- Schema definition for a node `get_node_schema`: As the nodes are quite extensive, we do not want to overload the agent by loading the schemas immediately for all nodes, only for the current one
- Execution `run_node_tool`: Transforms the selected node into a tool for the agent to use
- The toolchain can be enabled / disabled

image

When testing this PR, the following cases should be covered:
- The agent can find ANY node that is available in the workspace
- The agent will inform the user if the credentials are missing for a specific node
- The agent can perform tasks that require multiple different nodes

The agent will default to trial and error to make certain nodes work. This is currently expected and will be resolved by this [ticket](https://linear.app/n8n/issue/AGENT-20/feature-make-agent-learn-from-tool-usage)

To enable the feature add this to your .env file:
N8N_AGENTS_MODULES=node-tools-searcher

https://linear.app/n8n/issue/AI-2333/metatool-on-demand-tool-loading-for-agents

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/agents/agents.service.ts
```diff
@@ -276,12 +284,35 @@ export class AgentsService {
 			});
 		}
 
+		await this.attachNodeToolChain(agent, credentialProvider, agentId, projectId);
```
yehorkardash: I think this should be opt-in option in agent configuration, since not all agents would need this functionality

packages/cli/src/modules/agents/agents-tools.service.ts
```diff
@@ -0,0 +1,210 @@
...
+				}),
+			]),
+		)
+		.min(1)
+		.describe('Node IDs from search_nodes (e.g., ["n8n-nodes-base.gmail"])'),
+});
+
+const listCredentialsInputSchema = z.object({});
+
+const runNodeInputSchema = z.object({
```
yehorkardash, line 66: How viable it is to memoize tools and inject new tools into agent? I'm worried that with this implementation for each `run_node_tool` the agent will need to call `list_credentials`+`get_node_types` which incurs additional latency for simple calls and allows it to make more mistakes in `nodeParameters`.

Ideally, it would need to write `nodeParameters` and `credentials` once, do a test run and then never touch again, unless it needs to change some parameters

yehorkardash (changes requested):

packages/cli/src/modules/agents/agents.service.ts
bjorger: Agreed. Good catch :)

packages/cli/src/modules/agents/agents-tools.service.ts
```diff
@@ -0,0 +1,210 @@
...
+	nodeType: z.string().describe('Node type identifier from search_nodes'),
+	nodeTypeVersion: z.number().describe('Node type version from search_nodes'),
+	nodeParameters: z
```
yehorkardash, line 69: If this is a tool for one run, we could potentially drop inputData, since LLM can just hard code everything for this single too call

bjorger: As discussed outside of scope for P0

packages/cli/src/modules/agents/agents-tools.service.ts
bjorger: As discussed outside of scope for P0

yehorkardash:
@claude review

packages/cli/src/modules/agents/agents-tools.service.ts
```diff
@@ -0,0 +1,206 @@
...
+				);
+				return { results };
+			})
+			.build();
+	}
+
+	private buildListCredentialsTool(credentialProvider: CredentialProvider): BuiltTool {
+		return new Tool('list_credentials')
+			.description(
+				'List the credentials available to the user. Returns an array of credential names and types. ' +
```
yehorkardash, line 162: This description is specific to runtime agent, builder agent doesn't have run_node_tool

packages/cli/src/modules/agents/agents.service.ts
```diff
@@ -354,8 +356,17 @@ export class AgentsService {
 	 * Inject platform-level tools and storage into an agent instance.
 	 * Workflow and node tools are resolved earlier via `makeToolResolver()` inside
 	 * `fromSchema()`, so this method only handles host-side singletons.
+	 *
+	 * `nodeToolsEnabled` comes from the agent's `config.nodeTools.enabled` flag
+	 * (opt-in, defaults to false) — see {@link isNodeToolsEnabled}.
 	 */
-	private async injectRuntimeDependencies(agent: agents.Agent, agentId: string): Promise<void> {
+	private async injectRuntimeDependencies(
```
yehorkardash: Another PR is also going to add a new param [here](https://github.com/n8n-io/n8n/pull/28330/changes#diff-7ed434fd1adebf0f0d1b05b89bbc7f76c058aec3427d1d56b439c502d6228155R361)
Probably worth updating the method to accept params object, rather than a long list of params

packages/cli/src/modules/agents/builder/agents-builder-tools.service.ts
```diff
@@ -226,7 +226,10 @@ export class AgentsBuilderToolsService {
 		return [
 			buildCustomToolTool,
 			listWorkflowsTool,
-			...this.agentsToolsService.getSharedTools(credentialProvider),
+			...this.agentsToolsService.getSharedTools(
+				credentialProvider,
+				'Call this BEFORE generating code to know which .credential() value to use.',
```
yehorkardash: We don't use ts code for agent config, so the prompt should be a bit different, considering that tool node looks like that:
```json
{
      "type": "node",
      "name": "send_gmail",
      "description": "Send an email using Gmail. You can send emails to one or multiple recipients, with optional CC and BCC, and include a subject and message body in HTML format.",
      "node": {
        "nodeType": "n8n-nodes-base.gmail",
        "nodeTypeVersion": 2.2,
        "nodeParameters": {
          "resource": "message",
          ...
          }
        },
        "credentials": {
          "gmailOAuth2": {
            "id": "EsI18W8yyzsTb90c",
            "name": "Gmail account"
          }
        }
      },
      "inputSchema": {
        "type": "object",
        "properties": {
           .....
        }
      }
    }
```

yehorkardash (approved):

PR #28108, opened 2026-04-07
fix(ai-builder): Paginate list-credentials tool and drop unused fields

aalises:
## Summary
- Adds `limit`/`offset` pagination to the `list-credentials` tool (default 50, max 200) with a `total` count, so the AI can page through large credential sets instead of loading everything into the prompt
- Drops `createdAt` from `CredentialSummary` (no consumer uses it) and strips `updatedAt` from the tool output (kept on the interface for internal sorting in `setup-workflow.service.ts`)
- Only exposes `id`, `name`, `type` to the AI prompt — reducing per-credential token cost

Fixes the "prompt too long" error for admins with many credentials.

## Review / Merge checklist
- [x] Tests pass (8 new tests for pagination + 61 related tests green)
- [x] `CredentialDetail` still carries `createdAt`/`updatedAt` for `get-credential` single lookups
- [x] `setup-workflow.service.ts` sorting by `updatedAt` is unaffected (field stays on `CredentialSummary`)

## Test plan
- [ ] As an admin with 50+ credentials, open Instance AI and ask it to list credentials — verify it returns a paginated result with `total`
- [ ] Verify the AI can paginate through all credentials using `offset`
- [ ] Verify filtering by `type` still works
- [ ] Verify no timestamps appear in the AI's credential list output

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/credentials/__tests__/list-credentials.tool.test.ts
+++ b/packages/@n8n/instance-ai/src/tools/credentials/__tests__/list-credentials.tool.test.ts
@@ -26,7 +26,6 @@ function makeCredential(overrides?: Partial<CredentialSummary>): CredentialSumma
 		id: 'cred-1',
 		name: 'Gmail OAuth',
 		type: 'gmailOAuth2',
-		updatedAt: '2025-06-15T12:00:00.000Z',
 		...overrides,
 	};
 }
@@ -69,19 +68,6 @@ describe('list-credentials tool', () => {
 				{ id: 'cred-1', name: 'Gmail OAuth', type: 'gmailOAuth2' },
 			]);
 			expect(result.total).toBe(1);
-		});
-
-		it('does not include updatedAt in output', async () => {
-			const context = createMockContext();
-			(context.credentialService.list as jest.Mock).mockResolvedValue([makeCredential()]);
-
-			const tool = createListCredentialsTool(context);
-			const result = (await tool.execute!({}, {} as never)) as {
-				credentials: Array<Record<string, unknown>>;
-			};
-
-			expect(result.credentials[0]).not.toHaveProperty('updatedAt');
-			expect(result.credentials[0]).not.toHaveProperty('createdAt');
 		});
 
 		it('passes type filter to the list call', async () => {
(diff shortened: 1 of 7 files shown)


packages/@n8n/instance-ai/src/types.ts
```diff
@@ -88,12 +88,12 @@ export interface CredentialSummary {
 	id: string;
 	name: string;
 	type: string;
-	createdAt: string;
 	updatedAt: string;
```
OlegIvaniv: `updatedAt` was removed too

packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
```diff
@@ -824,7 +824,6 @@ export class InstanceAiAdapterService {
 						id: c.id,
 						name: c.name,
 						type: c.type,
-						createdAt: c.createdAt.toISOString(),
 						updatedAt: c.updatedAt.toISOString(),
```
OlegIvaniv: `updatedAt` was removed too

packages/@n8n/instance-ai/src/tools/credentials/__tests__/list-credentials.tool.test.ts
```diff
@@ -0,0 +1,133 @@
...
+		dataTableService: {} as InstanceAiContext['dataTableService'],
+	};
+}
+
+function makeCredential(overrides?: Partial<CredentialSummary>): CredentialSummary {
+	return {
+		id: 'cred-1',
+		name: 'Gmail OAuth',
+		type: 'gmailOAuth2',
+		updatedAt: '2025-06-15T12:00:00.000Z',
```
OlegIvaniv: `updatedAt` was removed

OlegIvaniv (changes requested):

packages/@n8n/instance-ai/src/types.ts
aalises: true, forgot to push it 🙈 , fixed!

OlegIvaniv (dismissed):

OlegIvaniv (approved):

PR #28110, opened 2026-04-07
fix(core): Fix retry activation in multi-main bypassing exponential backoff

ivov:
## Summary

On multi-main, when a workflow fails to activate on startup, we are supposed to keep retrying activation with exponential backoff. Instead, on retry we currently call `this.add()` _without_ passing `{ shouldPublish: false }`, so `shouldPublish` defaults to `true`, which causes the leader to publish a pubsub message to itself and return successfully without actually activating the workflow. The retry loop sees no error and stops, thinking it succeeded. If the leader then fails to activate the workflow when handling its own pubsub message, we permanently deactivate with no further retries, bypassing the exponential backoff mechanism.

This PR fixes the above and also clears queued activation timers on leader stepdown, preventing a former leader from firing stale retries.

## Testing

Workflow activation is ancient and in dire need of refactoring. Mocking `setTimeout`, `publisher`, `instanceSettings.isMultiMain`, etc. just to assert `add` was called with the right 4th arg would be restating the implementation in test form, which doesn't really prove anything. We could find some way to orchestrate this in e2e, but I think it's pragmatic to go ahead as-is until we streamline workflow publication. Please let me know if you disagree.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/active-workflow-manager.ts
+++ b/packages/cli/src/active-workflow-manager.ts
@@ -571,6 +571,7 @@ export class ActiveWorkflowManager {
 	@OnLeaderStepdown()
 	@OnShutdown()
 	async removeAllTriggerAndPollerBasedWorkflows() {
+		this.removeAllQueuedWorkflowActivations();
 		await this.activeWorkflows.removeAllTriggerAndPollerBasedWorkflows();
 	}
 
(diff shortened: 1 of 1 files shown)


despairblue (approved):
following this flow is mind bending

PR #28117, opened 2026-04-07
fix(core): Avoid permanent deactivation on transient isolate errors

ivov:
## Summary

Adds an `IsolateError` base class for expression isolate infrastructure errors. On workflow activation, these errors now queue the workflow for retry instead of permanently deactivating it, and do not increment the exponential backoff counter.

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/active-workflow-manager.ts
```diff
@@ -842,7 +855,7 @@ export class ActiveWorkflowManager {
 			} catch (error) {
 				this.errorReporter.error(error);
 				let lastTimeout = this.queuedActivations[workflowId].lastTimeout;
-				if (lastTimeout < WORKFLOW_REACTIVATE_MAX_TIMEOUT) {
+				if ((!error) instanceof IsolateError && lastTimeout < WORKFLOW_REACTIVATE_MAX_TIMEOUT) {
```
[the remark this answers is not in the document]
ivov: Wow, this was Biome reformatting during commit. Good catch!

packages/@n8n/expression-runtime/src/evaluator/expression-evaluator.ts
```diff
@@ -102,13 +103,13 @@ export class ExpressionEvaluator implements IExpressionEvaluator {
 	private getBridge(caller: object): RuntimeBridge {
 		const bridge = this.bridgesByCaller.get(caller);
 		if (!bridge) {
-			throw new Error('No bridge acquired for this context. Call acquire() first.');
+			throw new IsolateError('No bridge acquired for this context. Call acquire() first.');
```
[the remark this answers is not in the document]
ivov: In theory yes, but the whole reason we're here is that this exact error surfaced during workflow activation in production. The safe default is to retry.

packages/@n8n/expression-runtime/src/evaluator/expression-evaluator.ts
```diff
@@ -74,7 +75,7 @@ export class ExpressionEvaluator implements IExpressionEvaluator {
 		caller: object,
 		options?: EvaluateOptions,
 	): unknown {
-		if (this.disposed) throw new Error('Evaluator disposed');
+		if (this.disposed) throw new IsolateError('Evaluator disposed');
```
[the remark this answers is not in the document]
ivov: The evaluator is disposed during shutdown or leadership stepdown. If a workflow activation is in-flight when that happens, we hit this error, so retrying is correct. Permanently deactivating the workflow because the evaluator was disposed mid-activation is the worse outcome.

packages/nodes-base/nodes/N8n/n8n-api-coverage.json
```diff
@@ -190,6 +190,9 @@
 		"DELETE /data-tables/{dataTableId}/rows/delete": {
 			"status": "gap"
 		},
+		"GET /insights/summary": {
+			"status": "gap"
+		},
```
despairblue, line 195: What's this?

packages/@n8n/expression-runtime/src/pool/isolate-pool.ts
despairblue: Should this also be an IsolateError? 
If not can we add a comment to make sure nobody in the future turns this into an IsolateError?

despairblue (dismissed):
Two minor things, but approved generally.

packages/@n8n/expression-runtime/src/pool/isolate-pool.ts
ivov: It should, it's an isolate infra failure, updated!

packages/nodes-base/nodes/N8n/n8n-api-coverage.json
ivov: CI fix, discussed internally

despairblue (approved):

PR #28123, opened 2026-04-07
feat(editor): Improve Save button feedback in Workflow Settings modal

paulomation:
## Summary

- Save button starts in `outline` style (muted, transparent background) when no changes have been made
- Transitions to `solid` style (orange CTA) when any setting is modified
- Button returns to `outline` style after a successful save (modal closes)
- Button aligned to the right of the modal footer, consistent with other modals
- Relies on default transition styles (no custom CSS overrides)

Dirty state is tracked by snapshotting `workflowSettings` and `timeoutHMS` on modal open and comparing on every reactive change.

Closes [DS-516](https://linear.app/n8n/issue/DS-516)

## Test plan

- [ ] Open the Workflow Settings modal → Save button should appear as a subtle outlined button on the right
- [ ] Change any setting (e.g. Error Workflow, Timezone, Execution Order) → Save button should transition to the solid orange CTA
- [ ] Save → modal closes, re-open → Save button is outline again (no changes)
- [ ] Revert a change back to the original value → Save button should return to outline style

| | **No Change** | **With Change** |
|---|---|---|
| **Before** | Before - no change | Before - with change |
| **After** | After - no change | After - with change |

## Review / Merge checklist

- [ ] ~Docs updated or follow-up ticket created.~
- [x] Tests included.


RobTF9:
@claude review this PR for me

packages/frontend/editor-ui/src/app/components/WorkflowSettings.vue
```diff
@@ -1524,6 +1509,13 @@ onBeforeUnmount(() => {
 	}
 }
 
+.save-button {
```
RobTF9: I'd remove these overrides, just rely on any default styles for transitions like this.

RobTF9 (changes requested):
One comment about overrides, also I'd move the button the right hand side if possible. So that's consistent with other modals and overlaps

Image

PR #28125, opened 2026-04-07
fix(ai-builder): Improve post-build flow: setup, test, then publish

aalises:
## Summary
- Restructures the post-build flow in the system prompt as clear numbered steps: build → setup (if needed) → test → publish (only when user asks)
- `setup-workflow` is only called when there's actually something to configure (mocked credentials, missing parameters, unconfigured triggers)
- Removes auto-publish from the builder agent prompt — the builder's job ends at a successful submit
- Publishing is always the user's explicit decision after they've had a chance to test

## Related
https://www.notion.so/n8n/InstanceAI-asks-me-to-publish-before-i-get-a-chance-to-test-and-there-is-no-other-option-33b5b6e0c94f80c7b3aed05cdb5d226f
https://www.notion.so/n8n/Instance-AI-issues-reported-3285b6e0c94f802ca27cd2d2e067c5d7?p=33c5b6e0c94f8008abaac1b534b1b54a&pm=s


Cadiac (approved):
Nice, and thanks for addressing the test, I've had a few PRs fail to that now this afternoon.

We should probably (later) teach the agent to use the testing mode for webhooks etc, now it can't really test webhooks without publishing them

PR #28126, opened 2026-04-07
feat(core): Emit audit events for workflow activation on bootup

ivov:
## Summary

Write a dedicated audit event on workflow activation and deactivation on bootup. Intentionally not emitting a general workflow activation event as this triggers downstream services like the workflow indexer.

Context: https://www.notion.so/n8n/33b5b6e0c94f8010a4feecc6ad87914a

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/active-workflow-manager.ts
```diff
@@ -777,8 +788,21 @@ export class ActiveWorkflowManager {
...
 
+			if (dbWorkflow && (activationMode === 'init' || activationMode === 'leadershipChange')) {
+				void this.eventBus.sendAuditEvent({
+					eventName: 'n8n.audit.workflow.deactivated-on-bootup',
+					payload: {
+						workflowId,
+						workflowName: dbWorkflow.name,
+						deactivatedVersionId: dbWorkflow.activeVersionId ?? null,
+					},
+				});
```
despairblue, line 820: It would be nice to have a reason in the payload. Maybe the `e.name`?
And also the `activationMode`.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -525,6 +527,15 @@ export class ActiveWorkflowManager {
 					workflowName: dbWorkflow.name,
 					workflowId: dbWorkflow.id,
 				});
+
+				void this.eventBus.sendAuditEvent({
+					eventName: 'n8n.audit.workflow.activated-on-bootup',
```
despairblue:
```suggestion
					eventName: 'n8n.audit.workflow.activated',
```

Technically it's `on-init` or `on-leadership-change`, so maybe we want to drop the suffix and add the activation mode?

despairblue (approved):

PR #28135, opened 2026-04-07
feat(core): Add in-process mutex for SQLite advisory lock parity

afitzek:
## Summary

`DbLockService` previously relied solely on database transactions for SQLite serialization, which is insufficient — concurrent async callers within the same process can interleave inside the critical section even when each runs its own transaction. This PR adds an in-process async mutex for SQLite that provides the same serialization guarantees as Postgres advisory locks (`pg_advisory_xact_lock` / `pg_try_advisory_xact_lock`).

**What changed:**
- `withLock` and `tryWithLock` now acquire an in-process mutex before entering the transaction on SQLite, while Postgres continues to use advisory locks as before
- The mutex uses an ownership-token model: each acquisition creates an opaque token, and the release function only acts if its token still matches the current holder — making double-release a safe no-op
- Ownership transfers atomically on release (new token assigned to `held` before the waiter's promise resolves), eliminating a microtask-width window where `tryAcquireLock` could bypass the queue
- The Postgres code path is unchanged but slightly simplified by removing the now-unnecessary `if (type === 'postgresdb')` nesting inside the transaction callback

**Key implementation decisions:**
- The mutex is per-lock-ID (not global) so different lock IDs don't block each other
- FIFO ordering is maintained via a queue array
- Timeout support mirrors the Postgres `SET LOCAL lock_timeout` behavior with `setTimeout` + queue cleanup on expiry

## Review checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/db/src/services/__tests__/db-lock.service.test.ts
+++ b/packages/@n8n/db/src/services/__tests__/db-lock.service.test.ts
@@ -149,4 +149,221 @@ describe('DbLockService', () => {
 			expect(fn).toHaveBeenCalledWith(mockTx);
 		});
 	});
+
+	describe('SQLite in-process mutex', () => {
+		beforeEach(() => {
+			databaseConfig.type = 'sqlite';
+		});
+
+		afterEach(() => {
+			jest.useRealTimers();
+		});
+
+		it('should serialize concurrent withLock calls for the same lockId', async () => {
+			const executionOrder: string[] = [];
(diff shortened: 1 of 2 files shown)


phyllis-noester (approved):
The PR looks sound to me but I don't have a deep understanding of our db lock structure or how to easily test it without going down a rabbit role :) is it easy for you to test this? If so, could you record a video showing the change in action?

PR #28138, opened 2026-04-07
feat(ai-builder): Add binary check to detect changes in existing workflow (no-changelog)

lucamattiazzi:
## Review / Merge checklist

- [x] Tests included. 


burivuhster (commented):
My concern is that there are cases when workflow should not be changed, e.g. when users ask a question about something. This check will produce false positives in these scenarios.

So this requires either LLM-based check (we've introduced one covering this case recently: `response-matches-workflow-changes.ts`) OR some annotation of dataset examples, indicating whether the workflow changes are required in each case.

lucamattiazzi:
> My concern is that there are cases when workflow should not be changed, e.g. when users ask a question about something. This check will produce false positives in these scenarios.
> 
> So this requires either LLM-based check (we've introduced one covering this case recently: `response-matches-workflow-changes.ts`) OR some annotation of dataset examples, indicating whether the workflow changes are required in each case.

@burivuhster that makes sense, I actually wrote a version that adds an optional gate in front of the binary checks, so that this check can validate (using an llm) if it needs to run at all. the problem is that this would make this check not deterministic anymore.
the alternative I thought of is adding to the llm binary checks another function that works like `skipIf`, that runs if the llm decides that the check needs to run at all.
if neither seems ok, I can try to find another solution for this as well

lucamattiazzi:
> My concern is that there are cases when workflow should not be changed, e.g. when users ask a question about something. This check will produce false positives in these scenarios.
> 
> So this requires either LLM-based check (we've introduced one covering this case recently: `response-matches-workflow-changes.ts`) OR some annotation of dataset examples, indicating whether the workflow changes are required in each case.

I've implemented the script that adds the tags in the annotations of the dataset, and fixed the check to look for that specific tag before running

packages/@n8n/ai-workflow-builder.ee/evaluations/evaluators/binary-checks/checks/workflow-actually-changed.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const workflowActuallyChanged: BinaryCheck = {
+	name: 'workflow_actually_changed',
+	kind: 'deterministic',
+	async run(workflow: SimpleWorkflow, ctx: BinaryCheckContext) {
+		if (!ctx.existingWorkflow) {
+			return { pass: true, comment: 'There is no existing workflow, skipped' };
+		}
+		const tags: string[] = (ctx.annotations?.tags as []) ?? [];
+		if (!tags.includes(GATE_TAG_NAME)) {
```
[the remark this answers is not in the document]
lucamattiazzi: this is expected, to avoid false positives

packages/@n8n/ai-workflow-builder.ee/scripts/annotate-dataset-tags.ts
```diff
@@ -0,0 +1,246 @@
...
+	let completed = 0;
+	let errors = 0;
+	const startTime = Date.now();
+
+	const tasks = examples.map(
+		async (example) =>
+			await limit(async () => {
+				const index = ++completed;
+				try {
+					const inputs = example.inputs as Record<string, unknown>;
```
schrothbn, line 178: Probably worth checking how the current traces end up being annotated (open_code_note and stuff) and adjust to that? That's what the human annotators would use and they could prefill that for new traces as the instance-ai traces are shaped differently.

schrothbn (commented):
Looks good when CI is fixed (will approve then), 
A note about looking into the schema of the manual annotations, I think they're on a feedback key. 
That would allow to annotate new traces while they anyway have to look at them to look for signals as well and use the LLM version when we automate this completely.

packages/@n8n/ai-workflow-builder.ee/scripts/annotate-dataset-tags.ts
lucamattiazzi: oh ok! I tried to keep a similar structure to the one used by `no_unnecessary_code_nodes`, but I'll change that to align to this one instead!

lucamattiazzi:
> Looks good when CI is fixed (will approve then), A note about looking into the schema of the manual annotations, I think they're on a feedback key. That would allow to annotate new traces while they anyway have to look at them to look for signals as well and use the LLM version when we automate this completely.

the checks are working now (a typecheck in the script was failing)

PR #28140, opened 2026-04-07
refactor(ai-builder): Consolidate native tools into 10 action families (no-changelog)

aalises:
## Summary

Consolidates 60+ individual AI tools into **10 family tools** using a discriminated union on the `action` field. Instead of each operation living in its own tool (`list-workflows`, `get-workflow`, `delete-workflow`, …), related operations are grouped into a single tool that routes by action.

**Why:** The LLM orchestrator was choosing from ~45 tools per turn. Fewer, well-scoped tools improve selection accuracy and cut schema token overhead.

### Tool families

| Family tool | Example actions | Notes |
|---|---|---|
| `workflows` | `list`, `get`, `setup`, `publish`, `restore-version`, … | Builder gets all actions; orchestrator excludes `get-as-code` |
| `executions` | `list`, `get`, `run`, `debug`, `stop` | — |
| `credentials` | `list`, `get`, `delete`, `setup`, `test` | — |
| `data-tables` | `list`, `schema`, `query`, `create`, `insert-rows`, … | Orchestrator gets read-only subset |
| `workspace` | `list-projects`, `list-tags`, `create-folder`, … | — |
| `research` | `web-search`, `fetch-url` | Replaces standalone web-research tools |
| `nodes` | `list`, `search`, `describe`, `type-definition`, … | Orchestrator gets `explore-resources` only |
| `templates` | `search-structures`, `search-parameters`, `best-practices` | All agents (orchestrator + builder) |
| `task-control` | `update-checklist`, `cancel-task`, `correct-task` | Orchestrator only |

### What changed

- **Backend** — 10 new family tool files replace 57 individual tool files. Each uses `z.discriminatedUnion` for input validation. Surface-aware factories (`createAllTools`, `createOrchestratorDomainTools`) control which actions are available per agent role. Orchestration tool prompts updated to reference the new names.
- **Frontend** — Icon mapping, tool labels, result rendering, canvas preview, resource extraction, render hints, and i18n all updated to resolve family + action pairs.
- **Tests** — 198 new tests covering all 10 family tools; old per-tool test files removed.
- **Dead code cleanup** — Removed `filesystem.tool.ts`, its tests, and the `InstanceAiFilesystemService` interface + associated types (`FileEntry`, `FileContent`, `FileSearchMatch`, `FileSearchResult`). These became unreachable after `LocalFilesystemProvider` was removed — the only live filesystem path now goes through the gateway MCP server via `createToolsFromLocalMcpServer()`. Also removed the `filesystemService` field from `InstanceAiContext` and the fallback branches in both `createAllTools` and `createOrchestratorDomainTools`.
- **Bug fix** — Added `templates` tool to `createOrchestratorDomainTools()`. It was only given to sub-agents/builders, preventing the orchestrator from looking up workflow patterns and best practices during planning.

Net: **~1500 lines removed** across 91 files.

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/instance-ai/src/agent/instance-agent.ts
```diff
@@ -121,7 +119,7 @@ export async function createInstanceAgent(options: CreateInstanceAgentOptions):
 	// Build native n8n domain tools (context captured via closures — per-run)
 	const domainTools = createAllTools(context);
 
-	const orchestratorDomainTools = getOrchestratorDomainTools(domainTools);
+	const orchestratorDomainTools = createOrchestratorDomainTools(context);
```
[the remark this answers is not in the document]
aalises: Fixed — browser MCP tools are properly isolated. Line 172-176 filters them out from the orchestrator's tool set using `!browserToolNames.has(name)`.

packages/@n8n/instance-ai/src/tools/templates.tool.ts
```diff
@@ -0,0 +1,189 @@
...
+
+	// Specific technique lookup
+	const getDocFn = documentation[technique as WorkflowTechniqueType];
+	if (!getDocFn) {
+		// Check if it's a valid technique without docs
+		const description = TechniqueDescription[technique as WorkflowTechniqueType];
+		if (description) {
+			return {
+				technique,
+				message: `Technique "${technique}" (${description}) exists but does not have detailed documentation yet. Use search-template-structures to find example workflows instead.`,
```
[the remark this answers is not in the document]
aalises: Already correct — the message references `search-structures` which matches the action name.

packages/@n8n/instance-ai/src/tools/data-tables.tool.ts
```diff
@@ -0,0 +1,568 @@
...
+
+	return result;
+}
+
+async function handleCreate(
+	context: InstanceAiContext,
+	input: Extract<FullInput, { action: 'create' }>,
+	ctx: { agent?: { resumeData?: unknown; suspend?: unknown } },
+) {
+	const resumeData = ctx?.agent?.resumeData as ResumeData | undefined;
```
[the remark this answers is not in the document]
aalises: Fixed — `handleDelete` now checks `context.permissions?.deleteDataTable === 'blocked'` at line 257-259.

packages/@n8n/instance-ai/src/tools/data-tables.tool.ts
```diff
@@ -0,0 +1,568 @@
...
+async function handleDelete(
+	context: InstanceAiContext,
+	input: Extract<FullInput, { action: 'delete' }>,
+	ctx: { agent?: { resumeData?: unknown; suspend?: unknown } },
+) {
+	const resumeData = ctx?.agent?.resumeData as ResumeData | undefined;
+	const suspend = ctx?.agent?.suspend as ((payload: unknown) => Promise<void>) | undefined;
+
+	// State 1: First call — suspend for confirmation
+	if (resumeData === undefined || resumeData === null) {
```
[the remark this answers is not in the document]
aalises: Fixed — `handleDelete` checks `needsApproval` at line 261 and only suspends when approval is required, consistent with other handlers.

packages/@n8n/instance-ai/src/tools/filesystem.tool.ts
```diff
@@ -0,0 +1,313 @@
...
+		type: typeFilter,
+		recursive: isRecursive,
+	});
+	const truncated = fetched.length > limit;
+	const files = truncated ? fetched.slice(0, limit) : fetched;
+
+	return {
+		files,
+		truncated,
+		totalCount: files.length,
```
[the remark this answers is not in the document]
aalises: Fixed — the field was renamed to `returnedCount` which accurately reflects what it represents.

packages/@n8n/instance-ai/src/tools/orchestration/browser-credential-setup.tool.ts
```diff
@@ -306,12 +305,9 @@ export function createBrowserCredentialSetupTool(context: OrchestrationContext)
...
 			browserTools['ask-user'] = createAskUserTool();
 
-			// Add research tools (fetch-url, web-search) from the domain context
+			// Add consolidated research tool (web-search + fetch-url) from the domain context
 			if (context.domainContext) {
-				browserTools['fetch-url'] = createFetchUrlTool(context.domainContext);
-				if (context.domainContext.webResearchService?.search) {
-					browserTools['web-search'] = createWebSearchTool(context.domainContext);
-				}
+				browserTools.research = createResearchTool(context.domainContext);
```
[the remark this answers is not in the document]
aalises: False positive — `fetch-url` and `web-search` are the correct action names in the consolidated `research` tool (see `research.tool.ts` Zod schema at lines 19 and 38). The prompt references match the actual tool interface.

packages/@n8n/instance-ai/src/tools/workspace.tool.ts
```diff
@@ -0,0 +1,411 @@
...
+function buildInputSchema(context: InstanceAiContext) {
+	const baseActions = [
+		listProjectsAction,
+		listTagsAction,
+		tagWorkflowAction,
+		cleanupTestExecutionsAction,
+	] as const;
+
+	if (context.workspaceService?.listFolders) {
+		return sanitizeInputSchema(
```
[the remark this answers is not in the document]
aalises: False positive in context — `sanitizeInputSchema` is required because Anthropic rejects tool schemas without top-level `type: "object"`, and `z.discriminatedUnion` produces `oneOf` without it. The flattening makes all non-discriminator fields optional at the Zod level, but:

1. The LLM is guided by the tool **description** which lists required fields per action
2. Anthropic's API receives the JSON Schema (with per-action field descriptions preserved)
3. In practice, the LLM always sends valid inputs — invalid payloads from the model are not a real-world scenario

This is an intentional tradeoff to fix the `input_schema.type: Field required` API error.

packages/@n8n/instance-ai/src/tools/research.tool.ts
```diff
@@ -0,0 +1,202 @@
...
+		.number()
+		.int()
+		.positive()
+		.max(100_000)
+		.default(30_000)
+		.optional()
+		.describe('Maximum content length in characters (default 30000)'),
+});
+
+const inputSchema = sanitizeInputSchema(
```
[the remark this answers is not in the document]
aalises: False positive in context — `sanitizeInputSchema` is required because Anthropic rejects tool schemas without top-level `type: "object"`, and `z.discriminatedUnion` produces `oneOf` without it. The flattening makes all non-discriminator fields optional at the Zod level, but:

1. The LLM is guided by the tool **description** which lists required fields per action
2. Anthropic's API receives the JSON Schema (with per-action field descriptions preserved)
3. In practice, the LLM always sends valid inputs — invalid payloads from the model are not a real-world scenario

This is an intentional tradeoff to fix the `input_schema.type: Field required` API error.

packages/@n8n/instance-ai/src/tools/task-control.tool.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const correctTaskAction = z.object({
+	action: z.literal('correct-task').describe('Send a correction to a running background task'),
+	taskId: z.string().describe('The task ID to send the correction to (e.g. build-XXXXXXXX)'),
+	correction: z
+		.string()
+		.describe("The correction message from the user (e.g. 'use the Projects database')"),
+});
+
+const inputSchema = sanitizeInputSchema(
```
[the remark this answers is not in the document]
aalises: False positive in context — `sanitizeInputSchema` is required because Anthropic rejects tool schemas without top-level `type: "object"`, and `z.discriminatedUnion` produces `oneOf` without it. The flattening makes all non-discriminator fields optional at the Zod level, but:

1. The LLM is guided by the tool **description** which lists required fields per action
2. Anthropic's API receives the JSON Schema (with per-action field descriptions preserved)
3. In practice, the LLM always sends valid inputs — invalid payloads from the model are not a real-world scenario

This is an intentional tradeoff to fix the `input_schema.type: Field required` API error.

packages/@n8n/instance-ai/src/tools/executions.tool.ts
```diff
@@ -0,0 +1,233 @@
...
+		.optional()
+		.describe('Maximum number of items to return (default 10, max 50)'),
+});
+
+const stopAction = z.object({
+	action: z.literal('stop').describe('Cancel a running workflow execution'),
+	executionId: z.string().describe('ID of the execution to cancel'),
+});
+
+const inputSchema = sanitizeInputSchema(
```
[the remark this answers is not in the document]
aalises: False positive in context — `sanitizeInputSchema` is required because Anthropic rejects tool schemas without top-level `type: "object"`, and `z.discriminatedUnion` produces `oneOf` without it. The flattening makes all non-discriminator fields optional at the Zod level, but:

1. The LLM is guided by the tool **description** which lists required fields per action
2. Anthropic's API receives the JSON Schema (with per-action field descriptions preserved)
3. In practice, the LLM always sends valid inputs — invalid payloads from the model are not a real-world scenario

This is an intentional tradeoff to fix the `input_schema.type: Field required` API error.

mutdmour:
**Question: Parameter disambiguation after discriminated union flattening**

`sanitizeInputSchema` flattens each `z.discriminatedUnion` into a single flat `z.object` where all variant-specific fields become optional. This is necessary (Anthropic requires `type: "object"` at root), but the model loses the structural signal of which fields belong to which action.

For small tools like `research` (2 actions, completely distinct fields) this is fine. For larger families, there are some ambiguity risks worth considering:

**`workflows` (11 actions)** — `name` and `description` appear in both `publish` and `update-version`. `workflowId` appears in 9 actions, `versionId` in 4, `limit` in 2 (`list` and `list-versions`). After flattening, the model sees ~20 optional fields with no schema-level signal about which ones pair with which action.

**`data-tables` (11 actions)** — `name` means table name in `create` but column name in `add-column`. After flattening, it's one optional `name` field.

**`nodes` (6 actions)** — `explore-resources` contributes 7 fields (`credentialType`, `credentialId`, `methodName`, etc.) that all become top-level optionals. The model could plausibly attach `credentialId` to the `describe` action.

The failure mode is silent: extra/irrelevant fields are ignored (no error, no feedback to the model), and missing fields only fail downstream at the service layer with an opaque error that doesn't mention the schema mismatch. The model has no corrective signal.

Would it be worth enriching parameter `.describe()` strings with their action scope for the higher-risk tools? For example:

```ts
// Current
name: z.string().optional().describe('Name for this published version')

// With action scope
name: z.string().optional().describe('[publish, update-version] Name for this version')
```

Or listing parameters in the tool description bullets:

```
• publish — publish a workflow (accepts: workflowId, workflowName, versionId, name, description)
• list — list workflows (accepts: query, limit)
```

Not sure if this is causing real issues in practice — more of a defensive measure given the silent failure mode. Curious if you've seen the model misroute fields, or if the current descriptions are sufficient.

packages/@n8n/instance-ai/src/tools/credentials.tool.ts
```diff
@@ -0,0 +1,345 @@
...
+});
+
+const getAction = z.object({
+	action: z.literal('get').describe('Get credential metadata by ID'),
+	credentialId: z.string().describe('ID of the credential'),
+});
+
+const deleteAction = z.object({
+	action: z.literal('delete').describe('Permanently delete a credential by ID'),
+	credentialId: z.string().describe('ID of the credential to delete'),
```
mutdmour: Here is a good example where agent might be confused. 
The description is `ID of the credential to delete`
Below  for delete `ID of the credential to test`
And for getAction is `ID of the credential`
Depending on what comes into the santize function first, that will be the description. Which might be confusing for agent if it wants to call test, but description says "to delete"?

Maybe we can structure this differently to avoid this problem? How do we structure this so we intentionally set the descriptions? 

Maybe we have two Zod schemas? A global one that is a combo of all actions that we pass into mastra? And a per-action one that we use for validation? 
We should structure this so future agents don't fall into this trap

packages/@n8n/instance-ai/src/tools/workspace.tool.ts
```diff
@@ -0,0 +1,411 @@
...
+
+const listTagsAction = z.object({
+	action: z.literal('list-tags').describe('List all available tags'),
+});
+
+const tagWorkflowAction = z.object({
+	action: z
+		.literal('tag-workflow')
+		.describe('Assign tags to a workflow, creating missing tags automatically'),
+	workflowId: z.string().describe('ID of the workflow to tag'),
```
mutdmour: here same issue. `to tag` and `to clean up` below

packages/@n8n/instance-ai/src/tools/credentials.tool.ts
```diff
@@ -0,0 +1,345 @@
...
+	action: z.literal('search-types').describe('Search available credential types by keyword'),
+	query: z
+		.string()
+		.describe('Search keyword — typically the service name (e.g. "linear", "notion", "slack")'),
+});
+
+const setupAction = z.object({
+	action: z
+		.literal('setup')
+		.describe('Open the credential setup UI for the user to create or select credentials'),
```
mutdmour, line 78: Description of the action is not actually relevant to the LLM here. This is also maybe why we should restructure these.

What the LLM might see then
{
  "action": { "enum": ["list", "search-types", "setup"], "description": "Open the credential setup UI for the user to create or select credentials'" },
}

Which would be confusing

packages/@n8n/instance-ai/src/tools/credentials.tool.ts
```diff
@@ -0,0 +1,345 @@
...
+	return createTool({
+		id: 'credentials',
+		description:
+			'Manage credentials. Actions:\n' +
+			'• list — list credentials accessible to the current user (paginated, filterable by type)\n' +
+			'• get — get credential metadata by ID (never returns secrets)\n' +
+			'• delete — permanently delete a credential (irreversible)\n' +
+			'• search-types — search available credential types by keyword\n' +
+			'• setup — open the credential setup UI for the user to select or create credentials\n' +
+			'• test — test whether a credential is valid and can connect to its service',
```
mutdmour: maybe these could go into the description of the action parameter, automatically generated?

packages/@n8n/instance-ai/src/tools/credentials.tool.ts
mutdmour: Otherwise we just throw an error if there's a conflict at build time? 🤔

packages/@n8n/instance-ai/src/tools/nodes.tool.ts
```diff
@@ -17,15 +17,15 @@ const listAction = z.object({
...
 
 const searchAction = z.object({
 	action: z.literal('search').describe('Search node types by name or AI connection type'),
 	query: z
 		.string()
 		.optional()
-		.describe('Search query to match against node names, display names, aliases, and descriptions'),
+		.describe('Search query to filter by name or description (e.g. "slack", "http")'),
 	connectionType: z
 		.enum(AI_CONNECTION_TYPES)
```
mutdmour, line 30: enums are an interesting one. How do we handle conflicts here? Does not seem to be an issue today. But maybe we throw an error if there's a conflict to warn in case of future changes?

mutdmour (commented):
Amazing PR. Great to see so much code deleted. Wish we had evals to track how much this helps.

packages/@n8n/instance-ai/src/agent/sanitize-mcp-schemas.ts
```diff
@@ -30,30 +30,72 @@ export function sanitizeZodType(schema: z.ZodTypeAny): z.ZodTypeAny {
...
+			const sanitizedField = sanitizeZodType(entries[0].type).optional();
+
+			const withDesc = entries.filter(
+				(e): e is typeof e & { description: string } => !!e.description,
+			);
+			const uniqueDescs = new Set(withDesc.map((d) => d.description));
+
+			if (uniqueDescs.size > 1) {
+				// Descriptions conflict across actions — combine with action context
+				const combined = withDesc.map((d) => `For "${d.action}": ${d.description}`).join('. ');
```
mutdmour, line 142: My recommendation is we design this so future coding agents can get this right out of the gate next time. 

The next time a coding agent adds a new family, we will likely end up with 
{ properties: 
   "workflowId": { type: 'string', description: 'For "add": Workflow ID to add, for "delete": Workflow ID to delete }
}
Just like we had before the last few commits to harmonize them.

As such the simplest approach would be to throw a build (or type) error, prevent this from happening and give feedback to the agent.

packages/@n8n/instance-ai/src/agent/__tests__/sanitize-mcp-schemas.test.ts
```diff
@@ -270,38 +270,56 @@ describe('sanitizeMcpToolSchemas', () => {
...
+				z.object({
+					action: z.literal('get').describe('Get item by ID'),
+					id: z.string(),
+				}),
+			]);
+
+			const result = sanitizeZodType(union) as z.ZodObject<z.ZodRawShape>;
+			const actionField = result.shape.action;
+
+			expect(actionField.description).toBe('"list": List all items | "get": Get item by ID');
```
mutdmour, line 383: now that we are doing this we should clean up the descriptions of the tools themselves

packages/@n8n/instance-ai/src/agent/sanitize-mcp-schemas.ts
aalises: made the types stricter, the errors in this case should be enough for the LLM to know

packages/@n8n/instance-ai/src/tools/nodes.tool.ts
aalises: Added a strict-mode check in `6d7085529a` that compares enum members when the same field appears as a `z.enum` in multiple variants. Same pattern as the description check... throws at construction time for first-party tools, merges gracefully for external tools.

packages/@n8n/instance-ai/src/agent/__tests__/sanitize-mcp-schemas.test.ts
aalises: Done, the descriptions should be harmonisedd (name, description, versionId), data-tables (projectId, dataTableId, columnId, filter), and nodes (nodeType). + the new strict mode now enforces this going forward.

packages/@n8n/instance-ai/src/tools/nodes.tool.ts
mutdmour: Should we split them now? since the logic to merge "action" does not apply for external tools? or am I wrong?

packages/@n8n/instance-ai/src/agent/sanitize-mcp-schemas.ts
```diff
@@ -17,43 +17,133 @@ import { z } from 'zod';
...
-		const discriminatorValues: string[] = [];
+		// Phase 1: Collect metadata from all variants
+		const actionMeta: Array<{ value: string; description?: string }> = [];
+		const fieldMeta = new Map<
+			string,
+			Array<{ action: string; description?: string; type: z.ZodTypeAny }>
+		>();
 
 		for (const variant of variants) {
+			let actionValue = '';
```
mutdmour, line 54: if this function also applies to external tools, could this change how they behave? Not sure what external tools here might be? MCP tools?

packages/@n8n/instance-ai/src/tools/nodes.tool.ts
```diff
@@ -0,0 +1,370 @@
...
+	return createTool({
+		id: 'nodes',
+		description:
+			'Work with n8n node types. Actions:\n' +
+			'- list — list available node types, optionally filtered\n' +
+			'- search — search nodes by name or AI connection type\n' +
+			'- describe — get detailed node description (properties, credentials, I/O)\n' +
+			'- type-definition — get TypeScript type definitions for nodes\n' +
+			'- suggested — get curated node recommendations by workflow technique category\n' +
+			"- explore-resources — query real resources for a node's RLC parameters",
```
mutdmour: these can be removed now? and in other tools?

packages/@n8n/instance-ai/src/tools/research.tool.ts
```diff
@@ -0,0 +1,202 @@
...
+
+// ── Tool factory ────────────────────────────────────────────────────────────
+
+export function createResearchTool(context: InstanceAiContext) {
+	return createTool({
+		id: 'research',
+		description:
+			'Search the web or fetch page content. Actions:\n' +
+			'• web-search — search for information (API docs, error messages, guides)\n' +
+			'• fetch-url — fetch a page and extract content as markdown',
```
mutdmour: same here - are these still needed?

packages/@n8n/instance-ai/src/tools/nodes.tool.ts
aalises: true! will cleanup

packages/@n8n/instance-ai/src/tools/research.tool.ts
aalises: Cleaned up 🫡

mutdmour (approved):
LGTM

PR #28141, opened 2026-04-07
fix(Microsoft Teams Node): Downgrade OAuth2 scope from Group.ReadWrite.All to Group.Read.All

### LINKED ISSUE #27019: Teams Oauth Connector Bug
### Bug Description

Teams Oauth Connector has a bug / security vulnerability.  It is trying to request Group.ReadWriteAll, when that is NOT in the description of required permissions.  This is not a feature request, this is a bug of requiring permissions far beyond what is displayed.  The displayed permissions are correct, the bug is what it is requiring on line 7.

[n8n/packages/nodes-base/credentials/MicrosoftTeamsOAuth2Api.credentials.ts](https://github.com/n8n-io/n8n/blob/75edc5837f279b0a696b5cb802e36ef398509028/packages/nodes-base/credentials/MicrosoftTeamsOAuth2Api.credentials.ts#L7)

### To Reproduce

1.) Create teams connector (OAuth)
2.) Enter client ID and Secret from Azure App Registration
3.) Click Test
4.) Notice the azure requested permissions DO NOT MATCH what the n8n connector says it will require.

Image

### Expected behavior

Expected behavior after fixing the bug is that the permissions request will match what is stated in the app connector screen, and it will no longer attempt to request elevated and uneccessary permissions such as Group.ReadWriteAll

### Debug Info

# Debug info

## core

- n8nVersion: 2.4.8
- nodeJsVersion: 22.21.1
- nodeEnv: production
- database: postgres
- concurrency: -1
- consumerId: 7289138f-6c0d-42f0-8138-45de69b83239

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2026-03-13T16:21:29.873Z

### Operating System

Default (docker.n8n.io/n8nio/n8n:nightly)

### n8n Version

2.4.8

### Node.js Version

Default (docker.n8n.io/n8nio/n8n:nightly)

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

krisn0x:
## Summary

The Microsoft Teams OAuth2 credential requests `Group.ReadWrite.All` but the node only performs read operations on groups (listing teams, members, group membership). This is an over-privileged scope that used to block enterprise adoption.

**Changes:**
- `Group.ReadWrite.All` → `Group.Read.All` in default scopes
- Updated tests to match

https://linear.app/n8n/issue/NODE-4684
Fixes #27019

## Review / Merge checklist

- [x] Tests included.

Co-authored by Claude


packages/nodes-base/credentials/MicrosoftTeamsOAuth2Api.credentials.ts
```diff
@@ -4,7 +4,7 @@ const defaultScopes = [
 	'openid',
 	'offline_access',
 	'User.Read.All',
-	'Group.ReadWrite.All',
```
yehorkardash, line 7: Wouldn't this change break existing credentials, if the configured OAuth has only `Group.ReadWrite.All`?

dlavrenuek: @yehorkardash I've checked the spec and the code. According to the spec the `refresh token request` can include `scopes`, but this parameter is optional: https://datatracker.ietf.org/doc/html/rfc6749#section-6.

> The requested scope MUST NOT include any scope not originally granted by the resource owner, and if omitted is treated as equal to the scope originally granted by the resource owner.

I've found that in practice it means "you can downgrade scopes but not upgrade". Just from the spec this would mean that refreshing a token issued with `Group.ReadWrite.All` scope and passing the `Group.Read.All` scope would be an illegal request (unless MS treats the read scope as a subset of readwrite scope), but this is not relevant because we don't pass any scopes to the refresh token request. Following the execution chain we can see [in step 5](https://github.com/n8n-io/n8n/blob/master/packages/@n8n/client-oauth2/src/client-oauth2-token.ts#L73) that no scopes are added to the refresh request, so the token will get the same scopes. Only when the user manually clicks on "Reconnect" they will get a token with new scopes.


1. requestOAuth2 — entry point that signs and sends the request
[packages/core/.../request-helpers/oauth.ts#L229](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L229)

2. Trigger point — the request fails with the token-expired status code. Two paths, depending on which HTTP helper is used:

- n8n request path → catch on tokenExpiredStatusCode, calls retryWithNewToken [oauth.ts#L357](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L357)
- legacy request path → catch on tokenExpiredStatusCode, calls retryWithNewToken [oauth.ts#L379](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L379)
3. retryWithNewToken (closure defined inside requestOAuth2) → calls refreshOrFetchToken(refreshCtx)
[oauth.ts#L340](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L340) → [oauth.ts#L343](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L343)
4. refreshOrFetchToken branches on grant type
[oauth.ts#L116](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L116)
- clientCredentials → token.client.credentials.getToken() [oauth.ts#L151](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L151)
- otherwise (authorizationCode / PKCE) → token.refresh(tokenRefreshOptions) [oauth.ts#L153](https://github.com/n8n-io/n8n/blob/master/packages/core/src/execution-engine/node-execution-context/utils/request-helpers/oauth.ts#L153)
5. ClientOAuth2Token.refresh — builds the grant_type=refresh_token body and calls accessTokenRequest
[packages/@n8n/client-oauth2/src/client-oauth2-token.ts#L73](https://github.com/n8n-io/n8n/blob/master/packages/@n8n/client-oauth2/src/client-oauth2-token.ts#L73) → [client-oauth2-token.ts#L123](https://github.com/n8n-io/n8n/blob/master/packages/@n8n/client-oauth2/src/client-oauth2-token.ts#L123)
6. ClientOAuth2.accessTokenRequest — performs the actual HTTP POST to the token endpoint
[packages/@n8n/client-oauth2/src/client-oauth2.ts#L91](https://github.com/n8n-io/n8n/blob/master/packages/@n8n/client-oauth2/src/client-oauth2.ts#L91)

dlavrenuek: @yehorkardash I will test that the refresh flow does not request any scopes to confirm it

dlavrenuek: @yehorkardash confirming that the automatic token refresh request does not include any scopes, so any credentials connected before the change will have old scopes while refreshed successfully.

dlavrenuek (approved):

PR #28146, opened 2026-04-07
fix(Supabase Node): Don't display RPCs in an RLC for the table

RomanDavydchuk:
## Summary

Supabase's RLC to select the table would also load any RPCs defined, which may lead to confusion. This is because the method gets table name from OpenAPI exposed by Supabase (served at `/`). Example schema:
```json
{
  "swagger": "2.0",
  "info": {
    ...
  },
  "host": "tatqmzjyktunnpvvjynf.supabase.co:443",
  "basePath": "/",
  "schemes": [
    "https"
  ],
  "consumes": [
    ...
  ],
  "produces": [
    ...
  ],
  "paths": {
    "/": {
      "get": {
        "produces": [
          "application/openapi+json",
          "application/json"
        ],
        "responses": {
          "200": {
            "description": "OK"
          }
        },
        "summary": "OpenAPI description (this document)",
        "tags": [
          "Introspection"
        ]
      }
    },
    "/n8n-sample": {
      "get": {
        ...
      },
      "post": {
        ...
      },
      "delete": {
        ...
      },
      "patch": {
        ...
      }
    }
  },
  "definitions": {
    ...
  },
  "parameters": {
    ...
  },
  "externalDocs": {
    ...
  }
}
```

All keys in the `paths` object (except the first one - `/`) are assumed to be table names prefixed with `/`, but it can also contain RPCs that would have a `/rpc/` prefix. The change in this PR handles this

https://linear.app/n8n/issue/NODE-4743/community-issue-supabase-node-cannot-read-rpcfunction-parameters
Closes #27654

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Supabase/Supabase.node.ts
michael-radency: we should address this here as well by checking for `properties`
`if (!definitions[tableName]) return returnData;` before loop
or `for (const column of Object.keys(definitions[tableName]?.properties as IDataObject || {}))`

michael-radency (changes requested):

packages/nodes-base/nodes/Supabase/Supabase.node.ts
RomanDavydchuk: Fixed

michael-radency (approved):
LGTM

PR #28148, opened 2026-04-07
ci: Add concurrency group to docker-build-smoke workflow

mendral-agent:
## Summary

- Add `concurrency` with `cancel-in-progress: true` to the Docker Build Smoke Test workflow
- Prevents duplicate Docker builds when multiple pushes hit the same PR

## Problem

The `docker-build-smoke.yml` workflow runs on PRs that touch Docker-related files and on a daily schedule. It has no `concurrency` setting, so pushing twice to a PR that modifies Dockerfiles triggers two parallel full-chain Docker builds (pnpm install → pnpm build → base image → n8n + runners images), each on a 4-vCPU Blacksmith runner with all caches disabled.

The main PR workflow (`ci-pull-requests.yml`) already uses concurrency with cancel-in-progress, but this standalone workflow was missed.

## Fix

Added a concurrency group keyed on the PR number (for PR triggers) or git ref (for schedule/dispatch triggers):

```yaml
concurrency:
  group: docker-smoke-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

This ensures:
- Subsequent pushes to the same PR cancel the prior run
- Scheduled runs use `refs/heads/master` as the group key, so at most one nightly run proceeds
- Different PRs and manual dispatches run independently

## Verification

1. Push two commits in quick succession to a PR modifying a Dockerfile — the first run should be canceled.
2. Scheduled nightly runs should continue to execute normally.

---

> [!NOTE]
> Created by [Mendral](https://mendral.com). Tag @mendral-agent with feedback or questions.


shortstacked (approved):
🐱

PR #28157, opened 2026-04-08
fix(Microsoft Outlook Node): Prevent poll from skipping messages after API errors

BerniWittmann:
## Summary

The `poll()` method's catch block was missing a `return null` statement. When an API call failed in trigger mode with an existing `lastTimeChecked`, execution fell through to advance `lastTimeChecked = endDate`, permanently skipping any messages that arrived during the failed poll window. Adding `return null` in the catch block preserves the previous timestamp so those messages are retried on the next poll cycle.

https://linear.app/n8n/issue/NODE-4721

## Review / Merge checklist

- [x] Tests included.


RomanDavydchuk (approved):
LGTM and works as expected

PR #28159, opened 2026-04-08
fix(core): Omit empty scope from OAuth2 client credentials token request and improve error messaging

BGZStephen:
## Summary

When using a Generic OAuth2 API credential with **Grant Type: Client Credentials** and the Scope field left empty, n8n was sending `scope=` (empty string) in the token request body. This is non-standard per RFC 6749 §3.3 — scope is optional, and an explicit empty value is rejected by some OAuth2 servers with a `400 invalid_scope` or `401` error, preventing the credential from working at all.

Additionally, when token acquisition fails, the error message (`ResponseError: HTTP status 400`) gave no indication that it was the **token fetch** that failed — users saw an auth error and assumed the Authorization header was never sent to the API.

### Changes

- **Empty scope is no longer sent**: when the Scope field is blank, the `scope` parameter is omitted from the token request entirely rather than sent as `scope=`
- **Clearer error message on token fetch failure**: wraps `getToken()` to re-throw with `"Failed to acquire OAuth2 access token: <reason>"`, making it immediately clear the failure happened before any API request was attempted

### How to test

1. Create a **Generic OAuth2 API** credential:
   - Grant Type: `Client Credentials`
   - Authentication: `Send in Request Body`
   - Scope: _(leave empty)_
   - Fill in a real Access Token URL, Client ID, and Client Secret
2. Add an **HTTP Request** node using that credential
3. Execute — the token request should succeed (no `scope=` sent), and the API request should carry `Authorization: Bearer <token>`

To test the error message improvement: use invalid client credentials — the resulting error should now read `"Failed to acquire OAuth2 access token: ..."` instead of a raw `ResponseError`.

https://linear.app/n8n/issue/IAM-509

## Review / Merge checklist

- [x] Tests included.


guillaumejacquart (approved):
looks good !

afitzek (approved):

PR #28166, opened 2026-04-08
fix(core): Align VM expression engine error handler with legacy engine

ivov:
## Summary

The new expression engine's `tournament` error handler `E()` had different behavior than the legacy engine, causing many tests in the [VM expression E2E nightly job](https://github.com/n8n-io/n8n/actions/runs/24118384091) to fail.

For context, `tournament` wraps every expression with `try { return ...; } catch(e) { E(e, this); }`. The legacy `E()` handler (set via `setErrorHandler` in `expression.ts`) re-throws `ExpressionError` and `ExpressionExtensionError` and silently swallows everything else. The VM `E()` handler was re-throwing all non-`TypeError` errors, causing expressions that fail during paired-item resolution to crash in VM mode, where those silently return `undefined` in legacy mode.

Importantly, this means that paired-item resolution in this context (sub-nodes referencing main-flow nodes across AI connections) is apparently not supported. The legacy engine masks this by swallowing the error. Unclear if this is intentionally unsupported or an oversight.

> [!note]
> This fixes [many but not all](https://github.com/n8n-io/n8n/actions/runs/24128951449/job/70401107284) e2e issues.

https://linear.app/n8n/issue/CAT-2775

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
+++ b/packages/@n8n/expression-runtime/src/__tests__/integration.test.ts
@@ -377,58 +377,34 @@ describe('Integration: ExpressionEvaluator + IsolatedVmBridge', () => {
 
 	it('should handle throw null without crashing', () => {
 		const data = { $json: {} };
-		let error: Error | undefined;
-		try {
-			evaluator.evaluate('{{ (() => { throw null })() }}', data, caller);
-		} catch (e) {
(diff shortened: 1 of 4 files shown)


despairblue (approved):
Made some changes. Let me know if you disagree.

PR #28171, opened 2026-04-08
feat(core): Serve a nonce-based Content-Security-Policy on n8n's HTML pages

luigigubello:
**Update: 2026/08/17**

> ⚠️ **Disclosure:** AI agents were used to work on the n8n codebase. Given the kind of change, human review of the product-security trade-offs is important.

## Goal

Implementing a CSP Level 3 (nonce + `'strict-dynamic'`) for n8n.

https://linear.app/n8n/issue/SEC-1005

### Summary

This PR introduces a nonce-based Content-Security-Policy (nonce + `'strict-dynamic'`) on the HTML pages n8n renders itself.

```
script-src 'nonce-{random}' 'strict-dynamic' 'unsafe-eval'; object-src 'none'; base-uri 'none'
```

By default, for an initial delivery, n8n serves the header `Content-Security-Policy-Report-Only` and **no CSP is enforced**: violations are reported in the browser console without anything being blocked to avoid potential breaking changes. Pages that already serve a policy of their own (e.g. webhook responses, Form Trigger node, etc)  keep their `sandbox` policy untouched.

According to n8n's GitHub Security Advisories, 18 out of 119 are XSS (~15%). A nonce-based CSP blocks both injected `<script>` tags and inline event handlers, which is the shape most of those advisories take.

**‼️ Important before merging:**

1. This PR **changes the meaning of `N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY`**, which has been a boolean since **`1.100`**. It now holds a policy, like `N8N_CONTENT_SECURITY_POLICY`. **Boolean values keep their old behaviour and log a deprecation warning**, so no instance starts blocking on upgrade.
2. Every instance will start sending a new `Content-Security-Policy-Report-Only` header where there was none. It blocks nothing, but it will produce console violations. Users can set `report-uri` directive to collect this data.
3. The directive `object-src 'none'` blocks `<embed>`. This PR moves binary-data previews from `<embed>` to `<iframe>`, which `object-src` does not block, so previews survive, but any future `<embed>` will not. **This is a potential breaking change, even if `<embed>` is discouraged in modern web applications.**
4. The header `Content-Security-Policy` defined through the environment variable `N8N_CONTENT_SECURITY_POLICY` is deliberately **not** enforced for pages that serve already a Content-Security-Policy (in particular, `sandbox` directive for some nodes (e.g. Webhook Node, etc)).

### Considerations

The two env. variables (`N8N_CONTENT_SECURITY_POLICY` and `N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY`) are now symmetric: both hold a policy, in either accepted format (hermes.js format or W3C format, and `{}` (or `""`) switches that header off. They differ only in their default: nothing is enforced until you set one, and n8n's own policy is in `report-only` mode until you change it.

```
N8N_CONTENT_SECURITY_POLICY            default: {}           
N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY default: script-src 'nonce-{random}' 'strict-dynamic' 'unsafe-eval'; object-src 'none'; base-uri 'none'
```

Both accept a policy string (e.g. `script-src <nonce> 'strict-dynamic'`) and the `helmet.js` -format directives object n8n has always accepted (e.g.`{"script-src":["<nonce>","'strict-dynamic'"]}`). The format is detected from the value: anything starting with `{` is parsed as JSON. `{}`, an empty string and whitespace all mean "send no such header". Write `<nonce>` where the per-request nonce should go, or `default` to set the default n8n's policy.

Both headers print violations in the browser console; only the enforced one also blocks. So the `report-only` header is to evaluate and test the policy: add `report-uri` to either policy, or both, to send violations to a collector.

Two smaller fixes came out of testing this and are included:

### Minor changes

- The form 404, 409, invalid-token and no-action-required pages served **no** CSP at all. `curl -sI /form/anything` returned HTML with no policy. They now carry the same `sandbox` policy as the rest of the form flow.

### Unit Test

```bash
# policy parsing, both env vars, and the middleware that serves the headers
cd packages/cli
pnpm test src/security/__tests__ src/middlewares/__tests__/content-security-policy.test.ts

# pages that keep their own sandbox policy, and the form error pages
pnpm test \
  src/webhooks/__tests__/webhook-response-headers.test.ts \
  src/webhooks/__tests__/waiting-forms.test.ts \
  src/webhooks/__tests__/waiting-webhooks.test.ts \
  src/__tests__/response-helper.test.ts

# defaults of the two env vars
cd packages/@n8n/config && pnpm test test/config.test.ts

# binary-data previews render in an iframe, which object-src 'none' allows
cd packages/frontend/editor-ui
pnpm test src/features/ndv/runData/components/BinaryDataViewModal.test.ts
pnpm test src/features/ai/instanceAi/__tests__/ToolResultRenderer.test.ts
```

### Manual Tests

Build a Docker image from this branch, from the repo root:

```bash
pnpm build:docker
```

That builds the monorepo and then the image, tagged `n8nio/n8n:local`. 

---

**1. Default: report-only, nothing enforced**

```bash
docker run --rm -p 5678:5678 n8nio/n8n:local
curl -I -s -X GET 'http://localhost:5678' | grep -iE "Content-Security-Policy"
```

**Expected output:**

```
Content-Security-Policy-Report-Only: script-src 'nonce-{random-value}' 'strict-dynamic' 'unsafe-eval'; object-src 'none'; base-uri 'none'
```

There must be no `Content-Security-Policy` header. Run the command a second time: the `{random-value}` must differ.

---

**2. The nonce in the header matches the one in the page**

```bash
curl -s -D /tmp/h.txt http://localhost:5678/ -o /tmp/b.html
grep -io "nonce-[A-Za-z0-9_-]*" /tmp/h.txt
grep -o 'nonce="[^"]*"' /tmp/b.html
```

**Expected output:** the same nonce.

---

**3. Enforce n8n's policy instead of only trialling it**

```bash
docker run --rm -p 5678:5678 \
  -e N8N_CONTENT_SECURITY_POLICY=default \
  -e N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY="" \
  n8nio/n8n:local

curl -I -s -X GET 'http://localhost:5678' | grep -iE "Content-Security-Policy"
```

**Expected output:**

```
Content-Security-Policy: script-src 'nonce-{random-value}' 'strict-dynamic' 'unsafe-eval'; object-src 'none'; base-uri 'none'
```

---

**4. Your own policy, keeping the nonce**

```bash
docker run --rm -p 5678:5678 \
  -e N8N_CONTENT_SECURITY_POLICY="script-src <nonce>" \
  n8nio/n8n:local

curl -I -s -X GET 'http://localhost:5678' | grep -iE "Content-Security-Policy"
```

**Expected output:**

```
Content-Security-Policy: script-src 'nonce-{random-value}'
Content-Security-Policy-Report-Only: script-src 'nonce-{random-value}' 'strict-dynamic' 'unsafe-eval'; object-src 'none'; base-uri 'none'
```

The report-only header is still n8n's policy, because that variable keeps its default. Set it to `""`  to drop it.

---

**5. The `helmet.js` format still behaves exactly as before**

```bash
docker run --rm -p 5678:5678 \
  -e N8N_CONTENT_SECURITY_POLICY='{"frame-ancestors":["http://localhost:3000"]}' \
  n8nio/n8n:local

curl -I -s -X GET 'http://localhost:5678' | grep -iE "Content-Security-Policy"
```

**Expected output:**

```
Content-Security-Policy: frame-ancestors http://localhost:3000
Content-Security-Policy-Report-Only: script-src 'nonce-{random-value}' 'strict-dynamic' 'unsafe-eval'; object-src 'none'; base-uri 'none'
```

The enforced header carries exactly the configured directives.

---

**6. Pages that carry their own policy are left alone**

```bash
curl -I -s 'http://localhost:5678/form/does-not-exist' | grep -iE "Content-Security-Policy"
```

**Expected output:**

```
Content-Security-Policy: sandbox allow-downloads allow-forms allow-modals ... allow-scripts ...
```

Only the `sandbox` policy.

---

### Notes & Migration

- **Nothing is enforced by default.** The default policy is `report-only`, so an upgrade cannot break a running instance.
- **`N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY` behavior and scope changed.** It was a boolean from `1.100` until this PR, where you can use it to set the header `Content-Security-Policy-Report-Only`, aligned to `N8N_CONTENT_SECURITY_POLICY_REPORT` behavior.
    - Booleans (`true`, `false`, `1`, `0`) keep their original meaning and log a deprecation warning naming the replacement. In particular, `true` must not start *enforcing* a policy that an instance deliberately kept report-only.
    - Migrate by moving the policy into the variable that matches the intent: `N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY="script-src 'self'"` instead of `N8N_CONTENT_SECURITY_POLICY="script-src 'self'"` + `..._REPORT_ONLY=true`.
- `N8N_CONTENT_SECURITY_POLICY` keeps its `helmet.js` JSON format, with a regression test pinning that behaviour.
- `N8N_INSECURE_DISABLE_WEBHOOK_IFRAME_SANDBOX` and `N8N_INSECURE_DISABLE_FORM_HTML_SANDBOX` still disable the `sandbox` directive on those pages.
- The nonce is injected into built assets only, so `pnpm dev` (Vite dev server) does not serve nonces.


packages/cli/src/server.ts
```diff
@@ -430,8 +479,36 @@ export class Server extends AbstractServer {
...
+					// Apply CSP per-request so we can inject the nonce into
+					// `script-src` while merging any directives from
+					// `N8N_CONTENT_SECURITY_POLICY`. This avoids overwriting
+					// route-specific CSP set elsewhere and honors the
+					// `N8N_CONTENT_SECURITY_POLICY_REPORT_ONLY` flag.
+					const cspMiddleware = buildCspMiddleware(cspDirectives, cspReportOnly, nonce);
+
+					let indexHtml = '';
+					try {
+						indexHtml = await readFile(resolve(staticCacheDir, 'index.html'), 'utf8');
```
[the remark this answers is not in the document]
luigigubello: This is true, but not a good idea. If we want to use a cached `index.html`, we need to switch from a nonce-based CSP to a hash-based CSP, which I honestly prefer, but that requires strong, bulletproof hash-calculation automation. This is more of a product decision than a technical one. But if we want to go with a hash-based CSP, I need to rethink and rewrite the PR (close and create a new one).

> Cache the file contents once (e.g. lazily on first request) and only perform the nonce placeholder replacement per-request. 

But I can check this later, maybe cubic-ai is right :)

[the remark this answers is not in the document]
luigigubello: No. The current version `server.ts` doesn't cache `index.html`. This is not a good review comment for this PR. I can resolve the conversation: **won't fix**.

packages/frontend/editor-ui/src/features/ai/instanceAi/components/ToolResultFile.vue
```diff
@@ -1,12 +1,21 @@
...
 	mimeType: string;
 }>();
+
+const i18n = useI18n();
 </script>
 
 <template>
-	<embed :src="`data:${mimeType};base64,${data}`" :type="mimeType" :class="$style.file" />
+	<!-- An iframe rather than an embed, so a `object-src 'none'` CSP does not refuse it -->
+	<iframe
```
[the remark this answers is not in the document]
luigigubello: I don't resolve this for now, I am just moving from `<embed>` to `<iframe>`, the score is feature parity

packages/cli/src/security/content-security-policy.ts
```diff
@@ -0,0 +1,146 @@
+import type { Logger } from '@n8n/backend-common';
+import { DEFAULT_CONTENT_SECURITY_POLICY } from '@n8n/config';
+
+/**
+ * Token that users can place in a policy where they want n8n to substitute the
+ * per-request nonce, e.g. `N8N_CONTENT_SECURITY_POLICY="script-src <nonce>"`.
+ */
+export const NONCE_PLACEHOLDER = '<nonce>';
```
dlavrenuek, line 8: This is also used in https://github.com/n8n-io/n8n/pull/28171/changes#diff-e1ebdc279f7473503f8ac51d7e522a30dab59aa73383f25cf84982225b4dcd08R14 without either one importing the other one. Technically it means that if the placeholder is changed in one place then it will not work anymore. This is fine from my point of view since it's a token/constant that is not supposed to be changed.

packages/cli/src/security/content-security-policy.ts
```diff
@@ -0,0 +1,146 @@
...
+
+	return policy.length > 0 ? policy : undefined;
+};
+
+/**
+ * Decide which CSP headers to send from the two env vars. Only the report-only var
+ * holds a policy by default, so a new instance reports violations but cannot break
+ * on them.
+ */
+export const resolveContentSecurityPolicies = (
```
dlavrenuek, line 91: All of the `resolveContentSecurityPolicies` and if I see it correctly parse+serialize functionality implemented here reimplements what is already possible via zod schema for on the config level. Instead of custom functionality we should define a zod schema to parse the env values and have a clean type for the configuration properties in `security.config.ts`. You can see an example of how it works in f.e. `ssrf-protection.config.ts` but I also saw the same approach in other places

packages/cli/src/middlewares/content-security-policy.ts
```diff
@@ -0,0 +1,71 @@
...
+			enumerable: true,
+			configurable: true,
+		});
+
+		type WriteHead = Response['writeHead'];
+		const writeHead = res.writeHead.bind(res);
+
+		// Same technique as the `on-headers` package: there is no event for "headers
+		// about to be sent", so wrap the call that sends them.
+		res.writeHead = ((...args: Parameters<WriteHead>) => {
```
RicardoE105, line 79: Why not use the on-headers package? instead of monkey patching this

RicardoE105: I guess we do not want the dependency we can make this a helpers?

```
 *
 * Typed as `unknown[]` because `Parameters<>` collapses `writeHead`'s overloads to the
 * two-argument one, which cannot express the `(status, message, headers)` form.
 */
const copyWriteHeadHeaders = (res: Response, args: unknown[]) => {
      const headers = typeof args[1] === 'string' ? args[2] : args[1];
      if (typeof headers !== 'object' || headers === null || Array.isArray(headers)) return;

      for (const [name, value] of Object.entries(headers)) {
              if (typeof value === 'string' || typeof value === 'number' || Array.isArray(value)) {
                      res.setHeader(name, value);
              }
      }
};
```

RicardoE105: Also that should cover this issue raised by AI

 `writeHead(status[, message][, headers])` can carry headers that are not on the response yet, so a check that reads only `res` would miss a `content-type` set that way. Copy them across first; `writeHead` setting them again is a no-op. The array form is not handled: nothing in n8n passes one.

RicardoE105: Then in the wrapper we do  copyWriteHeadHeaders(res, args);

packages/cli/src/security/content-security-policy.ts
```diff
@@ -0,0 +1,146 @@
...
+ * Token that users can place in a policy where they want n8n to substitute the
+ * per-request nonce, e.g. `N8N_CONTENT_SECURITY_POLICY="script-src <nonce>"`.
+ */
+export const NONCE_PLACEHOLDER = '<nonce>';
+
+/**
+ * Placeholder baked into the built HTML at build time (see the `csp-nonce` Vite
+ * plugin), replaced with the per-request nonce when the page is served.
+ */
+export const HTML_NONCE_PLACEHOLDER = '{{CSP_NONCE}}';
```
RicardoE105: Should this value be a constant? that this file and the vite.config can import. Else there is a chance these two can drift by mistake.

luigigubello: Sorry, I misunderstood the question 😇

> Should this value be a constant? 

Yes, this is a constant, that is replaced with a random nonce. Where do you recommend to move it?

packages/cli/src/security/content-security-policy.ts
dlavrenuek: Also with the implementation here we have the env naming in two different places, one in config package and one in the security package. By moving everything to zod schemas in config this will also be solved

packages/cli/src/server.ts
```diff
@@ -443,21 +446,46 @@ export class Server extends AbstractServer {
...
 					headers: { accept },
 				} = req;
+				// A request with no `Accept` also gets the page. Only this handler fills in the
+				// nonce placeholders, so `express.static` would serve `index.html` with the
+				// placeholders intact, and no script on it could run under the CSP.
 				if (
 					method === 'GET' &&
-					accept &&
-					(accept.includes('text/html') || accept.includes('*/*')) &&
+					(!accept || accept.includes('text/html') || accept.includes('*/*')) &&
```
dlavrenuek, line 467: The accept header value could be sent as uppercase and servers treat it the same way. In practice this should be irrelevant because afaik browsers send it as `text/html` but may be there are some browsers that do it differently. Express actually has an `req.accepts` method, which catches other header values that are also parsed as html, like `text/*` -> parsed as `text/html` in express according to https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.1 . The accept.includes problem is preexisting but since we want to strenghten security I would change the check to


```suggestion
					(!accept || req.accepts('html') || accept.includes('*/*')) &&
```

probably `accept.includes('*/*')` is not necessary but does not hurt to keep it

packages/cli/src/middlewares/content-security-policy.ts
```diff
@@ -0,0 +1,99 @@
...
+			if (res.headersSent) return writeHead(...args);
+
+			copyWriteHeadHeaders(res, args);
+
+			if (isHtmlResponse(res) && !hasOwnPolicy(res)) {
+				if (enforced) {
+					res.setHeader(ENFORCED_HEADER, renderContentSecurityPolicy(enforced, getNonce()));
+				}
+				if (reportOnly) {
+					res.setHeader(REPORT_ONLY_HEADER, renderContentSecurityPolicy(reportOnly, getNonce()));
```
dlavrenuek, line 90: Is the new header supposed to be added for form and webhook requests when the sandbox policy is off? Right now it's not added and I can't follow the code why, because  `createContentSecurityPolicyMiddleware()` is added as a middleware globally and not per request, may be I am missing something or may be the header should not be added at all and everything is correct.

packages/frontend/@n8n/i18n/src/locales/en.json
```diff
@@ -434,6 +434,7 @@
 	"banners.workflowAutoDeactivated.message": "This workflow was automatically deactivated due to multiple crashed executions. Please review and reactivate it when ready.",
 	"binaryDataDisplay.backToList": "Back to list",
 	"binaryDataDisplay.backToOverviewPage": "Back to overview page",
+	"binaryDataDisplay.filePreview": "File preview",
```
dlavrenuek, line 437: This change looks unrelated to CSP, but does not hurt 😆

packages/frontend/editor-ui/src/features/ai/instanceAi/components/ToolResultFile.vue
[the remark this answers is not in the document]
dlavrenuek: I could not test it due to file tool handling now being broken in AI Assistant. Worth checking when it's fixed

packages/frontend/editor-ui/vite.config.mts
```diff
@@ -101,6 +103,18 @@ const plugins: UserConfig['plugins'] = [
...
+		// Build output only: the dev server serves no CSP.
+		name: 'csp-nonce',
+		transformIndexHtml(html, ctx) {
+			if (ctx.server) return html;
+			return html.replace(
+				/<script(?![^>]*\bnonce=)([^>]*)>/g,
+				`<script nonce="${HTML_NONCE_PLACEHOLDER}"$1>`,
+			);
+		},
+	},
```
dlavrenuek, line 117: vite supports adding the [nonce placeholder for build time](https://vite.dev/guide/features#content-security-policy-csp) (inside `defineConfig` and not in in plugins). This will also add the nonce placeholder to all stylesheet (`<link />`) files, that's not really required right now but will allow adding the `style-src: nonce-xxx` CSP directive without the need to extend our code

```
	html: {
		cspNonce: '{{CSP_NONCE}}',
	},
```

dlavrenuek (changes requested):
Works well, tested the UI, webhooks and forms with sandbox on/off and the OIDC login flow, all worked the same as before. Some comments and suggestions below

luigigubello:
Superseded by #36749.

Closing this in favour of #36749, which carries the identical commits (`d06f6ea`) on a branch in `n8n-io/n8n` instead of my fork. The move is purely mechanical: a fork branch can't be built into a Cloud preview instance, because the image build dispatches against `n8n-io/n8n` and fails with "No ref found" for a fork ref.

This branch is left in place so the discussion permalinks above keep resolving.

packages/frontend/editor-ui/vite.config.mts
luigigubello: Wow, this review is gold. I am working on this!

PR #28172, opened 2026-04-08
feat(ai-builder): Fix IF/Switch/Filter node misconfiguration in builder (no-changelog)

JoseBra:
## Summary
- Fix AI builder generating IF/Switch/Filter nodes with missing `conditions.options` (causes `caseSensitive` crash) and wrong `rules.rules` key (causes `Could not find property option` crash)
- Improve builder prompts with correct conditions structure, examples, and self-check checklist
- Make `options` and `combinator` required in the `FilterValue` type the LLM sees
- Add `filterNodeValidator` to catch structural misconfigurations as blocking errors at build time

## Results (notification-router × 5 runs)
| | Before | After |
|---|---|---|
| Structural crashes | 3/5 runs (60%) | 0/5 runs (0%) |
| Scenarios passed | 3/15 (20%) | 10/15 (67%) |
| Clean sweeps (3/3) | 1 | 3 |

Full suite (27 scenarios): 48% pass rate, zero crashes, all 8 workflows built.

### Before
Screenshot 2026-04-08 at 09 49 42
Screenshot 2026-04-08 at 09 54 49

### After
Screenshot 2026-04-08 at 09 50 02
Screenshot 2026-04-08 at 09 54 56

## Related Linear ticket
https://linear.app/n8n/issue/TRUST-35

## Test plan
- [x] 13 unit tests for `filterNodeValidator`
- [x] 11428 SDK tests pass (89 suites)
- [x] Before/after eval comparison (5 runs each, same workflow, same conditions)

### DIFF
--- a/packages/@n8n/instance-ai/evaluations/data/workflows/weather-monitoring.json
+++ b/packages/@n8n/instance-ai/evaluations/data/workflows/weather-monitoring.json
@@ -1,5 +1,5 @@
 {
-	"prompt": "Every hour, check the current weather for London, New York, and Tokyo using the OpenWeatherMap API (GET https://api.openweathermap.org/data/2.5/weather?q={city}&units=metric&appid=YOUR_API_KEY). Use 3 separate HTTP Request nodes, one per city. If any city has a temperature above 30°C, send a Telegram alert to chat ID -1001234567890 listing the hot cities. Log all readings to an Airtable base (base ID: 'appABC123def456', table ID: 'tblWeather789xyz') with columns: city, temperature, humidity, timestamp. Configure all nodes as completely as possible and don't ask me for credentials, I'll set them up later.",
+	"prompt": "Every hour, check the current weather for London, New York, and Tokyo using the OpenWeatherMap API (GET https://api.openweathermap.org/data/2.5/weather?q={city}&units=metric&appid=YOUR_API_KEY). Use 3 separate HTTP Request nodes, one per city. If any city has a temperature above 30°C, send a Telegram alert to chat ID -1001234567890 listing the hot cities. Log all readings to an Airtable table with columns: city, temperature, humidity, timestamp. Configure all nodes as completely as possible and don't ask me for credentials, I'll set them up later.",
 	"complexity": "complex",
 	"tags": [
 		"build",
--- a/packages/@n8n/workflow-sdk/src/generate-types/generate-types.ts
+++ b/packages/@n8n/workflow-sdk/src/generate-types/generate-types.ts
@@ -70,7 +70,7 @@ function generateFilterTypeDeclaration(exported: boolean): string {
 	return [
 		`${prefix} FilterOptionsValue = { caseSensitive?: boolean; leftValue?: string; typeValidation?: 'strict' | 'loose' };`,
 		`${prefix} FilterConditionValue = { id?: string; leftValue: unknown; operator: { type: string; operation: string }; rightValue: unknown };`,
-		`${prefix} FilterValue = { options?: FilterOptionsValue; conditions: FilterConditionValue[]; combinator?: 'and' | 'or' };`,
+		`${prefix} FilterValue = { options: FilterOptionsValue; conditions: FilterConditionValue[]; combinator: 'and' | 'or' };`,
 	].join('\n');
 }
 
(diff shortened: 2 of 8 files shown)


packages/@n8n/workflow-sdk/src/workflow-builder/plugins/validators/filter-node-validator.ts
```diff
@@ -0,0 +1,134 @@
...
+		if (!params) return issues;
+
+		const mapKey = findMapKey(graphNode, ctx);
+		const originalName = node.name;
+		const renamed = isAutoRenamed(mapKey, originalName);
+		const displayName = renamed ? mapKey : originalName;
+		const origForWarning = renamed ? originalName : undefined;
+		const nodeRef = formatNodeRef(displayName, origForWarning, node.type);
+
+		// IF and Filter nodes: conditions is directly on params
```
aalises, line 95: The `'conditions' in conditions` guard could mean that a malformed node like `{ conditions: { options: {...}, combinator: 'and' } }` (missing the inner `conditions` array) passes validation silently. Worth adding a check for missing/empty `conditions` array

packages/@n8n/workflow-sdk/src/generate-types/generate-types.ts
```diff
@@ -70,7 +70,7 @@ function generateFilterTypeDeclaration(exported: boolean): string {
 	return [
 		`${prefix} FilterOptionsValue = { caseSensitive?: boolean; leftValue?: string; typeValidation?: 'strict' | 'loose' };`,
 		`${prefix} FilterConditionValue = { id?: string; leftValue: unknown; operator: { type: string; operation: string }; rightValue: unknown };`,
```
aalises, line 72: Making `options` and `combinator` required is correct (matches the canonical type in `packages/workflow/src/interfaces.ts`), but `generate-types.test.ts` isn't updated in this PR. If it has snapshots covering `FilterValue`, CI will break (classic snapshot testing shenanigans, not a big fan...)

packages/@n8n/instance-ai/evaluations/data/workflows/weather-monitoring.json
```diff
@@ -1,5 +1,5 @@
 {
-	"prompt": "Every hour, check the current weather for London, New York, and Tokyo using the OpenWeatherMap API (GET https://api.openweathermap.org/data/2.5/weather?q={city}&units=metric&appid=YOUR_API_KEY). Use 3 separate HTTP Request nodes, one per city. If any city has a temperature above 30°C, send a Telegram alert to chat ID -1001234567890 listing the hot cities. Log all readings to an Airtable base (base ID: 'appABC123def456', table ID: 'tblWeather789xyz') with columns: city, temperature, humidity, timestamp. Configure all nodes as completely as possible and don't ask me for credentials, I'll set them up later.",
+	"prompt": "Every hour, check the current weather for London, New York, and Tokyo using the OpenWeatherMap API (GET https://api.openweathermap.org/data/2.5/weather?q={city}&units=metric&appid=YOUR_API_KEY). Use 3 separate HTTP Request nodes, one per city. If any city has a temperature above 30°C, send a Telegram alert to chat ID -1001234567890 listing the hot cities. Log all readings to an Airtable table with columns: city, temperature, humidity, timestamp. Configure all nodes as completely as possible and don't ask me for credentials, I'll set them up later.",
```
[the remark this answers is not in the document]
JoseBra: Intentional, the Airtable node was missconfigured with those IDs anyway, it's better to let the builder set their own defaults in this case.

packages/@n8n/workflow-sdk/src/workflow-builder/plugins/validators/filter-node-validator.ts
JoseBra: Good catch, fixed in the last commit ✅

packages/@n8n/workflow-sdk/src/generate-types/generate-types.ts
JoseBra: It only checks the existance of FilterValue, but not its content, we should be fine with the change.

aalises (approved):

PR #28184, opened 2026-04-08
feat: Implement session based permission modes in Computer Use

dlavrenuek:
## Summary

Updates the startup process of Computer Use to not require setting up permissions for different tool groups and removes the confirmation of the working directory. Instead these can be approved or configured when the connection is initialized.

Bildschirmfoto 2026-04-08 um 15 14 47

https://linear.app/n8n/issue/NODE-4797/change-permission-modes-to-be-bound-to-session-and-not-globally
https://linear.app/n8n/issue/NODE-4799/rename-filesystem-root-directory-to-ai-working-directory

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/computer-use/spec/local-gateway.md
+++ b/packages/@n8n/computer-use/spec/local-gateway.md
@@ -122,35 +122,51 @@ session modes are supported:
 
 ## Connection Flow
 
-### 1. Capability Preview & Configuration
+### 1. Silent startup
 
-Before the Local Gateway initiates a connection to the n8n instance, the user
-is shown a list of capabilities that the local machine supports. Capabilities
-that are not available on the machine (e.g. computer control on a headless
-server) are indicated as unavailable.
+The app starts silently. If no settings file exists, one is created
+automatically using the **Recommended** template (`filesystemDir` left empty).
+No prompts are shown.
 
-The user can enable or disable each capability individually. This gives the
-user explicit control over what the AI is permitted to do on their machine
-for this connection.
+### 2. Connect command
 
-The user must confirm the capability selection before the connection proceeds.
+The user runs the connect command (URL + token provided via CLI arguments or
+served automatically via daemon mode).
 
-### 2. Establishing a Connection
+### 3. Confirmation prompt
 
-After the user confirms, the Local Gateway connects to the n8n instance and
-registers the selected capabilities. The AI Agent is immediately aware of
-which tools are available and can use them in subsequent conversations.
+The app shows a single confirmation prompt displaying:
 
-### 3. Active Connection
+- The URL being connected to
+- The current permission table (one row per tool group)
+- The working directory
+
+The user is offered three choices: **Yes**, **Edit permissions / directory**,
+or **No**.
+
+### 4. Optional edit
+
+If the user chooses **Edit**, they can adjust per-group permission modes and/or
+the working directory for this session. These edits are **session-local** — they
+are not written back to the settings file.
+
+### 5. Session established
+
+The session is established with the confirmed settings. The AI is immediately
+aware of which tools are available and can use them in subsequent conversations.
+
+### 6. Active connection
 
 While connected:
 
 - The user can see that their Local Gateway is active.
 - The AI can invoke any of the registered capabilities as needed during a
   conversation.
+- Session rules (`allowForSession`) accumulate in memory for the duration of
+  the connection.
 - The connection persists across page reloads.
 
-### 4. Disconnection
+### 7. Disconnection
 
 The user can explicitly disconnect the Local Gateway at any time. After
 disconnection, the AI no longer has access to any local capabilities. If the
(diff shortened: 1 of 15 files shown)


packages/@n8n/computer-use/src/startup-config-cli.ts
```diff
@@ -131,83 +80,30 @@ function isAllDeny(permissions: Record<ToolGroup, PermissionMode>): boolean {
...
- * Apply a named template to a config, merging existing CLI/ENV overrides on top.
- * Useful for non-interactive pre-seeding (e.g. `--template yolo` in tests or CI).
- */
-export function applyTemplate(config: GatewayConfig, templateName: TemplateName): GatewayConfig {
-	const tpl = getTemplate(templateName);
-	return {
-		...config,
-		permissions: { ...tpl.permissions, ...config.permissions },
-	};
+	await fs.mkdir(nodePath.dirname(filePath), { recursive: true });
```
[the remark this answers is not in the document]
dlavrenuek: Computer Use creates the directory with current user's permissions - this directory also belongs to the user and should have default mode. We are not in the scope of n8n instance here.

packages/@n8n/computer-use/src/startup-config-cli.ts
```diff
@@ -131,83 +80,30 @@ function isAllDeny(permissions: Record<ToolGroup, PermissionMode>): boolean {
...
-/**
- * Return the template name for display purposes given a `--template` CLI flag value.
- * Falls back to 'default' for unknown values.
- */
-export function resolveTemplateName(raw: string | undefined): TemplateName {
-	if (raw === 'yolo' || raw === 'custom' || raw === 'default') return raw;
-	return 'default';
-}
+	const content = JSON.stringify(
+		{ permissions, filesystemDir: '', resourcePermissions: {} },
```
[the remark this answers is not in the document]
dlavrenuek: This is not correct, the stored permissions are used as default values for session based permissions and can be changed when the connection is initialized

packages/@n8n/computer-use/src/daemon.ts
```diff
@@ -136,26 +139,45 @@ async function handleConnect(req: http.IncomingMessage, res: http.ServerResponse
...
+			if (!stat.isDirectory()) {
+				jsonResponse(res, 400, { error: `Invalid directory: ${session.dir}` });
+				return;
+			}
+		} catch {
+			jsonResponse(res, 400, { error: `Invalid directory: ${session.dir}` });
+			return;
+		}
+
+		state.session = session;
```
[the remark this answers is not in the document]
dlavrenuek: that should not be a problem since session is overridden on new connection

packages/@n8n/computer-use/src/daemon.ts
```diff
@@ -136,26 +139,45 @@ async function handleConnect(req: http.IncomingMessage, res: http.ServerResponse
...
-			jsonResponse(res, 403, { error: 'Connection rejected by user.' });
-			return;
-		}
-	}
-
 	try {
 		const store = settingsStore ?? (await settingsStorePromise);
 		settingsStore ??= store;
 
+		const { GatewaySession } = await import('./gateway-session');
```
elsmr: Is there a reason for this to be a dynamic import?

packages/@n8n/computer-use/src/settings-store.ts
```diff
@@ -83,72 +77,80 @@ export class SettingsStore {
...
+		return { allow: rp?.allow ?? [], deny: rp?.deny ?? [] };
 	}
 
+	// ---------------------------------------------------------------------------
+	// Mutation methods
+	// ---------------------------------------------------------------------------
+
 	alwaysAllow(toolGroup: ToolGroup, resource: string): void {
 		const rp = this.getOrInitResourcePermissions(toolGroup);
 		if (!rp.allow.includes(resource)) {
```
elsmr, line 143: Not really from this PR but a Set/Map could make this faster if there's a lot of stored resources

packages/@n8n/computer-use/src/settings-store.ts
```diff
@@ -83,72 +77,80 @@ export class SettingsStore {
...
+	 * Persist updated defaults to the settings file (for future sessions).
+	 * Preserves existing resourcePermissions.
 	 */
-	check(toolGroup: ToolGroup, resource: string): PermissionMode {
-		const rp = this.persistent.resourcePermissions[toolGroup];
-		if (rp?.deny.includes(resource)) return 'deny';
-		if (rp?.allow.includes(resource)) return 'allow';
-		if (this.hasSessionAllow(toolGroup, resource)) return 'allow';
-		return this.getGroupMode(toolGroup);
+	async updateDefaults(permissions: Record<ToolGroup, PermissionMode>, dir: string): Promise<void> {
```
elsmr: I think this is not used

packages/@n8n/computer-use/spec/local-gateway.md
```diff
@@ -122,35 +122,51 @@ session modes are supported:
...
+- The working directory
+
+The user is offered three choices: **Yes**, **Edit permissions / directory**,
+or **No**.
+
+### 4. Optional edit
+
+If the user chooses **Edit**, they can adjust per-group permission modes and/or
+the working directory for this session. These edits are **session-local** — they
+are not written back to the settings file.
```
elsmr, line 151: So the only way to edit the "defaults" is by editing the settings file manually?

elsmr (commented):
Works nicely! 💯 some small comments/questions

packages/@n8n/computer-use/src/settings-store.ts
dlavrenuek: Good idea, created https://linear.app/n8n/issue/NODE-4804/use-set-internally-for-list-of-alloweddenied-resourses as a follow up

packages/@n8n/computer-use/spec/local-gateway.md
dlavrenuek: yes. When we have the native app it will also be possible to do it via settings there

packages/@n8n/computer-use/src/settings-store.ts
dlavrenuek: This is correct. I did not remove it because it would be used in the native app at some point but since we don't even know if we would have one I removed it now and can add it later

packages/@n8n/computer-use/src/daemon.ts
dlavrenuek: absolutely no reason, fixed

elsmr (approved):

PR #28185, opened 2026-04-08
feat(core): Configure OIDC settings via env vars

geemanjs:
## Summary
Configure OIDC via environment variables. 
```
N8N_SSO_MANAGED_BY_ENV
N8N_SSO_USER_ROLE_PROVISIONING
N8N_SSO_OIDC_CLIENT_ID
N8N_SSO_OIDC_CLIENT_SECRET
N8N_SSO_OIDC_DISCOVERY_ENDPOINT
N8N_SSO_OIDC_LOGIN_ENABLED
N8N_SSO_OIDC_PROMPT
N8N_SSO_OIDC_ACR_VALUES
```

> Note it is recommended to pass in the client secret via the `_FILE` method:
> ```N8N_SSO_OIDC_CLIENT_SECRET_FILE=/mnt/oidc-client-secret```

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/config/src/configs/instance-settings-loader.config.ts
```diff
@@ -26,4 +26,36 @@ export class InstanceSettingsLoaderConfig {
 	 */
 	@Env('N8N_INSTANCE_OWNER_PASSWORD_HASH')
 	ownerPasswordHash: string = '';
+
+	// --- SSO ---
```
geemanjs, line 30: I'm unsure if we should have a global config here - or if we put it into the specific configs for each are 🤔.

konstantintieber: What about `packages/@n8n/config/src/configs/sso.config.ts`?

konstantintieber: I would expect these env vars being part of `sso.config.ts`, since that is also where all of the other SSO specific configs live.

geemanjs: I agree - but then I think we have to put the `reading` / `writing` also into the module.. Which i'd be a fan of. It follows my "make the code easy to delete" model much better. Right now if you delete oidc you need to delete it from

I think the initial goal (correct me if I'm wrong @ireneea) was to have a single location where you can see bootstrapped values?

konstantintieber: > we have to put the reading / writing also into the module

What's the reading/writing you are referring to here? You mean reading the env vars and persisting them via the settings repo? What is implemented in `OidcInstanceSettingsLoader`?

I don't think the location of the env var definitions and the location of the settings loaded have to be in the same place.

By having these env vars inside `
packages/@n8n/config/src/configs/instance-settings-loader.config.ts`, they are detached from the existing env vars regarding SSO.
As a developer working on the SSO code, I find placing them elsewhere than `sso.config.ts` makes it unnecessarily hard to discover this feature.

EDIT: I think putting all "bootstrapping related" settings into the `
packages/@n8n/config/src/configs/instance-settings-loader.config.ts` also invites this file growing into unmanageable size, covering config for various domains.

geemanjs: If I read the env vars in the module - I think you should be able to read "everything to do with them" also in the module :eyes:

konstantintieber: > I think the initial goal (correct me if I'm wrong @ireneea) was to have a single location where you can see bootstrapped values?

With that reasoning we'd have to move several existing env vars to such a central place as well. E.g.  `MfaConfig`, `MultiMainSetupConfig`. How do these fit into that idea?

geemanjs: We kind of already have that though via the `GlobalConfig` object

ireneea: For this phase I think it makes more sense to keep them in the same file.

This should not be treated as the final solution for the settings loading, but the initial step to unblock things. The final solution should come from this [RFC](https://www.notion.so/n8n/Manage-instances-settings-at-scale-3335b6e0c94f80439901c7d41e8c3668?source=copy_link)

konstantintieber: Yes, which does not define all env vars itself, but has nested properties.
For SSO bootstrapping configuration, it would be much less thinking effort for me to go look in [GlobalConfig.ssoConfig](https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/config/src/index.ts#L200) to find these settings.

packages/cli/src/instance-settings-loader/loaders/oidc.instance-settings-loader.ts
```diff
@@ -0,0 +1,110 @@
...
+		return this.instanceSettingsLoaderConfig.ssoManagedByEnv;
+	}
+
+	async run(): Promise<'created' | 'skipped'> {
+		if (!this.instanceSettingsLoaderConfig.ssoManagedByEnv) return 'skipped';
+
+		const result = ssoEnvSchema.safeParse(this.instanceSettingsLoaderConfig);
+
+		if (!result.success) {
+			throw new OperationalError(result.error.issues[0].message);
```
ireneea: Is this the right type of error we should be throwing?

packages/cli/src/instance-settings-loader/loaders/oidc.instance-settings-loader.ts
```diff
@@ -0,0 +1,110 @@
...
+	) {
+		this.logger = this.logger.scoped('instance-settings-loader');
+	}
+
+	isConfiguredByEnv(): boolean {
+		return this.instanceSettingsLoaderConfig.ssoManagedByEnv;
+	}
+
+	async run(): Promise<'created' | 'skipped'> {
+		if (!this.instanceSettingsLoaderConfig.ssoManagedByEnv) return 'skipped';
```
ireneea: Would add a warning log if `ssoManagedByEnv` is false but some of the other settings are set. 

This would help understand miss-configuration issues. Typically "I've set the env vars but they are not being picked up"

packages/@n8n/config/src/configs/instance-settings-loader.config.ts
konstantintieber: It must be that I'm not too deep enough into the project overall.
I won't veto putting these env vars inside `instance-settings-loader.config.ts`. Their names look good anyway, which is the part that is the hardest to change later.

packages/cli/src/instance-settings-loader/loaders/oidc.instance-settings-loader.ts
geemanjs: Good question - it's what we throw in the `InstanceOwner` setting. We could always create a more specific error? `InstanceBootstrappingError?

ireneea (commented):
I think we can leave frontend, and readonly validations out of scope for this PR. The new env vars can just be treated as overrides on startup which is all we need for the current project. 

The fully fledged support for `Manage instances settings at scale` final implementation might look different.

packages/cli/src/instance-settings-loader/loaders/owner.instance-settings-loader.ts
```diff
@@ -67,7 +68,7 @@ export class OwnerInstanceSettingsLoader {
 		const result = ownerEnvSchema.safeParse(this.instanceSettingsLoaderConfig);
 
 		if (!result.success) {
-			throw new OperationalError(result.error.issues[0].message);
+			throw new InstanceBootstrappingError(result.error.issues[0].message);
```
ireneea, line 67: Nice one, I didn't realise there was the same issue in this loader. Thanks for fixing this as well

alielkhateeb:
> I think we can leave frontend, and readonly validations out of scope for this PR. The new env vars can just be treated as overrides on startup which is all we need for the current project.

@ireneea I still personally think making the UI read-only would remove a lot of confusion. Imagine spinning up an instance, the override happens, then someone modifies the instance owner credentials in the UI. Later, when n8n updates to a newer version and restarts, those credentials would get overridden again. Making it clear in the UI that these credentials are managed via env vars and cannot be edited there feels important to have within the scope of this PR to avoid confusing users.

That’s at least how I see it.

ireneea (dismissed):
Code looks good, also tested locally and works as expected 👍

konstantintieber (dismissed):

ireneea (dismissed):

konstantintieber (approved):

ireneea (approved):

PR #28189, opened 2026-04-08
feat(Moonshot Kimi Node): Add new node

DawidMyslak:
## Summary

[**Loom Recording**](https://www.loom.com/share/52ba867a40054de3b8466512d1d9ce6b)

Add a standalone Moonshot Kimi vendor node with two resources:

- **Text > Message a Model** - Full-featured chat with system prompts, multi-turn conversation, tool calling, built-in web search (`$web_search`), thinking mode, JSON response format, and image attachments (binary)
- **Image > Analyze Image** - Dedicated vision operation for image analysis via binary input

Key implementation details:
- Dynamic model selection via `/models` endpoint (resource locator)
- Token usage tracking for telemetry (`accumulateTokenUsage`)
- Tool calling loop with cancellation support and configurable max iterations
- Thinking mode toggle with explicit `thinking: { type: 'disabled' }` sent by default on all requests to prevent models like `kimi-k2.5` from silently consuming tokens on reasoning
- Conditional parameter sending (temperature, penalties) to avoid model-specific restrictions
- Binary-only image input (Moonshot API does not support URL-formatted images, only base64)
- Reuses `moonshotApi` credentials and logo from the chat model sub-node

https://linear.app/n8n/issue/NODE-4761

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/nodes-langchain/nodes/vendors/Moonshot/actions/versionDescription.ts
```diff
@@ -0,0 +1,72 @@
...
+import { NodeConnectionTypes, type INodeTypeDescription } from 'n8n-workflow';
+
+import * as image from './image';
+import * as text from './text';
+
+export const versionDescription: INodeTypeDescription = {
+	displayName: 'Moonshot Kimi',
+	name: 'moonshot',
+	icon: { light: 'file:moonshot.dark.svg', dark: 'file:moonshot.svg' },
+	group: ['transform'],
```
michael-radency, line 11: nit
this is kind of confusing, should we renamed them based on mode?

packages/@n8n/nodes-langchain/nodes/vendors/Moonshot/actions/image/analyze.operation.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export async function execute(this: IExecuteFunctions, i: number): Promise<INodeExecutionData[]> {
+	const model = this.getNodeParameter('modelId', i, '', { extractValue: true }) as string;
+	const text = this.getNodeParameter('text', i, '') as string;
+	const simplify = this.getNodeParameter('simplify', i, true) as boolean;
+	const options = this.getNodeParameter('options', i, {}) as { maxTokens?: number };
+
+	const content: ContentBlock[] = [];
+	const binaryPropertyNames = this.getNodeParameter('binaryPropertyName', i, 'data') as string;
+	for (const binaryPropertyName of splitByComma(binaryPropertyNames)) {
```
michael-radency: it should also support `binaryData` input from expressions, `splitByComma(binaryPropertyNames)` - this breaks it, we need to check for type and support string, array of strings, `binaryData`, array of `binaryData`

packages/@n8n/nodes-langchain/nodes/vendors/Moonshot/actions/text/message.operation.ts
```diff
@@ -0,0 +1,471 @@
...
+
+async function addAttachmentsToMessages(
+	this: IExecuteFunctions,
+	i: number,
+	messages: ChatMessage[],
+) {
+	const binaryPropertyNames = this.getNodeParameter('binaryPropertyName', i, 'data') as string;
+
+	const content: ContentBlock[] = [];
+	for (const binaryPropertyName of splitByComma(binaryPropertyNames)) {
```
michael-radency: same here

michael-radency (changes requested):
@DawidMyslak looks good 👍 
but we need to fix `binaryData` support

packages/@n8n/nodes-langchain/nodes/vendors/Moonshot/actions/versionDescription.ts
DawidMyslak: Can I double check what is confusing? I see 3 lines highlighted. 

Found this pattern in our examples:
```
		name: 'exampleChatMemory',
		icon: { light: 'file:example.svg', dark: 'file:example.dark.svg' },
		group: ['transform'],
```

DawidMyslak: Ah!
`icon: { light: 'file:moonshot.dark.svg', dark: 'file:moonshot.svg' },` 🤦

packages/@n8n/nodes-langchain/nodes/vendors/Moonshot/actions/image/analyze.operation.ts
DawidMyslak: I based this on the Anthropic and OpenAI nodes, which handle `binaryPropertyName` the same way (string splitting with `splitByComma`). Does that mean those nodes have the same issue? Do you have a reference to a node that handles it the way you're describing? Happy to align.

michael-radency (approved):
LGTM

PR #28199, opened 2026-04-08
feat(editor): Improve data table node

nikhilkuria:
## Summary

- Rename the "Update" table operation to "Rename" in the Data Table node UI to better reflect its purpose (only renames a table)
- Add a new "Clear" table operation that deletes all rows from a data table while retaining the table structure and ID
- Expose the clear operation via the public API as DELETE /data-tables/{dataTableId}/rows/clear

https://linear.app/n8n/issue/ADO-5041/feature-improve-data-table-node

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/data-table/data-table.service.ts
```diff
@@ -512,6 +512,19 @@ export class DataTableService {
 		return result;
 	}
 
+	async clearRows(dataTableId: string, projectId: string): Promise<{ deletedCount: number }> {
```
dariacodes, line 550: Could you add tests for this method to `data-table.service.integration.test.ts`?

dariacodes (commented):
Looks good! Added one comment about missing tests

packages/cli/src/modules/data-table/data-table.service.ts
nikhilkuria: Done 👍 almost 3 months late

dariacodes (approved):

PR #28203, opened 2026-04-08
fix(core): Unblock builder agent when correction arrives during HITL confirmation (no-changelog)

aalises:
## Summary

Fixes a deadlock where the AI builder agent gets stuck forever when a user sends a correction while the builder is waiting for HITL confirmation (e.g. `setup-credentials`).

**Root cause:** `drainCorrections()` is only called inside the stream chunk loop (`for await (const chunk of activeStream)`). Once the stream ends and the builder hits `await waitForConfirmation()`, no more chunks are produced, so corrections queued via `correct-background-task` are never processed... the builder blocks forever.

**Fix:** Add a `waitForCorrection` signal that races against `waitForConfirmation` in `resumable-stream-executor.ts`. When a correction arrives first, auto-confirm the suspended tool call with override data so the builder can resume and incorporate the correction.

- https://www.notion.so/n8n/Stuck-after-asking-for-correction-on-the-wf-33c5b6e0c94f8064a56bd218cc42e165

### DIFF
--- a/packages/@n8n/instance-ai/src/runtime/__tests__/background-task-manager.test.ts
+++ b/packages/@n8n/instance-ai/src/runtime/__tests__/background-task-manager.test.ts
@@ -209,6 +209,68 @@ describe('BackgroundTaskManager', () => {
 
 			expect(drainedCorrections[0]).toEqual([]);
 		});
+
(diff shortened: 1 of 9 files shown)


Cadiac (dismissed):
Nice find!

Cadiac (approved):

PR #28204, opened 2026-04-08
feat(core): Add audit logs for webhook waiting and resuming

konstantintieber:
## Summary

Demo video: https://www.loom.com/share/c5dbd17e14c44788877bbd4cdd83ad15

Disclaimer: lots of code was written by Claude. I thoroughly reviewed and amended the code.

Extends the audit log with two new events to cover "Send & Wait" and "Human in the Loop" node lifecycle — gaps identified in the existing audit trail.

- `n8n.audit.workflow.waiting` — emitted when an execution enters a waiting state..
- `n8n.audit.workflow.resumed` — emitted when a waiting execution is resumed. Includes node context and `resumeSource` (webhook) to distinguish how the execution was unblocked, plus `responseAt` timestamp. However: as of today, we only emit these two events for the webhook source.

The implementation designed to be extendable for more wait-use-cases in the future.

FYI: all types of nodes that have this waiting mechanism (whether it's "human in the loop" or "wait"), use the `putExecutionToWait()` API of our execution engine. This `putExecutionToWait` will set `runData.status = 'waiting'`. Then in the `workflowExecuteAfter` hook we amended inside `execution-lifecycle-hooks.ts`, there is now a new event being emitted for this `waiting` case.

### Testing locally

Helpful guide for testing locally: https://github.com/n8n-io/n8n/blob/29bd48acbd6d52d93b9f0424db3c9a9a7c2a80bf/packages/cli/test/integration/eventbus/README-manual-test-syslog.md

https://linear.app/n8n/issue/LIGO-378/be-add-audit-logs-for-send-and-waithuman-in-the-loop-nodes

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/execution-lifecycle/execution-lifecycle-hooks.ts
```diff
@@ -144,7 +145,23 @@ function hookFunctionsWorkflowEvents(hooks: ExecutionLifecycleHooks, userId?: st
 		eventService.emit('workflow-pre-execute', { executionId, data: workflowData, mode });
 	});
 	hooks.addHandler('workflowExecuteAfter', function (runData) {
-		if (runData.status === 'waiting') return;
+		if (runData.status === 'waiting') {
```
geemanjs, line 158: Probably not a now thing but we should look at a schema for passing some of this through. 

On `workflow-pre-execute` we have the whole workflow in `data`
On `execution-waiting` we have some parts of the workflow in specific fields
On `execution-resumed` the same parts as waiting
On `workflow-post-execute` we have the whole workflow in `workflow`


Any thoughts on which one we should go for? 

I think maybe we should use `workflow` and `node` on them?

konstantintieber: Wow now that I look into the different kinds of payloads we log per event I also would love to have a schema.
It also seems that including `workflowName` is not something other events do.
I think including just the `workflowId` is good enough.

EDIT: or rather, it's only logged on 4 of them, while all of them include the workflowId.

konstantintieber: Also I need to look into the `waitTill` property in that payload. I want to verify that it works.

packages/cli/src/events/maps/relay.event-map.ts
```diff
@@ -463,6 +463,30 @@ export type RelayEventMap = {
...
+		executionId: string;
+		workflowId: string;
+		workflowName: string;
+		nodeName: string;
+		nodeId?: string;
+		nodeType?: string;
+		/**
+		 * A value of null means waiting indefinitely (no timeout).
+		 */
+		waitTill: Date | null;
```
konstantintieber: TODO: remove `workflowName` from payload and manually test `waitTill` works the way it is designed to.
Also `null` value for `waitTill` doesn't really mean "wait indefinitely", it just means that we're waiting for something that's not a timer, e.g. a webhook request.
Will update that.

packages/cli/src/webhooks/waiting-webhooks.ts
```diff
@@ -323,6 +345,8 @@ export class WaitingWebhooks implements IWebhookManager {
 
 			const runExecutionData = execution.data;
 
+			this.emitExecutionResumedEvent(execution, executionId, lastNodeExecuted);
+
```
alielkhateeb: Just confirming, we are emitting an "execution resumed" event regardless if the execution succeeds or fails. We don't care we just want to see "Execution resumed" when it resumes, right?

konstantintieber: yes, we only care about the fact that it stopped waiting. There are existing events that are emitted for failure/success

packages/cli/src/events/maps/relay.event-map.ts
konstantintieber: Update:
- reduced the payload to the bare minimum here: https://github.com/n8n-io/n8n/pull/28204/commits/f1eb670d57ebcb30b482456e25c991b3baaef634
- verified waitTill is working as expected and improved docs [here](https://github.com/n8n-io/n8n/pull/28204/commits/3fb2607cf6b03c769c68e72059aa06b6c3d697d5) and [here](https://github.com/n8n-io/n8n/pull/28204/commits/17c1a479a11825b73978ed84cb96ab702373bb7e)

packages/cli/src/execution-lifecycle/execution-lifecycle-hooks.ts
```diff
@@ -155,7 +155,23 @@ function hookFunctionsWorkflowEvents(
...
+			const { executionId, workflowData: workflow } = this;
+			const lastNodeName = runData.data.resultData.lastNodeExecuted;
+			const lastNodeTaskData = lastNodeName
+				? runData.data.resultData.runData[lastNodeName]
+				: undefined;
+			const latestTask = lastNodeTaskData?.at(-1);
```
konstantintieber, line 164: Forgot to add unit tests for this logic. Coming in a sec

konstantintieber: Added in https://github.com/n8n-io/n8n/pull/28204/commits/2298cc819d6c2c8060f9573001e6c02fede209eb

geemanjs (approved):

PR #28206, opened 2026-04-08
build: patch avsc@5.7.6 to fix buffer.SlowBuffer removal in Node.js 22+

### LINKED ISSUE #28207: Bug: pnpm build fails on Node.js 22+ (Apple Silicon / ARM) due to removed buffer.SlowBuffer in avsc@5.7.6
## Bug Description

Running `pnpm build` fails on Node.js 22+ (including Apple Silicon M1/M2/M3 machines) with a misleading error:

```
ApplicationError: Class could not be found. Please check if the class is named correctly. { className: 'Kafka' }
```

The real cause is that `avsc@5.7.6` (a transitive dependency via `@kafkajs/confluent-schema-registry` → Kafka node) calls `new buffer.SlowBuffer()` at module load time. `buffer.SlowBuffer` was **removed in Node.js 22+**, which causes the `n8n-generate-metadata` VM isolation step to catch the `TypeError` and surface it as a misleading "class not found" error.

This affects **all contributors** trying to build locally on Node.js 22+, despite `package.json` declaring `engines: { node: ">=22.16.0" }`.

## Steps to Reproduce

1. Install Node.js 22.x (e.g. via nvm: `nvm install 22 && nvm use 22`)
2. Clone the repo and run `pnpm install`
3. Run `pnpm build`

## Expected behavior

Build completes successfully.

## Actual behavior

`n8n-nodes-base:build` fails:

```
ApplicationError: Class could not be found. Please check if the class is named correctly. { className: 'Kafka' }
```

Stack trace points to `loadClassInIsolation` catching a `TypeError: SlowBuffer is not a constructor` from `avsc/lib/types.js`.

## Environment

- **OS**: macOS (Apple Silicon / ARM64) — also reproducible on any platform with Node.js 22+
- **Node.js**: 22.x (any patch version)
- **pnpm**: 10.x

## Root Cause

`avsc@5.7.6` → `lib/types.js` line ~22:

```js
var SlowBuffer = buffer.SlowBuffer; // TypeError on Node 22+
```

`buffer.SlowBuffer` was removed in Node.js 22. The fix is already in `avsc@5.7.9` which uses `Buffer.allocUnsafeSlow` instead.

## Fix

PR #28206 adds a pnpm patch for `avsc@5.7.6` that backports this one-line fix:

```diff
-var SlowBuffer = buffer.SlowBuffer;
+var SlowBuffer = buffer.SlowBuffer || function SlowBuffer(size) { return Buffer.allocUnsafeSlow(size); };
```

## Workaround

Until the patch lands, you can manually patch the installed file:

```bash
node_modules/.pnpm/avsc@5.7.6/node_modules/avsc/lib/types.js
# Change line ~22 from:
#   var SlowBuffer = buffer.SlowBuffer;
# To:
#   var SlowBuffer = buffer.SlowBuffer || function SlowBuffer(size) { return Buffer.allocUnsafeSlow(size); };
```

hannasdev:
## Summary

Fixes #28207.

`avsc@5.7.6` (a transitive dependency via `@kafkajs/confluent-schema-registry` → Kafka node) calls `new buffer.SlowBuffer()` at module load time. `buffer.SlowBuffer` was [removed in Node.js 22+](https://nodejs.org/en/blog/release/v22.0.0), causing the `n8n-generate-metadata` step to crash with a misleading error:

```
ApplicationError: Class could not be found. Please check if the class is named correctly. { className: 'Kafka' }
```

This makes the entire `n8n-nodes-base` build fail for **all contributors on Node.js 22+**, including anyone on Apple Silicon (M1/M2/M3), despite `package.json` declaring `engines: { node: ">=22.16.0" }`.

### Fix

A pnpm patch for `avsc@5.7.6` that backports the fix already shipped in `avsc@5.7.9`: uses `Buffer.allocUnsafeSlow` as a fallback when `buffer.SlowBuffer` is unavailable.

```diff
-var SlowBuffer = buffer.SlowBuffer;
+// buffer.SlowBuffer was removed in Node.js 22+; fall back to Buffer.allocUnsafeSlow
+var SlowBuffer = buffer.SlowBuffer || function SlowBuffer(size) { return Buffer.allocUnsafeSlow(size); };
```

### Alternatives considered

- **Upgrade `avsc`**: `avsc@5.7.9` has this fix natively, but upgrading is a larger change and may introduce other compatibility issues with `@kafkajs/confluent-schema-registry`.
- **Upgrade `@kafkajs/confluent-schema-registry`**: Depends on upstream adoption of a newer `avsc`; not yet available.

### Related issue

`sqlite3@5.1.7` uses N-API (`prebuild-install -r napi`), so its binaries are forward-compatible across Node versions — a binary built for napi 3 or 6 runs on Node 22, 24, and 25 alike. However, the `postinstall` download can silently fail during `pnpm install` (e.g. network interruption or pnpm skipping lifecycle scripts), leaving no binary at all. If that happens, `prebuild-install -r napi` must be run manually from the sqlite3 package directory. Adding a `.nvmrc` would not fix this, but ensuring `pnpm install` lifecycle scripts complete successfully would.

## Review / Merge checklist
- [x] PR title and summary are descriptive. (no-changelog)
- [ ] Docs updated or follow-up ticket created. (`CONTRIBUTING.md` says Node 24+ but `package.json` engines requires `>=22.16.0` — minor inconsistency worth a follow-up)
- [ ] Tests included. (n/a — no meaningful unit test for a removed Node.js API in a transitive dependency; validation is the build passing on Node 22)


Matsuuu:
Hi @hannasdev ! Thank you for pointing this out and providing a fix for this.

It seems that we've overridden the `avsc` version on master with the version `5.7.9` [see here](https://github.com/n8n-io/n8n/commit/98534c6db9889d2d8c013359c4dc10f58f0c7c59#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R168).

Could you verify against the current master if the issue still persists, or if to override has already resolved the same issue this patch aims to resolve?

I will also try it out on my end but would love another pair of eyes / machine to try on 😸

Matsuuu:
Checking back: From my experience, this has been resolved on master via the override. But I won't close this PR before I've gotten validation from your side too

Matsuuu:
Resolved here https://github.com/n8n-io/n8n/commit/98534c6db9889d2d8c013359c4dc10f58f0c7c59#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R168

PR #28217, opened 2026-04-08
fix(core): Propagate formidable parse errors in Form Trigger

### LINKED ISSUE #28069: Form Trigger silently discards file uploads when formidable errors, producing misleading "No audio files found" instead of the real error
### Bug Description

When a file upload via the **Form Trigger** node causes formidable (the multipart parser) to error internally, the error is silently discarded. The trigger continues with empty binary data and the workflow proceeds to a downstream node that throws a confusing, unrelated error — in this case `"No uploaded audio files were found"` — instead of surfacing the actual upload failure.

The silent discard happens because the `_err` parameter in `createMultiFormDataParser`'s formidable callback is never checked:

```typescript
// packages/cli/src/webhooks/webhook-form-data.ts
form.parse(req, async (_err, data, files) => {
    // _err is ignored — any formidable error silently produces empty files
    normalizeFormData(data);
    normalizeFormData(files);
    resolve({ data, files });
});
```

When formidable errors, `files` is `undefined`. `normalizeFormData(undefined)` silently no-ops because JavaScript `for...in` over `null`/`undefined` is a no-op. The Form Trigger then emits an item with no binary data attached, and any downstream code that expects the uploaded file fails with a misleading error.

### To Reproduce

1. Set up a **Form Trigger** node with a file upload field.
2. Submit the form with a large file (tested: **174 MB**).
3. Observe the Form Trigger completes near-instantly (measured: **< 1 ms** in execution event log).
4. Observe downstream nodes receive an item with **empty binary data** (`input.binary === {}`).
5. No error is reported by the Form Trigger itself — the failure is silent.

**The same workflow processes small files successfully** (verified working up to ~10 MB in the same environment). Large files that **previously worked fine before upgrading to 2.13.4** now fail silently. The 174 MB file referenced above was confirmed working on a prior n8n version and broke after updating to 2.13.4 — suggesting a regression in how the Form Trigger handles large multipart uploads in this release.

### Expected behavior

When formidable errors during multipart parsing, the Form Trigger should **fail with the actual error** from formidable rather than silently resolving with empty data. The `_err` parameter should be propagated:

```typescript
form.parse(req, async (err, data, files) => {
    if (err) {
        reject(err);  // or throw — surface the real reason
        return;
    }
    normalizeFormData(data);
    normalizeFormData(files);
    resolve({ data, files });
});
```

This would give users a meaningful error message (e.g. a connection timeout, a stream error) instead of a confusing downstream failure.

### Debug Info

Execution event log for the failed upload (execution `24351`):

```
n8n.workflow.started       13:03:39.781   executionId: 24351
n8n.node.started           13:03:39.782   Form Trigger
n8n.node.finished          13:03:39.782   Form Trigger          ← completes in < 1ms
n8n.node.started           13:03:39.782   Normalize & Split Uploads (Code node)
n8n.runner.task.requested  13:03:39.784
n8n.node.finished          13:03:39.803   Normalize & Split Uploads
n8n.workflow.failed        13:03:39.804   "No uploaded audio files were found. [line 17]"
```

A genuine 174 MB multipart upload cannot complete in < 1 ms — the Form Trigger's near-instant completion confirms formidable errored immediately without writing any data. The binary data storage directory for execution `24351` contains no files, confirming zero binary data was stored.

Confirmed via source inspection on the running container:

```
# packages/cli/dist/webhooks/webhook-form-data.js (n8n 2.13.4)
const createMultiFormDataParser = (maxFormDataSizeInMb) => {
    return async function parseMultipartFormData(req) {
        const form = formidable({
            multiples: true,
            encoding: encoding,
            maxFileSize: maxFormDataSizeInMb * 1024 * 1024,
        });
        return await new Promise((resolve) => {
            form.parse(req, async (_err, data, files) => {
                normalizeFormData(data);
                normalizeFormData(files);
                resolve({ data, files });   // _err never checked
            });
        });
    };
};
```

### Operating System

Ubuntu 24.04.3 LTS (Noble Numbat) — DigitalOcean Droplet

### n8n Version

2.13.4

### Node.js Version

v24.13.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

jeanibarz:
## Summary

The multipart form data parser in `createMultiFormDataParser` silently discards errors from formidable. The callback's error parameter is prefixed with `_` and never checked:

```typescript
form.parse(req, async (_err, data, files) => {
    // _err is ignored — any formidable error silently produces empty files
    normalizeFormData(data);
    normalizeFormData(files);
    resolve({ data, files });
});
```

When formidable encounters an error (e.g. file exceeding `maxFileSize`, corrupted multipart stream, or any internal parse failure), `data` and `files` are empty/undefined. The promise resolves with empty objects, causing the Form Trigger to emit items with no binary data. Downstream nodes then fail with misleading errors (e.g. "No uploaded audio files were found") instead of surfacing the actual upload failure.

This PR checks the error parameter and rejects the promise when formidable reports an error, so the real failure reason reaches the user.

**What changed:**
- `webhook-form-data.ts`: Check formidable's error callback; reject instead of silently resolving
- `webhook-form-data.test.ts`: Updated the "file too large" test to expect rejection; added a second test verifying the error contains useful info (message + HTTP 413 status)

### Verification

I verified the fix by running real HTTP multipart requests through formidable (v3.5.4) with a file size exceeding `maxFileSize`:

**Before (old code — silent swallow):**
```
Had error from formidable: true
Error message: options.maxTotalFileSize (1024 bytes) exceeded, received 2048 bytes of file data
Data keys: []
File keys: []
=> BUG: Error silently swallowed. User gets empty data.
```

Formidable *does* report the error in its callback, but the old code ignores it and resolves with empty objects. The Form Trigger then emits an item with no binary data, and downstream nodes throw confusing, unrelated errors.

**After (new code — error propagated):**
```
Error surfaced: options.maxTotalFileSize (1024 bytes) exceeded, received 2048 bytes of file data
HTTP code: 413
=> FIX: User sees the actual error instead of empty data.
```

The error now rejects the promise, propagates through `parseRequestBody` → `executeWebhook`'s outer catch (line 927 of `webhook-helpers.ts`), and surfaces to the user as an `OperationalError` with the real formidable error as the cause.

**Success path (no regression):**
```
Data keys: [ 'name' ]
File keys: [ 'file' ]
=> Success path still works correctly.
```

Files under the limit still parse and resolve normally.

Fixes #28069

- Fixes #28069

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/webhooks/__tests__/webhook-form-data.test.ts
+++ b/packages/cli/src/webhooks/__tests__/webhook-form-data.test.ts
@@ -164,20 +164,36 @@ describe('webhook-form-data', () => {
 			testServer.assertHasBeenCalled();
 		});
 
-		it('should ignore file that is too large', async () => {
+		it('should reject with an error when file exceeds maxFileSize', async () => {
 			const oneByteInMb = 1 / 1024 / 1024;
 			const parseFn = createMultiFormDataParser(oneByteInMb);
 
 			await testServer
 				.sendRequestToHandler(async (req) => {
-					const parsedData = await parseFn(req);
-
-					expect(parsedData).toStrictEqual({
-						data: {},
-						files: {},
-					});
+					await expect(parseFn(req)).rejects.toThrow(/maxFileSize/);
 				})
 				.attach('file', oneKbData, 'file.txt');
+
+			testServer.assertHasBeenCalled();
+		});
+
+		it('should reject with an error when total file size exceeds limit', async () => {
+			// Simulate the real-world scenario from #28069: a file upload that
+			// exceeds the configured limit should surface the actual formidable
+			// error instead of silently returning empty data.
+			const twoKbData = Buffer.alloc(2 * 1024, 'x');
+			const oneKbInMb = 1 / 1024;
+			const parseFn = createMultiFormDataParser(oneKbInMb);
+
+			await testServer
+				.sendRequestToHandler(async (req) => {
+					const rejection = parseFn(req);
+					await expect(rejection).rejects.toThrow();
+					await expect(rejection).rejects.toMatchObject({
+						httpCode: 413,
+					});
+				})
+				.attach('file', twoKbData, 'large-upload.bin');
 
 			testServer.assertHasBeenCalled();
 		});
--- a/packages/cli/src/webhooks/webhook-form-data.ts
+++ b/packages/cli/src/webhooks/webhook-form-data.ts
@@ -27,8 +27,12 @@ export const createMultiFormDataParser = (maxFormDataSizeInMb: number) => {
 			// TODO: pass a custom `fileWriteStreamHandler` to create binary data files directly
 		});
 
-		return await new Promise((resolve) => {
-			form.parse(req, async (_err, data, files) => {
+		return await new Promise((resolve, reject) => {
+			form.parse(req, (error, data, files) => {
+				if (error) {
+					reject(error);
+					return;
+				}
 				normalizeFormData(data);
 				normalizeFormData(files);
 				resolve({ data, files });


BerniWittmann (approved):
🙌 Thanks for your contribution

PR #28231, opened 2026-04-09
feat(editor): Support showing full label in tooltip on hover of dropdown menu items (no-changelog)

CharlieKolb:
## Summary

Support showing a tooltip on very long names on dropdowns. Opt in currently to avoid changing existing behavior, also open to enabling this by default if preferred.

image

https://linear.app/n8n/issue/ADO-5043/improve-dropdown-menu-to-support-full-label-tooltip-on-hover

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.types.ts
+++ b/packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.types.ts
@@ -58,6 +58,8 @@ export type DropdownMenuItemProps<T = string, D = never> = {
 	subMenuOpen?: boolean;
 	/** extra data useful for rendering leading/trailing slot */
 	data?: D;
+	/** Show the full label text as a native tooltip on hover (useful for truncated labels) */
+	showFullLabelOnHover?: boolean;
 };
 
 export interface DropdownMenuProps<T = string, D = never> {
@@ -99,6 +101,8 @@ export interface DropdownMenuProps<T = string, D = never> {
 	searchDebounce?: number;
 	/** Text to be shown when no menu item exists */
 	emptyText?: string;
+	/** Show the full label text as a native tooltip on hover for all items (useful for truncated labels) */
+	showFullLabelOnHover?: boolean;
 }
 
 export interface DropdownMenuEmits<T = string> {
--- a/packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.types.ts
+++ b/packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.types.ts
@@ -58,8 +58,6 @@ export type DropdownMenuItemProps<T = string, D = never> = {
 	subMenuOpen?: boolean;
 	/** extra data useful for rendering leading/trailing slot */
 	data?: D;
-	/** Show the full label text as a native tooltip on hover (useful for truncated labels) */
-	showFullLabelOnHover?: boolean;
 };
 
 export interface DropdownMenuProps<T = string, D = never> {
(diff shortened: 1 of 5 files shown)


packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
```diff
@@ -182,6 +182,7 @@ watch(
 				<slot name="item-label" :item="props" :ui="labelProps">
 					<N8nText
 						:class="$style['item-label']"
+						:title="showFullLabelOnHover && label.length >= 20 ? label : undefined"
```
CharlieKolb: This is fairly arbitrary, I'm not familiar enough with the component to know if this applies in general, but for my use case I see 19 characters if truncated with overflow.

If there's a css-only solution that would be preferred, please let me know

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenu.types.ts
```diff
@@ -58,6 +58,8 @@ export type DropdownMenuItemProps<T = string, D = never> = {
 	subMenuOpen?: boolean;
 	/** extra data useful for rendering leading/trailing slot */
 	data?: D;
+	/** Show the full label text as a native tooltip on hover (useful for truncated labels) */
+	showFullLabelOnHover?: boolean;
```
MiloradFilipovic: I feel like showing tooltips for truncated labels should be the default behavior, not a configuration option. So I would suggest removing this prop.

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
MiloradFilipovic: Not that I know of :/
The simples thing would be to [compare scroll and client width](https://stackoverflow.com/a/10017343), but we can also introduce a library like [pretext](https://github.com/chenglou/pretext) that seems like an overkill but could come in handy in other places in the project.

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
```diff
@@ -282,6 +284,7 @@ watch(
 			<slot name="item-label" :item="props" :ui="labelProps">
 				<N8nText
 					:class="$style['item-label']"
+					:title="showFullLabelOnHover && label.length >= 20 ? label : undefined"
```
MiloradFilipovic: This seems to be repeated, should we introduce a `computed`?
```
 const titleAttr = computed(() =>
   props.showFullLabelOnHover && props.label.length >= 20 ? props.label : undefined
);
```

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
MiloradFilipovic: Nit: Any reason for using `title` over `N8nToolTip`?

MiloradFilipovic (commented):
Works nicely, I left a few comments so take a look what makes sense

packages/frontend/@n8n/design-system/src/v2/components/DropdownMenu/DropdownMenuItem.vue
CharlieKolb: Just cause I kept it optional - will try that 👌

CharlieKolb: I'm tempted to stick with the title on the text here 😅 Moving this risks breaking a use case which hardcodes a css override also technically, so I'd rather not touch it anyway

image

MiloradFilipovic: Alright, makes sense, it was a nit after all. I just see we are mixing those a bit all over the place but we should consult the design team for some guidelines. Good to go for this one

MiloradFilipovic (approved):
Thanks!

PR #28234, opened 2026-04-09
chore(instance-ai): Disable working memory and remove related code (no-changelog)

OlegIvaniv:
## Summary

Disables working memory in instance AI and removes all related infrastructure across the stack.

Working memory is turned off in the Mastra Memory config. The entire tracing layer around it is gone: `working-memory-tracing.ts`, the `internal_state` grouping wrapper in LangSmith tracing, `isInternalStateTool`, synthetic tool trace names for `updateWorkingMemory`, and `traceWorkingMemoryContext` calls that wrapped every `agent.stream()` invocation in the runtime, orchestration tools (delegate, plan-with-agent, data-table-agent, build-workflow-agent), and the resumable stream executor.

On the frontend, the memory panel (`InstanceAiMemoryPanel.vue`), its Pinia store, the brain toggle button in the header, and the memory API functions are all removed. The debug panel no longer shows a working memory section. The corresponding i18n keys are cleaned up.

On the backend, the `GET /memory/:threadId` and `PUT /memory/:threadId` endpoints are removed from the controller along with `getWorkingMemory`/`updateWorkingMemory` on the memory service. `getThreadContext` no longer returns a `workingMemory` field. The `InstanceAiUpdateMemoryRequest` DTO and the `workingMemory` field on `InstanceAiThreadContextResponse` are removed from api-types.

The "Working Memory" instruction sections are removed from both builder prompt variants.

The storage-layer `workingMemory` column on the resource entity and the TypeORM adapter methods are intentionally kept, they implement the Mastra storage interface contract and are harmless when the feature is disabled.

## Review / Merge checklist

- [ ] Tests included. 


aalises (dismissed):

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiDebugPanel.vue
```diff
@@ -308,11 +308,8 @@ onMounted(() => {
 					</div>
 					<template v-else-if="debugStore.threadContext">
 						<div :class="$style.contextSection">
-							<div :class="$style.contextLabel">
-								{{ i18n.baseText('instanceAi.debug.threads.workingMemory') }}
-							</div>
 							<pre :class="$style.contextContent">{{
-								debugStore.threadContext.workingMemory || '(empty)'
+								JSON.stringify(debugStore.threadContext, null, 2)
```
aalises: `threadContext` now only contains `{ threadId }`. this section renders meaningless data with no label. We could remove this section... (threadId is in the URL)

packages/cli/src/modules/instance-ai/instance-ai-memory.service.ts
```diff
@@ -146,26 +117,10 @@ export class InstanceAiMemoryService {
 	}
 
 	async getThreadContext(
-		userId: string,
+		_userId: string,
 		threadId: string,
 	): Promise<InstanceAiThreadContextResponse> {
```
aalises: This method now returns `{ threadId }`, which is data the caller already has. The endpoint, auth checks, and ownership verification still run for a no-op. We could remove this endpoint then, not useful anymore

aalises (approved):

PR #28239, opened 2026-04-09
fix(editor): Address high-severity issues in parity of Instance AI setup wizard (no-changelog)

aalises:
## Summary

Closes 5 parity gaps between the Instance AI `SetupWorkflow` wizard and the old AI builder setup panel. The wizard now handles credential testing, parameter validation, credential propagation, NDV edits, and AI agent node grouping correctly.

**NDV edit persistence:** `buildNodeParameters()` now falls back to `workflowsStore` node parameters, so edits made via the Node Detail View for nested/complex parameters (collection, fixedCollection, resourceMapper, etc.) are no longer silently dropped on Apply.

**Live parameter validation & field discovery:** Replaced static `card.hasParamIssues` with a live `cardHasParamWork()` helper backed by `trackedParamNames` — a sticky set that grows as `getNodeParametersIssues()` discovers new conditional required fields. Added `liveEscalatedCredTypes` to dynamically split shared credential cards into per-node cards when any node gains parameter work at runtime.

**Credential re-testing on selection change:** Added `testCredentialInBackground()` with store-derived state. `getEffectiveCredTestResult()` trusts the credential store first and only falls back to backend-provided results when the selection is unchanged. Testable credentials now require a positive test result for card completion. Removed the `credTestOverrides` mutable map entirely.

**Group-level credential state:** Replaced per-card `selections` ref with `credGroupSelections` keyed by credential group key. All credential mutations (select, deselect, delete, create, skip) go through `setCredentialForGroup()` / `clearCredentialForGroup()`, which sync credentials to all workflow store nodes in the group and trigger background testing.

**AI parent/subnode grouping:** Added `displayCards` computed that groups parent nodes with AI sub-nodes (connected via non-Main connections). Navigation and rendering use `displayCards`; apply and partial-apply logic stays on flat leaf cards. Each group section gets its own `ExpressionContextProvider` for correct expression resolution. Added `activeCredentialTarget` for targeted credential creation within grouped card sections.

- https://linear.app/n8n/issue/AI-2338
- https://linear.app/n8n/issue/AI-2339
- https://linear.app/n8n/issue/AI-2340
- https://linear.app/n8n/issue/AI-2341
- https://linear.app/n8n/issue/AI-2342

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAiWorkflowSetup.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+	'filter',
+	'assignmentCollection',
+]);
+
+// ---------------------------------------------------------------------------
+// Pure functions
+// ---------------------------------------------------------------------------
+
+export function credGroupKey(req: InstanceAiWorkflowSetupNode): string {
+	const credType = req.credentialType!;
```
mutdmour: would be better to do stricter type validation instead of `!`

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useCredentialGroupSelection.ts
mutdmour, line 1: Would be great to add some unit tests

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useCredentialTesting.ts
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			await credentialsStore.testCredential({
+				id: credentialId,
+				name: credentialName,
+				type: credentialType,
+				data: data as ICredentialDataDecryptedObject,
+			});
+		} catch {
+			// Test failure is tracked in the store as a side effect
```
mutdmour: Not really.  Two failure modes in testCredentialInBackground, both swallowed by the catch {}:
                                                                                                                                                                                                                               
  1. credentialsStore.testCredential() throws — credential gets stuck as 'pending' in the store (pre-existing bug in credentials.store.ts:414–427 — no catch/finally to clean up). The double-guard then blocks retries. Fix:  
  add catch in testCredential to set 'error' (fixes all callers).                                                                                                                                                              
  2. credentialsStore.getCredentialData() throws — no entry is written to the store at all. The card shows a spinner forever since getCredTestIcon returns 'spinner' for testable types with no result. Retries aren't blocked,
   but nothing triggers one — the user has no way to act. Fix: catch the getCredentialData failure and write 'error' to the store so the card shows a warning icon instead of an eternal spinner:                              
   
  } catch {                                                                                                                                                                                                                    
    if (credentialId) {                                                                                                                                                                                                        
      credentialsStore.credentialTestResults.set(credentialId, 'error');
    }                                                                                                                                                                                                                          
  }                                                                                                                                                                                                                          

  Neither issue is introduced by this PR — the catch {} pattern and the missing finally in testCredential are pre-existing. But this PR's automatic background testing on mount and selection change makes both more likely to 
  surface in practice.

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupCards.ts
```diff
@@ -0,0 +1,383 @@
...
+	// ---------------------------------------------------------------------------
+
+	const displayCards = computed((): DisplayCard[] => {
+		const directSubnodes = new Map<string, Set<string>>();
+		const connectionsByDest = workflowsStore.connectionsByDestinationNode;
+		for (const [destName, conns] of Object.entries(connectionsByDest)) {
+			if (!conns || typeof conns !== 'object') continue;
+			for (const connType of Object.keys(conns)) {
+				if (connType === NodeConnectionTypes.Main) continue;
+				const groups = (conns as Record<string, Array<Array<{ node: string }>>>)[connType];
```
mutdmour: we should validate the types here better instead of casting.

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAiWorkflowSetup.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export function toNodeUi(setupNode: InstanceAiWorkflowSetupNode): INodeUi {
+	return {
+		id: setupNode.node.id,
+		name: setupNode.node.name,
+		type: setupNode.node.type,
+		typeVersion: setupNode.node.typeVersion,
+		position: setupNode.node.position,
+		parameters: setupNode.node.parameters as INodeUi['parameters'],
+		credentials: setupNode.node.credentials as INodeUi['credentials'],
```
mutdmour: What if it's `undefined`? We should handle the types better

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAiWorkflowSetup.utils.ts
```diff
@@ -0,0 +1,104 @@
...
+	} as INodeUi;
```
mutdmour: `satisfies` would work better here

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -102,12 +102,34 @@ export async function buildSetupRequests(
...
+				.filter((n): n is string => n !== undefined);
+		}
+	}
+
+	// Dynamic credential resolution for nodes that use genericCredentialType
+	// or predefinedCredentialType (e.g. HTTP Request). The credential type name
+	// is stored in the node parameters rather than the description's credentials array.
+	if (credentialTypes.length === 0) {
+		const authentication = parameters.authentication as string | undefined;
+		if (authentication === 'genericCredentialType' && parameters.genericAuthType) {
```
mutdmour: would be better to validate the types here
```suggestion
		const authentication = parameters.authentication
		if (authentication === 'genericCredentialType' && typeof parameters.genericAuthType === 'string') {
```

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupActions.ts
```diff
@@ -0,0 +1,318 @@
...
+	function onCredentialSelected(
+		card: SetupCard,
+		updateInfo: { properties: { credentials?: Record<string, unknown> } },
+	) {
+		if (!card.credentialType) return;
+		const credentialData = updateInfo.properties.credentials?.[card.credentialType];
+		const credentialId =
+			typeof credentialData === 'string'
+				? undefined
+				: (credentialData as { id?: string } | undefined)?.id;
```
mutdmour: would be better to check the types first

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupActions.ts
```diff
@@ -0,0 +1,318 @@
...
+			return;
+		}
+
+		const { promise, cancel } = waitForToolResult(deps.requestId.value);
+		cancelApplyWait = cancel;
+		const toolResult = await promise;
+		cancelApplyWait = null;
+
+		if (toolResult) {
+			const error = toolResult.error as string | undefined;
```
mutdmour: could be more typesafe here.

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupActions.ts
```diff
@@ -0,0 +1,318 @@
...
+
+		isApplying.value = false;
+
+		if (toolResult && (toolResult.success as boolean)) {
+			applyServerResultToCanvas(toolResult);
+			isSubmitted.value = true;
+			isPartial.value = (toolResult.partial as boolean) ?? false;
+			deps.store.resolveConfirmation(deps.requestId.value, 'approved');
+		} else if (toolResult) {
+			applyError.value = (toolResult.error as string) ?? 'Apply failed';
```
mutdmour: could be more typesafe here

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupActions.ts
```diff
@@ -0,0 +1,318 @@
...
+		requestId: string,
+		timeoutMs = 60_000,
+	): { promise: Promise<Record<string, unknown> | null>; cancel: () => void } {
+		let stopWatch: (() => void) | null = null;
+		let timeoutId: ReturnType<typeof setTimeout> | null = null;
+
+		const promise = new Promise<Record<string, unknown> | null>((resolve) => {
+			const existing = deps.store.findToolCallByRequestId(requestId);
+			if (existing?.result !== undefined) {
+				resolve(existing.result as Record<string, unknown>);
```
mutdmour: could be more typesafe here

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -102,12 +102,34 @@ export async function buildSetupRequests(
...
+	if (credentialTypes.length === 0) {
 		const nodeCredTypes = node.credentials ? Object.keys(node.credentials) : [];
 		if (nodeCredTypes.length > 0) {
 			credentialTypes = nodeCredTypes;
-		} else if (nodeDesc?.credentials?.[0]?.name) {
-			credentialTypes = [nodeDesc.credentials[0].name];
+		} else if (nodeDesc?.credentials) {
+			// Only include credentials whose displayOptions match the current parameters
+			credentialTypes = nodeDesc.credentials
+				.filter((c: { name?: string; displayOptions?: unknown }) => !c.displayOptions)
```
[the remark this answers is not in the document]
mutdmour: cubic might have a point here. we have the nodes and their parameters in this function `buildSetupRequests`. for each node with displayOptions we can evaluate against the parameters. 
though not sure how important of a problem this is and how common it's.

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -102,12 +102,34 @@ export async function buildSetupRequests(
...
+				.map((c: { name?: string }) => c.name)
+				.filter((n): n is string => n !== undefined);
+		}
+	}
+
+	// Dynamic credential resolution for nodes that use genericCredentialType
```
mutdmour, line 132: there's some duplication with `instance-ai.adapter.service.ts` that could be improved.

mutdmour (commented):
Great work. Please leave add some unit tests to cover all the changes. Thanks

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupCards.ts
```diff
@@ -159,12 +159,17 @@ export function useSetupCards(
 			if (!conns || typeof conns !== 'object') continue;
 			for (const connType of Object.keys(conns)) {
 				if (connType === NodeConnectionTypes.Main) continue;
-				const groups = (conns as Record<string, Array<Array<{ node: string }>>>)[connType];
+				const groups = (conns as Record<string, unknown>)[connType];
```
mutdmour: types could still be better here. can use typeguards

mutdmour (commented):
LGTM. Thanks for addressing the feedback.

packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupCards.ts
aalises: Fixed! adjusted some other types

mutdmour (approved):

PR #28240, opened 2026-04-09
feat(core): Enable credential creation per project in public API 

sandra0503:
## Summary

Adds support for specifying a `projectId` when creating credentials via the public REST API (`POST /api/v1/credentials`). Without this, credentials always land in the user's personal project, blocking automated workflows that need to create credentials in team projects.

  **Changes:**
  - Extends `POST /api/v1/credentials` request body to accept an optional `projectId` field
  - Rafctoring:  delegates to CredentialService to unify the behaviour with internal API
  - Returns `404` when the specified project does not exist, and `403` when the user lacks access to it (previously both returned `400`)

https://www.loom.com/share/a835a8f3cf644c05b8fdf8b7d5ba957e

  Closes [LIGO-418](https://linear.app/n8n/issue/LIGO-418)

## Review / Merge checklist
- [x] Tests included. 


packages/cli/src/services/project.service.ee.ts
```diff
@@ -564,7 +564,10 @@ export class ProjectService {
 		};
 
 		if (!hasGlobalScope(user, scopes, { mode: 'allOf' })) {
-			const projectRoles = await this.roleService.rolesWithScope('project', scopes);
+			// Use the same EntityManager as the project lookup (including when callers pass a
```
sandra0503, line 567: I had postgres tests failing and this fixed it, I am not sure if this was a local problem?

packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
```diff
@@ -87,57 +82,17 @@ export async function getSharedCredentials(
...
-
-		await transactionManager.save<SharedCredentials>(newSharedCredential);
-
-		return savedCredential;
-	});
-
-	await Container.get(ExternalHooks).run('credentials.create', [encryptedData]);
+	const { scopes: _scopes, ...credential } = await Container.get(
+		CredentialsService,
+	).createUnmanagedCredential({ ...payload, projectId: payload.projectId ?? undefined }, user);
```
alielkhateeb, line 95: I think there may be a regression here (the refactor is nice overall, so please sanity-check in case I’m missing context).

Previously, `saveCredential` fetched the user’s personal project and used that ID for validation/sharing:
- Personal project lookup and usage:
https://github.com/n8n-io/n8n/blob/7399af34c975253917d06ff9f64b61b2fedf1101/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts#L110-L114

After the refactor, we now delegate to `createUnmanagedCredential` with `projectId ?? undefined`:

- New delegation path:
https://github.com/n8n-io/n8n/blob/b1eb9b8babd33dd6a7efdb2e7616e69f533dbb48/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts#L95

Inside CredentialsService.createCredential(), checkCredentialData() is called with opts.projectId ?? '':
- Empty fallback path:
https://github.com/n8n-io/n8n/blob/b1eb9b8babd33dd6a7efdb2e7616e69f533dbb48/packages/cli/src/credentials/credentials.service.ts#L1228

So if `projectId` is omitted in public API create credential, this will pass an empty project id into credential checks instead of the user’s personal project id (which used to be explicit in the old flow).

Sorry for the long comment 😂

alielkhateeb: Does it make sense to fix it on the type level, making `CreateCredentialsOptions` ALWAYS require projectId?

So instead of 
https://github.com/n8n-io/n8n/blob/bddb0836b9a1cc531a1cef7cb0ea4bd2977bd20b/packages/cli/src/credentials/credentials.service.ts#L77-L79

Make it something like:
```ts
type CreateCredentialOptions = Omit<CreateCredentialDto, 'projectId'> & {
	isManaged: boolean;
	projectId: string;
};
```

sandra0503: Hi Ali, thanks for checking! That's a good catch and actually a slightly behaviour wrong in the existing API. We should resolve the personal project earlier - not in save(). I will refactor that

alielkhateeb (dismissed):
Great work 🔥🚀 
I don't see this as a blocker tbh, so I approved anyway :)
I think it would be useful to have an integration test for that public API which has no project ID but has external secret reference

```ts
	test('should return 400 for external secret reference without projectId when permissions are missing', async () => {
		const payload = {
			name: 'test credential',
			type: 'githubApi',
			data: {
				accessToken: '={{ $secrets.myApiKey }}',
				user: 'test',
				server: 'testServer',
			},
		};

		const response = await authMemberAgent.post('/credentials').send(payload);

		expect(response.statusCode).toBe(400);
		expect(response.body.message).toBe(
			'Lacking permissions to reference external secrets in credentials',
		);
	});
```

AI wrote this 😂 what do you think?

sandra0503:
> ```ts
> 	test('should return 400 for external secret reference without projectId when permissions are missing', async () => {
> 		const payload = {
> 			name: 'test credential',
> 			type: 'githubApi',
> 			data: {
> 				accessToken: '={{ $secrets.myApiKey }}',
> 				user: 'test',
> 				server: 'testServer',
> 			},
> 		};
> 
> 		const response = await authMemberAgent.post('/credentials').send(payload);
> 
> 		expect(response.statusCode).toBe(400);
> 		expect(response.body.message).toBe(
> 			'Lacking permissions to reference external secrets in credentials',
> 		);
> 	});
> ```

Yes, let's add this in!

alielkhateeb (approved):

PR #28243, opened 2026-04-09
fix(ai-builder): Allow non-admin users to access Instance AI preferences (no-changelog)

aalises:
## Summary

The Instance AI settings page was entirely gated behind `instanceAi:manage` scope (admin-only), preventing non-admin users from configuring their own preferences. This PR lowers the access requirement so all users with `instanceAi:message` scope can reach the settings page, while admin-only sections remain hidden for non-admin users.

### Settings visibility by role

| Setting | Admin | All Users |
|---------|:-----:|:---------:|
| Enable/Disable AI toggle | ✅ | |
| Model credential selection | ✅ | ✅ |
| Model name | ✅ | ✅ |
| Local gateway toggle | ✅ | ✅ |
| Sandbox (provider, image, timeout) | ✅ | |
| Memory (last messages, embedder, top-k) | ✅ | |
| Search credential | ✅ | |
| Advanced (sub-agent steps, browser MCP, MCP servers) | ✅ | |
| Permissions (18 HITL modes) | ✅ | |

### Changes

- **Route guard**: lowered from `instanceAi:manage` to `instanceAi:message` so non-admin users can access the page
- **Sidebar item**: visibility now checks `instanceAi:message` instead of `instanceAi:manage`
- **Settings view**: conditionally renders admin-only sections (enable toggle, permissions, sandbox, memory, search, advanced) based on `instanceAi:manage` scope
- **Settings store**: skips admin-only API calls (`fetchSettings`, `fetchServiceCredentials`) for non-admin users to avoid 403 errors
- **Section components**: wires up existing but previously unused components (`ModelSection`, `LocalGatewaySection`, `SandboxSection`, `MemorySection`, `SearchSection`, `AdvancedSection`)
- **Save/Cancel footer**: added for draft-based changes from section components

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/instanceAi/views/SettingsInstanceAiView.vue
+++ b/packages/frontend/editor-ui/src/features/ai/instanceAi/views/SettingsInstanceAiView.vue
@@ -1,27 +1,24 @@
 <script lang="ts" setup>
-import { onMounted, ref, computed, watch } from 'vue';
-import { N8nHeading, N8nIcon, N8nOption, N8nSelect, N8nText } from '@n8n/design-system';
+import { onMounted, computed } from 'vue';
+import { N8nButton, N8nHeading, N8nIcon, N8nOption, N8nSelect, N8nText } from '@n8n/design-system';
 import { ElSwitch } from 'element-plus';
 import { useI18n } from '@n8n/i18n';
 import { useDocumentTitle } from '@/app/composables/useDocumentTitle';
-import { useRootStore } from '@n8n/stores/useRootStore';
-import { useToast } from '@/app/composables/useToast';
-import { fetchSettings, updateSettings } from '../instanceAi.settings.api';
-import type {
-	InstanceAiAdminSettingsResponse,
-	InstanceAiPermissions,
-	InstanceAiPermissionMode,
-} from '@n8n/api-types';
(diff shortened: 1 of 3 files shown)


packages/frontend/editor-ui/src/features/ai/instanceAi/views/SettingsInstanceAiView.vue
```diff
@@ -126,81 +75,134 @@ function getPermissionValue(key: keyof InstanceAiPermissions): InstanceAiPermiss
...
-				<div :class="$style.settingsRow">
-					<div :class="$style.settingsRowLeft">
-						<span :class="$style.settingsRowLabel">
-							{{ i18n.baseText('settings.n8nAgent.enable.label') }}
-						</span>
-						<span :class="$style.settingsRowDescription">
-							{{ i18n.baseText('settings.n8nAgent.enable.description') }}
-						</span>
+		<template v-else>
+			<template v-if="isAdmin">
```
Cadiac, line 83: Would it be useful to still see these, but have them disabled? That way users who need these changed could tell their admins what to change..

Cadiac (approved):
I'm fine to merge this as is, but we could (at some point) consider displaying all the settings but disabled for non-admins, at least that's how I like seeing settings pages :D

PR #28244, opened 2026-04-09
fix(Google Drive Node): Fix infinite pagination loop in v1 API request

BerniWittmann:
## Summary

Fixes a memory leak in the Google Drive trigger node caused by an infinite pagination loop in the v1 `googleApiRequestAllItems` function.

The function's `do-while` loop checked `responseData.nextPageToken` to decide whether to continue paginating, but never assigned `query.pageToken = responseData.nextPageToken` to advance to the next page. When the Google Drive API returned more than 100 results (one page), the same first page was re-fetched indefinitely, pushing ~100 items into an array per iteration until OOM.

This bug only affects the **v1 transport** used by `GoogleDriveTrigger.node.ts`. The v2 transport (used by the main Google Drive node) already had the correct implementation. Every other Google API pagination function in the codebase (Gmail, Calendar, Sheets, YouTube, etc.) also includes this assignment — the v1 Drive function was the only one missing it.

**Impact:** Instances with active Google Drive trigger workflows polling accounts with >100 modified files experienced unbounded memory growth and repeated OOM crashes, even when no workflow executions were triggered.

### How to test

1. Set up a Google Drive trigger node in `specificFile` mode polling a Google account with many files
2. Verify the trigger polls correctly without memory growth
3. Run the new unit test: `cd packages/nodes-base && pnpm test nodes/Google/Drive/test/v1/GenericFunctions.test.ts`

https://linear.app/n8n/issue/NODE-4565

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/Google/Drive/test/v1/GenericFunctions.test.ts
+++ b/packages/nodes-base/nodes/Google/Drive/test/v1/GenericFunctions.test.ts
@@ -1,0 +1,67 @@
+import { mockDeep } from 'jest-mock-extended';
+import type { IDataObject, IPollFunctions } from 'n8n-workflow';
+
+import { googleApiRequestAllItems } from '../../v1/GenericFunctions';
+
+describe('googleApiRequestAllItems', () => {
+	let mockContext: jest.Mocked<IPollFunctions>;
+
+	beforeEach(() => {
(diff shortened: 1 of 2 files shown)


DawidMyslak (approved):
LGTM 👍

PR #28251, opened 2026-04-09
feat(core): Add parse-file tool for structured attachments (no-changelog)

aalises:
## Summary

Add a native `parse-file` tool to Instance AI that exposes structured file attachments (CSV, TSV, JSON) through a secure, paginated API instead of injecting raw file bytes into the model prompt.
Screenshot 2026-04-10 at 13 12 47

### Key changes:
- **Attachment routing**: Structured attachments (csv/tsv/json) are replaced with a compact manifest in the prompt text; non-structured attachments keep the existing multimodal `file` path
- **`parse-file` tool**: Thin wrapper over a parser utility with format detection, column normalization, type inference, pagination, and output budgeting
- **Attachment-only messages**: `message` may now be empty when `attachments` is non-empty — synthesizes a stub directing the agent to inspect the first parseable file
- **Data-table agent**: `parse-file` added to tool subset, max steps increased 15 → 35, prompt updated with import flow (preview → create table → paginate + insert)
- **Security guardrails**: 512 KB decoded-size cap, 50 column max, 2000 cell budget, 40000 char budget, 5000 char cell limit, dangerous key rejection (`__proto__`, `constructor`, `prototype`)
- **Trace redaction**: Raw structured attachment data is excluded from prompt-build trace outputs


aalises:
> ## ⚠️ PR exceeds size limit (1,511 lines added)
> This PR adds **1,511 lines**, exceeding the 1,000-line limit.
> 
> Large PRs are harder to review and increase the risk of bugs going unnoticed. Please consider:
> 
> * Breaking this into smaller, logically separate PRs
> * Moving unrelated changes to a follow-up PR
> 
> If the size is genuinely justified (e.g. generated code, large migrations, test fixtures), a maintainer can override by commenting `/size-limit-override` and then pushing a new commit or re-running this check.

Screenshot 2026-04-10 at 13 16 26
😢

packages/@n8n/api-types/src/schemas/instance-ai.schema.ts
```diff
@@ -545,17 +545,17 @@ export type InstanceAiFilesystemResponse = z.infer<typeof instanceAiFilesystemRe
 // ---------------------------------------------------------------------------
 
 const instanceAiAttachmentSchema = z.object({
-	data: z.string(),
-	mimeType: z.string(),
-	fileName: z.string(),
+	data: z.string().max(700_000), // ~512 KB decoded + base64 overhead
```
Cadiac, line 548: We'll probably want to do something else than pass these as base64 later on, currently this means that we can't pass bigger attachments at all right if they don't fit into 700k chars of base64?

but as for this PR this will probably do, but I have a feeling we'll get feedback about large pdfs not working etc...

packages/@n8n/instance-ai/src/parsers/__tests__/structured-file-parser.test.ts
```diff
@@ -0,0 +1,544 @@
...
+
+	it('rejects malformed base64', () => {
+		const att: AttachmentInfo = {
+			data: '!!!not-base64!!!',
+			mimeType: 'text/csv',
+			fileName: 'test.csv',
+		};
+		// Buffer.from doesn't throw on bad base64, so this should still work as a valid CSV
+		// The real protection is the size check
+		expect(() => parseStructuredFile(att, 0, {})).not.toThrow();
```
Cadiac, line 344: ...so it doesn't reject on malformed base64? 👀

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -1604,34 +1612,64 @@ export class InstanceAiService {
...
-										})),
-									],
-								},
-							]
-						: fullMessage;
+				// Classify attachments: structured (csv/tsv/json) go through parse-file,
+				// non-structured keep the existing multimodal file path.
+				if (attachments && attachments.length > 0) {
+					const classified = classifyAttachments(attachments);
+					const nonStructured = attachments.filter((a) => !isStructuredAttachment(a));
```
Cadiac, line 1623: We're trusting on the filename / mimetype provided by user here, meaning they could trick this by sending some structured content with a mimetype for nonstructured (like foobar.jpg), and it would fully go to the LLM as is without pagination etc right?

I can't think of how this could really be abused though, so doesn't matter?

packages/@n8n/instance-ai/src/parsers/structured-file-parser.ts
```diff
@@ -0,0 +1,542 @@
...
+
+export function parseStructuredFile(
+	attachment: AttachmentInfo,
+	attachmentIndex: number,
+	input: ParseFileInput,
+): ParseFileOutput {
+	// Decode base64
+	let decoded: Buffer;
+	try {
+		decoded = Buffer.from(attachment.data, 'base64');
```
Cadiac, line 318: second decode here

packages/@n8n/instance-ai/src/parsers/structured-file-parser.ts
```diff
@@ -0,0 +1,542 @@
...
+	return attachments.map((att, index) => {
+		const format = detectFormat(att.fileName, att.mimeType);
+		if (!format) {
+			return { original: att, index, parseable: false };
+		}
+
+		// Check size
+		let decoded: Buffer;
+		try {
+			decoded = Buffer.from(att.data, 'base64');
```
Cadiac: We decode the buffer here and at `parseStructuredFile`, which I is a bit of a waste but maybe not substantial enough to warrant rewriting this?

Cadiac: probably not worth addressing now

Cadiac (dismissed):
left some optional comments but seems good

packages/@n8n/api-types/src/schemas/instance-ai.schema.ts
aalises: Yeah it won't scale for larger files. For now the 512KB decoded cap keeps it bounded, but we'll want to move to a reference-based approach (e.g. presigned URLs, not sure what service we use for storage or temp storage) when we tackle larger attachments...

packages/@n8n/instance-ai/src/parsers/__tests__/structured-file-parser.test.ts
aalises: Renamed the test to `'does not throw on malformed base64'` since `Buffer.from` silently ignores invalid base64 chars.

packages/@n8n/instance-ai/src/parsers/structured-file-parser.ts
aalises: Fixed. `classifyAttachments` now estimates decoded size from the base64 string length (`Math.ceil(len * 3 / 4)`) instead of fully decoding. 

💪 The actual decode only happens once in `parseStructuredFile` when the tool is invoked.

packages/@n8n/instance-ai/src/parsers/structured-file-parser.ts
aalises: Addressed!!. Classification now uses a base64 length estimate, so the full decode only happens once.

aalises:
> left some optional comments but seems good

addressed!

Cadiac (approved):
👍

PR #28254, opened 2026-04-09
fix(core): Release LS trace clients on run finalization (no-changelog)

OlegIvaniv:
## Summary

The module-level traceClients Map in langsmith-tracing.ts was only cleaned when a root run finished successfully via finishRun(). If a run was aborted, timed out, or the finishRun HTTP call failed, the Client and its entire RunTree hierarchy stayed in the Map indefinitely.

Add releaseTraceClient() that unconditionally deletes the entry, and call it from:
- finalizeMessageTraceRoot (normal message run completion)
- finalizeDetachedTraceRun (background task completion)
- deleteTraceContextsForThread (safety net on thread deletion)

### Measured impact (parallel workflow building benchmark: 3 tabs × 3 rounds)

| Metric | Before | After | 
|---|---|---|
| Heap leak (post-cleanup − baseline) | +152 MB | +80 MB |
| Round-over-round growth (R2→R3) | +75 MB | +6 MB |
| RunTree objects after cleanup | 127 | near 0 (expected) |

The remaining ~80 MB delta is one-time lazy loading (node type catalog, source maps, Zod schemas) that loads on first workflow build and stays flat across subsequent rounds.

## Review / Merge checklist

- [ ] Tests included. 


Cadiac (approved):
👍 👍

PR #28273, opened 2026-04-09
fix(ai-builder): Unify post-build credential setup into single setup-workflow flow

aalises:
## Summary
- Remove credential finalization (`setup-credentials` + `apply-workflow-credentials`) from the builder sub-agent — the builder now only builds and verifies
- Update workflow loop guidance to point to `setup-workflow` instead of the old `setup-credentials` → `apply-workflow-credentials` chain
- Add explicit deferral guardrail in system prompt: when `setup-workflow` returns `deferred: true`, do not retry with a different setup tool
- Tighten system prompt to clarify `setup-credentials` is only for standalone credential creation outside workflow context

Fixes duplicate setup prompts where users saw first a credential picker (`setup-credentials`), then the full setup panel (`setup-workflow`) after building a workflow.


Cadiac:
> ## ⚠️ Ownership acknowledgement required
> Please add or check the following item in your PR description before this can be merged:
> 
> ```
> - [x] I have seen this code, I have run this code, and I take responsibility for this code.
> ```

what :D

Cadiac (approved):
Seems like this makes sense, but didn't test locally

PR #28279, opened 2026-04-10
fix(ai-builder): Handle data table name conflict gracefully instead of looping (no-changelog)

JoseBra:
## Summary
- Fix `create-data-table` tool throwing `DataTableNameConflictError` when the table already exists during the builder's rebuild cycle, causing the agent to waste iterations retrying
- Return `denied: true` with guidance to use the existing table, which the agent handles gracefully
- Add unit tests for the create-data-table tool (8 tests covering all tool states)

## Before / After
| | Before | After |
|---|---|---|
| Build | 1/3 succeeded (parallel runs) | 3/3 succeeded |

## Context
Reported independently by product team — a user spent 45 min trying to build a workflow with a data table before giving up. The data table conflict during the builder's rebuild cycle wastes agent iterations and context, making it more likely to hit proxy timeouts.

## Related Linear ticket
https://linear.app/n8n/issue/AI-2358

## Review checklist

## Test plan
- [x] Unit tests: 8/8 passing (create, conflict handling, suspension, permission blocked, non-conflict errors)
- [x] Typecheck clean
- [x] Manual before/after testing: without fix 1/3 builds succeed, with fix 3/3 builds succeed


packages/@n8n/instance-ai/src/tools/data-tables/create-data-table.tool.ts
```diff
@@ -28,6 +28,20 @@ export const createDataTableResumeSchema = z.object({
...
 });
 
+/**
+ * Check if an error (or its cause chain) is a DataTableNameConflictError.
+ * The error class lives in packages/cli so we can't import it directly —
+ * instead we match on the class name through the cause chain.
+ */
+function isNameConflictError(error: unknown): boolean {
+	let current: unknown = error;
+	while (current instanceof Error) {
```
OlegIvaniv, line 38: That's neat!

OlegIvaniv (approved):

PR #28297, opened 2026-04-10
fix(core): Remove LocalFilesystemProvider, require computer use for filesystem access (no-changelog)

aalises:
## Summary

Remove `LocalFilesystemProvider` entirely from the instance AI module. Filesystem access now exclusively requires the `@n8n/computer-use` gateway daemon.

**Why:** The `LocalFilesystemProvider` gave the AI agent direct read access to the server's disk via Node.js `fs/promises`. Without explicit `N8N_INSTANCE_AI_FILESYSTEM_PATH` configuration, the agent could read **any file** the n8n process can access (`/etc/passwd`, `.env`, database files, etc.). The gateway model is inherently more secure — the user explicitly chooses which directory to share when starting the daemon.

**What's removed:**
- `LocalFilesystemProvider` class and its test (CLI backend)
- `N8N_INSTANCE_AI_FILESYSTEM_PATH` env var
- `InstanceAiFilesystemService` interface and 4 static filesystem tool files (dead code after removing the only implementation)
- `localGateway` and `localGatewayFallbackDirectory` from the API settings response and frontend stores
- Local filesystem status block from the settings UI (yellow dot + fallback directory)

**What stays (untouched):**
- Local Gateway / `@n8n/computer-use` — all SSE bridge infrastructure
- `localGatewayDisabled` admin/user preference (controls whether the gateway is available)
- `N8N_INSTANCE_AI_GATEWAY_API_KEY` env var
- Local sandbox provider (`'local'` in `SandboxProvider`) — code execution, unrelated

### ⚠️ Naming clarification for reviewers

Several removed fields have confusing names that do NOT mean what they sound like:

| Removed field | What it actually meant | What it sounds like |
|---|---|---|
| `localGateway: boolean` (API settings) | "Is the local filesystem fallback available?" | "Is the gateway connected?" |
| `localGatewayFallbackDirectory` (API settings) | Directory path for `LocalFilesystemProvider` | Something about gateway fallback |
| `isLocalGatewayEnabled` (frontend stores) | "Is the local filesystem fallback active?" | "Is the gateway enabled?" |

Gateway connection status was always tracked separately via SSE push events — these fields were exclusively about the now-removed local filesystem fallback.

https://linear.app/n8n/issue/AI-2359

## Review / Merge checklist

- [x] Tests included.


Cadiac (approved):
I'm relieved to see these tools gone ☺️ 
I think the description has the same inaccuracy as the ticket - N8N_INSTANCE_AI_FILESYSTEM_PATH being unset / empty string meant local FS wasn't available, not that it was constrained, user had to enter it some root path for it to activate - but that doesn't really matter as it's going away anyways.

I'm marking the title of this PR with no-changelog as this was never even released.

PR #28299, opened 2026-04-10
fix(core): Reduce planner workflow bias and fix data-table task routing (no-changelog)

OlegIvaniv:
## Summary

- Planner now uses data-table kind for all table operations (create, delete, modify) instead of falling back to delegate with incomplete tools
- Made blueprint data-table columns optional to support non-creation operations
- Updated accumulator to derive task title/spec from purpose when columns absent
- Added scope boundary to data-table agent to prevent cross-task side effects
- Orchestrator now handles single remaining tasks directly on replan instead of always creating a new plan via create-tasks
- Added standalone data-table operations section to best-practices guide

## Review / Merge checklist

- [ ] Tests included. 


Cadiac (approved):
🤞

PR #28309, opened 2026-04-10
fix(core): Add configurable retries and error details to S3

ivov:
https://linear.app/n8n/issue/CAT-2739


packages/core/src/binary-data/object-store/object-store.service.ee.ts
```diff
@@ -291,7 +293,12 @@ export class ObjectStoreService {
 				nextContinuationToken: response.NextContinuationToken,
 			};
 		} catch (e) {
-			throw new UnexpectedError('Request to S3 failed', { cause: e });
+			this.s3Error(e);
 		}
 	}
+
+	private s3Error(e: unknown): never {
```
tomi: nit: This method could be better named, e.g. `handleS3Error`

tomi (dismissed):
LGTM, just one nit

tomi (approved):
🚀

PR #28315, opened 2026-04-10
fix(core): Fix duplicate task request on runner defer

ivov:
## Summary

When the launcher [defers](https://github.com/n8n-io/task-runner-launcher/blob/main/docs/lifecycle.md) a task, the broker's `acceptOffer` catch block was pushing the request back into `pendingTaskRequests` but `settleTasks` never removes requests from that array in the first place, only sets `acceptInProgress = true`. This created a duplicate entry on every defer cycle.

This PR fixes the duplication and also adds a warning log when the broker has pending offers but none match a request's task type. My suspicion is after a worker restart, since only one launcher type reconnects, requests for the other type sit with no match until they time out.

https://linear.app/n8n/issue/CAT-2706

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/task-runners/task-broker/__tests__/task-broker.service.test.ts
+++ b/packages/cli/src/task-runners/task-broker/__tests__/task-broker.service.test.ts
@@ -359,6 +359,78 @@ describe('TaskBroker', () => {
 			expect(remainingRequests[0]).toEqual(request);
 		});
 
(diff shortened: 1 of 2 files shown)


ivov:
@tomi When you have a chance 🙏🏻

tomi (approved):
LGTM 🚀

PR #28317, opened 2026-04-10
feat(ai-builder): Improve sub-agent context passing with structured briefings and debriefings

aalises:
## Summary

Sub-agents (builder, research, data-table, delegate) were receiving context through ad-hoc string concatenation and returning bare `{ result: string }`. 

Each spawn site had its own copy-pasted briefing construction, the protocol forbade any diagnostic reasoning, and the orchestrator had no visibility into what the sub-agent actually did. This made retries blind, debugging difficult, and the codebase repetitive.

This PR simplifies the handover in both directions:

**Context flowing in**: A single shared `buildSubAgentBriefing()` replaces 4 copy-pasted briefing constructions across delegate, builder, research, and data-table agents. Sub-agents now receive structured context (task, conversation history, artifacts, retry history) plus awareness of what other tasks are running in parallel.

**Context flowing back**: A lightweight stream observer (`WorkSummaryAccumulator`) passively tracks tool call outcomes as they flow through the existing stream loop. The delegate tool now returns tool call counts, error counts, duration, and blockers alongside the result — so the orchestrator knows what happened, not just what the sub-agent said.

**Diagnostic reasoning**: The sub-agent protocol no longer forbids all prose. Sub-agents can now include 2-3 sentences of diagnostic context when something fails or requires a non-obvious decision. The orchestrator learns *why* something failed, enabling better retries.

### DIFF
--- a/packages/@n8n/instance-ai/src/agent/sub-agent-briefing.ts
+++ b/packages/@n8n/instance-ai/src/agent/sub-agent-briefing.ts
@@ -1,0 +1,95 @@
+import { formatPreviousAttempts, type IterationLog } from '../storage/iteration-log';
+
+// ── Types ───────────────────────────────────────────────────────────────────
+
+export interface RunningTaskSummary {
+	taskId: string;
+	role: string;
+	goal?: string;
+}
+
+export interface SubAgentBriefingInput {
+	/** The core task description. */
+	task: string;
+	/** Brief summary of the conversation so far. */
+	conversationContext?: string;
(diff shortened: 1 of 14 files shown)


OlegIvaniv (approved):

Cadiac (approved):
Seems like a nice implementation for this idea - not sure about its effectiveness, but lets try 👍

salomaoparkour:
Tks for this, @aalises ⚡

PR #28321, opened 2026-04-10
feat: Enable security policy settings via env vars

ireneea:
## Summary

Add environment variable support for security policy settings using the instance settings loader pattern. This allows MFA enforcement, personal space publishing, and personal space sharing to be pre-configured via env vars and applied on instance startup.

**New env vars:**
- `N8N_SECURITY_POLICY_MANAGED_BY_ENV` (boolean, default `false`) — master toggle; when false, all other security policy env vars are ignored
- `N8N_SECURITY_POLICY_MFA_ENFORCED` (boolean, default `false`) — enforce MFA for all users
- `N8N_SECURITY_POLICY_PERSONAL_SPACE_PUBLISHING` (boolean, default `true`) — allow personal space publishing
- `N8N_SECURITY_POLICY_PERSONAL_SPACE_SHARING` (boolean, default `true`) — allow personal space sharing

**How to test:**
1. Set `N8N_SECURITY_POLICY_MANAGED_BY_ENV=true` along with desired policy env vars
2. Start the instance
3. Verify security settings in UI match the env var values
4. Restart and confirm settings are re-applied on every startup

https://linear.app/n8n/issue/LIGO-436

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/config/src/configs/instance-settings-loader.config.ts
+++ b/packages/@n8n/config/src/configs/instance-settings-loader.config.ts
@@ -32,15 +32,15 @@ export class InstanceSettingsLoaderConfig {
 	 * On every startup the security policy will be overridden by env vars.
 	 * When false (default), security policy env vars are ignored even if set.
 	 */
-	@Env('N8N_SECURITY_POLICY_OVERRIDE')
-	securityPolicyOverride: boolean = false;
+	@Env('N8N_SECURITY_POLICY_MANAGED_BY_ENV')
+	securityPolicyManagedByEnv: boolean = false;
 
-	@Env('N8N_SECURITY_POLICY_MFA_ENFORCED')
-	securityPolicyMfaEnforced: boolean = false;
+	@Env('N8N_MFA_ENFORCED_ENABLED')
+	mfaEnforcedEnabled: boolean = false;
 
-	@Env('N8N_SECURITY_POLICY_PERSONAL_SPACE_PUBLISHING')
-	securityPolicyPersonalSpacePublishing: boolean = true;
+	@Env('N8N_PERSONAL_SPACE_PUBLISHING_ENABLED')
+	personalSpacePublishingEnabled: boolean = true;
 
-	@Env('N8N_SECURITY_POLICY_PERSONAL_SPACE_SHARING')
-	securityPolicyPersonalSpaceSharing: boolean = true;
+	@Env('N8N_PERSONAL_SPACE_SHARING_ENABLED')
+	personalSpaceSharingEnabled: boolean = true;
 }
--- a/packages/@n8n/config/src/configs/instance-settings-loader.config.ts
+++ b/packages/@n8n/config/src/configs/instance-settings-loader.config.ts
@@ -26,4 +26,21 @@ export class InstanceSettingsLoaderConfig {
 	 */
 	@Env('N8N_INSTANCE_OWNER_PASSWORD_HASH')
 	ownerPasswordHash: string = '';
+
+	/**
+	 * When true, security policy settings are managed via environment variables.
+	 * On every startup the security policy will be overridden by env vars.
+	 * When false (default), security policy env vars are ignored even if set.
+	 */
+	@Env('SECURITY_POLICY_OVERRIDE')
+	securityPolicyOverride: boolean = false;
+
+	@Env('SECURITY_POLICY_MFA_ENFORCED')
+	securityPolicyMfaEnforced: boolean = false;
+
+	@Env('SECURITY_POLICY_PERSONAL_SPACE_PUBLISHING')
+	securityPolicyPersonalSpacePublishing: boolean = true;
+
+	@Env('SECURITY_POLICY_PERSONAL_SPACE_SHARING')
+	securityPolicyPersonalSpaceSharing: boolean = true;
 }
--- a/packages/@n8n/config/src/configs/instance-settings-loader.config.ts
+++ b/packages/@n8n/config/src/configs/instance-settings-loader.config.ts
@@ -32,15 +32,15 @@ export class InstanceSettingsLoaderConfig {
 	 * On every startup the security policy will be overridden by env vars.
 	 * When false (default), security policy env vars are ignored even if set.
 	 */
-	@Env('SECURITY_POLICY_OVERRIDE')
+	@Env('N8N_SECURITY_POLICY_OVERRIDE')
 	securityPolicyOverride: boolean = false;
 
-	@Env('SECURITY_POLICY_MFA_ENFORCED')
+	@Env('N8N_SECURITY_POLICY_MFA_ENFORCED')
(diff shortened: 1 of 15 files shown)


packages/@n8n/config/src/configs/instance-settings-loader.config.ts
```diff
@@ -26,4 +26,21 @@ export class InstanceSettingsLoaderConfig {
...
+	 * On every startup the security policy will be overridden by env vars.
+	 * When false (default), security policy env vars are ignored even if set.
+	 */
+	@Env('N8N_SECURITY_POLICY_OVERRIDE')
+	securityPolicyOverride: boolean = false;
+
+	@Env('N8N_SECURITY_POLICY_MFA_ENFORCED')
+	securityPolicyMfaEnforced: boolean = false;
+
+	@Env('N8N_SECURITY_POLICY_PERSONAL_SPACE_PUBLISHING')
```
geemanjs: I think these should end with `_ENABLED` to be explicit - I also wonder if we need the `SECURITY_POLICY` parts? 

`N8N_PERSONAL_SPACE_PUBLISHING_ENABLED` feels good to me?

packages/@n8n/config/src/configs/instance-settings-loader.config.ts
```diff
@@ -26,4 +26,21 @@ export class InstanceSettingsLoaderConfig {
...
 	@Env('N8N_INSTANCE_OWNER_PASSWORD_HASH')
 	ownerPasswordHash: string = '';
+
+	/**
+	 * When true, security policy settings are managed via environment variables.
+	 * On every startup the security policy will be overridden by env vars.
+	 * When false (default), security policy env vars are ignored even if set.
+	 */
+	@Env('N8N_SECURITY_POLICY_OVERRIDE')
+	securityPolicyOverride: boolean = false;
```
geemanjs: @alielkhateeb and I discussed this at quite some length and concluded the `OVERRIDE` was confusing so settled with `MANAGED_BY_ENV`

If `MANAGED_BY_ENV=true` it is always managed by the environment never by the UI
If `MANAGED_BY_ENV=false` it is only manageable by the UI.

packages/cli/src/instance-settings-loader/loaders/security-policy.instance-settings-loader.ts
```diff
@@ -0,0 +1,60 @@
...
+		const {
+			securityPolicyOverride,
+			securityPolicyMfaEnforced,
+			securityPolicyPersonalSpacePublishing,
+			securityPolicyPersonalSpaceSharing,
+		} = this.instanceSettingsLoaderConfig;
+
+		if (!securityPolicyOverride) {
+			if (
+				securityPolicyMfaEnforced ||
```
geemanjs: It's possible here that this value could be being set to true or false (both are arguably valid configuration). It's basically if it's "different to what is the current stored value?"

packages/@n8n/config/src/configs/instance-settings-loader.config.ts
ireneea: I agree that makes sense 👍. `MANAGED_BY_ENV` is a better name. 

But just to clarify as part of this phase only the override behaviour is needed. Preventing UI changes is out of scope

ireneea: updated the name 👍

packages/@n8n/config/src/configs/instance-settings-loader.config.ts
ireneea: updated the name 👍

packages/@n8n/config/src/configs/instance-settings-loader.config.ts
```diff
@@ -26,4 +26,21 @@ export class InstanceSettingsLoaderConfig {
 	 */
 	@Env('N8N_INSTANCE_OWNER_PASSWORD_HASH')
 	ownerPasswordHash: string = '';
+
+	/**
+	 * When true, security policy settings are managed via environment variables.
+	 * On every startup the security policy will be overridden by env vars.
+	 * When false (default), security policy env vars are ignored even if set.
+	 */
+	@Env('N8N_SECURITY_POLICY_MANAGED_BY_ENV')
```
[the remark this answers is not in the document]
ireneea: This is new env var so there is no backward compatibility concern here

packages/cli/src/instance-settings-loader/loaders/security-policy.instance-settings-loader.ts
ireneea: Yeah that's a good point this is overly complicating things. I'm
 just going to switch this to a simple logger.debug instead 👍

packages/cli/src/instance-settings-loader/loaders/owner.instance-settings-loader.ts
```diff
@@ -51,14 +51,10 @@ export class OwnerInstanceSettingsLoader {
...
-		const { ownerManagedByEnv, ownerEmail, ownerPasswordHash } = this.instanceSettingsLoaderConfig;
+		const { ownerManagedByEnv } = this.instanceSettingsLoaderConfig;
 
 		if (!ownerManagedByEnv) {
-			if (ownerEmail || ownerPasswordHash) {
-				this.logger.warn(
-					'N8N_INSTANCE_OWNER_EMAIL or N8N_INSTANCE_OWNER_PASSWORD_HASH is set but N8N_INSTANCE_OWNER_MANAGED_BY_ENV is not enabled — ignoring owner env vars',
-				);
-			}
+			this.logger.debug('Owner is not managed by environment variables, skipping');
```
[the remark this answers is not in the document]
ireneea: That's intentional, the previous logic had some edge cases and and the value of the warning is not worth it. 

Users can still understand what's happening from the debug logs and take action

geemanjs (dismissed):

packages/cli/src/instance-settings-loader/__tests__/owner.instance-settings-loader.test.ts
```diff
@@ -41,31 +41,25 @@ describe('OwnerInstanceSettingsLoader', () => {
...
 
-		it('should skip and warn when ownerPasswordHash is set', async () => {
+		it('should skip when ownerPasswordHash is set', async () => {
 			const loader = createLoader({ ownerPasswordHash: validBcryptHash });
 
 			const result = await loader.run();
 
 			expect(result).toBe('skipped');
 			expect(ownershipService.setupOwner).not.toHaveBeenCalled();
-			expect(logger.warn).toHaveBeenCalledWith(
```
geemanjs, line 63: Any reason these were removed from the tests?

geemanjs: nvm - I see you removed the log below

geemanjs (approved):

PR #28323, opened 2026-04-10
 feat(core): Project based data table creation and transfer

sandra0503:
## Summary

Extends the public v1 data tables API so tables can be created in a specific project, listed with optional `projectId` filtering (with correct project-scope checks for non-admins).

**How to test**
- Manually: create data tables via public API with `projectId` in create body, list query filter, and PATCH body as documented in OpenAPI

https://www.loom.com/share/269ed04bf7d34a78ab176b7471e9840a

Closes [LIGO-416](https://linear.app/n8n/issue/LIGO-416/support-projectid-when-creating-data-table-via-the-public-api)

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/data-table/data-table.service.ts
```diff
@@ -136,6 +137,53 @@ export class DataTableService {
...
+						name,
+						projectId: targetProjectId,
+					})
+				) {
+					throw new DataTableNameConflictError(name);
+				}
+			}
+
+			const updates: { projectId?: string; name?: string } = {};
+			if (targetProjectId !== fromProjectId) updates.projectId = targetProjectId;
```
konstantintieber: Should we have validation to check whether any workflows are still referencing this data table?
Because as far as I know, workflows in this project that used to reference a data table that is now moved somewhere else, will break.

packages/cli/src/public-api/v1/handlers/data-tables/spec/paths/data-tables.yml
```diff
@@ -42,23 +44,36 @@ post:
   tags:
     - DataTable
   summary: Create a new data table
-  description: Create a new data table in your workspace.
+  description: Create a new data table in your personal project or a team project you have access to.
```
konstantintieber, line 45: In the video I see that you're sending a PATCH request, but there is no patch spec here.
Am I looking the wrong place?

packages/cli/src/modules/data-table/data-table.service.ts
```diff
@@ -136,6 +137,53 @@ export class DataTableService {
...
+					await em.existsBy(DataTable, {
+						name: table.name,
+						projectId: targetProjectId,
+					})
+				) {
+					throw new DataTableNameConflictError(table.name);
+				}
+			}
+
+			if (name !== undefined) {
```
konstantintieber: Thie name === undefined case, will this not be prevented by our openapi "required" validation already?
Or is name not marked as required there, so we need to check it here?

packages/cli/src/public-api/v1/handlers/data-tables/spec/paths/data-tables.yml
sandra0503: PATCH is defined here: https://github.com/n8n-io/n8n/blob/fe56a6583ec59ce0eadaac5f998b8325d87733f8/packages/cli/src/public-api/v1/handlers/data-tables/spec/paths/data-tables.dataTableId.yml#L25

packages/cli/src/modules/data-table/data-table.service.ts
sandra0503: The idea of this update function was that name and/or projectId can be passed in for updates! So neither of them would be required

packages/cli/src/modules/data-table/data-table.service.ts
sandra0503: That's a good question! I checked how we do it for variables and there are no checks in place whether workflows depend on a variable before updating the projectID.
We could block the request if workflows in the source project depend on a resource, but then we should apply the same rules for credentials, variables and data-tables

packages/cli/src/modules/data-table/data-table.service.ts
```diff
@@ -136,6 +137,53 @@ export class DataTableService {
...
+				if (
+					await em.existsBy(DataTable, {
+						name,
+						projectId: targetProjectId,
+					})
+				) {
+					throw new DataTableNameConflictError(name);
+				}
+			}
```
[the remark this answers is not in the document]
ireneea: Is it possible to have the an extra condition `existsBy` to exclude table matching `dataTableId`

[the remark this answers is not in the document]
ireneea: @sandra0503 what do you think of this ☝️

packages/cli/src/modules/data-table/data-table.service.ts
konstantintieber: I think it's hard to optimize this on the public PATCH API endpoint, where there is no UI involved.
If there was a dedicated `POST /data-table/{id}/transfer` endpoint (like we have for credentials) we could document explicitly that this will fail by default if the data table is still referenced, which can be disabled with an additional parameter.

I guess it'd help to understand for me: how much is the "transfer data table to other project via API a priority for our users"?
Is this something we could leave out for now still? Because removing things from public API is harder.

packages/cli/src/modules/data-table/data-table.service.ts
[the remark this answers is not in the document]
sandra0503: Sounds good!

packages/cli/src/modules/data-table/data-table.service.ts
sandra0503: SAP mentioned moving data tables as a reqirement (more good to have I believe). But it would be also consistent with variables and credentials endpoints. For workflows we have a dedicated transfer endpoint, maybe it's cleaner to have this here too as you suggest!

sandra0503: I tested this and the new requirements of dependency check and error handling would bloat this PR too much, so I will remove it from this PR. We can have it as a follow-up and double check if it's needed!

packages/cli/src/public-api/v1/handlers/data-tables/spec/paths/data-tables.yml
sandra0503: Creating a new data table is using POST

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -138,22 +159,30 @@ export = {
...
 		publicApiScope('dataTable:create'),
 		async (req: DataTableRequest.Create, res: express.Response): Promise<express.Response> => {
-			try {
-				const payload = CreateDataTableDto.safeParse(req.body);
-				if (!payload.success) {
-					return res.status(400).json({
-						message: payload.error.errors[0]?.message || 'Invalid request body',
-					});
-				}
+			const payload = PublicApiCreateDataTableDto.safeParse(req.body);
```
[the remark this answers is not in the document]
sandra0503: Errors of different types are now properly handled by middleware - no need to catch

packages/cli/src/services/project.service.ee.ts
```diff
@@ -588,7 +588,9 @@ export class ProjectService {
 
 		if (!hasGlobalScope(user, scopes, { mode: 'allOf' })) {
 			const projectRoles = await this.roleService.rolesWithScope('project', scopes);
-			where.type = 'team';
+			if (!projectIds) {
```
konstantintieber, line 592: Saw myself thinking for a bit here, what will be included in the list of projects if `projectIds` is in fact not defined.
The previous code was explicit that only `team` projects are included.
What's the types of projects that are included when `projectIds` is empty?
Only personal projects? Team and personal projects?

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -82,14 +89,28 @@ export = {
...
+					if (accessibleIds.length !== ids.length) throw new ForbiddenError();
+				}
+
+				let finalFilter: DataTableListFilter;
 				if (isGlobalOwnerOrAdmin) {
-					finalFilter = restFilter;
+					finalFilter = requestedProjectId
+						? { ...restFilter, projectId: requestedProjectId }
+						: restFilter;
+				} else if (requestedProjectId) {
```
konstantintieber: Nit code style: with there being three different if branches for what the value of `finalFilter` will be, understanding the logic for this state becomes harder to read.
Extracting this into a private method that's a pure function, or defining a pure function `getDataTableFilter` would help me here.

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -138,22 +159,30 @@ export = {
...
 
+			let projectId: string;
+			if (payload.data.projectId) {
+				const project = await Container.get(ProjectService).getProjectWithScope(
+					req.user,
+					payload.data.projectId,
+					['dataTable:create'],
+				);
+				if (!project) throw new ForbiddenError();
+				projectId = project.id;
```
konstantintieber: So this error here is about a user trying to create a data table in a project, where they lack the `dataTable:create` permission.
One more check that comes to mind for me that we can do before even, is whether a project with that ID exists.
In that case it wouldn't be forbidden, but bad request, right?

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -138,22 +159,30 @@ export = {
...
 					req.user.id,
 				);
+				projectId = project.id;
+			}
 
-				const result = await Container.get(DataTableService).createDataTable(
-					project.id,
-					payload.data,
-				);
+			const { projectId: _ignored, ...dto } = payload.data;
```
konstantintieber: This `projectId: _ignored` coupled with the `let projectId` above, makes it hard for me to follow what projectId is actually passed into the `createDataTable` method now.

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
sandra0503: Yes, we can check project existence first!

packages/cli/src/services/project.service.ee.ts
sandra0503: This is a small fix - without this guard you could not access your personal project with that function - it would only return team projects even if you directly request your personal project via ID.

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
sandra0503: This has been moved to helper functions in service file

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -138,22 +124,17 @@ export = {
...
-				const project = await Container.get(ProjectRepository).getPersonalProjectForUserOrFail(
-					req.user.id,
-				);
+			const { projectId: requestedProjectId, ...dto } = payload.data;
 
-				const result = await Container.get(DataTableService).createDataTable(
-					project.id,
-					payload.data,
-				);
+			const projectId = await resolveProjectIdForCreate(req.user, requestedProjectId);
```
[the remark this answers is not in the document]
sandra0503: As users are authenticated users this should be fine?

packages/cli/src/services/project.service.ee.ts
konstantintieber: I'm wondering would be a sementically named variable for the condition of `!projectIds`?

I'm guessing this one is wrong:
```ts
const isLookingForPersonalProject = !projectIds;
if (isLookingForPersonalProject) {
   where.type = 'team'
}
```

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -82,36 +76,28 @@ export = {
...
-						projectId: allAccessibleProjectIds,
-					};
+				if (requestedProjectId && !isGlobalOwnerOrAdmin) {
+					const ids = Array.isArray(requestedProjectId) ? requestedProjectId : [requestedProjectId];
+					const accessibleIds = await Container.get(ProjectService).getProjectIdsWithScope(
+						req.user,
+						['dataTable:listProject'],
+						ids,
+					);
+					if (accessibleIds.length !== ids.length) throw new ForbiddenError();
```
konstantintieber: While for data table creation, I think it's good to throw forbidden when passing a project id they don't have access to.
In the search case, I think we should instead not include those projectIds in the db query that fetches the data tables.

This is consistent with other search endpoints in our codebase:
- credential search ignores externalsecretstore that doesnt exist [code](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/credentials/credential-dependency.service.ts#L31)
- workflow search does an early return of empty array if project to filter by is not found [code](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/workflows/workflow.service.ts#L119)
  - they can do that there, since they don't accept a list of projectIds, but only a single one. I like that. Why do we support a list of projectIds to filter by for this new data table filter?

sandra0503: Changed it so they just get skipped! 
The filter type definition was already there and accepting an array, I didn't want to change it.

packages/cli/src/services/project.service.ee.ts
sandra0503: added an inline comment for this, hope it's more clear now

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
konstantintieber: The definition was there, but it was never used, right?
The `projectId` that was passed in was just ignored.
I think it'd be good to keep the API more simple if we see the opportunity.

packages/cli/src/services/project.service.ee.ts
konstantintieber: What if I pass a projectId that's a personal project? As admin I have access to personal projects.

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
sandra0503: Okay, we can be pretty sure that no one was using this field, right?

packages/cli/src/services/project.service.ee.ts
sandra0503: Before this change, it was setting `where.type = 'team'` in any case. Also the entire block is for non-admins 
```
if (!hasGlobalScope(user, scopes, { mode: 'allOf' })) {
```

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
konstantintieber: 100% certainty I'd say.

sandra0503: I changed to only accept one projectID. Therefore I had to adjust the dto and the filter parser. Let me know what you think!

packages/cli/src/public-api/v1/handlers/data-tables/data-tables.handler.ts
```diff
@@ -76,27 +75,24 @@ export = {
...
-					: await scopeProjectIds(req.user, requestedProjectId);
-
-				if (
-					!isGlobalOwnerOrAdmin &&
-					Array.isArray(accessibleProjectIds) &&
-					accessibleProjectIds.length === 0
-				) {
-					return res.json({ data: [], nextCursor: null });
+				if (requestedProjectId && !isGlobalOwnerOrAdmin) {
+					const accessibleIds = await Container.get(ProjectService).getProjectIdsWithScope(
```
konstantintieber: Nit: we can now use the `getProjectWithScope` method instead of `getProjectIdsWithScope`.

packages/@n8n/api-types/src/schemas/data-table.schema.ts
```diff
@@ -45,11 +45,11 @@ export type DataTableColumn = z.infer<typeof dataTableColumnSchema>;
 export type DataTableListFilter = {
 	id?: string | string[];
 	projectId?: string | string[];
-	name?: string;
+	name?: string | string[];
```
konstantintieber, line 48: Just double checking: why was this type extended to include `string[]`?
Because the implementation already supported it but the typing did not reflect this?

sandra0503: Yes, according to schema an array of strings gets accepted:
https://github.com/n8n-io/n8n/pull/28323/changes#diff-017269cc557c1a9d9f2eecc0d1894ced94c206917cec2561f90564323e104d9bR23

konstantintieber (dismissed):
Did not test it manually, since your video shows it working. Code looks good

konstantintieber (approved):

PR #28325, opened 2026-04-10
fix(core): Skip disabled Azure Key Vault secrets and handle partial fetch failures

alielkhateeb:
## Summary

- Skip secrets marked as disabled when listing Azure Key Vault secrets, avoiding unnecessary `getSecret` calls for secrets that cannot be retrieved.
- Switch from `Promise.all` to `Promise.allSettled` when fetching secret values so that a single inaccessible secret does not abort the entire reload — other secrets are still loaded and a warning is logged per failure.
- Only throw (and leave the cache unchanged) when every secret fetch fails, preserving the last known good state.

https://linear.app/n8n/issue/LIGO-449

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/providers/azure-key-vault/azure-key-vault.ts
+++ b/packages/cli/src/modules/external-secrets.ee/providers/azure-key-vault/azure-key-vault.ts
@@ -1,7 +1,7 @@
 import type { SecretClient } from '@azure/keyvault-secrets';
 import { Logger } from '@n8n/backend-common';
 import { Container } from '@n8n/di';
-import type { INodeProperties } from 'n8n-workflow';
+import { ensureError, type INodeProperties } from 'n8n-workflow';
 
 import type { AzureKeyVaultContext } from './types';
 import { DOCS_HELP_NOTICE } from '../../constants';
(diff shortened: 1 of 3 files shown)


packages/cli/test/integration/public-api/credentials.test.ts
alielkhateeb, line 1: This was just a quick formatting 😂

packages/cli/src/modules/external-secrets.ee/providers/azure-key-vault/azure-key-vault.ts
```diff
@@ -103,24 +103,38 @@ export class AzureKeyVault extends SecretsProvider {
...
+			}
+		}
 
-		this.cachedSecrets = secrets.reduce<Record<string, string>>((acc, cur) => {
-			if (cur.value === undefined) return acc;
-			acc[cur.name] = cur.value;
-			return acc;
-		}, {});
+		if (secretNames.length > 0 && Object.keys(updated).length === 0 && readErrors.length > 0) {
+			throw readErrors[0];
```
ireneea: Does it make sense to only throw the first read error?

alielkhateeb: Good question, I have given this a big thought so brace for a long comment 😂 sorry in advance!

We read vault secrets in parallel, each failure is already recorded with a `warn` log https://github.com/n8n-io/n8n/blob/83946a2c2146a02f73089fd4fb332cb39dd307e5/packages/cli/src/modules/external-secrets.ee/providers/azure-key-vault/azure-key-vault.ts#L127-L129

On the sync path, callers only handle single rejection (with 400) and generic failure responses, surfacing **one** error keeps that contract simple, without having to concatenate a huge error message for instance.

Last thing, we only throw in case every read failed AND nothing was cached, this would be a total refresh failure which in that case all failures are often (if not surely) the same root cause. Any single key succeeding will avoid throwing altogether.

That said, my other alternative was to throw something like:
```ts
throw new UnexpectedError('Could not read any secrets from Azure Key Vault', {
    cause: readErrors[0],
});
```
This is where I struggled to make a decision 😂 so if you think this would make things clearer I'm happy to do it (or do any other better ideas you have in mind really)

ireneea: I see 🤔 

Concatenation of all errors could be way to long for a large number of keys. So that's not a good option

I actually prefer your suggestion, I think it's a good middle ground between acknowledging what really happened (none of the secrets was read), and still providing insights in the error by picking the first one.     

```
throw new UnexpectedError('Could not read any secrets from Azure Key Vault', {
    cause: readErrors[0],
});
```

alielkhateeb: Makes sense 2b753d3a7cc59041f5d2d08aa10c726a1f50f36f

ireneea (approved):

PR #28328, opened 2026-04-10
feat(core): Make token exchange rate limits configurable via env vars (no-changelog)

afitzek:
## Summary

Makes rate limits on the token exchange authentication endpoints (`POST /auth/oauth/token` and `GET/POST /auth/embed`) configurable via environment variables instead of hardcoded values. Also adds a `redirectTo` parameter to the embed login endpoints so callers can specify where the user lands after authentication, and extracts the redirect URL validation logic from the SAML controller into a shared utility.

**New env vars:**
- `N8N_TOKEN_EXCHANGE_EMBED_LOGIN_PER_MINUTE` (default: `20`) — IP rate limit for `GET/POST /auth/embed`
- `N8N_TOKEN_EXCHANGE_TOKEN_EXCHANGE_PER_MINUTE` (default: `20`) — IP rate limit for `POST /auth/oauth/token`

This is part of the Phase 4 (Hardening) milestone for the OAuth 2.0 Token Exchange project. These endpoints are authentication surfaces, so configurable rate limiting allows operators to tune protection against brute-force attempts for their deployment profile without code changes.

- Rate limit values are read from `TokenExchangeConfig` at module load time via `Container.get()` at the top level, since decorator arguments must be static. This is consistent with how the config is used elsewhere.
- Extracted `validateRedirectUrl` from `SamlController` into `@/utils/validate-redirect-url` so both SAML and embed login share the same open-redirect protection.

## Related tickets

- https://linear.app/n8n/issue/IAM-474

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/@n8n/api-types/src/dto/auth/embed-login-body.dto.ts
+++ b/packages/@n8n/api-types/src/dto/auth/embed-login-body.dto.ts
@@ -4,4 +4,5 @@ import { Z } from '../../zod-class';
 
 export class EmbedLoginBodyDto extends Z.class({
 	token: z.string().min(1),
+	redirectTo: z.string().optional(),
 }) {}
--- a/packages/cli/src/modules/token-exchange/controllers/embed-auth.controller.ts
+++ b/packages/cli/src/modules/token-exchange/controllers/embed-auth.controller.ts
@@ -10,6 +10,9 @@ import { UrlService } from '@/services/url.service';
 
 import { TokenExchangeService } from '../services/token-exchange.service';
 import { TokenExchangeConfig } from '../token-exchange.config';
+import { Container } from '@n8n/di';
+
+const configService = Container.get(TokenExchangeConfig);
 
 @RestController('/auth/embed')
 export class EmbedAuthController {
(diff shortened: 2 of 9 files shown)


packages/cli/src/modules/token-exchange/controllers/embed-auth.controller.ts
```diff
@@ -7,9 +7,13 @@ import { AuthService } from '@/auth/auth.service';
 import { EventService } from '@/events/event.service';
 import { AuthlessRequest } from '@/requests';
 import { UrlService } from '@/services/url.service';
+import { validateRedirectUrl } from '@/utils/validate-redirect-url';
 
 import { TokenExchangeService } from '../services/token-exchange.service';
 import { TokenExchangeConfig } from '../token-exchange.config';
+import { Container } from '@n8n/di';
+
+const configService = Container.get(TokenExchangeConfig);
```
geemanjs, line 16: TIL you can access the container here 🤯

geemanjs (approved):

packages/@n8n/api-types/src/dto/auth/embed-login-body.dto.ts
```diff
@@ -4,4 +4,5 @@ import { Z } from '../../zod-class';
 
 export class EmbedLoginBodyDto extends Z.class({
 	token: z.string().min(1),
+	redirectTo: z.string().optional(),
```
geemanjs, line 7: Should we set a limit on the length in here?

PR #28329, opened 2026-04-10
feat: Add deployment_key table, entity, repository, and migration

BGZStephen:
## Summary

Adds the `deployment_key` table as pure DB infrastructure for the encryption key rotation epic. the `deployment_key` table is intended as a generic `key` store for various keys, including: `data_encryption`, `signing`, `fingerprint` to support the various activities within n8n.

**What's added:**
- `DeploymentKey` TypeORM entity extending `WithTimestampsAndStringId` (`id`, `createdAt`, `updatedAt` inherited)
- Columns: `type` (varchar 64), `value` (text), `algorithm` (varchar 20, nullable), `status` (varchar 20), `deprecatedAt` (datetime, nullable)
- `DeploymentKeyRepository` with `findActiveByType`, `findById`, and `findAllByType` methods
- Migration `1777000000000-CreateDeploymentKeyTable` registered for both SQLite and PostgreSQL
- Partial unique index `IDX_deployment_key_data_encryption_active` on `type` where `status = 'active' AND type = 'data_encryption'` — enforces at most one active `data_encryption` key at a time

https://linear.app/n8n/issue/IAM-484

## Review / Merge checklist

- [ ] Tests included.


packages/@n8n/db/src/migrations/common/1777000000000-CreateDeploymentKeyTable.ts
```diff
@@ -0,0 +1,25 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class CreateDeploymentKeyTable1777000000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column, createIndex } }: MigrationContext) {
+		await createTable('deployment_key').withColumns(
```
afitzek, line 5: I think the id column is missing in this table.

afitzek (dismissed):

packages/@n8n/db/src/repositories/deployment-key.repository.ts
```diff
@@ -0,0 +1,23 @@
...
+export class DeploymentKeyRepository extends Repository<DeploymentKey> {
+	constructor(dataSource: DataSource) {
+		super(DeploymentKey, dataSource.manager);
+	}
+
+	async findActiveByType(type: string): Promise<DeploymentKey | null> {
+		return await this.findOne({ where: { type, status: 'active' } });
+	}
+
+	async findById(id: string): Promise<DeploymentKey | null> {
```
guillaumejacquart: Not sure this provides much added value compared to the existing `findOneBy({ id })`.
We usually add a dedicated findById method if there is added value like decryption for the execution or something like that

packages/@n8n/db/src/migrations/common/1777000000000-CreateDeploymentKeyTable.ts
```diff
@@ -0,0 +1,58 @@
...
+
+		await createIndex(
+			'deployment_key',
+			['type'],
+			true,
+			'IDX_deployment_key_data_encryption_active',
+			"status = 'active' AND type = 'data_encryption'",
+		);
+
+		await createIndex(
```
guillaumejacquart, line 22: I'm probably missing something but why not a single index on type where status = active ?

guillaumejacquart (commented):
2 small comments

packages/@n8n/db/src/migrations/common/1777000000000-CreateDeploymentKeyTable.ts
BGZStephen: Mainly for flexibility, we have a desire to make data encryption and signing keys single active, but if we enforce single active for everything, we lose flexibility in the event we want to support other keys that maybe have multiple active

packages/@n8n/db/src/repositories/deployment-key.repository.ts
BGZStephen: Fair, let me remove

guillaumejacquart (approved):

PR #28330, opened 2026-04-10
feat: Add Telegram integration for agents

burivuhster:
## Summary

Add Telegram as a second chat platform integration for n8n agents, alongside the existing Slack integration, and refactor the integration.

**Backend — chat platform abstraction:**
- New `AgentChatIntegration` abstract base + `ChatIntegrationRegistry` (DI singleton). Each platform lives in one class under `integrations/platforms/`: `SlackIntegration`, `TelegramIntegration`. Subclasses own adapter construction, credential extraction, capability metadata, component normalization, and lifecycle hooks (`onAfterConnect`).
- `ChatIntegrationService.connect()` lost its switch/extract methods and delegates to the registry. `ComponentMapper.toCard()` calls `integration.normalizeComponents()` from the registry instead of a hardcoded `normalizeForPlatform`. `createRichInteractionTool()` reads `supportedComponents` and `description` from the registry. `AgentChatBridge` reads the `needsShortCallbackData` flag from the registry.
- Adding a new platform (Discord, Teams, …) is one `AgentChatIntegration` subclass + one `registry.register()` call in `agents.module.ts`.

**Backend — Telegram specifics:**
- Add `@chat-adapter/telegram` dependency.
- `TelegramIntegration` auto-detects webhook vs polling mode based on whether `WEBHOOK_URL` points to a public host. In webhook mode, `onAfterConnect` calls Telegram's `setWebhook` API. In polling mode (local dev with no tunnel) the adapter uses `getUpdates`.
- Generic `CallbackStore` (async, delete-on-resolve, TtlMap-backed) shortens button payloads to fit Telegram's 64-byte `callback_data` limit.
- Cache the reconstructed published runtime in `executeForChatPublished` under the same `runtimeKey(agentId, userId, integrationType)` as the draft path, so HITL button clicks on rich-interaction cards can find the runtime via `resumeForChat`. Without this, clicking a button on a published agent's card fails with "Agent is not compiled — cannot resume". Existing 30-min TTL bounds staleness after a republish.
- Thread `integrationType` through `executeForChat` / `executeForChatPublished` / `reconstructFromConfig` / `injectRuntimeDependencies` so each platform gets its own runtime with the right rich-interaction tool capabilities.
- Post-connect hook failures now shutdown the chat instance + dispose the bridge before propagating (no more leaked polling loops / adapter timers).
- Expired/missing short-callback key posts a user-facing "This action is no longer available" message instead of silently dropping the click.

**Frontend:**
- Generic integration API helpers (`connectIntegration`, `disconnectIntegration`, `getIntegrationStatus`) replace the Slack-only functions.
- `AgentIntegrationsPanel` is now data-driven over an `integrationConfigs` array — one entry per platform, same template.

**How to test:**
1. Create a Telegram bot via BotFather, add a `telegramApi` credential with the bot token.
2. Navigate to an agent → Integrations tab → connect the Telegram credential.
3. Message the bot in Telegram — agent should respond.
4. Trigger a tool that uses `rich_interaction` — buttons should render as an inline keyboard, and clicking one should resume the agent with the selected value.
5. Locally without a tunnel: connect works via polling mode (no `setWebhook` call); with `WEBHOOK_URL` pointing to a public host (e.g. ngrok), webhook mode is used.

https://linear.app/n8n/issue/AGENT-7

## Review / Merge checklist

- [ ] Tests included.


packages/cli/src/modules/agents/integrations/callback-store.ts
```diff
@@ -0,0 +1,87 @@
...
+ * This store maps 8-character hex keys to the original `{actionId, value}`
+ * so that buttons can use short IDs while the bridge can resolve them
+ * back to full payloads on action callbacks.
+ *
+ * Entries are deleted on resolve (one-time use). Unresolved entries
+ * auto-expire after the configured TTL (default: 1 hour).
+ *
+ */
+export class CallbackStore {
+	private readonly entries = new Map<
```
yehorkardash: I've added TtlMap recently, maybe you can use it here

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
```diff
@@ -120,11 +122,18 @@ export class ChatIntegrationService {
...
 			userId,
 			projectId,
+			integrationType,
 		);
 
 		// Initialize the Chat instance (connects adapters, state adapter, etc.)
 		await chat.initialize();
 
+		// Register Telegram webhook if running in webhook mode
+		if (integrationType === 'telegram' && this.getTelegramMode() === 'webhook') {
```
yehorkardash: I'm wondering if we should add `Integration` class with optional methods like `onBeforeConnect`, `onAfterConnect`, etc. to abstract this logic.
With more integrations this service will become more and more complex

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
```diff
@@ -355,6 +364,45 @@ export class ChatIntegrationService {
...
+		if (typeof token === 'string' && token) {
+			return token;
+		}
+		throw new Error(
+			'Could not extract a bot token from the Telegram credential. ' +
+				'Please ensure the credential has a valid access token from BotFather.',
+		);
+	}
+
+	private getTelegramMode(): 'webhook' | 'polling' {
```
yehorkardash: Should we make connection mode a parameter for telegram integration on frontend? Something like 'Auto', 'Polling', 'Webhook'?

packages/cli/src/modules/agents/integrations/component-mapper.ts
```diff
@@ -52,9 +52,39 @@ interface ComponentRenderContext {
...
-	 * Encodes `runId:toolCallId` in button IDs for the resume flow.
+	 * Normalize components for platforms that don't support certain types.
+	 * Converts select/radio_select → individual buttons, image → text link.
 	 */
+	private normalizeForPlatform(
+		components: SuspendComponent[],
+		platform?: string,
+	): SuspendComponent[] {
+		// Slack supports everything — no normalization needed
+		if (!platform || platform === 'slack') return components;
```
yehorkardash: I have same concerns here. Can we make it modular somehow to hide to abstract logic for specific integrations?

packages/cli/src/modules/agents/integrations/agent-chat-bridge.ts
```diff
@@ -103,7 +112,11 @@ export class AgentChatBridge {
 		private readonly logger: Logger,
 		private readonly n8nUserId: string,
 		private readonly n8nProjectId: string,
+		private readonly integrationType: string = 'slack',
```
yehorkardash: Should we have a default here? Can we make it a required parameter?

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
```diff
@@ -355,6 +364,45 @@ export class ChatIntegrationService {
...
+		const baseUrl = this.urlService.getWebhookBaseUrl();
+		const isPublic = baseUrl.startsWith('https://') && !baseUrl.includes('localhost');
+		return isPublic ? 'webhook' : 'polling';
+	}
+
+	private buildWebhookUrl(agentId: string, projectId: string, platform: string): string {
+		const base = this.urlService.getInstanceBaseUrl();
```
yehorkardash: should we use `getWebhookBaseUrl` instead?

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
```diff
@@ -120,11 +122,18 @@ export class ChatIntegrationService {
...
+			integrationType,
 		);
 
 		// Initialize the Chat instance (connects adapters, state adapter, etc.)
 		await chat.initialize();
 
+		// Register Telegram webhook if running in webhook mode
+		if (integrationType === 'telegram' && this.getTelegramMode() === 'webhook') {
+			const botToken = this.extractTelegramBotToken(decryptedData);
+			await this.registerTelegramWebhook(botToken, agentId, projectId);
```
yehorkardash: Can `chat` leak resources if `registerTelegramWebhook` throws?

packages/cli/src/modules/agents/integrations/agent-chat-bridge.ts
```diff
@@ -488,7 +523,17 @@ export class AgentChatBridge {
 	 * - `resume:{runId}:{toolCallId}:{index}` — generic per-tool resume button
 	 */
 	private async handleAction(event: ChatActionEvent): Promise<void> {
-		const { actionId, value, thread } = event;
+		let { actionId, value } = event;
+		const { thread } = event;
+
+		// Resolve short callback keys back to full action data
+		if (this.callbackStore) {
+			const resolved = await this.callbackStore.resolve(actionId);
```
yehorkardash, line 538: Would it make a sense to post a message like 'Link has expired' instead of silent noop?

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
burivuhster: I would leave this out of scope for now, we can add this later

packages/cli/src/modules/agents/integrations/platforms/telegram-integration.ts
```diff
@@ -0,0 +1,103 @@
...
+					normalized.push(c);
+			}
+		}
+		return normalized;
+	}
+
+	/** Webhook when we have a public URL, polling otherwise (local dev). */
+	private getMode(): 'webhook' | 'polling' {
+		const baseUrl = this.urlService.getWebhookBaseUrl();
+		const isPublic = baseUrl.startsWith('https://') && !baseUrl.includes('localhost');
```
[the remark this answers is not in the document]
burivuhster: Fine for P0

packages/cli/src/modules/agents/integrations/agent-chat-integration.ts
```diff
@@ -0,0 +1,88 @@
...
+	readonly needsShortCallbackData: boolean = false;
+
+	/** Build the Chat SDK adapter for this platform. */
+	abstract createAdapter(ctx: AgentChatIntegrationContext): Promise<unknown>;
+
+	/** Optional hook run AFTER `chat.initialize()`. Throwing triggers cleanup. */
+	onAfterConnect?(ctx: AgentChatIntegrationContext): Promise<void>;
+
+	/** Optional hook run BEFORE `chat.shutdown()`. Errors here are logged, not thrown. */
+	onBeforeDisconnect?(ctx: AgentChatIntegrationContext): Promise<void>;
```
yehorkardash: onBeforeDisconnect is not called anywhere

yehorkardash (commented):
Looks good, just one small comment

packages/cli/src/modules/agents/integrations/agent-chat-integration.ts
burivuhster: Good catch! I removed it, as it's not necessary so far

yehorkardash (approved):

PR #28331, opened 2026-04-10
feat(core): Add project-scoped insights permissions

fabian8n:

## Summary

Splits insights permissions from global-only to global + project-scoped, following the external secrets pattern.

**Changes:**
- Adds `insights:list` scope to all 4 project roles (admin, personal owner, editor, viewer)
- Extracts shared date filter logic from `InsightsController` into `InsightsBaseController`
- Creates new `InsightsProjectController` at `/insights/projects/:projectId/*` with `@ProjectScope('insights:list')`
- Registers the new controller in the insights module
- Existing global controller (`/insights/*` with `@GlobalScope`) is unchanged

**How it works:**
- Global owners/admins use `/insights/*` for instance-wide data (unchanged)
- Project members use `/insights/projects/:projectId/*` for project-scoped data
- `@ProjectScope` middleware handles authorization automatically via route params
- A user with access to multiple projects can query each independently

**No frontend changes** — the UI continues using the existing global endpoints. Frontend integration (KPI cards, project filtering, dashboard) comes in a follow-up PR (LIGO-358).

**How to test:**
- Existing insights tests pass (no regression)
- New project controller tests verify projectId routing, date validation, license checks, and route param vs query param precedence

https://linear.app/n8n/issue/LIGO-358

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/permissions/src/roles/scopes/project-scopes.ee.ts
```diff
@@ -50,6 +50,7 @@ export const REGULAR_PROJECT_ADMIN_SCOPES: Scope[] = [
 	'projectVariable:create',
 	'projectVariable:update',
 	'projectVariable:delete',
+	'insights:list',
```
MarcL, line 53: All of these permission changes look good. 👍

packages/cli/src/modules/insights/insights-base.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+import { UserError } from 'n8n-workflow';
+import { z } from 'zod';
+
+import { BadRequestError } from '@/errors/response-errors/bad-request.error';
+import { ForbiddenError } from '@/errors/response-errors/forbidden.error';
+import { InternalServerError } from '@/errors/response-errors/internal-server.error';
+
+import type { InsightsService } from './insights.service';
+
+export abstract class InsightsBaseController {
```
MarcL: While it does make some sense, no other controller uses this inheritance pattern.
Most of them use a "dependency injection" (DI) pattern.

As we've also got similar logic in the public API, it would make sense to pull these functions with the business logic into a small service which could be injected into the controller and also used by the public API (which can't inherit from this controller). It'll avoid us having duplicated logic.

packages/cli/src/modules/insights/__tests__/insights-project.controller.test.ts
```diff
@@ -0,0 +1,216 @@
...
+beforeAll(async () => {
+	await testDb.init();
+});
+
+afterAll(async () => {
+	await testDb.terminate();
+});
+
+describe('InsightsProjectController', () => {
+	const insightsByPeriodRepository = mockInstance(InsightsByPeriodRepository);
```
MarcL: The controller has the `InsightsService` injected into it but we're mocking the `InsightsByPeriodRepository` (which makes the DB calls). This is coupling the tests to the internals. If they change then the tests could break even though we've not changed the controller. Probably better to mock the `InsightsService` instead.

packages/cli/src/modules/insights/insights-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+	RestrictedInsightsByTime,
+} from '@n8n/api-types';
+import { InsightsDateFilterDto, ListInsightsWorkflowQueryDto } from '@n8n/api-types';
+import { AuthenticatedRequest } from '@n8n/db';
+import { Get, Licensed, Param, ProjectScope, Query, RestController } from '@n8n/decorators';
+
+import { InsightsBaseController } from './insights-base.controller';
+import { InsightsService } from './insights.service';
+
+@RestController('/insights/projects')
```
MarcL: We already have some API endpoints that exist on the `/projects` endpoint rather than this `/insights` endpoint. The pattern is `/projects/:projectId/{resouce}` e.g. `/projects/:projectId/folders` and `/projects/:projectId/data-tables`. It's worth following that pattern for consistency with other project endpoints.

MarcL: We also don't check the validity of the `projectId` being passed in. The folder controller has a middleware pattern which returns a 404 if no project exists. We should consider doing something similar (or refactor that middleware out to a shared function if that makes sense).

See: https://github.com/n8n-io/n8n/blob/9072365bdb838b9764f889be5ce3911d0e8391ce/packages/cli/src/controllers/folder.controller.ts#L42-L56

packages/cli/src/modules/insights/insights-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+		super();
+	}
+
+	@Get('/:projectId/summary')
+	@ProjectScope('insights:list')
+	async getProjectInsightsSummary(
+		_req: AuthenticatedRequest,
+		_res: Response,
+		@Param('projectId') projectId: string,
+		@Query query: InsightsDateFilterDto = {},
```
MarcL: The `InsightsDateFilterDto` for the incoming request query parameters contains an optional `projectId` parameter so this means we have 2 ways of passing in the `projectId`. I think the controller `projectId` in the path may take precedence but we should perhaps use another DTO here?

packages/cli/src/modules/insights/insights-base.controller.ts
fabian8n: yeah it is literally just for not duplicating the logic (code). even though it is only a couple of lines. i saw that we don't do this anywhere else. but i thought it might be a better pattern?

packages/cli/src/modules/insights/insights-project.controller.ts
fabian8n: yes that's correct. i created a test for that actually. but yeah maybe making this explicit is better

packages/cli/src/modules/insights/insights-base.controller.ts
MarcL: The compositional pattern is usually better than inheritance as it's generally easier to test and has less coupling.

packages/cli/src/modules/insights/insights-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+		@Param('projectId') projectId: string,
+		@Query query: InsightsDateFilterDto,
+	): Promise<InsightsByTime[]> {
+		const { startDate, endDate } = this.prepareDateFilters(query);
+
+		return (await this.insightsService.getInsightsByTime({
+			projectId,
+			startDate,
+			endDate,
+		})) as InsightsByTime[];
```
geemanjs, line 102: If the `response` contains more/different fields than user expects we should pick out the expected fields rather than cast the type. 

Here is another way:
https://www.notion.so/n8n/Request-Response-2265b6e0c94f8032a12ce711f5478a60?source=copy_link#2265b6e0c94f804581f5c4d68e9d5773

packages/cli/src/modules/insights/insights-project.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+		@Query query: InsightsDateFilterDto,
+	): Promise<RestrictedInsightsByTime[]> {
+		const { startDate, endDate } = this.prepareDateFilters(query);
+
+		return (await this.insightsService.getInsightsByTime({
+			insightTypes: ['time_saved_min'],
+			projectId,
+			startDate,
+			endDate,
+		})) as RestrictedInsightsByTime[];
```
geemanjs, line 120: Same here - The last thing we want is to send "data" back that we don't want in the response.

packages/cli/src/modules/insights/insights-date-filter.service.ts
```diff
@@ -0,0 +1,81 @@
...
+		try {
+			this.insightsService.validateDateFiltersLicense(dateFilters);
+		} catch (error: unknown) {
+			if (error instanceof UserError) {
+				throw new ForbiddenError(error.message);
+			}
+
+			throw new InternalServerError();
+		}
+	}
```
geemanjs, line 92: Any reason why we map these errors? The ones coming back from the insight service seem to give the right information

packages/cli/src/modules/insights/insights-base.controller.ts
geemanjs: I think the only reason we've gone for the inheritance model here is because the `insightsService` is in use. I agree with @MarcL RE the inheritance vs composition 👍 - we need a "good reason" to reach for inheritance usually and I don't think this is a "good reason"... 

Other than the `checkDatesFiltersAgainstLicense` these can be functional "utility" functions. 

(being picky) I also think theres a bit of a "smell" in here 👀 .. `prepareDateFilters` is doing more than it suggests:
1. Validating dates and throwing an error
2. Extracting and defaulting dates (now - 7d) - I think the defaulting should happen higher up when you pluck the values out of the request payload as it's "specific" to each call.
3. Checking licensing information

packages/cli/src/modules/insights/insights-base.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+		endDate: Date;
+	} {
+		const today = new Date();
+
+		if (!query.startDate) {
+			return {
+				startDate: DateTime.now().minus({ days: 7 }).toJSDate(),
+				endDate: today,
+			};
+		}
```
geemanjs: We should use `DateTime.utc()` for all of these or we'll be using the "system time zone" which could lead to mistakes.

geemanjs (commented):
Given the historic problems in insights - i'd suggest also writing an integration test around these additions.

fabian8n:
Addressed all review feedback in the latest commit. Here's what changed:

**1. Inheritance → DI (MarcL + geemanjs):** Deleted `InsightsBaseController`. Created `InsightsDateFilterService` as an injectable `@Service()`. Both controllers inject it. Follows the composition/DI pattern used everywhere else.

**2. Route pattern (MarcL):** Changed from `/insights/projects/:projectId/*` to `/projects/:projectId/insights/*` to match `/projects/:projectId/folders` and `/projects/:projectId/data-tables`.

**3. Project validation middleware (MarcL):** Added `@Middleware() validateProjectExists()` matching the folder controller pattern. Returns 404 for invalid projectId.

**4. Tests mock service, not repo (MarcL):** Rewrote project controller tests to mock `InsightsService` instead of `InsightsByPeriodRepository`.

**5. Separate DTOs (MarcL):** Created `InsightsProjectDateFilterDto` and `ListInsightsProjectWorkflowQueryDto` without `projectId` field, so the project endpoints have a clean API contract.

**6. DateTime.utc() (geemanjs):** Fixed in the new `InsightsDateFilterService`. Note: the original global controller's pre-existing code also uses `DateTime.now()` — fixing that across both is handled by the shared service now.

**Re: `as` casts (geemanjs):** These are pre-existing from the original global controller. I'd rather fix this properly across both controllers and the service return types in a follow-up rather than half-fix it here.

**Re: error mapping in checkDatesFiltersAgainstLicense (geemanjs):** The service throws `UserError` (from `n8n-workflow`) which doesn't map to HTTP status codes. Without the mapping, the error handler would return 500 instead of 403. Open to suggestions for a cleaner pattern.

**Still TODO:** Integration test (geemanjs). Will add in a follow-up commit.

PR #28333, opened 2026-04-10
chore: Add scoped JWT strategy for public API (no-changelog)

phyllis-noester:
## PR Summary

https://linear.app/n8n/issue/IAM-469

### What this PR does

Implements `ScopedJwtStrategy` — the auth strategy that validates scoped JWTs issued by the token exchange endpoint and authenticates public API requests with them.

Scoped JWTs can be presented in either `Authorization: Bearer <token>` or `x-n8n-api-key` headers. The strategy identifies them by their issuer (`n8n-token-exchange`) and is registered into `AuthStrategyRegistry` by `TokenExchangeModule` on startup, making it the second strategy after `ApiKeyAuthStrategy`.

### Changes

**`ScopedJwtStrategy` (new)**
- Extracts tokens from both `Authorization: Bearer` and `x-n8n-api-key` headers
- Performs a cheap `decode()` + issuer check before the expensive `verify()` call, so non-token-exchange JWTs pass through as `null` without signature overhead
- Resolves `sub` → `subject` and optionally `act.sub` → `actor` from the DB
- Loads scopes from the acting user's role (`actor ?? subject`) — not from the JWT payload — so permission changes take effect immediately without re-issuing tokens
- Sets `req.user` to the acting principal (actor if delegation is present, subject otherwise)
- Sets `req.tokenGrant = { scopes, subject, actor? }`

**`ApiKeyAuthStrategy` (updated)**
- Added `subject: apiKeyRecord.user` to `req.tokenGrant` — `TokenGrant.subject` is now required
- Added early-exit before the DB lookup: if the value in `x-n8n-api-key` is a JWT with an issuer other than `API_KEY_ISSUER`, the strategy returns `null` (abstain) instead of `false` (fail-fast). Without this, a token-exchange JWT in `x-n8n-api-key` would short-circuit the auth chain before `ScopedJwtStrategy` runs

**`TokenExchangeModule` (updated)**
- Registers `ScopedJwtStrategy` into `AuthStrategyRegistry` at the end of `init()`, behind the existing feature flag guard

### Key decisions

**Scopes from role, not JWT payload.** Scopes are resolved at request time from the acting user's `role.scopes` rather than from the `scope` claim embedded in the JWT. This means permission changes take effect immediately and the JWT itself only needs to carry identity claims (`sub`, `act`).

**Actor is optional.** If the JWT carries an `act` claim but the actor user ID is not found in the DB (e.g. the account was deleted after the token was issued), authentication continues with the subject acting as principal. Only if the actor is found *and* disabled does authentication fail. This keeps token-exchange tokens usable across user lifecycle events.

**`ApiKeyAuthStrategy` abstains on non-API-key JWTs.** The issuer check (`decoded.iss !== API_KEY_ISSUER`) is placed before the DB lookup so a token-exchange JWT in `x-n8n-api-key` gets `null` (pass through) rather than `false` (block). A null decode (non-JWT string) still returns `false` to preserve existing rejection behaviour for garbage values.

**Module self-registration (Option B).** `ScopedJwtStrategy` is registered by `TokenExchangeModule.init()` rather than in the public API bootstrap. This keeps the token-exchange module self-contained and ensures the strategy is only active when the module is enabled.
## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/public-api/v1/openapi.yml
```diff
@@ -140,6 +140,11 @@ components:
       type: apiKey
       in: header
       name: X-N8N-API-KEY
+    BearerAuth:
```
afitzek, line 143: This could be confusing for users that use an api key, because the api key is not accepted in this format.

packages/cli/src/modules/token-exchange/services/__tests__/scoped-jwt.strategy.integration.test.ts
```diff
@@ -0,0 +1,152 @@
...
+		await testDb.terminate();
+	});
+
+	it('returns null for a token with a non-token-exchange issuer', async () => {
+		const token = jwtService.sign({ iss: 'n8n', sub: '123' });
+		expect(await strategy.authenticate(makeBearerReq(token))).toBeNull();
+	});
+
+	it('returns false when subject user does not exist in DB', async () => {
+		const token = makeScopedJwt('non-existent-id');
```
afitzek: The user id should be a non existing uuid.

packages/cli/src/modules/token-exchange/services/__tests__/scoped-jwt.strategy.integration.test.ts
```diff
@@ -0,0 +1,152 @@
...
+			relations: { role: true },
+		});
+		const actorScopes = actorWithRole!.role.scopes.map((s) => s.slug);
+		expect(req.tokenGrant?.scopes).toEqual(expect.arrayContaining(actorScopes));
+		expect(req.tokenGrant?.scopes).toHaveLength(actorScopes.length);
+	});
+
+	it('continues without actor and uses subject scopes when actor ID is not in DB', async () => {
+		const subject = await createOwner();
+		const token = makeScopedJwt(subject.id, 'non-existent-actor');
```
afitzek: The actor id should be a non existing uuid.

afitzek (commented):
Code looks good, there are just 2 test problems that we need to fix.

afitzek (approved):
🚀

PR #28335, opened 2026-04-10
fix(core): Skip npm outdated check for verified-only community packages

BerniWittmann:
## Summary

In verified-only mode, `npm outdated` was running on every `listInstalledPackages` call, causing the UI to show an update CTA immediately after installing a verified node that had a published version newer than the latest verified version (e.g. ElevenLabs: verified 0.2.2, but 0.2.4 exists on npm). This happened because npm saw a newer registry version than the Strapi-pinned version that was installed.

This fix skips the `npm outdated` check when `unverifiedEnabled` is `false`, since update availability in that mode is determined by Strapi CMS versions on the frontend. The frontend install flow is also updated to only pin to the Strapi version when in verified-only mode.

## Testing

1. Set `N8N_UNVERIFIED_PACKAGES_ENABLED=false` (verified-only mode)
2. In A Workflow go to Add Node, search for ElevenLabs and install the ElevenLabs node
3. Verify the node installs at the Strapi-pinned version (currently 0.2.2)
4. Verify **no update CTA** is shown immediately after installation
5. Set `N8N_UNVERIFIED_PACKAGES_ENABLED=true` (unverified mode)
6. Go to Settings and uninstall the ElevenLabs Community Node
7. Install the ElevenLabs node again (similar to Step 3)
8. Verify it installs at the latest npm version (no version pinning) (currently 0.2.4)
9. Verify the update CTA behaves as expected (shows only if npm outdated reports a newer version) (right now should not show)

https://linear.app/n8n/issue/NODE-4652

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/features/settings/communityNodes/composables/useInstallNode.ts
```diff
@@ -81,7 +83,7 @@ export function useInstallNode() {
 
 		try {
 			loading.value = true;
-			if (props.type === 'verified') {
+			if (props.type === 'verified' && !settingsStore.isUnverifiedPackagesEnabled) {
 				await communityNodesStore.installPackage(
 					props.packageName,
 					true,
```
DawidMyslak, line 89: When `isUnverifiedPackagesEnabled` is `true` (the default), verified nodes now fall into the `else` branch and call `installPackage(packageName)` without `verify=true`, skipping backend checksum verification. Is this intentional?

DawidMyslak (changes requested):

packages/frontend/editor-ui/src/features/settings/communityNodes/composables/useInstallNode.ts
BerniWittmann: This is exactly the case with the elevenlabs node: 
Available NPM versions are 0.2.2 and 0.2.4 but only 0.2.2 is verified in Strapi. 

With `isUnverifiedPackagesEnabled = true` we want to install the latest version of that node (even if this version is not verified yet, 0.2.4 in this example), so we also have to skip checksum calculation (as we only have the checksum for an older version). 

So yes this is intentional

DawidMyslak (dismissed):

BerniWittmann:
Sorry @DawidMyslak had to resolve some merge conflicts, so apparently you have to give it another review :)

DawidMyslak (approved):
🚀

PR #28337, opened 2026-04-10
fix(core): Use closure-scoped evaluation contexts in VM expression bridge

despairblue:
## Summary

Replace global mutable state in the isolated-vm expression bridge with closure-scoped evaluation contexts using `evalClosureSync`. Each `execute()` call passes callback References as closure arguments (`$0`, `$1`, `$2`), and a global `buildContext()` function creates a fresh context per evaluation. This enables re-entrant and parallel expression evaluation without state corruption.

**Problem:** The VM bridge stored per-evaluation state (`__data`, callbacks) on `globalThis` inside the isolate. When host-side callbacks (e.g. `$parameter` proxy) triggered nested expression evaluations, the nested `execute()` overwrote the global state, corrupting the outer evaluation. This caused `$parameter` references and `$fromAI()` to silently return `undefined` in AI/langchain workflows.

**Fix:** Move all per-evaluation state into the `evalClosureSync` closure. Each evaluation gets its own callbacks and context object — they cannot interfere with each other. The `resetDataProxies()` function is replaced by `buildContext()` which accepts callbacks as parameters.

**Changes by commit:**
1. **Parameterize `createDeepLazyProxy`** — accept optional callbacks instead of reading globalThis
2. **Add `buildContext()`** — new context factory alongside existing `resetDataProxies()`
3. **Restore `vmExecuting` fallback** — temporary legacy fallback for re-entrant calls
4. **Switch to `evalClosureSync`** — the core change, replaces global state with closures
5. **Remove `resetDataProxies()`** — delete old code and global declarations
6. **Clean up** — remove temporary workarounds and debug logging

https://linear.app/n8n/issue/CAT-2775

## Review / Merge checklist

- [ ] ~[Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.~
- [x] Tests included.


packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -94,21 +94,6 @@ export class IsolatedVmBridge implements RuntimeBridge {
 	private config: Required<BridgeConfig>;
 	private logger: Required<BridgeConfig>['logger'];
 
-	// Script compilation cache for performance
-	// Maps expression code -> compiled ivm.Script
-	private scriptCache = new Map<string, ivm.Script>();
```
ivov, line 99: a

packages/@n8n/expression-runtime/src/runtime/context.ts
```diff
@@ -15,47 +15,44 @@ const SafeReferenceError = createSafeErrorSubclass(ReferenceError);
...
+export function buildContext(
+	getValueAtPath: any,
+	getArrayElement: any,
+	callFunctionAtPath: any,
+	timezone?: string,
+): Record<string, unknown> {
 	if (timezone && !IANAZone.isValidZone(timezone)) {
 		throw new Error(`Invalid timezone: "${timezone}"`);
 	}
 	Settings.defaultZone = timezone ?? 'system';
```
ivov, line 46: Timezone is still global, is that a concern? Outer eval starts in tz A, nested eval switches the global default to tz B, outer eval resumes, and any later Luxon use in the outer eval can silently use tz B?

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
```diff
@@ -540,65 +478,50 @@ export class IsolatedVmBridge implements RuntimeBridge {
...
+  return {
+    __isError: true,
+    name: e.name || "Error",
+    message: e.message || "",
+    stack: e.stack || "",
+    extra: extra
+  };
+}`;
+
+			const result = this.context.evalClosureSync(
```
ivov, line 519: We lost the caching of the tournament transforms. Thinking if there's a way to keep closure-scoped isolation. Maybe we can pre-compile a fn that accepts the closure args as params?

```js
globalThis.__evalExpr = function ($0, $1, $2, timezone) {
  var __ctx = buildContext($0, $1, $2, timezone);
  // ...
}
```

packages/@n8n/expression-runtime/src/runtime/context.ts
despairblue: Not a concern right now, timezone settings are per workflow, one isolate is only use for one workflow execution, it should not change.

I'm also not sure if we can change this, given the timezone setting for luxon is global.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: We have 2 caches:
1. expression string -> tournament transformed
2. tournament transformed -> precompiled script

The first one brought the main performance benefit during a cold start, the second one was not that big of a boost and that's the one we lost with this approach and can't really get back.

Benchmarks will show how much this really mattered. 
Then we can improve it.

packages/@n8n/expression-runtime/src/bridge/isolated-vm-bridge.ts
despairblue: accident?

ivov (approved):

PR #28338, opened 2026-04-10
refactor(instance-ai): Harmonize prompting between builders (no-changelog)

OlegIvaniv:
## Summary

Both `instance-ai` and `ai-workflow-builder.ee` had their own forked copies of the same prompting content, expression references, workflow rules, parameter guides, best-practices docs, node recommendations. When one got updated the other drifted. This PR moves all of that shared content into `@n8n/workflow-sdk/prompts` (both packages already depend on workflow-sdk) so there's a single source of truth, then deletes the 46 dead files that got replaced.

Also fixes a bug in the `search-nodes` tool where multi-word queries like `"data table tool"` returned 0 results when combined with `connectionType`, which meant the builder couldn't find auto-generated tool nodes (dataTableTool, googleCalendarTool, etc).

### What changed

Moved SDK reference constants (expression reference, additional functions, workflow rules, SDK patterns) into `workflow-sdk/src/prompts/sdk-reference/`. Both packages now import from there.

Moved all 11 parameter guides (IF, Switch, Set, HTTP, Gmail, etc.), 2 node tips (Webhook, Structured Output Parser), and 4 node recommendations (text/image/video/audio generation) into `workflow-sdk/src/prompts/node-guidance/`. Extracted types locally so there's no dependency on framework-specific packages.

Moved the `WorkflowTechnique` enum, `BestPracticesDocument` interface, and all 16 guide classes into `workflow-sdk/src/prompts/best-practices/`. This also wired up 5 guides that were previously commented out (data-analysis, enrichment, knowledge-base, human-in-the-loop, monitoring).

Replaced the ~300-line inline `SDK_RULES_AND_PATTERNS` block in instance-ai's builder prompt with composed imports from the shared parameter guides. The instance-ai-specific stuff (AI Agent subnodes, Code Node, Data Table, Google Sheets, web app pattern) stays local.

Deleted 46 original source files across both packages (~5,900 lines).

Fixed `searchByConnectionType` in the node search engine, it was calling `sublimeSearch` directly with the raw query string, which breaks on multi-word input. Extracted a `fuzzySearchNodes` method (with per-term splitting + display-name fallback) that both `searchByName` and `searchByConnectionType` now share. Also improved the `connectionType` param description so the agent actually knows to use `connectionType="ai_tool"` when looking for tool nodes to attach to an AI Agent.

### Layout

```
@n8n/workflow-sdk/src/prompts/
├── index.ts
├── sdk-reference/          expressions, functions, rules, patterns
├── node-guidance/
│   ├── parameter-guides/   11 guides + types
│   ├── node-tips/          2 tips + types
│   └── node-recommendations/  4 recs + types + format utility
└── best-practices/
    ├── types.ts            WorkflowTechnique enum, interfaces
    └── guides/             16 guide classes + registry
```

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/instance-ai/src/tools/nodes/search-nodes.tool.ts
```diff
@@ -13,7 +13,14 @@ export const searchNodesInputSchema = z.object({
 		.string()
 		.optional()
 		.describe(
-			'AI connection type to search for sub-nodes (e.g., "ai_languageModel", "ai_memory", "ai_tool", "ai_embedding", "ai_vectorStore")',
+			'Filter results by AI connection type. Use this when you need sub-nodes for an AI Agent: ' +
+				'"ai_tool" for tool nodes the agent can call (e.g., Google Calendar Tool, Slack Tool), ' +
+				'"ai_languageModel" for LLM providers, "ai_memory" for memory backends, ' +
+				'"ai_embedding" for embedding models, "ai_vectorStore" for vector stores. ' +
+				'IMPORTANT: Many regular nodes (Google Calendar, Slack, Gmail, etc.) have auto-generated ' +
```
mutdmour: We should make this simpler for the LLM to follow. Could we always surface the tools if available and no connection type is set?

packages/@n8n/ai-workflow-builder.ee/src/tools/best-practices/index.ts
```diff
@@ -1,38 +1,4 @@
...
-import { ChatbotBestPractices } from './chatbot';
-import { ContentGenerationBestPractices } from './content-generation';
-// import { DataAnalysisBestPractices } from './data-analysis';
-import { DataExtractionBestPractices } from './data-extraction';
-import { DataPersistenceBestPractices } from './data-persistence';
-import { DataTransformationBestPractices } from './data-transformation';
-import { DocumentProcessingBestPractices } from './document-processing';
-// import { EnrichmentBestPractices } from './enrichment';
-import { FormInputBestPractices } from './form-input';
-// import { HumanInTheLoopBestPractices } from './human-in-the-loop';
```
mutdmour, line 13: some of these disabled best practices are now enabled. was this intentional?

packages/@n8n/instance-ai/src/tools/best-practices/index.ts
```diff
@@ -1,35 +1,16 @@
...
-	[WorkflowTechnique.DATA_EXTRACTION]: getDataExtractionDocs,
-	[WorkflowTechnique.DATA_PERSISTENCE]: getDataPersistenceDocs,
-	[WorkflowTechnique.DATA_TRANSFORMATION]: getDataTransformationDocs,
-	[WorkflowTechnique.DOCUMENT_PROCESSING]: getDocumentProcessingDocs,
-	[WorkflowTechnique.ENRICHMENT]: undefined,
-	[WorkflowTechnique.FORM_INPUT]: getFormInputDocs,
-	[WorkflowTechnique.KNOWLEDGE_BASE]: undefined,
-	[WorkflowTechnique.NOTIFICATION]: getNotificationDocs,
-	[WorkflowTechnique.TRIAGE]: getTriageDocs,
-	[WorkflowTechnique.HUMAN_IN_THE_LOOP]: undefined,
```
mutdmour, line 32: disabled here, now enabled.

packages/@n8n/instance-ai/src/tools/nodes/node-search-engine.ts
```diff
@@ -112,13 +112,14 @@ export class NodeSearchEngine {
...
-	 * Always returns the latest version of a node.
-	 * @param query - The search query string
-	 * @param limit - Maximum number of results to return
-	 * @returns Array of matching nodes sorted by relevance
+	 * Fuzzy-search a list of nodes by query, handling multi-word queries by
+	 * splitting into individual terms and merging results. Falls back to
+	 * direct type-name / display-name matching for nodes the fuzzy search missed.
 	 */
-	searchByName(query: string, limit: number = 20): NodeSearchResult[] {
+	private fuzzySearchNodes(
```
mutdmour, line 119: would be great to add tests for related fixes here

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -0,0 +1,496 @@
...
+});
+
+// Declare branch nodes
+const branch1 = node({ type: 'n8n-nodes-base.httpRequest', ... });
+const branch2 = node({ type: 'n8n-nodes-base.httpRequest', ... });
+const processResults = node({ type: 'n8n-nodes-base.set', ... });
+
+// Connect branches to specific merge inputs using .input(n)
+export default workflow('id', 'name')
+  .add(trigger({ ... }))
```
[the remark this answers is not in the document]
OlegIvaniv: This is a false positive. The SDK's `.add(trigger({...}))` is a graph-building idiom to re-anchor the chain and start a new parallel branch from the same entry point. It's the same pattern used in `independent_sources` and `fan_in` above it. The `...` placeholders represent the same trigger node

packages/@n8n/ai-workflow-builder.ee/src/tools/best-practices/index.ts
OlegIvaniv: Thanks! Wasn't intentional, disabled them again

packages/@n8n/instance-ai/src/tools/nodes/search-nodes.tool.ts
OlegIvaniv: Switched `connectionType` to a `z.enum(AI_CONNECTION_TYPES)` so the LLM sees the exact valid values in the schema and removed the verbose description since the enum is self-documenting. I think always auto-surfacing tool variants is too much. Tool nodes are only relevant when building with AI Agent/provider nodes that support tool connections. Mixing them into every search would surface non-relevant tool nodes in the results. With the enum change and the multi-word search fix, LLM is already picking the right `connectionType` but let's keep an eye on traces for this tool

mutdmour (approved):
Looks good. thanks for addressing my feedback

PR #28340, opened 2026-04-10
fix(core): Refresh the ai assistant token if it's about to expire (no-changelog)

Cadiac:
## Summary

We were requesting a new token from AI assistant each turn, but as a single turn can take a while and as the tokens only have 10min TTL at production they could expire, leading into "Unauthorized" errors seen on the agent.

Daytona proxy always gets a new token already with a lazy get function for the token, and I implemented something similar here, but as I didn't want every interaction with the proxy endpoints (for instance the anthropic one gets called a lot) to result in a new token call I made us reuse the same token during the turn until its expiration time is near.

This complicates things a bit but should make it so that the token no longer expires during a long response turn / if user leaves the agent waiting on a HITL suspension for a while.

https://www.notion.so/n8n/aef5b6e0c94f82159cc58164aa417607?v=7965b6e0c94f823c918f88516500cb86&p=33e5b6e0c94f8080ae24ff196877207f&pm=s

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/instance-ai/instance-ai.service.ts
+++ b/packages/cli/src/modules/instance-ai/instance-ai.service.ts
@@ -362,23 +362,6 @@ export class InstanceAiService {
 	}
 
 	/**
-	 * Create a ProxyTokenManager that lazily fetches and caches proxy auth
-	 * tokens for a given user.  One manager is shared across the Anthropic
-	 * model, Brave Search, and LangSmith tracing configs within a single
-	 * orchestration run so every outbound request gets a valid token.
-	 */
-	private async createProxyTokenManager(user: User): Promise<{
-		client: Awaited<ReturnType<AiService['getClient']>>;
-		tokenManager: ProxyTokenManager;
-	}> {
-		const client = await this.aiService.getClient();
-		const tokenManager = new ProxyTokenManager(async () => {
-			return await client.getBuilderApiProxyToken({ id: user.id }, { userMessageId: nanoid() });
-		});
-		return { client, tokenManager };
-	}
-
-	/**
 	 * Build model config. When the AI service proxy is enabled, returns a native
 	 * Anthropic LanguageModelV2 instance pointing at the proxy.
 	 *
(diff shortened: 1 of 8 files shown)


packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -1077,9 +1075,31 @@ export class InstanceAiService {
...
+		let proxyBaseUrl: string | undefined;
+		if (this.aiService.isProxyEnabled()) {
+			const client = await this.aiService.getClient();
+			proxyBaseUrl = client.getApiProxyBaseUrl();
+			tokenManager = new ProxyTokenManager(async () => {
+				return await client.getBuilderApiProxyToken({ id: user.id }, { userMessageId: nanoid() });
+			});
+			searchProxyConfig = {
+				apiUrl: proxyBaseUrl + '/brave-search',
+				getAuthHeaders: async () => await tokenManager!.getAuthHeaders(),
```
mutdmour: nitpick - prefer type guards over `!`

packages/cli/src/modules/instance-ai/proxy-token-manager.ts
```diff
@@ -0,0 +1,92 @@
+const DEFAULT_TTL_MS = 10 * 60 * 1000; // 10 min — fallback if JWT has no exp
```
mutdmour: what if we increase the timeout service side? 10 minutes was arbitrary and more fitting for the ai builder workflows. 
Maybe the service should be returning the expiry time. But that's not a blocker for this.

mutdmour (dismissed):
LGTM. Minor comments. No blockers.

packages/cli/src/modules/instance-ai/proxy-token-manager.ts
Cadiac: We read the actual expiry on line 21 - this is a fallback that I think never gets used but claude wanted to be cautious. I think we could just error if the token has no expiry set somehow, as the expiry field should always be present on the JWT token.

If we change the expiry on assistant side this code should work just fine.

I did consider increasing the timeout to bypass the issue but I wasn't sure if there had been a good reason to keep it low for builder earlier.

mutdmour: Ya even if we increase it, but this would still be helpful.

Cadiac:
@mutdmour , could you have another look - I just made it an error if the expiry would be somehow missing to make it clear we're using the expiry from the token.

mutdmour (approved):
Nice now it's more flexible if we increase the TTL serverside

PR #28341, opened 2026-04-10
fix(Oracle DB Node): Handle the test failures

sudarshan12s:
## Summary

As part of fix for [issue](https://github.com/n8n-io/n8n/issues/26985),  the tests were failing because of some minor issue where the `returnData`is not returned in function `_getResponseForOutbinds`. 

Thanks @MDub3y for the earlier fixes. I made a minor change on top of this to fix the outbind tests. Can you take a look.. 

tests run using:
```

cd $WORK/packages/nodes-base/nodes/Oracle/Sql/test
export ORACLE_USER=DEMO
export ORACLE_PASSWORD=demo
export ORACLE_CONNECTSTRING=localhost:1521/freepdb1
export ORACLE_INTEG_TESTS=1
npx jest $WORK/packages/nodes-base/nodes/Oracle/Sql/test/*

```

Modified _getResponseForOutbinds to mutate the `returnData` with a for…of push; this keeps the max-argument fix (no spread) while preserving the original reference so `configureQueryRunner` sees the out-bind rows.
`• file: packages/nodes-base/nodes/Oracle/Sql/helpers/utils.ts`

Added the out-bind regression test: it now fakes a single out-bind row that expands to 250 k result items, reproducing the call-stack overflow.  Also added a test for the SELECT path. It mocks the driver returning 250 k rows to ensure the success-path aggregation.
`• file: packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts
`

https://github.com/n8n-io/n8n/issues/26985
https://github.com/n8n-io/n8n/commit/bd5a70215d19e46a53ef8705475616a737d4094f

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts
```diff
@@ -236,3 +238,85 @@ describe('Test getBindParameters ', () => {
...
+		const pool = { getConnection } as unknown as oracleDBTypes.Pool;
+		const constructExecutionMetaData = jest
+			.fn()
+			.mockImplementation((data: INodeExecutionData[]) => data);
+		const context = {
+			helpers: {
+				constructExecutionMetaData,
+			},
+		} as unknown as IExecuteFunctions;
+		const node = {} as any;
```
alexander-gekov: Instead of `as any` it can be `as unknown as INode`

alexander-gekov (changes requested):
Thanks for the contribution, left a small comment regarding use of `any` in test. Ready to unblock once that's resolved!

packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts
sudarshan12s: made changes to use INode

alexander-gekov (dismissed):
Looks good! Thanks for the fast change.

Qodo-Free-For-OSS:
Hi, The new `configureQueryRunner` unit tests allocate 250k-element arrays (and 250k row objects), which significantly increases memory/CPU consumption in CI and can make the Jest suite slower or flaky. These tests run unconditionally (not integration-gated) and the select test in particular constructs multiple large arrays in the same execution path.

**Severity:** remediation recommended | **Category:** performance

**How to fix:** Reduce test dataset size

> ### Issue description
> `packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts` introduces unit tests that allocate 250k-element arrays (and 250k row objects), increasing CI resource usage and potential flakiness.
> 
> ### Issue Context
> The purpose is to regression-test stack-overflow scenarios (previously caused by using spread/concat patterns). You can validate the same behavior with a smaller but still “large” dataset.
> 
> ### Fix Focus Areas
> - packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts[242-321]
> 
> ### Suggested changes
> - Reduce `chunkSize` to a smaller value that still exercises the non-spread/non-recursive path (e.g. 10_000–50_000).
> - Optionally add `jest.setTimeout(...)` if you want to keep very large sizes, but prefer reducing the allocation footprint.
> - Keep assertions that validate first/last elements and call counts.

---

sudarshan12s:
> Hi, The new `configureQueryRunner` unit tests allocate 250k-element arrays (and 250k row objects), which significantly increases memory/CPU consumption in CI and can make the Jest suite slower or flaky. These tests run unconditionally (not integration-gated) and the select test in particular constructs multiple large arrays in the same execution path.
> 
> **Severity:** remediation recommended | **Category:** performance
> 
> **How to fix:** Reduce test dataset size
> 
> **Agent prompt to fix** - you can give this to your LLM of choice:
> 
> > ### Issue description
> > `packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts` introduces unit tests that allocate 250k-element arrays (and 250k row objects), increasing CI resource usage and potential flakiness.
> > ### Issue Context
> > The purpose is to regression-test stack-overflow scenarios (previously caused by using spread/concat patterns). You can validate the same behavior with a smaller but still “large” dataset.
> > ### Fix Focus Areas
> > 
> > * packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts[242-321]
> > 
> > ### Suggested changes
> > 
> > * Reduce `chunkSize` to a smaller value that still exercises the non-spread/non-recursive path (e.g. 10_000–50_000).
> > * Optionally add `jest.setTimeout(...)` if you want to keep very large sizes, but prefer reducing the allocation footprint.
> > * Keep assertions that validate first/last elements and call counts.
> 
> _Found by [Qodo](https://qodo.ai) code review_

The 250k row test was needed to check the stack overflow [issue](https://github.com/n8n-io/n8n/issues/26985). The value was taken based on my node.js version.      I now run the test only when the environment variable, ORACLE_INTEG_TESTS is defined. So It will not be run by default.

sudarshan12s:
@alexander-gekov, I made some minor lint fixes and test updates on top of your last review to address the CI failures and AI suggestions. Can you take a look whenever you have a moment. 

Thanks!

alexander-gekov:
@claude review once

packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts
```diff
@@ -236,3 +240,89 @@ describe('Test getBindParameters ', () => {
 		expect(bindParameters).toEqual(expectedBindParams);
 	});
 });
+
```
alexander-gekov, line 241: These tests are fully mocked (no Oracle DB connection), so gating them behind `ORACLE_INTEG_TESTS` means they never run in CI and provide no regression safety.

Instead of allocating 250k items, you can test the behavioral contract directly with a spy-based approach and just a few items:

```ts
it('should mutate returnData in-place via push (not concat)', async () => {
  const returnData: INodeExecutionData[] = [];
  const pushSpy = jest.spyOn(returnData, 'push');
  const concatSpy = jest.spyOn(returnData, 'concat');

  // ... set up context, mock constructExecutionMetaData to return 3 items ...

  _getResponseForOutbinds.call(context, results, 'single', ['COL1'], returnData);

  expect(returnData.length).toBeGreaterThan(0);
  expect(pushSpy).toHaveBeenCalled();
  expect(concatSpy).not.toHaveBeenCalled();
});
```

This validates the exact regression (concat vs push) with zero memory overhead and can run unconditionally in CI.

alexander-gekov (changes requested):
Thanks for the fix! The core change (switching from `.concat()` to in-place `.push()` to preserve the `returnData` reference) is correct and addresses a real bug.

Two things to address on the test side:

1. **Tests should run unconditionally in CI** — they're currently gated behind `ORACLE_INTEG_TESTS` but don't actually connect to an Oracle DB (fully mocked). Without CI coverage, these provide no regression safety net.

2. **250k item allocation is unnecessary** — you can validate the fix with a spy-based approach using just a few items (see inline comment), which removes the memory/CI concern entirely and lets the tests run unconditionally.

packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts
sudarshan12s: Thanks for suggestion. I had to do for outbinds (_getResponseForOutbinds ) and select (configureQueryRunner). I will check the changes..

sudarshan12s: I made changes to mock `configureQueryRunner` to include `select` and `insert` with returning out binds.

I avoided `_getResponseForOutbinds` as i need to export it for test.  To make sure the test still catches the original regression, I added an assertion that fails if the code goes back to using `returnData.push(...executionData)`. This uses jest.spyOn(Array.prototype, 'push'), which does observe all array pushes during the configureQueryRunner call, but it is filtered to the mocked execution data entries and restored in a finally block.

I also kept the opt-in large dataset tests with 250_000 entries to simulate the original stack overflow case more directly, but they are skipped for regular CI.

Does this look reasonable, or does the prototype spy make the test feel too complex?

alexander-gekov (changes requested):
Thanks again for the updates - this is looking much better.

Could you please also remove the ORACLE_STRESS_TESTS env flag from this test file?
I’m totally fine keeping the 250k cases as manual stress coverage, but I’d prefer to mark them explicitly with describe.skip(...) (or it.skip(...)) rather than gating them via a new env variable.

sudarshan12s:
> Thanks again for the updates - this is looking much better.
> 
> Could you please also remove the ORACLE_STRESS_TESTS env flag from this test file? I’m totally fine keeping the 250k cases as manual stress coverage, but I’d prefer to mark them explicitly with describe.skip(...) (or it.skip(...)) rather than gating them via a new env variable.

Thanks for checking . I have explicitly skipped the 250k test.

packages/nodes-base/nodes/Oracle/Sql/test/utils.test.ts
```diff
@@ -236,3 +238,197 @@ describe('Test getBindParameters ', () => {
...
+		expect(result[0]?.json).toMatchObject({ COL1: 1 });
+		expect(result[2]?.json).toMatchObject({ COL1: 3 });
+		expect(constructExecutionMetaData).toHaveBeenCalledTimes(1);
+		expect(getConnection).toHaveBeenCalledTimes(1);
+		expect(close).toHaveBeenCalledTimes(1);
+	});
+});
+
+// eslint-disable-next-line n8n-local-rules/no-skipped-tests
+describe.skip('configureQueryRunner stack overflow regression', () => {
```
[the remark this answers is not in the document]
sudarshan12s: It is already verified by mock tests:

`should append out-bind execution data one item at a time without spread push `

and
 
`should return select execution data from the concat path`

alexander-gekov (approved):
Thanks for the changes! Approved.

PR #28344, opened 2026-04-10
feat(ai-builder): Add a binary check to validate node references (no-changelog)

lucamattiazzi:
## Summary

This PR adds a binary check that validates that all node references refer to nodes upstream.

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/evaluations/evaluators/binary-checks/checks/references-reachable-nodes.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/evaluations/evaluators/binary-checks/checks/references-reachable-nodes.ts
@@ -1,0 +1,63 @@
+import { getParentNodes, mapConnectionsByDestination } from 'n8n-workflow';
+import type { IConnections, INode } from 'n8n-workflow';
+
+import type { BinaryCheck, SimpleWorkflow } from '../types';
+import {
+	extractExpressionsFromParams,
+	extractNodeNamesFromExpression,
+} from './expressions-reference-existing-nodes';
+
+function nodeReferenceIsInvalid(parents: Set<string>, nodeReference: string): boolean {
+	return !parents.has(nodeReference);
+}
+
+function nodeIsInvalid(connectionsByDest: IConnections, node: INode): boolean {
+	if (!node.parameters) return false;
+
+	const expressions = extractExpressionsFromParams(node.parameters);
+	const referencedNames = expressions.flatMap((expr: string) =>
+		extractNodeNamesFromExpression(expr),
+	);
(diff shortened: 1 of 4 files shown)


packages/@n8n/ai-workflow-builder.ee/evaluations/evaluators/binary-checks/checks/references-reachable-nodes.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * This deterministic check checks that every time a node references
+ * another node by name that node must be an ancestor, i.e. it must
+ * be reachable by following the parent chain upstream, using the same
+ * logic that the frontend implements currently in `workflow-data-proxy.ts`
+ *
+ * Fails when:
+ * 1. The referenced node does not exist in the workflow. (this actually
+ * makes "expressions_reference_existing_nodes" redundant)
```
schrothbn, line 40: Not really redundant since you are referencing the code from the old check here? Probably a better idea to either remove the other check then and move the code here since this is the superset of the other one? Or extract the functions into a util file so we can reuse them throughout.

packages/@n8n/ai-workflow-builder.ee/evaluations/evaluators/binary-checks/checks/references-reachable-nodes.ts
```diff
@@ -0,0 +1,63 @@
...
+
+	const expressions = extractExpressionsFromParams(node.parameters);
+	const referencedNames = expressions.flatMap((expr: string) =>
+		extractNodeNamesFromExpression(expr),
+	);
+	const uniqReferencedNames = new Set<string>(referencedNames);
+
+	if (uniqReferencedNames.size === 0) return false;
+
+	const parents = new Set(getParentNodes(connectionsByDest, node.name, 'main', -1));
```
schrothbn, line 25: This might fail for subnodes (models, tools, memory,...) as they don't have any main connection. But they can technically reference nodes that are connected to their root node (agent, vector store, llm chains).

schrothbn (dismissed):
Looks good as is, comment on the redundancy of the old check, we could probably restructure this better.  One concern of false positives on subnodes. 
All non-blocking though because we move those over to the instance-ai soon and can do it then.

packages/@n8n/ai-workflow-builder.ee/evaluations/evaluators/binary-checks/checks/references-reachable-nodes.ts
lucamattiazzi: oh yes I was referring only to the check! makes sense to move everything in a single check as well then!

PR #28355, opened 2026-04-10
fix(Todoist Node): Migrate deprecated sync v9 endpoint to API v1

jeanibarz:
## Summary

Todoist has deprecated the `/sync/v9/sync` endpoint. It now returns **HTTP 410 Gone** with a deprecation message pointing users to `/api/v1/`. This breaks every Todoist operation that routes through `todoistSyncRequest`:

- **Move a task** (the symptom reported in #26450)
- Reminder CRUD (`ReminderCreate/Update/Delete/GetAll`)
- Sync handler on the v1 node
- **Quick Add** — which uses the related `/sync/v9/quick/add` path, also deprecated

Migrate `todoistSyncRequest` to POST to `https://api.todoist.com/api/v1/sync`. Introduce `todoistQuickAddRequest` for the new Quick Add path at `https://api.todoist.com/api/v1/tasks/quick_add`, which moved out of the sync namespace in the v1 API.

### Verification

Confirmed the deprecation and the migration target before writing code:

```
$ curl -s -X POST https://api.todoist.com/sync/v9/sync
This endpoint is deprecated.

If you're reading this on a browser, there's a good chance you can change
the v8 part on the URL to v1 and get away with it. ;)

If you're using the API directly, please update your use case to rely
on the new API endpoints, available under /api/v1/ prefixes.

$ curl -s -o /dev/null -w "%{http_code}\n" https://api.todoist.com/api/v1/sync
405  # (GET is 405, POST works with auth — endpoint is live)

$ curl -s -o /dev/null -w "%{http_code}\n" https://api.todoist.com/api/v1/tasks/quick_add
400  # (endpoint is live, 401 with empty POST body)
```

### What changed

- `GenericFunctions.ts` — `todoistSyncRequest` URL updated, `endpoint` parameter removed (no caller depended on it after rerouting Quick Add). New `todoistQuickAddRequest` helper.
- `v2/OperationHandler.ts` — `QuickAddHandler` now calls `todoistQuickAddRequest`.
- `v1/OperationHandler.ts` — URL comment on `MoveHandler` updated (v1 node is fixed transitively through the shared helper).
- Tests updated: `TodoistV2.node.test.ts` nock mocks now expect `/api/v1/sync` and `/api/v1/tasks/quick_add`. `OperationHandler.test.ts` QuickAddHandler tests use the new mock.
- Added four unit tests in `test/GenericFunctions.test.ts` covering both credential types (oAuth2 / apiKey) for both `todoistSyncRequest` and `todoistQuickAddRequest`.

### How to test

1. Check out this branch
2. From `packages/nodes-base`:
   ```
   pnpm jest --testPathPattern="nodes/Todoist"
   ```
   Expected: **69 tests pass, 3 suites pass** (4 new tests on top of the existing 65).
3. In a live n8n instance with a Todoist credential, run any workflow that uses Todoist **Move a task**, **Quick add a task**, or reminders — operations that failed with `410 Gone` before this patch should now succeed.

### Notes

- **Scope kept tight.** The v1 `MoveHandler` change is comment-only; the v1 node is fixed transitively because both v1 and v2 route through the shared `todoistSyncRequest` in `GenericFunctions.ts`.
- **Body shapes are unchanged.** The Todoist v1 Sync API is a rehost of the same sync engine: the `{commands: [{type, uuid, args, temp_id}], sync_token, resource_types, temp_id_mapping}` envelope is identical. Auth is also unchanged — `todoistApiRequest` has been hitting `/api/v1/` for `nodeVersion >= 2.2` with the same credential types for a while already.
- **Pre-existing gap (not fixed here):** the `Reminder*Handler` operations on the v2 node also use `todoistSyncRequest` but aren't covered in `TodoistV2.node.test.ts`. They'll benefit from the fix transitively, but adding integration coverage for them is out of scope for this bug fix and would be a separate PR.

Closes #26450

## Review / Merge checklist

- [x] Tests included (4 new unit tests + updated nock mocks covering both changed endpoints)
- [ ] Docs updated or follow-up ticket created (n/a — no user-facing doc changes; endpoint URLs are internal implementation detail)

### DIFF
--- a/packages/nodes-base/nodes/Todoist/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/Todoist/GenericFunctions.ts
@@ -52,7 +52,6 @@ export async function todoistSyncRequest(
 	this: Context,
 	body: any = {},
 	qs: IDataObject = {},
-	endpoint: string = '/sync',
 ): Promise<any> {
 	const authentication = this.getNodeParameter('authentication', 0, 'oAuth2');
 
@@ -60,13 +59,32 @@ export async function todoistSyncRequest(
 		headers: {},
 		method: 'POST',
 		qs,
-		uri: `https://api.todoist.com/sync/v9${endpoint}`,
+		uri: 'https://api.todoist.com/api/v1/sync',
 		json: true,
 	};
 
 	if (Object.keys(body as IDataObject).length !== 0) {
 		options.body = body;
 	}
+
+	try {
+		const credentialType = authentication === 'oAuth2' ? 'todoistOAuth2Api' : 'todoistApi';
+		return await this.helpers.requestWithAuthentication.call(this, credentialType, options);
+	} catch (error) {
+		throw new NodeApiError(this.getNode(), error as JsonObject);
+	}
+}
+
+export async function todoistQuickAddRequest(this: Context, body: IDataObject = {}): Promise<any> {
+	const authentication = this.getNodeParameter('authentication', 0, 'oAuth2');
+
+	const options: IRequestOptions = {
+		headers: {},
+		method: 'POST',
+		uri: 'https://api.todoist.com/api/v1/tasks/quick_add',
+		json: true,
+		body,
+	};
 
 	try {
 		const credentialType = authentication === 'oAuth2' ? 'todoistOAuth2Api' : 'todoistApi';
--- a/packages/nodes-base/nodes/Todoist/test/GenericFunctions.test.ts
+++ b/packages/nodes-base/nodes/Todoist/test/GenericFunctions.test.ts
@@ -2,14 +2,21 @@ import { mock } from 'jest-mock-extended';
 import type { IExecuteFunctions, INode } from 'n8n-workflow';
 
 import type { Context } from '../GenericFunctions';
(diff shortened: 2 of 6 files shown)


jeanibarz:
Apologies — I missed that @zozo123 already has an open PR for this issue ([#28048](https://github.com/n8n-io/n8n/pull/28048), opened 2026-04-05), which migrates the Move task handler via a different approach (switching to the REST `POST /tasks/{id}` endpoint instead of the v1 sync endpoint). My scouting step failed to surface it and I should not have opened this duplicate.

Closing this PR in deference to #28048. I did want to flag for maintainer awareness that the Todoist `/sync/v9/*` deprecation affects more than just Move — the same `todoistSyncRequest` helper is also used by `ReminderCreate/Update/Delete/GetAll` handlers and the v1 node's `SyncHandler`, and the sibling `/sync/v9/quick/add` path is used by `QuickAddHandler`. Those paths are all returning HTTP 410 Gone in production right now, independent of which approach is taken for Move. If the maintainer wants a follow-up issue opened for those, happy to file one.

cc @zozo123 — sorry for the noise.

PR #28358, opened 2026-04-11
fix(telegram Node): Use RFC 5987 encoding for non-ASCII filenames in Send Document

### LINKED ISSUE #28354: Regression: Telegram Send Document strips Chinese punctuation from binary file names
### Bug Description

When using the Telegram node with **Message → Send Document**, the file name stored in the binary data is not preserved after upload.

The issue affects **Chinese punctuation characters** in the file name. In my tests, Chinese punctuation is stripped or normalized after the file is sent through the Telegram node.

Example:

Binary file name before sending:

`认知天性【标题已矫正】.md`

File name returned by Telegram node output after sending:

`认知天性 标题已矫正.md`

Observed behavior:
- Chinese characters themselves are preserved
- Chinese punctuation characters are stripped or normalized
- This includes at least:
  - `【】`
  - `（）`
- From my testing, this seems to affect Chinese punctuation in general, not just one specific pair of brackets

This appears to be a regression:
- **At least n8n 1.109.1 worked correctly**
- **n8n 2.14.2 shows this issue**
- I have not tested every version in between yet

From a workflow perspective, the binary file name is correct before sending, but the file name returned after the Telegram upload is changed.

### To Reproduce

1. Prepare a binary file in n8n.
2. Set the binary file name to a value containing Chinese punctuation, for example:
   `认知天性【标题已矫正】.md`
3. Add a **Telegram** node with:
   - Resource: `Message`
   - Operation: `Send Document`
   - Binary File: enabled
   - Input Binary Field: the binary field containing the file
4. Execute the node.
5. Inspect the Telegram node output, especially:
   `result.document.file_name`

Actual result:
- The original binary file name is not preserved
- Chinese punctuation is stripped or normalized
- Example returned value:
  `认知天性 标题已矫正.md`

Additional notes:
- The same behavior also happens with other Chinese punctuation, such as `（）`
- **n8n 1.109.1** is confirmed working
- **n8n 2.14.2** reproduces the issue
- Intermediate versions have not been tested yet

### Expected behavior

The Telegram node should preserve the binary file name exactly as provided before upload, including Chinese punctuation characters.

For example, if the binary file name is:

`认知天性【标题已矫正】.md`

then the uploaded file name returned by the node should remain exactly:

`认知天性【标题已矫正】.md`

The same should apply to other Chinese punctuation characters as well.

Since this behavior works in **n8n 1.109.1** but fails in **n8n 2.14.2**, this looks like a regression.

### Debug Info

Please see the attached screenshots for the node configuration and output.

Confirmed versions:
- Working: `1.109.1`
- Broken: `2.14.2`

Versions in between:
- Not fully tested yet

Environment:
- Deployment type: `[Docker / npm / desktop / cloud / self-hosted]`
- Database: `[SQLite / Postgres / etc.]`
- OS: `[your OS]`
- Node.js version: `[if known]`

Additional notes:
- The binary metadata inside n8n shows the expected file name before sending
- The changed file name appears in the Telegram node output under `result.document.file_name`
- The issue affects Chinese punctuation characters in file names
- Chinese characters themselves are preserved

### Operating System

Ubuntu 24.04 LTS

### n8n Version

2.14.2

### Node.js Version

Official Docker Image

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

octo-patch:
## Summary

When a binary file has a non-ASCII filename (e.g. containing Chinese punctuation characters like `【】（）`), the Telegram **Send Document** operation silently strips or corrupts those characters. This is a regression introduced when the binary file upload was reworked.

**Root cause:** The `Content-Disposition` header for the file part is built by the `form-data` library using `filename="<raw>"`. When the filename contains non-ASCII bytes, some HTTP endpoints (including Telegram's Bot API) may strip or mangle them because raw UTF-8 in an ASCII-era header is technically non-standard.

**Fix:** Detect filenames containing non-ASCII characters and use [RFC 5987](https://datatracker.ietf.org/doc/html/rfc5987) percent-encoding (`filename*=UTF-8''<encoded>`). For ASCII-only filenames the existing plain-object code path is preserved unchanged. The form-data library's `options.header` string override is used to inject the custom Content-Disposition header.

**Changes:**
- `Telegram.node.ts` – new branch in `sendDocument` that builds a real `FormData` instance with RFC 5987-encoded filename header for non-ASCII filenames
- `Telegram.node.test.ts` – two new test cases covering the non-ASCII and ASCII-only code paths

Fixes #28354

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/mcp/mcp.controller.ts
+++ b/packages/cli/src/modules/mcp/mcp.controller.ts
@@ -1,6 +1,6 @@
 import { Logger } from '@n8n/backend-common';
 import { AuthenticatedRequest } from '@n8n/db';
-import { Head, Post, RootLevelController } from '@n8n/decorators';
+import { Get, Head, Post, RootLevelController } from '@n8n/decorators';
 import { Container } from '@n8n/di';
 import type { Request, Response } from 'express';
 import { ErrorReporter } from 'n8n-core';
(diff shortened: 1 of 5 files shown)


BerniWittmann:
I think the stripping of symbols is not an issue in n8n, but rather done upstream by telegram. 

See my comment in the issue: https://github.com/n8n-io/n8n/issues/28354#issuecomment-4294712751

As a result i will close this pr

PR #28360, opened 2026-04-11
fix(core): Use RFC-style default outbound User-Agent

### LINKED ISSUE #28280: WooCommerce node (and others) send bare "n8n" User-Agent string, causing requests to be blocked by WAFs
### Bug Description

n8n sends a bare n8n string as its User-Agent header on outbound HTTP requests. This is not a valid UA string and causes requests to be blocked by Web Application Firewalls (WAFs) that enforce UA string standards. Affected sites return 403 errors with the reason "Bad User-Agent," breaking integrations completely. The only workaround is to abandon native nodes and rebuild them as HTTP Request nodes with a manually set User-Agent header.

Image

### To Reproduce

1. Create a workflow using any native n8n node that makes outbound HTTP requests (e.g. WooCommerce)
2. Target a site protected by a WAF with UA string validation (e.g. SiteDistrict firewall)
3. Execute the workflow
4. Observe requests blocked with HTTP 403 and block reason: "Bad User-Agent"
5. Note: the issue also affects credential testing — the WooCommerce credential test fails with "resource could not be found" because the WAF blocks the test request before it reaches the server

Image

I've also used https://wizardacademy.org/wp-json/wc/v3 and https://wizardacademy.org/wp-json/wc/v3/ as the URL

### Expected behavior

n8n should send a properly formatted User-Agent string that identifies the application, version, and vendor per RFC convention, for example:
Mozilla/5.0 (compatible; n8n/1.0; +https://n8n.io/)
This would allow outbound requests to pass WAF validation and reach their target servers without being blocked.

### Debug Info

I cannot access the debug because I cannot save the node due to the inability to connect to WooCommerce.

# Debug info

## core

- n8nVersion: 1.123.27
- platform: docker (cloud)
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: 5

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 168 hours
- maxCount: 2500 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/146.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-04-10T02:00:28.232Z

### Operating System

unknown

### n8n Version

1.123.27

### Node.js Version

unknown

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

muhammadosama984:
## Summary
- replace the default outbound `User-Agent` from bare `n8n` to a WAF-friendly RFC-style token (`Mozilla/5.0 (compatible; n8n/<version>; +https://n8n.io/)`)
- add a dedicated utility to resolve n8n version for outbound headers while preserving explicit user overrides
- add unit and integration tests covering user-agent generation and runtime request behavior
- implemented from a Cursor-assisted workflow

- Fixes #28280
- https://github.com/n8n-io/n8n/issues/28280

## Test plan
- `cd packages/core && pnpm exec jest src/execution-engine/node-execution-context/utils/__tests__/outbound-user-agent.test.ts src/execution-engine/node-execution-context/utils/__tests__/outbound-user-agent.integration.test.ts src/execution-engine/node-execution-context/utils/__tests__/request-helper-functions.test.ts --no-cache`

## Review / Merge checklist
- [x] Tests included.
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if needed)

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/outbound-user-agent.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/outbound-user-agent.ts
@@ -1,0 +1,31 @@
+import { join } from 'node:path';
+
+const N8N_PRODUCT_URL = 'https://n8n.io/';
+
+function readN8nVersion(): string {
+	try {
+		// require self-caches, so repeated calls are cheap
+		// eslint-disable-next-line @typescript-eslint/no-require-imports
+		const pkg = require(join(__dirname, '../../../../package.json')) as { version: string };
+		return pkg.version;
+	} catch {
+		return '0.0.0';
+	}
+}
+
+/**
+ * Builds the default User-Agent for n8n-initiated HTTP requests when none is set.
+ * Uses a conventional browser-style token so strict WAFs accept the value.
+ *
+ * @see https://github.com/n8n-io/n8n/issues/28280
+ */
+export function buildDefaultN8nOutboundUserAgent(version: string): string {
+	return `Mozilla/5.0 (compatible; n8n/${version}; +${N8N_PRODUCT_URL})`;
+}
+
+/**
+ * Default User-Agent for n8n outbound HTTP requests (e.g. native integration nodes).
+ */
+export function getDefaultN8nOutboundUserAgent(): string {
+	return buildDefaultN8nOutboundUserAgent(readN8nVersion());
+}
--- a/packages/core/src/execution-engine/node-execution-context/utils/__tests__/outbound-user-agent.integration.test.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/__tests__/outbound-user-agent.integration.test.ts
@@ -1,0 +1,33 @@
+import nock from 'nock';
+
+import { getDefaultN8nOutboundUserAgent } from '../outbound-user-agent';
+import { httpRequest } from '../request-helper-functions';
+
+/** Exercises the full httpRequest → axios path for the default User-Agent. */
+describe('Outbound User-Agent (httpRequest integration)', () => {
+	const baseUrl = 'https://example.com';
+
+	beforeEach(() => {
+		nock.cleanAll();
+	});
+
+	afterEach(() => {
+		nock.cleanAll();
+	});
+
+	it('sends RFC-style default User-Agent when none is provided', async () => {
+		const scope = nock(baseUrl, {
+			reqheaders: {
+				'user-agent': getDefaultN8nOutboundUserAgent(),
+			},
+		})
+			.get('/outbound-user-agent-integration')
+			.reply(200, { success: true });
+
+		await expect(
+			httpRequest({ method: 'GET', url: `${baseUrl}/outbound-user-agent-integration` }),
+		).resolves.toEqual({ success: true });
+
+		scope.done();
+	});
+});
--- a/packages/core/src/execution-engine/node-execution-context/utils/__tests__/outbound-user-agent.test.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/__tests__/outbound-user-agent.test.ts
@@ -1,0 +1,30 @@
+import {
+	buildDefaultN8nOutboundUserAgent,
+	getDefaultN8nOutboundUserAgent,
+} from '../outbound-user-agent';
+
+describe('outbound-user-agent', () => {
+	it('buildDefaultN8nOutboundUserAgent uses a conventional product token', () => {
+		expect(buildDefaultN8nOutboundUserAgent('1.2.3')).toBe(
+			'Mozilla/5.0 (compatible; n8n/1.2.3; +https://n8n.io/)',
+		);
+	});
+
+	it('getDefaultN8nOutboundUserAgent returns a non-empty string', () => {
+		const ua = getDefaultN8nOutboundUserAgent();
+		expect(ua).toMatch(/^Mozilla\/5\.0 \(compatible; n8n\/.+; \+https:\/\/n8n\.io\/\)$/);
+	});
+
+	it('getDefaultN8nOutboundUserAgent falls back to 0.0.0 when package.json cannot be read', () => {
+		let ua: string | undefined;
+		jest.isolateModules(() => {
+			jest.mock('../../../../../package.json', () => {
+				throw new Error('Module not found');
+			});
+			// eslint-disable-next-line @typescript-eslint/no-require-imports
+			const mod = require('../outbound-user-agent') as typeof import('../outbound-user-agent');
+			ua = mod.getDefaultN8nOutboundUserAgent();
+		});
+		expect(ua).toBe('Mozilla/5.0 (compatible; n8n/0.0.0; +https://n8n.io/)');
+	});
+});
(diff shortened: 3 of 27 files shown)


packages/core/src/execution-engine/node-execution-context/utils/outbound-user-agent.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Resolves the running n8n version for outbound HTTP User-Agent strings.
+ */
+export function resolveN8nVersionForUserAgent(): string {
+	if (process.env.N8N_VERSION) {
+		return process.env.N8N_VERSION;
+	}
+	if (cachedVersion !== undefined) {
+		return cachedVersion;
+	}
+	cachedVersion = readN8nCliVersion() ?? readN8nCorePackageVersion() ?? '0.0.0';
```
BerniWittmann: This whole file feels over engineered. I would propose just reading the package.json version. require should self-cache if i'm not mistaken, and the different packages cli and core anyhow get the same version on releases

BerniWittmann (changes requested):
Thanks for addressing this issue 🙌 , however I have some comments on this PR. 
Also I think there are some nodes (e.g. GitHub Spotify, etc) that hard code the user Agent to n8n in their respective node files, which we should remove as well for them to utilize RFC conforming User Agent

muhammadosama984:
@BerniWittmann Addressed both review points: simplified outbound-user-agent.ts to use package.json via require caching, and removed hard-coded node User-Agent values so affected nodes use the RFC-style default.

BerniWittmann (changes requested):
Changes look good :+1: 
Thanks for the quick turnaround. 
I think there are still some hardcoded UA headers in credential files: 
`nodes-base/credentials/PerplexityApi.credentials.ts`
`nodes-base/credentials/DropcontactApi.credentials.ts`
`nodes-base/credentials/RundeckApi.credentials.ts`

Once these have been fixed, this is ready to be merged :)

muhammadosama984:
@BerniWittmann 
Implemented: removed hardcoded User-Agent headers from Dropcontact, Perplexity, and Rundeck credential files so they now use the RFC-style default outbound User-Agent.

pmboomer:
Happy to test this branch against our WAF‑protected WooCommerce and WordPress sites if that’s helpful. This would unblock several workflows for us.

BerniWittmann (approved):
Great work, thank you very much 🙌

muhammadosama984:
@BerniWittmann 
I noticed a few checks were missing in this PR. Let me know if you want me to help add them.

Joffcom:
We may need to revert the changes for nodes that specifically set the value as these were added specially for tracking those values.

We will need to let people know about this change as it could be breaking for them if we don't plan to revert the node specific options.

BerniWittmann:
@Joffcom you're raising a good point about compatibility

I think most of them are just removing the expliciteness (`n8n` user agents did not differ from default). Changing the default user agent should only affect WAF rules (other rules) where there is an exact matching on the UA, substring matches should not break. 

However, I found that this PR also changes some Nodes that used different than the default `n8n` UA aka should definitely be reverted
- HaloPSA Node (UA: `https://n8n.io`)
- OnFleet Node (UA: `n8n-onfleet`)
- OpenThesaurus Node (UA: `https://n8n.io`)
- Plivo Node (UA: `plivo-n8n`)

@muhammadosama984: Maybe it would also help if you squash your commits into one with a clear commit message about changing the default User Agent for the changelog for better indicating this change.

packages/nodes-base/nodes/HaloPSA/GenericFunctions.ts
```diff
@@ -73,7 +73,6 @@ export async function haloPSAApiRequest(
 		let options: IRequestOptions = {
 			headers: {
 				Authorization: `Bearer ${accessToken}`,
-				'User-Agent': 'https://n8n.io',
```
BerniWittmann, line 76: specific user agent

packages/nodes-base/nodes/Onfleet/GenericFunctions.ts
```diff
@@ -27,7 +27,6 @@ export async function onfleetApiRequest(
 	const options: IRequestOptions = {
 		headers: {
 			'Content-Type': 'application/json',
-			'User-Agent': 'n8n-onfleet',
```
BerniWittmann, line 30: specific user agent

packages/nodes-base/nodes/Onfleet/Onfleet.node.ts
```diff
@@ -131,7 +131,6 @@ export class Onfleet implements INodeType {
 				const options: IRequestOptions = {
 					headers: {
 						'Content-Type': 'application/json',
-						'User-Agent': 'n8n-onfleet',
```
BerniWittmann, line 134: specific user agent

packages/nodes-base/nodes/OpenThesaurus/GenericFunctions.ts
```diff
@@ -21,9 +21,6 @@ export async function openThesaurusApiRequest(
 ): Promise<any> {
 	try {
 		let options = {
-			headers: {
-				'User-Agent': 'https://n8n.io',
```
BerniWittmann, line 25: specific user agent

packages/nodes-base/nodes/Plivo/GenericFunctions.ts
```diff
@@ -25,9 +25,6 @@ export async function plivoApiRequest(
 	}>('plivoApi');
 
 	const options: IRequestOptions = {
-		headers: {
-			'user-agent': 'plivo-n8n',
```
BerniWittmann, line 29: specific user agent

BerniWittmann (changes requested):
I have marked the nodes/changes where an explicitly (non default) user agent was set

muhammadosama984:
@BerniWittmann All review comments have been addressed:

- **Restored specific custom User-Agent values** for the four nodes you flagged — HaloPSA (`https://n8n.io`), Onfleet (`n8n-onfleet`), OpenThesaurus (`https://n8n.io`), and Plivo (`plivo-n8n`) — these were incorrectly removed in the previous iteration and are now back to their original values.
- **Added a test** covering the `catch` fallback branch in `readN8nVersion` (the 1 missing line flagged by Codecov).
- **Squashed all commits into one** with a clear commit message for the changelog.

Would appreciate a re-review when you get a chance. Thank you!

BerniWittmann:
Hi @muhammadosama984 
We had an internal discussion on how to approach this, especially focusing on not breaking stuff for users who rely on the n8n user agent and tracking of requests coming from n8n. 
So i will close this pr in favor of https://github.com/n8n-io/n8n/pull/28771
But dont worry, I list you as Co-Author, for it to show up as your contribution. Thanks for all the effort 🙌

PR #15048, opened 2025-05-02
feat(Calendly Trigger Node): Deprecation notice for apiKey authentication

michael-radency:
## Summary

Calendly planning to deprecate API v1(apiKey authentication) in May 2025.

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/NODE-2571/blackcloak-calendly-node-fully-migrate-to-api-v2-before-end-of-may

## Review / Merge checklist

- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

PR #28367, opened 2026-04-12
feat(Calendly Trigger Node): Deprecate API key auth and default to OAuth2

SAY-5:
Calendly discontinued API key authentication on May 31, 2025. This updates the Calendly Trigger node to reflect that:

- The deprecation notice now uses past tense since the deadline has passed
- The API Key option in the auth dropdown is marked as \"(deprecated)\"
- New nodes default to OAuth2 instead of the broken API key method

Refs #15048


garritfra:
@Joffcom is this a breaking change that needs to be versioned?

Joffcom:
@garritfra yeah, it is changing a default value so if it was never changed it will break. This is one of those special cases though where if api keys don't work at all it would be ok. 

What I would do if API keys no longer work is version it and remove apikey as a selection so it only has OAuth.

SAY-5:
The failing `Unit tests / Nodes Unit Tests` is in `nodes/Google/GSuiteAdmin/test/group/delete.workflow`, unrelated to this Calendly change. The PR title check ran on an earlier title revision; the current title `feat(Calendly Trigger Node): ...` matches the same scope as the previously merged #15048.

garritfra (approved):

PR #28387, opened 2026-04-13
fix(editor): Only show role assignment warning modal when value actually changed

cstuncsik:
## Summary

The role assignment confirmation modal in SSO settings (SAML/OIDC) was appearing on every save, even when the Role Assignment dropdown hadn't changed. This was caused by `shouldPromptUserToConfirmUserRoleProvisioningChange()` mixing SSO login toggle state with provisioning modal concerns.

Replaces the old function with a reactive `roleAssignmentTransition` computed that purely compares the current dropdown value against the persisted config, returning `'none'`, `'backup'`, or `'switchToManual'`. Also updates modal copy per spec and adds cancel-revert behavior.

## How to verify

1. Go to **Settings → SSO** (SAML or OIDC)
2. With SSO already enabled and a role assignment method already selected, click **Save** without changing the Role Assignment dropdown → **No modal should appear**, form saves directly with a success toast
3. Change Role Assignment from "Assigned manually in n8n" to any SSO option (e.g. "Instance roles") → Click Save → **"Change role assignment method" modal** should appear with CSV download options
4. Change Role Assignment between two SSO options (e.g. "Instance roles" → "Instance and project roles") → Click Save → **Same backup modal** should appear
5. Change Role Assignment from any SSO option back to "Assigned manually in n8n" → Click Save → **"Manage user role provisioning in n8n only" modal** should appear (no CSV downloads, just checkbox confirmation)
6. In any modal, click **Cancel** → The dropdown should **revert to its previously saved value**

https://linear.app/n8n/issue/IAM-535

## Review / Merge checklist

- [x] Tests included.


cstuncsik:
Re: cubic-dev-ai review #2 (P1 on `show-project-roles-csv` blocking switchToManual):

Not valid. The `showProjectRolesCsv` prop has no effect on the switchToManual flow. In `ConfirmProvisioningDialog.vue`:

1. The CSV download section is inside `v-if="!isSwitchingToManual"`, so it never renders for manual transitions
2. The checkbox disabled condition starts with `!isSwitchingToManual &&`, so it is always enabled
3. The confirm button has the same guard

The prop is only consumed in the backup modal path, which is correct.

guillaumejacquart (approved):
I'm approving it, even though it seems it does something "smarter" than the specs to choose whether or not to display the roles csv download

PR #28389, opened 2026-04-13
chore: Migrate @testing-library/jest-dom to support Vitest in @n8n/composables

Matsuuu:
## Summary

As a part of the jest -> vitest migration, clean up unnecessary dependencies from @n8n/composables

https://linear.app/n8n/issue/CAT-2796

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/@n8n/composables/src/__tests__/setup.ts
```diff
@@ -1,4 +1,3 @@
-import '@testing-library/jest-dom';
```
alexgrozav, line 1: This library provides a lot of DOM utilities which we will likely use in the future.

Looks like there's a new way to import it: https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-vitest

Can we update to that?

alexgrozav (changes requested):

alexgrozav (approved):

PR #28407, opened 2026-04-13
fix(ai-builder): Use placeholders for user-provided values instead of hardcoding fake addresses

aalises:
## Summary

When a user asks the AI builder to create a workflow involving email (e.g. *"send me an email if it's going to rain"*), the builder hardcodes `sendTo: "user@example.com"` instead of collecting the real address. This happens because the builder prompt bans `placeholder()` universally — a rule designed for discoverable resources (spreadsheet IDs, calendar IDs) that incorrectly blocks user-provided values too.

This PR fixes the issue across four layers:

**Prompt semantics**: Scopes the `placeholder()` ban to discoverable resources only. The builder can now use `placeholder('Your email address')` for values that require direct user input, and is explicitly told never to hardcode fake addresses.

**Workflow-loop state machine**: Adds `hasUnresolvedPlaceholders` as first-class metadata alongside `mockedCredentialTypes`. Placeholder-bearing workflows skip runtime verification (which would fail on sentinel strings) and route to `setup-workflow` so the user can provide real values through the setup wizard.

**Setup service**: `buildSetupRequests()` now recursively scans node parameters for placeholder sentinels and surfaces them as parameter issues, ensuring the setup wizard shows input fields for them.

**Frontend setup wizard**: `isParamValueSet()` treats placeholder strings as "not set". `isCardComplete()` checks for remaining placeholders. `buildNodeParameters()` explicitly clears placeholder sentinels on apply so they don't silently survive.

[Notion P0: Didn't ask for email as part of the configuration steps](https://www.notion.so/n8n/33e5b6e0c94f80a79534e6f7f9c4dd65)

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
+++ b/packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
@@ -13,13 +13,23 @@ import {
 	WORKFLOW_SDK_PATTERNS,
 } from '../../workflow-builder';
 
+// ── Shared placeholder guidance (single source of truth) ────────────────────
+
+// prettier-ignore
+const PLACEHOLDER_RULE =
+	"**Do NOT use `placeholder()` for discoverable resources** (spreadsheet IDs, calendar IDs, channel IDs, folder IDs) — resolve real IDs via `explore-node-resources` or create them via setup workflows. For **user-provided values** that cannot be discovered or created (email recipients, phone numbers, custom URLs, notification targets), use `placeholder('descriptive hint')` so the setup wizard prompts the user after the build. Never hardcode fake values like `user@example.com`.";
+
+// prettier-ignore
+const PLACEHOLDER_ESCALATION =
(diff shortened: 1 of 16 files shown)


packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
```diff
@@ -13,13 +13,23 @@ import {
...
 } from '../../workflow-builder';
 
+// ── Shared placeholder guidance (single source of truth) ────────────────────
+
+// prettier-ignore
+const PLACEHOLDER_RULE =
+	"**Do NOT use `placeholder()` for discoverable resources** (spreadsheet IDs, calendar IDs, channel IDs, folder IDs) — resolve real IDs via `explore-node-resources` or create them via setup workflows. For **user-provided values** that cannot be discovered or created (email recipients, phone numbers, custom URLs, notification targets), use `placeholder('descriptive hint')` so the setup wizard prompts the user after the build. Never hardcode fake values like `user@example.com`.";
+
+// prettier-ignore
+const PLACEHOLDER_ESCALATION =
```
Cadiac, line 23: This seems to be pretty email-specific language, while we'd want this to work with all kinds of user identifiers - but maybe something to improve once we have proper evaluations in place to actually see the impact of different language here?

Cadiac (approved):

PR #28416, opened 2026-04-13
fix(core): Fix workflow-sdk validation for plain workflow objects

### LINKED ISSUE #28051: MCP update_workflow / validate_workflow fail with "builder.regenerateNodeIds is not a function"
**Environment**
- n8n Cloud version: 2.14.2
- MCP server endpoint: `https://[instance].app.n8n.cloud/mcp-server/http`
- Integration: Claude AI via MCP connector (Claude.ai)

**Description**

All three workflow modification tools exposed via the n8n MCP server fail consistently with the same internal error:

```
Failed to parse generated workflow code: builder.regenerateNodeIds is not a function
```

This affects:
- `validate_workflow`
- `update_workflow`
- `create_workflow_from_code`

The error occurs at the internal SDK parsing stage, before any workflow logic is evaluated. It is not related to the code format or content — the error reproduces with the most minimal possible input.

**Steps to Reproduce**

Call `validate_workflow` via MCP with the following minimal code:

```js
const myFlow = {
  name: "TEST",
  nodes: [
    {
      id: "sticky-test",
      name: "Test Note",
      type: "n8n-nodes-base.stickyNote",
      typeVersion: 1,
      position: [100, 100],
      parameters: { content: "Hello", height: 200, width: 300, color: 3 }
    }
  ],
  connections: {}
};
export default myFlow;
```

**Expected Result**
Workflow validates successfully and returns the parsed workflow JSON.

**Actual Result**
```json
{
  "valid": false,
  "errors": [
    "Failed to parse generated workflow code: builder.regenerateNodeIds is not a function"
  ]
}
```

**Additional Notes**
- The error is 100% reproducible across all code formats tested (with/without imports, with/without `export default`, plain objects, single nodes, multiple nodes)
- Tested before and after upgrading to v2.14.2 — bug persists
- Read-only MCP tools (`search_workflows`, `get_workflow_details`, `get_execution`) work correctly — only write tools are affected
- The `builder` object referenced in the error message appears to be missing the `regenerateNodeIds` method in the version of the Workflow SDK bundled with the MCP server

**Workaround**
None available via MCP. Workflow modifications must be done manually through the n8n UI.

MiloradFilipovic:
## Summary
- Extend `parseWorkflowCodeToBuilder` to support JSON workflow objects or throw a meaningful error otherwise.
- Introducing `WorkflowCodeParseError` error and using it's hints in MCP tool response to nudge MCP clients to use proper SDK syntax before retrying

**Context**: We had reports of users trying to call MCP tools that use this method by passing JSON instead of SDK code and getting cryptic errors (see attached community issue). My hunch is that this is caused by MCP clients not consulting SDK reference tool/resource and hallucinating the syntax when building workflows. This PR makes validation work with JSON objects but also, more importantly, gives hints on how to recover if parsing error is thrown.

Fixes ADO-5051
Fixes #28051

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/mcp/tools/workflow-builder/validate-workflow-code.tool.ts
```diff
@@ -104,9 +105,12 @@ export const createValidateWorkflowCodeTool = (
 			};
 			telemetry.track(USER_CALLED_MCP_TOOL_EVENT, telemetryPayload);
 
+			const hint = getSdkReferenceHint(error);
+
 			const output = {
 				valid: false,
 				errors: [errorMessage],
+				...(hint ? { hint } : {}),
```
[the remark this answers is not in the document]
MiloradFilipovic: Ah true, forgot about that

packages/@n8n/workflow-sdk/src/codegen/parse-workflow-code.ts
```diff
@@ -651,3 +652,33 @@ export function parseWorkflowCodeToBuilder(code: string): WorkflowBuilder {
...
+		result !== null &&
+		typeof (result as WorkflowBuilder).regenerateNodeIds === 'function'
+	) {
+		return result as WorkflowBuilder;
+	}
+
+	if (
+		typeof result === 'object' &&
+		result !== null &&
+		Array.isArray((result as Record<string, unknown>).nodes)
```
mutdmour: not sure why `as Record<string, unknown>` is needed?

packages/@n8n/workflow-sdk/src/codegen/parse-workflow-code.ts
```diff
@@ -651,3 +652,33 @@ export function parseWorkflowCodeToBuilder(code: string): WorkflowBuilder {
...
+	) {
+		return workflowFn.fromJSON(result as WorkflowJSON);
```
mutdmour: `as WorkflowJSON` is better set by a typeguard

if (isWorkflowJSON(result)) { reutrn workflowFn.fromJSON(result) }

packages/@n8n/workflow-sdk/src/codegen/parse-workflow-code.ts
```diff
@@ -651,3 +652,33 @@ export function parseWorkflowCodeToBuilder(code: string): WorkflowBuilder {
...
+ * - If the result is a plain object that looks like WorkflowJSON (has a `nodes` array), convert it via `workflow.fromJSON()`.
+ * - Otherwise, throw with a descriptive error.
+ */
+function asWorkflowBuilder(result: unknown): WorkflowBuilder {
+	if (
+		typeof result === 'object' &&
+		result !== null &&
+		typeof (result as WorkflowBuilder).regenerateNodeIds === 'function'
+	) {
+		return result as WorkflowBuilder;
```
mutdmour: better set as a typeguard `if (isWorkflowBuilder(result)) { return result }`

packages/@n8n/workflow-sdk/src/codegen/parse-workflow-code.ts
```diff
@@ -651,3 +652,33 @@ export function parseWorkflowCodeToBuilder(code: string): WorkflowBuilder {
...
+		typeof result === 'object' &&
+		result !== null &&
+		Array.isArray((result as Record<string, unknown>).nodes)
+	) {
+		return workflowFn.fromJSON(result as WorkflowJSON);
+	}
+
+	throw new SyntaxError(
+		'Code must export a workflow built with the workflow() SDK function, ' +
+			'or a valid workflow JSON object with a "nodes" array.',
```
mutdmour: Maybe the JSON note here is confusing for the agent. Better have it reach for the tool instead?

packages/@n8n/workflow-sdk/src/codegen/parse-workflow-code.ts
MiloradFilipovic: I don't think it is, will remove

packages/@n8n/workflow-sdk/src/codegen/parse-workflow-code.ts
MiloradFilipovic: Yeah, let's use same hint mechanism used for `WorkflowCodeParseError`

mutdmour (approved):

PR #28417, opened 2026-04-13
refactor(ai-builder): Move inline SDK patterns to on-demand discovery (no-changelog)

aalises:
## Summary

Improves builder prompt discoverability by moving ~350 lines of inline SDK code patterns out of the builder system prompt and into on-demand discovery mechanisms that the builder already uses.

### Benefits

- **~350 fewer tokens per builder invocation**: SDK code patterns (AI agents, data tables, Google Sheets, web app SPA, batch processing, parallel execution, etc.) are no longer embedded in the system prompt on every build. This reduces prompt size, improving focus and reducing cost.
- **Zero extra tool calls for node-specific guidance**: patterns for IF/Switch/Filter conditions, Data Table rules, SplitInBatches loops, etc. are now delivered via `@builderHint` annotations on the nodes themselves. The builder already calls `get-node-type-definition` and `search-nodes`, which surface these hints automatically.
- **Single source of truth**: node configuration guidance now lives with the node (via `@builderHint`), not duplicated in a prompt that can drift. New nodes with `@builderHint` annotations just work without needing prompt updates.
- **Workflow composition patterns available on demand**: the builder can now call `get-best-practices` (previously planner-only) to fetch code examples for complex workflow shapes when it needs them, rather than carrying all examples in context at all times.
- **Better discoverability**: enriched tool descriptions for `delegate`, `setup-workflow`, `publish-workflow`, and `test-credential` surface usage guidance at tool-load time instead of burying it in the system prompt.

### What changed

**Node hints (nodes-base):**
- Added `@builderHint` to Switch V3 (`rules.values` not `rules.rules` crash prevention)
- Added `@builderHint` to Filter V2 (conditions structure with `options`, `conditions`, `combinator`)
- Added `@builderHint` to SplitInBatches V3 (SDK factory pattern with `.onDone()` / `.onEachBatch()`)

**New best-practice guides (workflow-sdk):**
- `workflow_composition` — parallel execution, batch processing, fan-in, executeOnce, alwaysOutputData
- `ai_agent_patterns` — agents with tools, fromAi(), structured output, subnode factories
- `web_app` — file-based HTML, data injection, multi-route SPA pattern

**Builder prompt (instance-ai):**
- Removed `WORKFLOW_SDK_PATTERNS` from the builder prompt (all tagged pattern blocks)
- Removed inline examples: AI Agent subnodes, Code Node, Data Table, Google Sheets, Parallel Branches, Batch Processing, Multiple Triggers, Web App SPA full dashboard example, Google Sheets RLC
- Replaced with compact references to `get-best-practices` and `@builderHint`
- Strengthened `@builderHint` emphasis: "IMPORTANT: Always read @builderHint annotations"
- Updated both mandatory processes (tool-mode and sandbox) to reference `get-best-practices`
- Added `get-best-practices` to both builder tool sets (was planner-only)
- Fixed stale `web_app_pattern` references and `search-credential-types` reference

**Tool descriptions (instance-ai):**
- `delegate` — statelessness note
- `setup-workflow` — `deferred: true` behavior
- `publish-workflow` — user-only + sub-workflow exception
- `test-credential` — "call before referencing"
- Removed duplicated delegate guidance from system prompt


packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
```diff
@@ -43,362 +42,19 @@ const SDK_CODE_RULES = `## SDK Code Rules
...
-documentId: { __rl: true, mode: 'name', value: 'Sales Pipeline' },
-
-// WRONG — plain string
-documentId: 'YOUR_SPREADSHEET_ID',  // Not an RLC object
-
-// WRONG — expr() wrapper
-documentId: expr('{{ "spreadsheetId" }}'),  // RLC fields don't use expressions
-\`\`\`
-Always use the IDs from \`explore-node-resources\` results inside the RLC \`value\` field.
+For node-specific configuration patterns (Data Table, Google Sheets, HTTP Request auth, Code node, AI agent subnodes), read the @builderHint annotations from \`get-node-type-definition\` results.
```
mutdmour: is this needed at all? I mean it's read as part of the context when builder loads node definition. we have two reminders above.

packages/nodes-base/nodes/Switch/V3/SwitchV3.node.ts
```diff
@@ -59,6 +59,11 @@ export class SwitchV3 implements INodeType {
 			},
 			inputs: [NodeConnectionTypes.Main],
 			outputs: `={{(${configuredOutputs})($parameter)}}`,
+			builderHint: {
+				message: `The inner key for routing rules MUST be rules.values (NOT rules.rules). Using rules.rules crashes with "Could not find property option".
+Each rule needs an outputKey string and a complete conditions object with options, conditions array, and combinator.
+e.g.: { "rules": { "values": [{ "outputKey": "email", "conditions": { "combinator": "and", "options": { "caseSensitive": true, "leftValue": "", "typeValidation": "strict" }, "conditions": [{ "leftValue": "={{ $json.type }}", "operator": { "type": "string", "operation": "equals" }, "rightValue": "email" }] } }] } }`,
```
mutdmour, line 65: Can we investigate why SDK does not catch this as a type error? Instead of adding a builder hint. Should `rules.values` be required. Or that rules.rules does not match structure as extra key and should return warning back to agent.

packages/nodes-base/nodes/Filter/V2/FilterV2.node.ts
```diff
@@ -29,6 +29,13 @@ export class FilterV2 implements INodeType {
 			outputs: [NodeConnectionTypes.Main],
 			outputNames: ['Kept', 'Discarded'],
 			parameterPane: 'wide',
+			builderHint: {
+				message: `parameters.conditions must always contain these three sibling keys:
+- "combinator": "and" or "or", default to "and"
+- "conditions": [ {a list of condition objects} ]
+- "options": { "caseSensitive": true, "leftValue": "", "typeValidation": "strict" }
+Missing any of these crashes at runtime with "Cannot read properties of undefined (reading 'caseSensitive')".`,
```
mutdmour, line 37: this should not be a hint. this should be baked into the sdk, and a wanring returned to the agent. since this worked before, I assume there is a bug with the new agent.

packages/@n8n/workflow-sdk/src/prompts/best-practices/index.ts
```diff
@@ -52,4 +58,7 @@ export const bestPracticesRegistry: Record<
 	[WorkflowTechnique.NOTIFICATION]: new NotificationBestPractices(),
 	[WorkflowTechnique.KNOWLEDGE_BASE]: undefined,
 	[WorkflowTechnique.HUMAN_IN_THE_LOOP]: undefined,
+	[WorkflowTechnique.WORKFLOW_COMPOSITION]: new WorkflowCompositionBestPractices(),
+	[WorkflowTechnique.AI_AGENT_PATTERNS]: new AiAgentPatternsBestPractices(),
+	[WorkflowTechnique.WEB_APP]: new WebAppBestPractices(),
```
mutdmour: I believe these are the first one with SDK code. But this best practice registry is also consumed by the planning agent, which would be confused by the SDK code.
Maybe we need to split these?

packages/@n8n/workflow-sdk/src/prompts/best-practices/guides/workflow-composition.ts
mutdmour, line 1: this is where having a good eval set would be helpful to evaluate this change

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.tool.ts
```diff
@@ -189,6 +189,7 @@ export async function startBuildWorkflowAgentTask(
 		const toolNames = [
 			'search-nodes',
 			'get-suggested-nodes',
+			'get-best-practices',
```
mutdmour: I don't think we should add this tool to the builder agent. Especially without evaluations. We should stick to the formula we have optimized before.
Planning agent reads best practices (notifications, knowledge base, etc), none of which was optimized for the builder. Then the plan is sent to the builder. 
It also confuses with `get suggested nodes` which is a distilled version of those best practices.

aalises:
@mutdmour:

- Removed the redundant hint reminder at lines 56-57. The two earlier mentions are sufficient.
- Removed the 3 SDK-heavy guides (workflow-composition, ai-agent-patterns, web-app) from the registry and deleted the files. These contained SDK code that would confuse the planning agent.
- Removed `get-best-practices` from both builder tool lists (sandbox and non-sandbox). Restored `WORKFLOW_SDK_PATTERNS` inline in the builder prompt. Cleaned up all `get-best-practices` references in the builder prompt and submit-workflow tool.

Leaving as a follow-up the SDK-level validation (why it failed). Leaving the hints as a safety net for now...

packages/nodes-base/nodes/Switch/V3/SwitchV3.node.ts
mutdmour: @aalises I think this is worth investigating.
1. it's easy to forget this builder hint
2. if our validation is not getting applied correctly, it points to a deeper problem with the builder
I can take a look at this if you want.

aalises: Yeah I think this is some builder issue that'd be worth investigating.. and true that the builder hints are sometimes forgotten by the agents 😆

packages/@n8n/ai-workflow-builder.ee/src/tools/validate-configuration.tool.ts
mutdmour, line 1: can you please add some tests for this?

packages/@n8n/ai-workflow-builder.ee/src/tools/validate-configuration.tool.ts
```diff
@@ -51,19 +53,45 @@ export function createValidateConfigurationTool(
 				const fromAiViolations = validateFromAi(workflow, parsedNodeTypes);
 				const parameterViolations = validateParameters(workflow, parsedNodeTypes);
 
+				// Run workflow-sdk graph validators (includes filterNodeValidator
+				// which catches fixedCollection key mistakes like rules.rules vs rules.values)
```
mutdmour, line 57: comment is not useful here outside of the context of the switch node bug. can clarify this more or delete

packages/@n8n/ai-workflow-builder.ee/src/tools/validate-configuration.tool.ts
```diff
@@ -51,19 +53,45 @@ export function createValidateConfigurationTool(
 				const fromAiViolations = validateFromAi(workflow, parsedNodeTypes);
 				const parameterViolations = validateParameters(workflow, parsedNodeTypes);
 
+				// Run workflow-sdk graph validators (includes filterNodeValidator
```
mutdmour, line 56: I tripled check this. this tool does not seem to be imported or actually used. Can you clarify how this fixes anything?

PR #28421, opened 2026-04-13
fix(core): Position workflow correctly if opened while on a background tab (no-changelog)

Cadiac:
## Summary

On instance AI artifacts if the tab was on background while WF generation finished we failed to position the canvas correctly.
Add logic to reposition it correctly when visibility changes in that case. This shouldn't affect the regular canvas view in negative / noticeable way.

[Notion: Workflow artifact does zoom/fit wrong](https://www.notion.so/n8n/aef5b6e0c94f82159cc58164aa417607?v=7965b6e0c94f823c918f88516500cb86&p=33d5b6e0c94f802c914efc6f22bd6b64&pm=s)

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/workflows/canvas/components/Canvas.vue
```diff
@@ -720,6 +720,9 @@ async function onFitBounds(nodes: GraphNode[]) {
 }
 
 async function onFitView() {
+	if (document.hidden) {
+		fitViewWhileHidden = true;
+	}
 	await fitView({ maxZoom: defaultZoom, padding: 0.2 });
```
aalises, line 727: `fitView()` still runs even when `document.hidden` is true. I would early-return after setting the flag to avoid the unnecessary work? I think `onVisibilityChange.` will already compute the view correctly, but haven't tested

```ts
async function onFitView() {
	if (document.hidden) {
		fitViewWhileHidden = true;
		return;
	}
	await fitView({ maxZoom: defaultZoom, padding: 0.2 });
}
```

Cadiac: I'll have a look at this

Cadiac: Good suggestion 👍

aalises (approved):

PR #28422, opened 2026-04-13
fix(core): Handle git fetch failure during source control startup

### LINKED ISSUE #28311: Environments: expired Git HTTPS token prevents n8n startup after restart
### Bug Description

We use self-hosted n8n in Kubernetes and are testing the Environments feature: [Understand environments](https://docs.n8n.io/source-control-environments/understand/environments/)

We configured the Git source according to the docs using HTTPS: [Set up Git source](https://docs.n8n.io/source-control-environments/setup/)

After some time, the GitHub token used for the HTTPS remote expired. Once n8n-web restarted, it failed to start because it tried to connect to the Git remote during startup and exited on authentication failure.

This created a recovery deadlock: the UI was unavailable because n8n-web would not boot, and the only way to recover was to edit the database directly to remove or fix the stored Git configuration.

Relevant error:

```
remote: Invalid username or token. Password authentication is not supported for Git operations.
fatal: Authentication failed for 'https://github.com/[redacted]'
```
Full startup log excerpt:

```
{"level":"error","message":"Error: remote: Invalid username or token. Password authentication is not supported for Git operations.\nfatal: Authentication failed for 'https://github.com/[redacted]'\n\n    at Object.action (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/simple-git@3.28.0/node_modules/simple-git/src/lib/plugins/error-detection.plugin.ts:42:29)\n    at PluginStore.exec (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/simple-git@3.28.0/node_modules/simple-git/src/lib/plugins/plugin-store.ts:54:29)\n    at /usr/local/lib/node_modules/n8n/node_modules/.pnpm/simple-git@3.28.0/node_modules/simple-git/src/lib/runners/git-executor-chain.ts:124:42\n    at new Promise (<anonymous>)\n    at GitExecutorChain.handleTaskData (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/simple-git@3.28.0/node_modules/simple-git/src/lib/runners/git-executor-chain.ts:121:14)\n    at GitExecutorChain.attemptRemoteTask (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/simple-git@3.28.0/node_modules/simple-git/src/lib/runners/git-executor-chain.ts:97:40)\n    at processTicksAndRejections (node:internal/process/task_queues:103:5)\n    at GitExecutorChain.attemptTask (/usr/local/lib/node_modules/n8n/node_modules/.pnpm/simple-git@3.28.0/node_modules/simple-git/src/lib/runners/git-executor-chain.ts:61:18)","metadata":{"file":"start.js","function":"catch","timestamp":"2026-04-03T08:34:27.230Z"}}
{"level":"error","message":"Exiting due to an error.","metadata":{"file":"error-reporter.js","function":"defaultReport","timestamp":"2026-04-03T08:34:27.231Z"}}
{"level":"error","message":"remote: Invalid username or token. Password authentication is not supported for Git operations.\nfatal: Authentication failed for 'https://github.com/[redacted]'\n","metadata":{"file":"error-reporter.js","function":"defaultReport","timestamp":"2026-04-03T08:34:27.231Z"}}
```

### To Reproduce

1. Run self-hosted n8n in Kubernetes.
2. Enable and configure the Environments feature with a GitHub remote over HTTPS, using a token for authentication.
3. Let the token expire, revoke it, or otherwise make it invalid.
4. Restart the n8n-web instance/pod.
5. Observe that n8n-web fails during startup because Git authentication fails.
6. Observe that the instance cannot be recovered through the UI because the UI never becomes available.

### Expected behavior

If the configured Git remote becomes temporarily invalid, n8n-web should still start.

Invalid Git credentials should be handled as a recoverable error, for example by:

- starting normally and surfacing the Git problem in the UI/logs
- marking the source control connection as unhealthy
- allowing the user to update credentials or disconnect the Git source without direct database edits

It should not be possible for an expired Git token to block the whole instance from starting.

### Debug Info

# Debug info

## core

- n8nVersion: 2.11.3
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: postgres
- executionMode: scaling (multi-main)
- concurrency: -1

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: database

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/146.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-04-10T10:36:03.373Z

### Operating System

Linux (containerized deployment on Kubernetes/GKE)

### n8n Version

2.11.3

### Node.js Version

24.13.1

### Database

PostgreSQL

### Execution mode

queue

### Hosting

self hosted

alielkhateeb:
## Summary

> [!IMPORTANT]  
> I'd like to get more hands on this, I cannot really reproduce the issue with the steps provided because if someone has their HTTPs with git, a git folder must be already there, the way I reproduced this differed a bit from the ticket which is why I need someone to validate this with me please after watching the video below 👇 

https://www.loom.com/share/c9844d430fb140188f1a1cfb91addc1a

When n8n starts with source control (Environments) configured over HTTPS and the local git folder is absent or incomplete (e.g. fresh pod, ephemeral filesystem, or manually deleted folder), it calls `initRepository()` → `trackRemoteIfReady()` → `fetch()`. If the HTTPS token is expired or revoked, the authentication error was unhandled and propagated all the way up, crashing the process on startup.

This created a recovery deadlock: the UI never became available, so users couldn't fix the credentials through the UI and had to edit the database directly.

**Root cause:** `trackRemoteIfReady()` called `fetch()` without a try-catch, unlike the sibling method `ensureBranchSetup()` which already handles fetch errors gracefully.

**Fix:** Wrap the `fetch()` call in `trackRemoteIfReady()` with the same try-catch pattern used in `ensureBranchSetup()` — log a warning and return early, letting `sanityCheck()` surface the error once the instance is up.

**How to reproduce:**
1. Configure source control with a GitHub HTTPS remote and token
2. Revoke the token
3. Delete the local git folder (`~/.n8n/gitFolder` or equivalent)
4. Restart n8n — observe startup crash
5. With this fix: n8n starts, logs a warning, and the user can fix credentials via the UI

https://linear.app/n8n/issue/LIGO-453
closes https://github.com/n8n-io/n8n/issues/28311

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/modules/source-control.ee/__tests__/source-control-git.service.test.ts
+++ b/packages/cli/src/modules/source-control.ee/__tests__/source-control-git.service.test.ts
@@ -78,7 +78,7 @@ describe('SourceControlGitService', () => {
 		});
 
 		describe('when fetch fails during remote tracking setup', () => {
-			it('should log warning and not throw', async () => {
+			it('should log warning and not throw when tracking fetch failures are tolerated', async () => {
 				const mockLogger = mock<any>();
 				const gitService = new SourceControlGitService(mockLogger, mock(), mock());
 				const prefs = mock<SourceControlPreferences>({ branchName: 'main' });
@@ -90,12 +90,32 @@ describe('SourceControlGitService', () => {
 				const fetchError = new Error('Authentication failed for HTTPS remote');
 				jest.spyOn(gitService, 'fetch').mockRejectedValue(fetchError);
 
-				await gitService.initRepository(prefs, user);
+				await gitService.initRepository(prefs, user, {
+					tolerateTrackingFetchFailure: true,
+				});
 
 				expect(mockLogger.warn).toHaveBeenCalledWith(
 					'Failed to fetch during remote tracking setup',
 					{ error: fetchError },
 				);
+			});
+
+			it('should throw when tracking fetch failures are NOT tolerated', async () => {
+				const gitService = new SourceControlGitService(mock(), mock(), mock());
+				const prefs = mock<SourceControlPreferences>({ branchName: 'main' });
+				const user = mock<User>();
+				const git = mock<SimpleGit>();
+				gitService.git = git;
+				jest.spyOn(gitService, 'setGitCommand').mockResolvedValue();
+
+				const fetchError = new Error('Authentication failed for HTTPS remote');
+				jest.spyOn(gitService, 'fetch').mockRejectedValue(fetchError);
+
+				await expect(
+					gitService.initRepository(prefs, user, {
+						tolerateTrackingFetchFailure: false,
+					}),
+				).rejects.toThrow(fetchError);
 			});
 		});
 
@@ -154,18 +174,18 @@ describe('SourceControlGitService', () => {
 				expect(addRemoteSpy).toHaveBeenCalledWith('origin', originUrl);
 			});
 
-			it('should log warning and not throw when HTTPS credentials are invalid during fetch', async () => {
+			it('should throw error when HTTPS connection type is specified but no credentials found', async () => {
 				const mockPreferencesService = mock<SourceControlPreferencesService>();
-				const mockLogger = mock<any>();
+				const errorMessage = 'Error';
 				mockPreferencesService.getDecryptedHttpsCredentials.mockRejectedValue(
-					new Error('Authentication failed'),
+					new Error(errorMessage),
 				);
 				mockPreferencesService.getPreferences.mockReturnValue({
 					connectionType: 'https',
 					repositoryUrl: 'https://github.com/user/repo.git',
 				} as never);
 
-				const gitService = new SourceControlGitService(mockLogger, mock(), mockPreferencesService);
+				const gitService = new SourceControlGitService(mock(), mock(), mockPreferencesService);
 				const prefs = mock<SourceControlPreferences>({
 					repositoryUrl: 'https://github.com/user/repo.git',
 					connectionType: 'https',
@@ -175,12 +195,7 @@ describe('SourceControlGitService', () => {
 				const git = mock<SimpleGit>();
 				gitService.git = git;
 
-				await gitService.initRepository(prefs, user);
-
-				expect(mockLogger.warn).toHaveBeenCalledWith(
-					'Failed to fetch during remote tracking setup',
-					expect.objectContaining({ error: expect.any(Error) }),
-				);
+				await expect(gitService.initRepository(prefs, user)).rejects.toThrow(errorMessage);
 				expect(mockPreferencesService.getDecryptedHttpsCredentials).toHaveBeenCalled();
 			});
 		});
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/source-control.ee/__tests__/source-control-git.service.test.ts
```diff
@@ -77,6 +77,28 @@ describe('SourceControlGitService', () => {
 			});
 		});
 
+		describe('when fetch fails during remote tracking setup', () => {
```
geemanjs, line 80: Question - I wonder if we we should write a test that exactly matches the situation they were in. 

This test would fail because of the "log not being present" but it'd be good to have a test that mimics the situation they were in where on startup the system threw an error and shut down when the git config was incorrect.

packages/cli/src/modules/source-control.ee/source-control-git.service.ee.ts
```diff
@@ -274,7 +274,12 @@ export class SourceControlGitService {
 	private async trackRemoteIfReady(targetBranch: string) {
 		if (!this.git) return;
 
-		await this.fetch();
+		try {
+			await this.fetch();
```
geemanjs, line 281: Is there any logic above this function that catches this? Or do we only call `trackRemoteIfReady` on startup?

packages/cli/src/modules/source-control.ee/__tests__/source-control-git.service.test.ts
geemanjs: nvm the test further down does that 👍

packages/cli/src/modules/source-control.ee/source-control-git.service.ee.ts
ireneea: Here are the three possible scenario when we call `trackRemoteIfReady`:
- on startup
- on connection
- on multi-main pub/sub event

On startup and event handling the error is not caught. With this fix,  we end up with a git folder that is not tracking a remote repo. This has no real consequence because any user initiated operation (push,pull,connect) has the right checks in place via calls to `sanityCheck()` to deal with that.

ireneea (dismissed):

geemanjs (dismissed):

packages/cli/src/modules/source-control.ee/source-control-git.service.ee.ts
alielkhateeb: Good question @geemanjs , and makes sense @ireneea 

I wonder if it would be safer to divert the outcome according to where we are calling this from? Here is a commit that I can revert 😄 if we all agree it's better to have full control I'll keep it, if we prefer silently failing on fetch (only warn) at all times I'll revert 🙏 a44303d13d2a8e5c059e80ba8f8a17d50e495c89

ireneea (approved):

PR #28423, opened 2026-04-13
fix(Schedule Node): Use elapsed-time check to self-heal after missed triggers

### LINKED ISSUE #23711: Schedule trigger permanently drops valid triggers after changing schedule

BerniWittmann:
## Summary

The `recurrenceCheck` function used strict equality (`===`) to determine if an interval-based Schedule Trigger should fire. If a single trigger window was missed (instance restart, leadership change, Redis failover, downtime), the `lastExecution` value in `staticData` became permanently stale and the check never passed again — silently blocking the workflow for up to 364 days.

This replaces the strict equality with an elapsed-time `>=` check:
```
// Before: only fires on the exact expected tick
dayOfYear === (intervalSize + lastExecution) % 365

// After: fires once enough time has elapsed (self-healing)
(dayOfYear - lastExecution + 365) % 365 >= intervalSize
```

Applied to all four interval types: hours, days, weeks, months. The weeks same-week re-trigger behavior is preserved.

This also fixes the schedule-change variant (NODE-4625, PR https://github.com/n8n-io/n8n/pull/25671) as a side effect — when a user changes the interval config, the stale `lastExecution` from the old schedule no longer causes a permanent block.

### How to test
1. Create a workflow with Schedule Trigger set to "Every 2 hours"
2. Activate and let it fire once
3. Deactivate, then set `recurrenceRules` in `staticData` to a stale value via DB
4. Reactivate — the workflow should self-heal and fire on the next cron tick

- https://linear.app/n8n/issue/NODE-4831
- https://linear.app/n8n/issue/NODE-4625
- fixes https://github.com/n8n-io/n8n/issues/23711
- Closes https://github.com/n8n-io/n8n/pull/25671
- Fixes https://github.com/n8n-io/n8n/issues/23711

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/nodes-base/nodes/Schedule/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/Schedule/GenericFunctions.ts
@@ -53,13 +53,13 @@ export function recurrenceCheck(
 	const momentTz = moment.tz(timezone);
 	if (typeInterval === 'hours') {
 		const hour = momentTz.hour();
-		if (lastExecution === undefined || hour === (intervalSize + lastExecution) % 24) {
+		if (lastExecution === undefined || (hour - lastExecution + 24) % 24 >= intervalSize) {
 			recurrenceRules[index] = hour;
 			return true;
 		}
 	} else if (typeInterval === 'days') {
 		const dayOfYear = momentTz.dayOfYear();
-		if (lastExecution === undefined || dayOfYear === (intervalSize + lastExecution) % 365) {
+		if (lastExecution === undefined || (dayOfYear - lastExecution + 365) % 365 >= intervalSize) {
 			recurrenceRules[index] = dayOfYear;
 			return true;
 		}
@@ -67,7 +67,7 @@ export function recurrenceCheck(
 		const week = momentTz.week();
 		if (
 			lastExecution === undefined || // First time executing this rule
-			week === (intervalSize + lastExecution) % 52 || // not first time, but minimum interval has passed
+			(week - lastExecution + 52) % 52 >= intervalSize || // not first time, but minimum interval has passed
 			week === lastExecution // Trigger on multiple days in the same week
 		) {
 			recurrenceRules[index] = week;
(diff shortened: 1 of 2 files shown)


dlavrenuek (approved):
Looks good to me 🚀

PR #28426, opened 2026-04-13
chore: Suppress warning from style lint

Cadiac:
## Summary

Our style linter doesn't like `--text-color--subtler`, but it's a token in our design system currently anyways 🤔 
Suppress the error from those lints for now, as it runs on commit hook and blocks commits without `--no-verify`.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/instanceAi/components/AgentActivityTree.vue
+++ b/packages/frontend/editor-ui/src/features/ai/instanceAi/components/AgentActivityTree.vue
@@ -109,6 +109,7 @@ function resolveArtifactName(artifact: ArtifactInfo): string {
 
 <style lang="scss" module>
 .reasoningTrigger {
+	/* stylelint-disable-next-line @n8n/css-var-naming -- design-system token */
 	color: var(--text-color--subtler);
 }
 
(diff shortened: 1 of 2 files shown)


aalises:
Instead of suppressing the warning inline, we could instead add `text-color` to `PROPERTY_VOCABULARY` in the stylelint rule? It's a legitimate design system token defined in `_tokens.scss` so the linter should recognize it.

```diff
// packages/@n8n/stylelint-config/src/rules/css-var-naming.ts
const PROPERTY_VOCABULARY = new Set([
	'color',
	'color--text',
	'color--background',
	'color--foreground',
+	'text-color',
```

r00gm (approved):

Cadiac:
> Instead of suppressing the warning inline, we could instead add `text-color` to `PROPERTY_VOCABULARY` in the stylelint rule? It's a legitimate design system token defined in `_tokens.scss` so the linter should recognize it.
> 
> ```diff
> // packages/@n8n/stylelint-config/src/rules/css-var-naming.ts
> const PROPERTY_VOCABULARY = new Set([
> 	'color',
> 	'color--text',
> 	'color--background',
> 	'color--foreground',
> +	'text-color',
> ```

We could if those `text-color*` are intentional tokens in the design system vocabulary, yesterday when I opened this I assumed this to actually just be a mistake on the design system side, but now after seeing @Tuukkaa 's comment on slack it seems like these are intentional and adjusting the rules seems like the right way to go.

r00gm:
agree but at this point, i don't know anymore what's a legitimate token and what isn't to be honest

alexgrozav (approved):
Approved temporarily until we figure it out.

PR #28430, opened 2026-04-13
fix(core): Resolve additional keys lazily in VM expression engine

despairblue:
## Summary

When `N8N_EXPRESSION_ENGINE=vm` is enabled, credential types using `IAuthenticateGeneric` with expression-based authentication (e.g. Supabase, Airtable, Notion) fail with 401 errors. The root cause: `buildContext()` in the VM isolate only registered a hardcoded set of known properties. Keys like `$credentials` — passed as `additionalKeys` during credential resolution — were invisible inside the isolate, so expressions like `={{$credentials.serviceRole}}` silently resolved to `undefined`.

Many other keys were also missing: `$request`, `$response`, `$pageCount`, `$self`, `$evaluateExpression`, `$jmesPath`, `$getPairedItem`, `$rawParameter`, `$agentInfo`, etc.

**Fix:** Replace the plain context object in `buildContext()` with a `Proxy` whose `has`/`get` traps lazily resolve unknown keys from the host via the existing `getValueAtPath` callback. Each key is resolved at most once per evaluation and cached on the proxy target.

This approach:
- Eliminates all hardcoded key lists — no more sync issues when new keys are added to the data proxy
- Removes 7 eager `fetchPrimitive()` round-trips and 4 `exposeHostFunction()` probes from context setup
- Keys already on the target (builtins, `DateTime`, `$now`, `$item`, `$`) are returned immediately with no round-trip
- Results in a **17-39% improvement** on expression engine benchmarks vs baseline (because setup is cheaper)
- Only file changed in the runtime: `context.ts`. No bridge changes needed.

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/expression-runtime/src/runtime/context.ts
+++ b/packages/@n8n/expression-runtime/src/runtime/context.ts
@@ -45,7 +45,7 @@ export function buildContext(
 	}
 	Settings.defaultZone = timezone ?? 'system';
 
-	const ctx: Record<string, unknown> = {};
+	const target: Record<string, unknown> = {};
 
 	// Callback bundle passed to createDeepLazyProxy so proxies don't touch globalThis
 	const callbacks = { getValueAtPath, getArrayElement, callFunctionAtPath };
@@ -53,7 +53,7 @@ export function buildContext(
 	// __sanitize must be on the context because PrototypeSanitizer generates:
 	// obj[this.__sanitize(expr)] where 'this' is the context (via .call(ctx) wrapping)
 	// Use a non-writable property descriptor so override attempts throw instead of silently succeeding.
-	Object.defineProperty(ctx, '__sanitize', {
+	Object.defineProperty(target, '__sanitize', {
 		get: () => __sanitize,
 		set: () => {
 			throw new ExpressionError('Cannot override "__sanitize" due to security concerns');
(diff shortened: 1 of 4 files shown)


packages/@n8n/expression-runtime/src/runtime/context.ts
```diff
@@ -183,11 +105,88 @@ export function buildContext(
...
+	// so each key is fetched at most once per evaluation.
+	// -------------------------------------------------------------------------
+
+	// Track keys we've already probed so we never call applySync twice
+	// for the same key — even if the host returned undefined.
+	const probedKeys = new Set<string>();
+
+	function resolveFromHost(key: string): boolean {
+		if (probedKeys.has(key)) return false;
+		probedKeys.add(key);
```
ivov: We might be caching too early here. This runs before the try, so if `applySync` throws once (anything transient) then the key is marked probed forever. Maybe move the `add` into the success paths after we know we have a definitive answer, either a resolved value or a confirmed `undefined` from the host.

ivov (commented):
Nice fix! More maintainable also. I pushed a bit of unit test coverage. One question on caching only.

packages/@n8n/expression-runtime/src/runtime/context.ts
despairblue: nice finde, fixed and amended: 97ec22b0341bda268a88180ec8af2afa5e63f953

ivov (approved):
🚀

Matsuuu (approved):

PR #28433, opened 2026-04-13
feat(agent-sdk): Publish Agents

bjorger:
## Summary

This PR adds the ability to publish and unpublish agents. 
Publishing captures a snapshot of the current agent config and marks it as the active version. 
Unpublishing removes that snapshot and deactivates the agent. 
The system also tracks whether there are unsaved changes between the current config and the last published version.

How to test:
Navigate to the agents page
image

On the top right, when working on an agent, there is now a publish button (replaced save & cancel)
Changes are now captured automatically

When the user leaves the page with unpublished changes, a modal pops up to reminds the user:
image

In the agents list, published agents have a "published" tag present:
image

In the agents list, the user can publish / unpublish the agent via the context menu:
image

https://linear.app/n8n/issue/AGENT-19/feature-publish-the-agent

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/agents/agents.service.ts
```diff
@@ -174,6 +180,46 @@ export class AgentsService {
...
+			schema: agent.schema,
+			model: agent.model,
+			provider: agent.provider,
+			credentialId: agent.credentialId,
+			publishedById: userId,
+		});
+
+		// Stamp the current versionId as published — frontend compares these to detect changes.
+		agent.activeVersionId = agent.versionId;
+		await this.agentRepository.save(agent);
```
DeveloperTheExplorer: the publishAgent method does two separate writes — `savePublishedVersion()` then `agentRepository.save(agent)`. If the second fails, you'd have a published version row but `activeVersionId` never gets stamped. Same issue in `unpublishAgent`. Should these be wrapped in a transaction?

packages/cli/src/modules/agents/agents.controller.ts
```diff
@@ -197,6 +197,24 @@ export class AgentsController {
...
+	}
+
+	@Post('/:agentId/unpublish')
+	async unpublish(
+		req: AuthenticatedRequest<{ projectId: string }>,
+		_res: Response,
+		@Param('agentId') agentId: string,
+	) {
+		return await this.agentsService.unpublishAgent(agentId, req.params.projectId);
+	}
```
DeveloperTheExplorer, line 217: Should we check for permissions? As in, prevent users from publishing/unpublishing other people's agents?

I'm okay leaving it after P0 but if it's missed and forgotten later, it could be a security problem.

packages/frontend/editor-ui/src/features/agents/views/AgentBuilderView.vue
```diff
@@ -99,31 +96,25 @@ function startChat(msg: string) {
...
 	if (!localConfig.value) return;
-	await updateConfig(projectId.value, agentId.value, localConfig.value);
-	originalConfigJson.value = JSON.stringify(localConfig.value);
-	isDirty.value = false;
+	const result = await updateConfig(projectId.value, agentId.value, localConfig.value);
+	// Keep agent.versionId in sync so hasUnpublishedChanges stays accurate
+	if (agent.value && result.versionId !== undefined) {
+		agent.value = { ...agent.value, versionId: result.versionId };
+	}
+	showMessage({ title: locale.baseText('agents.builder.toast.saved'), type: 'success' });
```
DeveloperTheExplorer: Is this a toast? Might be noisy to show a toast every time autosave happens. We could follow-suit with how it's done in workflow builder

DeveloperTheExplorer (changes requested):

packages/cli/src/modules/agents/agents.service.ts
bjorger: Wrapped in a transaction in my last commit :)

packages/cli/src/modules/agents/agents.controller.ts
bjorger: Created a ticket for it.

packages/@n8n/api-types/src/agents.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * The snapshot of an agent at publish time. Returned by publish/unpublish
+ * endpoints as part of the agent payload so the UI can derive publish state
+ * (`not-published` / `published-no-changes` / `published-with-changes`) from
+ * `agent.versionId` vs `publishedVersion.publishedFromVersionId`.
+ */
+export interface AgentPublishedVersionDto {
+	schema: unknown;
```
mike12345567: Can we type this based on the actual agent schema JSON?

mike12345567 (approved):
LGTM - one NAB about typing but not a blocker.

packages/@n8n/api-types/src/agents.ts
bjorger: Good catch!

PR #28440, opened 2026-04-14
perf(community-packages): make reinstall missing community packages async so it will not block app start

smeng9:
## Summary

This merge request addresses an issue in runtime environments with slow networks. 

Reinstalling community packages from code start takes a lot of time and may eventually cause the health check to timeout.

By making the installation asynchronous, the service is now able to respond to frontend request while completing the missing package installing in the background.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/community-packages/community-packages.service.ts
+++ b/packages/cli/src/modules/community-packages/community-packages.service.ts
@@ -259,82 +259,92 @@ export class CommunityPackagesService {
 			});
 		});
 
-		this.missingPackages = [];
+		// Initialize the missing packages list
+		this.missingPackages = [...missingPackages].map(
+			(p) => `${p.packageName}@${p.version}`,
+		);
 
 		if (missingPackages.size === 0) return;
 
 		const { reinstallMissing } = this.config;
 		if (reinstallMissing) {
-			this.logger.info('Attempting to reinstall missing packages', {
+			this.logger.info('Attempting to reinstall missing packages in the background', {
 				missingPackages: [...missingPackages],
 			});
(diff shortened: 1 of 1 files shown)


Joffcom:
@smeng9 if this change was accepted it could lead to workflows failing to activate on startup due to the nodes not existing or where workers are being used if the main instance has the node and a worker doesn't the main will register the workflow and if the worker has not installed it the workflow fail to run.

Personally I think the wait on startup is acceptable in most cases and if there is a slow connection on your server I would strongly recommend persisting the storage for nodes.

Out of interest what is the longest amount of time you have seen for the nodes to reinstall and what is the connection like for your server?

smeng9:
Hi @Joffcom

I run a small server on render.com free tier with 0.1 CPU and 512 MB of memory. I have 3-4 custom community nodes, including n8n-nodes-cloudflare that has a lot of sub-dependencies, and the longest time I have seen is 15 minutes. I don't run the service very often so it requires a cold start.

For your concern `workflows failing to activate on startup due to the nodes not existing`, if we don't turn on the switch N8N_REINSTALL_MISSING_PACKAGES they will fail anyway. Users use this flag already knows there are miss packages on code start

Joffcom:
@smeng9 I think you helped to reiterate my point, this setting is enabled if you know packages might be missing and you don't want to manually install them and activate the workflows.

If we change this it will be a pretty big breaking change as workflows will need to be manually enabled.

I think in your specific case maybe using the API to install community nodes (available from 2.16.0) once the instance has started would solve the issue you are seeing, we can then update the docs to mention that this setting can cause the startup to be delayed.

smeng9:
Thanks! Using API to install community nodes once the instance has started would be a good alternative solution

PR #28470, opened 2026-04-14
fix(Gmail Trigger Node): Add maxResults limit to prevent OOM on large inboxes

BerniWittmann:
## Summary

The Gmail Trigger causes OOM on Cloud when polling many or large emails. Root causes:
1. Gmail list API can return up to 100 message IDs, then each is fetched as a full message (`format=raw` can be 10MB+ per email)
2. Full `Message` objects were kept in memory just for timestamp/duplicate detection
3. No user control over how many emails to fetch per poll

This PR adds **node version 1.4** with a configurable "Max Emails per Poll" field (default 10, max 50). The approach: list all IDs (cheap), limit full fetches (expensive), store unfetched IDs as `pendingMessageIds` for the next poll cycle.

### Key changes:
- **`maxResults` property** gated behind `@version >= 1.4`, disabled in manual mode
- **`MessageBookkeeping` type** — lightweight `{id, date}` for timestamp/dedup computation without retaining full message bodies
- **`pendingMessageIds` state** — unfetched IDs stored for next poll (when # of emails > maxResults), drained before listing new messages
- **Duplicate fix** — `possibleDuplicates` are now merged (not replaced) when `lastTimeChecked` doesn't advance during pending drain, preventing boundary duplicates from reappearing
- **Timestamp flooring** — `Math.floor()` on `lastTimeChecked` to prevent float precision issues in equality comparisons
- **24 tests** (12 existing pinned to 1.3, 12 new for 1.4) covering: budget limiting, pending drain, multi-poll cycles, same-timestamp edge cases, draft filtering, early-return paths

### How to test:
1. Create a Gmail Trigger v1.4 workflow with `maxResults=2`
2. Send 3+ emails to the monitored account
3. Verify: Poll 1 returns 2 emails, Poll 2 returns remaining, Poll 3 returns nothing (no duplicates)

https://linear.app/n8n/issue/NODE-4595

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
+++ b/packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
@@ -23,7 +23,9 @@ import type {
 	GmailWorkflowStaticData,
 	GmailWorkflowStaticDataDictionary,
 	Label,
+	ListMessage,
 	Message,
+	MessageBookkeeping,
 	MessageListResponse,
 } from './types';
 
@@ -33,7 +35,7 @@ export class GmailTrigger implements INodeType {
 		name: 'gmailTrigger',
 		icon: 'file:gmail.svg',
 		group: ['trigger'],
-		version: [1, 1.1, 1.2, 1.3],
+		version: [1, 1.1, 1.2, 1.3, 1.4],
 		description:
 			'Fetches emails from Gmail and starts the workflow on specified polling intervals.',
 		subtitle: '={{"Gmail Trigger"}}',
@@ -112,6 +114,23 @@ export class GmailTrigger implements INodeType {
 				builderHint: {
 					message:
 						'Set to false when the email body is needed for AI analysis, summarization, or content processing. When true, only returns snippet (preview text). When false, returns full email with {id, threadId, labelIds, headers, html, text, textAsHtml, subject, date, to, from, messageId, replyTo}.',
+				},
+			},
+			{
+				displayName: 'Max Emails per Poll',
+				name: 'maxResults',
+				type: 'number',
+				default: 10,
+				typeOptions: {
+					minValue: 1,
+					maxValue: 50,
+				},
+				description:
+					'Maximum number of emails to fetch each time the node polls for new messages. If more emails arrive between polls, the remaining ones will be picked up in subsequent polls.',
+				displayOptions: {
+					show: {
+						'@version': [{ _cnd: { gte: 1.4 } }],
+					},
 				},
 			},
 			{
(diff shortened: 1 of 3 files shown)


packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -33,7 +35,7 @@ export class GmailTrigger implements INodeType {
 		name: 'gmailTrigger',
 		icon: 'file:gmail.svg',
 		group: ['trigger'],
-		version: [1, 1.1, 1.2, 1.3],
+		version: [1, 1.1, 1.2, 1.3, 2],
```
RomanDavydchuk: For node versioning, if we introduce a "major" version, like `2`, it usually means the node is rewritten from scratch and actually lives in a separate file. See an example of such node [here](https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/nodes-langchain/nodes/vendors/OpenAi/OpenAi.node.ts). For smaller changes, where a `this.getNode().typeVersion` check is enough, we usually add a new light version, so this would be better as a `1.4` version, in my opinion

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -285,11 +304,103 @@ export class GmailTrigger implements INodeType {
 
 		const options = this.getNodeParameter('options', {}) as GmailTriggerOptions;
 		const filters = this.getNodeParameter('filters', {}) as GmailTriggerFilters;
+		const simple = this.getNodeParameter('simple') as boolean;
+
+		const isV2 = node.typeVersion >= 2 && this.getMode() !== 'manual';
```
RomanDavydchuk: nit: This var is not telling us whether this node is v2, but rather tells us if we should limit the number of messages fetched, so the name is a bit misleading

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -425,16 +514,29 @@ export class GmailTrigger implements INodeType {
...
+		// second boundary, so messages at the lastTimeChecked timestamp can reappear.
+		{
+			const prevDuplicates = new Set(nodeStaticData.possibleDuplicates ?? []);
+			if (prevDuplicates.size > 0) {
+				responseData = responseData.filter(({ json }) => {
+					if (!json || typeof json.id !== 'string') return false;
+					return !prevDuplicates.has(json.id);
+				});
+			}
 		}
```
RomanDavydchuk: Not sure why we need to have this code in a block

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -425,16 +514,29 @@ export class GmailTrigger implements INodeType {
 			return emailDate <= lastEmailDate ? duplicates.concat(message.id) : duplicates;
 		}, Array.from(emailsWithInvalidDate));
```
RomanDavydchuk: I know the PR does not touch this code, but I got a bit confused by it when reviewing, so maybe you can help me you. If I understand correctly, `nextPollPossibleDuplicates` will always be essentially equivalent to `allFetchedMessages.map(m => m.id)` but with extra steps because:
1. `lastEmailDate` is the latest valid timestamp out of all the messages
2. `nextPollPossibleDuplicates` is: all emails that have a timestamp before the `lastEmailDate` + emails with invalid timestamp
3. Since `lastEmailDate` is by definition the latest timestamp, all valid timestamp are before it. In other words `nextPollPossibleDuplicates` has all emails with a valid timestamp + all emails with an invalid timestamp, i.e. `allFetchedMessages`

Please correct me if I'm wrong, but if you agree - maybe we can simplify it and make it less confusing?

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -285,11 +304,103 @@ export class GmailTrigger implements INodeType {
...
+				`/gmail/v1/users/me/messages/${messageId}`,
+				{},
+				fetchQs,
+			)) as Message;
+
+			allFetchedMessages.push({
+				id: fullMessage.id,
+				internalDate: fullMessage.internalDate,
+				date: fullMessage.date,
+				headers: fullMessage.headers?.date ? { date: fullMessage.headers.date } : undefined,
```
RomanDavydchuk: These three fields are only needed to call `getEmailDateAsSeconds` which expects them, so maybe let's calculate it ahead of time and store a single date as seconds? Would reduce the size of stored data even more

RomanDavydchuk: On a second thought, I'm not even sure if it's a good idea to create a new "slim" object with 2 or 4 fields. In the previous implementation we had `allFetchedMessages.push(fullMessage)`, so it was reusing the `fullMessage` object and not creating a new one. Let me know what you think

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
RomanDavydchuk: Moreover, `duplicates.concat` creates a new array for each element in `allFetchedMessages`... Using RAM even more with no reason for it

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
RomanDavydchuk: Actually, I don't think it's necessary to have this at all, we already do duplicate filtering at line `432` before getting full messages

RomanDavydchuk (changes requested):
I think that this is a good approach to fix this issue, but I did leave some comments on what can be improved and a few questions I had during the review. This node's code is quite complicated, so my review wasn't "linear", let me know if any of my comments are unclear. If we can simplify the code as well that would be even better 🙌 
P.S. haven't looked at the tests yet

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
BerniWittmann: thanks for the hint :)

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
BerniWittmann: good point, removed/simplified this

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
BerniWittmann: but as we iterate over the list of emails, we keep every messages raw body. 
With this "slimmer" object, the allFetchedMessages array only includes the slim object and at max full message raw body, as the garbage collection of the full message can kick in then and improves the memory, aka fixes the initial memory leak.

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
BerniWittmann: You're correct, this has been simplified, thanks for the great review comments 🙌

packages/nodes-base/nodes/Google/Gmail/GmailTrigger.node.ts
```diff
@@ -394,47 +501,36 @@ export class GmailTrigger implements INodeType {
...
+		const lastEmailDate = allFetchedMessages.reduce(
+			(lastDate, message) => (message.date > lastDate ? message.date : lastDate),
+			0,
+		);
+
+		const nextPollPossibleDuplicates = allFetchedMessages.map((m) => m.id);
+
+		// For older versions, filter at the response level since the pre-fetch filter
+		// above is gated to v1.4+. v1.4+ already skipped these before fetching.
+		if (!shouldLimitMessages) {
```
RomanDavydchuk, line 525: Is it necessary to have separate filtering logic for `< 1.4` and `>= 1.4`? I think we get always filter out duplicated before fetching the full message, regardless of the version (this is an implementation detail, shouldn't affect the end result). It would make the logic simpler and even provide some small memory wins for older versions (as the duplicates are not fetched)

As always, let me know if I'm missing something

RomanDavydchuk (changes requested):
Thanks for incorporating my comments, also looked at the tests and they are very nice and thorough 🙌. I did leave one comment in the code regarding dedupe logic

Also tested manually this time, here's what I did and what I observed. My flow:
1. Configure trigger to listen with max emails per poll = 2
2. Send emails `1`, `2`, `3` to the email address that's being listened, in this order, a couple of seconds apart
3. Wait for the trigger to do the polling request
4. Trigger does the polling request, returns emails `3` and `2`
5. After that send email `4` (before the trigger makes the second poll)
6. Wait for the trigger to poll for the second time
7. Trigger polls for the second time, returns emails `1` and `4`
8. Wait for 1 more minute (the polling interval)
9. Observe that the trigger runs again and returns **email `1`**

So two things to note here:
1. I would expect the trigger to return emails `1` and `2` first, then `3` and `4`. Currently it first returns `3` and `2`, then `1` and `4`
2. Obviously it shouldn't trigger for the third time with email `1` again. I suspect there is some issue with the logic for the case when both pending and new messages are returned within the same execution? Or perhaps the issue in dedupe logic?

BerniWittmann:
@RomanDavydchuk Could you help me reproduce your issue: I tried the following workflow, but it returned all 4 emails correctly, without an additional poll. Maybe oyu could provide your workflow/reproduction steps? 
[Gmail trigger.json](https://github.com/user-attachments/files/26921904/Gmail.trigger.json)


Regarding order of emails, i understand that having a FIFO ordering makes some kind of sense/feels more natural, but existing behavior of the node is also oldest first, so I would prefer to keep ordering as is for this PR to not have feature creep or changing behavior as a side effect of this fix.

BerniWittmann:
@RomanDavydchuk I was able to produce a test that reproduced the issue you had: https://github.com/n8n-io/n8n/pull/28470/changes#diff-2e39d2f1162f5e83a5182fef21c6ede331cf21ab1c8210adbc7aefabe00084e6R1359

Maybe you could retry with the latest changes from this branch, and hopefully that fixes your issue?

RomanDavydchuk (approved):
Tested manually, looks like the issue is fixed 🎉 Also tried a couple of other scenarios and tested the old version (1.3) for regressions, all looks good 🙌

PR #28472, opened 2026-04-14
feat(core): Scale expression isolate pool to 0 after inactivity

ivov:
https://linear.app/n8n/issue/CAT-2781/


despairblue (approved):
added some extra tests and minor non-functional changes

PR #28473, opened 2026-04-14
fix(editor): Prevent setup wizard disappearing on requestId-driven remount

aalises:
## Summary

- Add unmount guard to `InstanceAiWorkflowSetup` so async mount work (`fetchWorkflowApi`, credential preloading) bails out when the component has already been destroyed by a `requestId` change — prevents the dead instance from overwriting the workflow store
- Show a loading spinner while async setup is in flight, preventing a blank gap during the unmount→remount transition

## How this fixes the disappearing setup wizard

The setup wizard component is keyed by `requestId` in the parent panel. When the backend re-suspends with a new `requestId` (e.g. after re-analysis), Vue destroys the old component instance and creates a new one.

The problem is that `onMounted` performs async work — it fetches credentials and the workflow via API — but never checks whether the component was unmounted while those fetches were in flight. This creates a race condition:

1. **Component A** mounts, starts `fetchWorkflowApi` (async)
2. Backend sends a new `requestId` → **Component A unmounts** (restores `previousWorkflow` to the store), **Component B mounts**
3. **Component A's fetch resolves** → calls `workflowsStore.setWorkflow()` with stale data, **clobbering Component B's state**
4. The wizard appears empty or disappears entirely

The fix adds an `isMounted` flag that is set to `false` in `onUnmounted`. After each `await` in `onMounted`, we check `if (!isMounted) return` to bail out early — the dead component instance never touches the workflow store.

Additionally, the template now shows a loading spinner (`"Loading setup…"`) while the async initialization is running, so the user sees feedback during the unmount→remount gap instead of a blank space.

## Related Linear ticket

https://linear.app/n8n/issue/AI-2364/fix-setup-wizard-component-disappears-due-to-requestid-driven-remount

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/InstanceAiWorkflowSetup.test.ts
+++ b/packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/InstanceAiWorkflowSetup.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { createTestingPinia } from '@pinia/testing';
 import { setActivePinia } from 'pinia';
 import userEvent from '@testing-library/user-event';
+import { waitFor } from '@testing-library/vue';
 import { createComponentRenderer } from '@/__tests__/render';
 import type { InstanceAiWorkflowSetupNode } from '@n8n/api-types';
 import InstanceAiWorkflowSetup from '../components/InstanceAiWorkflowSetup.vue';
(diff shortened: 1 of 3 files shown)


packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/InstanceAiWorkflowSetup.test.ts
```diff
@@ -75,6 +76,18 @@ vi.mock('@/features/workflows/canvas/experimental/composables/useExpressionResol
...
 const renderComponent = createComponentRenderer(InstanceAiWorkflowSetup);
 
+/** Render the component and wait for the async onMounted to complete (isStoreReady = true). */
+async function renderAndWait(
+	...args: Parameters<typeof renderComponent>
+): Promise<ReturnType<typeof renderComponent>> {
+	const result = renderComponent(...args);
+	// Wait for async onMounted (credential/workflow fetch) to complete so the wizard renders
+	await waitFor(() => {
+		expect(result.queryByText('instanceAi.workflowSetup.loading')).toBeNull();
```
scdekov, line 86: nit: use `data-test-id` for the selector

scdekov (approved):

PR #28474, opened 2026-04-14
fix(editor): Improve setup wizard placeholder detection and card completion scoping

aalises:
## Summary

- Scope `isCardComplete()` placeholder check to only `trackedParamNames` instead of the entire node parameters, so hidden placeholders in un-rendered params don't block card completion
- Improve backend `buildSetupRequests()` placeholder detection to always surface placeholders (even when param already has validation issues) and include the hint label in error messages
- Add `extractPlaceholderLabel()` utility to `@n8n/utils` for extracting human-readable hints from placeholder sentinels

## Related Linear tickets

- https://linear.app/n8n/issue/AI-2344/setup-workflow-no-placeholder-detection-for-ai-inserted-placeholder

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
@@ -5,7 +5,7 @@
  * Separated from the tool definition so the tool stays a thin suspend/resume
  * state machine, and this logic is testable independently.
  */
-import { hasPlaceholderDeep, extractPlaceholderLabel } from '@n8n/utils';
+import { findPlaceholderDetails } from '@n8n/utils';
 import type { IDataObject, NodeJSON, DisplayOptions } from '@n8n/workflow-sdk';
 import { matchesDisplayOptions } from '@n8n/workflow-sdk';
 import { nanoid } from 'nanoid';
@@ -70,11 +70,9 @@ export async function buildSetupRequests(
 
 	// Also treat placeholder values as parameter issues so the setup wizard surfaces them
 	for (const [paramName, paramValue] of Object.entries(parameters)) {
-		if (hasPlaceholderDeep(paramValue)) {
-			const label = extractPlaceholderLabel(paramValue);
-			const message = label
-				? `Placeholder "${label}" — please provide the real value`
-				: 'Contains a placeholder value — please provide the real value';
+		const details = findPlaceholderDetails(paramValue);
+		if (details.length > 0) {
+			const message = `Placeholder "${details[0].label}" — please provide the real value`;
 			if (parameterIssues[paramName]) {
 				parameterIssues[paramName].push(message);
 			} else {
--- a/packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
@@ -5,7 +5,7 @@
  * Separated from the tool definition so the tool stays a thin suspend/resume
  * state machine, and this logic is testable independently.
  */
-import { hasPlaceholderDeep } from '@n8n/utils';
+import { hasPlaceholderDeep, extractPlaceholderLabel } from '@n8n/utils';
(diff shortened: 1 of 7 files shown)


scdekov:
How does this fix https://linear.app/n8n/issue/AI-2368/fix-setup-wizard-error-indicator-persists-after-entering-valid? From what I see the reported issue is not related to the placeholder params but an empty required ones?

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -70,8 +70,16 @@ export async function buildSetupRequests(
 
 	// Also treat placeholder values as parameter issues so the setup wizard surfaces them
 	for (const [paramName, paramValue] of Object.entries(parameters)) {
-		if (!parameterIssues[paramName] && hasPlaceholderDeep(paramValue)) {
-			parameterIssues[paramName] = ['Contains a placeholder value - please provide the real value'];
+		if (hasPlaceholderDeep(paramValue)) {
+			const label = extractPlaceholderLabel(paramValue);
+			const message = label
+				? `Placeholder "${label}" — please provide the real value`
+				: 'Contains a placeholder value — please provide the real value';
```
scdekov: Should we simplify this to extract the placeholder before the if and then check if there's a value or not in the if condition? Rn this seems like a dead code ie it shouldn't be possible to not have label if this line is reached.

packages/@n8n/utils/src/placeholder.ts
```diff
@@ -19,3 +22,46 @@ export function hasPlaceholderDeep(value: unknown): boolean {
 	}
 	return false;
 }
+
+/**
+ * Extract the human-readable hint label from the first placeholder sentinel found in a value.
+ * Handles both direct string placeholders and nested objects/arrays (recurses to find the first match).
+ * Returns `undefined` if no placeholder is found or the label is empty.
+ */
+export function extractPlaceholderLabel(value: unknown): string | undefined {
```
scdekov: Looks like we already have this logic in `packages/frontend/editor-ui/src/features/ai/assistant/composables/useBuilderTodos.ts` called `findPlaceholderDetails`. Maybe we should just move it here?

packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/InstanceAiWorkflowSetup.test.ts
```diff
@@ -75,6 +76,18 @@ vi.mock('@/features/workflows/canvas/experimental/composables/useExpressionResol
 
 const renderComponent = createComponentRenderer(InstanceAiWorkflowSetup);
 
+/** Render the component and wait for the async onMounted to complete (isStoreReady = true). */
```
scdekov: Doesn't seem to be related to the current changes.

scdekov (changes requested):

aalises:
> How does this fix https://linear.app/n8n/issue/AI-2368/fix-setup-wizard-error-indicator-persists-after-entering-valid? From what I see the reported issue is not related to the placeholder params but an empty required ones?

True! must've slipped in the stacking of PRs (like the unrelated changes), will remove the unrelated changes and the link to this issue

packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/InstanceAiWorkflowSetup.test.ts
aalises: stacked changes shenanigans

packages/@n8n/utils/src/placeholder.ts
aalises: done! consolidated in utils

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
aalises: yep! done and removed the dead code

scdekov (approved):

PR #28475, opened 2026-04-14
fix(core): Fix public API package update process

sandra0503:
- Updated the update method to include verification for community packages. Makes sure to retrieve the checksum for the respective version
- Improved error messages to clarify when a package is not verified.

Screenshot 2026-04-14 at 14 22 07

Screenshot 2026-04-14 at 14 25 27

Fixes [LIGO-447](https://linear.app/n8n/issue/LIGO-447/community-packages-install-endpoint-doesnt-work-on-cloud)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
+++ b/packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
@@ -107,6 +107,7 @@ export class CommunityPackagesLifecycleService {
 		let checksum: string | undefined;
 
 		if (verify) {
+			await this.communityNodeTypesService.getCommunityNodeTypes();
 			checksum = this.communityNodeTypesService.findVetted(name)?.checksum;
 			if (!checksum) {
 				throw new BadRequestError(`Package ${name} is not vetted for installation`);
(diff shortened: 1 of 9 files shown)


packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
```diff
@@ -99,13 +100,14 @@ export class CommunityPackagesLifecycleService {
...
 
-		if (version && !isValidVersionSpecifier(version)) {
+		if (version && !valid(version)) {
 			throw new BadRequestError(`Invalid version: ${version}`);
 		}
 
 		let checksum: string | undefined;
 
 		if (verify) {
+			await this.communityNodeTypesService.getCommunityNodeTypes();
```
konstantintieber: It looks like no one is using the resolved value of `getCommunityNodeTypes`. Why is this call here?

sandra0503: findVetted (line 111) relies on `this.communityNodeTypes.values()`. Before it worked if these values were in memory, but it would fail if it didn't get populated

konstantintieber: What part of the `getCommunityNodeTypes` does `findVetted` rely on?
Right now, this `getCommunityNodeTypes` call contains a sideeffect.
It'd be nice if `findVetted` would work as expected independently if someone called `getCommunityNodeTypes` right before it.

konstantintieber: Otherwise, it may be worth ensuring that `getCommunityNodes` is called before `findVetted`,  by requiring a value like the resolved value of `getCommunityNodeTypes` as parameter to `findVetted`.

packages/cli/src/public-api/v1/handlers/community-packages/spec/paths/community-packages.packageName.yml
```diff
@@ -23,6 +23,11 @@ patch:
             version:
               type: string
               description: Specific semver version to update to
+            verify:
+              type: boolean
+              description: >
+                Whether to verify the package against the n8n-vetted package list.
+                Required when the instance has N8N_UNVERIFIED_PACKAGES_ENABLED=false.
```
geemanjs: A part of me feels verify should default to true. But I guess we can't do that without bumping the api version?

packages/cli/src/modules/community-packages/community-packages.lifecycle.service.ts
sandra0503: Sure, we can pull this fetch into `findVetted` so it will be an independen function

packages/cli/src/public-api/v1/handlers/community-packages/spec/paths/community-packages.packageName.yml
sandra0503: Yeah I was thinking about that too. Probably it should be true by default. We just introduced the flag with 2.17.0, so I think it's still safe to change

sandra0503: I set the default to true now

konstantintieber (approved):
LGTM

PR #28478, opened 2026-04-14
fix(core): Filter stale credentials from setup wizard requests

aalises:
## Summary

The setup wizard was asking users to configure credentials for nodes that don't need them (e.g., an HTTP Request node calling a public API with `authentication: 'none'`).

**Root cause:** Stale `node.credentials` entries, left over from earlier builder revisions, were being surfaced unconditionally in two places:
1. `getNodeCredentialTypes()` in the adapter had a "Step 4" that added all existing credentials without checking if they're still needed
2. The fallback in `buildSetupRequests()` checked `node.credentials` first, bypassing the displayOptions-filtered path

**Fix:**
- Removed the unconditional existing-credentials step from the adapter: Steps 1-3 (displayOptions filtering, credential issues, dynamic resolution) already cover all legitimate detection paths
- Removed the unfiltered `node.credentials` fallback in `buildSetupRequests()`: now only the displayOptions-filtered path from the node description is used

## Related Linear ticket

https://linear.app/n8n/issue/AI-2366


packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
```diff
@@ -1689,7 +1689,7 @@ export class InstanceAiAdapterService {
 				return filteredIssues;
 			},
 
-			getNodeCredentialTypes: async (nodeType, typeVersion, parameters, existingCredentials) => {
+			getNodeCredentialTypes: async (nodeType, typeVersion, parameters, _existingCredentials) => {
```
scdekov, line 1693: Why do we keep this param?

scdekov:
How did you test this? I'm still seeing the same issue linked notion doc


https://github.com/user-attachments/assets/7c38f70a-73b3-4068-ae48-2d4a3c266de8

aalises:
Thanks for flagging @scdekov,  you were right, the first pass only addressed the display side 🙈 fixed it and added some more fixes

If you re-run your repro against the Berlin-weather flow from the Notion doc, the HTTP Request node should no longer carry httpSslAuth after the first setup/apply pass.

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -31,6 +31,76 @@ export function createCredentialCache(): CredentialCache {
...
+		try {
+			const dynamic = await context.nodeService.getNodeCredentialTypes(
+				node.type,
+				typeVersion,
+				parameters,
+				node.credentials as Record<string, unknown> | undefined,
+			);
+			for (const t of dynamic) types.add(t);
+		} catch {
+			// Ignore — fall through to description-based detection.
```
Cadiac: could we use the logger from our context here and log a warning here or something - at least for now to understand when these happen?

Cadiac: even on debug loglevel if we expect this to be too spammy for `warn`, but doesn't seem like this should happen?

packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
Cadiac, line 1693: Unused?

Cadiac (commented):
I couldn't figure out a good way to reproduce this locally and validate the fix, but I _think_ it makes sense as the fix, but how confident are you that this fix of filtering the credentials like this is necessary?

I think we're anyways better having this on monday, so happy to give this an approval but maybe consider adding the logger on the catch block.

scdekov:
@aalises Are we missing the filtering of the 'none' auth credential on wf update? I did't get a setup request for it on the first message but after I asked to add one more node I got cards for both the new card which was correct but also for the first request which did not require credentials.

https://github.com/user-attachments/assets/532e441e-24de-4f30-81ca-ca833199a013

Should be reproducible by first asking for `generate workflow that fetches a random joke from https://icanhazdadjoke.com/` and then follow up with `add another request to openrouter to rate how funny the joke is`

Not related to this change but 2 more issues I see:
* The builder picks openai credential for openrouter http request which is not correct. I can look into this as part of https://linear.app/n8n/issue/AI-2363/fix-setup-wizard-doesnt-handle-header-auth-well-asks-for-name-and
* Some request setup cards get 2 checkmarks when completed 
image

Cadiac:
for what its worth I just tested on the original prompt reported on notion, and that case now worked on 5df2eec5643b729cfb620f31b3831e872185f905

aalises:
> @aalises Are we missing the filtering of the 'none' auth credential on wf update? I did't get a setup request for it on the first message but after I asked to add one more node I got cards for both the new card which was correct but also for the first request which did not require credentials.
> 
>  Screen.Recording.2026-04-20.at.9.13.29.mov 
> Should be reproducible by first asking for `generate workflow that fetches a random joke from https://icanhazdadjoke.com/` and then follow up with `add another request to openrouter to rate how funny the joke is`
> 
> Not related to this change but 2 more issues I see:
> 
> * The builder picks openai credential for openrouter http request which is not correct. I can look into this as part of https://linear.app/n8n/issue/AI-2363/fix-setup-wizard-doesnt-handle-header-auth-well-asks-for-name-and
> * Some request setup cards get 2 checkmarks when completed
> 
> image

Does this happen also on master? would maybe need a follow-up in another PR

scdekov:
> > @aalises Are we missing the filtering of the 'none' auth credential on wf update? I did't get a setup request for it on the first message but after I asked to add one more node I got cards for both the new card which was correct but also for the first request which did not require credentials.
> > Screen.Recording.2026-04-20.at.9.13.29.mov
> > Should be reproducible by first asking for `generate workflow that fetches a random joke from https://icanhazdadjoke.com/` and then follow up with `add another request to openrouter to rate how funny the joke is`
> > Not related to this change but 2 more issues I see:
> > 
> > * The builder picks openai credential for openrouter http request which is not correct. I can look into this as part of https://linear.app/n8n/issue/AI-2363/fix-setup-wizard-doesnt-handle-header-auth-well-asks-for-name-and
> > * Some request setup cards get 2 checkmarks when completed
> > 
> > image
> 
> Does this happen also on master? would maybe need a follow-up in another PR

It does happen on master but the description of this PR says it is that issue that is being fixed. I guess it fixes half of it (on the first message it doesn't show the 'none' auth for the http request node) so up to you if we should create a new issue for the second part or fix it here

aalises:
> > > @aalises Are we missing the filtering of the 'none' auth credential on wf update? I did't get a setup request for it on the first message but after I asked to add one more node I got cards for both the new card which was correct but also for the first request which did not require credentials.
> > > Screen.Recording.2026-04-20.at.9.13.29.mov
> > > Should be reproducible by first asking for `generate workflow that fetches a random joke from https://icanhazdadjoke.com/` and then follow up with `add another request to openrouter to rate how funny the joke is`
> > > Not related to this change but 2 more issues I see:
> > > 
> > > * The builder picks openai credential for openrouter http request which is not correct. I can look into this as part of https://linear.app/n8n/issue/AI-2363/fix-setup-wizard-doesnt-handle-header-auth-well-asks-for-name-and
> > > * Some request setup cards get 2 checkmarks when completed
> > > 
> > > image
> > 
> > 
> > Does this happen also on master? would maybe need a follow-up in another PR
> 
> It does happen on master but the description of this PR says it is that issue that is being fixed. I guess it fixes half of it (on the first message it doesn't show the 'none' auth for the http request node) so up to you if we should create a new issue for the second part or fix it here

Can you test again? pushed a fix where the builder save paths now strip credential entries that don't match a node's current parameters before saving, so a stale credential reference the LLM re-emits on a follow-up turn can't survive long enough to make a setup card reappear on a node that no longer needs one.

Cadiac (approved):

PR #21952, opened 2025-11-17
fix(core): Use execFile instead of exec to prevent command injection

elsmr:
## Summary

Replaces string-based `exec()` calls with arg array-based `execFile()` calls throughout the community packages module to eliminate potential shell injection vulnerabilities.

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/NODE-3890

## Review / Merge checklist

- [ ] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `release/backport` (if the PR is an urgent fix that needs to be backported)

---

> [!NOTE]
> Switch npm command execution to execFile with argument arrays, updating service API, controller/CLI calls, and tests accordingly.
> 
> - **Core/Service (`community-packages.service.ts`)**:
>   - Replace `exec` with `execFile` (promisified) for all npm/tar invocations.
>   - Change `executeNpmCommand` signature from string to `string[]` args and call `execFile('npm', args, ...)`.
>   - Update `downloadPackage()` to use `execFile` for `npm pack`, `tar`, and `npm install` with array args.
>   - Return `getNpmInstallArgs()` as an array instead of a joined string.
> - **Controller/CLI**:
>   - Controller: call `executeNpmCommand(['outdated', '--json'], { doNotHandleError: true })`.
>   - CLI command: prune via `executeNpmCommand(['prune'])`.
> - **Tests**:
>   - Update mocks and expectations to use `execFile` and array-style args across integration/unit tests, including npm pack/tar/install sequences.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d56403b5d19b6b6ea6b700a707c0f5e8b96f77ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #28481, opened 2026-04-14
fix(core): Can't install community nodes on Windows

### LINKED ISSUE #22706: installing community nodes get "spawn npm ENOENT" Issue in n8n on Windows
### Bug Description

When installing community nodes in n8n under the Windows environment, the "spawn npm ENOENT" error often occurs. The core reason is that n8n does not adapt to Windows script suffix resolution when calling npm.

### To Reproduce

Setting->Community nodes->Install

### Expected behavior

Update n8n version to  1.122.4, install community node get "spawn npm ENOENT". It worked fine in the previous version.

### Debug Info

This problem can be solved once and for all by modifying the n8n source code to add the `shell: true` parameter to the `execFile` call. Here are the detailed steps.

## Core Solution: Source Code Modification to Adapt to Windows npm Calls

n8n calls npm through `asyncExecFile` (a Promise wrapper for `execFile`). On Windows, the `npm.cmd` suffix needs to be parsed through the Shell. Adding the `shell: true` parameter to all npm calls enables compatibility.

### Step 1: Locate the Core Configuration File

The n8n community package management logic is located in the compiled JS file (no need to modify TS source code), with a fixed path:

```Plain Text
C:\Users\YourUsername\AppData\Roaming\npm\node_modules\n8n\dist\modules\community-packages\community-packages.service.js
```

Open this file with VS Code or Notepad++, and search for "asyncExecFile('npm'" to locate 3 npm call points.

### Step 2: Modify the Three execFile Calls

Add the `shell: true` option to each `asyncExecFile('npm'` call. The specific modifications are as follows:

#### 1. Modify the executeNpmCommand Method

Find the `execOptions` configuration block and add `shell: true`:

```Plain Text
// Original code
const execOptions = {
    cwd: this.downloadFolder,
    env: { NODE_PATH: process.env.NODE_PATH, PATH: process.env.PATH, APPDATA: process.env.APPDATA, NODE_ENV: 'production' },
};

// Modified code
const execOptions = {
    cwd: this.downloadFolder,
    env: { NODE_PATH: process.env.NODE_PATH, PATH: process.env.PATH, APPDATA: process.env.APPDATA, NODE_ENV: 'production' },
    shell: true, // Add this line
};
```

#### 2. Modify the npm pack Call in downloadPackage

Add `shell: true` to the configuration of the `npm pack` call:

```Plain Text
// Original code
const { stdout: tarOutput } = await asyncExecFile('npm', ['pack', `${packageName}@${packageVersion}`, `--registry=${registry}`, '--quiet'], { cwd: this.downloadFolder });

// Modified code
const { stdout: tarOutput } = await asyncExecFile('npm', ['pack', `${packageName}@${packageVersion}`, `--registry=${registry}`, '--quiet'], { 
    cwd: this.downloadFolder,
    shell: true // Add this line
});
```

#### 3. Modify the npm install Call in downloadPackage

Add `shell: true` to the configuration of the `npm install` call:

```Plain Text
// Original code
await asyncExecFile('npm', ['install', ...this.getNpmInstallArgs()], { cwd: packageDirectory });

// Modified code
await asyncExecFile('npm', ['install', ...this.getNpmInstallArgs()], { 
    cwd: packageDirectory,
    shell: true // Add this line
});
```

### Step 3: Restart n8n to Verify the Effect

1. Save the modified file;
2. End all `node.exe` processes in Task Manager (completely shut down n8n background processes);
3. Re-execute `n8n start` to launch the service;
4. Install a community node directly in n8n (e.g., @fictionking/n8n-nodes-stream-respond). If it can be installed normally, the modification is effective.

> (Note: Some parts of the document may be generated by AI)

### Operating System

Windows 11

### n8n Version

Version 1.122.4

### Node.js Version

v22.11.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

RomanDavydchuk:
## Summary

Community nodes installation is broken on Windows due to changes in this PR: #21952. `execFile` does not work on Windows without `{ shell: true }` because on Windows `npm.cmd` script is used to run npm and a shell is needed to run the `.cmd` script. But `{ shell: true }` introduces potential security issues, which is why it's not a good approach. This PR fixes npm commands on Windows by directly invoking `npm-cli.js` file via `node` executable. This allows us to still use `execFile` where args are provided as an array (which is more secure) without adding `{ shell: true }`

### The flow (analogous to what `npm.cmd` does, see [here](https://github.com/nodejs/node/blob/main/deps/npm/bin/npm.cmd)):
1. Determine the location of `node` executable using `process.execPath`
2. Try running `npm-prefix.js` which reads npm config and returns the prefix path of where `npm-cli.js` 
3. If successful, use that prefix to determine `npm-cli.js` location
4. Otherwise, try two potential "default" locations
5. If able to get `npm-cli.js` location - run it with the actual command
6. Otherwise - throw an error

### Notes
- Change only affects Windows, for Linux/Mac npm commands are executed in the same way as right now
- The "discovery" process runs only once per instance lifetime and the path gets cached afterwards

### Demo of this fix on Windows:
https://github.com/user-attachments/assets/215b642c-45a8-459f-adda-642afad48a3c

https://linear.app/n8n/issue/NODE-4291/community-nodes-cant-be-installed-on-windows
Closes #22706

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/community-packages/npm-utils.ts
```diff
@@ -100,8 +197,7 @@ export async function executeNpmCommand(
 	});
 
 	try {
-		const { stdout } = await asyncExecFile('npm', fullArgs, cwd ? { cwd } : undefined);
-		return typeof stdout === 'string' ? stdout : stdout.toString();
+		return await executeNpmCli(args, cwd);
```
ShireenMissi: Should this be ?
```suggestion
		 return await executeNpmCli(fullArgs, cwd);
```

ShireenMissi (changes requested):
Mostly looks good! The build is failing for me locally so I can't run the code locally
I also left a question

packages/cli/src/modules/community-packages/npm-utils.ts
RomanDavydchuk: Good catch, `fullArgs` was introduced in `master` after I submitted the PR, so I missed it when merging `master`

RomanDavydchuk:
@ShireenMissi the build was failing because I missed a `}` when resolving a merge conflict with `master` 😅 Should be all good now. Also the comment regarding `fullArgs` is valid - good catch

ShireenMissi (approved):
LGTM 🎉

PR #28486, opened 2026-04-14
fix(editor): Save button never shows saved state after saving credentials

elsmr:
## Summary

Restore the old behavior of the Save button:

When a credential was saved, show a "Saved" label instead of the Save button

image

https://linear.app/n8n/issue/NODE-4841/save-button-never-shows-saved-state-after-saving-credentials

## Review / Merge checklist

- [ ] Tests included. 


dlavrenuek (commented):
Code looks good but it looks like all of the e2e tests that touch credentials fail now because there is no button element any more

dlavrenuek (approved):
🎉

PR #28495, opened 2026-04-14
fix(core): Rework Instance ai settings (no-changelog)

Cadiac:
## Summary

- If proxy is enabled hide & prevent setting sandbox / model on settings
- Respect the local gateway settings correctly
  - Proper working global toggle for admins to disable local gateway for everyone
  - User level toggle for enabling / disabling local gateway personally for everyone
  - Enforce these settings on the backend
  - Don't request the gw's /events endpoint if local gateway is disabled 
- Make settings work for members (only allowing the personal toggle & model picker on self hosted)
- Style fixes at the UI
- Validate request bodies on local gw endpoints using our preferred `@Body` decorators instead of doing it manually

## Review / Merge checklist

- [ ] Tests included. 


dlavrenuek:
So we don't forget it: when Instance AI or Computer Use are disabled globally (admin changes the setting from enabled to disabled) - all clients should be disconnected immediately. An example of [the existing call is here](https://github.com/n8n-io/n8n/blob/883d747337ca827476760b8ae559f5b6891efb84/packages/cli/src/modules/instance-ai/instance-ai.service.ts#L1062)

packages/cli/src/modules/instance-ai/instance-ai-settings.service.ts
```diff
@@ -367,9 +391,10 @@ export class InstanceAiSettingsService {
 	}
 
 	/** Whether the local gateway is disabled for a given user (admin override OR user preference). */
-	isLocalGatewayDisabledForUser(userId: string): boolean {
+	async isLocalGatewayDisabledForUser(userId: string): Promise<boolean> {
+		if (!this.enabled) return true;
 		if (this.config.localGatewayDisabled) return true;
-		const prefs = this.userPreferences.get(userId);
+		const prefs = await this.loadUserPreferences(userId);
```
Cadiac, line 397: Since the setting is for disabling the local gw before when this wasn't async and actually checking db it would return undefined -> "not disabled" on first reads after instance startup

Cadiac:
> So we don't forget it: when Instance AI or Computer Use are disabled globally (admin changes the setting from enabled to disabled) - all clients should be disconnected immediately

Addressed in 9b97410435dae28d795d527d0a63fb90d72d6be7

packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -625,24 +646,21 @@ export class InstanceAiController {
 	}
 
 	@Post('/gateway/init', { skipAuth: true })
-	gatewayInit(req: Request) {
+	async gatewayInit(req: Request, _res: Response, @Body payload: InstanceAiGatewayCapabilitiesDto) {
```
dlavrenuek, line 659: @Cadiac this does not seem to work properly

> @Body payload: InstanceAiGatewayCapabilitiesDto

payload is null when I run the code

Cadiac: Found the issue - same one as I've done a few times in the past, but looks like claude hasn't learned it yet - we can't export the DTO's as types. Should be fixed in f30b65b56724207b3357582314721d000489c608

aalises:
there's some schema/DTO duplication in `instance-ai.schema.ts`

`InstanceAiGatewayCapabilitiesDto` and `InstanceAiFilesystemResponseDto` seem copied duplicates of `instanceAiGatewayCapabilitiesSchema` / `instanceAiFilesystemResponseSchema` (same field list, separate declarations).

I would derive one from the other (e.g. have the DTO class wrap the existing schema), or delete the old Zod schemas if they are no longer used... or they might drift?

Cadiac:
> there's some schema/DTO duplication in `instance-ai.schema.ts`
> 
> `InstanceAiGatewayCapabilitiesDto` and `InstanceAiFilesystemResponseDto` seem copied duplicates of `instanceAiGatewayCapabilitiesSchema` / `instanceAiFilesystemResponseSchema` (same field list, separate declarations).
> 
> I would derive one from the other (e.g. have the DTO class wrap the existing schema), or delete the old Zod schemas if they are no longer used... or they might drift?

Good catch, definitely just claude being silly here and this isn't how we've been doing these schemas elsewhere. Addressed in c8dff7c633

aalises (approved):

PR #28510, opened 2026-04-15
fix(MCP Server Trigger Node): Propagate tool errors with isError flag in MCP responses

schrothbn:
## Summary
  - When a tool execution fails (e.g. expired OAuth2 token returning HTTP 401), the MCP Server Trigger was returning the error as a
  normal successful text response without isError: true, making it impossible for MCP clients to distinguish tool errors from
  successful results
  - Add MessageFormatter.isErrorResult() to detect error results from both queue mode (error objects from worker: { error: {
  message, name } }) and direct mode (error strings like "NodeApiError: ..." or "There was an error: ...")
  - Pass the isError flag through formatToolResult at all three call sites in McpServer.ts so the MCP protocol response correctly
  signals errors to clients

https://linear.app/n8n/issue/AI-2278

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/mcp/McpTrigger/protocol/MessageFormatter.ts
+++ b/packages/@n8n/nodes-langchain/nodes/mcp/McpTrigger/protocol/MessageFormatter.ts
@@ -1,25 +1,48 @@
 import type { McpToolResult } from './types';
 
 export class MessageFormatter {
-	static formatToolResult(result: unknown): McpToolResult {
+	static formatToolResult(result: unknown, isError = false): McpToolResult {
+		let content: McpToolResult['content'];
+
 		if (typeof result === 'object' && result !== null) {
-			return { content: [{ type: 'text', text: JSON.stringify(result) }] };
+			content = [{ type: 'text', text: JSON.stringify(result) }];
+		} else if (typeof result === 'string') {
+			content = [{ type: 'text', text: result }];
+		} else if (result === null || result === undefined) {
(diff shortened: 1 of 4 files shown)


packages/@n8n/nodes-langchain/nodes/mcp/McpTrigger/protocol/MessageFormatter.ts
```diff
@@ -1,25 +1,48 @@
...
+	 * Detect whether a tool result represents an error.
+	 *
+	 * In direct mode, N8nTool catches errors and returns `e.toString()` producing
+	 * strings like `"NodeApiError: Bad request"`. ToolHttpRequest catches HTTP errors
+	 * and returns `"There was an error: \"...\"" or `"HTTP 401 There was an error: \"...\""`.
+	 *
+	 * In queue mode, the job processor wraps errors as `{ error: { message, name } }`.
+	 */
+	static isErrorResult(result: unknown): boolean {
+		if (typeof result === 'object' && result !== null && 'error' in result) {
```
aalises, line 39: Narrow this check to the worker error envelope, e.g. require `error` to be a non-null object with a string `message`. Valid tool results can contain an `error` field and should not be marked `isError`.
image

aalises (dismissed):
I'd tighten error detection so successful tool outputs with an `error` field are not marked as failed MCP calls :'(

aalises (approved):

PR #28512, opened 2026-04-15
fix(Freshservice Node): guard against missing error description (#18296)

aayushbaluni:
## Summary

Fixes #18296.

**Root cause:** `freshserviceApiRequest` accesses `error.error.description` without checking if `error.error` exists, causing `TypeError: Cannot read properties of undefined` when the API returns an error without the expected nested structure.

**Fix:** Added optional chaining (`?.`) to safely access nested error properties.

## Changes

- `packages/nodes-base/nodes/Freshservice/GenericFunctions.ts`: Added null-safe access to error response properties

## Testing

- Verified fix prevents TypeError when API returns error without description field
- Existing error handling for validation failures continues to work


Joffcom:
Hi @aayushbaluni — thanks for the contribution! Two things before we can route this for review:

1. **Add a regression test.** Per our [contributing guide](../blob/master/CONTRIBUTING.md), bug fixes need a test that locks in the fix. A unit test in `packages/nodes-base/nodes/Freshservice/test/` that calls `freshserviceApiRequest` with an error response missing the `description` field and asserts it no longer throws would be ideal.
2. **Update the PR description to match the template.** Your `## Changes` / `## Testing` sections have good content but please rename to the template's `## Summary` and `## Related Linear tickets, Github issues, and Community forum posts` (your Fixes #18296 belongs under the latter) plus add the `## Review / Merge checklist` from `.github/pull_request_template.md`.

Thanks!

PR #28518, opened 2026-04-15
feat(core): Persist deployment_key entries for stability across restarts and key rotation

BGZStephen:
## Summary

This PR delivers two Linear tickets (IAM-485 and IAM-486) as a single unit because the IAM-486 work is directly layered on top of IAM-485.

### Problem

n8n derives four signing secrets directly from the encryption key at every boot:

| Secret | Used for | Derivation |
|--------|----------|------------|
| `instanceId` | Telemetry, licence | `SHA256(encryptionKey.slice(mid))` |
| `hmacSignatureSecret` | Webhook resumption URL signing | `SHA256("hmac-signature:" + encryptionKey)` |
| `jwtSecret` | User session JWT signing | `SHA256("n8n:" + encryptionKey)` |
| `signingSecret` | Binary-data presigned URL signing | `SHA256("url-signing:" + encryptionKey)` |

Rotating the encryption key immediately and silently invalidates all active user sessions, in-flight webhook resumption URLs, shared binary-data links, and instance identity used for telemetry/licensing.

### Solution

A **two-phase initialisation pattern** that decouples each secret from the encryption key after first boot:

1. **Constructor** — always derives the value from the encryption key (backward-compatible fallback; no change on upgrade).
2. **`initialize(repo)`** — called once after DB migrations complete. Precedence: `env var → active DB row → insertOrIgnore (persist) → race-condition fallback read`.

The derived value is persisted to the `deployment_key` table on first boot. Every subsequent boot reads from DB, so the secret is stable regardless of what happens to the encryption key.

### Concurrency safety (multi-main)

`DeploymentKeyRepository.insertOrIgnore()` issues `INSERT OR IGNORE RETURNING *`. On a unique-index conflict (two mains racing at startup), the insert is silently dropped and `null` is returned; the caller reads the winner's row. No exceptions, no locks.

---

### `packages/@n8n/db` (IAM-485)
- New entity `DeploymentKey` — `id`, `type`, `value`, `status`, `algorithm`, timestamps
- New repository `DeploymentKeyRepository` with `findActiveByType()` and `insertOrIgnore()`
- New migration `1777000000000-CreateDeploymentKeyTable` — table + unique partial indexes for `instance.id`, `signing.hmac`, `signing.jwt`, `signing.binary_data`

### `packages/core / InstanceSettings` (IAM-485 + IAM-486)
- `instanceId` persisted to `deployment_key` (type `instance.id`) via `initialize(repo)`
- `hmacSignatureSecret` made mutable; persisted via same `initialize(repo)` call (type `signing.hmac`)
- Duck-typed inline repo interface avoids circular package dependency (`@n8n/db` → `n8n-core` at runtime)

### `packages/core / BinaryDataConfig` (IAM-486)
- New `initialize(repo)` — env-var fast-path → DB read → insertOrIgnore → race fallback; type `signing.binary_data`

### `packages/cli / JwtService` (IAM-486)
- `jwtSecret` made `private`
- New `initialize(repo)` — same pattern; type `signing.jwt`; respects `N8N_USER_MANAGEMENT_JWT_SECRET`

### `packages/cli / start.ts` (IAM-486)
- Calls `initialize(DeploymentKeyRepository)` for `InstanceSettings`, `JwtService`, and `BinaryDataConfig` in sequence after DB migrations, before licence init

### Tests
- `instance-settings.test.ts` — 8 new tests (4× `instance.id`, 4× `signing.hmac`)
- `jwt.service.test.ts` — 4 new `initialize()` tests
- `binary-data.config.test.ts` — 4 new `initialize()` tests

Each set covers: env-var fast-path, existing DB row, persist-on-first-boot, concurrent-insert race condition.

---

## Upgrade path

No breaking changes. Derived values are identical to pre-upgrade; they are transparently persisted on first boot of the new version. Sessions, webhook URLs, and binary-data links continue to work without interruption.

## Environment variables

| Variable | Behaviour |
|----------|-----------|
| `N8N_ENCRYPTION_KEY` | Still the derivation source on first boot |
| `N8N_USER_MANAGEMENT_JWT_SECRET` | Pinned; DB write skipped entirely |
| `N8N_HMAC_SIGNATURE_SECRET` | Pinned; DB write skipped entirely |
| `N8N_BINARY_DATA_SIGNING_SECRET` | Pinned; DB write skipped entirely |
| `N8N_INSTANCE_ID` | Pinned; DB write skipped entirely |

## Related Linear tickets

https://linear.app/n8n/issue/IAM-485
https://linear.app/n8n/issue/IAM-486
https://linear.app/n8n/issue/IAM-487
https://linear.app/n8n/issue/IAM-488

## Review / Merge checklist

- [x] Tests included.
- [ ] PR labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` if urgent.

### DIFF
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -13,6 +13,7 @@ import { McpServer } from '@n8n/n8n-nodes-langchain/mcp/core';
 import glob from 'fast-glob';
 import { createReadStream, createWriteStream, existsSync } from 'fs';
 import { mkdir } from 'fs/promises';
+import { BinaryDataConfig } from 'n8n-core';
 import { jsonParse, sleep, type IWorkflowExecutionDataProcess } from 'n8n-workflow';
 import path from 'path';
 import replaceStream from 'replacestream';
@@ -32,6 +33,7 @@ import { Publisher } from '@/scaling/pubsub/publisher.service';
 import { PubSubRegistry } from '@/scaling/pubsub/pubsub.registry';
 import { Subscriber } from '@/scaling/pubsub/subscriber.service';
 import { Server } from '@/server';
+import { JwtService } from '@/services/jwt.service';
 import { OwnershipService } from '@/services/ownership.service';
 import { ExecutionsPruningService } from '@/services/pruning/executions-pruning.service';
 import { UrlService } from '@/services/url.service';
@@ -230,6 +232,8 @@ export class Start extends BaseCommand<z.infer<typeof flagsSchema>> {
 		}
 
 		await this.instanceSettings.initialize(Container.get(DeploymentKeyRepository));
+		await Container.get(JwtService).initialize(Container.get(DeploymentKeyRepository));
+		await Container.get(BinaryDataConfig).initialize(Container.get(DeploymentKeyRepository));
 
 		await this.initLicense();
 
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -1,7 +1,12 @@
 /* eslint-disable @typescript-eslint/no-unsafe-call */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 import { LICENSE_FEATURES } from '@n8n/constants';
-import { AuthRolesService, ExecutionRepository, SettingsRepository } from '@n8n/db';
+import {
+	AuthRolesService,
+	DeploymentKeyRepository,
+	ExecutionRepository,
+	SettingsRepository,
+} from '@n8n/db';
 import { Command } from '@n8n/decorators';
 import { Container } from '@n8n/di';
 import { McpServer } from '@n8n/n8n-nodes-langchain/mcp/core';
@@ -223,6 +228,8 @@ export class Start extends BaseCommand<z.infer<typeof flagsSchema>> {
 		} else {
 			await this.initOrchestration();
 		}
+
+		await this.instanceSettings.initialize(Container.get(DeploymentKeyRepository));
 
 		await this.initLicense();
 
--- a/packages/cli/src/services/jwt.service.ts
+++ b/packages/cli/src/services/jwt.service.ts
@@ -9,10 +9,7 @@ export class JwtService {
 	private jwtSecret: string = '';
 	private readonly jwtSecretFromEnv: boolean;
 
-	constructor(
-		{ encryptionKey }: InstanceSettings,
-		private readonly globalConfig: GlobalConfig,
-	) {
+	constructor({ encryptionKey }: InstanceSettings, globalConfig: GlobalConfig) {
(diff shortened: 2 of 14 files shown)


packages/core/src/instance-settings/instance-settings.ts
```diff
@@ -52,11 +52,12 @@ export class InstanceSettings {
...
 	/**
 	 * Fixed ID of this n8n instance, for telemetry.
-	 * Derived from encryption key. Do not confuse with `hostId`.
+	 * Derived from encryption key on first boot, then read from DB.
+	 * Do not confuse with `hostId`.
 	 *
 	 * @example '258fce876abf5ea60eb86a2e777e5e190ff8f3e36b5b37aafec6636c31d4d1f9'
 	 */
-	readonly instanceId: string;
+	instanceId: string;
```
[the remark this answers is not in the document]
BGZStephen: no, this is fine

packages/cli/src/services/jwt.service.ts
```diff
@@ -23,6 +23,38 @@ export class JwtService {
...
+	async initialize(repo: {
+		findActiveByType(type: string): Promise<{ value: string } | null>;
+		insertOrIgnore(entity: {
+			type: string;
+			value: string;
+			status: string;
+			algorithm: null;
+		}): Promise<void>;
+	}): Promise<void> {
+		if (process.env.N8N_USER_MANAGEMENT_JWT_SECRET) {
```
guillaumejacquart: Why not check for `globalConfig.userManagement.jwtSecret` instead ?

packages/cli/src/commands/start.ts
```diff
@@ -224,6 +231,10 @@ export class Start extends BaseCommand<z.infer<typeof flagsSchema>> {
 			await this.initOrchestration();
 		}
 
+		await this.instanceSettings.initialize(Container.get(DeploymentKeyRepository));
```
guillaumejacquart, line 234: I think so, but are we sure that we're not gonna need that on workers / webhook ?

packages/cli/src/services/jwt.service.ts
BGZStephen: mainly because `globalConfig` is provided by the constructor as opposed to as a dependency available on JWT Service, eventually I think we'll move away from constructor to just use this initialize

guillaumejacquart: Ah yes, I did not notice. Why isn't it a private readonly injected like we do elsewhere ? 
Anyway i think we could also use `Container.get(GlobalConfig).userManagement.jwtSecret` here no ? 
I'm just wary of making sure that config fetching is future proof (like what if in the future this config can be set from a file instead of a envvar)

BGZStephen: Yep, makes sense, I've reworked to increase consistency

guillaumejacquart (approved):

PR #28519, opened 2026-04-15
feat(API): Add missing credential endpoints (GET by ID and test)

alielkhateeb:
## Summary

💡 I recommend reviewing this PR one commit at a time, it's easier to follow:
- **Commit 1:** Adds `credential:read` to `API_KEY_RESOURCES` so API keys can be granted this scope
- **Commit 2:** Adds `GET /credentials/{id}` endpoint
- **Commit 3:** Adds `POST /credentials/{id}/test` endpoint

https://www.loom.com/share/5f31a1cd114a45d6911a5aa8f34c6297

Adds two missing Public API endpoints for credentials:

- **`GET /credentials/{id}`** — Fetch a single credential by ID. Previously you had to call `GET /credentials` (paginated list) and filter client-side.
- **`POST /credentials/{id}/test`** — Test that a stored credential is valid. Useful for CI/CD pipelines that need to verify credentials work before deploying workflows that depend on them.

Also adds `credential:read` to `API_KEY_RESOURCES` so API keys can be granted this scope.

**How to test:**

```bash
# Get a single credential
curl -s "http://localhost:5678/api/v1/credentials/<id>" \
  -H "X-N8N-API-KEY: <your-key>" | jq .

# Test a credential
curl -s -X POST "http://localhost:5678/api/v1/credentials/<id>/test" \
  -H "X-N8N-API-KEY: <your-key>" | jq .
# Returns: {"status":"OK","message":"Connection successful!"}
# Or:      {"status":"Error","message":"..."}
```

Both endpoints require the `credential:read` scope on the API key and respect project-level access control.

https://linear.app/n8n/issue/LIGO-390

## Review / Merge checklist

- [x] Tests included.

---
🤖 PR Summary generated by AI

### DIFF
--- a/packages/@n8n/permissions/src/constants.ee.ts
+++ b/packages/@n8n/permissions/src/constants.ee.ts
@@ -71,7 +71,7 @@ export const API_KEY_RESOURCES = {
 	project: ['create', 'update', 'delete', 'list'] as const,
 	user: ['read', 'list', 'create', 'changeRole', 'delete', 'enforceMfa'] as const,
 	execution: ['delete', 'read', 'retry', 'list', 'get', 'stop'] as const,
-	credential: ['create', 'update', 'move', 'delete', 'list'] as const,
+	credential: ['create', 'read', 'update', 'move', 'delete', 'list'] as const,
 	sourceControl: ['pull'] as const,
 	workflowTags: ['update', 'list'] as const,
 	executionTags: ['update', 'list'] as const,
--- a/packages/cli/src/credentials/credentials.service.ts
+++ b/packages/cli/src/credentials/credentials.service.ts
@@ -43,6 +43,7 @@ import {
 import { CredentialTypes } from '@/credential-types';
 import { createCredentialsFromCredentialsEntity, CredentialsHelper } from '@/credentials-helper';
 import { BadRequestError } from '@/errors/response-errors/bad-request.error';
+import { CredentialNotFoundError } from '@/errors/credential-not-found.error';
 import { ForbiddenError } from '@/errors/response-errors/forbidden.error';
 import { NotFoundError } from '@/errors/response-errors/not-found.error';
 import { ExternalHooks } from '@/external-hooks';
@@ -817,6 +818,84 @@ export class CredentialsService {
 
 	async test(userId: User['id'], credentials: ICredentialsDecrypted) {
 		return await this.credentialsTester.testCredentials(userId, credentials.type, credentials);
+	}
+
+	/**
+	 * Build credentials payload ready to pass to credential testing.
+	 *
+	 * - If `credentialsToTest` is not provided, uses stored decrypted credential data.
+	 * - If `credentialsToTest` is provided, normalizes it for testing:
+	 *   - fills payload data for sharees when needed
+	 *   - restores redacted values from stored decrypted data
+	 */
+	private async prepareCredentialsForTest({
+		storedCredential,
+		user,
+		credentialsToTest,
+	}: {
+		storedCredential: CredentialsEntity;
+		user?: User;
+		credentialsToTest?: ICredentialsDecrypted;
+	}): Promise<ICredentialsDecrypted> {
+		const decryptedData = this.decrypt(storedCredential, true);
+		const mergedCredentials: ICredentialsDecrypted = credentialsToTest
+			? deepCopy(credentialsToTest)
+			: {
+					id: storedCredential.id,
+					name: storedCredential.name,
+					type: storedCredential.type,
+					data: decryptedData,
+				};
+
+		if (user && credentialsToTest) {
+			await this.replaceCredentialContentsForSharee(
+				user,
+				storedCredential,
+				decryptedData,
+				mergedCredentials,
+			);
+
+			if (mergedCredentials.data) {
+				mergedCredentials.data = this.unredact(
+					mergedCredentials.data,
+					decryptedData,
+					this.getCredentialTypeProperties(storedCredential.type),
+				);
+			}
+		}
+
+		return mergedCredentials;
+	}
+
+	async testById(userId: User['id'], credentialId: string) {
+		const storedCredential = await this.credentialsFinderService.findCredentialById(credentialId);
+
+		if (!storedCredential) {
+			throw new CredentialNotFoundError(credentialId);
+		}
+
+		const credentials = await this.prepareCredentialsForTest({ storedCredential });
+		return await this.test(userId, credentials);
+	}
+
+	async testWithCredentials(user: User, credentials: ICredentialsDecrypted) {
+		const storedCredential = await this.credentialsFinderService.findCredentialForUser(
+			credentials.id,
+			user,
+			['credential:read'],
+		);
+
+		if (!storedCredential) {
+			throw new CredentialNotFoundError(credentials.id);
+		}
+
+		const mergedCredentials = await this.prepareCredentialsForTest({
+			storedCredential,
+			user,
+			credentialsToTest: credentials,
+		});
+
+		return await this.test(user.id, mergedCredentials);
 	}
 
 	// Take data and replace all sensitive values with a sentinel value.
--- a/packages/cli/src/credentials/credentials.service.ts
+++ b/packages/cli/src/credentials/credentials.service.ts
@@ -820,53 +820,6 @@ export class CredentialsService {
 		return await this.credentialsTester.testCredentials(userId, credentials.type, credentials);
 	}
 
-	/**
-	 * Build credentials payload ready to pass to credential testing.
-	 *
-	 * - If `credentialsToTest` is not provided, uses stored decrypted credential data.
-	 * - If `credentialsToTest` is provided, normalizes it for testing:
-	 *   - fills payload data for sharees when needed
-	 *   - restores redacted values from stored decrypted data
-	 */
-	private async prepareCredentialsForTest({
-		storedCredential,
-		user,
-		credentialsToTest,
-	}: {
-		storedCredential: CredentialsEntity;
-		user?: User;
-		credentialsToTest?: ICredentialsDecrypted;
-	}): Promise<ICredentialsDecrypted> {
-		const decryptedData = this.decrypt(storedCredential, true);
-		const mergedCredentials: ICredentialsDecrypted = credentialsToTest
-			? deepCopy(credentialsToTest)
-			: {
-					id: storedCredential.id,
-					name: storedCredential.name,
-					type: storedCredential.type,
-					data: decryptedData,
-				};
-
-		if (user && credentialsToTest) {
-			await this.replaceCredentialContentsForSharee(
-				user,
-				storedCredential,
-				decryptedData,
-				mergedCredentials,
-			);
-
-			if (mergedCredentials.data) {
-				mergedCredentials.data = this.unredact(
-					mergedCredentials.data,
-					decryptedData,
-					this.getCredentialTypeProperties(storedCredential.type),
-				);
-			}
-		}
-
-		return mergedCredentials;
-	}
-
 	async testById(userId: User['id'], credentialId: string) {
 		const storedCredential = await this.credentialsFinderService.findCredentialById(credentialId);
 
@@ -1364,5 +1317,52 @@ export class CredentialsService {
 		const scopes = await this.getCredentialScopes(user, credential.id);
 
 		return { ...credential, scopes };
+	}
+
+	/**
+	 * Build credentials payload ready to pass to credential testing.
+	 *
+	 * - If `credentialsToTest` is not provided, uses stored decrypted credential data.
+	 * - If `credentialsToTest` is provided, normalizes it for testing:
+	 *   - fills payload data for sharees when needed
+	 *   - restores redacted values from stored decrypted data
+	 */
+	private async prepareCredentialsForTest({
+		storedCredential,
+		user,
+		credentialsToTest,
+	}: {
+		storedCredential: CredentialsEntity;
+		user?: User;
+		credentialsToTest?: ICredentialsDecrypted;
+	}): Promise<ICredentialsDecrypted> {
+		const decryptedData = this.decrypt(storedCredential, true);
+		const mergedCredentials: ICredentialsDecrypted = credentialsToTest
+			? deepCopy(credentialsToTest)
+			: {
+					id: storedCredential.id,
+					name: storedCredential.name,
+					type: storedCredential.type,
+					data: decryptedData,
+				};
+
+		if (user && credentialsToTest) {
+			await this.replaceCredentialContentsForSharee(
+				user,
+				storedCredential,
+				decryptedData,
+				mergedCredentials,
+			);
+
+			if (mergedCredentials.data) {
+				mergedCredentials.data = this.unredact(
+					mergedCredentials.data,
+					decryptedData,
+					this.getCredentialTypeProperties(storedCredential.type),
+				);
+			}
+		}
+
+		return mergedCredentials;
 	}
 }
--- a/packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
+++ b/packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
@@ -5,10 +5,12 @@ import { CredentialsRepository } from '@n8n/db';
 import { Container } from '@n8n/di';
 import { hasGlobalScope } from '@n8n/permissions';
 import type express from 'express';
+import type { ICredentialsDecrypted } from 'n8n-workflow';
 import { z } from 'zod';
 
 import { CredentialTypes } from '@/credential-types';
 import { EnterpriseCredentialsService } from '@/credentials/credentials.service.ee';
+import { CredentialsService } from '@/credentials/credentials.service';
 import { CredentialsHelper } from '@/credentials-helper';
 import { ResponseError } from '@/errors/response-errors/abstract/response.error';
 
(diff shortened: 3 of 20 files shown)


packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
```diff
@@ -92,6 +94,51 @@ export = {
...
+			res: express.Response,
+		): Promise<express.Response<Partial<CredentialsEntity> | { message: string }>> => {
+			const { id: credentialId } = req.params;
+
+			const credential = await getCredential(credentialId);
+			if (!credential) {
+				return res.status(404).json({ message: 'Not Found' });
+			}
+
+			return res.json(sanitizeCredentials(credential as CredentialsEntity));
```
geemanjs: I feel like on all public api responses we should be "plucking" specifics out of mapping / service responses to prevent:
1. Accidental Breaking changes
2. Leaking data we explicitly don't want to. 

I feel like each endpoint should have a zod schema that we parse into dropping anything not explicitly in it.

geemanjs: It also feels weird to have the `as` cast in here

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
```diff
@@ -92,6 +94,51 @@ export = {
...
+			const credentialsService = Container.get(CredentialsService);
+			const decryptedData = credentialsService.decrypt(storedCredential as CredentialsEntity, true);
+			const mergedCredentials: ICredentialsDecrypted = {
+				id: storedCredential.id,
+				name: storedCredential.name,
+				type: storedCredential.type,
+				data: decryptedData,
+			};
+
+			const credentialTestResult = await credentialsService.test(req.user.id, mergedCredentials);
```
geemanjs: I also think this logic should be in a "service" somewhere that is called by both the "internal api" and the "public api". Rather than doing specific behaviour in the public api 👀

konstantintieber: +100. This is long standing issue of the existing public API of credentials. It leads to different bugs/security issues in one or the other.

alielkhateeb: @konstantintieber @geemanjs valid point, but I want to challenge a bit
The problem is that sometimes the API behaviours differ. In this case if we would unify the logic, we'd find ourselves battling in the service method with if conditions, and I am against doing that.

Example in this particular scenario: 
Public API: we load the stored credential by ID, and we test that
Internal: accepts payload, handles redacted values, sharee fallback/merge logic, then tests

The only common thing here is the "testing" part which is indeed in the service already `credentialsService.test`

Did I miss something and looking at the wrong place? 😄

konstantintieber: The latest public API rfc also proposes the same service being used for public and internal APIs: https://www.notion.so/n8n/Public-API-v2-3125b6e0c94f8053bc13c37dd5fec6f0?source=copy_link#33e5b6e0c94f80c89003c4649b1f64ba

Even if the public API endpoint behaves differently than an internal API endpoint, it'd be easier to keep API behaviour consistent and maintainable if they use the same service class (in that case potentially with two separate methods in the service, or parameters that change the behaviour of a single service method).

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
alielkhateeb: I fully agree, but I chose to follow the pattern in this file otherwise I'd be making much more changes than the scope of what's required here, but now that someone else is triggered I'll go ahead and make the mapper 😂 

I was so triggered by the multiple places in this file that do type casting :) so thank you for this comment 😂 pushing a mapper soon - and about the zod schema we already have https://github.com/n8n-io/n8n/blob/bb609f0ecf34db4e7f895188cb7c5f1b621b041e/packages/%40n8n/api-types/src/schemas/credential-created.schema.ts#L7 that I can reuse :)

alielkhateeb: 32ab9fcf7366df4faf173707e3b73743cbc94dd4

packages/cli/src/public-api/v1/handlers/credentials/spec/paths/credentials.id.yml
```diff
@@ -1,3 +1,29 @@
...
+    '200':
+      description: Operation successful.
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/create-credential-response.yml'
+    '401':
+      $ref: '../../../../shared/spec/responses/unauthorized.yml'
+    '404':
+      $ref: '../../../../shared/spec/responses/notFound.yml'
```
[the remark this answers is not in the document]
alielkhateeb: added here fd8f7169ad95a0d83402bd933ed206e6a507303a

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
alielkhateeb: Honestly speaking the fact that making an endpoint available to the public API requires one to write logic is already absurd 😂 It'd be great if internal apis get that decorator @IsPublic() and we're done 😂 but anyway:

I'll move these to a service tomorrow. And possibly create a ticket (or find the ticket if it already exists) to move the logic to services instead of having it in handlers (for other endpoints)

alielkhateeb: Just pushed 9a9a38f1b527c521b4e01f6a3d420a7d4f9dcc11
I am quite confused with why we have a credentials service in the public API and another in credentials "module", I assumed we want everything centralised in the credentials service that is [here](https://github.com/n8n-io/n8n/blob/9a9a38f1b527c521b4e01f6a3d420a7d4f9dcc11/packages/cli/src/credentials/credentials.service.ts)

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
```diff
@@ -92,17 +96,61 @@ export = {
...
 		publicApiScope('credential:create'),
 		async (
 			req: CredentialRequest.Create,
 			res: express.Response,
-		): Promise<express.Response<Partial<CredentialsEntity>>> => {
+		): Promise<express.Response<PublicApiCredentialResponse>> => {
 			const savedCredential = await saveCredential(req.body, req.user);
-
-			return res.json(sanitizeCredentials(savedCredential));
+			return res.json(savedCredential);
```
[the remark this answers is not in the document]
alielkhateeb: I think this is irrelevant now, please double check

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
alielkhateeb: After discussion, the plan is to eventually get rid of those public API services and rely on one source of truth for all business logic. This is documented [here](https://www.notion.so/n8n/Public-API-v2-3125b6e0c94f8053bc13c37dd5fec6f0?source=copy_link)

packages/@n8n/permissions/src/constants.ee.ts
```diff
@@ -71,7 +71,7 @@ export const API_KEY_RESOURCES = {
 	project: ['create', 'update', 'delete', 'list'] as const,
 	user: ['read', 'list', 'create', 'changeRole', 'delete', 'enforceMfa'] as const,
 	execution: ['delete', 'read', 'retry', 'list', 'get', 'stop'] as const,
-	credential: ['create', 'update', 'move', 'delete', 'list'] as const,
+	credential: ['create', 'read', 'update', 'move', 'delete', 'list'] as const,
```
konstantintieber, line 74: Something I just thought about, where I'm not sure which action to take:
The fact that this PR adds the `credential:read` scope to API scopes, means that no API key that was created before merging this PR will be able to use `GET /credentials/{id}`.

I don't think this is blocking the PR in any way.

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -819,6 +820,84 @@ export class CredentialsService {
...
 
+	/**
+	 * Build credentials payload ready to pass to credential testing.
+	 *
+	 * - If `credentialsToTest` is not provided, uses stored decrypted credential data.
+	 * - If `credentialsToTest` is provided, normalizes it for testing:
+	 *   - fills payload data for sharees when needed
+	 *   - restores redacted values from stored decrypted data
+	 */
+	private async prepareCredentialsForTest({
```
konstantintieber: Nit code style: private methods should come after public ones that use them.

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -819,6 +820,84 @@ export class CredentialsService {
...
+		if (!storedCredential) {
+			throw new CredentialNotFoundError(credentialId);
+		}
+
+		const credentials = await this.prepareCredentialsForTest({ storedCredential });
+		return await this.test(userId, credentials);
+	}
+
+	async testWithCredentials(user: User, credentials: ICredentialsDecrypted) {
+		const storedCredential = await this.credentialsFinderService.findCredentialForUser(
```
konstantintieber, line 835: Nice job on moving this logic from the CredentialsController to the CredentialService.

packages/cli/src/credentials/credentials.service.ts
```diff
@@ -819,6 +820,84 @@ export class CredentialsService {
...
+			credentials.id,
+			user,
+			['credential:read'],
+		);
+
+		if (!storedCredential) {
+			throw new CredentialNotFoundError(credentials.id);
```
konstantintieber, line 842: The [current controller implementation](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/credentials/credentials.controller.ts#L152) throws `new ForbiddenError()`, will this `CredentialNotFoundError` be mapped to the same status code as before?

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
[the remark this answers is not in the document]
konstantintieber: @alielkhateeb are you referring to this zod.safeParse call that was added last week?
https://github.com/n8n-io/n8n/blob/master/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts#L124

packages/cli/src/credentials/credentials.service.ts
alielkhateeb: @konstantintieber Yes we catch the CredentialNotFoundError and throw a ForbiddenError here
https://github.com/n8n-io/n8n/blob/99cd970fd6bd2b8a98e4747187b07d924b046e65/packages/cli/src/credentials/credentials.controller.ts#L146-L148

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
```diff
@@ -113,7 +159,7 @@ export = {
 		async (
 			req: CredentialRequest.Update,
 			res: express.Response,
-		): Promise<express.Response<Partial<CredentialsEntity>>> => {
+		): Promise<express.Response<PublicApiCredentialResponse | { message: string }>> => {
```
konstantintieber: This is the first time that I see the return type of a public API endpoint by a union type of the response and `{ message: string }`.
What's the use case for this one?

packages/cli/src/credentials/credentials.service.ts
alielkhateeb: @konstantintieber Agreed, I have in mind adding `@typescript-eslint/member-ordering` at some point 😂 I need to create a ticket for it

alielkhateeb: 60a68d1847fae58be48c990e12bb70b2aac817d6

packages/cli/src/public-api/v1/handlers/credentials/credentials.mapper.ts
```diff
@@ -0,0 +1,28 @@
...
+	},
+): PublicApiCredentialResponse {
+	const parsed = publicApiCredentialResponseSchema.safeParse({
+		...credential,
+		resolvableAllowFallback: credential.resolvableAllowFallback ?? false,
+		resolverId: credential.resolverId ?? null,
+	});
+
+	if (!parsed.success) {
+		throw new UnexpectedError('Public API credential response failed validation');
```
konstantintieber: This error message seems unconventional to me. As far as I know we don't prefix errors that happen on public endpoints with 'Public API ...'.

When debugging such errors, we already know from the url path that it originates from the public api.

packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
```diff
@@ -89,7 +89,7 @@ export async function getSharedCredentials(
 export async function saveCredential(
 	payload: { type: string; name: string; data: ICredentialDataDecryptedObject; projectId?: string },
 	user: User,
-): Promise<CredentialsEntity> {
+): Promise<PublicApiCredentialResponse> {
```
konstantintieber, line 92: The CredentialsService relying on public API types as return values seems unintuitive to me. The service should by design not be specific to public or internal API.
What do you think about the public API handler calling the `toPublicApiCredentialResponse` on the `CredentialsEntity` instead?

packages/cli/src/public-api/v1/handlers/credentials/credentials.handler.ts
alielkhateeb: This is because instead of throwing an error we are still returning an object with a message, I can change that to throw the appropriate error instead, this should not have any impact since we recently centralised error handling.

alielkhateeb: @sandra0503 Is there a ticket to actually change all those instances of `res.status(xxx)` to throw the appropriate error?

konstantintieber: I see, thx for explaining.
I think throwing the error should be preferred, to avoid patterns where many return signature on our API also include `{ message: string }` for some error cases.

alielkhateeb: @konstantintieber d2b359748466c32ba0d183f3d00e7099c65c7406 this should be how it is for ALL handlers, at some point 😂

packages/cli/src/public-api/v1/handlers/credentials/credentials.mapper.ts
alielkhateeb: Absolutely, this is something AI did that I missed 🤦

alielkhateeb: @konstantintieber 242988a47743403ad57dd1eee42a549ba1d9a26a also included unit tests for this method (which were missing too)

packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
alielkhateeb: Yeah I totally agree, this was meant to keep the service return shape backward compatible since we used to parse in-line:
https://github.com/n8n-io/n8n/blob/a3292b738ab904df3c4dc2d0e8ff7f45e9147789/packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts#L124-L129

And I did not like that 😂, I will look into making this happen in the handler and adjust the types here accordingly

alielkhateeb: 418b4daa1ca0904709e3aaf2af8429ea431f4c6c What do you think of this? I don't like the "hardcoded" type but for now it should be fine?

packages/@n8n/permissions/src/constants.ee.ts
alielkhateeb: But isn't that exactly what we expect?
We added this new scope and guarded `GET /credentials/{id}` with `publicApiScope('credential:read')` so it makes sense that previously generated API keys would NOT have access to it, right?

konstantintieber: Yes, it's technically correct.
But from a UX perspective of a user having an API token that is able to update a credential via the public API, but not fetch the same credential, is unfortunate.
From a business perspective, I'd want my `update` token also be able to `read`, but I wasn't able to select that as scope until now.
Customers will need to update the API key of their integration in order to call the read endpoint.

packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts
konstantintieber: My bad, I think here it's actually fine to use the PublicApiCredentialsResponse as return type in the `packages/cli/src/public-api/v1/handlers/credentials/credentials.service.ts`, because that one is specific to the public API.
I should've looked better at the file path. I just read "service" and assumed this would be a proper service pattern. But it turns out it's just the service for the public API, which we want to get rid of in the future anyway.

konstantintieber: The previous implementation achieved the same result with less code, and in an environment that is isolated to the public API. So I think it's actually the better one to keep.

alielkhateeb: @konstantintieber Reverted back, like you said it really doesn't matter we'll get rid of that at some point.

I just agree that http responses should not appear in any service, that's why I implemented the change above, but again, doesn't really matter 😂

packages/@n8n/permissions/src/constants.ee.ts
alielkhateeb: @konstantintieber Ah, got it!
True that sounds a bit weird from a UX perspective

konstantintieber (approved):
Great work!

PR #28522, opened 2026-04-15
feat(Linear Trigger Node): Add signing secret validation

burivuhster:
## Summary

Add an optional `signingSecret` field to both `LinearApi` and `LinearOAuth2Api` credentials, and verify incoming webhook signatures in the Linear Trigger node.

**How it works:**
- Both Linear credential types now have a `Signing Secret` password field
- `LinearTriggerHelpers.ts` verifies the `Linear-Signature` header using HMAC-SHA256 (same pattern as Slack)
- Validates `webhookTimestamp` from the body payload within a 60-second window (per [Linear docs](https://linear.app/developers/webhooks))
- Backward compatible: verification is skipped when no signing secret is configured

**How to test:**
1. Create a Linear credential and note the new "Signing Secret" field
2. Activate a workflow having Linear Trigger, n8n will register webhook in Linear
3. Go to your Linear instance's settings -> API -> Webhooks, find a webhook created by n8n, and copy its signing secret into the credential

- https://linear.app/n8n/issue/AGENT-23/feature-add-signing-secret-option-to-linear-credentials

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/nodes-base/nodes/Linear/LinearTriggerHelpers.ts
```diff
@@ -0,0 +1,49 @@
...
+					return null;
+				}
+
+				const hmac = createHmac('sha256', signingSecret);
+
+				if (Buffer.isBuffer(req.rawBody)) {
+					hmac.update(req.rawBody);
+				} else {
+					const rawBodyString =
+						typeof req.rawBody === 'string' ? req.rawBody : JSON.stringify(req.rawBody);
```
dlavrenuek: Can there be a case where `rawBody` would not be a Buffer or string? If it is an object then `JSON.stringify(...)` will only produce the same string for hash comparison if the body did not contain any extra whitespace characters compared to what `JSON.stringigy` produces f.e. if it is a nicely formatted json with new lines and identation

packages/nodes-base/nodes/Linear/test/LinearTriggerHelpers.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+			expect(result).toBe(true);
+		});
+
+		it('should return true when signature is valid', async () => {
+			mockWebhookFunctions.getCredentials.mockResolvedValue({
+				signingSecret: testSigningSecret,
+			});
+
+			(timingSafeEqual as jest.Mock).mockReturnValue(true);
```
dlavrenuek: This test and `'should return false when signature is invalid'` do not verify the correct implementation. What this test does is asserting that `verifySignature` returns true if `timingSafeEqual` returns true, which is one very specific implementation detail. I suggest using real hmac hash using the specified test signature key + test request body

dlavrenuek (changes requested):
Hi @burivuhster, thank you for this nice improvement! The code looks good, I only have one question there and a suggestion on test cases.

packages/nodes-base/nodes/Linear/LinearTriggerHelpers.ts
burivuhster: Good catch! You are right, `getRawBody` that we use produce only `Buffer` or `string`

dlavrenuek (approved):
Thank you for the improvements and addressing my comments 🚀

PR #28523, opened 2026-04-15
fix(editor): Address design feedback for favoriting (no-changelog)

CharlieKolb:
## Summary

Adjust gap between resource groups, use raw icon for star and provide a custom filled alternative if favorites. Adjust this icon and sidebar section collapse chevron from small to medium icon size.

Also always show the project favoriting icon and remove the tooltip on the star.

n/a

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectHeader.vue
```diff
@@ -451,27 +453,15 @@ const onSelect = (action: string) => {
 						</div>
 					</template>
 				</div>
-				<N8nTooltip
+				<N8nIcon
```
RobTF9: This needs to be an icon button, not just an Icon. Color is also way too subtle in light mode and dark mode.

Image

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectNavigation.vue
```diff
@@ -237,7 +237,7 @@ onBeforeUnmount(() => {
 				</N8nText>
 				<N8nIcon
```
RobTF9, line 247: Also should be a button, not an icon. Needs to be centered with the text and less subtle

Image

RobTF9 (changes requested):
Added some changes to make, also nice to haves would be a way to favorite/unfavorite from the sidebar itself.

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectHeader.vue
CharlieKolb: To confirm, you want the lil border around it? 
image

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectNavigation.vue
CharlieKolb: Note that this is structured this way to allow clicking on Projects itself to also collapse/expand. I figured it looking like an actual button is misleading in this case

RobTF9 (changes requested):
Can you remove this popover as well please? Just show the favourites like we do the other items in the sidebar:

Image

packages/frontend/editor-ui/src/features/collaboration/projects/components/ProjectNavigation.vue
RobTF9: Ah, missed the whole button wrapper. In that case - it needs a background color on hover, and a focus state (same as the menutitems), the chevron should be the same color as the the other ones (on the settings popover).

CharlieKolb: image
Like this?

RobTF9 (approved):

PR #28533, opened 2026-04-15
feat(core): Add KeyManagerService for encryption key lifecycle management

BGZStephen:
## Summary

Introduces the `encryption-key-manager` module with `KeyManagerService` — a single authoritative
service for managing the full lifecycle of data encryption keys stored as rows in `deployment_key`.

This is **T06** of the encryption key rotation epic. T01 created the `deployment_key` table and
`DeploymentKeyRepository`; this task builds the service layer on top.

### What's added

**`packages/cli/src/modules/encryption-key-manager/`**

| File | Purpose |
|------|---------|
| `encryption-key-manager.module.ts` | `@BackendModule` entry point, registered in the framework's module registry |
| `key-manager.service.ts` | `@Service()` implementing the full key lifecycle API |
| `__tests__/key-manager.service.test.ts` | 18 unit tests covering all methods and acceptance criteria |

**`packages/@n8n/backend-common/src/modules/`**

- `modules.config.ts` — `'encryption-key-manager'` added to `MODULE_NAMES`
- `module-registry.ts` — `'encryption-key-manager'` added to `defaultModules`

### API surface (`KeyManagerService`)

| Method | Description |
|--------|-------------|
| `getActiveKey()` | Returns the active encryption key; throws if none or if the DB invariant is violated (multiple active rows) |
| `getKeyById(id)` | Returns a key by ID, or `null` |
| `getLegacyKey()` | Returns the legacy `aes-256-cbc` key used to decrypt rows with `NULL encryptionKeyId` |
| `addKey(value, algorithm, setAsPrimary?)` | Inserts a new key; demotes current active to `'deprecating'` first when `setAsPrimary=true` |
| `setPrimaryKey(id)` | Promotes a key to `'active'`, demoting the current active first |
| `markDeprecating(id)` | Transitions a key to `'deprecating'` and records `deprecatedAt` |
| `markInactive(id)` | Transitions a key to `'inactive'` (T13 will add usage-count guard) |
| `removeKey(id)` | Deletes the key row (T13 will add usage-count guard) |

### Key design decisions

- **Partial unique index safety**: `markDeprecating()` is always called _before_ any key is set to
  `'active'`, ensuring the `UNIQUE(type) WHERE status = 'active'` constraint is never violated.
- **Invariant detection**: `getActiveKey()` uses `find()` (not `findOne()`) so it can detect and
  throw on multiple active rows rather than silently returning an ambiguous result.
- **ID generation**: Entity IDs are auto-generated via `generateNanoId()` in the `@BeforeInsert()`
  hook on `WithTimestampsAndStringId` — no manual UUID generation needed.
- **T13 stubs**: `markInactive` and `removeKey` contain `// TODO: T13` comments where usage-count
  guards will be added; no `ConflictError` is thrown yet.

https://linear.app/n8n/issue/IAM-489

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+	async getKeyById(id: string): Promise<KeyInfo | null> {
+		const key = await this.deploymentKeyRepository.findOne({ where: { id } });
+		if (!key || !key.algorithm) return null;
+		return { id: key.id, value: key.value, algorithm: key.algorithm };
+	}
+
+	/** Returns the legacy CBC key (used to decrypt rows with NULL encryptionKeyId). */
+	async getLegacyKey(): Promise<KeyInfo> {
+		const key = await this.deploymentKeyRepository.findOne({
+			where: { type: 'encryption', algorithm: 'aes-256-cbc' },
```
guillaumejacquart: Should we use a custom ID or something to return the legacy key ? Instead of relying on the algorithm name

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+		});
+		if (activeKeys.length === 0) {
+			throw new NotFoundError('No active encryption key found');
+		}
+		if (activeKeys.length > 1) {
+			throw new Error('Encryption key invariant violated: multiple active keys found');
+		}
+		const key = activeKeys[0];
+		if (!key.algorithm) {
+			throw new NotFoundError('Active encryption key has no algorithm set');
```
guillaumejacquart: Maybe we could use a more fitting error type here

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+		return { id: key.id, value: key.value, algorithm: key.algorithm! };
+	}
+
+	/**
+	 * Inserts a new encryption key row.
+	 * If setAsPrimary is true, transitions the current active key to 'deprecating' first,
+	 * then inserts the new key as 'active'. Otherwise inserts as 'inactive'.
+	 */
+	async addKey(value: string, algorithm: string, setAsPrimary = false): Promise<{ id: string }> {
+		if (setAsPrimary) {
```
guillaumejacquart: i'm seeing the wording `markInactive` below. Maybe we should use `setAsActive` ?

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+		});
+		const saved = await this.deploymentKeyRepository.save(entity);
+		return { id: saved.id };
+	}
+
+	/**
+	 * Sets the given key as active.
+	 * Transitions the current active key to 'deprecating' first.
+	 */
+	async setPrimaryKey(id: string): Promise<void> {
```
guillaumejacquart: maybe `setActiveKey`

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+	/** Transitions key to 'inactive'. Usage count guard to be added in T13. */
+	async markInactive(id: string): Promise<void> {
+		// TODO: T13 will add usage check — throw ConflictError if usage count > 0
+		await this.deploymentKeyRepository.update(id, { status: 'inactive' });
+	}
+
+	/** Removes the key row. Usage count guard to be added in T13. */
+	async removeKey(id: string): Promise<void> {
+		// TODO: T13 will add usage check — throw ConflictError if usage count > 0
+		await this.deploymentKeyRepository.delete(id);
```
guillaumejacquart: do we allow having no active keys ? or should we add some checks here ?

BGZStephen: this actually isn't needed for the moment, so will delete this.

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
BGZStephen: I don't think so, as the ID's are auto generated it'd feel a bit odd to have a static one, it seems reasonable business logic to expect only one CBC key here

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
BGZStephen: Any suggestions? Not Found doesn't seem too wrong unless there is a better one

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
guillaumejacquart: all righty, I guess it makes sense since we're in control of the algorithm and cbc is actually legacy

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
guillaumejacquart: I was thinking of a dedicated error introduced by the service, or maybe an unexpected error, as this is unexpected no ?

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,88 @@
...
+	/**
+	 * Inserts a new encryption key row.
+	 * If setAsActive is true, transitions the current active key to 'inactive' first,
+	 * then inserts the new key as 'active'. Otherwise inserts as 'inactive'.
+	 */
+	async addKey(value: string, algorithm: string, setAsActive = false): Promise<{ id: string }> {
+		if (setAsActive) {
+			const current = await this.deploymentKeyRepository.findActiveByType('encryption');
+			if (current) {
+				await this.markInactive(current.id);
```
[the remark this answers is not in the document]
guillaumejacquart: That feedback seems reasonable

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
BGZStephen: I actually removed this for the timebeing, the algorithm checks isn't actually right for this function, so it's overkill

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -0,0 +1,85 @@
...
+			where: { type: 'data_encryption', status: 'active' },
+		});
+		if (activeKeys.length === 0) {
+			throw new NotFoundError('No active encryption key found');
+		}
+		if (activeKeys.length > 1) {
+			throw new Error('Encryption key invariant violated: multiple active keys found');
+		}
+		const key = activeKeys[0];
+		return { id: key.id, value: key.value, algorithm: key.algorithm! };
```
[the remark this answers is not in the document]
BGZStephen: This should not be the concern of this function, and will be handled outside of this function

packages/@n8n/db/src/repositories/deployment-key.repository.ts
```diff
@@ -28,4 +28,28 @@ export class DeploymentKeyRepository extends Repository<DeploymentKey> {
 		const entity = this.create(entityData);
 		await this.createQueryBuilder().insert().values(entity).orIgnore().execute();
 	}
+
+	/** Atomically deactivates any existing active key of the same type, then saves the given entity as active. */
+	async insertAsActive(entity: DeploymentKey): Promise<DeploymentKey> {
```
guillaumejacquart: this feels a little weird that the function is called `insertAsActive` for its argument can be a deployment key with status: 'active' (nothing prevents it).
Maybe we could:
- ensure the entity is active by setting entity.status = 'active' before the save
- and/or passing a subset of fields, without the status, as argument

guillaumejacquart (commented):
One last comment

packages/@n8n/db/src/repositories/deployment-key.repository.ts
BGZStephen: Ok, let me tighten the type to avoid this, should be easy enough

guillaumejacquart (approved):
Thanks for handling the feedbacks

PR #28545, opened 2026-04-15
fix(schedule): enforce dom+dow constraint handling

Aijeyomah:
## Summary

- Add Schedule Trigger v1.4 with AND matching for constrained day-of-month and day-of-week in custom cron

## Before fix
Screenshot 2026-04-15 at 14 05 32

## After fix
Screenshot 2026-04-15 at 22 25 19

closes #27238 

## Review / Merge checklist

- [x] Tests included. 


Aijeyomah:
@krisn0x Please review

Aijeyomah:
@sandra0503 @BerniWittmann Please review 🙏

BerniWittmann:
@Aijeyomah There have been now quite some changes and fixes in the ScheduleTrigger node (as you can see from the conflicts)
Is this issue still happening / the change request still relevant?

ShireenMissi:
Thanks for taking the time to put this together, we appreciate the contribution. After discussing internally, we're not going to merge this one.

The current behavior matches the POSIX cron spec: when both day-of-month and day-of-week are restricted (neither is *), the job fires when either one matches. Changing this to AND semantics would make n8n's cron parsing behave differently from cron everywhere else, which would likely cause more confusion than it solves.

We're looking into ways to make this clearer in the UI. For anyone needing AND behavior in the meantime, the workaround is to restrict only one field in the expression and check the other with an IF node after the trigger.

Closing, but thanks again for the effort here!

PR #28567, opened 2026-04-16
fix(Google Gemini Node): Filter out undefined parts when processing API responses

RomanDavydchuk:
## Summary

If Google's API response contained undefined parts, for some reason, the node would fail because `'functionCall' in p` throws an error when `p` is undefined. This PR adds extra checks to make sure we properly filter out unwanted parts of the response

https://linear.app/n8n/issue/NODE-4822/community-issue-googlegemini-node-typeversion-11-crashes-with-cannot
Closes #28215 

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/actions/text/message.operation.ts
```diff
@@ -538,7 +549,10 @@ export async function execute(this: IExecuteFunctions, i: number): Promise<INode
 			break;
 		}
 
-		contents.push(...response.candidates.map((c) => c.content));
+		contents.push.apply(
+			contents,
+			response.candidates.map((c) => c.content),
+		);
```
michael-radency, line 560: claude:
push.apply offers no performance advantage over spread in modern JS engines. Both translate to the same operation: passing array elements as individual arguments to push.

The old push.apply pattern came from pre-ES6 days when spread didn't exist. It was used as a workaround. Today, contents.push(...arr) is the idiomatic equivalent and compiles to the same bytecode in V8.

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/GoogleGemini.node.test.ts
```diff
@@ -730,6 +731,54 @@ describe('GoogleGemini Node', () => {
 				);
 			});
 
+			it('should skip undefined parts when parsing tool calls', async () => {
+				executeFunctionsMock.getNodeParameter.mockImplementation((parameter: string) => {
```
michael-radency: maybe better name 
"should handle undefined entries in parts array"

michael-radency (dismissed):
LGTM, added some non-blocking comments

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/actions/text/message.operation.ts
RomanDavydchuk: I think this got changed because I ran eslint autofix, and we have a lint rule to prefer `apply` over spread

The reasoning: if the array being spread is very big, it can cause stack overflow, because each array item is converted into a function argument, therefor it takes a lot of memory on the stack

This doesn't apply here, because `contents` is always rather small, but I'll leave it like this to keep the linter happy

packages/@n8n/nodes-langchain/nodes/vendors/GoogleGemini/GoogleGemini.node.test.ts
RomanDavydchuk: Agree, updated it

michael-radency (approved):

PR #28573, opened 2026-04-16
fix(core): Guard against undefined config properties in credential overwrites

Joffcom:
## Summary

Fixes an intermittent `TypeError: Cannot read properties of undefined (reading 'skipTypes')` crash that caused nodes to fail when running workflows with credentials.

**Root causes (two bugs):**

1. `CredentialsOverwrites.get()` always returned `{}` instead of `undefined` when no overwrites existed for a credential type. This made the early-return guard at line 142 dead code, meaning every credential lookup unnecessarily reached the `skipTypes` access.

2. The `skipTypes` access at line 150 was unsafe. Based on the error stack — "reading 'skipTypes' from undefined" — the `overwrite` config object itself was `undefined`. This can occur during rolling deploys or self-hosted partial package updates where `credentials-overwrites.ts` is updated before the corresponding `@n8n/config` version (which introduced `skipTypes`) is active.

**Fixes:**
- `get()` now returns `undefined` when no entries are found, restoring the intended early-return behaviour
- Line 150 uses optional chaining (`overwrite?.skipTypes?.includes(type)`) to safely handle both `overwrite` and `skipTypes` being undefined

**Tests added:** two new cases covering `skipTypes === undefined` and `overwrite === undefined`.

https://linear.app/n8n/issue/NODE-4854/intermittent-skiptypes-error

## Review / Merge checklist

- [x] Tests included.

---
🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/__tests__/credentials-overwrites.test.ts
+++ b/packages/cli/src/__tests__/credentials-overwrites.test.ts
@@ -113,6 +113,36 @@ describe('CredentialsOverwrites', () => {
 			expect(result).toEqual(data);
 		});
 
+		it('should not crash when skipTypes is undefined', () => {
+			// Simulate version mismatch where skipTypes is not present on the config object
+			globalConfig.credentials.overwrite.skipTypes =
+				undefined as unknown as CommaSeparatedStringArray<string>;
+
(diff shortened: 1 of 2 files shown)


Matsuuu (approved):
Awesome 💯

PR #28581, opened 2026-04-16
fix: Validate loadOptions routing to prevent SSRF via absolute URLs

BGZStephen:
## Summary

Adds validation to `loadOptions` routing so that user-supplied `routing.request.url` values are restricted to relative paths or expression strings (those beginning with `=`). Absolute URLs — including `http://`, `https://`, `file://`, and protocol-relative `//` forms — are now rejected with a `400 Bad Request`.

Additionally, `baseURL` is stripped from any user-supplied `routing.request` object before it reaches the executor, ensuring it cannot override the node's `requestDefaults.baseURL`.

Validation is applied at two layers:

- **DTO level** (`OptionsRequestDto`): the `loadOptions.routing.request.url` field is validated via a Zod refinement before the request reaches the service.
- **Service level** (`DynamicNodeParametersService.sanitizeLoadOptionsRouting`): a dedicated private method enforces the same URL constraints and strips `baseURL` before the routing object is passed to the workflow executor.

### Happy path — relative URL is accepted

1. Start n8n locally.
2. Open any workflow and add a node that loads options via routing (e.g. a custom community node or the built-in HTTP Request node with dynamic parameter routing).
3. Click the parameter dropdown that triggers an options load.
4. Verify that options load successfully (no regression).

### Blocked path — absolute URL is rejected

1. Log in to n8n and open any workflow.
2. Open **DevTools → Network**.
3. Trigger any dropdown that fires a `POST /rest/dynamic-node-parameters/options` request.
4. Right-click the request → **Copy as cURL**.
5. Paste the curl command into a terminal and modify the `loadOptions` body to include an absolute URL:

```json
"loadOptions": {
  "routing": {
    "request": {
      "url": "http://127.0.0.1:5678/rest/settings"
    }
  }
}
```

6. Run the modified curl command.
7. Verify the response is **`400 Bad Request`** with message `routing.request.url must be a relative path, not an absolute URL`.

### Blocked path — `baseURL` override is stripped

Repeat step 5–6 above but instead of (or in addition to) an absolute `url`, include `"baseURL": "http://attacker.example.com"` in `routing.request`. Confirm the request either completes without hitting the attacker URL, or returns a 400 if the `url` field is also invalid.

https://linear.app/n8n/issue/IAM-548

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/api-types/src/dto/dynamic-node-parameters/options-request.dto.ts
+++ b/packages/@n8n/api-types/src/dto/dynamic-node-parameters/options-request.dto.ts
@@ -3,6 +3,9 @@ import { z } from 'zod';
 
 import { BaseDynamicParametersRequestDto } from './base-dynamic-parameters-request.dto';
 
+const isRelativeOrExpressionUrl = (url: string) =>
+	url.startsWith('=') || (!/^[a-zA-Z][a-zA-Z0-9+\-.]*:\/\//.test(url) && !url.startsWith('//'));
+
 export class OptionsRequestDto extends BaseDynamicParametersRequestDto.extend({
 	loadOptions: z
 		.object({
@@ -10,7 +13,13 @@ export class OptionsRequestDto extends BaseDynamicParametersRequestDto.extend({
 				.object({
 					operations: z.any().optional(),
 					output: z.any().optional(),
-					request: z.any().optional(),
+					request: z
+						.object({ url: z.string().optional() })
+						.passthrough()
+						.refine((req) => req?.url === undefined || isRelativeOrExpressionUrl(req.url), {
+							message: 'routing.request.url must be a relative path, not an absolute URL',
+						})
+						.optional(),
 				})
 				.optional(),
 		})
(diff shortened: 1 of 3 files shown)


guillaumejacquart (commented):
I think this will lead to regressions.
I tried this template: https://n8n.io/workflows/2465-building-your-first-whatsapp-chatbot
Which used an openai embedding node (EmbeddingsOpenAi) that uses this loadOptions mechanism (dynamic node parameters) with expressions expressions. 
Got this error on node loading: 
image
Example impacted nodes:

Node | File:line | URL
-- | -- | --
OpenAI Chat Model (Model dropdown) | packages/@n8n/nodes-langchain/nodes/llms/LMChatOpenAi/LmChatOpenAi.node.ts:145 | {{ $parameter.options?.baseURL?.split("/").slice(-1).pop() \|\| $credentials?.url?.split("/").slice(-1).pop() \|\| "v1" }}/models
Embeddings OpenAI (Model dropdown) | packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOpenAI/EmbeddingsOpenAi.node.ts:29 | Same shape as above
OpenAI Assistant (Model + Assistant dropdowns) | packages/@n8n/nodes-langchain/nodes/agents/OpenAiAssistant/OpenAiAssistant.node.ts:132, 194 | Same shape as above

BGZStephen:
Thanks @guillaumejacquart, I'll close this amd advise it needs to wait for proper SSRF protection coming from global configuration

PR #28584, opened 2026-04-16
fix(ClickUp Node): Unclear error message when using OAuth credentials

### LINKED ISSUE #14277: ClickUp trigger consistently returns errors and is unstable
### Bug Description

"ClickUp Trigger: Webhook configuration already exists" or "refreshToken is required" is consistently presented to the user - both are widely reported issues with seemingly no fix available. This makes the entire Click Up module useless as any small changes to the workflow breaks everything.

### To Reproduce

i dont have reproduction steps for you please just try to use these and build a workflow

### Operating System

xxx

### n8n Version

xxx

### Node.js Version

xxx

### Database

SQLite (default)

### Execution mode

main (default)

RomanDavydchuk:
## Summary

ClickUp doesn't support token refresh for OAuth access token, so it does not return the refresh token when authorizing and the access token has no expiry ([docs](https://developer.clickup.com/docs/authentication#step-3-request-a-token)). Besides that, if a user tries to access a resource (task, list, etc) by ID and it's not a valid ID, ClickUp's API returns a 401 with `{ "err": "Team(s) not authorized", "ECODE": "OAUTH_023" }`. n8n by default would try to initiate a token refresh, but will throw an error because the refresh token is missing [here](https://github.com/n8n-io/n8n/blob/c6534fa0b389a394e7591d3fc5ec565409279004/packages/%40n8n/client-oauth2/src/client-oauth2-token.ts#L77). This PR fixes it by adding a flag to disable token refresh on 401. This way, the behavior and errors are consistent between using OAuth and Access Token

#### Video of behavior before the fix:
https://github.com/user-attachments/assets/448df6cd-acb3-4e46-9830-58ba2e32a678

#### Video of behavior after the fix:
https://github.com/user-attachments/assets/4f96a9f4-caa6-4f51-9efa-30ee60bd71a8

https://linear.app/n8n/issue/NODE-4545/community-issue-clickup-trigger-consistently-returns-errors-and-is
Closes #14277

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/ClickUp/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/ClickUp/GenericFunctions.ts
@@ -41,6 +41,9 @@ export async function clickupApiRequest(
 			const oAuth2Options: IOAuth2Options = {
 				keepBearer: false,
 				tokenType: 'Bearer',
+				// ClickUp's access token don't expire and
+				// ClickUp does not return refresh tokens
+				skipTokenRefresh: true,
 			};
 			return await this.helpers.requestOAuth2.call(
 				this,
--- a/packages/nodes-base/nodes/ClickUp/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/ClickUp/GenericFunctions.ts
@@ -41,7 +41,7 @@ export async function clickupApiRequest(
 			const oAuth2Options: IOAuth2Options = {
 				keepBearer: false,
 				tokenType: 'Bearer',
-				// ClickUp's access token don't expire and
+				// ClickUp's access token doesn't expire and
 				// ClickUp does not return refresh tokens
 				skipTokenRefresh: true,
 			};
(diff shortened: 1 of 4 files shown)


packages/nodes-base/nodes/ClickUp/GenericFunctions.ts
```diff
@@ -41,6 +41,9 @@ export async function clickupApiRequest(
 			const oAuth2Options: IOAuth2Options = {
 				keepBearer: false,
 				tokenType: 'Bearer',
+				// ClickUp's access token don't expire and
```
DawidMyslak:
```suggestion
				// ClickUp's access token doesn't expire and
```

DawidMyslak (dismissed):

RomanDavydchuk:
@DawidMyslak GH dismissed your approval after I applied your suggestion 😅 Not the best UX, can you reapprove please?

DawidMyslak (approved):

PR #28586, opened 2026-04-16
refactor: Use napi-rs/image instead of sharp for screenshots

Cadiac:
## Summary

`npx @n8n/computer-use ...` commands failed silently with exit 1 on machines that have a system-wide `libvips` (e.g. installed via homebrew). Sharp's install script (`useGlobalLibvips()`in `sharp/lib/libvips.js`) detects a compatible system `libvips` and prefers to build from source against it - but building requires `node-gyp`, which isn't in sharp's deps. npx swallows the install error, so the user sees no banner, no logs, and no clue what went wrong.

I did consider adding `node-gyp` as a dependency but decided against it as that bundles a full C++ toolchain footprint for what amounts to one `toJpeg` helper. 

`@napi-rs/image` (MIT) is a pure-Rust image library with statically linked prebuilds per platform. It has no system-library detection path and no build-from-source fallback, so install either succeeds with a prebuilt binary or fails loudly. Prebuild coverage is equivalent to sharp for our use case (desktop tool: macOS arm64/x64, Windows x64/arm64, Linux x64/arm64 glibc+musl). Admittedly I have no experience from using this library before, but `napi-rs` ecosystem seems reasonably actively developed.

## Review / Merge checklist

- [ ] Tests included. 


Cadiac:
I'll test this a bit now to ensure it works, this didn't get enough testing last week.

dlavrenuek:
Works perfectly, thank you @Cadiac 🚀

Cadiac:
Also tested earlier locally that this did in fact did the issue I was seeing on my machine

dlavrenuek (approved):

dlavrenuek:
Looks like I clicked on "comment" instead of "approve" last time 🙈

PR #28594, opened 2026-04-16
fix(core): Guard event log parsing against unbounded memory growth

guillaumejacquart:
## Summary

Adds a configurable working-set guard to `readLoggedMessagesFromFile` so startup recovery aborts parsing a single event log file when the in-memory message count exceeds `N8N_EVENTBUS_LOGWRITER_MAXMESSAGESPERPARSE` (default `10000`). Healthy logs are unaffected because confirm records prune the working set as the file streams; only legacy files full of unconfirmed messages (pre-`PR #27334`) trip the guard, which prevents the startup crash loop on Starter-plan containers.

**How to test:** unit tests in `message-event-bus-log-writer.test.ts` cover both the bloat-abort path and the healthy paired-confirm path. To verify manually, seed `~/.n8n/n8nEventLog.log` with >10k orphaned `n8n.workflow.started` lines, restart n8n, and observe the warn `Event log ... exceeded 10000 in-memory messages during parse` instead of an OOM.

https://linear.app/n8n/issue/IAM-528

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/eventbus/message-event-bus-writer/message-event-bus-log-writer.ts
+++ b/packages/cli/src/eventbus/message-event-bus-writer/message-event-bus-log-writer.ts
@@ -205,51 +205,27 @@ export class MessageEventBusLogWriter {
 	): Promise<ReadMessagesFromLogFileResult> {
 		if (logFileName && existsSync(logFileName)) {
 			try {
+				const stream = createReadStream(logFileName);
 				const rl = readline.createInterface({
-					input: createReadStream(logFileName),
+					input: stream,
 					crlfDelay: Infinity,
 				});
+				// Safety guard: abort if the in-memory working set grows too large.
+				// Healthy files stay small because confirm lines prune loggedMessages as
+				// they stream; legacy files with orphaned messages (pre-PR #27334) are
+				// the pathological case this guards against.
+				const maxMessagesPerParse = this.globalConfig.eventBus.logWriter.maxMessagesPerParse;
+				let aborted = false;
 				rl.on('line', (line) => {
-					try {
-						const json = jsonParse(line);
-						if (isEventMessageOptions(json) && json.__type !== undefined) {
-							const msg = this.getEventMessageObjectByType(json);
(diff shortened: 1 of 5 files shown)


packages/cli/src/eventbus/message-event-bus-writer/message-event-bus-log-writer.ts
```diff
@@ -205,51 +205,40 @@ export class MessageEventBusLogWriter {
 	): Promise<ReadMessagesFromLogFileResult> {
 		if (logFileName && existsSync(logFileName)) {
 			try {
+				const stream = createReadStream(logFileName);
+				stream.on('error', (error) => {
+					if ((error as NodeJS.ErrnoException).code !== 'ERR_STREAM_DESTROYED') {
```
phyllis-noester, line 210: so we swallow all errors and never throw?

cstuncsik (approved):

phyllis-noester (approved):
this doesn't work for rolling updates/in multi main or with workers right? not your change but ingesting the logs from a local file in general

packages/cli/src/eventbus/message-event-bus-writer/message-event-bus-log-writer.ts
guillaumejacquart: Yep, errors are logged but not propagated. I think this makes sense, as we don't want this recovery process (furthermore a single log file reading) to hard fail the whole instance. It's a not critical recovery AFAIK (for instance, we've allowed ourselves to tamper with those files in cloud medic system to solve instances failures)

guillaumejacquart:
> this doesn't work for rolling updates/in multi main or with workers right? not your change but ingesting the logs from a local file in general

Yes you're right, this feature is built upon single node scope with persistent disk. each instance would try and recover. I think it's fine though (although probably improvable) as instance will replay the failed events that they initiated (so a main instance would reply a workflow execution failure, which is expected)

PR #28595, opened 2026-04-16
fix(HubSpot Trigger Node): Add missing property selectors

### LINKED ISSUE #27504: HubSpot Trigger Node - Ticket Property Changed
## Describe the problem/error/question
Hi team! When I add the HubSpot Trigger - "Ticket Property Changed" I am unable to get a list of the properties to use.

## What is the error message (if any)?
n/a

## Please share your workflow/screenshots/recording

Image

## Share the output returned by the last node
N/A this is a trigger node

### core

- n8nVersion: 2.13.2
- platform: docker (cloud)
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: 20

### storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

### pruning

- enabled: true
- maxAge: 720 hours
- maxCount: 25000 executions

### client

- isTouchDevice: false

Generated at: 2026-03-24T13:25:15.195Z

RomanDavydchuk:
## Summary

HubSpot Trigger node has 5 events of type `propertyChanged` and the user needs to specify which property should be monitored for changes. However, "Ticket Property Changed" and "Conversation Property Change" events didn't have the selector to do that. This PR adds them and refactors the code a bit to be more DRY

https://linear.app/n8n/issue/NODE-4709/community-issue-hubspot-trigger-node-ticket-property-changed
Closes #27504 

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Hubspot/HubspotTrigger.node.ts
```diff
@@ -251,53 +310,22 @@ export class HubspotTrigger implements INodeType {
...
-				return returnData;
+				return await getEntityProperties.call(this, '/properties/v2/deals/properties');
+			},
+			async getTicketProperties(this: ILoadOptionsFunctions): Promise<INodePropertyOptions[]> {
+				return await getEntityProperties.call(this, '/properties/v2/tickets/properties');
+			},
+			async getConversationProperties(
+				this: ILoadOptionsFunctions,
+			): Promise<INodePropertyOptions[]> {
+				return await getEntityProperties.call(this, '/properties/v2/conversations/properties');
```
BerniWittmann: As far as i understood the Hubspot Api docs, `conversations` are not one of the "objects" that can be used for this endpoint (https://developers.hubspot.com/docs/api-reference/legacy/crm/using-object-apis#object-type-id-values) 
DId you try running this, and got a valid response? then I'm probably not fully understanding the docs here.

packages/nodes-base/nodes/Hubspot/HubspotTrigger.node.ts
```diff
@@ -251,53 +310,22 @@ export class HubspotTrigger implements INodeType {
 
 	methods = {
 		loadOptions: {
-			// Get all the available contacts to display them to user so that they can
-			// select them easily
 			async getContactProperties(this: ILoadOptionsFunctions): Promise<INodePropertyOptions[]> {
-				const returnData: INodePropertyOptions[] = [];
```
BerniWittmann, line 257: Great job at removing all this repeated code 👍

packages/nodes-base/nodes/Hubspot/HubspotTrigger.node.ts
```diff
@@ -13,6 +13,29 @@ import { NodeConnectionTypes, NodeOperationError } from 'n8n-workflow';
...
 
+export async function getEntityProperties(
+	this: ILoadOptionsFunctions,
+	endpoint: string,
+): Promise<INodePropertyOptions[]> {
+	const returnData: INodePropertyOptions[] = [];
+	const properties = await hubspotApiRequest.call(this, 'GET', endpoint, {});
+
+	if (!Array.isArray(properties)) {
+		return returnData;
```
BerniWittmann: I'm thinking about whether we should let it error in this case instead to give better feedback to the user, instead of just showing an empty list. 
But I think you are probably the better judge on that, based on how similar situations are handled in other nodes :)

BerniWittmann (changes requested):

packages/nodes-base/nodes/Hubspot/HubspotTrigger.node.ts
RomanDavydchuk: Good call! The weird thing is that the API did return a valid response, but according to the docs there are only three valid properties for `conversations`:
image

https://developers.hubspot.com/docs/api-reference/legacy/conversations/guide#webhooks

I will update the code to use a predefined list of properties

RomanDavydchuk: Updated the options, verified that it's working, here's an example event I received during testing:
image

packages/nodes-base/nodes/Hubspot/HubspotTrigger.node.ts
RomanDavydchuk: Good idea, updated the code to throw, here's how it would look:
image

packages/nodes-base/nodes/Hubspot/HubspotTrigger.node.ts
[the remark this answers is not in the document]
RomanDavydchuk: Fixed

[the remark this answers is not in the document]
RomanDavydchuk: Fixed

packages/nodes-base/nodes/Hubspot/V1/GenericFunctions.ts
```diff
@@ -135,6 +135,8 @@ export const propertyEvents = [
 	'contact.propertyChange',
 	'company.propertyChange',
 	'deal.propertyChange',
+	'ticket.propertyChange',
+	'conversation.propertyChange',
```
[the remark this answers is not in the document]
RomanDavydchuk: Fixed

BerniWittmann (approved):
LGTM 🚀

PR #28631, opened 2026-04-17
fix(Jenkins Node): Properly load job parameters

### LINKED ISSUE #27991: Jenkins "Trigger with Parameters" operation fails to pre-load job parameters
### Bug Description

The Jenkins node is unable to retrieve parameters for a "Parameterized Build".

The dropdown "Name or ID" remains empty with a "No data" message despite the job being correctly configured in Jenkins.

Critical: Switching to "Expression" mode to manually enter the parameter name (e.g., DRUPAL_VERSION) does not work. The execution fails or the parameter is not passed to Jenkins, making it impossible to trigger parameterized jobs.

### To Reproduce

Define a job in Jenkins (e.g., drupal-10-dockerhub-harbor) with a String Parameter: DRUPAL_VERSION.

In n8n, configure a Jenkins node:

Resource: Job

Operation: Trigger with Parameters

Job Name or ID: Select the job above.

Observe the "Parameters" section:

Click "Add Parameter" -> "Name or ID" shows "No data".

Attempt Workaround:

Toggle "Name or ID" to Expression.

Type DRUPAL_VERSION manually.

Set a value.

Execute Step: The trigger fails or Jenkins receives a call without the parameters.

### Expected behavior

n8n should fetch available parameters via the Jenkins API (/job/JOBNAME/api/json) and populate the dropdown.

Manual entry via Expression should be correctly mapped to the API call payload.

### Debug Info

# Debug info

## core

- n8nVersion: 2.14.2
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: 0f0d8f19-bcb8-49c6-aea0-370507794062

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2026-04-02T20:38:48.302Z

### Operating System

Ubuntu 24.04 - Docker 27.3.1

### n8n Version

2.14.2

### Node.js Version

v24.13.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

RomanDavydchuk:
## Summary

Jenkins can expose build parameters in different places depending on job type and Jenkins/plugin version, so reading only one field is incomplete.
- `actions` is common for older/freestyle-style jobs
- `property` is common for Pipeline and newer configurations
- Both can represent the same `ParametersDefinitionProperty`, just in different parts of the job JSON
- If we read only actions, many valid parameterized jobs return “No data” even though parameters exist
- Reading both makes the node compatible across real-world Jenkins setups and avoids false negatives in the UI
- Deduplication is needed because `property` may include parameters that are also returned in `actions`

#### Manual
To manually test:
1. Run Jenkins, for example using Docker: `docker run -p 8080:8080 -p 50000:50000 jenkins/jenkins:lts`
2. Create a pipeline job
3. Go to "Configure" section of the job and check "This project is parameterized"
4. Add any parameters you like for testing
5. Go to n8n, add Jenkins node
6. Select "Trigger with Parameters" operation, select the pipeline job you created
7. Try adding the parameter. Check if parameter name is loaded or not (you should see the parameters you created in Jenkins)

#### Before the fix (no parameters are loaded for the pipeline job):
https://github.com/user-attachments/assets/d6922282-ea53-4d60-a37d-bed6f4f345e2

#### After the fix (parameters are loaded for both pipeline and freestyle jobs):
https://github.com/user-attachments/assets/7ec83651-3c68-436b-a8b7-ba4b052a915f

https://linear.app/n8n/issue/NODE-4823/community-issue-jenkins-trigger-with-parameters-operation-fails-to-pre
Closes #27991

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/nodes/Jenkins/Jenkins.node.ts
```diff
@@ -452,15 +452,22 @@ export class Jenkins implements INodeType {
...
+				const endpoint = `/job/${job}/api/json?tree=actions[parameterDefinitions[*]],property[parameterDefinitions[*]]`;
+				const result = await jenkinsApiRequest.call(this, 'GET', endpoint);
+				const allParameters = [...(result.actions ?? []), ...(result.property ?? [])];
+				const seenParameterNames = new Set<string>();
+				for (const { _class, parameterDefinitions } of allParameters) {
+					if (!_class?.includes('ParametersDefinitionProperty')) {
+						continue;
+					}
+
+					for (const { name, type } of parameterDefinitions) {
```
yehorkardash, line 467: not sure if Jenkins API may allow having null here, but maybe worth adding guard against null/undefined here

packages/nodes-base/nodes/Jenkins/Jenkins.node.ts
```diff
@@ -452,15 +452,22 @@ export class Jenkins implements INodeType {
...
 				const job = this.getCurrentNodeParameter('job') as string;
 				const returnData: INodePropertyOptions[] = [];
-				const endpoint = `/job/${job}/api/json?tree=actions[parameterDefinitions[*]]`;
-				const { actions } = await jenkinsApiRequest.call(this, 'GET', endpoint);
-				for (const { _class, parameterDefinitions } of actions) {
-					if (_class?.includes('ParametersDefinitionProperty')) {
-						for (const { name, type } of parameterDefinitions) {
+				const endpoint = `/job/${job}/api/json?tree=actions[parameterDefinitions[*]],property[parameterDefinitions[*]]`;
+				const result = await jenkinsApiRequest.call(this, 'GET', endpoint);
+				const allParameters = [...(result.actions ?? []), ...(result.property ?? [])];
```
yehorkardash, line 457: Can Jenkins return both `result.actions` and `result.property` leading to duplicate parameters?

yehorkardash (commented):
Looks good to me, just one clarification question

packages/nodes-base/nodes/Jenkins/Jenkins.node.ts
RomanDavydchuk: Yes, here's one of the API responses from actual Jenkins service, as you can see both `property` and `action` have a parameter `str_freestyle`
image

packages/nodes-base/nodes/Jenkins/Jenkins.node.ts
RomanDavydchuk: Added `Array.isArray` check for `parameterDefinitions` which covers nullish values as well

yehorkardash (approved):

PR #28636, opened 2026-04-17
fix(nodes-base): Bundle xlsx to fix npx install resolution failure

### LINKED ISSUE #28617: `npx n8n` fails with ETARGET for xlsx@0.20.2 while npm install n8n works
### Bug Description

Running `npx n8n` fails with `ETARGET` for `xlsx@0.20.2`, even though direct installs succeed.

Direct installs that work:

- `npm install n8n@2.16.1`
- `npm install n8n-nodes-base@2.16.0`
- `npm install https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz`

Published package metadata also looks correct:

- `n8n@2.16.1` depends on `n8n-nodes-base@2.16.0`
- published `n8n-nodes-base@2.16.0` contains:
  `"xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz"`

So this appears to be specific to the `npx` / `npm exec` path.

### To Reproduce

1. Use Node.js `v24.14.1`
2. Use npm `11.11.0`
3. Run `npx n8n`
4. Confirm installation when prompted

The command fails with:

```text
npm ERR! code ETARGET
npm ERR! notarget No matching version found for xlsx@0.20.2.

### Expected behavior

`npx n8n` should install and run successfully.

### Debug Info

From the npm debug log, the failure goes through registry resolution during `libnpmexec`:

```text
verbose stack xlsx: No matching version found for xlsx@0.20.2.
verbose stack     at module.exports (.../npm-pick-manifest/lib/index.js:210:23)
verbose stack     at RegistryFetcher.manifest (.../pacote/lib/registry.js:130:54)
verbose stack     at async RegistryFetcher.resolve (.../pacote/lib/registry.js:55:5)
verbose stack     at async #extractOrLink (.../@npmcli/arborist/lib/arborist/reify.js:722:7)
verbose stack     at async Arborist.reify (.../@npmcli/arborist/lib/arborist/reify.js:121:5)
verbose stack     at async .../libnpmexec/lib/with-lock.js:58:19
```

### Operating System

macOS 26.4.1

### n8n Version

2.16.1

### Node.js Version

24.14.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

ssam18:
Running `npx n8n` fails with `ETARGET No matching version found for xlsx@0.20.2` on npm 11 because `libnpmexec` resolves URL-based dependencies differently from a regular `npm install`, causing it to look up the SheetJS CDN package in the npm registry where it doesn't exist. Adding `xlsx` to `bundledDependencies` embeds it directly in the published tarball, so npm never needs to resolve it externally. Fixes #28617


michael-radency:
Thanks for the fix @ssam18!

However, we can't merge this as-is. Bundling xlsx directly into the published tarball changes our distribution relationship with SheetJS: while we already declare it as a dependency, embedding it inside our tarball makes us a distributor of the package. SheetJS uses PolyForm Noncommercial 1.0, which has specific terms around redistribution that require legal review before we can proceed.

Closing for now — thanks for digging into this!

PR #28646, opened 2026-04-18
feat(Schedule Node): Emit deduplication key when enabled (no-changelog)

mfsiega:
## Summary
The Schedule node emits a dedupe key when the feature flag is enabled.

For now, it just gets ignored. The next PR will pick it up and use it.

NOTE: there's no real e2e testing in this PR, that will come in the next change.

Three changes:
- `registerCron` accepts an `onTick` callback. This callback now supports a `scheduledTime` param which the `cron` library generates to indicate the canonical time when the cron was scheduled. The closure keeps track of the next scheduled time and updates it on each tick.
- The Schedule node accepts this `scheduledTime` and uses it to build the deduplication key. This then gets emitted, so when the trigger fires, it carries the deduplication key.
- Interval-based Schedules ("every 10 minutes") get deterministic jitter based on a hash of the workflowId+nodeId, rather than random when we convert to cron. For example: instead of picking a random value 0-59 for seconds, we pick it based on the hash. This means multiple instances of the same trigger have the same jitter.

The next PR will consume the emitted trigger from the workflow runner and actually apply deduplication on the database level.

Design doc and full plan: https://www.notion.so/n8n/3455b6e0c94f81a4afccfa6efadaa637

**Already merged**: feature flag and database migration for dedupe key column.
**Next up**: use the dedupe key to prevent duplicate executions.

## Related Linear tickets

https://linear.app/n8n/issue/CAT-2677

## Review / Merge checklist

- [ ] Docs updated or follow-up ticket created. _(N/A — inert producer until PR 4.)_
- [x] Tests included.


packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -441,7 +443,18 @@ export class ScheduleTrigger implements INodeType {
 			}
 		}
 
-		const executeTrigger = (recurrence: IRecurrenceRule, skipRecurrenceCheck = false) => {
+		const workflowId = this.getWorkflow().id;
+		const nodeId = this.getNode().id;
+
+		const configDedupEnabled =
+			Container.get(ExecutionsConfig).scheduledExecutionDeduplicationEnabled;
+		const dedupEnabled = configDedupEnabled && Boolean(workflowId);
```
mfsiega, line 453: I don't understand this line. Why is the truthiness of the workflowId involved? When would the workflowId ever be false anyway?

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -462,7 +475,17 @@ export class ScheduleTrigger implements INodeType {
 				Timezone: `${timezone} (UTC${momentTz.format('Z')})`,
 			};
 
-			this.emit([this.helpers.returnJsonArray([resultData])]);
+			const deduplicationKey =
+				dedupEnabled && scheduledT
+					? `${workflowId}:${nodeId}:${scheduledT.toISOString()}`
+					: undefined;
```
mfsiega, line 483: Should it be undefined, or null? I thought we were writing null to the database. Does it make a difference at this level?

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -462,7 +475,17 @@ export class ScheduleTrigger implements INodeType {
...
+
+			this.emit(
+				[this.helpers.returnJsonArray([resultData])],
+				undefined,
```
mfsiega: It would be helpful to annotate these args with the `/* paramName= */` pattern, so we know what we're setting undefined.

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -478,7 +501,9 @@ export class ScheduleTrigger implements INodeType {
 						expression: cronExpression,
 						recurrence,
 					};
-					this.helpers.registerCron(cron, () => executeTrigger(recurrence));
+					this.helpers.registerCron(cron, (scheduledT) =>
+						executeTrigger(recurrence, false, scheduledT),
```
mfsiega: Use the `/* paramName= */` pattern here as well, so we know what we're setting false.

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -47,7 +47,7 @@ export class ScheduledTaskManager {
 		}, activeInterval * Time.minutes.toMilliseconds);
 	}
 
-	registerCron(ctx: CronContext, onTick: () => void) {
+	registerCron(ctx: CronContext, onTick: (scheduledT: Date) => void) {
```
mfsiega: We should add some documentation explaining what scheduledT is, and what it's for. I'm not sure where exactly those docs belong. Also feel free to give it a more verbose, descriptive name.

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -72,10 +72,15 @@ export class ScheduledTaskManager {
...
-		const job = new CronJob(
+		let scheduledT: Date | null = null;
+		const job: CronJob = new CronJob(
 			expression,
 			() => {
+				const firedFor = scheduledT;
+				scheduledT = computeNext();
+
 				if (!this.instanceSettings.isLeader) return;
+				if (firedFor === null) return;
```
mfsiega: I don't understand what's going on here. Why the early return?

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -72,10 +72,15 @@ export class ScheduledTaskManager {
 			return;
 		}
 
-		const job = new CronJob(
+		let scheduledT: Date | null = null;
+		const job: CronJob = new CronJob(
```
mfsiega: This has gotten a bit harder to read because it's so long. Maybe breaking something into a descriptively-named helper function would be useful?

packages/nodes-base/nodes/Schedule/test/ScheduleTrigger.node.test.ts
```diff
@@ -170,5 +172,108 @@ describe('ScheduleTrigger', () => {
...
+				);
+				// The ISO timestamp segment must match the cron-scheduled fire time exactly.
+				const iso = (fourthArg as string).slice(`${workflowId}:${nodeId}:`.length);
+				const scheduledT = new Date(iso);
+				expect(scheduledT.getUTCMinutes()).toBe(0);
+				expect(scheduledT.getUTCSeconds()).toBe(0);
+				expect(scheduledT.getUTCMilliseconds()).toBe(0);
+			});
+
+			it('should not emit a deduplication key when the workflow id is undefined', async () => {
```
mfsiega: Again, I don't totally understand why we're singling out this case?

packages/nodes-base/nodes/Schedule/test/ScheduleTrigger.node.test.ts
```diff
@@ -170,5 +172,108 @@ describe('ScheduleTrigger', () => {
...
+					workflow: { active: true },
+				});
+
+				jest.advanceTimersByTime(2 * HOUR);
+
+				expect(emit).toHaveBeenCalledTimes(1);
+				expect(emit.mock.calls[0][3]).toBeUndefined();
+			});
+
+			it('should not emit a deduplication key when the feature flag is disabled', async () => {
```
mfsiega: I don't actually think we need this test - since emitting the key is harmless, it's not so important that we verify it, I think.

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
mfsiega: Kept the guard but added a short comment: workflowId should always be defined in scheduled mode, but if it isn't we skip emitting the key.

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
mfsiega: `undefined` matches `ITriggerFunctions.emit`'s param type (`deduplicationKey?: string`). The runner translates absence to `null` for the DB column on insert.

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
mfsiega: Annotated as `/* responsePromise= */ undefined, /* donePromise= */ undefined`.

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
mfsiega: Annotated as `/* skipRecurrenceCheck= */ false`.

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Renamed to `scheduledTime` and added a JSDoc on `onTick` explaining it's the canonical scheduled fire time (computed via `nextDate()` rather than `now`, so it's drift-free and stable across instances).

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Added an inline comment: `firedFor === null` would mean `computeNext()` swallowed an error at registration. Unreachable in practice (the cron expression has already been validated by the constructor), but the guard preserves the invariant: never invoke `onTick` without a canonical scheduled time.

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Extracted the cron callback into a named `handleTick` helper.

packages/nodes-base/nodes/Schedule/test/ScheduleTrigger.node.test.ts
mfsiega: Removed.

packages/nodes-base/nodes/Schedule/test/ScheduleTrigger.node.test.ts
mfsiega: Removed.

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -47,7 +47,13 @@ export class ScheduledTaskManager {
 		}, activeInterval * Time.minutes.toMilliseconds);
 	}
 
-	registerCron(ctx: CronContext, onTick: () => void) {
+	/**
+	 * @param onTick - Called on each cron firing on the leader instance.
+	 *   Receives `scheduledTime`: the canonical scheduled fire time for this
+	 *   tick, pre-computed from the cron expression rather than read from
+	 *   `now`, so it is drift-free and stable across instances.
```
mfsiega:
```suggestion
	 *   Receives `scheduledTime`: the canonical scheduled fire time for this
	 *   tick.
```

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -72,24 +75,32 @@ export class ScheduledTaskManager {
...
-				onTick();
-			},
-			undefined,
-			true,
-			timezone,
-		);
+			onTick(firedFor);
+		};
+
+		const job: CronJob = new CronJob(expression, handleTick, undefined, true, timezone);
```
mfsiega: Let's document what "undefined" and "true" mean with the `/* paramName= */ pattern.

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -72,24 +75,32 @@ export class ScheduledTaskManager {
...
+		const computeNext = (): Date => cronTime.sendAt().toJSDate();
+		let scheduledTime: Date = computeNext();
+
+		const handleTick = () => {
+			// Capture the time this firing was scheduled for, then advance
+			// `scheduledTime` to the next upcoming fire.
+			const firedFor = scheduledTime;
+			scheduledTime = computeNext();
+
+			if (!this.instanceSettings.isLeader) return;
```
mfsiega: Does it make sense to advance firedFor if we aren't the leader? Seems like we should return immediately if we aren't the leader, and not do anything else?

packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
```diff
@@ -28,97 +27,139 @@ function mockMomentTz(values: {
...
-		});
-		expect(result).toEqual('30 */30 * * * *');
+		const result = toCronExpression(
+			{
+				field: 'minutes',
+				minutesInterval: 30,
+			},
+			TEST_SEED,
+		);
+		expect(result).toEqual('56 */30 * * * *');
```
mfsiega, line 73: For reference - it seems like previously we were overriding randomInt such that it was returning 30. Now we are using the value that happens to be produced by our stable hash.

It might be clearer if we add that comment to each test - the seconds/minutes/etc are generated by our stable hash, while the minutes/hours/etc are based on the input interval.

packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
```diff
@@ -28,97 +27,139 @@ function mockMomentTz(values: {
...
+			{
+				field: 'months',
+				monthsInterval: 3,
+			},
+			TEST_SEED,
+		);
+		expect(result1).toEqual('56 19 14 4 */3 *');
+	});
+
+	it('should be deterministic for the same seed and produce different output for different seeds', () => {
```
mfsiega: Why do we have this test?

packages/nodes-base/nodes/Schedule/GenericFunctions.ts
```diff
@@ -83,28 +84,41 @@ export function recurrenceCheck(
...
+/**
+ * Deterministic integer in `[min, max)`, derived from `seed` and `label`.
+ *
+ * Used to fill in the otherwise-unspecified parts of a cron expression
+ * (e.g. the second within a minute) in a way that's stable across
+ * instances. Two mains computing the same schedule for the same node
+ * therefore produce identical cron expressions, identical fire times,
+ * and identical deduplication keys.
+ */
+const stableInt = (seed: string, label: string, min: number, max: number): number => {
```
mfsiega, line 105: I think it's worth explaining why we have the seed and label params here, since it's not clear why we need them based on the doc here.

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Already applied in `70ecf7dc34` ("Apply suggestion from @mfsiega").

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Annotated as `/* onComplete= */ undefined, /* start= */ true`.

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Good call — moved the `isLeader` check to the top of `handleTick`. Safe in practice because leadership transitions deregister crons (`job.stop()`), so a follower never holds a live `CronJob` long enough for stale `scheduledTime` to matter.

packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
mfsiega: Added per-test comments calling out which positions of each cron expression are filled by the stable hash vs taken from the input interval.

packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
mfsiega: Removed. The other tests already lock in determinism by asserting specific outputs for specific seeds; the second half of the assertion was essentially testing the hash function itself.

packages/nodes-base/nodes/Schedule/GenericFunctions.ts
mfsiega: Added `@param` docs for `seed` (stable identity of the entity, e.g. `${workflowId}:${nodeId}`) and `label` (per-field disambiguator so the second and minute don't collide).

packages/nodes-base/nodes/Schedule/ScheduleTrigger.node.ts
```diff
@@ -441,7 +443,20 @@ export class ScheduleTrigger implements INodeType {
...
+		const configDedupEnabled =
+			Container.get(ExecutionsConfig).scheduledExecutionDeduplicationEnabled;
+		// The workflowId should always be defined, but if it isn't we skip
+		// the deduplication key.
+		const dedupEnabled = configDedupEnabled && Boolean(workflowId);
+
+		const executeTrigger = (
+			recurrence: IRecurrenceRule,
+			skipRecurrenceCheck = false,
+			scheduledTime?: Date,
```
tomi, line 458: When can this be undefined?

packages/nodes-base/nodes/Schedule/test/GenericFunctions.test.ts
```diff
@@ -28,97 +27,146 @@ function mockMomentTz(values: {
...
-		value: (min: number, max: number) => Math.floor((min + max) / 2),
-	});
+// Cron expressions are 6 fields: `<sec> <min> <hr> <dom> <mon> <dow>`.
+// Fields the user does not explicitly pin (e.g. `triggerAtMinute`) are
+// filled in by `stableInt(seed, label, ...)`; every other field comes
+// from the input interval.
+//
+// For `seed = 'test-key'`, `stableInt` produces:
+//   second=56, minute=19, hour=14, dayOfMonth=4
+const TEST_SEED = 'test-key';
```
tomi, line 37: Nice tests 👌

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -72,22 +75,36 @@ export class ScheduledTaskManager {
...
+			// `scheduledTime` to the next upcoming fire.
+			const firedFor = scheduledTime;
+			scheduledTime = computeNext();
+
+			this.logger.debug('Executing cron for workflow', {
+				workflowId,
+				nodeId,
+				cron: summary,
+				instanceRole: this.instanceSettings.instanceRole,
+			});
```
tomi, line 98: nit: I think we could include the firedFore here in the log as well

tomi (approved):
LGTM, only questions / nit comments

PR #28649, opened 2026-04-18
feat(core): Warn and skip on duplicate scheduled executions

mfsiega:
## Summary

Consume the deduplication key on execution creation to deduplicate executions.

Threads the key to the point where we create the execution, and handle the uniqueness violation error if it occurs.

Also adds integration tests. TODO: manually try to force a leader confusion to test that executions really get deduplicated.

## Related Linear tickets

https://linear.app/n8n/issue/CAT-2677

## Review / Merge checklist

- [ ] Docs updated or follow-up ticket created. _(Follow-up: add an operator-facing entry for the flag once we're ready to recommend turning it on.)_
- [x] Tests included.


azender1:
This looks like the same execution-boundary class as the IMAP duplicate trigger issue, just through the scheduler path.

If the same scheduled activation is picked up more than once (or overlaps due to concurrency), and there’s no single authoritative execution claim before launching the workflow, multiple runs can all proceed as valid.

At that point the system can’t distinguish between:
- already executing
- already completed
- needs execution

I reproduced this pattern outside n8n with a minimal model:

- same logical event delivered concurrently
- no global execution claim
→ duplicate side effects

Adding a simple “claim before execute” step using a stable event_id resulted in exactly-once behavior — only the first execution ran, others skipped based on prior claim.

This suggests the scheduler path may share the same missing boundary as trigger-based duplication: execution state isn’t authoritative across concurrent activations.

packages/cli/test/migration/1778000000000-add-execution-deduplication-key.test.ts
mfsiega, line 1: We decided a migration test is unnecessary. We can get rid of this, I think. I think we got rid of it in a previous PR before we rebased.

packages/core/src/execution-engine/scheduled-task-manager.ts
```diff
@@ -108,6 +108,22 @@ export class ScheduledTaskManager {
 			timezone,
 		);
 
+		const computeNext = (): Date | null => {
```
mfsiega: Does this still need to be here? I thought this changed with the previous PRs.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -364,20 +365,35 @@ export class ActiveWorkflowManager {
...
+								nodeId: node.id,
+								deduplicationKey: error.deduplicationKey,
+							});
+							return undefined;
+						}
+						throw error;
+					});
 
 				void executePromise.then((executionId) => {
+					if (executionId === undefined) return;
```
mfsiega, line 400: Is it true that executionId was previously never undefined? Or were we not handling it? Or is this redundant handling?

packages/cli/test/migration/1778000000000-add-execution-deduplication-key.test.ts
mfsiega: Removed in 15acd2d3ed.

packages/core/src/execution-engine/scheduled-task-manager.ts
mfsiega: Right, this was leftover dead code from before master switched to the `cronTime.sendAt()` approach (lines 78-83 above) — the `-X ours` rebase preserved master's version inline but appended the old block below it, leaving a duplicate `computeNext` declaration and an undefined `scheduledT` reference. Removed in 15acd2d3ed.

packages/cli/src/active-workflow-manager.ts
mfsiega: `workflowExecutionService.runWorkflow` returns `Promise<string>` (it returns `workflowRunner.run()` which is typed `Promise<string>`), so previously `executionId` here was always defined. The new `.catch` handler a few lines up returns `undefined` on `DuplicateExecutionError` — that's what makes `executePromise` resolve to `string | undefined` now. Without this guard we'd emit `workflow-executed` with `executionId: undefined` for skipped duplicates, which is wrong (no execution actually happened). Same reasoning applies to the `donePromise` branch below.

packages/cli/src/__tests__/active-workflow-manager.test.ts
```diff
@@ -369,6 +376,73 @@ describe('ActiveWorkflowManager', () => {
...
+					mode,
+					activation,
+				);
+				const context = getTriggerFunctions(workflow, node, additionalData, mode, activation);
+
+				context.emit(triggerData, undefined, undefined, 'wf-1:node-1:1700000000000');
+
+				await new Promise((resolve) => setTimeout(resolve, 0));
+
+				expect(scopedLogger.warn).toHaveBeenCalledTimes(1);
```
mfsiega: We don't need assertions on log statements.

packages/cli/src/errors/duplicate-execution.error.ts
mfsiega, line 1: Just to check, this this the right place? Or does it belong in `n8n-workflow`?

azender1:
Great to see this getting native treatment in n8n core. I posted in the community yesterday about this exact problem — built SafeAgent as a Python execution guard that handles this at the tool call boundary for AI agents. The deduplicationKey approach here is the right pattern. Happy to share notes on edge cases we hit, particularly around SQLite unique violation detection across restarts.

azender1:
Also worth noting — the SQLite unique constraint works well for 
single-node but needs careful transaction isolation on Postgres 
under concurrent retries. Happy to share what we found if useful 
for the distributed case.

packages/cli/src/executions/execution-persistence.ts
```diff
@@ -43,26 +45,33 @@ export class ExecutionPersistence {
...
+				}
+
+				await this.fsStore.write(
+					{ workflowId: id, executionId },
+					{ data, workflowData: workflowSnapshot, workflowVersionId },
+				);
 				return executionId;
+			});
+		} catch (error) {
+			if (executionEntity.deduplicationKey && isUniqueViolationError(error)) {
```
mfsiega: I guess technically we could be violating some other uniqueness constraint, right? And I guess verifying that we passed in a deduplicationKey is a heuristic to be more confident that *this* is the uniqueness violation. Is there any way to know with even more confidence? For example, checking the error message?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -364,20 +365,35 @@ export class ActiveWorkflowManager {
...
+						node,
+						data,
+						additionalData,
+						mode,
+						responsePromise,
+						deduplicationKey,
+					)
+					.catch((error: unknown) => {
+						if (error instanceof DuplicateExecutionError) {
+							this.logger.warn('Scheduled execution skipped: duplicate deduplication key', {
```
mfsiega: Is there any other error reporting we should be doing here, based on the sort of error handling we see around this sort of thing elsewhere in the codebase?

packages/cli/src/active-workflow-manager.ts
mfsiega, line 400: We should explain when this happens.

packages/cli/src/active-workflow-manager.ts
```diff
@@ -388,6 +404,10 @@ export class ActiveWorkflowManager {
 
 				if (donePromise) {
 					void executePromise.then((executionId) => {
+						if (executionId === undefined) {
```
mfsiega, line 413: As above, there should be a short comment explaining when this would happen.

azender1: For what it's worth — in SafeAgent we put the execution guard 
at the outermost boundary, not inside the workflow package. 
The reasoning: if the guard lives inside the workflow, a crash 
before the guard runs still leaves you exposed. External = 
catches everything including mid-execution failures.

Happy to share the request_id + unique constraint approach 
we landed on if useful.

mfsiega: Hi @azender1 - thanks for your input, and for sharing your pointers! This code will go through the usual process of review by maintainers. If you do have specific concerns or suggestions, feel free to leave them.

packages/cli/src/__tests__/active-workflow-manager.test.ts
mfsiega: Done in 2d41ccd322 — kept the behavioral assertion (`eventService.emit` not called) and dropped the log assertions.

packages/cli/src/errors/duplicate-execution.error.ts
mfsiega: I think this is the right place. `packages/cli/src/errors` already houses sibling execution-lifecycle errors (`execution-not-found-error.ts`, `execution-already-resuming.error.ts`, `aborted-execution-retry.error.ts`, `deduplication.error.ts`), and `DuplicateExecutionError` is only thrown/caught by cli code (`ExecutionPersistence` and `ActiveWorkflowManager`). `n8n-workflow/src/errors` is for the base classes and generic workflow-level errors.

packages/cli/src/executions/execution-persistence.ts
mfsiega: Good point — the `deduplicationKey` heuristic is correct today (it's the only user-facing unique constraint on `execution_entity`), but it's fragile if someone adds another unique column later. Tightened in 2d41ccd322 to also check that the error message contains `deduplicationKey`. Both Postgres (`duplicate key value violates unique constraint ...`) and SQLite (`UNIQUE constraint failed: execution_entity.deduplicationKey`) include the column name in the message. Added a test case for the "unique violation on a different column" case to lock the new behavior in.

packages/cli/src/active-workflow-manager.ts
mfsiega: I don't think so. `errorReporter.error()` is used elsewhere in this file (lines 211, 564, 911, 981, 998) for *unexpected* errors — Sentry-grade noise. A duplicate scheduled execution is an *expected* outcome of the dedup mechanism (the whole point of this PR), so a `warn` log is the right level. Reporting to Sentry would create noise on every duplicate fire. Happy to revisit if there's a specific reporting pattern you had in mind.

packages/cli/src/active-workflow-manager.ts
mfsiega: Added an explanatory comment in 2d41ccd322.

packages/cli/src/active-workflow-manager.ts
mfsiega: Added in 2d41ccd322.

packages/cli/src/executions/execution-persistence.ts
```diff
@@ -43,26 +45,37 @@ export class ExecutionPersistence {
...
+
+				await this.fsStore.write(
+					{ workflowId: id, executionId },
+					{ data, workflowData: workflowSnapshot, workflowVersionId },
+				);
 				return executionId;
+			});
+		} catch (error) {
+			if (
+				executionEntity.deduplicationKey &&
```
tomi: I think this might be a bit overly complex. Maybe we could have a `isDuplicateExecutionError` and use that?

packages/cli/src/active-workflow-manager.ts
```diff
@@ -364,20 +365,39 @@ export class ActiveWorkflowManager {
...
+					)
+					.catch((error: unknown) => {
+						if (error instanceof DuplicateExecutionError) {
+							const context = {
+								workflowId: workflowData.id,
+								nodeId: node.id,
+								deduplicationKey: error.deduplicationKey,
+							};
+							this.logger.warn('Scheduled execution skipped: duplicate deduplication key', context);
+							this.errorReporter.warn(error, { extra: context });
```
tomi: Let's use `shouldBeLogged: false` here since it's already logged

packages/cli/src/active-workflow-manager.ts
```diff
@@ -364,20 +365,39 @@ export class ActiveWorkflowManager {
...
+							return undefined;
+						}
+						throw error;
+					});
 
 				void executePromise.then((executionId) => {
+					// `executionId` is undefined when the catch above swallowed a
```
tomi, line 398: I wonder if we should make the "dedup" branch a more distinctive from the normal flow?

packages/cli/test/integration/executions/execution-persistence.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	beforeEach(async () => {
+		await testDb.truncate(['ExecutionEntity']);
+	});
+
+	afterAll(async () => {
+		await testDb.terminate();
+	});
+
+	describe('create with deduplicationKey', () => {
```
tomi, line 25: Nice tests 👏

packages/cli/src/executions/execution-persistence.ts
mfsiega: Specifically - isDuplicateExecutionError can just be a helper function in execution-persistence.ts. We don't need a generic isUniqueViolationError which involves this QueryFailedError hence the n8n-local-rules/misplaced-n8n-typeorm-import violation.

packages/cli/src/active-workflow-manager.ts
mfsiega: I also want to be extra sure that executionId is never undefined before this change. I find that a bit surprising.

azender1:
The isDuplicateExecutionError helper approach tomi suggested 
is cleaner — same pattern we ended up using in SafeAgent. 
Keeps the uniqueness check self-contained and testable in 
isolation without pulling in QueryFailedError dependencies.

Two edge cases worth flagging from what we hit in SafeAgent:

1. SQLite unique violation detection changes behavior when 
extended result codes are disabled — the error code comes back 
as SQLITE_CONSTRAINT instead of SQLITE_CONSTRAINT_UNIQUE. 
Worth testing with extended_result_codes off.

2. On Postgres under high concurrency, two transactions can 
both pass the deduplicationKey check before either commits — 
the unique constraint catches it, but only if the isolation 
level is at least READ COMMITTED with the insert happening 
inside the same transaction as the check. If the key is 
checked then inserted in separate statements, there's a 
window.

Both are edge cases, not blockers. Happy to share the test 
cases we used to reproduce them if useful.

packages/cli/src/executions/execution-persistence.ts
mfsiega: Done.

packages/cli/src/active-workflow-manager.ts
mfsiega: Done.

packages/cli/src/active-workflow-manager.ts
mfsiega: I couldn't find a way to make this feel much cleaner or simpler, so leaving as-is for now.

We asked whether there was any other case where we could get an undefined executionId here. `WorkflowExecutionService.runWorkflow` returns `Promise<string>` so it seems we should only get undefined when it's being generated by the `catch` above. Not sure if we want to do anything to make that more readable (or leave it as a follow-up).

One thing to call out - we invoke the donePromise with the undefined execution id. Schedule trigger doesn't use donePromise so it's not a problem right now, but in the future nodes emitting a dedupe key *and* using donePromise will have to handle undefined executionId. Not sure if that's already the case or not.

tomi (approved):
Code LGTM 👌 Will test manually a bit next

tomi:
Works nicely. Deduplicates duplicate schedule times 👌

PR #28655, opened 2026-04-19
feat(core): Implement propertiesOrOptionsChanged detection in isDirty via config hash snapshotting

imnick48:
## Summary

`isDirty` in `find-start-nodes.ts` has a hardcoded `const propertiesOrOptionsChanged = false` placeholder. Dirty-detection for parameter changes currently works only because the **frontend** (`useNodeDirtiness.ts`) sends `dirtyNodeNames` to the backend, which pre-cleans run data before `isDirty` is ever called. This means API-triggered partial executions without a frontend can silently reuse stale output when a node's configuration has changed.

**Root cause:** The backend has no way to know whether a node's config changed since its last execution — it has no record of what the config looked like at execution time.

**Fix:** When a node executes, compute a deterministic SHA-256 hash of its execution-relevant configuration (`type`, `typeVersion`, `parameters`, `credentials`, `disabled`, `executeOnce`, `retryOnFail`, `maxTries`, `waitBetweenTries`, `alwaysOutputData`, `onError`, `continueOnFail`) and store it in a new optional `configHash` field on `ITaskData`. On subsequent partial executions, `isDirty` compares the current node's hash against the stored one to detect configuration drift.

Fields excluded from hashing: `id`, `name`, `position`, `notes`, `notesInFlow`, `webhookId` — these are cosmetic and do not affect execution behaviour.

**Backwards compatibility:** If `configHash` is absent on stored run data (pre-existing executions), the check is skipped entirely and falls through to the existing dirty-detection logic. No behaviour change for old data.

**Scope:** This PR only implements `propertiesOrOptionsChanged`. The `parentNodeGotDisabled` check remains a TODO — it requires passing `DirectedGraph` into `isDirty`, which is a larger refactor and will be addressed in a follow-up.

**To test:** Run a partial execution via the API (no frontend), change a node parameter, run again — the node should now re-execute rather than reuse cached output.

### Changed files

| Package | File | Change |
|---|---|---|
| `n8n-workflow` | `interfaces.ts` | Add optional `configHash?: string` to `ITaskData` |
| `@n8n/core` | `compute-node-config-hash.ts` | New utility — deterministic SHA-256 of execution-relevant node config |
| `@n8n/core` | `find-start-nodes.ts` | Replace hardcoded `false` with actual hash comparison |
| `@n8n/core` | `workflow-execute.ts` | Write `configHash` into `taskData` after node execution |
| `@n8n/core` | `compute-node-config-hash.test.ts` | New unit tests for the hash utility |
| `@n8n/core` | `find-start-nodes.test.ts` | New test cases for the `isDirty` hash comparison path |

None — this addresses a known internal TODO (`// TODO: implement` in `find-start-nodes.ts`).

## Review / Merge checklist
- [x] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created. — Not applicable, no user-facing behaviour change.
- [x] Tests included. — Unit tests for `computeNodeConfigHash` and new `isDirty` cases covering hash match (not dirty), hash mismatch (dirty), and missing `configHash` (backwards compat).


chuenchen309:
This has sat since April with no human review, so here's an independent check of the premise. The gap it describes is real and still open on `master` today, and I think the reason it's stalled is a design question the description doesn't quite surface — worth stating explicitly so a maintainer can rule on it quickly.

**The gap, verified end to end on current `master`:**

1. `packages/@n8n/api-types/src/dto/workflows/manual-run.dto.ts:45` — `dirtyNodeNames: z.array(z.string()).optional()`. It is optional on the public manual-run payload.
2. `packages/cli/src/manual-execution.service.ts:195` — passed straight through to `workflowExecute.runPartialWorkflow2(...)`.
3. `packages/core/src/execution-engine/workflow-execute.ts:211` — `dirtyNodeNames: string[] = []`, and at 285–286 that array is the only thing driving the dirty-node cleanup:
   ```ts
   const dirtyNodes = graph.getNodesByNames(dirtyNodeNames);
   runData = cleanRunData(runData, graph, dirtyNodes);
   ```
4. `packages/core/src/execution-engine/partial-execution-utils/find-start-nodes.ts:14-15` — still:
   ```ts
   // TODO: implement
   const propertiesOrOptionsChanged = false;
   ```

So a caller that omits `dirtyNodeNames` gets an empty array, `cleanRunData` removes nothing on that basis, and the backend's own dirtiness check is a hardcoded `false`. There is no server-side fallback: parameter-change detection today is entirely whatever the editor chose to send. For an API-driven partial execution that's a silent stale-output reuse, exactly as described.

**The decision this is really asking for.** The fix adds `configHash` to `ITaskData`, which is part of stored run data rather than a purely in-memory structure — so it changes what gets written into every execution record from here on, and the hash's field list (`type`, `typeVersion`, `parameters`, `credentials`, `disabled`, `executeOnce`, `retryOnFail`, `maxTries`, `waitBetweenTries`, `alwaysOutputData`, `onError`, `continueOnFail`) becomes a de-facto contract: change the list later and every stored hash silently mismatches, marking everything dirty once. The backwards-compatibility story here is good — absent `configHash` skips the check — but it only covers *old* data, not a future revision of the hash input. That's a maintainer call about storage and versioning, not something a reviewer can settle, and I suspect it's why this hasn't moved.

If the answer is "yes but don't persist it", the same detection is achievable by hashing the config the frontend already knows about and keeping it out of `ITaskData` — worth asking before more work goes in either direction.

Two smaller notes: the `n8n-assistant` checklist asked for tests and this does carry them (`compute-node-config-hash.test.ts` plus new `find-start-nodes.test.ts` cases), so that item looks satisfied; and since the branch is ~3.5 months behind, it'll want a rebase — though the specific TODO it targets is untouched on `master`, so the change itself hasn't gone stale.

*AI-assisted review; the four call sites above were read on current `master` rather than inferred. I have no stake in this PR.*

chuenchen309:
@imnick48 — worth being clear about what closed this, because it wasn't the design question I raised above and it wasn't the code.

The bot closed it on path alone: `packages/core` is the workflow execution engine, and per the contributing guidelines changes there aren't taken as community PRs. That rule catches every file this touches except `packages/workflow/src/interfaces.ts`, so no amount of reworking the hash or rebasing would have changed the outcome. Four months of it sitting in review queue before that is its own thing, but the close itself is a routing rule, not a verdict on the work.

The same message points at the way through, and it's worth taking literally:

> If you've hit a bug in one of these areas, please open an issue describing it and we'll pick it up from there — that's genuinely useful to us.

The gap is still open on `master` as of today — `packages/core/src/execution-engine/partial-execution-utils/find-start-nodes.ts` still reads:

```ts
export function isDirty(node: INode, runData: IRunData = {}, pinData: IPinData = {}): boolean {
	// TODO: implement
	const propertiesOrOptionsChanged = false;
```

so an issue describing the user-visible symptom would land on something real and unclaimed. The four-step chain in my earlier comment is the backend half of that argument if it's useful to you — the part that matters for a bug report is the observable end: a partial execution driven through the API without `dirtyNodeNames` reuses stale output after a node's parameters change, because the server has no fallback detection of its own.

I'd file it myself, but the bug template requires debug info, n8n version and hosting details from a running instance, and I don't have one — I came at this from the code. You do, so the report will be a lot more concrete coming from you.

*AI-assisted; the guideline text, the bot's close reason and the `master` source above were read today rather than recalled.*

PR #28666, opened 2026-04-20
refactor(ai-builder): Unify sub-agent handoff under one typed contract

aalises:
## Summary

Collapse the scattered briefing/debriefing scaffolding that Instance AI's six sub-agent spawn sites each reimplemented into a single typed handoff contract at `packages/@n8n/instance-ai/src/agent/handoff.ts`.

- Per-kind Zod input schemas + discriminated `SubAgentHandoff` union
- One `renderHandoff()` XML renderer; one `observeOutcome()` observer
- `PlannedTask.spec: string` → `PlannedTask.handoff: PlannedHandoff` (typed per kind)
- `PlannedTaskRecord.outcome: Record<string, unknown>` → typed `SubAgentOutcome`
- `toPlannedTaskArg()` projects the wire-format; SSE payloads stay byte-identical

Inspired by OpenAI Agents SDK `handoff({ inputType })` + ambient `RunContext`, adapted to Mastra's `Agent.stream(string)`.

image

## What changes for the model

Four of six spawn paths render byte-identical XML. The two outliers (`plan-with-agent`, `browser-credential-setup`) previously built ad-hoc markdown and now go through `renderHandoff` — wrapped in `<task>…</task>` and receive `<thread-state>` when siblings are running.

## What doesn't change

No wire-format, SSE payload, or `@n8n/api-types` changes. Internal refactor only.

## Why

Forces exhaustive dispatch at compile time, kills `outcome as WorkflowBuildOutcome` casts, and gives future contributors one obvious pattern instead of three.


packages/@n8n/instance-ai/src/tools/orchestration/delegate.tool.ts
```diff
@@ -145,13 +156,16 @@ export async function startDetachedDelegateTask(
...
-	const briefingMessage = await buildDelegateBriefing(
-		context,
+	const handoff = buildDelegateHandoff({
 		role,
-		input.spec,
-		input.artifacts,
-		input.conversationContext,
-	);
+		instructions:
+			'Complete the delegated task using the provided tools. Return concrete results only.',
```
mutdmour: what is "concrete" results? Maybe examples would help here. Not sure what this is solving. Or problems it's creating.

packages/@n8n/instance-ai/src/tools/orchestration/blueprint-accumulator.ts
```diff
@@ -119,31 +105,47 @@ function workflowItemToTask(
...
-		kind: 'delegate',
-		spec: di.description,
 		deps: di.dependsOn,
 		tools: di.requiredTools,
+		handoff: makeHandoff(di.id, {
+			kind: 'delegate',
+			input: {
+				role: di.title,
+				instructions:
+					'Complete the delegated task using the provided tools. Return concrete results only.',
```
mutdmour: same question here

packages/@n8n/instance-ai/src/agent/handoff.ts
```diff
@@ -0,0 +1,411 @@
...
+			if (formatted) parts.push(formatted);
+		} catch {
+			// Best effort — iteration log is never load-bearing for the briefing.
+		}
+	}
+
+	return parts.join('\n\n');
+}
+
+function taskBlockFor(h: SubAgentHandoff): string {
```
mutdmour: this refactor seems to fall short here. instead of a generic handoff, the prompts are now split from their related tools. would be be better to leave those in their tools, or refactor so this is part of the tool

packages/@n8n/instance-ai/src/agent/handoff.ts
```diff
@@ -0,0 +1,411 @@
...
+		case 'manage-data-tables':
+			return h.input.goal;
+		case 'planner':
+			return renderPlannerTask(h.input);
+		case 'browser-credential-setup':
+			return renderBrowserCredTask(h.input);
+	}
+}
+
+function conversationContextFor(h: SubAgentHandoff): string | undefined {
```
mutdmour: same issue here

packages/@n8n/instance-ai/src/tools/orchestration/delegate.tool.ts
aalises: modified the prompt to be clearer on what's expected for a handoff

packages/@n8n/instance-ai/src/agent/handoff.ts
aalises: I did this on purpose. Keeping dispatch in handoff.ts means TypeScript forces me to handle every kind, so adding a new one and the compiler breaks until I do. Moving renderers back into the tools loses that check but not a biggie.                                                            

re. the prompts, they aren't really tool logic, as they turn the handoff into the message the child reads (it's more of an application). Putting them next to the schema they format felt alright, but happy to move them if you still think it's better that way.

mutdmour: Can we require passing a callback to the handoff? `buildTaskBlock: (handoff: SubAgentHandoff) => string` for every tool. 
Yes, not a fan of mixing the prompts here since they are unrelated, out of context and makes it harder to trace/understand wha the tool is doing.

aalises: done

aalises:
@mutasem FYI, added a credentials snapshot (`availableCredentials` + `credentialsSnapshotAt`) to the builder handoff so sub-agent builders can skip their redundant `credentials(action="list")` call; captured once per dispatch, rendered into the briefing's `<artifacts>` block, and both builder prompts now treat it as authoritative.

Let's see how it works

packages/@n8n/instance-ai/src/storage/planned-task-storage.ts
```diff
@@ -35,56 +36,170 @@ const plannedHandoffSchema = z.discriminatedUnion('kind', [
...
 	messageGroupId: z.string().optional(),
-	status: z.enum(['active', 'awaiting_replan', 'completed', 'cancelled']),
-	tasks: z.array(plannedTaskRecordSchema),
+	status: z.enum(['awaiting_approval', 'active', 'awaiting_replan', 'completed', 'cancelled']),
+	tasks: z.array(legacyPlannedTaskRecordSchema),
 });
 
+type LegacyPlannedTaskRecord = z.infer<typeof legacyPlannedTaskRecordSchema>;
+
+function normalizeLegacyTask(task: LegacyPlannedTaskRecord): unknown {
```
mutdmour: why do we need to maintain this legacy logic?

mutdmour (commented):
Can you please add more tests? I see a lot of changes tools with no test changes

aalises:
tests added and legacy removed @mutdmour

packages/@n8n/instance-ai/src/storage/planned-task-storage.ts
aalises: true, will remove it

PR #28675, opened 2026-04-20
fix(Simple Memory Node): Scope memory nodes session id to single memory node (no-changelog)

lucamattiazzi:
## Summary

This fix changes the `session_id` value for all memory nodes when the value is inherited from previous nodes to avoid multiple memory nodes in the same workflow to read/write from the same memory.

https://linear.app/n8n/issue/AI-2377

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/nodes-langchain/utils/helpers.ts
```diff
@@ -49,6 +49,33 @@ export function getPromptInputByType(options: {
...
+	'@n8n/n8n-nodes-langchain.memoryMotorhead': 1.4,
+	'@n8n/n8n-nodes-langchain.memoryXata': 1.5,
+	'@n8n/n8n-nodes-langchain.memoryZep': 1.4,
+};
+
+function shouldScopeSessionKey(ctx: ISupplyDataFunctions | IWebhookFunctions): boolean {
+	const node = ctx.getNode();
+	if (!node) return false;
+	const minVersion = SESSION_KEY_SCOPING_MIN_VERSION[node.type];
+	if (minVersion === undefined) return false;
```
yehorkardash: Does that mean that scoping won't work automatically for the new memory nodes that are not in this list?

packages/@n8n/nodes-langchain/utils/helpers.ts
```diff
@@ -101,6 +128,10 @@ export function getSessionId(
 		}
 	}
 
+	if (selectorType === autoSelect && shouldScopeSessionKey(ctx)) {
```
yehorkardash, line 138: Would the scoping work if a user connects one memory node to agent and subagent? That maybe a common use case if a user wants to reuse the configuration. Not sure if that's possible, but one solution I can come up with is using caller node name for that

packages/@n8n/nodes-langchain/utils/helpers.ts
lucamattiazzi: that's right, I should return `true` for nodes that are not in `SESSION_KEY_SCOPING_MIN_VERSION`, to be sure that newer types of nodes will still use the scoping

packages/@n8n/nodes-langchain/utils/helpers.ts
lucamattiazzi: I think that the way you suggest could work, since then the expression would resolve correctly, but it would be (I think) visually confusing, since it would look like the two agents are sharing the same memory. the user could simply duplicate the memory node instead.
If you know that sharing the memory node is something that users usually do, I'll use your proposal though!

packages/@n8n/nodes-langchain/utils/helpers.ts
lucamattiazzi: fixed it!

packages/@n8n/nodes-langchain/utils/helpers.ts
yehorkardash: Yeah, I think some advanced users do that, I'm just not sure if can easily get name of the agent caller node. If that's difficult we can live with the current solution

lucamattiazzi: I've run a couple of tests using the parent.name to scope, but seems there are other points where the memory is loaded, and in those cases the parent would not be the agent node (ChatTrigger.node.ts for example), and I'm afraid that it might not show the exact content of the memory node. I can try to find a good solution in a different PR if we see that it's something that users need!

yehorkardash (dismissed):

packages/@n8n/nodes-langchain/utils/helpers.ts
yehorkardash: Okay then! Could you add a documenting comment in the code describing that there's a limitation to this solution when two nodes use same memory node

lucamattiazzi: @yehorkardash added the comment!

yehorkardash (approved):

PR #28678, opened 2026-04-20
fix(ai-builder): Scope artifacts panel to resources produced in-thread

aalises:
## Summary

Linear: https://linear.app/n8n/issue/AI-2379

The Instance AI Artifacts panel showed a workspace-sized list of workflows/data-tables whenever the agent called a list tool for context (e.g. `workflows action=list`). The registry that powers the panel was also used for markdown name→link resolution, so it was deliberately comprehensive — list results, writes, and mutations all landed in the same map.

This PR splits that single registry into two collections with distinct purposes:

- **`producedArtifacts`** (keyed by **resource id**) — resources the agent **built, submitted, created, or mutated** in this thread. Populated only by producer-shaped tool-call results (`result.workflowId`, `result.workflow`, `result.table`, `result.dataTableId + projectId`). Powers the Artifacts panel, canvas preview tabs, and timeline/activity name resolution.
- **`resourceNameIndex`** (keyed by **lowercased name**) — every named resource seen in any tool call, including list results. Powers markdown name→link replacement so references to listed-but-untouched resources still resolve to clickable pills.

### Notable behavior changes

1. **Keying change:** `producedArtifacts` is keyed by **resource id** (the old registry was keyed by lowercased name). Two produced workflows with the same name no longer clobber each other... each has its own entry under its own id.
2. **Merge semantics on repeat writes:** `recordProduced` does an id-keyed upsert that preserves optional fields the new call omits. A patch `build-workflow` call that omits a name no longer regresses a known name to "Untitled"; an explicit `workflowName` on a later `submit-workflow` still wins. This keeps real-time artifact updates working without the regression.
3. **Branch-based split:** `extractFromToolCall` decides which collection(s) to write to based on the *shape* of the result: array branches (list results) only touch `resourceNameIndex`; single-id branches (writes) touch both.

### Observed trace
Thread id `f8c01d16-95eb-4a49-a394-28589a7f459f` (LangSmith `6dc2cfb8-615c-4bca-a842-884abd91717f`) — "super many artifacts" for a Linear-classifier build. The agent listed the user's entire workspace via `workflows action=list` for context, and every workspace workflow landed in the panel.

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/createInstanceAiHarness.ts
+++ b/packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/createInstanceAiHarness.ts
@@ -163,14 +163,16 @@ export async function createInstanceAiHarness(): Promise<InstanceAiHarness> {
 	// --- Mock store ---
 	const messages = ref<InstanceAiMessage[]>([]) as Ref<InstanceAiMessage[]>;
 	const isStreaming = ref(false);
-	const resourceRegistry = ref(new Map<string, ResourceEntry>());
+	const producedArtifacts = ref(new Map<string, ResourceEntry>());
+	const resourceNameIndex = ref(new Map<string, ResourceEntry>());
 
 	const threadMetadata = new Map<string, Record<string, unknown>>();
 
 	const store = reactive({
 		messages,
 		isStreaming,
-		resourceRegistry,
+		producedArtifacts,
+		resourceNameIndex,
 		currentThreadId: 'thread-1',
 		getThreadMetadata: (threadId: string) => threadMetadata.get(threadId),
 		updateThreadMetadata: async (threadId: string, metadata: Record<string, unknown>) => {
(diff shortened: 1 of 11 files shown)


packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/createInstanceAiHarness.ts
```diff
@@ -213,21 +215,34 @@ export async function createInstanceAiHarness(): Promise<InstanceAiHarness> {
 	// --- Convenience actions ---
 
 	function registerWorkflow(id: string, name = `Workflow ${id}`) {
-		const next = new Map(store.resourceRegistry);
-		next.set(name.toLowerCase(), { type: 'workflow', id, name });
-		store.resourceRegistry = next;
+		const entry: ResourceEntry = { type: 'workflow', id, name };
+		const nextProduced = new Map(store.producedArtifacts);
```
mutdmour, line 219: i wonder why we recreate the map every time? some reactivity rule? maybe if we just use objects instead of Map this would be simpler.

mutdmour (approved):
LGTM. An E2E here to make sure this bug does not happen again would be nice, but not a must. Artifact tests I am working on should cover this.

PR #28679, opened 2026-04-20
feat: Removes computer use setup logic on Assistant AI opt-in flow and minor UX changes (no-changelog)

Joco-95:
## Summary
- Removes computer use setup logic on Assistant AI opt-in flow
- Implements computer use enable choice on opt-in flow

https://github.com/user-attachments/assets/a29a6608-7562-4c55-99f4-17b63c551e09

## Review / Merge checklist

- [ ] Tests included. 


dlavrenuek:
@Joco-95 on the video the computer use checkbox is already active when assistant AI is activated. If it's `opt-in` then it should be disabled per default and require use action (check the checkbox) to enable computer use

Cadiac:
we should also flip the setting flags from "disabled" to "enabled", but I think that's beyond this PR..

Joco-95:
> @Joco-95 on the video the computer use checkbox is already active when assistant AI is activated. If it's `opt-in` then it should be disabled per default and require use action (check the checkbox) to enable computer use

Updated the default state of Computer use to disabled

Cadiac (approved):
Seems to work 👍 
This makes the feature much harder to discover for now, but I guess the chat prompt support is coming 🤷

PR #28684, opened 2026-04-20
fix(core): Fix instance-ai planner and prompts after tool consolidation (no-changelog)

OlegIvaniv:
## Summary
#28140 consolidated ~57 individual AI tools into 10 action-family tools (`workflows(action="list")` instead of `list-workflows`, etc.), it also quietly dropped several things we didn't catch in review: 
1. **The planner sub-agent was unwired.** Before #28140, `createOrchestrationTools` registered `plan: createPlanWithAgentTool(context)` and kept `create-tasks` separately for replanning. After #28140, `plan` was rewired to `createPlanTool` (the `create-tasks` tool under a different key) and the `createPlanWithAgentTool` import was deleted. So when the orchestrator called `plan` with a freeform user request, it hit `create-tasks`' schema (`{ tasks: [...] }`) and failed validation. The planner prompt, `add-plan-item` streaming, and progressive task rendering were all unreachable. 
2. **The planner's internal tool whitelist still used old tool names.** `PLANNER_DOMAIN_TOOL_NAMES` listed `list-credentials`, `list-data-tables`, `get-best-practices`, `search-nodes`, etc. None of those exist in `context.domainTools` anymore — it only holds family keys (`workflows`, `nodes`, `credentials`, `data-tables`, `templates`, `research`, `workspace`, `ask-user`). So every whitelist lookup silently missed and the planner got zero discovery tools. Dead code masquerading as working code. 
3. **Prompts still name the old tools.** `build-workflow-agent.prompt.ts`, `data-table-agent.prompt.ts`, `plan-agent-prompt.ts`, and `workflow-loop/guidance.ts` all still instruct agents to call things like `list-credentials`, `search-nodes`, `run-workflow`, `publish-workflow`. System prompt was updated in #28140, the sub-agent prompts were not. 
4. **`nodes(type-definition)` has an inconsistent input key.** The `describe` and `explore-resources` actions take `nodeType: string`. `type-definition` takes `nodeIds: array` — and inside each array element the object variant uses `nodeId`. Three near-synonyms in one tool. LangSmith trace showed the builder agent sending `{ action: "type-definition", nodeType: "n8n-nodes-base.telegramTrigger" }` (the wrong key) and crashing the handler with `Cannot read properties of undefined (reading 'map')`. The crash slips past Zod because `sanitizeInputSchema` flattens the discriminated union for Anthropic compatibility — after flattening, every variant-specific field is optional on the outer object and there's no action→field enforcement left. 

## What this PR does 
- Restores the wiring: ```ts plan: createPlanWithAgentTool(context), 'create-tasks': createPlanTool(context), ``` 
- Updates `PLANNER_DOMAIN_TOOL_NAMES` / `PLANNER_RESEARCH_TOOL_NAMES` to the new family names. 
- Rewrites stale tool references in the four prompt files to `family(action="…")` syntax, matching how `system-prompt.ts` already phrases things. 
- Updates a couple of tests that pinned the old tool-name strings (`guidance.test.ts`, `report-verification-verdict.tool.test.ts`). 
- Fixes `instance-agent.test.ts` — it was still trying to `jest.mock('../tool-access', …)` even though that module was removed in #28140. Swapped to mocking `createOrchestratorDomainTools` on `../../tools`. 
- Renames `nodes(type-definition)` input fields so every action on the tool uses the same `nodeType(s)` term: `nodeIds` → `nodeTypes` (the array), and the inner object variant's `nodeId` → `nodeType`. Output elements follow suit for symmetry. Once every action agrees on `nodeType`, the old `nodeIds` key no longer exists on the flattened schema, so Zod rejects the wrong shape at validation time instead of letting it crash the handler. The frontend only reads `definitions[].content`, so the output rename has no UI impact. 

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/instance-ai/src/agent/__tests__/instance-agent.test.ts
+++ b/packages/@n8n/instance-ai/src/agent/__tests__/instance-agent.test.ts
@@ -33,7 +33,10 @@ jest.mock('../../memory/memory-config', () => ({
 
 jest.mock('../../tools', () => ({
 	createAllTools: jest.fn((context: { runLabel?: string }) => ({
-		'list-workflows': { id: `list-${context.runLabel ?? 'unknown'}` },
+		workflows: { id: `workflows-${context.runLabel ?? 'unknown'}` },
+	})),
+	createOrchestratorDomainTools: jest.fn((context: { runLabel?: string }) => ({
+		workflows: { id: `workflows-${context.runLabel ?? 'unknown'}` },
 	})),
 	createOrchestrationTools: jest.fn((context: { runId: string }) => ({
 		plan: { id: `plan-${context.runId}` },
@@ -56,10 +59,6 @@ jest.mock('../sanitize-mcp-schemas', () => ({
 
 jest.mock('../system-prompt', () => ({
 	getSystemPrompt: jest.fn().mockReturnValue('system prompt'),
-}));
-
-jest.mock('../tool-access', () => ({
-	getOrchestratorDomainTools: jest.fn((tools: Record<string, unknown>) => tools),
 }));
 
 const { createInstanceAgent } =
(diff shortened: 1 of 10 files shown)


Cadiac (dismissed):
Yeah this was clearly missing 👍

Cadiac (approved):

PR #28689, opened 2026-04-20
fix(editor): Fix instance-ai setup parameter issues not resetting on input (no-changelog)

scdekov:
## Summary

The ParameterInputs are getting the parameter value for validation purposes from the `expressionLocalResolveCtx.workflow.getNode()` which points to `workflowsStore.workflowObject` which was not update when parameters are changing.

Prompt to reproduce the issue: "create new workflow that makes a new http request. Open the setup once the wf is created"

https://linear.app/n8n/issue/AI-2392/fix-parameter-setup-error-persists-after-a-valid-value-is-entered

## Review / Merge checklist

- [ ] Tests included. 


packages/frontend/editor-ui/src/features/ai/instanceAi/composables/useSetupCardParameters.ts
```diff
@@ -95,6 +95,19 @@ export function useSetupCardParameters(
 		if (canvasNode) {
 			canvasNode.parameters = { ...canvasNode.parameters, [paramName]: parameterData.value };
 		}
+
+		// 3. `workflowsStore.workflowObject` holds a deep copy of the nodes (see
+		// `createWorkflowObject(..., copyData=true)` in `workflows.store.ts`), and
+		// `ParameterInput` reads its node through that copy via
+		// `expressionLocalResolveCtx.workflow.getNode()`. Without syncing, the
+		// per-input issue indicator would keep checking stale parameters.
+		const workflowObjectNode = workflowsStore.workflowObject.getNode(nodeName);
```
aalises, line 104: A parameter value now lives in 3 places that must stay in sync on every keystroke: `paramValues`, the Pinia store node, and `workflowObject`'s internal node dict (the new write here). 

The next code path that writes a parameter and forgets this third node dict will reintroduce the exact same bug. 

Worth a follow-up to either drop `copyData=true` so the copy aliases the store, or funnel all setup-card writes through one setter so there's 1 source of truth for those

scdekov: Dropping `copyData=true` will break other functionalities so I don't think it's the way to go. Extracting into a dedicated setter does make sense though but I'd prefer to leave this for a bigger refactoring that we should do after the initial release anyways.

aalises (approved):

aalises (approved):

PR #28691, opened 2026-04-20
feat(core): Expose instance registry via proxy service (no-changelog)

afitzek:
## Summary

Introduces `InstanceRegistryProxyService` — a CLI-level bridge that exposes the
instance-registry module to the rest of the CLI through a stable interface
(`InstanceRegistryProvider`), without coupling consumers to the module's
feature-flag state.

### Why

The `instance-registry` module (gated by `N8N_ENV_FEAT_INSTANCE_REGISTRY`)
tracks main/worker/webhook processes in distributed deployments. Until now,
any CLI code that wanted cluster information had to import
`InstanceRegistryService` directly — a hard dependency on an optional module.

This PR lays the consumer-facing surface for that module:

- `InstanceRegistryProvider` — the well-known interface (`getAllInstances`,
  `getLocalInstance`).
- `InstanceRegistryProxyService` — a singleton, always-available
  `@Service()` in `src/services/` that implements the interface.
- When the module initializes, it registers `InstanceRegistryService` as the
  proxy's provider. When the feature flag is off, no provider is registered
  and the proxy returns safe defaults (`[]` / `null`), so consumers can call
  it unconditionally.

Scope is deliberately narrow. Version-mismatch computation
(`getVersionMismatch()`) and rewiring
`InstanceRegistryController.getClusterInfo()` to the proxy are handled in
follow-up tickets.

### How to test manually

There is no user-visible change in this PR. The existing REST endpoint
`GET /rest/instance-registry` still returns `{ instances, versionMismatch: null }`
exactly as before (hardcoded `null` in the controller, unchanged).

Smoke verification is covered by unit tests
(`packages/cli/src/services/__tests__/instance-registry-proxy.service.test.ts`):

- Without a registered provider → `getAllInstances()` returns `[]`,
  `getLocalInstance()` returns `null`.
- With a registered provider → both methods delegate.
- `getLocalInstance()` forwards a `null` from the provider unchanged.
- Re-registering swaps in the latest provider (latest-wins).

If desired, you can also run n8n locally with
`N8N_ENV_FEAT_INSTANCE_REGISTRY=true` and confirm the existing endpoint
still responds normally — behavior is unchanged in this PR.

- **Proxy lives in `packages/cli/src/services/`, not inside the module folder.**
  The module file cannot eagerly import the proxy (blocked by
  `n8n-local-rules/no-top-level-relative-imports-in-backend-module`), and the
  proxy must remain resolvable from DI even when the feature flag is off.
  Placing it under `src/services/` lets CLI consumers import it
  unconditionally while the module performs a lazy dynamic `import(...)`
  inside its `init()` to call `registerProvider()`.
- **`InstanceRegistryProvider.getLocalInstance()` returns `InstanceRegistration | null`.**
  The proxy widens the underlying service's non-nullable return type so the
  "no provider registered" state is expressible through the same contract.
  `InstanceRegistryService`'s non-null return remains structurally compatible.
- **Latest-wins on re-registration.** Simple and sufficient for the current
  single-module topology; documented via a test.

closes https://linear.app/n8n/issue/IAM-330

## Review / Merge checklist

- [X] Tests included.

### DIFF
--- a/packages/cli/src/services/instance-registry-proxy.service.ts
+++ b/packages/cli/src/services/instance-registry-proxy.service.ts
@@ -1,0 +1,26 @@
+import type { InstanceRegistration } from '@n8n/api-types';
+import { Service } from '@n8n/di';
+
+export interface InstanceRegistryProvider {
+	getAllInstances(): Promise<InstanceRegistration[]>;
+	getLocalInstance(): InstanceRegistration | null;
+}
+
+@Service()
+export class InstanceRegistryProxyService implements InstanceRegistryProvider {
+	private provider: InstanceRegistryProvider | null = null;
+
+	registerProvider(provider: InstanceRegistryProvider): void {
+		this.provider = provider;
+	}
+
+	async getAllInstances(): Promise<InstanceRegistration[]> {
+		if (!this.provider) return [];
+		return await this.provider.getAllInstances();
+	}
+
+	getLocalInstance(): InstanceRegistration | null {
+		if (!this.provider) return null;
+		return this.provider.getLocalInstance();
+	}
+}
--- a/packages/cli/src/modules/instance-registry/instance-registry.module.ts
+++ b/packages/cli/src/modules/instance-registry/instance-registry.module.ts
@@ -24,7 +24,13 @@ export class InstanceRegistryModule implements ModuleInterface {
 		await import('./instance-registry.controller');
 
 		const { InstanceRegistryService } = await import('./instance-registry.service');
-		await Container.get(InstanceRegistryService).init();
+		const instanceRegistryService = Container.get(InstanceRegistryService);
+		await instanceRegistryService.init();
+
+		const { InstanceRegistryProxyService } = await import(
+			'@/services/instance-registry-proxy.service'
+		);
+		Container.get(InstanceRegistryProxyService).registerProvider(instanceRegistryService);
 
 		const { StaleMemberCleanupService } = await import('./stale-member-cleanup.service');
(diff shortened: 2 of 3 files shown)


packages/cli/src/services/instance-registry-proxy.service.ts
```diff
@@ -0,0 +1,26 @@
+import type { InstanceRegistration } from '@n8n/api-types';
```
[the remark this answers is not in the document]
afitzek: This was descoped during development.

packages/cli/src/services/instance-registry-proxy.service.ts
```diff
@@ -0,0 +1,26 @@
...
+@Service()
+export class InstanceRegistryProxyService implements InstanceRegistryProvider {
+	private provider: InstanceRegistryProvider | null = null;
+
+	registerProvider(provider: InstanceRegistryProvider): void {
+		this.provider = provider;
+	}
+
+	async getAllInstances(): Promise<InstanceRegistration[]> {
+		if (!this.provider) return [];
```
guillaumejacquart, line 18: do we want a special return type for no provider ? in order to make a difference between no instances registered and no provider configured ?

guillaumejacquart (commented):
1 comment

packages/cli/src/services/instance-registry-proxy.service.ts
afitzek: I think that is not necessary, because services using this proxy shouldn't be concerned about this.

guillaumejacquart (approved):

PR #28694, opened 2026-04-20
feat(agents): Node to message a new agent

mike12345567:
## Summary

Adds a **Message an Agent** node to `nodes-base`, letting workflows send a message to an n8n SDK agent (from the new `agents` module) and receive its response, structured output, tool calls, and token usage.

Pick an agent from a searchable dropdown (or supply an ID), provide a message, and the node returns:

| Field | Description |
|-------|-------------|
| `response` | The agent's final assistant text |
| `structuredOutput` | Parsed structured output if the agent defines one |
| `usage` | `{ promptTokens, completionTokens, totalTokens }` |
| `toolCalls` | Tools invoked during the run (name, input, result) |
| `finishReason` | Why the agent stopped |

Agents must be **published** to run from a workflow — drafts are rejected with a clear error. Each item gets an isolated thread (`${executionId}-${itemIndex}`) so multi-item batches don't bleed into each other's context.

## Screenshots

image
image

## Review / Merge checklist


packages/nodes-base/nodes/MessageAnAgent/MessageAnAgent.node.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export class MessageAnAgent implements INodeType {
+	description: INodeTypeDescription = {
+		displayName: 'Message an Agent',
+		name: 'messageAnAgent',
+		icon: 'fa:robot',
+		group: ['transform'],
+		version: 1,
+		hidden: true,
+		description: 'Send a message to an SDK agent and receive its response',
```
yehorkardash, line 20: would it work if we also add `usableAsTool:true`?

mike12345567: Done good idea! Let these agents be sub-agents.

yehorkardash (approved):

PR #28697, opened 2026-04-20
fix(Alibaba Cloud Chat Model Node): Add credential-level url field for AI gateway compatibility

DawidMyslak:
## Summary

Adds a hidden `url` field to `AlibabaCloudApi` credentials that computes the base URL from the selected region via expression. This gives the AI gateway a `urlField` to target when injecting synthetic credentials — previously the URL was constructed in code with no field for the gateway to override.

Both Alibaba Cloud nodes now read `credentials.url`:
- **Chat Model** appends `/compatible-mode/v1` for the OpenAI-compatible API
- **Vendor node** uses the url directly for its own API paths

Dead code removed: `getBaseUrl`, `getBaseHost`, `BASE_URL_EXPRESSION`.

No version bump — internal plumbing only, same URLs produced from same inputs.

## Related

https://linear.app/n8n/issue/NODE-4867

## Review / Merge checklist

- [x] Tests included.


alexander-gekov (approved):
Confirming this works with the AI Gateway

PR #28701, opened 2026-04-20
feat(agents): Add Build/Test chat toggle to iterate on agents after creation

bjorger:
## Summary

Adds a Build / Test chat toggle to the agent detail view so users can keep iterating with the builder agent after initial creation (previously a one-shot flow). The sidebar reshapes itself depending on which chat mode is active, giving the config editor first-class screen real estate while the builder is writing to it.

Builder View:
image

Test View:
image

https://linear.app/n8n/issue/AGENT-25/feature-builder-agent-toggle-and-iteration

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/agents/composables/useAgentChatStream.ts
```diff
@@ -0,0 +1,237 @@
...
+			const dbMessages = await fetchMessages(
+				rootStore.restApiContext,
+				params.projectId.value,
+				params.agentId.value,
+			);
+			if (dbMessages.length > 0) {
+				messages.value = convertDbMessages(dbMessages);
+			}
+		} catch {
+			// Silently ignore — just start with empty chat
```
mike12345567: Theres a few silent ignores in here, I think if it fails to load history we should report it to the user.

packages/frontend/editor-ui/src/features/agents/components/AgentChatPanel.vue
```diff
@@ -31,302 +30,47 @@ const emit = defineEmits<{
...
-	if (!lastMsg || lastMsg.role === 'user') return 'waitingFirstChunk';
-	return 'receiving';
+const {
+	messages,
+	isStreaming,
+	messagingState,
+	loadHistory,
+	clearHistory,
+	sendMessage,
+	stopGenerating,
```
mike12345567, line 42: Should we `stopGenerating` on unmount? Incase the user navigates away, stop the request.

packages/frontend/editor-ui/src/features/agents/views/AgentBuilderView.vue
```diff
@@ -344,16 +399,26 @@ watch(agentId, initialize, { immediate: true });
...
-						key="chat"
-						:project-id="projectId"
-						:agent-id="agentId"
-						mode="inline"
-						endpoint="chat"
-						:initial-message="initialPrompt"
-						@config-updated="onConfigUpdated"
-					/>
+					<div v-else key="chat" :class="$style.chatHost">
+						<AgentChatPanel
```
mike12345567, line 423: Since both of these are mounted at the same time, does that mean that we'll end up loading/configuring both panels?

packages/frontend/editor-ui/src/features/agents/composables/useAgentApi.ts
```diff
@@ -223,6 +223,30 @@ export const clearBuilderMessages = async (
 	);
 };
 
+export const getChatMessages = async (
+	context: IRestApiContext,
+	projectId: string,
+	agentId: string,
+): Promise<unknown[]> => {
```
mike12345567: We could likely return a type for this, a typed DTO for the endpoint?

mike12345567 (approved):
Few minor comments, great improvement, thanks for tidying up/breaking down some of the frontend components as well!

packages/frontend/editor-ui/src/features/agents/views/AgentBuilderView.vue
bjorger: Yes - they are. Good catch I will now only load the history once the panel is visible

mike12345567 (approved):

PR #28702, opened 2026-04-20
fix(core): Allow SAML connection test without saving config first

guillaumejacquart:
## Summary

The SAML connection test failed with "No IdP metadata configured" unless the user first saved their config, which contradicted the "test before save" UX the test endpoint was meant to provide. The ACS callback was still reading metadata from the persisted preferences, so the inbound leg of the round-trip had nothing to validate against.

The fix caches the submitted metadata server-side under a short-lived (10 min) random token, embeds that token in the SAML `RelayState` (`?t=<id>`), and on ACS callback rebuilds a temporary IdP instance from the cached metadata when a testId is present. Tokens are single-use and pruned on access.

### How to test

1. On an enterprise license, go to Settings → SSO → SAML.
2. Paste IdP metadata XML (or a metadata URL) without saving.
3. Click **Test connection**, authenticate at the IdP.
4. ACS should render the success template with attributes instead of "No IdP metadata configured".
5. Save a valid config, toggle SAML on, sign in via `/sso/saml/initsso` — existing login flow still works.

https://linear.app/n8n/issue/IAM-539

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/modules/sso-saml/saml.service.ee.ts
```diff
@@ -40,13 +40,17 @@ import { SamlValidator } from './saml-validator';
...
+const TEST_CONFIG_TTL_MS = 10 * 60 * 1000;
+
 @Service()
 export class SamlService {
 	private identityProviderInstance: IdentityProviderInstance | undefined;
 
 	// eslint-disable-next-line @typescript-eslint/consistent-type-imports
 	private samlify: typeof import('samlify') | undefined;
 
+	private readonly pendingTestConfigs = new Map<string, { metadata: string; expiresAt: number }>();
```
afitzek: We should use the cache service, otherwise in a multi main scenario, the instance might not have the test config.

packages/cli/src/modules/sso-saml/saml.controller.ee.ts
```diff
@@ -200,27 +207,29 @@ export class SamlController {
...
-		if (config) {
-			metadata = config.metadata;
-			if (!metadata && config.metadataUrl) {
-				metadata =
-					(await this.samlService.fetchMetadataFromUrl(config.metadataUrl, config.ignoreSSL)) ?? '';
-			}
+		let relayState = getServiceProviderConfigTestReturnUrl();
+		if (metadata) {
+			const testId = this.samlService.storePendingTestConfig(metadata);
+			relayState = `${relayState}?t=${testId}`;
```
afitzek: Will this work if the relayState is already an url encoded value, that we add `?` or would we need to concat with `&`.

packages/cli/src/modules/sso-saml/saml.service.ee.ts
guillaumejacquart: Ah yes, good point :)

packages/cli/src/modules/sso-saml/saml.service.ee.ts
```diff
@@ -314,9 +311,54 @@ export class SamlService {
...
+		return testId;
+	}
+
+	/**
+	 * Retrieves and removes the pending test metadata associated with the given
+	 * token. Returns undefined if the token is unknown or expired.
+	 */
+	async consumePendingTestConfig(testId: string): Promise<string | undefined> {
+		const key = `${TEST_CONFIG_CACHE_PREFIX}${testId}`;
+		const metadata = await this.cacheService.get<string>(key);
```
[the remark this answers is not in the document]
guillaumejacquart: i think that's all right for testing purposes

afitzek (dismissed):
Nice, thanks for addressing the comments 👍

afitzek (approved):

PR #28709, opened 2026-04-20
feat(core): Broadcast workflow updates from MCP tools to open editors

dariacodes:
## Summary

**Problem**: When a user has a workflow opened in a tab and it is updated through an MCP client at the same time, they don't receive live updates on the canvas.
**Solution**: Send workflow update push events from the MCP server - reuse the system that we already have.

Another problem that this PR solves is that if any user is currently editing the same workflow on the canvas, we now return an error from MCP:
image

Closes [ADO-5022](https://linear.app/n8n/issue/ADO-5022/feature-update-canvas-when-workflow-is-updated-on-the-backend)

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/mcp/tools/workflow-builder/update-workflow.tool.ts
```diff
@@ -177,6 +179,8 @@ export const createUpdateWorkflowTool = (
 				source: 'n8n-mcp',
 			});
 
+			await collaborationService.broadcastWorkflowUpdate(workflowId, user.id);
```
MiloradFilipovic: Should we wrap these broadcasts in a separate `try`/`catch` to avoid sending error response to MCP clients if it's just the broadcast that fails?

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
```diff
@@ -286,7 +286,7 @@ export const useCollaborationStore = defineStore(STORES.COLLABORATION, () => {
 	}
 
 	async function handleWorkflowUpdate() {
-		if (isCurrentTabWriter.value || !collaboratingWorkflowId.value) {
```
MiloradFilipovic, line 289: Is this now opening a possibility that local changes can be overwritten by MCP push?

MiloradFilipovic (dismissed):
Looking good, I left a couple of comments so take a look at what makes sense.

packages/cli/src/modules/mcp/tools/workflow-builder/update-workflow.tool.ts
dariacodes: Good point! And we do not have to await it either. Will update it

packages/frontend/editor-ui/src/features/collaboration/collaboration/collaboration.store.ts
dariacodes: Good catch! We discussed this and I totally forgot to solve this problem. I decided to send an error from MCP while the workflow is being edited - pushed this in https://github.com/n8n-io/n8n/pull/28709/commits/4543a333203cb60059e892578032f02189112e63 and added it to the description. And this also allowed to revert the commit that was overcomplicating things with the exclude logic, so it looks better now. What do you think?

There is also another conflict problem with workflow being edited by two different MCP clients... Maybe we should start sending `expectedChecksum`, but that would be a big change. I propose to leave it for now.

packages/cli/src/modules/mcp/tools/workflow-builder/delete-workflow.tool.ts
```diff
@@ -47,12 +49,16 @@ export const createArchiveWorkflowTool = (
...
 		try {
+			await collaborationService.ensureWorkflowEditable(workflowId);
+
 			const workflow = await workflowService.archive(user, workflowId, { skipArchived: true });
 
 			if (!workflow) {
 				throw new Error("Workflow not found or you don't have permission to archive it.");
 			}
 
+			void collaborationService.broadcastWorkflowUpdate(workflowId, user.id).catch(() => {});
```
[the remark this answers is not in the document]
MiloradFilipovic: I would be ok with leaving this empty, but maybe we can use `note` (like [here](https://github.com/n8n-io/n8n/blob/86ceb68a05f8b30ea22126b3b09a311a8521b3d7/packages/cli/src/modules/mcp/tools/workflow-builder/create-workflow-from-code.tool.ts#L64)) property to add a hint that something auxiliary failed.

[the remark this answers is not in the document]
dariacodes: Ah, I was wondering whether we have a way of reporting something like that. Tbh, I think here we shouldn't do it as it's not directly related to MCP functionality

packages/cli/src/collaboration/collaboration.service.ts
```diff
@@ -274,6 +274,18 @@ export class CollaborationService {
 		return await this.state.getWriteLock(workflowId);
 	}
 
+	/**
+	 * Throws if any user currently holds the write lock for the given workflow.
+	 */
+	async ensureWorkflowEditable(workflowId: Workflow['id']): Promise<void> {
```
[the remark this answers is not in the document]
dariacodes: This method is now used after access checks, so no leaking. I don't want to add access checks to this method as we already have them in the callers

MiloradFilipovic (approved):
I think the strategy change makes sense and the UX is now in line with the editor.
Can you just take a look if the second comment from Cubic needs addressing, otherwise looks good to me.

dariacodes:
@MiloradFilipovic Thanks for reviewing! I already addressed it in the last commit, so time to merge 🎉

PR #28711, opened 2026-04-20
fix(core): Support mixed bracket/dot notation in external secrets expressions

### LINKED ISSUE #28516: Breaking change in 2.9.0 when saving/sharing Credentials, using external vault
### Bug Description

After migrating from the v1 to the v2, we have an issue with all our credentials: **we can't update or share them anymore.**
The error we have: `Could not find a valid external secret vault name inside "={{ $secrets.azureKeyVault["postgres-n8n-data"] }}" used in "password"`

We using an external secret provider (azure key vault).
An example of value in a credential : `{{ $secrets.azureKeyVault["postgres-n8n-data"] }}`. We have to use the `[ ]` because of the `-` in the secret name in the azure key vault (this is a internal convention).

In the UI, the expression is valid
Image

The issue comes from this PR https://github.com/n8n-io/n8n/pull/25406, released in the 2.9.0
A new check of the format forget to check the possibility to have a mix of `.` and `[]`

Expression | Dot regex ((?=\\.)) | Bracket regex ($secrets[ )
-- | -- | --
$secrets.vault.key | ✅ matches vault | ❌
$secrets["vault"]["key"] | ❌ | ✅ matches vault
$secrets.vault["key"] | ❌ (next char is [) | ❌ ($secrets[ not present)

So there is a diff between the validation in the UI and the validation in the back for credential fields.
Also, there is a lack of documentation regarding the possible format for the credential fields when using external secrets (https://docs.n8n.io/external-secrets/#use-secrets-in-n8n-credentials)

### To Reproduce

1. Deploy a [2.8.4](https://github.com/n8n-io/n8n/releases/tag/n8n%402.8.4) version of n8n
2. Configure an external vault (example : azure key vault)
3. Add a new secret inside the vault: `postgres-n8n-data` (dash are important here) to store the password connection
4. Create a new postgre credential on n8n and inside `Password` field, set `{{ $secrets.azureKeyVault["postgres-n8n-data"] }}` as expression.
5. Update n8n to latest version 
6. Try to update the credential or to share it, you will have an error

### Expected behavior

`{{ $secrets.azureKeyVault["postgres-n8n-data"] }}` is valid as it's green on the UI (so valid expression)

### Debug Info

N/A

### Operating System

Kubernetes

### n8n Version

2.15.1

### Node.js Version

24

### Database

PostgreSQL

### Execution mode

queue

### Hosting

self hosted

alielkhateeb:
## Summary

https://www.loom.com/share/4c8e1a06e8234140b2cf5dc886bfd0cd

Fixes a bug where credential expressions using **dot notation for the provider** and **bracket notation for the secret key** — e.g. `$secrets.vault['common/n8n']` — would fail to save. The provider key was not being extracted because the regex lookahead only matched `.` and not `[` as the next character after the provider name.

**Changes:**
- Updated the dot-notation regex in `extractProviderKeysFromExpression` to match provider keys followed by either `.` or `[`
- Relaxed the bracket-notation regex to capture any string and validate it against `PROVIDER_KEY_PATTERN` separately (handles provider keys with special characters like `-`)
- Added regression tests for mixed notation patterns

**To test:**
1. Enable an external secrets provider (e.g. Vault)
2. Create a credential that references a secret using mixed notation: `={{ $secrets.vault['secret-name'] }}`
3. Verify the credential saves successfully

https://linear.app/n8n/issue/LIGO-435
fixes #28516

## Review / Merge checklist

- [x] Tests included.

---

🤖 PR Summary generated by AI


packages/cli/src/credentials/__tests__/external-secrets.utils.test.ts
```diff
@@ -16,6 +16,18 @@ describe('External secrets utils', () => {
...
 
+		it('extracts single provider from mixed notation with bracket secret access', () => {
+			expect(extractProviderKeysFromExpression("={{ $secrets.vault['secret-name'] }}")).toEqual([
+				'vault',
+			]);
+		});
+
+		it('extracts single provider from mixed notation with bracket provider access', () => {
+			expect(
+				extractProviderKeysFromExpression('={{ $secrets["aws-secrets-manager"].secret }}'),
```
alielkhateeb: I know that secret vaults won't contain `-` but users can still use bracket notation to refer to secret vaults, this is why the regex is slightly different than the fix described in the ticket.

konstantintieber:
Do we support single quotes and double quote prop access in other expressions as well?
As far as I know double quote would only appear if entered manually. Our autocomplete via codemirror would only use single quotes

packages/cli/src/credentials/external-secrets.utils.ts
```diff
@@ -38,15 +38,20 @@ export function extractProviderKeysFromExpression(expression: string): string[]
...
 		const expressionContent = expression[1]; // Content inside {{ }}
 
-		// Match all dot notation occurrences: $secrets.providerKey
+		// Match provider keys in dot notation, including mixed notation:
+		// - $secrets.providerKey.secret
+		// - $secrets.providerKey['secret']
+		// - $secrets.providerKey["secret"]
 		const dotMatches = expressionContent.matchAll(
-			new RegExp(`\\$secrets\\.(${PROVIDER_KEY_PATTERN})(?=\\.)`, 'g'),
+			new RegExp(`\\$secrets\\.(${PROVIDER_KEY_PATTERN})(?=\\.|\\[)`, 'g'),
```
geemanjs: Visualisation of this regex:
Image

packages/cli/src/credentials/external-secrets.utils.ts
```diff
@@ -38,15 +38,20 @@ export function extractProviderKeysFromExpression(expression: string): string[]
...
 		for (const match of dotMatches) {
 			providerKeys.add(match[1]);
 		}
 
-		// Match all bracket notation occurrences: $secrets['providerKey'] or $secrets["providerKey"]
+		// Match bracket notation occurrences where the provider key is bracket-accessed:
+		// - $secrets['providerKey']
+		// - $secrets["providerKey"]
 		const bracketMatches = expressionContent.matchAll(
 			new RegExp(`\\$secrets\\[['"](${PROVIDER_KEY_PATTERN})['"]\\]`, 'g'),
```
geemanjs: Visualisation of this regex: 
Image

Question - Should we add `_` in the `PROVIDER_KEY_PATTERN`?

alielkhateeb: Thanks for the visualisations these are pretty useful 😍 
I do not think we should add `_` because we only allow letters and numbers in the provider keys
image

alielkhateeb: Even the hyphen here is questionable I know we historically allowed them, but I can't recall if we released it to the public before restricting this to alphanumeric or not 😂 if we haven't then the `PROVIDER_KEY_PATTERN` can be modified to remove the hyphen.

geemanjs (dismissed):

packages/cli/src/credentials/external-secrets.utils.ts
ireneea: We released it after restricting to alphanumeric 😉

alielkhateeb: AWESOME! removing the hyphen matching ...

alielkhateeb: @geemanjs 31d56622db644988de5bb62305bbdefc86826308 this makes the regex much simpler
I also extracted this to api-types instead of having this comment "Keep in sync with" I hated that 😂

geemanjs: I nearly commented that suggestion - but then I thought if we update the validation - we still need to support the "time before" we changed the validation. Lets keep it though the old YAGNI probably holds strong enough.

geemanjs (dismissed):

packages/cli/src/credentials/external-secrets.utils.ts
alielkhateeb: Now lets hope I did not break any tests 😂 I expect tests to fail honestly I'd be surprised if they did not fail

ireneea (approved):

PR #28712, opened 2026-04-20
fix(RabbitMQ Trigger Node): Redeliver messages on failed executions

### LINKED ISSUE #28605: RabbitMQ trigger & Delete from queue options, not working as expected.
### Bug Description

RabbitMQ trigger node behaves differently than stated.

When setting the 'delete from queue when' to 'Specifiek later in workflow':
In production: message wil delete from the queue when workflow crashes. But it should be added back to the queue on crash ( would be nice ), else it should stay NACK, untill it reaches timeout, and then sent back to the queue.

In Editor mode: its hard to test, but it will stay NACK when crashed, and will stay NACK untill timeout reaches, and sends back to queue. 

### To Reproduce

In produciton(published). 
1. make a rabbitmq trigger, and set it to a queue.
2. set it to the option, delete when delete message node is activated
3. set timer of 20 seconds, then make it crash on purpose, without reaching the delete from queue node.
4.  Msg should be deleted from the queue, instead of moving back to the queue.

### Expected behavior

I expect the message to be sent back to the queue on crashed workflow, and will retry on next trigger, untill it reaches the delete form queue node, then it should be deleted. 

### Debug Info

Image

Image

These images were a production flow, where the flow purposely failed before reaching the node delete from queue, yet it shows the msg was still deleted from the queue. 

### Operating System

Windows 11

### n8n Version

n8n@2.16.1

### Node.js Version

unclear where to find it.

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

BerniWittmann:
## Summary

When the RabbitMQ Trigger is configured with **Delete From Queue When = Specific later in workflow** and a workflow crashes before reaching the explicit *Delete from Queue* action node, the message was being ack'd anyway, silently dropping it from the broker.

`handleMessage` used to await only the `sendResponse` hook for this acknowledge mode. The engine resolves that hook unconditionally at execution teardown (`resolveExecutionResponsePromise` in `packages/cli/src/active-executions.ts`), so the hook fires on every run — including crashes, before `sendResponse` was ever called. That made the hook useless as an ack signal.

This PR races the `sendResponse` hook against the final `IRun` (`workflowExecuteAfter`) and acts on whichever resolves first:

- **Hook wins** — the *Delete from Queue* node fired `sendResponse` mid-execution → `ack` immediately, without waiting for downstream nodes.
- **IRun wins with error** — the workflow crashed before the Delete node ran → `nack` so the broker redelivers. This is safe because `workflowExecuteAfter` fires strictly before the teardown-level hook resolve, so the IRun reliably wins the race on crash.
- **IRun wins clean** — workflow completed without firing the Delete node (misconfiguration) → `ack`, preserving pre-fix behaviour for this non-bug case.

**Behavior change (at-least-once semantics restored):** users who were unknowingly losing messages during failed runs will now see those messages redelivered — the documented intent of this feature. Happy-path timing is unchanged: the ack still happens as soon as the Delete-from-Queue node fires.

### How to test

1. Run a local broker: `docker run --rm -p 5672:5672 rabbitmq:3`.
2. Create a workflow: RabbitMQ Trigger (`acknowledge = laterMessageNode`) → Code node that throws → RabbitMQ (Delete from Queue).
3. Activate, publish a message, observe the queue:
   - **Scenario A** (no crash, Delete from Queue runs): queue returns to 0 as soon as Delete fires, even if later nodes keep running.
   - **Scenario B** (Code node crashes): before this PR the queue still returns to 0 (message lost); after this PR the message is redelivered on the next consumer.

 Queue before fix:

Bildschirmfoto 2026-04-20 um 16 43 25

 Queue after fix:

Bildschirmfoto 2026-04-20 um 16 44 19

https://linear.app/n8n/issue/NODE-4863
Fixes #28605

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/RabbitMQ/GenericFunctions.ts
+++ b/packages/nodes-base/nodes/RabbitMQ/GenericFunctions.ts
@@ -218,17 +218,52 @@ export async function handleMessage(
 
 		let responsePromise: IDeferredPromise<IRun> | undefined = undefined;
 		let responsePromiseHook: IDeferredPromise<IExecuteResponsePromiseData> | undefined = undefined;
-		if (acknowledgeMode !== 'immediately' && acknowledgeMode !== 'laterMessageNode') {
-			responsePromise = this.helpers.createDeferredPromise();
-		} else if (acknowledgeMode === 'laterMessageNode') {
+		if (acknowledgeMode === 'laterMessageNode') {
 			responsePromiseHook = this.helpers.createDeferredPromise<IExecuteResponsePromiseData>();
+			// Also await execution end so we can nack when the run errors
+			// before the Delete-from-Queue node fires sendResponse. The engine
+			// unconditionally resolves the hook at teardown via
+			// resolveExecutionResponsePromise with an empty {} — the payload
+			// shape is what tells a real sendResponse apart from that cleanup.
+			responsePromise = this.helpers.createDeferredPromise<IRun>();
+		} else if (acknowledgeMode !== 'immediately') {
+			responsePromise = this.helpers.createDeferredPromise();
 		}
-		if (responsePromiseHook) {
-			this.emit([[item]], responsePromiseHook, undefined);
-		} else {
-			this.emit([[item]], undefined, responsePromise);
-		}
-		if (responsePromise && acknowledgeMode !== 'laterMessageNode') {
+		this.emit([[item]], responsePromiseHook, responsePromise);
+		if (acknowledgeMode === 'laterMessageNode' && responsePromise && responsePromiseHook) {
+			// The hook resolves both from a node calling sendResponse
+			// (mid-execution, carries the item JSON) AND from teardown cleanup
+			// (resolves with {}). Treat only a non-empty payload as a real
+			// sendResponse from Delete-from-Queue — that's the fast-ack path.
+			// For the empty cleanup, fall through to the final IRun and decide
+			// ack vs nack based on error state.
+			type RaceResult = { kind: 'hook'; isRealSendResponse: boolean } | { kind: 'run'; data: IRun };
+			const hookRace: Promise<RaceResult> = responsePromiseHook.promise.then((data) => ({
+				kind: 'hook',
+				isRealSendResponse:
+					data !== null &&
+					data !== undefined &&
+					typeof data === 'object' &&
+					Object.keys(data as object).length > 0,
+			}));
+			const runRace: Promise<RaceResult> = responsePromise.promise.then((data) => ({
+				kind: 'run',
+				data,
+			}));
+			const first = await Promise.race([hookRace, runRace]);
+
+			if (first.kind === 'hook' && first.isRealSendResponse) {
+				channel.ack(message);
+			} else {
+				const runData = first.kind === 'run' ? first.data : await responsePromise.promise;
+				if (runData?.data?.resultData?.error) {
+					channel.nack(message);
+				} else {
+					channel.ack(message);
+				}
+			}
+			messageTracker.answered(message);
+		} else if (responsePromise && acknowledgeMode !== 'laterMessageNode') {
 			// Acknowledge message after the execution finished
 			await responsePromise.promise.then(async (data: IRun) => {
 				if (data.data.resultData.error) {
@@ -239,11 +274,6 @@ export async function handleMessage(
 						return;
 					}
 				}
-				channel.ack(message);
-				messageTracker.answered(message);
-			});
-		} else if (responsePromiseHook && acknowledgeMode === 'laterMessageNode') {
-			await responsePromiseHook.promise.then(() => {
 				channel.ack(message);
 				messageTracker.answered(message);
 			});
--- a/packages/nodes-base/nodes/RabbitMQ/test/GenericFunctions.test.ts
+++ b/packages/nodes-base/nodes/RabbitMQ/test/GenericFunctions.test.ts
@@ -1,6 +1,12 @@
 import type { Channel, Connection, ConsumeMessage, Message } from 'amqplib';
 import { mock, mockDeep } from 'jest-mock-extended';
-import type { INode, IRun, ITriggerFunctions, IWorkflowMetadata } from 'n8n-workflow';
+import type {
+	IExecuteResponsePromiseData,
+	INode,
+	IRun,
+	ITriggerFunctions,
+	IWorkflowMetadata,
+} from 'n8n-workflow';
 
 const mockChannel = mock<Channel>();
 const mockConnection = mock<Connection>({ createChannel: async () => mockChannel });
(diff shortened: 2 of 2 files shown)


packages/nodes-base/nodes/RabbitMQ/GenericFunctions.ts
```diff
@@ -218,17 +218,30 @@ export async function handleMessage(
...
+			// before Delete-from-Queue) so the broker redelivers, ack on
+			// clean completion.
+			const data: IRun = await responsePromise.promise;
+			if (data?.data?.resultData?.error) {
+				channel.nack(message);
+			} else {
+				channel.ack(message);
+			}
+			messageTracker.answered(message);
+		} else if (responsePromise && acknowledgeMode !== 'laterMessageNode') {
```
RomanDavydchuk: I don't think this is quite correct: `responsePromise` is a promise that's resolved once the execution finishes and we always await it here. `responsePromise` will be resolved _before_ `responsePromiseHook` only when the execution **errors before the "response node"**. In other cases - `responsePromiseHook` will resolve first. It means that for successful execution it functions essentially the same as the `executionFinishesSuccessfully` response mode

To see it in action try this workflow:
```json
{
  "nodes": [
    {
      "parameters": {
        "queue": "n8n-dev",
        "options": {
          "acknowledge": "executionFinishesSuccessfully"
        }
      },
      "type": "n8n-nodes-base.rabbitmqTrigger",
      "typeVersion": 1,
      "position": [
        0,
        0
      ],
      "id": "9c86df32-563f-4518-bdec-c72d9f5aa4bb",
      "name": "RabbitMQ Trigger",
      "credentials": {
        "rabbitmq": {
          "id": "XwRkoqtkG9W3JpgX",
          "name": "RabbitMQ (localhost)"
        }
      }
    },
    {
      "parameters": {
        "operation": "deleteMessage"
      },
      "type": "n8n-nodes-base.rabbitmq",
      "typeVersion": 1.1,
      "position": [
        448,
        0
      ],
      "id": "afdaa27c-f810-4b62-9199-f061950a2217",
      "name": "RabbitMQ",
      "credentials": {
        "rabbitmq": {
          "id": "XwRkoqtkG9W3JpgX",
          "name": "RabbitMQ (localhost)"
        }
      }
    },
    {
      "parameters": {
        "amount": 50
      },
      "type": "n8n-nodes-base.wait",
      "typeVersion": 1.1,
      "position": [
        672,
        0
      ],
      "id": "eaeea8ea-7cc1-4011-80f1-3118d2064e67",
      "name": "Wait1",
      "webhookId": "db539f29-b4e7-4aee-b277-c8a785cf939e"
    },
    {
      "parameters": {
        "amount": 10
      },
      "type": "n8n-nodes-base.wait",
      "typeVersion": 1.1,
      "position": [
        224,
        0
      ],
      "id": "e2ed54b9-51f5-4a0b-b900-b4dbe7179551",
      "name": "Wait",
      "webhookId": "a3e7f1bc-5d30-4bd7-b5c1-f3d27eb84efb"
    }
  ],
  "connections": {
    "RabbitMQ Trigger": {
      "main": [
        [
          {
            "node": "Wait",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "RabbitMQ": {
      "main": [
        [
          {
            "node": "Wait1",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Wait": {
      "main": [
        [
          {
            "node": "RabbitMQ",
            "type": "main",
            "index": 0
          }
        ]
      ]
    }
  },
  "pinData": {},
  "meta": {
    "templateCredsSetupCompleted": true,
    "instanceId": "935ffbc4ed6068e5224bb41b3fac79658dd867ba2639d61c1773533283b0e647"
  }
}
```

There is a 10s delay between the workflow start and the "response node" execution, but the message actually stays in the queue for ~1m, because that's how long it takes for the whole workflow to execute:
Image

The first thought I have is to use `Promise.any` and decide what to do based on which promise got resolved first

RomanDavydchuk (changes requested):
Did some manual testing, the bug that's reported is fixed, but I encountered an issue while testing the happy path, left a comment in the code. Let me know if you can reproduce it. Overall I like a little refactor because it makes the code a bit easier to reason about, once this issue is fixed I think it would be good

packages/nodes-base/nodes/RabbitMQ/GenericFunctions.ts
BerniWittmann: Great catch, added a test and fix for this issue, to prevent a regression of this behavior 🙇‍♂️

packages/nodes-base/nodes/RabbitMQ/GenericFunctions.ts
```diff
@@ -218,17 +218,42 @@ export async function handleMessage(
...
+			const hookRace: Promise<RaceResult> = responsePromiseHook.promise.then(() => ({
+				kind: 'hook',
+			}));
+			const runRace: Promise<RaceResult> = responsePromise.promise.then((data) => ({
+				kind: 'run',
+				data,
+			}));
+			const first = await Promise.race([hookRace, runRace]);
+
+			if (first.kind === 'run' && first.data?.data?.resultData?.error) {
```
RomanDavydchuk: Looks like my initial thought was wrong ☹️ 
Tested manually both happy path and a case when the workflow crashes before delete node, the happy path works now but the initial issue came back

Workflow I used to test:
```json
{
  "nodes": [
    {
      "parameters": {
        "queue": "n8n-dev",
        "options": {
          "acknowledge": "laterMessageNode"
        }
      },
      "type": "n8n-nodes-base.rabbitmqTrigger",
      "typeVersion": 1,
      "position": [
        0,
        0
      ],
      "id": "1f4007fe-a69a-42c5-88d1-6b811f854c75",
      "name": "RabbitMQ Trigger",
      "credentials": {
        "rabbitmq": {
          "id": "XwRkoqtkG9W3JpgX",
          "name": "RabbitMQ (localhost)"
        }
      }
    },
    {
      "parameters": {
        "amount": 10
      },
      "type": "n8n-nodes-base.wait",
      "typeVersion": 1.1,
      "position": [
        208,
        0
      ],
      "id": "5d570c15-5d66-4a71-9be7-ded7e4ec5eec",
      "name": "Wait",
      "webhookId": "6ebf43ad-988f-4107-abad-60249cc643b8"
    },
    {
      "parameters": {
        "operation": "deleteMessage"
      },
      "type": "n8n-nodes-base.rabbitmq",
      "typeVersion": 1.1,
      "position": [
        624,
        0
      ],
      "id": "ef3b245b-1ffd-4472-8019-a5eacedfcd39",
      "name": "RabbitMQ",
      "credentials": {
        "rabbitmq": {
          "id": "XwRkoqtkG9W3JpgX",
          "name": "RabbitMQ (localhost)"
        }
      }
    },
    {
      "parameters": {
        "jsCode": "throw {};"
      },
      "type": "n8n-nodes-base.code",
      "typeVersion": 2,
      "position": [
        416,
        0
      ],
      "id": "ed63e854-d65f-4380-b3f6-8d72200ab20c",
      "name": "Code in JavaScript"
    }
  ],
  "connections": {
    "RabbitMQ Trigger": {
      "main": [
        [
          {
            "node": "Wait",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "Wait": {
      "main": [
        [
          {
            "node": "Code in JavaScript",
            "type": "main",
            "index": 0
          }
        ]
      ]
    },
    "RabbitMQ": {
      "main": [
        []
      ]
    },
    "Code in JavaScript": {
      "main": [
        [
          {
            "node": "RabbitMQ",
            "type": "main",
            "index": 0
          }
        ]
      ]
    }
  },
  "pinData": {},
  "meta": {
    "templateCredsSetupCompleted": true,
    "instanceId": "935ffbc4ed6068e5224bb41b3fac79658dd867ba2639d61c1773533283b0e647"
  }
}
```

Stepping through the code with a debugger shows that in this case the `hookRace` is resolved first for some reason 🤷 

Image

`runRace` is still pending:
Image

To be honest, I'm not sure how to go around this ☹️

RomanDavydchuk (changes requested):
The changes you implemented are really close to what I imagined might work, but after testing, looks like it doesn't(

Left more details in the comment below

BerniWittmann:
Thanks @RomanDavydchuk you were right, but was able to fix it and improve the tests

RomanDavydchuk (approved):
LGTM and works well now 🎉

PR #28714, opened 2026-04-20
feat(core): Make SSO connection settings configurable via env vars

ireneea:
## Summary

Allows SAML and OIDC connection settings to be configured via environment variables. When env vars are set, the corresponding SSO settings become read-only in the UI and cannot be modified via the API.

- Unified OIDC and SAML instance settings loaders into a single SSO loader
- Added controller-level guards so protected fields cannot be updated when managed by env
- Frontend forms display a callout and become read-only when controlled by env vars

## Demo

[run through](https://www.loom.com/share/203daa5312094e8b8f713f54dc18c52d)

https://linear.app/n8n/issue/LIGO-382

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/@n8n/config/src/configs/instance-settings-loader.config.ts
```diff
@@ -75,4 +75,17 @@ export class InstanceSettingsLoaderConfig {
...
+	/** XML metadata string from the identity provider. */
+	@Env('N8N_SSO_SAML_METADATA')
+	samlMetadata: string = '';
+
+	/** URL to fetch SAML metadata from (mutually exclusive with metadata). */
+	@Env('N8N_SSO_SAML_METADATA_URL')
+	samlMetadataUrl: string = '';
+
+	@Env('N8N_SSO_SAML_LOGIN_ENABLED')
+	samlLoginEnabled: boolean = false;
```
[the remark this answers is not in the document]
ireneea: This is on purpose in order to have a parity with what can be set via the UI. This avoids exposing potential config that are not supported or experimental. 

The key idea here is that what users can do via the UI should be doable via the env vars (not more, not less)

packages/cli/src/instance-settings-loader/loaders/sso.instance-settings-loader.ts
```diff
@@ -0,0 +1,227 @@
...
+		authenticationContextClassReference: input.oidcAcrValues
+			? input.oidcAcrValues
+					.split(',')
+					.map((v) => v.trim())
+					.filter(Boolean)
+			: [],
+	}));
+
+@Service()
+export class SsoInstanceSettingsLoader {
```
konstantintieber, line 86: This class has gotten a bit large.
There is no case for configuring both OIDC and SAML on the same instance.
I see potential to make this code easier to maintain by having separtate SamlInstanceSettingsLoader and OidcInstanceSettingsLoader services.

ireneea: That's what I had initially but it made it a bit awkward to deal with validation of the fact that they should be mutually exclusive, and also detect when we should disable them or set the protocol back to `email`

ireneea: Basically `SamlInstanceSettingsLoader` had to now about `OidcInstanceSettingsLoader` and vice versa. And both loader were also updating the same settings

packages/cli/src/instance-settings-loader/loaders/sso.instance-settings-loader.ts
```diff
@@ -0,0 +1,227 @@
...
+
+import { PROVISIONING_PREFERENCES_DB_KEY } from '@/modules/provisioning.ee/constants';
+import { OIDC_PREFERENCES_DB_KEY } from '@/modules/sso-oidc/constants';
+import { SAML_PREFERENCES_DB_KEY } from '@/modules/sso-saml/constants';
+import {
+	getCurrentAuthenticationMethod,
+	setCurrentAuthenticationMethod,
+} from '@/sso.ee/sso-helpers';
+
+const PROVISIONING_MODES = ['disabled', 'instance_role', 'instance_and_project_roles'] as const;
```
geemanjs, line 18: Can we pluck these types out from anywhere that isn't the "instance settings loader"? 

I worry that they may be updated elsewhere but not in the loader.

geemanjs (approved):

packages/cli/src/instance-settings-loader/loaders/sso.instance-settings-loader.ts
konstantintieber: Could this be avoided by throwing an error when both OIDC and SAML env vars are set?
I saw in your video that your motivation for this was that in the UI, you can theoretically still have OIDC settings next to SAML, while they are unused.
But I think that doesn't cover a real world use case and we don't need to support having both SAML and OIDC set.

konstantintieber (approved):

packages/cli/src/instance-settings-loader/loaders/sso.instance-settings-loader.ts
ireneea: On it

ireneea: Actually will address it in the next PR just for the sake of getting this merged

PR #28715, opened 2026-04-20
feat(core): Add observability for the VM expression engine

ivov:
https://linear.app/n8n/issue/CAT-2276


ivov:
CI blocked by: https://github.com/n8n-io/tournament/pull/25

despairblue (dismissed):
Added some commits:
* added clarifying docs
* stricter config validation
* remove counters that are already included in the histogram
* pre-register all metrics with all labels to prevent race conditions in the future if somebody is adding a label that could lead to runtime errors or metrics being dropped if the labels don't match the registration
* rename the error missing sentinel to `none` which seems clearer when seen in grafana
* switch to seconds, which is standard for prometheus and in our codebase to
* honor the metrics prefix
* remove the metrics maps (which already live in the prom client)

I have another follow up that I left out of this PR to not blow it up, I was thinking about redoing the traces setup which is right now unused due to piggy backing on the metrics, so we may as well remove it or don't piggy back. Not sure yet.

Best to check commit by commit, let me know if you disagree with anything. 🙏🏾

ivov:
Can you reapprove? Fixed a test

despairblue (approved):

PR #28716, opened 2026-04-20
feat(core): Bootstrap encryption key set from environment

guillaumejacquart:
## Summary

On first startup with `N8N_ENCRYPTION_KEY_IDENTIFICATION_ENABLED=true`, seed the legacy `instanceSettings.encryptionKey` into the `deployment_key` table as an active `aes-256-cbc` row. This gives T10 call sites a known `encryptionKeyId` to route decryption through; the NULL sentinel becomes purely historical after first run. The bootstrap runs via the existing `encryption-key-manager` module's `init()` hook (post-migrations, pre-request-serving), is leader-only for multi-main safety, and idempotent — a second run finds the existing active key and returns.

To test: set `N8N_ENCRYPTION_KEY_IDENTIFICATION_ENABLED=true`, start n8n against a fresh DB, verify one row exists in `deployment_key` with `type='data_encryption'`, `algorithm='aes-256-cbc'`, `status='active'`, and `value` matching the instance's encryption key. Restart and confirm no duplicate row is created. With the flag unset, no row is inserted.

Covered by unit tests (flag-off, non-leader, seed, idempotency, invariant-error rethrow) and an integration test against a real SQLite DB (seed + second-run no-op).

https://linear.app/n8n/issue/IAM-491

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/modules/encryption-key-manager/encryption-bootstrap.service.ts
```diff
@@ -0,0 +1,39 @@
...
+		private readonly instanceSettings: InstanceSettings,
+		private readonly logger: Logger,
+	) {
+		this.logger = this.logger.scoped('encryption-key-manager');
+	}
+
+	async run(): Promise<void> {
+		if (process.env.N8N_ENCRYPTION_KEY_IDENTIFICATION_ENABLED !== 'true') return;
+
+		if (!this.instanceSettings.isLeader) return;
```
BGZStephen: On a phased rollout there is a chance the leader may change `after` this `run` has already happened, there is a helper for this situation, called insertOrIgnore, it is used within `packages/core/src/instance-settings/instance-settings.ts` to handle race conditions on init, might be worth trying to reproduce that pattern here

packages/cli/src/modules/encryption-key-manager/encryption-bootstrap.service.ts
```diff
@@ -0,0 +1,39 @@
...
+	constructor(
+		private readonly keyManager: KeyManagerService,
+		private readonly instanceSettings: InstanceSettings,
+		private readonly logger: Logger,
+	) {
+		this.logger = this.logger.scoped('encryption-key-manager');
+	}
+
+	async run(): Promise<void> {
+		if (process.env.N8N_ENCRYPTION_KEY_IDENTIFICATION_ENABLED !== 'true') return;
```
BGZStephen: I'm not entirely sure we need a feature flag for this, it's a pretty safe operation as it just populates the DB with the legacy key

packages/cli/src/modules/encryption-key-manager/encryption-bootstrap.service.ts
guillaumejacquart: Good call — switched to the findActiveByType → insertOrIgnore pattern from instance-settings.ts (62fd7e6). Dropped the isLeader guard since the partial unique index now handles concurrency, and added a concurrent-run integration test (Promise.all([run(), run(), run()]) asserts exactly one active row). The logic lives on KeyManagerService.bootstrapLegacyKey to keep repo access in the service layer.

packages/cli/src/modules/encryption-key-manager/encryption-bootstrap.service.ts
guillaumejacquart: Agreed — dropped it in 62fd7e6. Since the bootstrap is now race-safe and idempotent, there's nothing risky to gate. One thing worth flagging: IAM-491 explicitly asks for the flag as part of the staged rollout plan. If we're dropping it here, I'll update the ticket to note that the rollout strategy changed.

BGZStephen: Yeah, I may have been overzealous on feature flagging, the only thing I think we will be, is the APIs and front-end to gate the feature from the UI

BGZStephen (approved):

packages/cli/src/modules/encryption-key-manager/encryption-bootstrap.service.ts
guillaumejacquart: oh my, sorry i let claude reply on the PR (hence the emdash xD). Agreed for the flag though :)

packages/cli/src/modules/encryption-key-manager/encryption-bootstrap.service.ts
```diff
@@ -0,0 +1,21 @@
...
+	constructor(
+		private readonly keyManager: KeyManagerService,
+		private readonly instanceSettings: InstanceSettings,
+		private readonly logger: Logger,
+	) {
+		this.logger = this.logger.scoped('encryption-key-manager');
+	}
+
+	async run(): Promise<void> {
+		await this.keyManager.bootstrapLegacyKey(this.instanceSettings.encryptionKey);
```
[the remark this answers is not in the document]
guillaumejacquart: That's on purpose, we decided to remove the flagging

PR #28722, opened 2026-04-20
fix(core): Shift existing rules on create to avoid order conflict

cstuncsik:
## Summary

Fixes a 409 conflict when creating a role mapping rule that lands on an already-occupied `order`. The backend now splices the new rule into the requested position and atomically renumbers the sequence using the existing two-phase `applyOrder` transaction. Omitting `order` appends the rule; out-of-range values clamp to the end of the list.

### How to verify in the browser

**Prerequisites:** n8n running locally with SAML or OIDC provisioning licensed (`N8N_LICENSE_ACTIVATION_KEY` or dev license), and an SSO config that uses **Rules in n8n** for role assignment.

1. Sign in as an owner and open **Settings → SSO → Provisioning**.
2. Under **Instance role rules**, click **Add rule**, set an expression (e.g. `$claims.admin === true`) and a role, then save the SSO config.
3. Click **Add rule** again — add another instance rule **on top of** the one you just created.
4. Save the SSO config.
   - **Before this PR:** red error toast *"A role mapping rule already exists with type 'instance' and order 0. Use a different order value."*
   - **After this PR:** the rule is saved. Reload the page and confirm both rules are listed in the intended order with sequential priorities (`1.`, `2.`).
5. Drag a new local rule above an existing one, save, reload — the dragged rule should land at the top and the other rules should shift down without errors.
6. Delete the middle rule, add a new one, save — no conflict, priorities remain contiguous.
7. Repeat steps 2–4 for **Project role rules** (if enabled) — same behaviour.

https://linear.app/n8n/issue/IAM-560

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/@n8n/api-types/src/dto/roles/__tests__/create-role-mapping-rule.dto.test.ts
+++ b/packages/@n8n/api-types/src/dto/roles/__tests__/create-role-mapping-rule.dto.test.ts
@@ -1,0 +1,39 @@
+import { CreateRoleMappingRuleDto } from '../create-role-mapping-rule.dto';
+
+const VALID_BODY = {
+	expression: 'claims.admin === true',
+	role: 'global:member',
+	type: 'instance' as const,
+};
+
+describe('CreateRoleMappingRuleDto', () => {
+	describe('Valid requests', () => {
+		test.each([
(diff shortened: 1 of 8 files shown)


afitzek (approved):
👍

PR #28728, opened 2026-04-20
feat(Slack Node): Allow users to configure OAuth2 scopes

Joffcom:
## Summary

Replaces the previously hardcoded OAuth2 `user_scope` list in the Slack credential with a user-configurable field. By default, the credential behaves identically to before — the same full set of user scopes is requested automatically. Users who need a restricted or custom scope set can enable the **Custom Scopes** toggle to expose and edit the `User Scope` field before connecting.

Changes:
- Adds `customScopes` boolean toggle (default: `false`) to the Slack OAuth2 credential
- Adds `userScope` string field (shown when `customScopes` is `true`) pre-populated with the default scope list
- Adds an explicit empty `scope` hidden field to prevent bot-scope inheritance from the base credential
- Updates `authQueryParameters` to dynamically use the custom `userScope` when enabled, or fall back to the default scope list otherwise
- Adds unit tests for the credential properties

To test:
1. Create a new Slack OAuth2 credential — verify it works unchanged with default scopes
2. Enable **Custom Scopes**, remove a scope, re-authorize — verify only the specified scopes are requested

https://linear.app/n8n/issue/NODE-3815

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

> Tested locally on 2.16 without this then ran with this with various reconnect options to ensure this doesn't break anything by removing the bot scope under the scopes field. We removed this as it can't be used in the node to send messages as the bot user.


packages/nodes-base/credentials/test/SlackOAuth2Api.credentials.test.ts
```diff
@@ -0,0 +1,67 @@
+import { SlackOAuth2Api } from '../SlackOAuth2Api.credentials';
+
+describe('SlackOAuth2Api Credential', () => {
+	const credential = new SlackOAuth2Api();
+
+	const defaultUserScopes = [
```
DawidMyslak: This duplicates the scope list from the source file. If we ever add or remove a scope, the test could silently drift. Would it be worth exporting `userScopes` from the credential and importing it here instead?

packages/nodes-base/credentials/SlackOAuth2Api.credentials.ts
```diff
@@ -54,18 +54,49 @@ export class SlackOAuth2Api implements ICredentialType {
...
 			type: 'hidden',
-			default: 'chat:write',
+			default: '',
+		},
+		{
+			displayName: 'Custom Scopes',
+			name: 'customScopes',
+			type: 'boolean',
+			default: false,
+			description: 'Define custom OAuth2 scopes instead of using the defaults',
```
DawidMyslak: Nit: Discord and Teams both use "Define custom scopes" - should we simplify it?

Image

DawidMyslak (changes requested):
Looks good overall, 2 minor comments 🙏

packages/nodes-base/credentials/test/SlackOAuth2Api.credentials.test.ts
Joffcom: Solid point and I have just fixed this.

packages/nodes-base/credentials/SlackOAuth2Api.credentials.ts
Joffcom: That is fair, I forgot I made the skill and I should have used it. Updated.

DawidMyslak (approved):

PR #28737, opened 2026-04-20
fix(core): Restore credential data deep copy in overwrites

Joffcom:
## Summary

Restores the deep copy behavior for credential data in `CredentialsOverwrites.applyOverwrite()`.

PR #28573 changed `CredentialsOverwrites.get()` to return `undefined` instead of `{}` when no credential overwrites are configured. This caused `applyOverwrite()` to early-return the **original credential data reference** instead of a `deepCopy()`.

In multi-main queue mode, multiple workflow executions run concurrently on the same process. The `getDecrypted()` method in `CredentialsHelper` has an `await` point at `dynamicCredentialsProxy.resolveIfNeeded()`. With the original reference being shared across concurrent requests, one execution could observe a partially-mutated credential object from another — resulting in an empty or corrupted token being passed to downstream API calls (e.g. Slack's `not_authed` error).

**Fix:** Change `if (entries.length === 0) return undefined` → `if (entries.length === 0) return {}`.

This restores the 2.17.2 path where `applyOverwrite` always calls `deepCopy(data)`, giving each concurrent execution its own isolated copy. The `skipTypes` optional chaining safety fix from PR #28573 is preserved.

**To test:**
1. Deploy to a multi-main + Postgres + queue mode instance
2. Run a workflow using Slack API credential — should succeed without `not_authed`
3. Verify credential test still shows green for a valid Slack token

https://linear.app/n8n/issue/NODE-4869

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI


Matsuuu (approved):
The reasoning seems reasonable. Let's try this out tomorrow with a patch release

BerniWittmann (approved):

Joffcom:
Closing as doesn't appear to be needed.

Issue #26356, opened 2026-02-27
AI Agent fails using tools when one of connected Tool is Deactivated


tarik23:
### Bug Description

When using an AI Agent node (chat trigger or not) with the OpenRouter Chat Model (can't test on other types of chat model nodes), if two or more Tool nodes are connected and one of them is set to Deactivated, the agent fails execution when tries to use one of active tool.

The agent does not complete the task and stops with the error:

```Max iterations (10) reached. The agent could not complete the task within the allowed number of iterations.```

This issue occurs even when no Memory node is connected to the AI Agent.

<img width="1333" height="887" alt="Image" src="https://github.com/user-attachments/assets/0109b93a-78d1-4aeb-bef8-7f886406f154" />

<img width="1858" height="959" alt="Image" src="https://github.com/user-attachments/assets/44233bb5-db65-473a-b29a-21c0fd9322d2" />

Spent two weeks to find out that "deactivated tool" is the problem...

### To Reproduce

1. Create Chat trigger and an AI Agent node.
2. Connect two Tool nodes to it (tested on tools as Subworkflows)
3. Use OpenRouter Chat Model.
4. Deactivate one of the Tool nodes. (using "D" key for example)
5. Write message to chat where you ask to use tool (that is now active)
6. You get error on AI agent node "Max iterations (10) reached. The agent could not complete the task within the allowed number of iterations."

### Expected behavior

Agent should use tool and end task successful

### Debug Info

# Debug info

## core

- n8nVersion: 2.9.4
- platform: docker (self-hosted)
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: postgres
- executionMode: regular
- concurrency: -1
- license: enterprise (production)
- consumerId: 9078bc9f-30a7-4edc-b0e4-8d59f3e03261

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/144.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2026-02-27T15:19:44.870Z

### Operating System

Debian

### n8n Version

2.9.4

### Node.js Version

24.13.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

Dom-Tursini:
[...]

PR #28743, opened 2026-04-20
fix: Filter out disabled AI Agent tool parents when mapping connected tools

Dom-Tursini:
## Summary

This fixes a tool/source-node desync in getConnectedTools() for AI Agent tool connections when disabled AI tool nodes remain connected.

Previously, `toolkitConnections` excluded disabled AI tool nodes, while `parentNodes` still included them. Because the helper correlates both arrays by index, disabling any AI tool node caused later tools to receive the wrong `sourceNodeName`.

This change filters disabled nodes out of `parentNodes` so both arrays use the same active-node set and stay index-aligned.

A regression test was added to cover the case where a disabled tool node appears between active tool connections.

How to test:
1. Create an AI Agent with multiple connected AI tool nodes.
2. Disable one tool node that is not the last in the connection order.
3. On the affected code path, later tools are mapped to the wrong `sourceNodeName`.
4. Verify with this change that later tools remain mapped to the correct source node.
5. Run `pnpm test -- utils/tests/helpers.test.ts --runInBand` in `packages/@n8n/nodes-langchain`.

#26356 

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/nodes-langchain/utils/tests/helpers.test.ts
+++ b/packages/@n8n/nodes-langchain/utils/tests/helpers.test.ts
@@ -322,6 +322,39 @@ describe('getConnectedTools', () => {
 		});
 	});
 
+	it('should ignore disabled parent tool nodes when mapping source node metadata', async () => {
+		const mockParentNodes = [
+			{ name: 'RegularTool', type: 'test', typeVersion: 1, disabled: false },
+			{ name: 'Disabled Tool', type: 'test', typeVersion: 1, disabled: true },
+			{ name: 'MCP Client Tool', type: 'test', typeVersion: 1, disabled: false },
+		];
+		const mockTools = [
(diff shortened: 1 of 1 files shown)


Dom-Tursini:
Closing this as the fix has been merged via #29460  by @schrothbn . I’m glad to see the index-desync I identified here was resolved using the same logic. Hopefully, the regression tests I included here can still be useful for the core team.

Dom-Tursini:
Leaving this here for provenance/context.

This PR originally contained the same fix that later shipped via #29460 filtering disabled AI tool nodes out of parentNodes so the parentNodes/toolkitConnections indexes stay aligned.

After rebasing onto master, the diff no longer shows that line because the same change had already been merged via #29460 .

I understand this may have been handled through internal tracking, so I just wanted to record the connection between the two PRs for future context.

PR #28754, opened 2026-04-21
refactor(ai-builder): Streamline instance-ai prompt surface (no-changelog)

OlegIvaniv:
## Summary

Audited every persistent prompt in `@n8n/instance-ai` against Anthropic's current Claude 4.x prompting guidance and found the usual suspects: the same "output discipline" block copy-pasted across six sub-agent personas, a ~200 line Web App SPA example sitting in every builder turn, three duplicate statements of the plan/create-tasks decision rule, heavy `NEVER`/`MUST`/`CRITICAL` density that Sonnet 4.6+ tends to overtrigger on, and a tool-mode builder that hadn't seen a production trace in a week. This PR works through that list.

## What changed

The work landed as small focused commits so each piece can be reverted independently if a regression shows up in traces:

- Extracted the shared sub-agent snippets (`SUBAGENT_OUTPUT_CONTRACT`, `UNTRUSTED_CONTENT_DOCTRINE`, `ASK_USER_FALLBACK`) into `shared-prompts.ts` and imported them where needed, instead of six near-duplicates.
- Deduplicated the plan/create-tasks rule in the orchestrator system prompt, it's stated once now, with the runtime guard in `plan.tool.ts` doing the actual enforcement.
- Trimmed tension markers across personas where they weren't actually load-bearing, per Anthropic's Opus 4.5/4.6 guidance. Kept the strong language on untrusted-content rules, webhook path discipline, credential mocking, places where trace data shows it's still needed.
- Moved the Web App SPA example (HTML + multi-route SDK code) out of the builder prompt and behind `nodes(action="suggested")`, so it only loads when the task actually involves a web app instead of on every Slack notifier build.
- Retired the tool-mode workflow builder. Trace scan showed zero production hits over the 7-day window. Removed `build-workflow.tool.ts`, the dual-mode scaffolding, and the sandbox override banner, sandbox is the only path now.
- Follow-up hardening after a code review surfaced two real issues in the streamlined surface:
  - The deleted `build-workflow.tool.ts` held the only enforcement of the `createWorkflow` / `updateWorkflow` permission modes. Admins setting either to `'blocked'` were silently losing that protection, since the schema still exposes both with default `require_approval`. Re-added the guard at the top of `submit-workflow.tool.ts`.
  - The builder was stripping credential ids from pre-loaded workflow code via regex and letting `resolveCredentials` re-resolve them on save. That resolver does last-write-wins by credential type, so a user with two Slack credentials could have saves silently rewired to the wrong one. Changed the regex to emit raw `{ id, name }` objects, which preserves the original binding.
  - The `create-tasks` replan guard was reading its signal by substring-matching raw user chat text, which a user could spoof by pasting the literal marker. Added a trusted `isReplanFollowUp` flag on `OrchestrationContext`, set by the replan pipeline, and read by the guard instead. Now aligned with the `UNTRUSTED_CONTENT_DOCTRINE` the same PR introduced.
  - Fixed a `plan.tool.test.ts` `afterEach` that assigned `undefined` back to `process.env` (Node stringifies that to the literal `"undefined"` and it leaks into later tests).
- Final sweep: dropped the tool-mode leftovers still sitting in `src/workflow-builder/`. Deleted `patch-code.ts` (`applyPatches`), `sdk-prompt-sections.ts`, and the string-manipulation helpers in `extract-code.ts`, all tool-mode-only. Trimmed `parse-validate.ts` to just `partitionWarnings` (the only function still consumed by `submit-workflow`), pruned the barrel and types to match. About 1000 lines of dead code gone in a single commit.

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
aalises, line 1: There's some leftover from the tool mode retirement, seems to be an empty-section (the stray `// Patch-mode builder prompt` header with nothing underneath)

aalises:
On the `tools.md` docs I would maybe cleanup old references. Still describes tool-mode, `build-workflow` tool, `patches` param, and `parseAndValidate()`

aalises (dismissed):
some e2e flakyness? and conflicts. LMK once you want a re-approval

aalises (approved):

PR #28760, opened 2026-04-21
feat(core): Add public API endpoints for folder management

sandra0503:
## Summary

Add `POST /projects/{projectId}/folders` and `GET /projects/{projectId}/folders`
to the public API, enabling folder creation and listing through API keys.

- Reuse `FolderService` business logic from the internal API to avoid duplication
- Add shared `resolveProjectId` middleware for project validation and authorization
- Register `folder` as an API key resource with scopes for owner and member roles
- Map `FolderNotFoundError` to 404 in the handler (it extends `OperationalError`,
  not `ResponseError`, so it would otherwise surface as 500)

## Review notes

- The `resolveProjectId` helper follows the same 404/403 pattern as `data-tables`
- `skip`/`take` defaults are handled by `ListFolderQueryDto`, not the OpenAPI spec,
  to avoid type coercion issues from the OpenAPI validator

Implements [LIGO-438](https://linear.app/n8n/issue/LIGO-438/add-missing-create-and-list-folders-endpoints)

## Screenshots

Screenshot 2026-04-21 at 09 28 50
Screenshot 2026-04-21 at 09 29 05

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/public-api/v1/handlers/folders/spec/paths/folders.yml
```diff
@@ -0,0 +1,110 @@
...
+        application/json:
+          schema:
+            $ref: '../schemas/folder.yml'
+    '400':
+      $ref: '../../../../shared/spec/responses/badRequest.yml'
+    '401':
+      $ref: '../../../../shared/spec/responses/unauthorized.yml'
+    '403':
+      $ref: '../../../../shared/spec/responses/forbidden.yml'
+    '404':
```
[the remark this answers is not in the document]
sandra0503: https://github.com/n8n-io/n8n/pull/28760/changes#diff-5b2fc8e8970802009b0d5d14d84cd673f7ced18817f521bc5faed4b13d9a4e66R109

packages/cli/src/public-api/v1/shared/middlewares/global.middleware.ts
```diff
@@ -155,3 +157,19 @@ export const isLicensed = (feature: BooleanLicenseFeature) => {
 		return res.status(403).json({ message: new FeatureNotLicensedError(feature).message });
 	};
 };
+
+export async function resolveProjectId(
```
alielkhateeb: This function is only doing validation/permission checks, it is also not used as a middleware. Is it intended to be a middleware? Or just a utility for validation?

packages/cli/src/public-api/v1/shared/middlewares/global.middleware.ts
```diff
@@ -155,3 +157,19 @@ export const isLicensed = (feature: BooleanLicenseFeature) => {
...
+	scopes: Scope[],
+): Promise<string> {
+	const exists = await Container.get(ProjectRepository).findOneBy({ id: projectId });
+	if (!exists) {
+		throw new NotFoundError(`Project with ID "${projectId}" not found`);
+	}
+
+	const project = await Container.get(ProjectService).getProjectWithScope(user, projectId, scopes);
+	if (!project) throw new ForbiddenError();
+
```
alielkhateeb: Should we return 404 here instead of 403? Right now we reveal that the project exists but the user lacks access. If we want to avoid leaking resource existence in the public API, we should return 404 in both cases. What do you think?
Also, what do we do in other cases? :joy:

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
```diff
@@ -0,0 +1,57 @@
...
+	isLicensed,
+	resolveProjectId,
+} from '../../shared/middlewares/global.middleware';
+
+export = {
+	createFolder: [
+		isLicensed('feat:folders'),
+		apiKeyHasScopeWithGlobalScopeFallback({ scope: 'folder:create' }),
+		async (req: AuthenticatedRequest<{ projectId: string }>, res: Response) => {
+			const projectId = await resolveProjectId(req.user, req.params.projectId, ['folder:create']);
```
alielkhateeb: Continuing on my other comment, we currently rely on upstream checks (and DB FK) for projectId validity. Should we add an explicit existence check in `createFolder` instead, so this service throws a clear domain error (`ProjectNotFoundError` for instance) if it’s called from a new path later? That way we won't even need `resolveProjectId` at all, and we'll catch `ProjectNotFoundError` here and throw `NotFoundError` HTTP response accordingly. What do you think?

packages/cli/src/public-api/v1/shared/middlewares/global.middleware.ts
sandra0503: In the other places we return 403 as well. I think it's okay to leak the project existence because it's still an authenticated user who is a member of the instance. What do you think?

alielkhateeb: Alright then if we are consistently returning 403 in other places in this case then it's better to keep it consistent I agree

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
sandra0503: I was thinking of doing that, but in this case it might be better to keep folder service clean. The REST Api controller is already checking permissions which would mean we would check it twice on this path

`packages/cli/src/controllers/folder.controller.ts`:
```
@ProjectScope('folder:create')
@Licensed('feat:folders')
async createFolder() {
...
}
```

alielkhateeb: Yes agreed if we will keep the 404 and 403 distinct then lets keep the checks outside and leave the folder service clean. The only open thing now is whether you want this check to stay as a middleware or switch it to a utility/validation helper

alielkhateeb: Problem with relying on the FK as a guard is that eventually it becomes a bit less user friendly, but that's a topic not related to this PR anyway 😂

packages/cli/src/public-api/v1/shared/middlewares/global.middleware.ts
sandra0503: It's just a utility, I can move it to a better place

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
sandra0503: Right now it's using the new utility `resolveProjectId`? I suggest to keep that (but move the utility to another file based on your comment above)

packages/cli/src/public-api/v1/handlers/folders/__tests__/folders.handler.test.ts
```diff
@@ -0,0 +1,248 @@
...
+		mockFolderService = mockInstance(FolderService);
+		mockProjectRepository = mockInstance(ProjectRepository);
+		mockProjectService = mockInstance(ProjectService);
+
+		jest.spyOn(Container, 'get').mockImplementation((serviceClass) => {
+			if (serviceClass === FolderService) return mockFolderService as any;
+			if (serviceClass === ProjectRepository) return mockProjectRepository as any;
+			if (serviceClass === ProjectService) return mockProjectService as any;
+			return {} as any;
+		});
```
alielkhateeb: I usually try to avoid using type-casting and especially avoid using `any`. In this case, I think it's safe to remove `as any` I don't see why we have them, am I missing something?

packages/cli/src/public-api/v1/handlers/folders/__tests__/folders.handler.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+		mockProjectRepository.findOneBy.mockResolvedValue({ id: projectId } as never);
```
alielkhateeb: TL;DR I would rather properly mock the project object than using `as never`.

Reason is using `as never` defeats type-safety, it technically silences the compiler from detecting any type issues. My mental model usually when I see a TS error is to fix that error properly and not cast/escape it, escaping it that way makes refactoring/maintaining that code later super difficult.

sandra0503: Yes, actually we have a nice `mockProject` function, I used that now

packages/cli/src/public-api/v1/handlers/folders/__tests__/folders.handler.test.ts
sandra0503: Not needed indeed, I removed it here as well as in other places of the file where possible

packages/cli/src/public-api/v1/handlers/folders/__tests__/folders.handler.test.ts
```diff
@@ -0,0 +1,249 @@
...
+import { FolderService } from '@/services/folder.service';
+import { ProjectService } from '@/services/project.service.ee';
+import * as middlewares from '@/public-api/v1/shared/middlewares/global.middleware';
+import { mockProject } from '@test/mock-objects';
+
+const mockMiddleware = jest.fn(async (_req: unknown, _res: unknown, next: unknown) =>
+	(next as () => void)(),
+) as unknown as middlewares.ScopeTaggedMiddleware;
+jest.spyOn(middlewares, 'isLicensed').mockReturnValue(mockMiddleware as any);
+jest.spyOn(middlewares, 'apiKeyHasScopeWithGlobalScopeFallback').mockReturnValue(mockMiddleware);
```
alielkhateeb, line 20: I noticed this is the way we mockMiddleware in other files too, but I don't think it's good to have all these type-casts and abuse of `any` type. I will create a tech debt issue for this

packages/cli/src/public-api/v1/shared/middlewares/global.middleware.ts
```diff
@@ -155,3 +157,19 @@ export const isLicensed = (feature: BooleanLicenseFeature) => {
...
+	return project.id;
```
alielkhateeb: More of a nit: this is not really "resolving" project ID it looks like `validateProjectAccess` and I believe it should not return anything, we already have project ID through the params everywhere we consume this, so making it an only validation method sounds cleaner to me.

Only case I would keep "resolve" and the return is if we expect this to actually resolve personal project IDs but I doubt we will do that ever, we ALWAYS receive project ID in params, don't we?

alielkhateeb (approved):
🚀 👏 Amazing work 🫶

pedrokehl:
@sandra0503 With this feature, users can't know what's inside each folder, right? Could this be added?

PR #28761, opened 2026-04-21
feat(editor): Add dev-panel for DOM-annotated feedback prompts (no-changelog)

Tuukkaa:
## Summary

Adds a dev-only overlay to the editor-ui that lets developers pick DOM
elements (single click, shift-click multi-select, or drag-to-select), attach
a free-form prompt to each, and copy the whole batch as formatted markdown
to paste into Claude Code (or any other tool).

### Highlights

- **Pick + annotate + copy** workflow: numbered markers overlay each
  annotated element; the copy button emits a markdown block with page path,
  viewport, DOM path, source file/line (via
  [`vite-plugin-vue-inspector`](https://github.com/webfansplz/vite-plugin-vue-inspector)),
  Vue component name, test id, and the user's prompt.
- **Persistent annotations**: stored in `localStorage` with a 7-day TTL so
  they survive refresh and HMR; re-resolved against the current DOM when
  navigation or mutations shift things around.
- **Feature flag overrides**: a flag button in the toolbar opens a panel that
  lists all PostHog flags discovered in the SDK + persisted blobs, lets you
  set/clear overrides (boolean or variant, with custom-value input), and
  writes them to the same `N8N_EXPERIMENT_OVERRIDES` localStorage key the
  Playwright suite already uses. Reload-and-apply button included.
- **Hidden by default**: toggle with <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>D</kbd>
  (uses `event.ctrlKey` + `event.key` so it works on macOS and on
  non-QWERTY layouts like Colemak). Visibility is remembered across reloads.
- **Dev-only**: mounted behind `import.meta.env.DEV`, so Vite tree-shakes it
  out of production bundles. No backend, no extra process.

### Preview
image
image
image

### Why

Speeds up the "I noticed something on this screen, let me describe it
precisely to an AI" loop. Copying N annotations at once beats round-tripping
a single screenshot, and every annotation carries enough source metadata
that an AI assistant can usually jump straight to the right file.

The flag override panel folds in a workflow that previously lived in a
bookmarklet, so toggling experiments no longer requires editing
localStorage by hand or keeping a separate snippet around.

### Reviewer notes

- No production code paths are affected. Nothing new in \`packages/cli\`,
  \`nodes-base\`, etc.

N/A — internal dev tooling.

## Review / Merge checklist

- [ ] Tests included. _(manual only for now — the overlay is behind \`import.meta.env.DEV\`)_
- [ ] PR Labeled with \`Backport to Beta\`, \`Backport to Stable\`, or \`Backport to v1\` — N/A


packages/frontend/editor-ui/src/app/dev/dev-panel/DevPanel.vue
```diff
@@ -0,0 +1,889 @@
...
+	loadForPath(currentPath.value);
+	domObserver = new MutationObserver(scheduleReresolve);
+	domObserver.observe(document.body, { childList: true, subtree: true });
+});
+
+onUnmounted(() => {
+	window.removeEventListener('scroll', scheduleBump, true);
+	window.removeEventListener('resize', scheduleBump);
+	window.removeEventListener('keyup', handleShiftKeyUp, true);
+	window.removeEventListener('keydown', handleToggleShortcut, true);
```
aalises: Panel is hidden by default, but the `MutationObserver` on `document.body`, scroll/resize listeners, and history patch are installed unconditionally on mount. 

For devs who never open the panel, this is constant overhead on every app-wide DOM mutation. Can we gate these on `visible` (attach on togglin on the panel, detach on toggle-off)?

packages/frontend/editor-ui/src/app/dev/dev-panel/DevPanel.vue
```diff
@@ -0,0 +1,889 @@
...
+		pendingMulti.value = [];
+		editingId.value = null;
+	}
+}
+
+function updatePath() {
+	currentPath.value = window.location.pathname;
+}
+
+let restoreHistory: (() => void) | null = null;
```
aalises: Ctrl+Shift+D iirc is "Bookmark all tabs" in Chrome/Edge on Windows/Linux, and the handler doesn't check the event target, so pressing it while typing in an input silently eats the keystroke. Could we either pick a less-claimed chord, or early-return when `event.target` is editable?

packages/frontend/editor-ui/src/app/dev/dev-panel/DevPanel.vue
```diff
@@ -0,0 +1,889 @@
...
+		history.pushState = originalPush;
+		history.replaceState = originalReplace;
+	};
+}
+
+function loadForPath(path: string) {
+	const stored = loadAnnotations(path);
+	annotations.value = stored.map((a) => ({
+		...a,
+		elements: a.contexts
```
aalises: I wouldn't monkey-patch `history.pushState`/`replaceState` globally which is fragile. Since good ole' vue-router is already in the app, using `useRouter().afterEach(...)` would give us the same without touching the global.

packages/frontend/editor-ui/src/app/dev/dev-panel/DevPanel.vue
Tuukkaa: Done in fc47feb56c — observers (scroll, resize, keyup, MutationObserver, router subscription) are now attached lazily inside an `activateObservers` helper that runs when `visible` flips true and tears down on toggle-off. Only the Ctrl+Shift+D `keydown` listener stays mounted unconditionally so the toggle still works.

packages/frontend/editor-ui/src/app/dev/dev-panel/DevPanel.vue
Tuukkaa: Fixed in fc47feb56c — added an `isEditableTarget` check that early-returns when the event target is an `<input>`, `<textarea>`, `<select>`, or anything `contenteditable`, so the chord no longer eats keystrokes during typing. Keeping Ctrl+Shift+D for now since the panel is dev-only and the editable check covers the main hazard.

packages/frontend/editor-ui/src/app/dev/dev-panel/DevPanel.vue
Tuukkaa: Fixed in fc47feb56c — dropped the `pushState`/`replaceState` monkey-patch and now subscribe via `router.afterEach` (importing the router singleton from `@/app/router`). `popstate` is also no longer needed since the router already handles browser back/forward.

aalises:
The flag panel should seed from `window.featureFlags.getAll()` before falling back to PostHog/localStorage; otherwise server-evaluated flags can be missing or the panel can appear empty in normal dev sessions.

I’d put this in `packages/frontend/editor-ui/src/app/dev/dev-panel/useFeatureFlags.ts`, then merge it into `refresh()`:

```suggestion
function readEvaluatedFlags(): Record<string, FlagValue> {
	try {
		return window.featureFlags?.getAll?.() ?? {};
	} catch {
		return {};
	}
}
```

```ts
const evaluated = readEvaluatedFlags();
const persisted = readPersistedFlags();
const sdk = readSdkFlags();
const ph: Record<string, FlagValue> = { ...persisted, ...sdk, ...evaluated };
```

Tuukkaa:
Good catch — fixed in b2ff9aa216. `readEvaluatedFlags()` reads `window.featureFlags.getAll()` and is merged with highest priority over the SDK / persisted-blob fallbacks, so server-evaluated flags now show up even when the PostHog client hasn't populated localStorage yet.

aalises (approved):

PR #28769, opened 2026-04-21
fix: Stop persisting client id in session storage to fix duplicate tab problem (no-changelog)

dariacodes:
## Summary

This PR reverts the change introduced in https://github.com/n8n-io/n8n/pull/25646 which assumed it would be nice to persist client id on page refresh and missed the duplicate tab problem where this causes issues with sending and receiving push events. Full info in the Linear ticket.

How to test: duplicate a tab and notice that push connection is broken.

Closes [ADO-5095](https://linear.app/n8n/issue/ADO-5095/bug-duplicate-tabs-share-the-same-pushref-causing-missed-push-updates)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/@n8n/stores/src/useRootStore.ts
+++ b/packages/frontend/@n8n/stores/src/useRootStore.ts
@@ -35,18 +35,6 @@ export type RootStoreState = {
 };
 
 export const useRootStore = defineStore(STORES.ROOT, () => {
-	// Generate or retrieve client ID from sessionStorage
-	const getClientId = (): string => {
-		const storageKey = 'n8n-client-id';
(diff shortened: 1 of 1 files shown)


MiloradFilipovic (approved):
Seems to be working fine. What about the original issue that we had? Will this revert bring it back? I am not able to see any connection issues when duplicating tabs but just want to double check.

dariacodes:
@MiloradFilipovic I couldn't find any original issue that actually needed this change. I think I thought it would be nice to keep the id and was worried about how updates would work, but they seem to work correctly after this fix, plus fixing the issue I mentioned, which I didn't originally find when introducing this change.

PR #28770, opened 2026-04-21
fix(editor): Fix sub-workflow folder placement and connection loss

### LINKED ISSUE #28751: [Bug] Convert to sub-workflow: child workflow not placed in parent folder and internal node connections lost (regression vs 2.6.3)
### Bug Description

After upgrading from **n8n 2.6.3** to **2.16.1**, using **Convert to sub-workflow** on a valid node selection produces a new workflow with two problems:

1. **Folder:** The new sub-workflow is **not created inside the same folder** as the parent workflow (e.g. under Personal → `test`). It appears at the **root** of the workflow list instead.
2. **Connections:** Inside the generated sub-workflow, **edges between nodes are missing** on the canvas (nodes exist but are not wired as in the selection).

On **2.6.3**, the same workflow and the same operation did **not** exhibit either issue.

### To Reproduce

1. Open **Personal** (or equivalent project space).
2. Create a folder, e.g. **`test`**.
3. Inside **`test`**, create a **new workflow** and add several **ordinary** nodes (e.g. Set → HTTP Request → Set) with **normal linear connections** (no Merge as single entry, no If/Switch branching out of the selection if possible).
4. **Select** the last contiguous segment of nodes (or the full chain you intend to extract).
5. **Right-click** the canvas background → **Convert to sub-workflow**.
6. Open the **newly created** workflow from the sidebar.

### Expected behavior

1. The new sub-workflow should be created **in the same folder** as the parent workflow (`test`), or at least the behavior should match **2.6.3** (where it did).
2. The sub-workflow should **preserve all internal connections** between the moved nodes, matching the graph before conversion.

### Actual behavior

1. The sub-workflow is created under the **root** of the workflow list (not inside `test`).
2. The sub-workflow **does not show connections** between nodes (broken or empty graph wiring in the UI).

### Environment

- **n8n version:** 2.16.1 (previously **2.6.3** without this issue)
- **Deployment:** [self-hosted Docker]
- **Database:** [SQLite]
- **Browser:** [Chrome]
- **OS:** [macOS]

### To Reproduce

1. Open **Personal** (or equivalent project space).
2. Create a folder, e.g. **`test`**.
3. Inside **`test`**, create a **new workflow** and add several **ordinary** nodes (e.g. Set → HTTP Request → Set) with **normal linear connections** (no Merge as single entry, no If/Switch branching out of the selection if possible).
4. **Select** the last contiguous segment of nodes (or the full chain you intend to extract).
5. **Right-click** the canvas background → **Convert to sub-workflow**.
6. Open the **newly created** workflow from the sidebar.

### Expected behavior

1. The new sub-workflow should be created **in the same folder** as the parent workflow (`test`), or at least the behavior should match **2.6.3** (where it did).
2. The sub-workflow should **preserve all internal connections** between the moved nodes, matching the graph before conversion.

### Debug Info

1. The sub-workflow is created under the **root** of the workflow list (not inside `test`).
2. The sub-workflow **does not show connections** between nodes (broken or empty graph wiring in the UI).

### Operating System

macOS

### n8n Version

2.16.1

### Node.js Version

24.14.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

Omc12:
## Summary

This PR fixes a regression in the "Convert to sub-workflow" feature where newly created sub-workflows were placed at the root directory instead of the parent workflow's folder, and internal node connections were stripped during extraction.

**Root Cause:**
The `useWorkflowExtraction` composable was using `injectWorkflowDocumentStore()`, which depends on Vue's provide/inject context. Since the conversion flow is triggered via `WorkflowExtractionNameModal.vue` (mounted globally in `AppModals.vue`), it sits outside the `WorkflowLayout.vue` provider tree. This caused the injection to return `null`, leaving the extraction logic without the necessary context to identify the parent folder or the connection graph.

**Changes:**
- Migrated `useWorkflowExtraction` to resolve the `workflowDocumentStore` dynamically via Pinia's store registry using the active `workflowId`.
- Implemented the store reference as a `computed` property to maintain reactivity and ensure global modals can access the correct workflow context.
- Removed redundant optional chaining on the computed ref to resolve ESLint `no-unnecessary-condition` workspace errors.

**How to Test:**
1. Create a folder in your workspace (e.g., "Test Folder").
2. Create a new workflow inside that folder.
3. Add a few nodes (e.g., `Set` -> `HTTP Request` -> `Set`) and connect them.
4. Select the nodes, right-click, and select **Convert to sub-workflow**.
5. Give it a name and confirm.
6. **Verify Folder:** Ensure the new sub-workflow appears inside "Test Folder" in the workflow list.
7. **Verify Connections:** Open the new sub-workflow and confirm all nodes remain wired correctly.

fixes #28751

## Review / Merge checklist

- [ ] Tests included.


CharlieKolb (approved):
Thank you for the contribution! Much appreciated

PR #28786, opened 2026-04-21
refactor(editor): Add hydrate and reset to workflow document store

alexgrozav:
## Summary

Adds `hydrate(workflow: IWorkflowDb)` and `reset()` to the workflow document store so callers have an explicit path for populating and clearing a document from an `IWorkflowDb`. This is the last prep step before `workflow` and `workflowObject` can be removed from `workflowsStore` in a follow-up PR.

- `workflowDocument.store.ts` — `hydrate` fans out to every composable's public setter with safe defaults mirroring the original `setWorkflow` fallback pattern; `reset` does the same with empty-workflow defaults.
- `useWorkflowDocumentActive.setActiveVersion(value)` — single-field setter needed by the `setWorkflowActiveVersion` shim.
- `useWorkflowDocumentTags.setTags`/`addTags` now accept `Array<string | ITag>` and normalize internally, matching `IWorkflowDb.tags` without leaking normalization to callers.
- `workflows.store.ts` — `setWorkflow`, `resetWorkflow`, `setWorkflowActiveVersion` become thin shims delegating to the document store; `workflow.value` and `workflowObject` stay for now. `isNewWorkflow` gets a TODO documenting the planned move and the decision to preserve the `workflowsListStore` coupling.

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/app/stores/workflowDocument.store.ts
```diff
@@ -159,6 +161,61 @@ export function useWorkflowDocumentStore(id: WorkflowDocumentId) {
 			workflowDocumentPinData.setPinData({});
 		}
 
+		function hydrate(workflow: IWorkflowDb) {
```
autologie, line 174: Wdyt about throwing exception or printing warning on console if `workflow.id` and `workflow.version` don't match the store's workflow ID? That would help us find potential bugs early 🤔

packages/frontend/editor-ui/src/app/stores/workflowDocument/useWorkflowDocumentActive.ts
```diff
@@ -31,11 +31,19 @@ export function useWorkflowDocumentActive() {
 		applyActiveState(state);
 	}
 
+	function setActiveVersion(value: WorkflowHistory | null) {
```
autologie, line 34: The fact that the workflow document store is identified by the combination of workflow ID and versionID makes me wonder if it conceptually fits well to have workflow history in workflow document store.

To illustrate, let's say we have
- store instance `vAVHIwUr38TG2ofH@latest`
- store instance `vAVHIwUr38TG2ofH@9ef1c980`
- store instance `vAVHIwUr38TG2ofH@dea23828`

Here, I'm unsure which store (or all?) should receive setActiveVersion call when setting active version for `vAVHIwUr38TG2ofH`. Perhaps we need a new type of store to track WorkflowHistory or could it be the responsibility of workflow list store.

autologie (commented):
General thoughts on active version/workflow history that I wanted to bring up

packages/frontend/editor-ui/src/app/stores/workflowDocument/useWorkflowDocumentActive.ts
alexgrozav: Interesting. I haven't thought of that very thoroughly, but we discussed this during a architectural meeting. The idea was the following:
- When working inside the editor, always use latest. 
- When loading for diff views and such, always use versioned.

I guess for the latest one, there could be a mismatch between store version and active version. What do you think?

packages/frontend/editor-ui/src/app/stores/workflowDocument.store.ts
alexgrozav: Great idea! Added the sanity check.

packages/frontend/editor-ui/src/app/stores/workflowDocument/useWorkflowDocumentActive.ts
alexgrozav: I've added a check and also added a comment to show that "latest" is the source of truth for it. Let's leave it as-is for now since other fields have the same problem (i.e. sharedWithProjects, metadata, etc) and have this as a convention.

autologie: Thanks for the clarification.

> When working inside the editor, always use latest.

Debugging past execution in the canvas might add complication here, but I guess we can think about it later. I'm happy to proceed as is.

autologie (approved):

PR #28790, opened 2026-04-21
feat(core): Add public API endpoints for folder management (folderId)

sandra0503:
## Summary
Add folder CRUD endpoints to the public API, scoped under `/projects/{projectId}/folders`:
- `GET /projects/{projectId}/folders/{folderId}` -- get folder details with sub-folder and workflow counts
- `PATCH /projects/{projectId}/folders/{folderId}` -- update folder name or parent
- 
All endpoints reuse `FolderService` business logic from the internal API. A shared
`resolveProjectId` middleware handles project validation and authorization.

### Permissions changes
- Add scopes `folder:read` and `folder:update` to `OWNER_API_KEY_SCOPES` and `MEMBER_API_KEY_SCOPES`

Closes [LIGO-466](https://linear.app/n8n/issue/LIGO-466/add-missing-get-and-patch-folders-endpoints)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
+++ b/packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
@@ -1,4 +1,9 @@
-import { CreateFolderDto, DeleteFolderDto, ListFolderQueryDto } from '@n8n/api-types';
+import {
+	CreateFolderDto,
+	DeleteFolderDto,
+	ListFolderQueryDto,
+	UpdateFolderDto,
+} from '@n8n/api-types';
 import { Container } from '@n8n/di';
 import { UserError } from 'n8n-workflow';
 
@@ -24,6 +29,8 @@ type FolderHandlers = {
 	createFolder: FoldersEndpoint<{ projectId: string }>;
 	getFolders: FoldersEndpoint<{ projectId: string }>;
 	deleteFolder: FoldersEndpoint<{ projectId: string; folderId: string }>;
+	getFolder: FoldersEndpoint<{ projectId: string; folderId: string }>;
+	updateFolder: FoldersEndpoint<{ projectId: string; folderId: string }>;
 };
 
 const folderHandlers: FolderHandlers = {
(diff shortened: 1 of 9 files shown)


packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
```diff
@@ -89,6 +96,59 @@ const folderHandlers: FolderHandlers = {
...
+	updateFolder: [
+		isLicensed('feat:folders'),
+		apiKeyHasScopeWithGlobalScopeFallback({ scope: 'folder:update' }),
+		async (req, res) => {
+			const { projectId } = req.params;
+			await assertProjectScope(req.user, projectId, ['folder:update']);
+
+			const payload = UpdateFolderDto.safeParse(req.body);
+			if (payload.error) {
+				throw new BadRequestError(payload.error.errors[0].message);
```
geemanjs, line 127: Are we safe to do this? It always worries me when theres a hard coded array lookup 👀

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
```diff
@@ -89,6 +96,59 @@ const folderHandlers: FolderHandlers = {
...
+			}
+
+			try {
+				await Container.get(FolderService).updateFolder(
```
geemanjs: Would it make more sense here to return the updated folder from the call to `updateFolder`? I think this is a relatively common pattern

packages/cli/test/integration/public-api/folders.test.ts
```diff
@@ -525,3 +525,211 @@ describe('DELETE /projects/:projectId/folders/:folderId', () => {
 		expect(response.statusCode).toBe(500);
 	});
 });
+
+describe('GET /projects/:projectId/folders/:folderId', () => {
```
geemanjs, line 529: This style of testing makes me much happier 👍 nice work!

packages/cli/test/integration/public-api/folders.test.ts
```diff
@@ -525,3 +525,211 @@ describe('DELETE /projects/:projectId/folders/:folderId', () => {
...
+		const teamProject = await createTeamProject('No Access');
+		const folder = await createFolder(teamProject, { name: 'Team Folder' });
+
+		const response = await authMemberAgent.get(`/projects/${teamProject.id}/folders/${folder.id}`);
+
+		expect(response.statusCode).toBe(403);
+	});
+
+	test('should return folder details with counts', async () => {
+		testServer.license.enable('feat:folders');
```
geemanjs, line 580: Given the `testServer` outlives the lifecycle of the test do we need to disable the feature in an `afterEach`?

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
```diff
@@ -89,6 +96,59 @@ const folderHandlers: FolderHandlers = {
...
 	],
+	getFolder: [
+		isLicensed('feat:folders'),
+		apiKeyHasScopeWithGlobalScopeFallback({ scope: 'folder:read' }),
+		async (req, res) => {
+			const { projectId } = req.params;
+			await assertProjectScope(req.user, projectId, ['folder:read']);
+
+			try {
+				const folder = await Container.get(FolderService).findFolderInProjectOrFail(
```
alielkhateeb: We already run this as part of `getFolderAndWorkflowCount` https://github.com/n8n-io/n8n/blob/39f3cdc10331b8e416c36f5ed6c5af94d1e04dc1/packages/cli/src/services/folder.service.ts#L249

I find it even strange that we do not return the folder in this method, I suggest: 
- Return folder too from this method. (saves us a trip to the DB)
- Renaming `getFolderAndWorkflowCount` to something like `findFolderWithContentCounts`. (because it represents what it actually does)
- Remove that line and destructure `folder` from `findFolderWithContentCounts` 

I hope I'm not missing something 😂

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
sandra0503: Yes, if there is a Zod error from safeParse then the array contains at least one error

geemanjs: Ahh perfect - I wonder if we should have a utility function for extracting it (I guess we do this a lot)...

`extractFirstZodError(error)`

```
// Some folder somewhere else
function extractFirstZodError(payload) {
  return payload.error.errors[0].message
}
```

Just hides away the ugly and we can maybe wrap it in something more defensive just in case zod behaviour every changes in a future version. 

Probably out of scope of what you are trying to do in this PR though 🙄

packages/cli/test/integration/public-api/folders.test.ts
sandra0503: I had a look and this is handled in `setupTestServer`: `testServer.license.reset()` L144

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
sandra0503: Yes, makes sense! But I think we need to look it up in the service, or is there another way?

https://github.com/n8n-io/n8n/pull/28790/changes#diff-3ae04c14fbbd26f1aadc816731534f9eee489c18ac8f08a961605bd02d28b384R94

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
sandra0503: Good point, that's much better as we have the data already

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
sandra0503: Yeah that would be good. I was also thinking about what if the zod implementation would change, then we have to adjust this everywhere

packages/cli/src/public-api/v1/handlers/folders/folders.handler.ts
geemanjs: I think thats fine 👍.

geemanjs (approved):

alielkhateeb (approved):
🚀

PR #28793, opened 2026-04-21
fix(ai-builder): Prevent duplicate workflow creation on parallel submits in instance AI

aalises:
## Summary

Wrap the sandbox `submit-workflow` tool with a per-filePath identity layer so parallel calls for the same file collapse onto a single workflow instead of creating duplicates.

**Incident:** in a single assistant turn, the builder sub-agent emitted 5 parallel `submit-workflow` calls within 115 ms. Three dropped `workflowId` → each hit the create branch of `InstanceAiWorkflowService` → duplicate rows with the same name. Only the last workflow had credentials wired up, so only it worked.

**Root cause:** the sandbox wrapper at `build-workflow-agent.tool.ts` didn't enforce identity. Between dispatch and first successful create, no shared state existed for concurrent calls to read, so they all raced into create.

**Fix:** the first call for a given resolved `filePath` synchronously installs a deferred in a `Map<filePath, Promise<workflowId>>` before dispatching. Concurrent calls for the same path await the first result and inject the bound `workflowId` — forcing the update branch regardless of what the LLM passed. Map is scoped to the builder-task closure and dies with the task — no cross-module coordinator, eviction hook, or TTL sweep needed.

- https://linear.app/n8n/issue/AI-2406

## Review / Merge checklist

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
@@ -148,6 +148,45 @@ export const submitWorkflowInputSchema = z.object({
 	name: z.string().optional().describe('Workflow name (required for new workflows)'),
 });
 
+export const submitWorkflowOutputSchema = z.object({
+	success: z.boolean(),
+	workflowId: z.string().optional(),
+	workflowName: z.string().optional(),
+	/** Node names whose credentials were mocked via pinned data. */
+	mockedNodeNames: z.array(z.string()).optional(),
+	/** Credential types that were mocked (not resolved to real credentials). */
+	mockedCredentialTypes: z.array(z.string()).optional(),
+	/** Map of node name → credential types that were mocked on that node. */
+	mockedCredentialsByNode: z.record(z.array(z.string())).optional(),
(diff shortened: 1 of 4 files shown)


packages/@n8n/instance-ai/src/tools/workflows/submit-workflow.tool.ts
```diff
@@ -148,6 +148,45 @@ export const submitWorkflowInputSchema = z.object({
...
+
+export type SubmitWorkflowInput = z.infer<typeof submitWorkflowInputSchema>;
+export type SubmitWorkflowOutput = z.infer<typeof submitWorkflowOutputSchema>;
+
+/**
+ * Resolve a raw `filePath` tool argument into an absolute path under the sandbox root.
+ * Exported so identity wrappers can key state by the same resolved path the tool uses.
+ */
+export function resolveSandboxWorkflowFilePath(
+	rawFilePath: string | undefined,
```
Cadiac, line 175: this seems suspectible to path traversal to reach folders outside the root - is that a problem here?

Cadiac (approved):
👍 I guess that works, I still don't understand why we end up in situations where the agent calls submit-workflow many times in quick succession?

PR #28804, opened 2026-04-21
fix(core): Scope live and test webhook handlers to their route family

BerniWittmann:
## Summary

Each live/test webhook URL prefix (`/form`, `/webhook`, `/mcp` and the `-test` variants) now resolves to a handler that only serves webhooks whose node declares a matching `webhookDescription.nodeType`. Requests that target a node from a different family return a standard not-found response.

The `expectedNodeType` for a route is passed into `createWebhookHandlerFor`, which threads it through `WebhookRequestHandler` to `IWebhookManager.executeWebhook`. `LiveWebhooks` and `TestWebhooks` remain single services per family (no subclassing) and apply the guard after resolving the webhook. A small shared helper in `node-type-matcher.ts` keeps the comparison consistent across the two managers.

### How to test

1. Activate a Form Trigger workflow, then `curl -i http://localhost:5678/form/<id>` → 200 and `curl -i http://localhost:5678/webhook/<id>` → 404.
2. Activate a regular Webhook, then `curl -i http://localhost:5678/webhook/<path>` → 200 and `curl -i http://localhost:5678/form/<path>` → 404.
3. Chat Trigger on `/webhook/<id>/chat` and MCP Trigger on `/mcp/<id>` continue to work.
4. `cd packages/cli && pnpm test src/webhooks/__tests__/live-webhooks.test.ts src/webhooks/__tests__/test-webhooks.test.ts`
5. `pnpm --filter=n8n-playwright test:local tests/e2e/api/form-endpoint-isolation.spec.ts`

https://linear.app/n8n/issue/NODE-4871

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/webhooks/live-webhooks.ts
+++ b/packages/cli/src/webhooks/live-webhooks.ts
@@ -5,15 +5,6 @@ import type { Response } from 'express';
 import { Workflow, CHAT_TRIGGER_NODE_TYPE } from 'n8n-workflow';
 import type { INode, IWebhookData, IHttpRequestMethods, IWorkflowBase } from 'n8n-workflow';
 
-import { authAllowlistedNodes } from './constants';
-import { sanitizeWebhookRequest } from './webhook-request-sanitizer';
-import type {
-	IWebhookResponseCallbackData,
-	IWebhookManager,
-	WebhookAccessControlOptions,
-	WebhookRequest,
-} from './webhook.types';
-
 import { NotFoundError } from '@/errors/response-errors/not-found.error';
 import { WebhookNotFoundError } from '@/errors/response-errors/webhook-not-found.error';
 import { NodeTypes } from '@/node-types';
@@ -21,6 +12,17 @@ import * as WebhookHelpers from '@/webhooks/webhook-helpers';
 import { WebhookService } from '@/webhooks/webhook.service';
 import * as WorkflowExecuteAdditionalData from '@/workflow-execute-additional-data';
 import { WorkflowStaticDataService } from '@/workflows/workflow-static-data.service';
+
+import { authAllowlistedNodes } from './constants';
+import { matchesExpectedNodeType } from './node-type-matcher';
+import type { ExpectedWebhookNodeType } from './node-type-matcher';
+import { sanitizeWebhookRequest } from './webhook-request-sanitizer';
+import type {
+	IWebhookResponseCallbackData,
+	IWebhookManager,
+	WebhookAccessControlOptions,
+	WebhookRequest,
+} from './webhook.types';
 
 /**
  * Service for handling the execution of live webhooks, i.e. webhooks
@@ -62,6 +64,7 @@ export class LiveWebhooks implements IWebhookManager {
 				// we need to use webhookId for matching
 				isChatWebhookNode(type, webhookId),
 		);
+
 		return webhookNode?.parameters?.options as WebhookAccessControlOptions;
 	}
 
@@ -71,6 +74,7 @@ export class LiveWebhooks implements IWebhookManager {
 	async executeWebhook(
 		request: WebhookRequest,
 		response: Response,
+		expectedNodeType?: ExpectedWebhookNodeType,
 	): Promise<IWebhookResponseCallbackData> {
 		const httpMethod = request.method;
 		const path = request.params.path;
@@ -145,6 +149,16 @@ export class LiveWebhooks implements IWebhookManager {
 			const webhookData = this.webhookService
 				.getNodeWebhooks(workflow, workflow.getNode(webhook.node) as INode, additionalData)
 				.find((w) => w.httpMethod === httpMethod && w.path === webhook.webhookPath) as IWebhookData;
+
+			if (
+				expectedNodeType &&
+				!matchesExpectedNodeType(expectedNodeType, webhookData?.webhookDescription.nodeType)
+			) {
+				throw new WebhookNotFoundError(
+					{ path, httpMethod, webhookMethods: await this.getWebhookMethods(path) },
+					{ hint: 'production' },
+				);
+			}
 
 			// Get the node which has the webhook defined to know where to start from and to
 			// get additional data
--- a/packages/cli/src/webhooks/test-webhooks.ts
+++ b/packages/cli/src/webhooks/test-webhooks.ts
@@ -1,4 +1,3 @@
-import { TEST_WEBHOOK_TIMEOUT } from '@/constants';
 import { OnPubSubEvent } from '@n8n/decorators';
 import { Service } from '@n8n/di';
 import type express from 'express';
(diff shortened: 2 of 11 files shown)


packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -62,9 +64,26 @@ export class LiveWebhooks implements IWebhookManager {
...
 		);
+
+		if (webhookNode && !this.matchesNodeTypeForNode(webhookNode)) {
+			return undefined;
+		}
+
 		return webhookNode?.parameters?.options as WebhookAccessControlOptions;
 	}
 
+	protected matchesNodeType(declared: 'form' | 'webhook' | 'mcp' | undefined): boolean {
```
dlavrenuek: for `declared` we can use the type `IWebhookDescription["nodeType"]` to not have type duplication

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -62,9 +64,26 @@ export class LiveWebhooks implements IWebhookManager {
...
+		if (this.expectedNodeType === 'form') return declared === 'form';
+		if (this.expectedNodeType === 'mcp') return declared === 'mcp';
+		return declared === undefined || declared === 'webhook';
+	}
```
dlavrenuek: If `expectedNodeType` is set to `"webhook"` in `LiveWebhooks` instead of undefined then this check can be greatly simplified 


```suggestion
	protected matchesNodeType(declared: IWebhookDescription["nodeType"] = 'webhook'): boolean {
		return this.expectedNodeType === declared;
	}
```

dlavrenuek: To confirm that the behavior would be the same I did create a "decision table" (I think this is what it's called)

**Current implementation** (`declared` first row, `expectedNodeType` first column each row)

| | form | webhook | mcp | undefined |
|---|---|---|---|---|
| form | 1 | 0 | 0 | 0 |
| mcp | 0 | 0 | 1 | 0 |
| undefined | 0 | 1 | 0 | 1 |

**Alternative** (`declared` first row, `expectedNodeType` first column each row )

| | form | webhook | mcp | undefined |
|---|---|---|---|---|
| form | 1 | 0 | 0 | 0 |
| mcp | 0 | 0 | 1 | 0 |
| webhook | 0 | 1 | 0 | 1 |

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -62,9 +64,26 @@ export class LiveWebhooks implements IWebhookManager {
...
+
+	private matchesNodeTypeForNode(node: INode): boolean {
+		const nodeType = this.nodeTypes.getByNameAndVersion(node.type, node.typeVersion);
+		const declared = nodeType.description.webhooks?.find((w) => w.nodeType)?.nodeType;
+		return this.matchesNodeType(declared);
```
dlavrenuek: Looking at how this check works `declared` will be set to the first defined webhook nodeType if `nodeType` is set or undefined. It means that if a node has multiple webhooks with different `nodeType` (f.e. a form webhook and a regular webhook) this will always find the first one and fail for all other types.

I did go through all of our built in nodes and the only such case is the wait node, but it's webhooks run through waiting webhooks and not `live-webhooks`. But since there is no restriction that you must only use one `nodeType` or webhook type per node it could be possible that there are community nodes with multiple webhook nodeType. This method is only executed for setting up CORS headers, so it's better to return configured cors options even if the type does not match than to not return the configured options (removing the check completely will be better than to have it). Alternatively on the normal execution path we search for the specific webhook instead the first one with a `nodeType`: https://github.com/n8n-io/n8n/pull/28804/changes#diff-30d7a5c748dcae18f480e6b34a681454a7a19442e6a0e46588ded8ada4149650R166

packages/cli/src/webhooks/live-webhooks.ts
```diff
@@ -201,3 +227,13 @@ export class LiveWebhooks implements IWebhookManager {
 		return webhook;
 	}
 }
+
+@Service()
+export class LiveFormWebhooks extends LiveWebhooks {
```
dlavrenuek: So far we have one service per file across the code base, I would stick to this pattern. So if we add `LiveFormWebhooks` it should be in `live-form-webhooks.ts`, same for `LiveMcpWebhooks`. But I'm not 100% sure that it's the best to have these separate classes. Looking at changes in `abstract-server.ts` there will be 3 objects created for each webhook service class, so it's 3 times more memory usage than before (only 3x memory needed for the service + WebhookRequestHandler).

What do you think of determining the `expectedNodeType` from `request` object, which is passed to `executeWebhook` (or path passed to `findAccessControlOptions`)? This could be done by using the configured paths similar to how the webhook paths are defined in https://github.com/n8n-io/n8n/blob/95696c7106371c47b8c3d8414e2b4d63ddd84dcb/packages/cli/src/abstract-server.ts#L89

packages/cli/src/webhooks/test-webhooks.ts
```diff
@@ -49,17 +50,31 @@ const SINGLE_WEBHOOK_TRIGGERS = [
...
+		if (this.expectedNodeType === 'form') return declared === 'form';
+		if (this.expectedNodeType === 'mcp') return declared === 'mcp';
+		return declared === undefined || declared === 'webhook';
+	}
+
+	private matchesNodeTypeForNode(node: INode): boolean {
+		const nodeType = this.nodeTypes.getByNameAndVersion(node.type, node.typeVersion);
+		const declared = nodeType.description.webhooks?.find((w) => w.nodeType)?.nodeType;
+		return this.matchesNodeType(declared);
+	}
```
dlavrenuek: These methods are now duplicated in two classes. May be we could extract them to an abstract class or separate helper functions?

dlavrenuek (changes requested):

packages/cli/src/webhooks/live-webhooks.ts
BerniWittmann: makes sense :+1:

packages/cli/src/webhooks/test-webhooks.ts
BerniWittmann: this is now outdated

BerniWittmann:
@dlavrenuek I have reworked the approach and simplified it now, thanks for the review :)

packages/testing/playwright/tests/e2e/api/form-endpoint-isolation.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+						name: 'Form Trigger',
+						webhookId,
+					},
+				],
+				connections: {},
+			};
+
+			const { workflowId, createdWorkflow } = await api.workflows.createWorkflowFromDefinition(
+				workflow,
+				{ makeUnique: false },
```
dlavrenuek, line 63: When using `makeUnique: false` the webhook id/path will not be autogenerated, that's exactly what we want in this test. But there is a side effect: if the test fails for some reason, which sometimes happens in e2e tests, it will rerun but every run afterwards will fail at the activate step due to webhook conflict error (there is already a published workflow with the same webhooks from run/try 1). So for this kind of tests you should track what workflows were created and delete them f.e. in `afterEach` - we already have this approach in some webhook-related tests

dlavrenuek (changes requested):
The code looks good now. I have only found one non-obvious functionality that should be changed slightly in the test

packages/testing/playwright/tests/e2e/api/form-endpoint-isolation.spec.ts
BerniWittmann: ah thats a good insight, good to know, thank you :) added the after each clanup

dlavrenuek (dismissed):
Very nice implementation, thank you @BerniWittmann 🚀

BerniWittmann:
@dlavrenuek Had to update the e2e tests, as the mcp trigger e2e tests were still using the old "permissive" `webhook` endpoint instead of the `mcp` endpoint (as the UI states)

dlavrenuek (approved):

PR #28811, opened 2026-04-21
feat: Use proxy for agent builder

yehorkardash:
## Summary

This PR adds a new settings section and support of AI assistant proxy
image
Self-hosted users can configure agent builder model through UI or with env variables
image

When self-hosted user start using the agent-builder they will see the following message:
image

Modes:
- Cloud - use proxy by default when `N8N_AI_ASSISTANT_BASE_URL` is set
- Self-hosted - use custom model if it's selected in UI
- Self-hosted - fallback to `N8N_AI_ANTHROPIC_KEY` or `ANTHROPIC_API_KEY` when there's not assistant url or custom model set

Testing:
1. Set N8N_AI_ASSISTANT_BASE_URL and configure license. Verify that agent builder can be accessed without configuration
2. Remove N8N_AI_ASSISTANT_BASE_URL any `ANTHROPIC*` envs. Verify that the user is asked to configure a model
3. Set `ANTHROPIC_API_KEY`. Verify that user. Verify that agent builder can be accessed without configuration

https://linear.app/n8n/issue/AGENT-28/feature-use-ai-assistant-proxy-for-agent-builder

## Review / Merge checklist

- [ ] Tests included. 


yehorkardash:
@claude review

packages/cli/src/modules/agents/builder/agents-builder-settings.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Admin endpoints for the agent builder model + lightweight per-user
+ * readiness endpoint. Lives under its own base path so the FE can talk to it
+ * independently of any other AI feature.
+ */
+@RestController('/agent-builder')
+export class AgentsBuilderSettingsController {
+	constructor(private readonly settingsService: AgentsBuilderSettingsService) {}
+
+	@Get('/settings')
+	@GlobalScope('agent:read')
```
burivuhster: Should this be `agent:manage`?

packages/@n8n/permissions/src/roles/scopes/global-scopes.ee.ts
```diff
@@ -150,6 +151,14 @@ export const GLOBAL_OWNER_SCOPES: Scope[] = [
...
 
 export const GLOBAL_MEMBER_SCOPES: Scope[] = [
+	'agent:create',
+	'agent:read',
+	'agent:update',
+	'agent:list',
+	'agent:delete',
+	'agent:publish',
+	'agent:unpublish',
+	'agent:execute',
```
burivuhster: Do we need these in global member scopes?

packages/cli/src/modules/agents/builder/agents-builder-settings.controller.ts
```diff
@@ -0,0 +1,51 @@
...
+		// pattern as `CreateDestinationDto` in the log-streaming module.)
+		const parseResult = AgentBuilderAdminSettingsUpdateDto.safeParse(req.body);
+		if (!parseResult.success) {
+			throw new BadRequestError(parseResult.error.errors[0]?.message ?? 'Invalid request body');
+		}
+		await this.settingsService.updateAdminSettings(parseResult.data);
+		return await this.settingsService.getAdminSettings();
+	}
+
+	@Get('/status')
```
burivuhster, line 42: Does this endpoint need any permissions check?

packages/cli/src/modules/agents/builder/agents-builder-settings.service.ts
```diff
@@ -0,0 +1,263 @@
...
+	// eslint-disable-next-line @typescript-eslint/no-require-imports
+	const { ProxyAgent } = require('undici') as typeof Undici;
+	const dispatcher = new ProxyAgent(proxyUrl);
+	return (async (url: string | URL | Request, init?: RequestInit) =>
+		await globalThis.fetch(url, {
+			...init,
+			// @ts-expect-error dispatcher is a valid undici option for Node.js fetch
+			dispatcher,
+		})) as typeof globalThis.fetch;
+}
```
burivuhster: I would prefer common `proxyFetch` or `getProxyAgent` from `@n8n/ai-utilities`, as proxy logic might be tricky and easy to miss edge cases like `NO_PROXY` for example

packages/cli/src/modules/agents/builder/agents-builder-settings.service.ts
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * Admin-instance-wide settings for the agent builder. Decides which model
+ * the builder LLM runs on:
+ *   1. `mode: 'custom'` — admin-picked credential. Wins regardless of platform.
+ *   2. `mode: 'default'` + AI proxy enabled — runs through the n8n AI assistant proxy.
+ *   3. `mode: 'default'` + env-var backstop set (`N8N_AI_ANTHROPIC_KEY` /
+ *      `ANTHROPIC_API_KEY`) — direct Anthropic API calls.
+ *
+ * Self-contained: does not import anything from the instance-ai module.
```
burivuhster: IMHO this comment does not add much value

packages/frontend/editor-ui/src/features/agents/components/settings/AgentBuilderModelSection.vue
```diff
@@ -0,0 +1,267 @@
...
+	const settings = store.effectiveSettings;
+	if (settings.mode !== 'custom') return;
+	const currentChatHubProvider = CATALOG_TO_CHATHUB[settings.provider];
+	if (currentChatHubProvider === provider) {
+		store.setCustomSelection({
+			provider: settings.provider,
+			credentialId: credentialId ?? '',
+			modelName: settings.modelName,
+		});
+	}
```
burivuhster: It's unclear to me what this part does. Can you please add a brief comment with explanation?

burivuhster (commented):
Looks good! Couple of questions/suggestions

packages/cli/src/modules/agents/builder/agents-builder-settings.controller.ts
yehorkardash: It's a read request and used on frontend to check if agent is configured so not only admins should have this scope

burivuhster: Got it. But it's about global settings for all agents, not about specific agent, right? Maybe we need a separate global scope for reading settings

yehorkardash: I can put admin settings under 'agent:manage' and then update `/status` to use a new `agent:readSettings`. wdyt?

burivuhster: Makes sense! If both writing and reading settings would be protected with `agent:manage` scope, maybe for status only we might not need any scope, just let every authenticated user request it.

burivuhster (approved):
Thanks for addressing my feedback! LGTM!

PR #28820, opened 2026-04-21
feat(agents): prevent multiple agents from sharing a Telegram credential

burivuhster:
## Summary

Telegram's Bot API allows only one active webhook per bot token, so two agents using the same Telegram credential would silently clobber each other's webhook. This PR blocks the connect flow upfront with a clear error.

- **Backend** (`ChatIntegrationService.connect`): new `ensureTelegramCredentialAvailable` guard runs before the adapter is built, throwing `ConflictError` (HTTP 409) if the credential is already claimed. Two detection signals:
  - **DB check** — new `AgentRepository.findByIntegrationCredential(type, credentialId, excludeAgentId)` finds other agents in the same instance; error message names the owning agent.
  - **Telegram `getWebhookInfo` probe** — catches leftover webhooks (stale state, cross-instance, Telegram-trigger workflows) that aren't claimed by any agent in our DB. Fails open on network errors so a Telegram API blip can't block a legitimate connect.
- **Frontend** (`AgentIntegrationsPanel.vue`): the inline error now hides the "Edit credential" link on 409 responses — editing the credential wouldn't resolve a conflict.
- Startup reconnect is first-come-first-served: the existing per-integration `try/catch` in `reconnectAll` logs the `ConflictError` and skips the losing agent, no crash.

Ref: https://linear.app/n8n/issue/AGENT-7

## Test plan

- [ ] Create two published agents A and B in the same project
- [ ] Connect A to a Telegram credential — expect \"Connected\"
- [ ] On B, select the same Telegram credential and click Connect — expect inline error *\"Telegram credential is already connected to agent \"A\"\"* with no \"Edit credential\" link
- [ ] Restart the backend while both agents have the same credential persisted — expect first agent reconnects successfully, second logs the ConflictError and is skipped (no crash)
- [ ] On A with no other agents owning the credential, manually set a foreign webhook URL on the bot via curl, then try to reconnect A — expect inline 409 \"Telegram bot already has a webhook registered elsewhere: &lt;url&gt;\"


packages/cli/src/modules/agents/integrations/chat-integration.service.ts
```diff
@@ -104,6 +105,12 @@ export class ChatIntegrationService {
 			webhookUrlFor: (platform) => this.buildWebhookUrl(agentId, projectId, platform),
 		};
 
+		// Telegram: each bot token can serve only one webhook at a time. Enforce
+		// one-agent-per-credential before we touch Telegram's API.
+		if (integrationType === 'telegram') {
```
yehorkardash: can we abstract it? I imagine other chat providers that rely on webhooks may have the same issue

burivuhster: Good call! Moved this into `onBeforeConnect` hook

yehorkardash (approved):

PR #28823, opened 2026-04-21
fix(editor): Hide credential check icon while Instance AI setup card has pending params (no-changelog)

aalises:
## Summary

In the Instance AI workflow setup wizard, the small green ✓ next to the card title rendered as soon as the credential test succeeded, regardless of whether required parameters were still empty. Users read that checkmark as "step complete" even though fields like Telegram's `chatId` were still blank.

The small icon is driven by `getCredTestIcon(currentCard)`, which only reflects credential test state. The fix gates the `check` variant on `!cardHasParamWork(currentCard)` so it only renders once there is no pending parameter work. Spinner and warning variants are unchanged; the full "Complete" label (already gated on `isCardComplete`) is also unchanged. A `data-test-id` is added on the icon for test coverage.

- `InstanceAiWorkflowSetup.vue:577`: gate success ✓ on `!cardHasParamWork(currentCard)`.
- `InstanceAiWorkflowSetup.test.ts`: regression test: tested credential + pending param issues → no ✓ in header.

https://linear.app/n8n/issue/AI-2411/bug-card-shows-completed-icon-when-its-not-really-completed

## Review / Merge checklist

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiWorkflowSetup.vue
+++ b/packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiWorkflowSetup.vue
@@ -574,7 +574,8 @@ const nodeNamesTooltip = computed(() => nodeNames.value.join(', '));
 						:class="$style.loading"
 					/>
 					<N8nIcon
-						v-else-if="getCredTestIcon(currentCard) === 'check'"
+						v-else-if="getCredTestIcon(currentCard) === 'check' && !cardHasParamWork(currentCard)"
+						data-test-id="instance-ai-workflow-setup-cred-check"
 						icon="check"
 						size="small"
 						:class="$style.success"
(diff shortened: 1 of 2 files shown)


packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiWorkflowSetup.vue
```diff
@@ -574,7 +574,8 @@ const nodeNamesTooltip = computed(() => nodeNames.value.join(', '));
 						:class="$style.loading"
 					/>
 					<N8nIcon
-						v-else-if="getCredTestIcon(currentCard) === 'check'"
+						v-else-if="getCredTestIcon(currentCard) === 'check' && !cardHasParamWork(currentCard)"
```
scdekov, line 577: Should we just drop this case and leave it to the one with `isCardComplete` condition few lines bellow?

scdekov (approved):

PR #28824, opened 2026-04-21
fix(agents): reliable Telegram markdown delivery via MarkdownV2 + streaming fixes

burivuhster:
## Summary

Makes agent replies land reliably in Telegram. Three layers of fix, the first one in our code and the other two as `pnpm patch` diffs against the chat SDK.

**Bridge (our code):** `AgentChatBridge.consumeStream` now drives the chat SDK's own `StreamingMarkdownRenderer` so only balanced-markdown prefixes are yielded mid-stream — unclosed \`**bold\`, \`` `code `\`, \`[link](\` etc. are held back until their close arrives. On segment end we yield the `finish()` tail so the chat SDK's final raw-accumulated edit is guaranteed balanced too. Slack's native-streaming path is unchanged (it passes raw deltas straight to its Block Kit adapter).

**`@chat-adapter/telegram@4.26.0` patch:**
- `TELEGRAM_MARKDOWN_PARSE_MODE`: \`Markdown\` → \`MarkdownV2\`.
- Custom `stringifyMarkdownV2` emitting correct V2 markers (single \`*\`/\`_\`/\`~\`) with full literal-text escaping.
- `fromAst` walkAst transforms for block-level CommonMark V2 can't render: heading → bold paragraph, thematicBreak → empty.
- Smart `truncateMessage` backs off to nearest `\n` or space within 512 chars so the 4096-char limit never lands mid-entity (the bug that was leaving `[text](url` unterminated and tripping the V2 parser).
- Route card fallback text through `fromMarkdown` so card text gets V2 escaping too.
- Lightweight outgoing-call logging for `editMessageText` / `sendMessage` — was instrumental during this iteration; low-risk to keep.

**`chat@4.26.0` patch:**
- Join card parts with `\n\n` (paragraph break) instead of `\n` (soft break), so CommonMark emphasis markers in different parts don't pair across parts.

Also bumps the four chat SDK packages from `^4.20.2` → `^4.26.0` (tree in lockfile).

Ref: https://linear.app/n8n/issue/AGENT-7

## Test plan

- [ ] Send a message to a Telegram-connected agent, have it reply with heavy markdown (headings, bold, italic, code, lists, tables, blockquotes, links). Final message renders correctly in Telegram, no `fallbackStream edit failed` warnings, no stuck `...` placeholder.
- [ ] Send a message that triggers a rich UI card reply. Card body renders without weird nested emphasis; buttons work.
- [ ] Prompt the agent to produce a response longer than 4096 chars. Message is truncated cleanly at a line/word boundary (not inside a link), no parse errors.
- [ ] Slack integration remains unchanged — markdown responses look identical to before.
- [ ] Unit tests: `cd packages/cli && pnpm test src/modules/agents` — 197/197 pass (includes 6 new edge-case tests for the committable-text stream: empty stream, reasoning-delta filtering, `message`/`error` chunk termination, suspended-without-followup, per-segment renderer reset).

## Notes for reviewers

- The two patches are versioned at exactly `4.26.0`; updating the chat SDK in the future will need a patch refresh.
- The outgoing-call `console.log` inside the telegram patch is intentionally plain stdout — handy during debugging; can be behind a flag or removed in a follow-up if preferred.


burivuhster:
We decided to disable streaming for now

PR #28835, opened 2026-04-21
feat(ai-builder): Add LangSmith integration for workflow eval tracking (no-changelog)

JoseBra:
## Summary
- Wires Instance AI workflow execution evals into LangSmith: JSON test cases synced to a dataset with deterministic example IDs, runs recorded as experiments with per-scenario feedback, and traceable spans for `workflow_build` + `scenario_execution`
- CLI detects `LANGSMITH_API_KEY` and switches to `evaluate()` mode; without the key the same CLI falls back to a direct loop and still writes `eval-results.json` + HTML report
- Experiment-level aggregates (`duration_s`, `avg_build_s`, `avg_exec_s`, `unique_builds`) surfaced at the experiment row via a post-run `updateProject` that merges with LangSmith's auto-populated metadata (keeps the Splits column intact)
- HTML report restored + added to CI artifacts alongside `eval-results.json`
- Bug fixes from review: direct-mode HTML being empty (fileSlug propagation), `--output-dir` flag being a no-op, dead `setSessionCookie` code, verifier had no timeout, workflow cleanup not in `finally`, dataset-sync swallowing auth/network errors

Closes https://linear.app/n8n/issue/TRUST-24

## Test plan
- [x] Full local run (27 scenarios, 8 builds, ~8 min wall clock): `dotenvx run -f ../../../.env.local -- pnpm eval:instance-ai --verbose`
- [x] Filtered local run: `--filter contact-form` — resolves to 1 split, runs 5 scenarios only
- [x] Direct-loop fallback (no `LANGSMITH_API_KEY`): produces valid JSON + HTML report
- [x] Verified via LangSmith MCP that feedback aggregates reconcile with our metadata (`total_build_s ≈ avg_build_s × unique_builds`)
- [x] Typecheck + lint clean
- [ ] CI run (this PR) completes and posts the eval comment


packages/@n8n/instance-ai/evaluations/cli/index.ts
```diff
@@ -1,33 +1,90 @@
...
+// Instance AI workflow eval CLI
+//
+// Runs workflow execution evaluations. When LANGSMITH_API_KEY is set, uses
+// LangSmith's evaluate() for experiment tracking and tracing. Otherwise
+// falls back to a direct loop with the same eval-results.json output.
+// ---------------------------------------------------------------------------
+
+import type { InstanceAiEvalExecutionResult } from '@n8n/api-types';
 import { mkdirSync, writeFileSync } from 'fs';
+import { Client } from 'langsmith';
```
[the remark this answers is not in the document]
JoseBra: The "lazy-load heavy modules" rule is aimed at native modules and large parsers — LangSmith is neither, and   this file is the eval CLI's entry point, quite isolated (invoked once per run, then does minutes of work). Lazy-loading buys maybe 100ms of startup against code that's harder to read.

packages/@n8n/instance-ai/evaluations/cli/index.ts
```diff
@@ -40,113 +97,576 @@ async function main(): Promise<void> {
...
+	execErrors: string[];
+	evalResult?: InstanceAiEvalExecutionResult;
+}
 
-	// Print console summary
-	printSummary(results);
+function collectResultsFromExperiment(rows: Array<{ run: Run }>): CollectedResult[] {
+	return rows.map((row) => {
+		const output = parseTargetOutput(row.run.outputs);
+		const inputs = runInputsSchema.parse(row.run.inputs ?? {});
```
aalises: `parse()` throws on type mismatch (zod `.default()` only fills `undefined`). One malformed row aborts the whole `.map` and loses every experiment result. Use `safeParse` per row.

packages/@n8n/instance-ai/evaluations/cli/index.ts
```diff
@@ -1,33 +1,90 @@
...
 
-	const totalScenarios = testCases.reduce((sum, tc) => sum + tc.scenarios.length, 0);
-	console.log(
-		`Running ${String(testCases.length)} workflow test case(s) with ${String(totalScenarios)} scenario(s)\n`,
-	);
+type TargetOutput = Omit<z.infer<typeof targetOutputSchema>, 'evalResult'> & {
+	evalResult?: InstanceAiEvalExecutionResult;
+};
 
+function isEvalResult(v: unknown): v is InstanceAiEvalExecutionResult {
```
aalises, line 75: Fake type guard: asserts any non-null object is a full `InstanceAiEvalExecutionResult`. HTML report accesses `.errors` / `.hints` / `.nodeResults` unguarded — crashes mid-render on schema drift. Validate with zod or guard each field access.

packages/@n8n/instance-ai/evaluations/cli/index.ts
```diff
@@ -40,113 +97,576 @@ async function main(): Promise<void> {
...
+		avg_exec_s: Math.round(avgExecMs / 100) / 10,
+		unique_builds: uniqueBuilds,
+	};
+
+	try {
+		const project = await lsClient.readProject({ projectName: experimentName });
+		// `updateProject` replaces `extra` wholesale — preserve it so auto-set
+		// fields (splits, etc.) survive.
+		const existingExtra = (project.extra ?? {}) as Record<string, unknown>;
+		const existingMetadata = (existingExtra.metadata ?? {}) as Record<string, unknown>;
```
aalises: Unchecked `as` casts (violates repo TS rules). If `project.extra` or `.metadata` isn't an object the spread silently drops/corrupts LangSmith auto-fields (incl. the Splits column this PR claims to preserve).

packages/@n8n/instance-ai/evaluations/cli/args.ts
```diff
@@ -38,6 +44,9 @@ const cliArgsSchema = z.object({
 	filter: z.string().optional(),
 	keepWorkflows: z.boolean().default(false),
 	outputDir: z.string().optional(),
+	dataset: z.string().default('instance-ai-workflow-evals'),
+	concurrency: z.number().int().positive().default(99),
```
aalises: Default 99 is risky locally: only builds are capped (4), executions fan out to `maxConcurrency`, so up to 99 scenarios hit n8n in parallel. Direct-loop path caps at 4 — modes diverge sharply.

packages/@n8n/instance-ai/evaluations/cli/index.ts
```diff
@@ -40,113 +97,576 @@ async function main(): Promise<void> {
...
+	const datasetName = await syncDataset(lsClient, args.dataset, logger, args.filter);
+
+	const buildLimiter = pLimit(MAX_CONCURRENT_BUILDS);
+	const buildCache = new Map<string, Promise<BuildResult>>();
+	const buildDurations = new Map<string, number>();
+
+	async function getOrBuild(
+		prompt: string,
+	): Promise<{ build: BuildResult; buildDurationMs?: number }> {
+		const existing = buildCache.get(prompt);
```
aalises: `buildCache` keyed by raw `prompt`: two test case files with identical prompts silently share a build (and one cleanup). Key by `fileSlug` instead.

aalises (commented):
Awesome! some comments

packages/@n8n/instance-ai/evaluations/cli/index.ts
JoseBra: Fixed in the next commit. `parseTargetOutput` now uses `safeParse` and returns `undefined` on malformed input; all iteration loops (`reshapeLangSmithRuns`, `updateExperimentAggregates`, `writePerRunPassMetrics`, `feedbackExtractor`) skip bad rows instead of aborting.

packages/@n8n/instance-ai/evaluations/cli/index.ts
JoseBra: Tightened `isEvalResult` to check the fields we actually touch downstream (`nodeResults`, `errors`, `hints`). Not a full zod schema since we produce this data ourselves, but enough to catch obvious schema drift.

packages/@n8n/instance-ai/evaluations/cli/index.ts
JoseBra: Replaced both casts with an `isPlainObject()` runtime guard so non-object values fall back to `{}` instead of being spread.

packages/@n8n/instance-ai/evaluations/cli/args.ts
JoseBra: Good call on the divergence. Lowered the default to 16 so LangSmith-mode parallelism is closer to direct-loop's effective ~20 scenarios. CI still runs at --concurrency 4 explicitly.

packages/@n8n/instance-ai/evaluations/cli/index.ts
JoseBra: Intentional — two files with identical prompts produce identical workflows, so sharing the build is correct. Cleanup handles it fine (one workflow id in `claimedWorkflowIds`, one deletion). Keying by fileSlug would force a redundant build for no gain. Happy to reconsider if you have a case where the dedup is wrong.

JoseBra:
/size-limit-override

Asking for an override rather than splitting. This PR wires Instance AI evals into LangSmith — dataset sync, experiment tracking, multi-run pass@k / pass^k, traceable spans, CI reporting — and it's hard to carve up without either shipping half-working pieces or creating a dependency chain where each PR reviewer has to hold the next one in their head.

The bulk of the lines come from three places worth flagging:

- `langsmith/dataset-sync.ts` (~315 lines) is a new module, but it's genuinely just "sync JSON files to a LangSmith dataset."
- `report/workflow-report.ts` (~500 lines) recovers the HTML file reporter we had for local debugging. Self-contained code, mostly CSS/HTML templating — doesn't interact with anything else.
- `cli/index.ts` grew because it now handles two execution paths (LangSmith + direct fallback) with multi-run aggregation. A structural refactor of this file is coming in a follow-up — it's already on my list — but doing it here would make this PR even bigger and muddy the story.

Everything stays inside `packages/@n8n/instance-ai/evaluations/` and the eval CI workflow — no cross-package interfaces change. Happy to walk through it if that helps.

aalises (approved):

Matsuuu (approved):

PR #28836, opened 2026-04-21
fix(editor): Fix NDV not loading parameters after refresh on instance AI (no-changelog)

Cadiac:
## Summary

Remove logic attempting to open 'output' as the fallback for `NodeSettings.openPanel` state, as no such panel actually exists. This seems to be a recent mistake introduced in #28412 and this PR undoes those changes.

Now NDV falls back to "params" tab correctly, so when the artifact view refreshes (after update) it no longer ends up in limbo state where the tab it tries to open is nonexisting "output" tab.

https://linear.app/n8n/issue/INS-23/bug-node-preview-empty

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/types/nodeSettings.ts
+++ b/packages/frontend/editor-ui/src/app/types/nodeSettings.ts
@@ -4,5 +4,4 @@ export type NodeSettingsTab =
 	| 'communityNode'
 	| 'docs'
 	| 'action'
-	| 'credential'
-	| 'output';
+	| 'credential';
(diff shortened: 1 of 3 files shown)


mutdmour (approved):
LGTM. An E2E would be great to add here. to avoid this bug happening again. Can be just a demo page test, rather than an instance ai test

Cadiac:
This would've only come up in a pretty full e2e case where the artifact workflow viewer gets first loaded and then refreshed (resulting in the default value on nodeSettings) and then NDV opened - I think this unit test catches the actual issue we faced on a reasonable level.

PR #28838, opened 2026-04-21
fix(editor): Show credential dropdown when assigned credential is deleted

Etyamor:
## Summary

When a credential assigned to a node (e.g. HTTP Request with Basic Auth) is deleted from the Credentials page, reopening the node shows a disabled "No credentials yet" dropdown instead of the normal credential selector. The user cannot select another credential or create a new one from that dropdown.

The fix adds an `options.length === 0` guard to the empty state condition so it only renders when there are truly no credentials available. When other credentials exist or the dropdown has a "Create new" option, the normal selector is shown instead.

## Review / Merge checklist

- [ ] Tests included.


Etyamor:
Seems like the issue was fixed and deployed by @BerniWittmann recently

PR #28839, opened 2026-04-21
feat(Alibaba Cloud Node): Switch to dynamic model list fetched from API

DawidMyslak:
## Summary

Switch the Alibaba Cloud Model Studio node from hardcoded static model dropdowns to dynamic lists fetched from the DashScope API at runtime, using the `resourceLocator` pattern. Node version bumped to 1.1 with `@version` gating for backward compatibility.

https://www.loom.com/share/da7cca9db1c945dfaa36fb5289fcce57

### How to test

1. Create a new Alibaba Cloud Model Studio node — verify the model dropdown fetches models dynamically from the API.
2. Select different resources (Text, Image, Video) and confirm each operation shows relevant models filtered by category.
3. Use the "ID" mode to enter a custom model ID manually.
4. Open an existing workflow with a v1 Alibaba Cloud node — verify it still shows the static dropdown with hardcoded options.

https://linear.app/n8n/issue/NODE-4868

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


DawidMyslak:
@alexander-gekov I simplified the model list parsing after realising that Linear description was not accurate, Alibaba response is quite standardized, model name always comes from the same field.

~~Could you please re-approve?~~
Ah, it was not approved yet 😆 .

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/methods/listSearch.ts
```diff
@@ -0,0 +1,98 @@
...
+const PAGE_SIZE = 100;
+
+async function baseModelSearch(
+	this: ILoadOptionsFunctions,
+	modelFilter: (model: string) => boolean,
+	filter?: string,
+): Promise<INodeListSearchResult> {
+	const allIds: string[] = [];
+
+	for (let page = 1; ; page++) {
```
alexander-gekov: Dynamic fetching of model works but it does seem slow, I am guessing it's due to the pagination here.

Is there a way to get the Total Page Count beforehand and then do a `Promise.all` if we know that?

alexander-gekov: Also should we maybe have some more error handling around this logic

alexander-gekov (dismissed):
Left comment regarding the pagination fetching but otherwise looks solid! I think cubic comments we're not relevant, but double check

packages/@n8n/nodes-langchain/nodes/vendors/AlibabaCloud/methods/listSearch.ts
DawidMyslak: Gemini is using way larger page size. I will try to copy that approach.

DawidMyslak:
@alexander-gekov Replaced pagination loop with a single request (`page_size: 200`) - matches the Gemini node approach. The model list fits in one call.

alexander-gekov (approved):

PR #28872, opened 2026-04-22
feat(Webhook Node): Add "Only Run If" option to filter requests

ivov:
## Summary

This PR adds an "Only run if" option to the Webhook node for expression-based rejection of requests to prevent them from generating executions.

Motivation is a relentless flood of requests that trigger no-op executions because the sender cannot filter on their end. Context: https://n8nio.slack.com/archives/C069KJBJ8HE/p1776848342458439

Capture 2026-06-19 at 10 24 09@2x

## How to test

> Tested on both `legacy` and `vm` expression engines

1. Import this workflow: https://n8n.to/only-run-if
2. Start it and call it in production
3. Start it and call it manually

```sh
# production - should trigger
curl -i -X POST http://localhost:5678/webhook/run-if \
  -H 'Content-Type: application/json' \
  -d '{"campaign_id":"user-research-invite"}'

# production - should not trigger
curl -i -X POST http://localhost:5678/webhook/run-if \
  -H 'Content-Type: application/json' \
  -d '{"campaign_id":"something-else"}'

# manual - should trigger
curl -i -X POST http://localhost:5678/webhook-test/run-if \
  -H 'Content-Type: application/json' \
  -d '{"campaign_id":"user-research-invite"}'

# manual - should not trigger
  -H 'Content-Type: application/json' \
  -d '{"campaign_id":"something-else"}'
```

n/a

## Review / Merge checklist

- [x] Tests included. 


shortstacked:
Will we lose track of how many attempts this makes? Would it be still possible to have a metric/tracker to know how many times this particular endpoint would be hit, even if it's filtering out?

ivov:
> Will we lose track of how many attempts this makes? Would it be still possible to have a metric/tracker to know how many times this particular endpoint would be hit, even if it's filtering out?

Existing Prom middleware covers webhook routes, counting every inbound request before this new filter runs.

shortstacked:
> > Will we lose track of how many attempts this makes? Would it be still possible to have a metric/tracker to know how many times this particular endpoint would be hit, even if it's filtering out?
> 
> Existing Prom middleware covers webhook routes, counting every inbound request before this new filter runs.

Is that visible to a user though, or only on internal? Something for the insights view or similar.

ivov:
Closing to remove from my queue, let's reopen if/when there's interest

ivov:
CI looks unrelated, had it retriggering, I'll keep retriggering after review

packages/nodes-base/nodes/Webhook/description.ts
```diff
@@ -298,6 +298,16 @@ export const optionsProperty: INodeProperties = {
...
 		},
+		{
+			displayName: 'Only Run If',
+			name: 'onlyRunIf',
+			type: 'string',
+			default: '',
+			placeholder: "={{ $json.body.campaign_id === 'user-research-invite' }}",
+			// eslint-disable-next-line n8n-nodes-base/node-param-description-miscased-json
+			description:
+				'Expression evaluated against the incoming request. The workflow will run only if the expression returns true. <code>$json</code> exposes the request as <code>{ body, headers, params, query }</code>. Requests that do not match receive a 200 response, without creating an execution. If the expression fails to evaluate, the request is allowed through and the error is logged.',
```
Flexicon, line 309: Seems fairly verbose in the UI:

Image

Maybe we could shorten it to just two or so sentences? I wonder if we could trim it a bit somehow? But it already does seem to contain essential information only.

packages/nodes-base/nodes/Webhook/description.ts
```diff
@@ -298,6 +298,16 @@ export const optionsProperty: INodeProperties = {
 			default: false,
 			description: 'Whether to ignore requests from bots like link previewers and web crawlers',
 		},
+		{
+			displayName: 'Only Run If',
+			name: 'onlyRunIf',
+			type: 'string',
+			default: '',
+			placeholder: "={{ $json.body.campaign_id === 'user-research-invite' }}",
```
Flexicon: Do we use the `=` prefix in the UI? I know that under-the-hood we store expressions like this, but I thought that we don't really display hints in this kind of format in the UI.

Flexicon (dismissed):
LGTM!

packages/nodes-base/nodes/Webhook/description.ts
ivov: [a16dea7](https://github.com/n8n-io/n8n/pull/28872/commits/a16dea77b7dec16c77a019cb2a39e0e3142dc1b6)

packages/nodes-base/nodes/Webhook/description.ts
ivov: Long but essential to know, wouldn't know what to cut

Flexicon: Yeah agreed. Just thought I'd raise.

Flexicon (dismissed):
LGMT. Tested locally - works like a charm!

Flexicon (approved):

PR #28874, opened 2026-04-22
feat(core): Annotate LangSmith traces with user feedback from Instance AI (no-changelog)

mutdmour:
## Summary

Wires the existing thumbs up/down rating in the AskAssistantChat UI (and its optional text comment) into LangSmith as a `user_score` feedback record on the `message_turn` root run. Previously these signals only fired PostHog telemetry — LangSmith is additive.

- New `POST /instance-ai/feedback/:threadId/:responseId` endpoint. Thread-access gated, fire-and-forget, swallows LangSmith errors so UI is never blocked.
- LangSmith run/trace IDs are now persisted on the existing `instance_ai_run_snapshots` row (new migration adds two nullable columns). This survives the in-memory `traceContextsByRunId` cleanup that happens at run finalization, so feedback submitted after the turn completes can still find its anchor.
- Deterministic `feedbackId` (uuid-v5 of `user_score:<responseId>`) so a thumbs-down followed later by a text comment upserts the *same* LangSmith record instead of creating a second one.
- New reusable helper `submitLangsmithUserFeedback(options)` in `@n8n/instance-ai/tracing` — reuses the existing cached direct/proxy `Client` instances and routes proxy headers through the same `AsyncLocalStorage` mechanism as trace creation.

**Not changed**: event schema, SSE transport, `traceContextsByRunId` lifecycle, `MessageRating.vue`, existing PostHog telemetry.

No Linear ticket — internal observability improvement.

## Test plan

- [ ] Fresh-DB smoke: run migrations on sqlite + postgres, confirm `langsmithRunId` / `langsmithTraceId` columns exist on `instance_ai_run_snapshots`.
- [ ] End-to-end with `LANGSMITH_API_KEY` set: send a message, click thumbs-up on the response, confirm a `user_score = 1` feedback record appears on the run in LangSmith with expected `sourceInfo`.
- [ ] End-to-end thumbs-down → text comment flow: confirm a single record (same `feedbackId`) with `score = 0` and `comment = <text>`, not two records.
- [ ] Restart the server between run completion and feedback click: confirm feedback still lands (durability).
- [ ] Confirm no user-visible behavior change when LangSmith is disabled (no env key, no proxy): endpoint still returns `{ ok: true }`, nothing hits the network.
- [ ] Confirm no user-visible behavior change when LangSmith returns an error: endpoint still returns `{ ok: true }`, error logged as `warn`.

## Verification already run

- `pnpm build` at repo root — 57/57 packages ✅
- Typecheck clean: `@n8n/api-types`, `@n8n/instance-ai`, `cli`, `frontend/editor-ui`
- Lint clean: `@n8n/instance-ai`, `@n8n/api-types`, `@n8n/db`, editor-ui feature dir, touched cli files
- Unit tests: 56 in `@n8n/instance-ai` tracing (3 new), 70 in cli controller (5 new), 7 new `DbSnapshotStorage.findLangsmithAnchor`, 27 in `useResponseFeedback` (5 new)

### DIFF
--- a/packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
+++ b/packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
@@ -2,10 +2,13 @@ import type { MigrationContext, ReversibleMigration } from '../migration-types';
 
 export class AddLangsmithIdsToInstanceAiRunSnapshots1777100000000 implements ReversibleMigration {
 	async up({ schemaBuilder: { addColumns, column } }: MigrationContext) {
-		// LangSmith run/trace IDs are UUID v4 (e.g. "f47ac10b-58cc-4372-a567-0e02b2c3d479").
 		await addColumns('instance_ai_run_snapshots', [
-			column('langsmithRunId').varchar(36),
-			column('langsmithTraceId').varchar(36),
+			column('langsmithRunId')
+				.varchar(36)
+				.comment('LangSmith run ID (UUID v4, e.g. "f47ac10b-58cc-4372-a567-0e02b2c3d479").'),
+			column('langsmithTraceId')
+				.varchar(36)
+				.comment('LangSmith trace ID (UUID v4, e.g. "f47ac10b-58cc-4372-a567-0e02b2c3d479").'),
 		]);
 	}
 
--- a/packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
+++ b/packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
@@ -2,6 +2,7 @@ import type { MigrationContext, ReversibleMigration } from '../migration-types';
 
 export class AddLangsmithIdsToInstanceAiRunSnapshots1777100000000 implements ReversibleMigration {
 	async up({ schemaBuilder: { addColumns, column } }: MigrationContext) {
+		// LangSmith run/trace IDs are UUID v4 (e.g. "f47ac10b-58cc-4372-a567-0e02b2c3d479").
 		await addColumns('instance_ai_run_snapshots', [
 			column('langsmithRunId').varchar(36),
 			column('langsmithTraceId').varchar(36),
--- a/packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
+++ b/packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
@@ -1,0 +1,14 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class AddLangsmithIdsToInstanceAiRunSnapshots1777100000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { addColumns, column } }: MigrationContext) {
+		await addColumns('instance_ai_run_snapshots', [
+			column('langsmithRunId').varchar(36),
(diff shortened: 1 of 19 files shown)


packages/cli/src/modules/instance-ai/instance-ai.controller.ts
```diff
@@ -357,6 +358,26 @@ export class InstanceAiController {
...
+		req: AuthenticatedRequest,
+		_res: Response,
+		@Param('threadId') threadId: string,
+		@Param('responseId') responseId: string,
+		@Body payload: InstanceAiFeedbackRequestDto,
+	) {
+		this.requireInstanceAiEnabled();
+		await this.assertThreadAccess(req.user.id, threadId);
+		// Fire-and-forget: never surface LangSmith errors to the UI. The service
+		// logs its own failures; the .catch here guards against unhandled
```
aalises, line 373: Maybe having no .catch here makes any rejection from the findLangsmithAnchor DB call or some throw from getClient, bubble up as an unhandled rejection. It'd be worth wrapping it with a .catch((error) => this.logger.warn(...))

aalises: ah lol I see it got fixed 😆 nvm then @mutdmour was looking at an older version

mutdmour: Screenshot 2026-04-22 at 13 38 48
It already handles that within the function.

aalises (dismissed):

packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
```diff
@@ -0,0 +1,14 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class AddLangsmithIdsToInstanceAiRunSnapshots1777100000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { addColumns, column } }: MigrationContext) {
+		await addColumns('instance_ai_run_snapshots', [
+			column('langsmithRunId').varchar(36),
+			column('langsmithTraceId').varchar(36),
```
tomi: What format are these? UUID? Is 36 enough? Can you add a comment that shows an example of how an ID looks like?

mutdmour: - Format: UUID v4 (36 chars with dashes, e.g. f47ac10b-58cc-4372-a567-0e02b2c3d479)
  - varchar(36) is exactly right — UUID v4 is always 36 characters                                                                                                                                                                                  
  - LangSmith's https://github.com/langchain-ai/langsmith-sdk generates these via uuid.v4() internally; both id and trace_id are UUIDs  
  
  adding a comment now

packages/@n8n/db/src/migrations/common/1777100000000-AddLangsmithIdsToInstanceAiRunSnapshots.ts
```diff
@@ -0,0 +1,15 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class AddLangsmithIdsToInstanceAiRunSnapshots1777100000000 implements ReversibleMigration {
+	async up({ schemaBuilder: { addColumns, column } }: MigrationContext) {
+		// LangSmith run/trace IDs are UUID v4 (e.g. "f47ac10b-58cc-4372-a567-0e02b2c3d479").
```
tomi: I actually adding a `.comment('...')` on the column itself. That way it's documented on the schema itself

mutdmour: Ops. on it.

tomi (approved):
Migrations LGTM 👌 Didn't review anything else

PR #28897, opened 2026-04-22
feat(API): Add encryption key management endpoints

guillaumejacquart:
## Summary

Introduces two owner-only REST endpoints under `/api/v1/encryption/keys`, gated by a new `encryptionKey:manage` global scope:

- `GET /encryption/keys` — lists deployment keys (optionally filtered by `type`), returning id/type/algorithm/status/createdAt. Raw key material is never exposed.
- `POST /encryption/keys` — accepts `{ type: 'data_encryption' }`, server-generates a 256-bit key, encrypts it with the instance encryption key via `Cipher`, and atomically promotes it to active (deactivating the previous active key).

Includes DTO validation (via `Z.class` zod wrappers), controller + service unit tests, and an integration test covering the happy path, 400s for bad input, and 403s for non-owners. To test: authenticate as owner, `GET /encryption/keys` and `POST /encryption/keys` with `{ "type": "data_encryption" }`.

https://linear.app/n8n/issue/IAM-495

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/@n8n/api-types/src/dto/encryption/create-encryption-key.dto.ts
+++ b/packages/@n8n/api-types/src/dto/encryption/create-encryption-key.dto.ts
@@ -1,0 +1,7 @@
+import { z } from 'zod';
+
+import { Z } from '../../zod-class';
+
+export class CreateEncryptionKeyDto extends Z.class({
+	type: z.literal('data_encryption'),
+}) {}
--- a/packages/cli/src/modules/encryption-key-manager/encryption-key-manager.module.ts
+++ b/packages/cli/src/modules/encryption-key-manager/encryption-key-manager.module.ts
@@ -6,6 +6,7 @@ import { Container } from '@n8n/di';
 export class EncryptionKeyManagerModule implements ModuleInterface {
 	async init() {
 		await import('./key-manager.service');
+		await import('./encryption-key.controller');
 
 		const { EncryptionBootstrapService } = await import('./encryption-bootstrap.service');
 		await Container.get(EncryptionBootstrapService).run();
--- a/packages/cli/src/modules/encryption-key-manager/encryption-key-manager.module.ts
+++ b/packages/cli/src/modules/encryption-key-manager/encryption-key-manager.module.ts
@@ -9,13 +9,11 @@ function isKeyRotationApiEnabled(): boolean {
 @BackendModule({ name: 'encryption-key-manager', instanceTypes: ['main'] })
 export class EncryptionKeyManagerModule implements ModuleInterface {
 	async init() {
-		await import('./key-manager.service');
-
 		if (isKeyRotationApiEnabled()) {
+			await import('./key-manager.service');
 			await import('./encryption-key.controller');
+			const { EncryptionBootstrapService } = await import('./encryption-bootstrap.service');
+			await Container.get(EncryptionBootstrapService).run();
 		}
-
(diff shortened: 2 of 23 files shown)


packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -59,25 +64,52 @@ export class KeyManagerService {
...
+	}
+
+	/**
+	 * Encrypts the given plaintext value with the instance encryption key and inserts
+	 * it as a new deployment key row. If setAsActive, atomically deactivates the
+	 * previous active key; otherwise the new key is inserted as inactive.
+	 */
+	async addKey(
+		plaintextValue: string,
+		algorithm: string,
```
afitzek: I think we have a type for the supported algorithms, shall we use this here.

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
```diff
@@ -59,25 +64,52 @@ export class KeyManagerService {
...
+		setAsActive = false,
+	): Promise<DeploymentKey> {
+		const encryptedValue = this.cipher.encrypt(plaintextValue);
```
afitzek: This would encrypt the new data encryption key, with the current data encryption key, this only works for the first key.

afitzek (commented):
I think we need to be careful, when we encrypt decrypt the data encryption keys themself.

packages/cli/src/modules/encryption-key-manager/key-manager.service.ts
guillaumejacquart: make sense, thanks

packages/@n8n/api-types/src/dto/encryption/create-encryption-key.dto.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod';
+
+import { Z } from '../../zod-class';
+
+export class CreateEncryptionKeyDto extends Z.class({
+	type: z.literal('data_encryption'),
```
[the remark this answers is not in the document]
guillaumejacquart: the ticket specs changed. I'll update the ticket

packages/cli/src/modules/encryption-key-manager/encryption-key-manager.module.ts
```diff
@@ -2,11 +2,19 @@ import type { ModuleInterface } from '@n8n/decorators';
...
+function isKeyRotationApiEnabled(): boolean {
+	return process.env.N8N_ENV_FEAT_ENCRYPTION_KEY_ROTATION === 'true';
+}
+
 @BackendModule({ name: 'encryption-key-manager', instanceTypes: ['main'] })
 export class EncryptionKeyManagerModule implements ModuleInterface {
 	async init() {
 		await import('./key-manager.service');
 
+		if (isKeyRotationApiEnabled()) {
```
afitzek, line 12: Should we initialize anything when the feature flag is not set?

afitzek (commented):
One comment on the feature flag.

packages/cli/src/modules/encryption-key-manager/encryption-key-manager.module.ts
guillaumejacquart: I wrapped the whole init code with the env var check. Thanks !

afitzek (approved):

PR #28902, opened 2026-04-22
fix: Preserve node positions in `layoutWorkflowJSON` for update paths

dariacodes:
## Summary

When updating a workflow through MCP or the instance-ai deep agent, we currently run `layoutWorkflowJSON()` as if the workflow were being laid out from scratch. That is fine for new workflows, but it is the wrong behavior for update paths where nodes already have intentional positions. In practice, that can move related elements such as sticky notes after a simple content edit:
image

The key idea is that the update tool should just update the workflow content - it shouldn’t try to clean things up or rearrange the canvas on its own. If we do want cleanup or normalization, that should be a separate tool that people can choose to use explicitly.

This PR makes the update path preserve existing layout instead of recalculating it blindly:
- `layoutWorkflowJSON()` now accepts partially positioned workflows via `LayoutableWorkflowJSON` / `LayoutableNodeJSON`
- when `preservePositions` is enabled, explicitly positioned nodes are kept in place and reused during dagre layout calculations
- sticky notes are reanchored relative to those preserved nodes instead of drifting because the layout pass treated everything as new

### How to test

1. Use an MCP client or the instance-ai deep agent to update a workflow that already has manually positioned nodes and a sticky note.
2. Change content only, for example update a sticky note's text or another node field without asking for layout changes.
3. Confirm the content change is applied, existing positioned nodes stay where they are, and sticky notes remain anchored correctly.
4. Confirm that new workflows or nodes without positions still receive automatic layout.

Closes [ADO-5104](https://linear.app/n8n/issue/ADO-5104/bug-sticky-note-position-is-always-changed-when-updated-through-layout)

## Review / Merge checklist

- [x] Tests included. 


MiloradFilipovic (dismissed):
Nice work!

PR #28905, opened 2026-04-22
fix(editor): Copy to editor not using pinned data from the execution

### LINKED ISSUE #28439: workflow processes only one item when importing execution data with multiple JSON items
### Bug Description

When importing execution data that contains multiple JSON items into a drive trigger based workflow, the workflow only executes once instead of processing each item individually.

Image

### To Reproduce

1. Import execution data that contains multiple JSON items (e.g., 2 items)

2. Execute the workflow using the imported data

### Expected behavior

workflow should process both json instead one one.

### Debug Info

# Debug info

## core

- n8nVersion: 2.15.1
- nodeJsVersion: 24.13.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 120 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2026-04-14T02:00:23.272Z

### Operating System

Ubuntu 24.04

### n8n Version

2.15.1

### Node.js Version

24.13.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

RomanDavydchuk:
## Summary

When execution data is copied into the editor, trigger pin data is restored in memory but the workflow is not marked as dirty. Because manual run executes the workflow stored in the database, the restored pin data is not saved and is ignored, so trigger-based runs fall back to default manual behavior (processing a single item, same as running a manual test execution for a trigger). 

This PR marks the workflow state as dirty after importing execution data, so it's saved to the DB and the workflow uses pinned data properly

#### To reproduce the issue or test the fix
1. Create a workflow with a trigger that can be triggered in manual execution (like Google Drive Trigger) and any node after it (for example Noop node)
2. Activate the workflow
3. Trigger the workflow so it produces more than 1 item
4. Copy the execution to editor
5. Run the workflow manually
6. Observe that:
     - **Before the fix:** the node after the trigger returns only 1 item and this matches what you would get if you run the workflow manually without any pinned data
     - **After the fix:** the node returns items that are identical to the data in the prod execution

https://linear.app/n8n/issue/NODE-4875/community-issue-workflow-processes-only-one-item-when-importing
Closes #28439

## Review / Merge checklist

- [x] Tests included. 


michael-radency (approved):
LGTM

PR #28907, opened 2026-04-22
test(core): Add E2E coverage for Instance AI workflow setup apply actions (no-changelog)

aalises:
## Summary

Adds end-to-end Playwright coverage for the Instance AI workflow-setup confirmation UI — the **confirm-mode apply**, **confirm-mode defer**, **full-wizard apply**, and **full-wizard Later / partial apply** flows triggered from the setup card. Also pins the backend contract at the unit layer so a future regression in `resumeData` → `applyNodeChanges` → persisted workflow is caught outside of e2e.

Both specs run in the `sqlite:e2e` container project and stay green across 10 consecutive sequential runs (`--repeat-each=5 --workers=1`).

## Test cases

| # | Spec | Test | Flow | What it proves |
|---|------|------|------|----------------|
| 1 | `instance-ai-workflow-setup.spec.ts` | `should render confirm mode when a single valid credential auto-applies` | Seed workflow with a Slack node + one matching credential → AI opens setup → review details → Apply | Confirm mode renders when a single valid credential can auto-apply, the "review details" link expands the full card, and Apply persists the auto-resolved credential to the workflow |
| 2 | `instance-ai-workflow-setup.spec.ts` | `should defer the whole setup when Later is clicked in confirm mode` | Same confirm-mode seed → AI opens setup → Later | Later in confirm mode transitions to the deferred state and does NOT mutate the workflow — the Slack node still carries no credential |
| 3 | `instance-ai-workflow-setup-actions.spec.ts` | `should apply parameter and credential edits and persist them to the workflow` | Seed workflow with an empty HTTP URL + Slack node pre-credentialed → AI opens setup → user fills URL → Apply | Filling the required param flips the card to the "Complete" step-check badge, and Apply persists both the parameter edit (HTTP `url`) and the existing credential assignment to the workflow JSON |
| 4 | `instance-ai-workflow-setup-actions.spec.ts` | `should partially apply completed cards when Later is clicked on the last step` | Seed same workflow without a Slack cred → AI opens setup → user fills HTTP URL → Next → Prev → Next → Later on Slack card | Prev navigation returns to the HTTP card with the filled value intact (wizard keeps `paramValues` as the source of truth across step changes), and partial apply persists the completed card's edits (HTTP `url`) while the untouched Slack node keeps its empty credentials |

All four specs exercise the real `workflows(setup)` suspend/resume contract end-to-end — the orchestrator LLM, the `POST /instance-ai/confirm` endpoint, `applyNodeChanges` → `workflowService.update`, the SSE tool-result stream back to the FE, and the post-apply workflow GET.

## Behind the scenes

A few things had to line up to make the apply path observable and deterministic from a Playwright harness:

- **Wait on the applying spinner, not the setup-card.** The wizard template in `InstanceAiWorkflowSetup.vue` unmounts synchronously the moment `isApplying` flips true, so waiting for the setup card to hide resolves *before* the backend save — the subsequent workflow GET would race the commit. The `instance-ai-workflow-setup-applying` spinner stays visible for the full apply window (POST → tool-result SSE → save commit), so watching it appear and disappear guarantees persistence before we assert.
- **Debounced parameter commits.** `ParameterInput` debounces `@update:model-value` by 200 ms, so a short `waitForTimeout(300)` after `.fill()` lets the value propagate into `paramValues` before Apply fires.
- **Per-test unique workflow + credential names.** The orchestrator's `list-workflows` tool is sensitive to same-named workflows across parallel workers during recording — unique names prevent the LLM from slipping into a disambiguation response.
- **Slack `users.profile.get` success mock in the fixture.** The FE's `buildNodeCredentials` intentionally excludes any credential whose test has failed. Seed tokens in these tests are fake, so the fixture installs a success mock on mockserver *before* loading recordings — mockserver's FIFO matching serves it ahead of any recorded Slack response. Same shape as the existing LLM mocks.

## Regression tests

Two unit tests pin the tool-layer contract:

- `applyNodeChanges` — merged `nodeParameters` values land on the saved workflow JSON.
- `workflows(setup)` resume path — `resumeData.nodeParameters` is forwarded to `applyNodeChanges` unchanged.

Plus three `credGroupKey` cases covering same-credential-type node grouping, HTTP Request distinct-URL splits, and cross-type isolation.

## Related Linear tickets

https://linear.app/n8n/issue/INS-113

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/instance-ai/src/tools/__tests__/workflows.tool.test.ts
```diff
@@ -363,6 +363,46 @@ describe('workflows tool', () => {
...
+				connections: {},
+			});
+
+			const tool = createWorkflowsTool(context, 'full');
+			await tool.execute!({ action: 'setup', workflowId: 'wf1' }, {
+				agent: {
+					resumeData: {
+						approved: true,
+						action: 'apply',
+						nodeParameters: { 'HTTP Request': { url: 'https://example.com/api' } },
```
scdekov, line 397: Should we set the url to a different value here? It's a little confusing that the old(getAsWorkflowJSON) and the new values are the same

packages/testing/playwright/tests/e2e/instance-ai/instance-ai-workflow-setup-actions.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+			// HTTP card remains (URL still missing), so the wizard is a single
+			// step — "1 of 1".
+			await expect(n8n.instanceAi.getWorkflowSetupStepCounter()).toContainText('1 of 1');
+
+			// Fill the HTTP URL. n8n's ParameterInput debounces updates by
+			// 200ms; the wait lets the value propagate to `paramValues` before
+			// we proceed so the apply payload includes the URL.
+			await expect(n8n.instanceAi.getWorkflowSetupParameterIssues()).toBeVisible();
+			await n8n.instanceAi.getWorkflowSetupParameterInput().fill('https://example.com/api');
+			await n8n.page.waitForTimeout(300);
```
scdekov: Do we need this? Seems like the next expect has a timeout

scdekov (dismissed):

packages/@n8n/instance-ai/src/tools/__tests__/workflows.tool.test.ts
aalises: done!

packages/testing/playwright/tests/e2e/instance-ai/instance-ai-workflow-setup-actions.spec.ts
aalises: true, removed!

scdekov (approved):

PR #28909, opened 2026-04-22
feat(core): Accept token-exchanged scoped JWTs on the instance MCP server

afitzek:
The instance MCP server previously authenticated callers with an MCP API key
or an MCP OAuth access token. This PR extends MCP authentication to also
accept scoped JWTs issued by the token-exchange module (RFC 8693), so
embedding partners can drive MCP tools on behalf of a user with the same
delegated tokens they already use against the public API.

To keep a single authentication pipeline, `McpServerApiKeyService.verifyApiKey`
now delegates to `AuthStrategyRegistry` via a new `buildTokenGrant(token, options?)`
primitive on `AuthStrategy`. `ApiKeyAuthStrategy` and `ScopedJwtStrategy` both
implement it; the MCP call site asks the registry for a grant scoped to
`audience: 'mcp-server-api'`. `UserWithContext` gains an optional `actor` so
delegated calls carry both the acting principal and the subject they act on
behalf of.

Relevant context from the token-exchange initiative (see memory & Notion RFC):

- Phase 2b introduced `AuthStrategyRegistry` + the `TokenGrant` model for the
  public API.
- This ticket (IAM-574, Phase 3) reuses that pipeline on the MCP surface so the
  MCP server and public API share one auth chain.
- The same scoped JWT a partner mints via `/auth/oauth/token` now authenticates
  against `/mcp-server/http`.

- **Separation of extraction from verification**: `authenticate(req)` becomes a
  thin wrapper that extracts the token and calls `buildTokenGrant(token)`. The
  new primitive is reusable from any surface that has already parsed the
  bearer token (MCP middleware does this itself).
- **`audience`/`issuer` options on `AuthStrategyOptions`**: let the same
  `ApiKeyAuthStrategy` service both `audience: 'public-api'` (default) and
  `audience: 'mcp-server-api'` (MCP) without duplication. The DB filter and
  `jwtService.verify` both enforce the requested audience, so an MCP caller
  cannot present a public-API key and vice versa.
- **Registry registration is now unconditional**: `ApiKeyAuthStrategy`
  registration moved from `loadPublicApiVersions()` (only runs when
  `isApiEnabled()`) to `Server.configure()`. Without this, disabling the
  public REST API would 401 every MCP API key. The test harness
  (`test-server.ts`) mirrors the new placement.
- **Delegation plumbing**: `verifyApiKey` returns `{ user: actor ?? subject, actor }`.
  `user` is the acting principal (used for downstream req.user), `actor` is
  kept available for telemetry/audit plumbing.

### How to test manually

Prerequisites: an n8n instance with the token-exchange feature flag enabled
(`N8N_ENV_FEAT_TOKEN_EXCHANGE=true`) and a trusted external key configured
so you can mint scoped JWTs via `POST /auth/oauth/token`.

1. **MCP API key path (AC1) — still works**:
   - Obtain your MCP API key from the UI (Settings → Personal → MCP Server API Key).
   - `curl -X POST http://localhost:5678/mcp-server/http -H "Authorization: Bearer <MCP_API_KEY>" -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}'`
   - Expect 200 / JSON-RPC response.
   - Present a **public-API** key to the same endpoint — expect 401
     (audience mismatch).

2. **Scoped JWT path (AC2) — new**:
   - Mint a scoped JWT by exchanging a signed subject token against
     `/auth/oauth/token` (RFC 8693 `urn:ietf:params:oauth:grant-type:token-exchange`).
   - Call MCP with `Authorization: Bearer <SCOPED_JWT>` — expect 200.
   - Mint a scoped JWT with an `act.sub` claim mapping to another user. Call
     MCP, confirm the tools act as the actor.

3. **Regression path — public API disabled**:
   - Set `N8N_PUBLIC_API_DISABLED=true` (or otherwise disable the public API).
   - Restart, repeat step 1 — MCP API keys must still authenticate (this
     used to fail before the registration move). Automated regression
     coverage: `mcp-api-key.service.public-api-disabled.integration.test.ts`.

### Automated test coverage

- `mcp-api-key.service.integration.test.ts` — AC1, AC2 (with/without actor),
  rejection matrix (garbage, public-API key, deleted row, expired JWT,
  disabled subject/actor, unknown subject), and a defense-in-depth test for
  a DB row whose stored JWT has a tampered `aud` claim.
- `mcp-api-key.service.public-api-disabled.integration.test.ts` — regression
  guard; verified to fail when the registration is moved back under an
  `isApiEnabled()` guard.
- `mcp-api-key.service.test.ts` — unit coverage of the registry delegation
  and error mapping.
- `mcp-server-middleware.service.test.ts` — new delegated scoped-JWT case
  proving MCP middleware sets `req.user` to the actor.
- `scoped-jwt.strategy(.integration).test.ts` — refactored to test
  `buildTokenGrant` directly + the `authenticate` wrapper.
- `api-key-auth.strategy.integration.test.ts` — added `audience` option
  matrix (default rejects MCP key, `mcp-server-api` rejects public key,
  `mcp-server-api` accepts MCP key).
- `auth-strategy.registry.test.ts` — new `buildContextFromToken` suite.

## Related Links

- Linear ticket: https://linear.app/n8n/issue/IAM-574
- Token-exchange RFC: https://www.notion.so/n8n/32e5b6e0c94f800fa2cce56f39929a3e
- Design doc: https://www.notion.so/n8n/3305b6e0c94f808c9a18f62bb72fb618

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/mcp/__tests__/mcp-api-key.service.integration.test.ts
+++ b/packages/cli/src/modules/mcp/__tests__/mcp-api-key.service.integration.test.ts
@@ -57,7 +57,7 @@ describe('McpServerApiKeyService.verifyApiKey (integration)', () => {
 		await testDb.terminate();
 	});
 
-	it('accepts a valid MCP API key and returns the owning user (AC1)', async () => {
+	it('accepts a valid MCP API key and returns the owning user', async () => {
 		const owner = await createOwner();
 		const apiKey = await createMcpApiKey(owner);
 
@@ -68,7 +68,7 @@ describe('McpServerApiKeyService.verifyApiKey (integration)', () => {
 		expect(result.context).toBeUndefined();
 	});
 
-	it('accepts a scoped JWT and returns the subject when no actor is present (AC2)', async () => {
+	it('accepts a scoped JWT and returns the subject when no actor is present', async () => {
 		const subject = await createOwner();
 
 		const result = await service.verifyApiKey(makeScopedJwt(jwtService, subject.id));
@@ -78,7 +78,7 @@ describe('McpServerApiKeyService.verifyApiKey (integration)', () => {
 		expect(result.context).toBeUndefined();
 	});
 
-	it('accepts a scoped JWT with delegation and returns the actor as user (AC2)', async () => {
+	it('accepts a scoped JWT with delegation and returns the actor as user', async () => {
 		const subject = await createOwner();
 		const actor = await createMember();
 
--- a/packages/cli/src/modules/mcp/__tests__/mcp-api-key.service.integration.test.ts
+++ b/packages/cli/src/modules/mcp/__tests__/mcp-api-key.service.integration.test.ts
@@ -1,0 +1,177 @@
+import { testDb } from '@n8n/backend-test-utils';
+import type { User } from '@n8n/db';
+import { ApiKey, ApiKeyRepository, UserRepository } from '@n8n/db';
+import { Container } from '@n8n/di';
+import { randomString } from 'n8n-workflow';
+
+import { ScopedJwtStrategy } from '@/modules/token-exchange/services/scoped-jwt.strategy';
+import { TOKEN_EXCHANGE_ISSUER } from '@/modules/token-exchange/token-exchange.types';
+import { ApiKeyAuthStrategy } from '@/services/api-key-auth.strategy';
+import { AuthStrategyRegistry } from '@/services/auth-strategy.registry';
+import { JwtService } from '@/services/jwt.service';
+import { createMember, createOwner, createOwnerWithApiKey } from '@test-integration/db/users';
+
+import { McpServerApiKeyService } from '../mcp-api-key.service';
+
+function makeScopedJwt(
+	jwtService: JwtService,
+	sub: string,
+	opts: { actSub?: string; expired?: boolean } = {},
+): string {
+	const now = Math.floor(Date.now() / 1000);
+	return jwtService.sign({
+		iss: TOKEN_EXCHANGE_ISSUER,
+		sub,
+		...(opts.actSub && { act: { sub: opts.actSub } }),
+		iat: opts.expired ? now - 7200 : now,
+		exp: opts.expired ? now - 1 : now + 3600,
+		jti: `jti-${randomString(8)}`,
+	});
+}
+
+async function createMcpApiKey(user: User): Promise<string> {
+	return (await Container.get(McpServerApiKeyService).createMcpServerApiKey(user)).apiKey;
+}
+
+describe('McpServerApiKeyService.verifyApiKey (integration)', () => {
+	let service: McpServerApiKeyService;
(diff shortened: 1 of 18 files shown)


packages/cli/src/modules/mcp/mcp-api-key.service.ts
```diff
@@ -66,36 +68,26 @@ export class McpServerApiKeyService {
...
+					user: tokenGrant.actor ?? tokenGrant.subject,
+					actor: tokenGrant.actor,
 				};
 			}
-			return { user };
+
+			return {
+				user: null,
+				context: {
+					reason: 'invalid_token',
```
guillaumejacquart, line 87: Are we sure that this reason change will not impact existing flows ?

guillaumejacquart (commented):
1 comment

packages/cli/src/modules/mcp/__tests__/mcp-api-key.service.integration.test.ts
```diff
@@ -0,0 +1,177 @@
...
+		const apiKey = await createMcpApiKey(owner);
+
+		const result = await service.verifyApiKey(apiKey);
+
+		expect(result.user?.id).toBe(owner.id);
+		expect(result.actor).toBeUndefined();
+		expect(result.context).toBeUndefined();
+	});
+
+	it('accepts a scoped JWT and returns the subject when no actor is present (AC2)', async () => {
```
phyllis-noester: We should remove the (AC2) parts of the tests IMO

packages/cli/src/modules/mcp/mcp-api-key.service.ts
```diff
@@ -66,36 +68,26 @@ export class McpServerApiKeyService {
...
-				apiKeys: {
-					apiKey,
-					audience: API_KEY_AUDIENCE,
-				},
-			},
-			relations: ['role'],
-		});
-	}
-
 	async verifyApiKey(apiKey: string): Promise<UserWithContext> {
```
phyllis-noester, line 71: should we rename this/move it out of the api key service to make avoid mixing different concerns?

packages/cli/src/modules/mcp/mcp-api-key.service.ts
afitzek: I think this change should be fine, it does not mean, that there was no user found. We basically reduce the surface of information we are handing out.

packages/cli/src/modules/mcp/mcp-api-key.service.ts
afitzek: I think this is fine here. I thought about moving this part to the middleware of the mcp module, but I think this mcp api key service is fine to also handle verification of its own api keys. So this function allows us to specify which api keys are verified by this mcp specific api key service.

phyllis-noester (approved):
🚀

PR #28910, opened 2026-04-22
feat(agents): Stream tool-call lifecycle for immediate agent UI feedback

bjorger:
## Summary

- Stream three stages of each tool call (selected / executing / done) so the
  UI updates as the agent runs instead of only at the result.
- Pass `toolCallId` through so early and late stream events correlate and
  don't produce duplicate tool cards.
- Rewrite the custom-tool builder prompt around one principle ("handlers
  run in a V8 isolate — treat them as pure functions") and document
  `ctx.suspend` for HITL.
- Stop `build_custom_tool` from mutating `schema.tools`; registering the
  tool in the config is the agent's job via `patch_config` / `write_config`.

## Review / Merge checklist

- [ ] Tests included. 


mike12345567 (approved):
LGTM!

PR #28911, opened 2026-04-22
fix(MySql Node): Fix table name filtering

### LINKED ISSUE #27499: Unable to filter mySQL table list in node editor
### Bug Description

Image

Having issues that searching for a table in the workflow node editor, it doesn't seem to filter any of our tables, we've updated to 2.13.2, apologies I wasn't able to find this logged in the list

### To Reproduce

1. Go Workflow editor
2. Add MSQL Node
3. Select drop down list of tables, enter a value to search for tables, it does not seem to be filterable

### Expected behavior

Table list to be filtered

### Debug Info

Not sure

### Operating System

AlmaLinux v9.7.0 STANDARD kvm

### n8n Version

2.13.2

### Node.js Version

22

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

dlavrenuek:
## Summary

Takes provided filter options when selecting the table in MySql node

https://linear.app/n8n/issue/NODE-4882/community-issue-unable-to-filter-mysql-table-list-in-node-editor
Fixes #27499

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/nodes-base/nodes/MySql/v2/methods/listSearch.ts
+++ b/packages/nodes-base/nodes/MySql/v2/methods/listSearch.ts
@@ -3,7 +3,10 @@ import type { IDataObject, ILoadOptionsFunctions, INodeListSearchResult } from '
 import type { MysqlNodeCredentials } from '../helpers/interfaces';
 import { createPool } from '../transport';
 
-export async function searchTables(this: ILoadOptionsFunctions): Promise<INodeListSearchResult> {
+export async function searchTables(
+	this: ILoadOptionsFunctions,
+	filter?: string,
+): Promise<INodeListSearchResult> {
 	const credentials = await this.getCredentials<MysqlNodeCredentials>('mySql');
 
 	const nodeOptions = this.getNodeParameter('options', 0) as IDataObject;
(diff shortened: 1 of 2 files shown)


packages/nodes-base/nodes/MySql/v2/methods/listSearch.ts
```diff
@@ -13,8 +16,13 @@ export async function searchTables(this: ILoadOptionsFunctions): Promise<INodeLi
...
 		const connection = await pool.getConnection();
 
-		const query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE table_schema = ?';
-		const values = [credentials.database];
+		let query = 'SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE table_schema = ?';
+		const values: string[] = [credentials.database];
+
+		if (filter) {
+			query += ' AND TABLE_NAME LIKE ?';
+			values.push(`%${filter}%`);
```
BerniWittmann, line 24: Only issue i found is wapping it like this retains matching characters in the filter. 
e.g. filter = `user_data` would also find tables named `userXdata`
which might be confusing for the user when the expectation is to only return one table.

BerniWittmann (commented):
Just one thing i found, where I'm unsure whether it is intentional behavior or not.

packages/nodes-base/nodes/MySql/v2/methods/listSearch.ts
dlavrenuek: The same is true for `%` since we are using `LIKE`. do you think that we should escape these characters?

BerniWittmann: From what i understand this method is only used for selecting the table in the node right now:
Bildschirmfoto 2026-04-23 um 14 42 18
As the user anyhow has to click the correct table there, i would argue it is okay to keep them unescaped aka as wildcards for now.

BerniWittmann (approved):

PR #28914, opened 2026-04-22
fix(core):  Fixes user search functionality in ProjectSettings for pr…

sandra0503:
Allow project admins to list all users

## Summary

**Problem**: Project admins could not invite users to their projects. When a project admin opened Project Settings and tried to search for users to add as members, the backend returned a 403

**Fix**:
- Backend: Instead of throwing a ForbiddenError for non-admins, the controller now checks whether the requesting user has project:update scope on any project. If they do, the request is allowed.
- Frontend: `shouldFetchAllUsers` computed property checks if the user is a global admin/owner or has canUpdateProject scope. If true, the projectId filter is omitted so the search returns all instance users (enabling invitations).

Fixes [LIGO-479](https://linear.app/n8n/issue/LIGO-479/cant-invite-users-to-a-project)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/controllers/__tests__/users.controller.test.ts
+++ b/packages/cli/src/controllers/__tests__/users.controller.test.ts
@@ -1,3 +1,4 @@
+import type { UsersListFilterDto } from '@n8n/api-types';
 import type { AuthenticatedRequest, User, UserRepository } from '@n8n/db';
 import { mock } from 'jest-mock-extended';
 import type { Response } from 'express';
@@ -7,6 +8,7 @@ import type { JwtService } from '@/services/jwt.service';
 import type { ProvisioningService } from '@/modules/provisioning.ee/provisioning.service.ee';
 import type { ProjectService } from '@/services/project.service.ee';
 import type { UrlService } from '@/services/url.service';
+import { ForbiddenError } from '@/errors/response-errors/forbidden.error';
 import { NotFoundError } from '@/errors/response-errors/not-found.error';
 
 import { UsersController } from '../users.controller';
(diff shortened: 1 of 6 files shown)


packages/cli/src/controllers/users.controller.ts
```diff
@@ -128,9 +128,14 @@ export class UsersController {
 				throw new NotFoundError('Project not found');
 			}
 		} else if (!['global:owner', 'global:admin'].includes(req.user.role.slug)) {
-			throw new ForbiddenError(
-				'Listing all users is limited to instance administrators. Filter by project to list project members.',
-			);
+			// Project admins need to search all users to invite them into their projects
+			const isProjectAdmin =
+				(await this.projectService.getProjectIdsWithScope(req.user, ['project:update'])).length > 0;
```
geemanjs, line 133: Its more a nit - but it would be more performant to do a "count" at the db level for this.

packages/cli/src/controllers/users.controller.ts
```diff
@@ -128,9 +128,14 @@ export class UsersController {
...
+			if (!isProjectAdmin) {
+				throw new ForbiddenError(
+					'Listing all users is limited to instance administrators and project admins. Filter by project to list project members.',
+				);
+			}
```
alielkhateeb, line 138: This whole part is duplicated in `users.handler`, and right now the same bug would exist in public API's `getUsers`, right?

I suggest we introduce a `assertGetUsersAccess` or similar in `user.service` (NOT the one in public-api module 😂) and consume it in both, this would avoid future access issues from being missed in one or the other, does that make sense?

packages/cli/src/controllers/__tests__/users.controller.test.ts
```diff
@@ -67,6 +69,81 @@ describe('UsersController', () => {
 		});
 	});
 
+	describe('listUsers', () => {
```
geemanjs, line 72: I think this is probably just my opinion but I really hate these kind of tests 😆. I much prefer things like this done with "integration" level with a real db and calling the controller directly 👀.

geemanjs (approved):

packages/cli/src/controllers/users.controller.ts
sandra0503: You're right. We should have the same behaviour for public API. Would it be okay to do this in a follow-up? To include this fix in backports today and unblock the issue in the UI. Or should we do it in this PR?

alielkhateeb: @sandra0503 Yes sure, I would have of course preferred to backport this too, but it is much less likely for project admins who are not owners to use the public API to get users

packages/cli/src/controllers/__tests__/users.controller.test.ts
sandra0503: I see what you're mean as they're heavily mocked. I will merge for now but move the permission logic into a service in a follow-up, okay?

alielkhateeb: @geemanjs Not just your opinion 😂 I agree with you!

PR #28915, opened 2026-04-22
feat: Add interactive tools to builder agent

yehorkardash:
## Summary

This PR adds UI HITL elements to builder agent that allow user to select credentials for LLM and nodes. It also adds Q&A tool
The PR also extends a variety of chunks that streaming events could send, to simplify turn tracking on FE

| Area | Files | +/− | Total churn |
|---|---:|---:|---:|
| editor-ui agents feature | 17 | +2396 / −238 | **2634** |
| cli agents module | 16 | +984 / −322 | **1306** |
| @n8n/agents (SDK runtime) | 15 | +230 / −291 | **521** |
| @n8n/api-types (wire) | 11 | +314 / −1 | **315** |

## Review / Merge checklist

- [x] Tests included. 


yehorkardash:
@claude review

packages/@n8n/agents/src/runtime/agent-runtime.ts
```diff
@@ -125,9 +125,15 @@ type ToolCallOutcome =
 	| {
 			outcome: 'success';
 			toolEntry: ToolResultEntry;
+			/**
+			 * Output as the LLM sees it (after `toModelOutput`). Same as
+			 * `toolEntry.output` when no `toModelOutput` transform is configured.
+			 * Surfaced on the `tool-result` wire chunk so consumers see what the
+			 * LLM saw (rather than the larger raw output).
+			 */
+			modelOutput: unknown;
```
mike12345567, line 134: Is there anyway this could be typed? Is it just a string or does it have a structure of some sort? Is this the provider specific output or something like that?

packages/cli/src/modules/agents/builder/agents-builder.service.ts
```diff
@@ -115,4 +208,35 @@ export class AgentsBuilderService {
...
+	// Private — open-suspension lookup
+	// ---------------------------------------------------------------------------
+
+	/**
+	 * Return the parsed state of the most recent non-expired suspended
+	 * checkpoint for this agent, or `null` if there isn't one. Each pending
+	 * tool call inside the state already carries its own `runId`, so callers
+	 * don't need a separate runId from this helper.
+	 */
+	async findOpenCheckpoint(agentId: string): Promise<SerializableAgentState | null> {
```
mike12345567, line 222: This just returns the latest agent checkpoint for an agent, should this be user scoped?

packages/cli/src/modules/agents/agents.controller.ts
```diff
@@ -399,61 +420,80 @@ export class AgentsController {
...
+		const agent = await this.agentsService.findById(agentId, projectId);
+		if (!agent) throw new NotFoundError(`Agent "${agentId}" not found`);
+
+		const credentialProvider = new AgentsCredentialProvider(this.credentialsService, projectId);
+
+		const { send } = initSseStream(res);
+
+		try {
+			const suspended = await pumpChunks(
+				this.agentsBuilderService.resumeBuild(
```
mike12345567, line 479: Should this be scoped to a user in someway as well, if you have the `runId` and `agentId` you could theoretically resume someone else's session and get information back.

mike12345567 (commented):
Just a couple of questions!

packages/cli/src/modules/agents/builder/agents-builder.service.ts
yehorkardash: Currently builder agent memory is not scoped by user, which means agent builder chat history and checkpoints are shared between all users. Makes sense to scope builder agent conversation by user id, but that's for another PR

packages/cli/src/modules/agents/agents.controller.ts
yehorkardash: Yeah, we should start scoping agent builder conversations by user id, but I didn't focus on this change in the PR

packages/cli/src/modules/agents/builder/agents-builder.service.ts
mike12345567: Ah yeah ok makes sense, could we create a linear ticket for this - just to track its something we need to sort out!

packages/@n8n/agents/src/runtime/agent-runtime.ts
yehorkardash: It comes from tool output and `toModelOutput` returning `unknown`, because tools usually don't have schema of what they can return.

mike12345567: Ah - yeah we could probably do something about that - but not for now!

mike12345567 (approved):

PR #28920, opened 2026-04-22
fix(editor): Use current pin data for fetched executions (no-changelog)

keshav-005:
## Summary
- use the current workflow document pin data when hydrating fetched executions after a run finishes
- add a regression test covering stale workflow-store pin data versus current document-store pin data

## Test plan
- corepack pnpm exec vitest run src/app/composables/usePushConnection/handlers/executionFinished.test.ts
- verified the new fetchExecutionData regression passes in the targeted handler test file

Fixes n8n-io/n8n#28109


ivov:
Thanks for the PR. The same change landed on master independently as part of #29395 released in n8n@2.20.0. Closing as superseded, appreciate the contribution.

PR #28921, opened 2026-04-22
fix(ai-builder): Clarify Switch onCase indices and SplitInBatches nesting

luka-zivkovic:
## Summary

Two examples in the workflow-builder's SDK-reference prompt were wrong. Both made the LLM produce code that wouldn't compile. Both live in the same file: `packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts`.

### Switch example

The old example used string keys (`onCase('urgent', ...)`) and finished with `onDefault(archive)`. Neither exists. `onCase` takes a numeric positional index — entry `0` is the first rule, entry `1` is the second, and so on. There is no default method; a fallback is just another rule whose condition matches everything.

The example now uses numeric indices, adds a third rule as the catch-all, and has a one-line note above it explaining the index-to-`outputKey` mapping.

### SplitInBatches example

`SplitInBatchesBuilder` isn't a valid branch target — you can't pass it into `onTrue`, `onFalse`, or `onCase`. The prompt never said so, and the LLM kept trying. The fix is to branch *inside* the batch loop instead of around it:

​```js
sibNode
  .onEachBatch(prepareItem.to(ifElse.onTrue(actionA).onFalse(actionB)).to(nextBatch(sibNode)))
  .onDone(finalizeResults);
​```

Added that snippet and a short "Nesting constraint" note above the existing example.

## How this came up

A RAG chatbot build in `instance-ai` burned a lot of LLM calls across two builder sub-agents before giving up. Reading the iteration logs, the same two errors kept bouncing back:

- `Type 'string' is not assignable to type 'number'` on `onCase`, then attempts to call a non-existent `onDefault`.
- `SplitInBatchesBuilder is not assignable to type 'IfElseTarget'` every time the builder tried to batch inside a conditional.

The builder was faithfully copying broken prompt examples. Nothing in the SDK was wrong — just the reference material. Fixing the two examples closes the loop without touching any runtime code.


packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -226,6 +230,18 @@ export default workflow('id', 'name')
 </parallel_execution>
 
 <batch_processing>
+
+**Nesting constraint:** \`SplitInBatchesBuilder\` is NOT a valid \`IfElseTarget\` or \`SwitchCaseTarget\` — you cannot pass a split-in-batches builder into \`onTrue\`, \`onFalse\`, or \`onCase\`. Do not batch inside a conditional; batch first, branch per-item inside \`.onEachBatch(...)\` via a \`NodeChain\`:
```
mutdmour: This is pretty confusing. Not clear to me what the issue is.
Anyways, if this is a real blocker, I would add a validation error here instead of confusing "Do not" prompt. Also in general we should aim for positive prompts. 

Also LLM should not need to know SplitInBatchesBuilder. Also split in batches can come after true/false branch from if.

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -193,9 +197,9 @@ const routeByPriority = switchCase({
 export default workflow('id', 'name')
   .add(startTrigger)
   .to(routeByPriority
-    .onCase('urgent', processUrgent.to(notifyTeam.to(escalate)))
-    .onCase('normal', processNormal)
-    .onDefault(archive));
+    .onCase(0, processUrgent.to(notifyTeam.to(escalate)))
```
mutdmour, line 199: good catch

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -173,7 +173,10 @@ export default workflow('id', 'name')
 
 <multi_way_routing>
 
-Switch rules use \`rules.values\` (NOT \`rules.rules\`). Each rule needs \`outputKey\` and a complete \`conditions\` object:
+Switch rules use \`rules.values\` (NOT \`rules.rules\`). Each rule needs \`outputKey\` and a complete \`conditions\` object.
```
mutdmour: we have a validation rule for switch/if etc.  can we expand there instead of prompt here.

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -173,7 +173,10 @@ export default workflow('id', 'name')
...
+
+\`onCase(index, target)\` takes a **numeric index** — entry 0 targets the first rule's \`outputKey\`, entry 1 the second, and so on. String keys do not compile. To cover a fallback route, add a final rule whose condition matches everything and wire it via its positional index.
```
mutdmour: I would drop the negative example there. numeric index is clear enough

mutdmour: for the fallback route, is not that what fallback output is for? we should have a separate example for that

luka-zivkovic: Good call. I removed the catch-all rule workaround and added a dedicated fallback example right after it. Shape matches the fluent-builder pattern in the workflow-builder.test.ts

luka-zivkovic: Removed

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
luka-zivkovic: Yes, i agree. But i think that i would keep this PR scoped to the prompt fix, and maybe open a follow up ticket for the validator work

mutdmour: can we remove `Switch rules use \`rules.values\` (NOT \`rules.rules\`)`? that should be caught in the validator

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -193,9 +196,35 @@ const routeByPriority = switchCase({
...
+\`\`\`javascript
+const routeByStatus = switchCase({
+  version: 3.2,
+  config: {
+    name: 'Route by Status',
+    parameters: {
+      rules: {
+        values: [
+          { outputKey: 'active', conditions: { options: { caseSensitive: true, leftValue: '', typeValidation: 'strict' }, conditions: [{ leftValue: '={{ $json.status }}', operator: { type: 'string', operation: 'equals' }, rightValue: 'active' }], combinator: 'and' } },
+          { outputKey: 'inactive', conditions: { options: { caseSensitive: true, leftValue: '', typeValidation: 'strict' }, conditions: [{ leftValue: '={{ $json.status }}', operator: { type: 'string', operation: 'equals' }, rightValue: 'inactive' }], combinator: 'and' } },
```
mutdmour, line 214: should not we be using `expr()` instead of `={{}}`?

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -226,6 +255,18 @@ export default workflow('id', 'name')
 </parallel_execution>
 
 <batch_processing>
+
+To branch per item inside a batch, nest the \`ifElse\` inside \`.onEachBatch(...)\` via a \`NodeChain\`:
```
mutdmour: LLM is not aware of NodeCahin

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
```diff
@@ -226,6 +255,18 @@ export default workflow('id', 'name')
...
+
+\`\`\`javascript
+// To branch per batch item, chain the conditional inside onEachBatch:
+sibNode
+  .onEachBatch(prepareItem.to(ifElse.onTrue(actionA).onFalse(actionB)).to(nextBatch(sibNode)))
```
mutdmour: why is this needed? `onEachBatch(ifElse)` is buggy? it should work AFAIK

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
luka-zivkovic: I noticed that the LLM was generating workflows where a split-in-batches loop was placed inside an if-else branch and it was basically using the loop itself as the true or false target. This fix wanted to basically say don't put batch loop inside the condition, first batch, then branch per-item inside. Prompts batch-processing section showed a happy-path loop, so wanted to add guidance on where loops could live. So now i reworked it as a single positive sentence.

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
luka-zivkovic: Other conditions in the workflow-patterns.ts uses this format so that's why i used it as well

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
luka-zivkovic: Not buggy, just a type error. onEachBatch takes a node or chain and IfElseBuilder is a separate type that isn't assignable to either. I had this bug from the trace
SplitInBatchesBuilder is not assignable to type 'IfElseTarget'
and it kept appearing every time the builder tried to batch inside a conditional
Basically, this should just steer the LLM away from the case where it tries this:
ifElse.onTrue(splitInBatches(...)) (so basically passing a whole batch loop as a branch target. The example shows the correct shape instead where we batch first and then branch per item inside onEachBatch.

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
luka-zivkovic: dropped "via a NodeChain"

packages/@n8n/workflow-sdk/src/prompts/sdk-reference/workflow-patterns.ts
luka-zivkovic: dropped "Switch rules use rules.values (NOT rules.rules)"

PR #28936, opened 2026-04-23
feat(core): Add cluster check reconciliation cycle (no-changelog)

afitzek:
## Summary

Wires up the Cluster Instance Registry's `CheckService`: a leader-only reconciliation loop that discovers and runs cluster-wide health checks, plus an on-demand path that lets any instance compute the same results for the REST/UI surface without side effects.

**Why.** Part of the Cluster Instance Registry RFC (IAM-334, Phase 3: Health Check Registry & Leader Service). Follow-up to the existing registry plumbing (instance registration, storage backends, decorator/metadata in `@n8n/decorators`) and the first built-in check (`VersionMismatchCheck`). The RFC commits to computing user-facing check results on demand; this PR implements that contract and exposes the results to the UI.

**Key design decisions.**

- **Two entry points, one code path.** `CheckService.runChecks()` is the shared primitive: it fetches current state, reads the last-known snapshot, computes the diff, invokes every `@ClusterCheck()` class, and returns aggregated results. The leader's `reconcile()` loop calls it every 180s and layers side-effect dispatch (warnings, audit events, push broadcasts) + `saveLastKnownState` on top. The REST controller calls it directly on every request.
- **Single writer for `lastKnownState`.** Only the leader's scheduled loop writes it. `runChecks()` reads it to build diff context but never mutates it — otherwise UI polling would race the leader and corrupt dedup baselines across handoffs.
- **Recompute on demand for the UI.** Checks are cheap (one Redis `SMEMBERS` + `MGET`, sub-ms CPU). Persisting results would just duplicate the registry data and introduce TTL/staleness semantics for no real gain. Any main can answer, no leader routing required.
- **Response shape shows successful checks too.** `GET /instance-registry/` now returns a `checks` summary keyed by check name — `{ status: 'succeeded' | 'failed', warnings, executedAt }` — so the UI can render the full picture (green ticks for healthy checks, warnings for failing ones) rather than just a warnings-only feed.
- **Thrown checks surface as a synthetic warning.** When a check's `run()` throws, the controller emits a `cluster.check-execution-failed` warning so the UI never silently drops a failing check.

**Scope.** The module is gated behind `N8N_ENV_FEAT_INSTANCE_REGISTRY=true`; no user-visible changes when the flag is off. Ships one built-in check (`VersionMismatchCheck`); additional checks (hostId clash, split-brain, lifecycle) land in follow-up IAM-333 tickets.

## How to test manually

Start with `export N8N_ENV_FEAT_INSTANCE_REGISTRY=true` visit http://localhost:5678/settings/instance-registry to see the computed check.

closes  https://linear.app/n8n/issue/IAM-334

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/cli/src/modules/instance-registry/checks/check.service.ts
+++ b/packages/cli/src/modules/instance-registry/checks/check.service.ts
@@ -1,0 +1,296 @@
+import type { InstanceRegistration, PushMessage } from '@n8n/api-types';
+import { Logger } from '@n8n/backend-common';
+import type {
+	ClusterCheckAuditEvent,
+	ClusterCheckContext,
+	ClusterCheckPushNotification,
+	ClusterCheckResult,
+	ClusterCheckWarning,
+	ClusterStateDiff,
+	IClusterCheck,
+} from '@n8n/decorators';
+import {
+	ClusterCheckMetadata,
+	OnLeaderStepdown,
+	OnLeaderTakeover,
+	OnShutdown,
+} from '@n8n/decorators';
+import { Container, Service } from '@n8n/di';
+import { InstanceSettings } from 'n8n-core';
+
+import type { EventNamesAuditType } from '@/eventbus/event-message-classes';
+import type { EventPayloadAudit } from '@/eventbus/event-message-classes/event-message-audit';
+import { MessageEventBus } from '@/eventbus/message-event-bus/message-event-bus';
+import { Push } from '@/push';
+
+import { InstanceRegistryService } from '../instance-registry.service';
+import { REGISTRY_CONSTANTS } from '../instance-registry.types';
+
+/**
+ * Leader-only service that reconciles cluster state and runs health checks
+ * on a fixed interval. Discovers checks via `ClusterCheckMetadata`, fans them
+ * out with error isolation, and forwards each result into logs, the audit
+ * event bus, and the push channel.
+ */
+@Service()
+export class CheckService {
+	private reconcileInterval: NodeJS.Timeout | undefined;
+
+	private isShuttingDown = false;
+
+	private readonly checks: IClusterCheck[] = [];
+
+	private readonly logger: Logger;
+
+	constructor(
+		logger: Logger,
+		private readonly instanceSettings: InstanceSettings,
+		private readonly instanceRegistryService: InstanceRegistryService,
+		private readonly clusterCheckMetadata: ClusterCheckMetadata,
+		private readonly messageEventBus: MessageEventBus,
+		private readonly push: Push,
+	) {
+		this.logger = logger.scoped('instance-registry');
+	}
+
+	init() {
+		this.discoverChecks();
+		if (this.instanceSettings.isLeader) this.startReconciliation();
+	}
+
+	@OnLeaderTakeover()
+	startReconciliation() {
+		if (this.isShuttingDown) return;
+		if (this.reconcileInterval) return;
+
+		void this.runReconcileSafely();
+
+		this.reconcileInterval = setInterval(
+			async () => await this.runReconcileSafely(),
+			REGISTRY_CONSTANTS.RECONCILIATION_INTERVAL_MS,
+		);
+
+		this.logger.debug('Cluster check reconciliation scheduled');
+	}
+
+	@OnLeaderStepdown()
+	stopReconciliation() {
+		clearInterval(this.reconcileInterval);
+		this.reconcileInterval = undefined;
+	}
+
+	@OnShutdown()
+	shutdown() {
+		this.isShuttingDown = true;
+		this.stopReconciliation();
+	}
+
+	private discoverChecks() {
+		const checkClasses = this.clusterCheckMetadata.getClasses();
+
+		for (const CheckClass of checkClasses) {
+			try {
+				const check = Container.get(CheckClass);
+				this.checks.push(check);
+			} catch (error) {
+				this.logger.error(`Failed to instantiate cluster check "${CheckClass.name}"`, {
+					error,
+				});
+			}
+		}
+
+		this.logger.info(`Discovered ${this.checks.length} cluster checks`, {
+			names: this.checks.map((c) => c.checkDescription.name),
+		});
+	}
+
+	private async runReconcileSafely() {
+		try {
+			await this.reconcile();
+		} catch (error) {
+			this.logger.warn('Reconciliation cycle failed', { error });
+		}
+	}
+
+	/**
+	 * Evaluates all registered cluster checks against the current cluster state
+	 * and returns their aggregated results. Side-effect free: does not dispatch
+	 * warnings/audit events/push notifications and does not persist state.
+	 *
+	 * Safe to call from any instance (leader or follower), e.g. from the REST
+	 * controller serving the cluster overview UI. The leader's scheduled
+	 * reconciliation loop is the single writer of `lastKnownState`; this method
+	 * only reads it to build the diff context for checks.
+	 */
+	async runChecks(): Promise<{
+		currentState: Map<string, InstanceRegistration>;
+		results: Array<{
+			checkName: string;
+			checkDisplayName?: string;
+			result?: ClusterCheckResult;
+			failed?: true;
+		}>;
+	}> {
+		if (this.checks.length === 0) {
+			return { currentState: new Map(), results: [] };
+		}
+
+		const instances = await this.instanceRegistryService.getAllInstances();
+		const currentState = new Map<string, InstanceRegistration>(
+			instances.map((i) => [i.instanceKey, i]),
+		);
+
+		const previousState = await this.instanceRegistryService.getLastKnownState();
+		const diff = computeDiff(previousState, currentState);
+
+		const context: ClusterCheckContext = { currentState, previousState, diff };
+
+		const settled = await Promise.allSettled(
+			this.checks.map(async (check) => await check.run(context)),
+		);
+
+		const results: Array<{
+			checkName: string;
+			checkDisplayName?: string;
+			result?: ClusterCheckResult;
+			failed?: true;
+		}> = [];
+		for (let i = 0; i < settled.length; i++) {
+			const outcome = settled[i];
+			const check = this.checks[i];
+			if (outcome.status === 'fulfilled') {
+				results.push({
+					checkName: check.checkDescription.name,
+					checkDisplayName: check.checkDescription.displayName,
+					result: outcome.value,
+				});
+			} else {
+				this.logger.error('Cluster check failed', {
+					checkName: check.checkDescription.name,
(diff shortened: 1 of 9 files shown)


packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	@OnLeaderTakeover()
+	startReconciliation() {
+		if (this.isShuttingDown) return;
+		if (this.reconcileInterval) return;
+
+		void this.runReconcileSafely();
+
+		this.reconcileInterval = setInterval(
```
guillaumejacquart: i've seen setTimeout used elsewhere for background processes, to make sure that no new process starts before the last one finishes. I don't think that applies here, considering the timeout and nature of the checks, but still worth mentioning it

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,296 @@
...
+			async () => await this.runReconcileSafely(),
+			REGISTRY_CONSTANTS.RECONCILIATION_INTERVAL_MS,
+		);
+
+		this.logger.debug('Cluster check reconciliation scheduled');
+	}
+
+	@OnLeaderStepdown()
+	stopReconciliation() {
+		clearInterval(this.reconcileInterval);
```
guillaumejacquart: Do we want to stop a running reconciliation process here ?

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,296 @@
...
+		this.reconcileInterval = undefined;
+	}
+
+	@OnShutdown()
+	shutdown() {
+		this.isShuttingDown = true;
```
guillaumejacquart, line 84: Do we want to stop a reconciliation process here, or wait for it to finish ?

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,296 @@
...
+					checkDisplayName: check.checkDescription.displayName,
+					result: outcome.value,
+				});
+			} else {
+				this.logger.error('Cluster check failed', {
+					checkName: check.checkDescription.name,
+					checkDisplayName: check.checkDescription.displayName,
+					error: outcome.reason,
+				});
+				results.push({
```
guillaumejacquart: Very nit, but i don't like having almost the same push in the if and else. What about preparing the data conditionally, and having a single push after ?

packages/cli/src/modules/instance-registry/instance-registry.controller.ts
```diff
@@ -1,20 +1,63 @@
...
+					code: warning.code,
+					message: warning.message,
+					severity: warning.severity,
+					context: warning.context,
+				}));
+			}
+
+			if (cur.failed === true) {
+				acc[cur.checkName].status = 'failed';
+				acc[cur.checkName].warnings = [
```
guillaumejacquart: this would override the warning in case there already was any from line 32. Not sure if it's a problem

packages/cli/src/modules/instance-registry/instance-registry.controller.ts
```diff
@@ -1,20 +1,63 @@
...
 	@Get('/')
 	@GlobalScope('orchestration:read')
 	async getClusterInfo(): Promise<ClusterInfoResponse> {
-		const instances = await this.instanceRegistryService.getAllInstances();
+		const [instances, { results }] = await Promise.all([
+			this.instanceRegistryService.getAllInstances(),
+			this.checkService.runChecks(),
+		]);
+
+		const checks = results.reduce((acc, cur) => {
```
guillaumejacquart, line 24: some refactor suggestion:
```ts
const now = Date.now();

const checks = results.reduce((acc, cur) => {
    const { checkName, result, failed } = cur;

    // Initialize if it doesn't exist
    if (!acc[checkName]) {
        acc[checkName] = {
            status: 'succeeded',
            check: checkName,
            executedAt: now,
            warnings: [],
        };
    }

    const entry = acc[checkName];

    // Handle Warnings from results
    const newWarnings = result?.warnings?.map(w => ({
        check: checkName,
        ...w // Assuming the warning object structure matches
    })) ?? [];

    if (newWarnings.length > 0) {
        entry.status = 'failed';
        entry.warnings.push(...newWarnings);
    }

    // Handle Execution Failure
    if (failed) {
        entry.status = 'failed';
        entry.warnings.push({
            check: checkName,
            code: 'cluster.check-execution-failed',
            message: 'Failed to execute cluster check, please check error logs for details',
            severity: 'warning',
        });
    }

    return acc;
}, {} as ClusterCheckSummary);
```

guillaumejacquart (commented):
some minor comments

packages/cli/src/modules/instance-registry/instance-registry.controller.ts
afitzek: Nice! 👍

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,310 @@
...
+		private readonly clusterCheckMetadata: ClusterCheckMetadata,
+		private readonly messageEventBus: MessageEventBus,
+		private readonly push: Push,
+	) {
+		this.logger = logger.scoped('instance-registry');
+	}
+
+	init() {
+		this.discoverChecks();
+		if (this.instanceSettings.isLeader) this.startReconciliation();
```
BGZStephen, line 59: Is this required? Or does the decorator `@OnLeaderTakeover()` catch leader starts where the starting instance *is* the leader? Just checking for a double fire

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,310 @@
...
+
+	@OnShutdown()
+	shutdown() {
+		this.isShuttingDown = true;
+		this.stopReconciliation();
+	}
+
+	private scheduleNextReconcile(signal: AbortSignal) {
+		if (signal.aborted) return;
+		this.reconcileTimer = setTimeout(async () => {
```
BGZStephen, line 90: Do we need to also check for an existing timer instead of overwriting it?

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,310 @@
...
+			instances.map((i) => [i.instanceKey, i]),
+		);
+
+		const previousState = await this.instanceRegistryService.getLastKnownState();
+		const diff = computeDiff(previousState, currentState);
+
+		const context: ClusterCheckContext = { currentState, previousState, diff };
+
+		const settled = await Promise.allSettled(
+			this.checks.map(async (check) => await check.run(context)),
```
BGZStephen, line 157: Is this ok to be unbounded?

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	private async reconcile(signal: AbortSignal) {
+		if (this.checks.length === 0) return;
+
+		if (signal.aborted) return;
+
+		const { currentState, results } = await this.runChecks();
+
+		if (signal.aborted) return;
```
BGZStephen, line 200: Why are there two checks for `if (signal.aborted) return;`?

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,310 @@
...
+
+		for (const { checkName, result } of results) {
+			this.processResult(checkName, result);
+		}
+
+		try {
+			if (signal.aborted) return;
```
BGZStephen, line 207: A 3rd `if (signal.aborted) return;`? Can't decide if this code is trolling me...

packages/cli/src/modules/instance-registry/checks/check.service.ts
```diff
@@ -0,0 +1,310 @@
...
+				});
+			});
+	}
+
+	private broadcastPush(checkName: string, notification: ClusterCheckPushNotification) {
+		try {
+			this.push.broadcast({
+				type: notification.type,
+				data: notification.data,
+			} as PushMessage);
```
BGZStephen, line 259: Surprised you need a cast here

packages/cli/src/modules/instance-registry/checks/check.service.ts
afitzek: These checks are fine to be unbounded, because these are registered checks in the code. So each check is a specific class that registers itself using a decorator.

packages/cli/src/modules/instance-registry/checks/check.service.ts
afitzek: In each tick of the event loop the abort signal can be send. So by checking in multiple points in the code if we should abort the running execution.

packages/cli/src/modules/instance-registry/checks/check.service.ts
afitzek: In each tick of the event loop the abort signal can be send. So by checking in multiple points in the code if we should abort the running execution.

packages/cli/src/modules/instance-registry/checks/check.service.ts
BGZStephen: right, so something outside of the function is mutating the signal?

packages/cli/src/modules/instance-registry/checks/check.service.ts
afitzek: This is required, because the `OnLeaderTakeover` does not handle the leader during starts. For in single-main mode, we would miss the `OnLeaderTakeover` event. Also `startReconciliation` is a noop, because it checks on the existence of a reconcileController and exists early if it is already started.

BGZStephen: Perfect, thanks!

packages/cli/src/modules/instance-registry/checks/check.service.ts
afitzek: I think that should not really be a problem, but I'll add a clearTimeout just to be sure. 👍

packages/cli/src/modules/instance-registry/checks/check.service.ts
afitzek: Since we are dropping the UI part of this, I'll drop the whole push notifications for now. 👍

BGZStephen (approved):

guillaumejacquart (approved):

PR #28945, opened 2026-04-23
refactor(editor): Move workflow serialization to workflowDocument store

alexgrozav:
## Summary

Relocates `getWorkflowDataToSave` and `getNodeDataToSave` out of `useWorkflowHelpers` into the `workflowDocument` store as `serialize` / `serializeNode`, with the node-serialization logic extracted to a pure utility (`workflowDocument/serializeNode.ts`) so callers like `useCanvasOperations` can bypass Pinia's action wrapping when needed. 14 call sites across `useWorkflowSaving`, `useRunWorkflow`, `useWorkflowCommands`, `executionFinished`, `workflows.store`, `ActionsDropdownMenu`, `useCanvasOperations`, and the internal callers in `useWorkflowHelpers` now use the store methods instead. Behavior is unchanged — the functions still read the same document state; the methods became synchronous since they never awaited anything, and all callers were updated to drop the extraneous `await`.

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/app/stores/workflowDocument/useWorkflowDocumentNodeSerialization.ts
autologie, line 1: The function `serializeNode` isn't Vue composable and works independent of workflowDocument. Could `app/utils/nodes/nodeTransforms.ts` be a better place to define this?

packages/frontend/editor-ui/src/app/utils/nodeTypeUtils.ts
autologie, line 1: There's already `packages/frontend/editor-ui/src/app/utils/nodeTypesUtils.ts`, is it intentional to create a new file?

autologie (commented):
A few nitpicks regarding file names

packages/frontend/editor-ui/src/app/stores/workflowDocument/useWorkflowDocumentNodeSerialization.ts
alexgrozav: Fair point. Moving to `nodeTransforms.ts` as it might be reused again in the future.

packages/frontend/editor-ui/src/app/utils/nodeTypeUtils.ts
alexgrozav: Good catch. Just a bad refactoring prompt 😣

autologie (dismissed):
👍

autologie (approved):

PR #28961, opened 2026-04-23
test: Retry SSE webhook setup on 404

ivov:
`Test: E2E VM Expressions Nightly` failed 13 of 15 recent runs with isolated expressions, consistently on `mcp-trigger.spec.ts` specifically the SSE transport tests.

This is because there's a race between workflow activation and the first request that hits the webhook. In multi-main mode, a request to `/rest/workflows/:id/activate` causes themain that receives it to publish a message to the leader over Redis. The leader then evaluates the webhook's path expressions and insert the webhook in the DB.

In the legacy expression engine, expression evaluation is sub-ms, so the leader almost always wins the race. But in the `vm` engine, that same evaluation means acquiring a V8 isolate bridge and having each `$parameter["path"]` etc. lookup cross the isolate boundary via a callback back to Node. This is much slower, consistently enough for the test client to beat the leader, so the client gets a 404. 

The fix mirrors the same retry loop into `sseSetup`. But importantly, the activation race remains regardless of expression engine. The `vm` engine just widens the window enough to expose it. A deeper fix is architectural and should be tackled separately.

### DIFF
--- a/packages/testing/playwright/services/mcp-api-helper.ts
+++ b/packages/testing/playwright/services/mcp-api-helper.ts
@@ -177,6 +177,29 @@ export class McpApiHelper {
 	 */
 	async sseSetup(
 		path: string,
+		options?: {
+			headers?: Record<string, string>;
+			maxNotFoundRetries?: number;
+			notFoundRetryDelayMs?: number;
+		},
+	): Promise<McpSession> {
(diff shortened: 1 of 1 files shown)


despairblue (approved):
got carried away fixing nits, sorray 😕 

thanks for fixing this 🙏🏾

PR #28964, opened 2026-04-23
fix(WooCommerce Trigger Node): Guard HMAC verification against missing secret

BerniWittmann:
## Summary

The `webhook()` method crashed with an unhandled `TypeError` when `workflowStaticData.secret` was `undefined`, because `createHmac('sha256', undefined)` throws synchronously. This caused HTTP 500 responses to WooCommerce without any failed execution visible in the UI.

This change:
- Guards the secret lookup and throws a `NodeOperationError` with an actionable message (deactivate + re-activate) when the stored secret is missing — surfaces as a failed execution the user can see.
- Changes signature mismatches and missing signature headers from a silent 200 drop to a proper `401 Unauthorized` response (without logging a failed execution, since the caller isn't trusted).
- Adds unit tests covering all 5 branches of `webhook()`.

To test: activate a WooCommerce Trigger workflow, hit its webhook URL with a missing / invalid / valid `x-wc-webhook-signature` header and verify the responses (no stack traces, 401 on bad auth, 200 on valid). Unit tests in `WooCommerceTrigger.node.test.ts` are the authoritative coverage.

https://linear.app/n8n/issue/NODE-4821
closes #28041

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
+++ b/packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
@@ -9,6 +9,7 @@ import type {
 } from 'n8n-workflow';
 import { NodeConnectionTypes, NodeOperationError } from 'n8n-workflow';
 
+import { verifySignature } from '../../utils/webhook-signature-verification';
 import { getAutomaticSecret, woocommerceApiRequest } from './GenericFunctions';
 
 export class WooCommerceTrigger implements INodeType {
@@ -133,7 +134,7 @@ export class WooCommerceTrigger implements INodeType {
 									{ force: true },
 								);
 							} catch (error) {
-								this.logger.debug('Failed to delete orphaned webhook during checkExists', {
+								this.logger.warn('Failed to delete orphaned webhook during checkExists', {
 									webhookId: webhook.id,
 									error,
 								});
@@ -197,12 +198,12 @@ export class WooCommerceTrigger implements INodeType {
 			});
 		}
 
-		const providedSignature = headerData['x-wc-webhook-signature'] as string | undefined;
-		const computedSignature = providedSignature
-			? createHmac('sha256', secret).update(req.rawBody).digest('base64')
-			: undefined;
+		const isValid = verifySignature({
+			getExpectedSignature: () => createHmac('sha256', secret).update(req.rawBody).digest('base64'),
+			getActualSignature: () => (headerData['x-wc-webhook-signature'] as string) ?? null,
+		});
 
-		if (!providedSignature || providedSignature !== computedSignature) {
+		if (!isValid) {
 			const res = this.getResponseObject();
 			res.status(401).send('Unauthorized').end();
 			return { noWebhookResponse: true };
--- a/packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
+++ b/packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
@@ -7,7 +7,7 @@ import type {
 	INodeTypeDescription,
 	IWebhookResponseData,
 } from 'n8n-workflow';
-import { NodeConnectionTypes } from 'n8n-workflow';
+import { NodeConnectionTypes, NodeOperationError } from 'n8n-workflow';
 
 import { getAutomaticSecret, woocommerceApiRequest } from './GenericFunctions';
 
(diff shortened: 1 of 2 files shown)


packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
```diff
@@ -164,17 +164,30 @@ export class WooCommerceTrigger implements INodeType {
...
 			return {};
 		}
 
-		const computedSignature = createHmac('sha256', webhookData.secret as string)
-			.update(req.rawBody)
-			.digest('base64');
-		if (headerData['x-wc-webhook-signature'] !== computedSignature) {
-			// Signature is not valid so ignore call
-			return {};
+		const secret = webhookData.secret as string | undefined;
```
yehorkardash, line 193: Usually we skip verification if secret is not set for backwards compatibility reasons, but I see that we added secret verification from the very start. Do we know the reason a secret disappeared from storage?

BerniWittmann: I think it can disappear via two ways: 

1. Workflow exporting & importing or duplicating (staticData is not copied over)
2. Setting up the webhook on woocommerce first, then activating the workflow (it will then only store the webhook id, because the webhook already exists, but not the secret)

yehorkardash: Can we update checkExists method to update a webhook with secret when it's missing?

BerniWittmann: Yeah we could do that, aka self heal the existing webhook. 

I would maybe propose to instead in checkExists when no secret is found for an existing webhook to simply delete the webhook (similar to `CurrentsTrigger.node.ts`), which should then automatically call create and set up a fresh webhook

Regardless of the solution i would still keep the improved error handling. 

What do you think about that?

yehorkardash: I think we do self healing in checkExists rather than drop webhook completely, but I'm not 100% sure, you can check how we do it in other nodes. 

As for error handling, I agree

BerniWittmann: It's still self healing. But instead of having an update call in check exists, we notice in check exists the secret is missing and remove the webhook. And then create function automatically kicks in and creates it fresh. 
CurrentTrigger node does it that way as well

packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
```diff
@@ -164,17 +184,30 @@ export class WooCommerceTrigger implements INodeType {
...
+					'The stored webhook secret could not be found. Deactivate and re-activate the workflow so n8n can re-register the webhook with WooCommerce.',
+			});
 		}
+
+		const providedSignature = headerData['x-wc-webhook-signature'] as string | undefined;
+		const computedSignature = providedSignature
+			? createHmac('sha256', secret).update(req.rawBody).digest('base64')
+			: undefined;
+
+		if (!providedSignature || providedSignature !== computedSignature) {
```
yehorkardash: `!==` on the base64 strings is timing-attack-vulnerable.
You can use packages\nodes-base\utils\webhook-signature-verification.ts to do the comparison

packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
```diff
@@ -122,6 +122,26 @@ export class WooCommerceTrigger implements INodeType {
...
+							try {
+								await woocommerceApiRequest.call(
+									this,
+									'DELETE',
+									`/webhooks/${webhook.id}`,
+									{},
+									{ force: true },
+								);
+							} catch (error) {
+								this.logger.debug('Failed to delete orphaned webhook during checkExists', {
```
yehorkardash: Would probably make it `warn()`

packages/nodes-base/nodes/WooCommerce/WooCommerceTrigger.node.ts
BerniWittmann: very good catch 🙌

yehorkardash (approved):

PR #28973, opened 2026-04-23
fix(core): Persist execution context before writing to db

phyllis-noester:
## Summary

**Change:** Before `WorkflowRunner.run()` calls `activeExecutions.add()` (which triggers the DB write via
  `executionPersistence.create()`), it now builds a `Workflow` + `additionalData` and calls `establishExecutionContext()`. Guarded by
  `if (data.executionData)` so manual executions without pre-populated data skip the call.

  **Files:**
  - `packages/cli/src/workflow-runner.ts` — the fix, plus new `n8n-core` import
  - `packages/cli/src/__tests__/workflow-runner.test.ts` — 3 tests (ordering, masked-data-persisted, undefined-guard)
  - `packages/core/src/execution-engine/index.ts` — exports `establishExecutionContext`
  - `packages/core/src/execution-engine/__tests__/execution-context.test.ts` — worker idempotency test

  # Why

  **Problem:** In queue mode, raw `Authorization: Bearer <JWT>` headers from webhook triggers were written to PostgreSQL before
  anything masked them. The worker masked in memory later, but the DB held the raw token until (and unless) the worker wrote an updated
   record. A worker crash left the raw JWT as the final row.

  **Why in `WorkflowRunner.run()`:** Every execution path funnels through here (webhooks, chat, wait-tracker resumes, manual, CLI, MCP,
   AI-builder). Placing the call at this choke point gives uniform coverage and co-locates context establishment with the
  enqueue-vs-local decision and the DB write.

  **Why it's safe to run twice:** `establishExecutionContext` early-exits if `runtimeData` is already set. The worker still calls it
  inside `WorkflowExecute.processRunExecutionData`, but finds `runtimeData` populated and returns immediately — no double-processing.

  **Why the minimal build (not the thread-through refactor):** `runMainProcess()` does more than `establishExecutionContext` needs
  (staticData loading, pinData, timeouts). Lifting that would change `loadStaticData` semantics in queue mode. Building one extra
  `Workflow` + `additionalData` per execution is cheap and keeps the diff tightly scoped to the security fix.

## Testing

  These steps verify that webhook headers are redacted in the persisted execution
  data **while the execution is still running** (not just after completion).

  ### Prerequisites

  - Docker running
  - A workflow JSON that starts with a Webhook trigger and has a long-running step
  afterwards (e.g. a `Wait` node set to several minutes). An example workflow file
  is attached to this PR.

  ### 1. Start n8n in queue mode

  From `packages/testing/containers/`:

  ```bash
  args=(--queue)
  while IFS= read -r line; do
    args+=(--env "$line")
  done < <(grep -v '^\s*#\|^\s*$' ../../../.env)
  npx tsx ./n8n-start-stack.ts "${args[@]}"
  ```

  Note the project name printed in the logs (e.g. `n8n-stack-htdxh`) — you'll need
  it to connect to Postgres.

  ### 2. Import and activate the test workflow

  1. Open the n8n URL printed by the stack starter.
  2. Import the attached workflow JSON.
  3. Activate the workflow so the webhook is live.

  ### 3. Trigger the webhook

  Call the webhook URL with some headers that should be redacted (e.g.
  `Authorization`):

  ```bash
  curl -X POST <webhook-url> \
    -H "Authorization: Bearer secret-token-123" \
    -H "Cookie: session=abc123" \
    -H "X-Custom: visible-value"
  ```

  The workflow will start and pause on the long-running node, keeping the execution
  in a running state.

  ### 4. Inspect the persisted execution data

  While the execution is still running, open a psql shell against the stack's
  Postgres container:

  ```bash
  docker exec -it <project-name>-postgres psql -U n8n_user -d n8n_db
  ```

  Find the running execution's ID:

  ```sql
  SELECT id, status FROM execution_entity WHERE status = 'running' ORDER BY id DESC
  LIMIT 5;
  ```

  Check the persisted payload for that execution:

  ```sql
  SELECT data FROM execution_data WHERE "executionId" = '<id>';
  ```

  ### 5. Verify

  - Sensitive headers (e.g. `Authorization`, `Cookie`) appear **redacted** in the
  `data` column.
  - Non-sensitive headers (e.g. `X-Custom`) appear with their original values.
  - The row exists **before the execution finishes** — confirming the context is
  persisted pre-completion.

[My workflow.json](https://github.com/user-attachments/files/27061086/My.workflow.json)

https://linear.app/n8n/issue/IAM-547/fixcli-redact-jwt-tokens-before-execution-db-write-in-queue-mode

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/workflow-runner.ts
+++ b/packages/cli/src/workflow-runner.ts
@@ -18,6 +18,7 @@ import type {
 	ExecutionError,
 	IDeferredPromise,
 	IExecuteResponsePromiseData,
+	INode,
 	IPinData,
 	IRun,
 	WorkflowExecuteMode,
@@ -156,6 +157,28 @@ export class WorkflowRunner {
 		await hooks?.runHook('workflowExecuteAfter', [fullRunData]);
 	}
 
+	/**
+	 * Persist a failed execution record for an already-registered execution and
+	 * finalize it, so a pre-flight failure surfaces as a normal failed run.
+	 */
+	private async failExecution(
+		data: IWorkflowExecutionDataProcess,
+		executionId: string,
+		error: ExecutionError & { node?: INode },
+		responsePromise?: IDeferredPromise<IExecuteResponsePromiseData>,
+	): Promise<void> {
+		const runData = this.failedRunFactory.generateFailedExecutionFromError(
+			data.executionMode,
+			error,
+			error.node,
+		);
+		const lifecycleHooks = getLifecycleHooksForRegularMain(data, executionId);
+		await lifecycleHooks.runHook('workflowExecuteBefore', [undefined, data.executionData]);
+		await lifecycleHooks.runHook('workflowExecuteAfter', [runData]);
+		responsePromise?.reject(error);
+		this.activeExecutions.finalizeExecution(executionId);
+	}
+
 	/** Run the workflow
 	 * @param realtime This is used in queue mode to change the priority of an execution, making sure they are picked up quicker.
 	 */
@@ -166,6 +189,9 @@ export class WorkflowRunner {
 		restartExecutionId?: string,
 		responsePromise?: IDeferredPromise<IExecuteResponsePromiseData>,
 	): Promise<string> {
+		// Register a new execution
+		const executionId = await this.activeExecutions.add(data, restartExecutionId);
+
 		// Establish the execution context before persisting to the DB.
 		// activeExecutions.add() -> executionPersistence.create() writes
 		// data.executionData to the DB; any header masking or runtimeData
@@ -177,6 +203,7 @@ export class WorkflowRunner {
 		// the outer IRunExecutionData is created with `executionData: null`
 		// so the trigger-item stack is undefined here; nothing to mask yet,
 		// the worker will establish context once it populates the stack.
+
 		if (data.executionData?.executionData) {
 			// Deliberately lightweight: no pinData, no staticData loading,
 			// no additionalData. establishExecutionContext only needs the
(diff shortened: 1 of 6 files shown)


packages/cli/src/workflow-runner.ts
```diff
@@ -160,6 +166,41 @@ export class WorkflowRunner {
...
+				id: data.workflowData.id,
+				name: data.workflowData.name,
+				nodes: data.workflowData.nodes,
+				connections: data.workflowData.connections,
+				active: data.workflowData.activeVersionId !== null,
+				nodeTypes: this.nodeTypes,
+				staticData: data.workflowData.staticData,
+				settings: data.workflowData.settings ?? {},
+			});
+			await establishExecutionContext(
```
afitzek: I'm wondering if we need to handle errors that are thrown here. Should these be counted against workflow executions? I think if we throw an exception here, we are not seeing a failed workflow execution in the history, but just an error through what ever trigger we received this from.

afitzek (commented):
I think we need to consider the error handling path for this. Maybe we can wrap this into a try catch similar to the credentialsPermissionChecker.check.

packages/cli/src/workflow-runner.ts
phyllis-noester: makes sense! I was a bit torn how to handle the possibly un-redacted data in case establishing the execution context fails; for that reason I wipe the data now before creating a failed execution but that makes it harder to debug for customers; let me know what you think :)

afitzek (approved):
Nice work!! 🚀

PR #28978, opened 2026-04-23
fix(ai-builder): Stop treating empty defaults as satisfying required for the Split node

aalises:
## Summary

**Evals now pass cleanly, no more builder failures on Split Out.**

- [Before vs after comparison](https://smith.langchain.com/o/f6d01556-807a-45e9-8732-d76e0788584b/datasets/131d179d-5e53-447e-b03e-070107e86069/compare?selectedSessions=de2fa2ad-d9d5-42a8-a513-01752293e7c5%2Cda94cf45-3191-45cb-92c1-b62d3437ff3c&source=de2fa2ad-d9d5-42a8-a513-01752293e7c5&example=5f283e8c-2d90-48f4-86be-2ecfcc0deb45) (LangSmith)
- [Passing run](https://smith.langchain.com/o/f6d01556-807a-45e9-8732-d76e0788584b/datasets/131d179d-5e53-447e-b03e-070107e86069/compare?selectedSessions=da94cf45-3191-45cb-92c1-b62d3437ff3c) (LangSmith)

### What was broken

Running instance-ai evals, Split Out kept failing three different ways:
- **Empty required field**: `fieldToSplitOut` was left blank and the workflow blew up at execution.
- **Junk values**: the LLM would fill the field with `$json` or `$binary` which neither is valid.
- **Buffer crash**: the HTTP Request mock returned `{"type":"Buffer","data":[...]}` and the Notion node crashed trying to read `$json.title`.

### What the PR fixes

1. **Codegen truth-telling.** The SDK said `fieldToSplitOut?: string` (optional) because the field had a `default: ''`, even though it was marked `required: true`. The LLM believed the schema, left it empty, and everything downstream broke. Fixed so empty-string / empty-array defaults no longer count as satisfying `required`. Both the Zod schema *and* the TypeScript type now mark it required.
2. **Guidance for the LLM.** Split Out had no `builderHint` on `fieldToSplitOut`, so once the field was required the LLM guessed `$json` or `$binary`. Added a concrete hint: *"use the key name inside `$json` that points at an array — e.g. `issues` for `{ issues: [...] }`. Don't pass `$json` or prefix with `$json.`. `$binary` is only for binary data."*
3. **Eval mock shape.** The mock HTTP response always came back as a Buffer, which didn't match what axios returns on its default JSON path. HTTP Request nodes crashed. Aligned the mock with axios's `responseType`: stream → `Readable`, arraybuffer → `Buffer`, text → string, default → parsed object (with a `__bodyResolved` flag so the node skips its decode pipeline).
4. **New eval scenario.** Added an Airtable → Slack test case with a named-array envelope (`{ records: [...] }`) so there's a clean positive signal for Split Out behavior alongside the trickier GitHub bare-array scenario.

### How to test

- Unit tests: `pnpm test src/generate-types/` in `@n8n/workflow-sdk` (324 tests), `pnpm test src/execution-engine/__tests__/eval-mock-helpers.test.ts` in `core` (44 tests), `pnpm test SplitOut` in `nodes-base` (9 tests).
- Evals: `pnpm eval:instance-ai --filter airtable-split-to-slack --iterations 3` — the LLM should set `fieldToSplitOut: 'records'` and all scenarios should pass (confirmed, see links above).

https://linear.app/n8n/issue/INS-121

## Review / Merge checklist


JoseBra (dismissed):

Cadiac (dismissed):
This fix seems reasonable, but I'm actually a bit surprised we treat empty strings as defaults this way 🤔

Cadiac (approved):

JoseBra (approved):

PR #28984, opened 2026-04-23
feat(core): Add delete folder endpoint to public API

alielkhateeb:
## Summary

https://www.loom.com/share/138262ab555643db848e78034ce99e79

Adds a \`DELETE /projects/{projectId}/folders/{folderId}\` endpoint to the public API, which was missing from the initial folder management implementation.

The endpoint:
- Requires the \`folder:delete\` API key scope
- Accepts an optional \`transferToFolderId\` query parameter to move folder contents before deletion
- Returns \`204 No Content\` on success
- Returns \`404\` if the folder or project is not found
- Returns \`400\` for invalid input (e.g. attempting to delete a non-empty folder without a transfer target)

**Changes:**
- Adds \`deleteFolder\` handler with license and scope checks
- Adds OpenAPI spec for \`DELETE /projects/{projectId}/folders/{folderId}\`
- Adds \`folder:delete\` scope to \`API_KEY_RESOURCES\` and owner/member/admin API key scope lists
- Extracts \`PublicAPIHandler\` type into a shared file
- Replaces unit tests with comprehensive integration tests covering all folder endpoints

**Testing:**
Run the integration tests:
```bash
cd packages/cli && pnpm test test/integration/public-api/folders.test.ts
```

Or test manually with curl:

Delete a folder and transfer its contents to another folder:
```bash
curl -X DELETE \
  "http://localhost:5678/api/v1/projects/{projectId}/folders/{folderId}?transferToFolderId={targetFolderId}" \
  -H "X-N8N-API-KEY: your-api-key"
```

Delete a folder and archive its contents (no transfer):
```bash
curl -X DELETE \
  "http://localhost:5678/api/v1/projects/{projectId}/folders/{folderId}" \
  -H "X-N8N-API-KEY: your-api-key"
```

https://linear.app/n8n/issue/LIGO-467

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/public-api/v1/handlers/folders/__tests__/folders.handler.test.ts
alielkhateeb, line 1: I don't see a reason to have a separate unit tests file, integration test for folders (public API) [here](https://github.com/n8n-io/n8n/blob/5f0fa6ba04dae4988071e354d1fe23458c16dac0/packages/cli/test/integration/public-api/folders.test.ts) are more than sufficient, it is even confusing to have some paths covered here and others covered there, so I moved everything to the integration tests

packages/cli/src/public-api/v1/shared/handler.types.ts
alielkhateeb, line 1: Created a tech debt ticket to use this in other handlers https://linear.app/n8n/issue/LIGO-490/use-publicapihandler-type-in-all-public-api-handlers

packages/cli/src/public-api/v1/shared/handler.types.ts
```diff
@@ -0,0 +1,7 @@
+import type { AuthenticatedRequest } from '@n8n/db';
+import type { Response } from 'express';
+
+export type PublicAPIHandler<TParams extends Record<string, string>> = (
```
sandra0503, line 4: That's great!

sandra0503:
Looks good to me!

sandra0503 (approved):

packages/cli/src/public-api/v1/shared/handler.types.ts
```diff
@@ -0,0 +1,7 @@
...
+	req: AuthenticatedRequest<TParams>,
+	res: Response,
+) => Promise<Response | undefined>;
```
konstantintieber, line 7: What are use cases for a public api handler resolving to undefined?

konstantintieber: @alielkhateeb boop

alielkhateeb: Sorry, I need to check why I didn’t get notified about the first comment 😂

The undefined isn’t really “used” in this PR, it’s there to keep the type flexible for Express-style handlers. Some handlers call `res.send(...)` or `res.end()` without returning the Response, so the return type becomes undefined.

Example from our codebase:
https://github.com/n8n-io/n8n/blob/c0d827bf3bcceaccee26fd6c83bfaf8b379237ba/packages/cli/src/public-api/v1/handlers/workflows/workflows.handler.ts#L49

I don’t feel strongly about enforcing a strict return type → Response, `undefined` is mainly for consistency. Happy to restrict it if you prefer as part of https://linear.app/n8n/issue/LIGO-490/use-publicapihandler-type-in-all-public-api-handlers.

konstantintieber: I feel that if this interface enforces returning a response, we end up with more consistent code, and writing assertions for these handlers is also easier, since they end up becoming more pure.
Doing `res.status().send()` makes one have to write an assertion for the `send` call on the mocked response, while returning a `Response` object enables devs to write simpler assertions on the returned value of the handler.

Looking at this handler implementation for example, we see consistently the pattern of returning a response, rather than using `res.status().send`: https://github.com/n8n-io/n8n/blob/master/packages/cli/src/public-api/v1/handlers/executions/executions.handler.ts

I think this reads nicer than `res.status().send`.

alielkhateeb: As mentioned, I don't have a strong opinion, both options have benefits, I agree with what you mentioned and YAGNI applies here for if we do streaming, ever 😂 

I'll restrict handler return type to not be undefined in my next PR 👍 I'll ask you for a review 🏅

PR #28989, opened 2026-04-23
feat(core): Centralise user listing permissions for internal and publ…

sandra0503:
## Summary

Move permission checks into user service to make sure public API and internal follow the same rules.

Removes unit tests from users.controller as they were heavily mocked and not bringing much value. Tests were updated for the service.

Closes [LIGO-479](https://linear.app/n8n/issue/LIGO-479/cant-invite-users-to-a-project)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/services/__tests__/user.service.test.ts
+++ b/packages/cli/src/services/__tests__/user.service.test.ts
@@ -17,11 +17,14 @@ import { mock } from 'jest-mock-extended';
 import { v4 as uuid } from 'uuid';
 
 import { BadRequestError } from '@/errors/response-errors/bad-request.error';
+import { ForbiddenError } from '@/errors/response-errors/forbidden.error';
+import { NotFoundError } from '@/errors/response-errors/not-found.error';
 import { UrlService } from '@/services/url.service';
 import { UserService } from '@/services/user.service';
 import type { UserManagementMailer } from '@/user-management/email';
 
 import type { OwnershipService } from '../ownership.service';
+import type { ProjectService } from '../project.service.ee';
 import type { PublicApiKeyService } from '../public-api-key.service';
 import type { RoleService } from '../role.service';
 import { JwtService } from '../jwt.service';
@@ -47,6 +50,7 @@ describe('UserService', () => {
 	const roleService = mock<RoleService>();
 	const mailer = mock<UserManagementMailer>();
 	const publicApiKeyService = mock<PublicApiKeyService>();
+	const projectService = mock<ProjectService>();
 	const jwtService = mockInstance(JwtService, {
 		sign: jest.fn().mockReturnValue('mock-jwt-token'),
 	});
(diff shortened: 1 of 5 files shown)


packages/cli/src/services/user.service.ts
```diff
@@ -66,6 +70,25 @@ export class UserService {
...
+			}
+		} else if (!['global:owner', 'global:admin'].includes(user.role.slug)) {
+			const hasProjectUpdateScope =
+				(await this.projectService.getProjectIdsWithScope(user, ['project:update'])).length > 0;
+			if (!hasProjectUpdateScope) {
+				throw new ForbiddenError(
+					'Listing all users is limited to instance administrators and project admins. Filter by project to list project members.',
+				);
+			}
+		}
```
alielkhateeb, line 93: Now that we have it in a service method, I can see and understand what it does better 😂 

Here is a suggestion and why its better IMHO (Always happy to be challenged haha):
- This improves control-flow clarity
- Early returns prevents accidental checks (if someone adds logic after the if-else)
- Naming `isInstanceAdmin` documents intent better 

```suggestion
		if (projectId) {
			const project = await this.projectService.getProjectWithScope(user, projectId, ['project:list']);
			if (!project) {
				throw new NotFoundError('Project not found');
			}
			return;
		}

		const isInstanceAdmin = ['global:owner', 'global:admin'].includes(user.role.slug);
		if (isInstanceAdmin) {
			return;
		}

		const hasProjectUpdateScope =
			(await this.projectService.getProjectIdsWithScope(user, ['project:update'])).length > 0;
		if (!hasProjectUpdateScope) {
			throw new ForbiddenError(
				'Listing all users is limited to instance administrators and project admins. Filter by project to list project members.',
			);
		}
```

packages/cli/src/services/__tests__/user.service.test.ts
```diff
@@ -649,4 +654,33 @@ describe('UserService', () => {
...
+		});
+
+		it('should throw ForbiddenError for member without project admin scope', async () => {
+			const member = Object.assign(new User(), { role: GLOBAL_MEMBER_ROLE });
+			projectService.getProjectIdsWithScope.mockResolvedValueOnce([]);
+
+			await expect(userService.assertGetUsersAccess(member)).rejects.toThrow(ForbiddenError);
+		});
+
+		it('should throw NotFoundError when filtering by unknown projectId', async () => {
```
alielkhateeb, line 688: Should we add a test for the case when `projectId` is valid?

alielkhateeb (commented):
This is so much better 🥇💯🫶 Thank you for tackling this in lightning speed 🚀 
Just a couple of comments, mostly opinionated comments 😂 and we're good to go 🥳

packages/cli/src/services/user.service.ts
sandra0503: Sure, I prefer your version as well :)

alielkhateeb (approved):
🚀 Super clean 👏

geemanjs (approved):

PR #28993, opened 2026-04-23
fix(Merge Node): Improve SQL Query mode memory efficiency and error reporting

BerniWittmann:
## Summary

Community users running the Merge v3 node in **SQL Query** mode against larger item streams (25k+ rows) hit `Isolate is disposed` with no output. The existing isolate-based sandbox already provides the security boundary we want, but the data-transport and error-handling paths around it were exhausting the isolate on reasonable workloads.

This PR makes three coordinated changes to `combineBySql`:

1. **Data now crosses the isolate boundary via `context.evalClosure` arguments instead of a script-source JSON literal.** Previously the full dataset was embedded as a JSON string inside the script text, then re-parsed inside the isolate — peak allocation was roughly 3× the dataset. Now rows, the per-call database id, and the query string are passed as closure arguments with `arguments: { copy: true }` (V8 structured clone), dropping it to ~1×. Prototypes and functions are still stripped at the boundary, and because the arguments are local parameters of the implicit closure — not writes to the shared context's globals — concurrent `combineBySql` invocations on the singleton isolate cannot see or overwrite each other's inputs.
2. **Clearer error classification and sandbox recovery.** Disposed-isolate and timeout errors are detected and surfaced as actionable messages (`Dataset too large for the SQL sandbox` / `SQL query exceeded the 30 second execution limit`) with hints pointing at upstream filtering or batching. On dispose we also reset the cached isolate so the next execution starts fresh instead of reusing a poisoned singleton.
3. **Execution timeout raised from 5 s to 30 s**, matching real-world large-dataset query durations.

### How to test

1. Workflow: Code node → Merge v3 (SQL Query).
2. Code node:
   ```js
   return Array.from({ length: 25000 }, (_, i) => ({
     json: { id: i, val: i * 3, category: i % 100, name: `row-${i}`, flag: i % 2 === 0 },
   }));
   ```
3. Merge SQL: `SELECT category, SUM(val) AS total FROM input1 GROUP BY category ORDER BY total DESC`
4. Pre-fix: fails with `Isolate is disposed`. Post-fix: returns 100 rows in ~1–3 s.
5. Stress path: bump row count + column count past the 64 MB ceiling — new error reads `Dataset too large for the SQL sandbox`, and the very next small-dataset execution still succeeds (cache was reset).

Automated coverage: `pnpm test combineBtSql` in `packages/nodes-base` — 31/31 pass, including a new regression case that runs a `COUNT(*)` over 25k rows.

https://linear.app/n8n/issue/NODE-4830

closes #27861

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/Merge/v3/actions/mode/combineBySql.ts
+++ b/packages/nodes-base/nodes/Merge/v3/actions/mode/combineBySql.ts
@@ -14,7 +14,11 @@ import { getResolvables, updateDisplayOptions } from '@utils/utilities';
 
 import { numberInputsProperty } from '../../helpers/descriptions';
 import { modifySelectQuery, rowToExecutionData } from '../../helpers/utils';
-import { loadAlaSqlSandbox, runAlaSqlInSandbox } from '../../helpers/sandbox-utils';
+import {
+	loadAlaSqlSandbox,
+	resetSandboxCache,
+	runAlaSqlInSandbox,
+} from '../../helpers/sandbox-utils';
 
 type OperationOptions = {
 	emptyQueryResult: 'success' | 'empty';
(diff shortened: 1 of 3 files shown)


packages/nodes-base/nodes/Merge/v3/actions/mode/combineBySql.ts
```diff
@@ -78,17 +82,24 @@ const displayOptions = {
...
+	const message = isDisposed
+		? 'Dataset too large for the SQL sandbox'
+		: isTimeout
+			? 'SQL query exceeded the 30 second execution limit'
+			: 'Issue while executing query';
+	const description = isDisposed
+		? 'Try filtering or aggregating upstream, or split the input into smaller batches before the Merge node.'
+		: isTimeout
+			? 'Simplify the query (remove unnecessary JOINs) or reduce the number of input rows.'
+			: raw;
```
elsmr, line 100: nit: could be more readable to extract a small function that returns `{message, description}`

elsmr (approved):
Great fix! 💯

PR #29005, opened 2026-04-23
feat: Enable log streamin settings via env vars

ireneea:
## Summary

Adds two env vars for declaratively managing log streaming destinations:

- `N8N_LOG_STREAMING_MANAGED_BY_ENV` — when `true`, destinations are reconciled from env on every startup
- `N8N_LOG_STREAMING_DESTINATIONS` — JSON array of destinations (`webhook` / `syslog` / `sentry`)

Reconciliation matches entries by `id` or `(type, label)`: existing rows are updated in place, new ones inserted, unmatched rows deleted. Invalid JSON or schema failures raise an `InstanceBootstrappingError` and halt startup.

When `managedByEnv` is enabled:
- The controller rejects create/update/delete with `403 Forbidden`
- `managedByEnv` is exposed on frontend settings; the UI renders the destinations list read-only with a notice explaining env-managed state

[loom demo](https://www.loom.com/share/d5d79346cf884b4a8664f64e94022de0)

closes LIGO-464

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
+++ b/packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
@@ -8,8 +8,10 @@ import { MessageEventBusDestinationTypeNames } from 'n8n-workflow';
 import { v4 as uuid } from 'uuid';
 import { z } from 'zod';
 
-import { EventDestinations } from '@/modules/log-streaming.ee/database/entities';
 import { EventDestinationsRepository } from '@/modules/log-streaming.ee/database/repositories/event-destination.repository';
+import { EventDestinations } from '@/modules/log-streaming.ee/database/entities';
+
+import { InstanceBootstrappingError } from '../instance-bootstrapping.error';
 
 // The env schema mirrors the UI fields exactly (see logStreaming.constants.ts):
 // fields hidden from the UI are not accepted here, and closed enums (method,
@@ -171,7 +173,14 @@ export class LogStreamingInstanceSettingsLoader {
 			'logStreamingManagedByEnv is enabled — reconciling log streaming destinations from env vars',
 		);
 
-		const items = this.parseAndValidate(this.config.logStreamingDestinations);
+		let items: EnvDestination[] = [];
+		try {
+			items = this.parseAndValidate(this.config.logStreamingDestinations);
+		} catch (error) {
+			const message = (error as Error).message ?? 'Unknown error';
+			this.logger.error(message);
+			throw new InstanceBootstrappingError(message);
+		}
 
 		await this.eventDestinationsRepository.manager.transaction(async (tx) => {
 			await this.reconcile(tx, items);
--- a/packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
+++ b/packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
@@ -3,7 +3,12 @@ import { InstanceSettingsLoaderConfig } from '@n8n/config';
 import type { EntityManager } from '@n8n/db';
 import { Service } from '@n8n/di';
 import type { MessageEventBusDestinationOptions } from 'n8n-workflow';
-import { MessageEventBusDestinationTypeNames } from 'n8n-workflow';
+import {
+	MessageEventBusDestinationSentryOptionsSchema,
+	MessageEventBusDestinationSyslogOptionsSchema,
+	MessageEventBusDestinationTypeNames,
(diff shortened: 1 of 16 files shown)


packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
```diff
@@ -0,0 +1,317 @@
...
+			}
+
+			if (item.id) {
+				if (seenIds.has(item.id)) {
+					throw new Error(
+						`N8N_LOG_STREAMING_DESTINATIONS has duplicate id "${item.id}" at index ${index}`,
+					);
+				}
+				seenIds.add(item.id);
+			} else if (item.label) {
```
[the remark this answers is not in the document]
ireneea: fixed 👍

packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
```diff
@@ -0,0 +1,319 @@
...
+					);
+				}
+				seenLabelType.add(key);
+			}
+		});
+
+		return items;
+	}
+
+	private async reconcile(tx: EntityManager, items: EnvDestination[]): Promise<void> {
```
geemanjs: I know this somewhat goes against the future direction of instance settings - but if its managed by env - should we "remove all" and then "re-insert". This reconciliation piece seems rather complex when "deleting all" and "recreating all" at startup should also work?

ireneea: That's actually a good point, because at the end of the day as long as we end up with the desired state that all that matters.

The only benefit of doing the reconciliation is if there places we actually care about keeping track of the entities. I'll double check but I don't think that would be the case

ireneea: The only side effect would be the `createdAt` basically the same destinations could be recreated on every restart which makes the field a bit useless. But I don't think that's really a big deal

ireneea: Updated 👍

packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
```diff
@@ -0,0 +1,241 @@
...
+
+const circuitBreakerSchema = z
+	.object({
+		maxFailures: z.number().int().positive().optional(),
+		failureWindow: z.number().int().min(100).optional(),
+	})
+	.strict()
+	.optional();
+
+const webhookParameterItemSchema = z
```
geemanjs: I feel like we have major duplication of logic here 🤔 - the validation should be in the "services" we are configuring - and or just reused here rather than us having to update in two places.

e.g if we changed syslog to have a new field it would be very easy to forget updating here.

geemanjs: Likewise if we added a new "type" of log sink. 

A better way would probably be to expose a `validateConfiguration` function on the services that we can use from both the UI and here.

ireneea: I've only re-used the same schemas rather than sharing common validation logic. This is because the env json uses a "user friendly" type rather than the `__type` property used the rest of the system.

packages/cli/src/instance-settings-loader/loaders/log-streaming.instance-settings-loader.ts
```diff
@@ -0,0 +1,165 @@
...
+import { InstanceBootstrappingError } from '../instance-bootstrapping.error';
+
+// Env var format reuses the canonical destination DTOs from `n8n-workflow`,
+// with three deliberate adjustments per variant:
+//  1. Replace the internal `__type: '$$MessageEventBusDestination…'` discriminator
+//     with a friendlier `type: 'webhook' | 'syslog' | 'sentry'`.
+//  2. Tighten `id` to a UUID (the DTO only requires a non-empty string).
+//  3. Switch to strict mode so unknown keys in the env JSON fail loudly
+//     rather than being silently dropped on insert.
+const envWebhookSchema = MessageEventBusDestinationWebhookOptionsSchema.omit({ __type: true })
```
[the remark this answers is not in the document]
ireneea: This was intentionally removed because we don't want to add validation that is not present in other part of the system. This should be a thin layer that builds on the existing code

geemanjs (approved):

PR #29007, opened 2026-04-23
feat(core): Add endpoint to toggle mcp access for multiple workflows

MiloradFilipovic:
## Summary
Implement the new `PATCH /mcp/workflows/toggle-access ` endpoint used to bulk toggle mcp access for workflows:

- The endpoint replaces the old `PATCH /mcp/workflows/:workflowId/toggle-access`
- The endpoint accepts one of three scopes: an explicit `workflowIds` list, a `projectId`, or a `folderId`

Other changes:
- Update logic has been moved from `mcp.settings.controller` to `bulkSetAvailableInMCP` method on `mcp.settgins.service`
- Added new methods in `workflow-finder.service.ts‎` to handle user project scope (since we are not able to use `@ProjectScope` decorator)

Closes ADO-5108

## How to test

Old flow (updating single workflow) should still work using UI controls.

I guess easiest way to test bulk update would be using browser dev console:

Paste the following helper function:

```js
const mcpToggle = async (body) => {
  const res = await fetch('/rest/mcp/workflows/toggle-access', {
    method: 'PATCH',
    credentials: 'include',
    headers: { 
        'browser-id': localStorage.getItem('n8n-browserId'), 
    },
    body: JSON.stringify(body),
  });
  const json = await res.json();
  console.log(res.status, json);
  return json;
};

```

Then, bulk update can be tested by:

```js
await mcpToggle({ availableInMCP: true, workflowIds: ['<id1>', '<id2>'] });
```
```js
await mcpToggle({ availableInMCP: true, projectId: '<projectId>' });
```
```js
await mcpToggle({ availableInMCP: false, folderId: '<folderId>' });
```

Then check if workflows are actually updated.

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/mcp/mcp.settings.service.ts
```diff
@@ -36,4 +57,102 @@ export class McpSettingsService {
...
+		dto: UpdateWorkflowsAvailabilityDto,
+	): Promise<BulkSetAvailableInMCPResult> {
+		const { availableInMCP, workflowIds, projectId, folderId } = dto;
+
+		const scopeCount = [workflowIds, projectId, folderId].filter(Boolean).length;
+		if (scopeCount !== 1) {
+			throw new BadRequestError('Provide exactly one of workflowIds, projectId or folderId');
+		}
+
+		const requestedSize = workflowIds ? new Set(workflowIds).size : 0;
```
dariacodes: Can we remove this variable and inline it right where it's used? It was a bit hard to understand what it is for

packages/cli/src/modules/mcp/mcp.settings.service.ts
```diff
@@ -36,4 +57,102 @@ export class McpSettingsService {
...
+				});
+
+				for (const row of rows) {
+					const nextSettings = removeDefaultValues(
+						{ ...(row.settings ?? {}), availableInMCP },
+						this.globalConfig.executions.timeout,
+					);
+
+					if (isEqual(row.settings, nextSettings)) {
+						noOpIds.push(row.id);
```
dariacodes: Here we're comparing settings after running the `removeDefaultValues` method. Would it be cleaner to only compare `availableInMCP` first and form a new settings object without default values, only if it differs from what we want?

packages/cli/src/modules/mcp/mcp.settings.service.ts
```diff
@@ -36,4 +57,102 @@ export class McpSettingsService {
...
+		const baselineSize = workflowIds ? requestedSize : candidateIds.length;
+
+		if (candidateIds.length === 0) {
+			return { updatedCount: 0, updatedIds: [], skippedCount: baselineSize };
+		}
+
+		const writtenIds: string[] = [];
+		const noOpIds: string[] = [];
+
+		await this.workflowRepository.manager.transaction(async (trx) => {
```
dariacodes: I have a concern about running everything in a single transaction. For ~10k workflows projects this can get pretty heavy: long-running transaction, lots of locks, bigger rollback risk, plus blocking other queries. 
What about switching to keeping each bulk in a separate transaction? Since each workflow can be updated independently and we're filtering out those that already have the right values, I think it’s fine to allow partial success here, if we report it clearly.

packages/cli/src/modules/mcp/mcp.settings.service.ts
```diff
@@ -36,4 +57,102 @@ export class McpSettingsService {
...
+						{ settings: nextSettings, updatedAt: now },
+					);
+
+					writtenIds.push(row.id);
+				}
+			}
+		});
+
+		// To make the endpoint idempotent, we still count no-ops (workflows that are already in the requested state) as updated
+		const confirmedIds = [...writtenIds, ...noOpIds];
```
dariacodes, line 145: For big projects this array can be too big to return to the client. What if we return this only for `workflowIds` mode or just cap it to 1000?

packages/cli/src/modules/mcp/mcp.settings.service.ts
```diff
@@ -36,4 +57,102 @@ export class McpSettingsService {
...
+						noOpIds.push(row.id);
+						continue;
+					}
+
+					await trx.update(
+						WorkflowEntity,
+						{ id: row.id },
+						{ settings: nextSettings, updatedAt: now },
+					);
+
```
dariacodes, line 131: I think we're missing firing of some events like running external pre-save and post-save hooks (`workflow.update`, `workflow.afterUpdate`), telemetry (`workflow-saved` with `settingsChanged`)

dariacodes (commented):
Nice work! I added a few comments, let me know if you want to jump in a call to discuss any of them

packages/cli/src/modules/mcp/mcp.settings.service.ts
MiloradFilipovic: Ah, good point. Will update

packages/cli/src/modules/mcp/mcp.settings.service.ts
MiloradFilipovic: Yeah makes sense + `workflowsId` mode is capped anyway

packages/cli/src/modules/mcp/mcp.settings.service.ts
MiloradFilipovic: Good one, I think one transaction per chunk is the way to go.

packages/cli/src/modules/mcp/mcp.settings.service.ts
MiloradFilipovic: I left that out on purpose for a couple of reasons:
- We don't do it on other places where we bulk update workflows (e.g. [here](https://github.com/n8n-io/n8n/blob/7f54ff4acf73940afc6a59df851177d2c7b1cafe/packages/cli/src/workflows/workflow.service.ee.ts#L575) and [here](https://github.com/n8n-io/n8n/blob/7f54ff4acf73940afc6a59df851177d2c7b1cafe/packages/cli/src/workflows/workflow.service.ee.ts#L449))
- Would be too expensive for a lot of workflows unless we introduce some kind of summary event, which i don't think is in scope here
- I figured mcp access is meta information so conceptually makes sense not to trigger save-related hooks on update

Let me know if this makes sense.

dariacodes: Yeah, makes sense. It's always better to have some sort of audit logs, but when it's too expensive and not very important, we can skip it. Anyway, the main logic of changing workflows has these, so we should be fine

dariacodes (approved):
Looks good! One question, but non-blocking: I noticed that `updatedCount` is actually `processedCount` as it returns the same count for updating the same data. Shall we rename it or shall we return an honest number of updated rows?

MiloradFilipovic:
> Looks good! One question, but non-blocking: I noticed that `updatedCount` is actually `processedCount` as it returns the same count for updating the same data. Shall we rename it or shall we return an honest number of updated rows?

Good callout. I did like this with the single-workflow case in mind, where noop is not a failure, so UI can continue working without updates. But now looking at it, it makes sense to distinguish between really updated workflows and skipped ones. Will add it to ne next PR that implements UI for this

PR #29013, opened 2026-04-23
feat(core): Wire Cipher to encryption key proxy for key rotation support

BGZStephen:
## Summary

Adds `encryptV2` and `decryptV2` async methods to the `Cipher` service, along with the `EncryptionKeyProxy` abstraction that bridges `packages/core` to `KeyManagerService` without creating a circular dependency.

This is a prerequisite for encryption key rotation. The existing synchronous `encrypt`/`decrypt` methods are unchanged — no call sites are broken. Call sites will be migrated to the V2 methods in follow-up PRs.

### What changed

- **`Cipher.encryptV2`** – async; when `N8N_ENV_FEAT_ENCRYPTION_KEY_ROTATION=true` and a proxy is registered, fetches the active DB key and prefixes ciphertext with the key ID (`<keyId>:<ciphertext>`). Falls back to the standard AES-256-CBC path otherwise.
- **`Cipher.decryptV2`** – async; detects the key-ID prefix to select the correct decryption key; falls back to the legacy AES-256-CBC path for un-prefixed (pre-rotation) data. `customEncryptionKey` always bypasses the proxy.
- **`EncryptionKeyProxy`** – new concrete `@Service()` in `packages/core`, constructor-injected into `Cipher`. `EncryptionKeyManagerModule` calls `proxy.setProvider(keyManager)` at init to wire up the concrete implementation without introducing a circular dependency. Follows the `DynamicCredentialsProxy` setter pattern.

### How to test

1. Run the instance without the feature flag — all existing behaviour is unchanged (`encrypt`/`decrypt` are unmodified).
2. Set `N8N_ENV_FEAT_ENCRYPTION_KEY_ROTATION=true`, call `encryptV2` with a proxy registered — verify the output is prefixed with a key ID.
3. Call `decryptV2` on a prefixed value — confirm correct plaintext is returned.
4. Call `decryptV2` on an un-prefixed (legacy) value — confirm the CBC fallback works.

https://linear.app/n8n/issue/IAM-493

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/core/src/encryption/cipher.ts
+++ b/packages/core/src/encryption/cipher.ts
@@ -1,4 +1,4 @@
-import { Container, Service } from '@n8n/di';
+import { Service } from '@n8n/di';
 
 import { InstanceSettings } from '@/instance-settings';
 import { assertUnreachable } from '@/utils/assertions';
@@ -14,11 +14,8 @@ export class Cipher {
 		private readonly instanceSettings: InstanceSettings,
 		private readonly cipherAES256GCM: CipherAes256GCM,
 		private readonly cipherAES256CBC: CipherAes256CBC,
+		private readonly encryptionKeyProxy: EncryptionKeyProxy,
 	) {}
-
-	private get encryptionKeyProxy(): EncryptionKeyProxy | undefined {
-		return Container.has(EncryptionKeyProxy) ? Container.get(EncryptionKeyProxy) : undefined;
-	}
 
 	encrypt(data: string | object, customEncryptionKey?: string): string {
 		const key = customEncryptionKey ?? this.instanceSettings.encryptionKey;
@@ -34,18 +31,19 @@ export class Cipher {
 	async encryptV2(data: string | object, customEncryptionKey?: string): Promise<string> {
 		const plaintext = typeof data === 'string' ? data : JSON.stringify(data);
 
-		if (!customEncryptionKey && process.env.N8N_ENV_FEAT_ENCRYPTION_KEY_ROTATION === 'true') {
-			const proxy = this.encryptionKeyProxy;
-			if (proxy) {
-				const keyInfo = await proxy.getActiveKey();
-				const plaintextKey = this.decryptWithInstanceKey(keyInfo.value);
-				const ciphertext = this.encryptWithKey(
-					plaintext,
-					plaintextKey,
-					keyInfo.algorithm as CipherAlgorithm,
-				);
-				return `${keyInfo.id}:${ciphertext}`;
-			}
+		if (
+			!customEncryptionKey &&
+			process.env.N8N_ENV_FEAT_ENCRYPTION_KEY_ROTATION === 'true' &&
+			this.encryptionKeyProxy.isConfigured()
+		) {
+			const keyInfo = await this.encryptionKeyProxy.getActiveKey();
+			const plaintextKey = this.decryptWithInstanceKey(keyInfo.value);
+			const ciphertext = this.encryptWithKey(
+				plaintext,
+				plaintextKey,
+				keyInfo.algorithm as CipherAlgorithm,
+			);
+			return `${keyInfo.id}:${ciphertext}`;
 		}
 
 		const key = customEncryptionKey ?? this.instanceSettings.encryptionKey;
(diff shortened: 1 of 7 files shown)


packages/core/src/encryption/cipher.ts
```diff
@@ -13,15 +14,58 @@ export class Cipher {
...
+		const key = customEncryptionKey ?? this.instanceSettings.encryptionKey;
+		return this.decryptWithKey(data, key, 'aes-256-cbc');
+	}
+
+	async encryptV2(data: string | object, customEncryptionKey?: string): Promise<string> {
+		const plaintext = typeof data === 'string' ? data : JSON.stringify(data);
+
+		if (
+			!customEncryptionKey &&
+			process.env.N8N_ENV_FEAT_ENCRYPTION_KEY_ROTATION === 'true' &&
```
guillaumejacquart: is this really necessary ? I would have say the proxy being configured is a sufficient enough check

packages/core/src/encryption/cipher.ts
```diff
@@ -13,15 +14,58 @@ export class Cipher {
...
+			return `${keyInfo.id}:${ciphertext}`;
+		}
+
+		const key = customEncryptionKey ?? this.instanceSettings.encryptionKey;
+		return this.encryptWithKey(plaintext, key, 'aes-256-cbc');
+	}
+
+	async decryptV2(data: string, customEncryptionKey?: string): Promise<string> {
+		if (!customEncryptionKey && this.encryptionKeyProxy.isConfigured()) {
+			const colonIdx = data.indexOf(':');
```
guillaumejacquart, line 59: Are we certain that ':' is specific enough that it cannot exist already in the data we're gonna decrypt ? 
I'm wondering if we shouldn't fallback to decrypt with the instance encryption key if we don't find the key id

packages/core/src/encryption/encryption-key-proxy.ts
```diff
@@ -0,0 +1,43 @@
...
+	setProvider(provider: IEncryptionKeyProvider | undefined): void {
+		this.provider = provider;
+	}
+
+	isConfigured(): boolean {
+		return this.provider !== undefined;
+	}
+
+	async getActiveKey(): Promise<KeyInfo> {
+		return await this.provider!.getActiveKey();
```
guillaumejacquart: i would handle the case if provider is not setup here just to be safe. even if it's throwing an error, i think it's better to be explicit. I think that's what we do in the other proxies

guillaumejacquart (commented):
Looks good 👍  some comments

packages/core/src/encryption/cipher.ts
BGZStephen: Yes, as the outputs from encrypt and decrypt will be base64 encoded before the prefix is applied, the colon is a sufficient delimeter

packages/core/src/encryption/cipher.ts
```diff
@@ -13,15 +14,54 @@ export class Cipher {
...
+
+		if (!customEncryptionKey && this.encryptionKeyProxy.isConfigured()) {
+			const keyInfo = await this.encryptionKeyProxy.getActiveKey();
+			const plaintextKey = this.decryptWithInstanceKey(keyInfo.value);
+			const ciphertext = this.encryptWithKey(
+				plaintext,
+				plaintextKey,
+				keyInfo.algorithm as CipherAlgorithm,
+			);
+			return `${keyInfo.id}:${ciphertext}`;
```
afitzek, line 46: Should we version this format with a prefix?

BGZStephen: I don't think so, as it's just the ID of the key, if we want to roll out new keys (new algorithm, etc) it's just a case of updating the algorithm for any new keys

packages/core/src/encryption/cipher.ts
```diff
@@ -13,15 +14,54 @@ export class Cipher {
...
 		return this.encryptWithKey(plaintext, key, 'aes-256-cbc');
 	}
 
-	decrypt(data: string, customEncryptionKey?: string) {
+	decrypt(data: string, customEncryptionKey?: string): string {
+		const key = customEncryptionKey ?? this.instanceSettings.encryptionKey;
+		return this.decryptWithKey(data, key, 'aes-256-cbc');
+	}
+
+	async encryptV2(data: string | object, customEncryptionKey?: string): Promise<string> {
```
afitzek, line 31: Should we support the customEncryptionKey parameter there?

BGZStephen: I personally don't think we need to, but left it in there after a chat with Guillaume about it

packages/core/src/encryption/cipher.ts
```diff
@@ -13,15 +14,54 @@ export class Cipher {
...
+				const keyId = data.slice(0, colonIdx);
+				const ciphertext = data.slice(colonIdx + 1);
+				const keyInfo = await this.encryptionKeyProxy.getKeyById(keyId);
+				if (!keyInfo) throw new Error(`Encryption key not found: ${keyId}`);
+				const plaintextKey = this.decryptWithInstanceKey(keyInfo.value);
+				return this.decryptWithKey(ciphertext, plaintextKey, keyInfo.algorithm as CipherAlgorithm);
+			}
+			const keyInfo = await this.encryptionKeyProxy.getLegacyKey();
+			const plaintextKey = this.decryptWithInstanceKey(keyInfo.value);
+			return this.decryptWithKey(data, plaintextKey, 'aes-256-cbc');
```
afitzek: This needs the algorithm from keyInfo I think.

packages/core/src/encryption/cipher.ts
BGZStephen: I decided to keep this to allow it to go out safely behind a feature glaf before we're ready to turn it on

packages/core/src/encryption/cipher.ts
BGZStephen: true

afitzek (approved):
Nice 🚀

PR #29014, opened 2026-04-23
feat(Oracle DB Vector Store Node): Add support for vectorstore and embedding node support

sudarshan12s:
### Oracle Vector Store node
- Work with Oracle Database Vector Support directly from n8n. 
- Provides the standard Insert, Load, Retrieve, and Retrieve-as-tool modes with distance-strategy selection.
- Supports metadata filtering, now with deep merge logic so node-level filters and ad-hoc runtime filters (including nested $and/$or arrays) are combined instead of overwritten.
- Wraps the Oracle connection pool with a lightweight adapter so each operation borrows a pool connection on demand, and gracefully handles “No rows found” by returning an empty result set.

### Oracle ONNX Embedding node

- Allow to pick models via a resource locator (list or explicit ID).
- Uses a pooled Oracle connection per embedding request, ensuring handles are returned cleanly, and produces standard AI Embedding outputs ready for downstream vector-store nodes.

**Note**: The database need to have ONNX model loaded to use this node. Details on loading ONNX model into Oracle Database are [here](https://docs.oracle.com/en/database/oracle/oracle-database/26/vecse/import-onnx-models-oracle-ai-database-end-end-example.html).

### Testing
Added an extensive mock test suite covering insert paths, list ONNX Models, filter merging of similarity search, and pool lifecycle for vectorstore node. 

`pnpm --filter @n8n/n8n-nodes-langchain test -- VectorStoreOracleDB EmbeddingsOracleDb

insertdocsconfigvectorstore

InsertDocumentsintovectorstore

onnxembeddingnode

vectorstoreconfigwithAIAgent

vectorstorewithAIAgent

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
```diff
@@ -0,0 +1,251 @@
...
+		const config: OracleDBVSArgs = {
+			client,
+			tableName,
+			query,
+			filter: filter as ExtendedOracleDBVectorStore['FilterType'] | undefined,
+		};
+
+		config.distanceStrategy = context.getNodeParameter(
+			'options.distanceStrategy',
+			0,
```
dlavrenuek: This should be `itemIndex` instead of `0`. Since `options.distanceStrategy` can contain an expression it can actually have different values for different items


```suggestion
			itemIndex,
```

packages/nodes-base/package.json
```diff
@@ -881,7 +881,7 @@
     "@types/mime-types": "catalog:",
     "@types/mssql": "^9.1.5",
     "@types/nodemailer": "^7.0.3",
-    "@types/oracledb": "^6.9.1",
+    "@types/oracledb": "^6.10.0",
```
dlavrenuek, line 890: This only updates `@types/oracledb` to ^6.10.0 but the actual `oracledb` is still defined as `6.9.0` in the same file. The lockfile shows that `"@oracle/langchain-oracledb":"0.2.0",` has a peer dependency on `oracledb: ^6.10.0`, which means that `oracledb` should be updated

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+const createLazyOraclePool = (
+	context: IExecuteFunctions | ISupplyDataFunctions,
+	credentials: OracleDBNodeCredentials,
+) => {
+	const getPool = async () => await configureOracleDB.call(context, credentials);
+	// OracleVS expects an oracledb.Pool, but it only uses getConnection/close.
+	// We provide a lightweight wrapper that matches that subset of the interface.
+	return new LazyOraclePool(getPool) as unknown as oracledb.Pool;
```
dlavrenuek, line 141: This cast breaks typescript contracts since `LazyOraclePool` does not fulfill `oracledb.Pool` type. When accessing any properties other than `getConnection` or `close` there will be no build time TS error but a runtime error. I can not see why a lazy pool is even needed, anywhere where it's called a real connection will be established, or do I miss anything?

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
```diff
@@ -0,0 +1,251 @@
...
+	constructor(private readonly getPool: () => Promise<oracledb.Pool>) {}
+
+	async getConnection(): Promise<oracledb.Connection> {
+		const pool = await this.getPool();
+		return await pool.getConnection();
+	}
+
+	async close(): Promise<void> {
+		// ConnectionPoolManager owns lifecycle; keep the shared pool alive.
+	}
```
dlavrenuek, line 131: While I did not test it, this code looks like it will lead to a memory leak (close does not reach the actual pool). If there is no benefit in using a `LazyOraclePool`, let's just use a real one and remove `LazyOraclePool`

packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOracleDB/listModels.ts
```diff
@@ -0,0 +1,39 @@
...
+	this: ILoadOptionsFunctions,
+	filter = '',
+): Promise<INodeListSearchResult> {
+	const credentials = await this.getCredentials('oracleDBApi');
+	const pool = await configureOracleDB.call(this, credentials as OracleDBNodeCredentials);
+	const connection = await pool.getConnection();
+
+	try {
+		const result = await connection.execute<MiningModelRow>(
+			'select model_name from user_mining_models',
```
dlavrenuek: While it is unlikely it's not impossible and some users might encounter this case: if someone has a lot of user_mining_models (for example 100000) this operation will select all of the rows and then filter the results programmatically. I suggest using `filter` directly in the statement via bind params

dlavrenuek (changes requested):
Thank you for the improvement @sudarshan12s, I did test the implementation and it works as expected 🎉
I've added some remarks on the code. The lockfile is also broken, but I can fix that before merging (assuming the fork has default access for maintainers)

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
sudarshan12s: The `LazyOraclePool` is intentional because i wanted to use n8n  ConnectionPoolManager in nodes-base which manages pools.

We cannot pass a concrete `oracledb.Pool` at initialization time because n8n resolves pools through `ConnectionPoolManager` when a connection is needed. `LazyOraclePool` adapts `OracleVS`’s pool-based API to n8n’s managed pool lifecycle, so the pool is fetched from `ConnectionPoolManager` only when `OracleVS` actually needs a connection.

I agree the full `oracledb.Pool` cast is not ideal. I can think of doing some changes in  `OracleVS` in coming releases of `langchain-oracledb` module to accept a connection provider/factory API instead of requiring a concrete `oracledb.Pool` to avoid run time issues .

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
sudarshan12s: I have written tests to show  how multiple OracleVectorStore Nodes can use same pool .  It can also be used by Oracle SQL node as the pools are managed by n8n `ConnectionPoolManager` (`connection-pool-manger.ts`) 


I have created a `OracleVectorStore` node in `n8n` for document ingestion.  This called `oracledb.createPool `from 
`packages/nodes-base/nodes/Oracle/Sql/transport/index.ts `


The cleanup of pool is also issued after 5 mins from 
`ConnectionPoolManager` (`connection-pool-manger.ts`)
So it called `await pool.close();` from `packages/nodes-base/nodes/Oracle/Sql/transport/index.ts `

The `LazyOraclePool` `close` would not be hence required..

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
sudarshan12s: Thanks i added it. I followed other vectorstore nodes like `VectorStorePGVector.node.ts` thinking its not normal to modify this per item.

packages/nodes-base/package.json
sudarshan12s: updated `oracledb` to 6.10.0

packages/@n8n/nodes-langchain/nodes/embeddings/EmbeddingsOracleDB/listModels.ts
sudarshan12s: I used a sql with filter

`'SELECT model_name FROM user_mining_models WHERE INSTR(UPPER(model_name), :modelNameFilter) > 0 ORDER BY model_name',`

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
sudarshan12s: just to add on `OracleVS`:

In `langchain-oracledb`, `OracleVS` accepts `oracledb.Pool | oracledb.Connection`, but in practice it only needs that client to acquire a connection and execute SQL. For the current `OracleVS` usage, the adapter surface is sufficient and covered by tests: connections are returned after use, the managed pool remains open, and `ConnectionPoolManager` cleanup closes the underlying pool.

sudarshan12s:
Hi @dlavrenuek   — could you take a look at my PR and share your reviews/feedback when you get a chance?

Thanks for your time.

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
dlavrenuek: Thank you for the insights. I can follow the explanation but still don't see the need for this implementation. Instead of

```ts
const client = createLazyOraclePool(context, credentials);
```

You can use 

```ts
const client = await configureOracleDB.call(context, credentials);
```

`configureOracleDB` already uses `ConnectionPoolManager`. Since  `close()` should be a NOOP the same logic as now in the code can be achieved by turning `releaseVectorStoreClient` into a NOOP

```
	releaseVectorStoreClient(_vectorStore) {},
```

Right now the client is created in `populateVectorStore` and in `getVectorStoreClient`, these are only called on execution. Unless there is anything else I miss I suggest removing the lazy pool class

dlavrenuek: PS: I did test these changes locally and these work as expected

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.test.ts
```diff
@@ -0,0 +1,596 @@
+// cspell:ignore langchain vectorstores vectorstore oracledb XEPDB Demis Hassabis
```
dlavrenuek: How did `Demis Hassabis` end up here? 😄 But on the serious side: let's not use real people names in here, please replace these by `John Doe` or similar

dlavrenuek (changes requested):
Sorry for the delay @sudarshan12s - the pool handling requires some time for setup and testing and the last days/weeks have been pretty busy. I did fix the lockfile and lint errors but still see some changes needed. Let me know when these are ready and I promise to do a review in a timely manner next time

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.test.ts
sudarshan12s: sorry, I corrected my comment.

dlavrenuek: Did you may be forget to push your changes or are these still in progress?

sudarshan12s: ohh I get it now.  I have removed the names and just used Author1, Author2.

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
sudarshan12s: Using:

`const client = await configureOracleDB.call(context, credentials);`

would pass a concrete pool instance to `OracleVS`.

`OracleVS.fromDocuments()` roughly does:

```
await vss.initialize();
const vectors = await embeddings.embedDocuments(texts);
await vss.addVectors(vectors, documents, options);
```

So for large inputs or slow embedding providers, `ConnectionPoolManager` may mark the pool as stale and close it during the embedding step, before `addVectors()` asks for a connection. The lazy wrapper avoids holding that possibly stale pool reference. Each `OracleVS.getConnection()` call resolves the current managed pool through `ConnectionPoolManager`.

Does it make sense?


One related edge case I noticed is cleanup while database work is still in progress. If a connection is busy(injesting documents / executing long queries,..)  when `ConnectionPoolManager` attempts to close the pool, that should be handled separately from this PR. I’d prefer to address that in a follow-up PR so this change stays focused on adapting `OracleVS` to n8n’s managed-pool lifecycle.

sudarshan12s:
On the failed Snyk checks, I can’t view the details due to a permissions error.

I ran a local `pnpm audit --prod`. I don’t see a vulnerability path reported related to this PR changes. 

Could someone with Snyk access share the affected package and dependency path?

Thanks.

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
sudarshan12s: I raised [PR](https://github.com/n8n-io/n8n/pull/31059) for the edge case above . 

As per  lazy pool class , I am retaining it so that a stale pool is not stored inside OracleVS. But i am working on improvements on OracleVS to consider pool [provider](https://github.com/oracle/langchain-oracle/pull/229) as a function. Once that change is in, I can then remove lazy pool class and just use the following for `client` argument:

`client: async () => await configureOracleDB.call(context, credentials)
`

That returns an oracledb.Pool and  it returns the current managed pool at the time `OracleVS` needs a connection, instead of storing one concrete pool up front.

I hope this makes sense.. Let me know if i am missing something. 

Thanks for your time.

dlavrenuek (approved):
Thank you for the contribution 🎉

packages/@n8n/nodes-langchain/nodes/vector_store/VectorStoreOracleDB/VectorStoreOracleDB.node.ts
dlavrenuek: Ah, having OracleVS itself it's own lifecycle management is definitely a problem because we can not close the pool itself

PR #29019, opened 2026-04-23
fix(core): Broadcast MCP availability change via push connection 

MiloradFilipovic:
## Summary
This PR fixes a bug that made workflow checksums de-sync if toggling mcp access from a different tab. The bug can be reproduced by (video in [the Linear ticket](https://linear.app/n8n/issue/ADO-5111/bug-workflow-autosave-breaks-if-mcp-toggle-is-updated-from-another)):

1. Opening workflow list in one browser tab (list tab)
2. Opening a workflow in another tab (canvas tab)
3. Update workflow's MCP access on the list tab
4. Change something in the canvas tab and wait for autosave to kick in -> warning modal shows and save fails

### The fix
Broadcast a new `workflowMcpAvailabilityChanged` push event on MCP toggle so open tabs (including the writer) can patch just the availableInMCP flag and refresh their `checksum`. 
I decided not to reuse the existing `workflowUpdated` because it deliberately skips the writer tab and does a full-workflow refetch that would overwrite unsaved edits.

Fixes ADO-5111

## How to test

Follow steps to reproduce and make sure autosave still works after toggling MCP access.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/mcp/mcp.settings.controller.ts
+++ b/packages/cli/src/modules/mcp/mcp.settings.controller.ts
@@ -17,6 +17,7 @@ import { UpdateWorkflowAvailabilityDto } from './dto/update-workflow-availabilit
 import { McpServerApiKeyService } from './mcp-api-key.service';
 import { McpSettingsService } from './mcp.settings.service';
 
+import { CollaborationService } from '@/collaboration/collaboration.service';
 import { NotFoundError } from '@/errors/response-errors/not-found.error';
 import { listQueryMiddleware } from '@/middlewares';
 import type { ListQuery } from '@/requests';
@@ -32,6 +33,7 @@ export class McpSettingsController {
 		private readonly mcpServerApiKeyService: McpServerApiKeyService,
 		private readonly workflowFinderService: WorkflowFinderService,
 		private readonly workflowService: WorkflowService,
+		private readonly collaborationService: CollaborationService,
 	) {}
 
 	@GlobalScope('mcp:manage')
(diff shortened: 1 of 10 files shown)


packages/cli/src/modules/mcp/mcp.settings.controller.ts
```diff
@@ -123,6 +125,22 @@ export class McpSettingsController {
 
 		const updatedWorkflow = await this.workflowService.update(req.user, workflowUpdate, workflowId);
 
+		// Notify other tabs / users that have this workflow open so they can
+		// resync their local `availableInMCP` and `expectedChecksum` without
+		// waiting for a full reload. Non-blocking: a broadcast failure must
+		// not fail the write.
+		try {
```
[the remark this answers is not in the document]
MiloradFilipovic: Makes sense, will update

packages/frontend/editor-ui/src/app/composables/usePushConnection/handlers/workflowMcpAvailabilityChanged.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	// Only the editor tab needs to resync the document store + checksum.
+	if (workflowId !== workflowsStore.workflowId) return;
+
+	const workflowDocumentStore = useWorkflowDocumentStore(createWorkflowDocumentId(workflowId));
+	workflowDocumentStore.mergeSettings({ availableInMCP });
+
+	try {
+		// Refresh the expectedChecksum so the next normal save doesn't 409.
+		const workflow = await getWorkflow(rootStore.restApiContext, workflowId);
```
dariacodes: Shall we send the checksum along with the event so there is no need to refetch it here?

dariacodes (commented):
Thanks for fixing this bug! Setting one specific setting outside canvas will always require a need for such a push event, so I'm a bit concerned about the way of creating separate push events for each settings (if we have more toggles in the future). Would it be better to have a generic "settings updated" event?

packages/frontend/editor-ui/src/app/composables/usePushConnection/handlers/workflowMcpAvailabilityChanged.ts
MiloradFilipovic: Let me revisit this, I thought about it but seemed like we need to calculate it in-place so wanted to over-complicate the solution.

MiloradFilipovic: Alright, added the `checksum` to the payload and removed extra `GET`

dariacodes (approved):
I tested it - works well!
Thanks for addressing feedback 🌟

PR #29021, opened 2026-04-23
fix(task-runner-python): allow strptime imports

### LINKED ISSUE #29010: Python native Code node: datetime.strptime() fails in task runner sandbox with "'mappingproxy' object has no attribute '__import__'"
## Bug description

When using a Python native `Code` node, calling `datetime.strptime()` fails with:

```text
'mappingproxy' object has no attribute '__import__'
```

This appears to be caused by the Python task runner sandbox, not by the workflow data itself.

## To reproduce

Create a workflow with a Python native `Code` node.

This works:

```python
from datetime import datetime

return {
  "ok": True
}
```

This fails:

```python
from datetime import datetime

datetime.strptime("2026-04-23T16:04:28+0000", "%Y-%m-%dT%H:%M:%S%z")

return {
  "ok": True
}
```

I was also able to reproduce it directly inside the task runner environment:
- `from datetime import datetime` works
- `datetime.strptime(...)` throws the same error

## Error

```text
Error: 'mappingproxy' object has no attribute '__import__'
    at throwExecutionError (.../nodes/Code/throw-execution-error.ts:11:9)
    at PythonTaskRunnerSandbox.runUsingIncomingItems (.../nodes/Code/PythonTaskRunnerSandbox.ts:69:30)
    at processTicksAndRejections (node:internal/process/task_queues:104:5)
    at ExecuteContext.execute (.../nodes/Code/Code.node.ts:201:12)
```

From the Python runner traceback:

```text
Traceback (most recent call last):
  File "/opt/runners/task-runner-python/.venv/lib/python3.13/site-packages/src/task_executor.py", line 257, in _per_item
    exec(compiled_code, globals)
  File "<per_item_task_execution>", line 6, in <module>
  File "<per_item_task_execution>", line 3, in _user_function
AttributeError: 'mappingproxy' object has no attribute '__import__'
```

## Expected behavior

`datetime.strptime()` should work in Python native Code nodes the same way other standard-library datetime parsing methods do.

## Actual behavior

`datetime.strptime()` crashes inside the Python task runner sandbox with:

```text
'mappingproxy' object has no attribute '__import__'
```

## Environment

- n8n version: `2.17.6`
- Deployment: Self-hosted
- Execution mode: `queue`
- Python runner: external task runners
- Custom runner image based on: `n8nio/runners:2.17.6`
- Python version inside runner: `3.13.13`

## Additional notes

I tested these combinations:

- `runOnceForAllItems` + `from datetime import datetime` => works
- `runOnceForEachItem` + `from datetime import datetime` => works
- `runOnceForEachItem` + `datetime.strptime(...)` => fails
- `datetime.fromisoformat(...)` works as a workaround for the same timestamp format

So the issue seems specifically related to `datetime.strptime()` in the sandboxed Python runner, likely due to an internal import path interacting badly with the sandboxed `__builtins__`.

## Workaround

Using `datetime.fromisoformat(...)` instead of `datetime.strptime(...)` works for ISO-like timestamps such as:

```python
from datetime import datetime

dt = datetime.fromisoformat("2026-04-23T16:04:28+0000")
```

### To Reproduce

Create a workflow with a Python native `Code` node.

This works:

```python
from datetime import datetime

return {
  "ok": True
}
```

This fails:

```python
from datetime import datetime

datetime.strptime("2026-04-23T16:04:28+0000", "%Y-%m-%dT%H:%M:%S%z")

return {
  "ok": True
}
```

I was also able to reproduce it directly inside the task runner environment:
- `from datetime import datetime` works
- `datetime.strptime(...)` throws the same error

### Expected behavior

`datetime.strptime()` should work in Python native Code nodes the same way other standard-library datetime parsing methods do.

### Debug Info

# Debug info

## core

- n8nVersion: 2.17.6
- nodeJsVersion: 24.14.1
- nodeEnv: production
- database: postgres
- concurrency: -1
- consumerId: b048c93d-46e8-48c0-9876-8e56904f8f06

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: database

## pruning

- enabled: true
- maxAge: 168 hours
- maxCount: 50000 executions

## client

- isTouchDevice: false

Generated at: 2026-04-23T16:16:28.162Z

### Operating System

Ubuntu 20.04.6 LTS

### n8n Version

2.17.6

### Node.js Version

24.14.1

### Database

PostgreSQL

### Execution mode

queue

### Hosting

self hosted

corazzione:
## Summary
- stop wrapping filtered Python builtins in `MappingProxyType`
- keep the safe `__import__` override in place while restoring stdlib compatibility for `datetime.strptime()`
- add regression coverage for the `strptime()` per-item execution path and update the builtins unit expectation

## Root cause
The Python task runner injects `__builtins__` as a `MappingProxyType`. That breaks stdlib code paths used by `datetime.strptime()`, which expect builtins to expose import machinery the same way normal exec globals do. In practice this crashes with `"mappingproxy object has no attribute __import__"` before user code can complete.

## Changes
- return a plain filtered builtins dict from `_filter_builtins()`
- preserve the sandboxed safe import function on `__import__`
- add an integration regression test for `datetime.strptime()` in per-item mode
- update the builtins unit test to match the new return type

## Validation
- `git diff --check`
- local Python reproduction outside the runner: `MappingProxyType(dict(__builtins__))` reproduces the reported `__import__` failure, while `dict(__builtins__)` allows the same `datetime.strptime()` call to succeed
- attempted: `uv run pytest tests/unit/test_task_executor.py -k filter_builtins`
- attempted: `uv run pytest tests/integration/test_execution.py -k datetime_strptime_works`

## Notes
The `uv run pytest` commands are blocked in this environment because the task runner package assumes Unix process primitives (`ForkServerProcess`) and the available environment here is Windows, so the full runner test suite could not be completed locally.

Fixes #29010

### DIFF
--- a/packages/@n8n/task-runner-python/src/task_executor.py
+++ b/packages/@n8n/task-runner-python/src/task_executor.py
@@ -436,7 +436,7 @@ class TaskExecutor:
 
         filtered["__import__"] = TaskExecutor._create_safe_import(security_config)
 
-        return types.MappingProxyType(filtered)
+        return filtered
 
     @staticmethod
     def _sanitize_sys_modules(security_config: SecurityConfig):
--- a/packages/@n8n/task-runner-python/tests/integration/test_execution.py
+++ b/packages/@n8n/task-runner-python/tests/integration/test_execution.py
@@ -185,6 +185,27 @@ async def test_per_item_with_continue_on_fail(broker, manager):
     assert "division by zero" in done_msg["data"]["result"][0]["json"]["error"]
 
 
+@pytest.mark.asyncio
+async def test_per_item_datetime_strptime_works(broker, manager):
+    task_id = nanoid()
+    items = [{"json": {"value": "2026-04-23T16:04:28+0000"}}]
+    code = textwrap.dedent("""
+        from datetime import datetime
+
+        parsed = datetime.strptime(_item['json']['value'], "%Y-%m-%dT%H:%M:%S%z")
+        return {'parsed': parsed.isoformat()}
+    """)
+    task_settings = create_task_settings(code=code, node_mode="per_item", items=items)
+    await broker.send_task(task_id=task_id, task_settings=task_settings)
+
+    done_msg = await wait_for_task_done(broker, task_id)
+
+    assert done_msg["taskId"] == task_id
+    assert done_msg["data"]["result"] == [
+        {"json": {"parsed": "2026-04-23T16:04:28+00:00"}, "pairedItem": {"item": 0}}
+    ]
+
+
 # ========== Security ===========
 
 
--- a/packages/@n8n/task-runner-python/tests/unit/test_task_executor.py
+++ b/packages/@n8n/task-runner-python/tests/unit/test_task_executor.py
@@ -1,6 +1,5 @@
 import pytest
 import json
-import types
 from unittest.mock import MagicMock, patch
 
 from src.task_executor import TaskExecutor
@@ -209,7 +208,7 @@ class TestFilterBuiltins:
             runner_env_deny=False,
         )
 
-    def test_returns_mapping_proxy(self):
+    def test_returns_dict(self):
         result = TaskExecutor._filter_builtins(self._make_security_config())
 
-        assert isinstance(result, types.MappingProxyType)
+        assert isinstance(result, dict)


ivov:
@corazzione Thanks for the PR. We went a different route to keep the builtins read-only (#29776) since the proxy was deliberate. Closing in favor of that one.

PR #29022, opened 2026-04-23
fix(Http Request Node): Preserve multipart filenames for binary uploads

### LINKED ISSUE #28854: HTTP Request node: File not attached when sending multipart-form-data (400 "No file field in request")
### Bug Description

When using the HTTP Request node (v4.4) to send a binary file via multipart-form-data, the request is sent but the file is not properly attached. The receiving API returns 400 - "No file field in request".

The same request works correctly when executed via curl, confirming the API endpoint is functioning properly.

Additional observation:
In the Webhook node, I set "Field Name for Binary Data" to invoice_file, but the output still shows the binary data under data0. This may be related - if the binary property key isn't being set correctly, the HTTP Request node may not be finding the binary data at the expected location.

Screenshot attached showing:

Image

"Field Name for Binary Data" set to invoice_file
Output showing binary data under data0 instead

### To Reproduce

1. Create a workflow with binary data available (e.g., PDF from a previous node)
2. Add an HTTP Request node configured to send the file via multipart-form-data
3. Execute the workflow
4. API returns 400 error: "No file field in request"

**Node Configuration**
{
  "parameters": {
    "method": "POST",
    "url": "https://example.com/api/endpoint",
    "authentication": "genericCredentialType",
    "genericAuthType": "httpBearerAuth",
    "sendBody": true,
    "contentType": "multipart-form-data",
    "bodyParameters": {
      "parameters": [
        {
          "parameterType": "formBinaryData",
          "name": "file",
          "inputDataFieldName": "data0"
        }
      ]
    }
  },
  "type": "n8n-nodes-base.httpRequest",
  "typeVersion": 4.4
}

**Error Response:**
{
  "errorMessage": "Bad request - please check your parameters",
  "errorDescription": "No file field in request",
  "errorDetails": {
    "rawErrorMessage": ["400 - \"{\\\"error\\\":\\\"No file field in request\\\"}\\n\""],
    "httpCode": "400"
  }
}

### Expected behavior

The file should be attached to the multipart-form-data request with the field name "file", and the API should receive it correctly.

Actual behavior: The request is sent, but the file field is missing from the request body. The API returns a 400 error indicating no file was received.

### Debug Info

The API works correctly with curl:
curl -X POST "https://example.com/api/endpoint" \
  -H "Authorization: Bearer TOKEN" \
  -F "file=@/path/to/document.pdf"

### Operating System

Linux Ubuntu LTS 24.4

### n8n Version

2.17.3

### Node.js Version

v24.14.1

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

corazzione:
## Summary
- preserve a filename when the HTTP Request node uploads binary data as multipart form-data
- add regression coverage for multipart binary requests and the missing-fileName case

## Root cause
When the incoming binary data had no `fileName`, the multipart part was serialized without a `filename` attribute. Some APIs only recognize that part as a file upload when `filename` is present, so they responded as if no file field had been sent.

## Changes
- default the multipart upload filename to `file` when binary metadata does not include `fileName`
- add tests to verify the node sends a `FormData` payload and keeps a fallback filename in the multipart body

## Validation
- `pnpm --dir packages/nodes-base exec jest nodes/HttpRequest/test/node/HttpRequestV3.test.ts --runInBand`
- `pnpm --dir packages/nodes-base exec eslint nodes/HttpRequest/V3/HttpRequestV3.node.ts nodes/HttpRequest/test/node/HttpRequestV3.test.ts`
- `git diff --check`

Fixes #28854


michael-radency (approved):
Thanks for the fix @corazzione , LGTM 👍

BerniWittmann (approved):

PR #29024, opened 2026-04-23
test(benchmark): Question-driven Playwright benchmark suite with tiered topology and rich diagnostics

shortstacked:
## Summary

Refactors the Playwright infrastructure benchmark suite into a question-driven architecture: each spec answers ONE scaling question, the topology is declared per-spec via a single `benchConfig()` helper, and the reporter groups results by question with per-stage variant rows.

Specs are organised in four explicit tiers, each answering a different *kind* of question:

| Tier | Topology | Purpose |
|------|----------|---------|
| **Peak** | 1m direct, no workers | Architectural ceiling — no queue tax, no worker dispatch |
| **Actual** | 1m + 1w queue mode | Real-world minimum HA topology — what a basic production setup actually delivers |
| **Scaling** | 2m + 2w queue mode | HA distribution check — does doubling capacity ~double the actual baseline? |
| **Cost** | matches the baseline | What does turning on configuration X cost vs the baseline? (e.g. OTEL overhead) |

## Spec layout

| Trigger | Tier | Spec | Question |
|---|---|---|---|
| kafka | Peak | `single-instance-ceiling` | How much can we process on a single instance? |
| kafka | Peak | `steady-rate-breaking-point` | At what input rate does the system fall behind? |
| kafka | Actual | `queue-mode-sustained-rate` | Can queue mode sustain 250 msg/s steady? |
| kafka | Actual | `burst-drain-capacity` | How fast can we drain a backlog? |
| kafka | Actual | `node-count-scaling` | How does throughput scale with workflow complexity? |
| kafka | Actual | `output-size-impact` | What is the impact of node output size on throughput? |
| webhook | Peak | `webhook-single-instance` | What is the single-instance webhook ingestion ceiling? |
| webhook | Scaling | `webhook-main-scaling` | Does webhook ingestion scale linearly with main count? |
| webhook | Cost | `webhook-otel-overhead` | What is the runtime cost of enabling OTEL? |

### Architecture
- **One config helper** (`benchConfig()`) replaces four factory functions and two env profiles. Specs declare topology with `{ kafka, mains, workers, tracing }` flags.
- **Caddy load balancer policy** is configurable via `lbPolicy`; benchmarks default to `round_robin` so multi-main load actually distributes (UI tests keep the default `first` for debuggability).
- **Auto-applied multi-main env** (`N8N_MULTI_MAIN_SETUP_ENABLED`) when `mains > 1`.

### Reporter
- Question-grouped tables — each `describe()` title is a heading; rows underneath show the variants.
- New `Variant` and `Verdict` columns sourced from metric dimensions.
- Empty columns trimmed per-question — no more walls of `—`.
- Test scenario printed as a one-line context under each question heading.
- Markdown output to `$GITHUB_STEP_SUMMARY` mirrors the console table.

### Per-stage / variant rows
- Staged-rate runs (`steady-rate-breaking-point`) emit one reporter row per stage, tagged with rate and verdict (`kept up` / `falling behind` / `saturated`) so the breaking point is obvious.
- Ramp specs (`node-count-scaling`, `output-size-impact`) emit one row per variant.

### Diagnostics
- **cAdvisor service** for per-container CPU/memory/IO; surfaces in a `[CONTAINERS]` block.
- **pg_stat_statements** breakdown now ranked by total ms/s (calls × avg) rather than raw call count, so trivial high-frequency statements don't crowd out the work that actually moves throughput.
- **`[PG SATURATION]` block**: total query CPU including the long tail, buffer hit ratio, bgwriter / WAL pressure, `pg_stat_io` per-backend-type IO.
- **Jaeger trace artifacts** (`jaeger-traces.json`) when a spec opts into `tracing: true` (e.g. `webhook-otel-overhead`). Includes an `import-jaeger-traces.mjs` helper for local replay.
- **`pg_stat_statements.track_planning`** and **`track_io_timing`** are always on. Originally gated behind a flag for fear of overhead, but measured impact at our load was negligible — the planner/IO time data is too useful to omit.

### Load generation
- Webhook harness uses `pipelining: 3` so the load generator doesn't artificially cap throughput at `connections / latency` when the server is faster than autocannon's connection model.
- Webhook warm-up moved into `setupBenchmarkRun` so the baseline counter is captured *after* the probe POST — probe traffic doesn't pollute the measurement window.

## CI

The full suite runs on `blacksmith-8vcpu-ubuntu-2204` via `.github/workflows/test-e2e-infrastructure-reusable.yml`, single container at a time. Each spec spawns its own topology.

## Test plan

- [x] Typecheck and lint clean (`pnpm --filter=n8n-playwright typecheck && pnpm --filter=n8n-playwright lint`)
- [x] All 9 specs run end-to-end on the Blacksmith runner
- [x] Reporter renders a Question-grouped table in the GitHub Step Summary
- [x] `webhook-otel-overhead` produces a `jaeger-traces.json` artifact
- [ ] Compare actual-tier numbers against current production benchmarks for sanity


shortstacked:
@claude Review please

packages/testing/playwright/tests/infrastructure/benchmarks/harness/webhook-throughput-harness.ts
Matsuuu: Plan and workerPlan are options here, but [are not destructured ever](https://github.com/n8n-io/n8n/blob/749d6c4332122d637b9b0935a1f75df8231efe52/packages/testing/playwright/tests/infrastructure/benchmarks/harness/webhook-throughput-harness.ts#L53) in the actual `runWebhookThroughputTest` function. Should we remove them from the interface as well as from the callsites?

packages/testing/containers/services/postgres.ts
```diff
@@ -45,12 +46,37 @@ export const postgres: Service<PostgresResult> = {
...
 				'max_connections=200',
+				'-c',
+				'shared_preload_libraries=pg_stat_statements',
+				'-c',
+				'pg_stat_statements.track=all',
+				'-c',
+				'pg_stat_statements.max=10000',
+				// Track plan time so query CPU includes planner overhead (often >50%
+				// for TypeORM's parameterized-but-not-prepared statements).
+				'-c',
```
Matsuuu, line 57: PR says 

> BENCH_DEEP_DIAGNOSTICS=true env opt-in for pg_stat_statements.track_planning and track_io_timing — adds noticeable PG overhead so it's gated.

But here it seems to just be always on?

packages/testing/playwright/playwright-projects.ts
```diff
@@ -43,73 +43,117 @@ const CONTAINER_CONFIGS: Array<{ name: string; config: N8NConfig }> = [
...
+		...(opts.tracing && {
+			N8N_OTEL_ENABLED: 'true',
+			N8N_OTEL_EXPORTER_OTLP_ENDPOINT: 'http://jaeger:4318',
+			N8N_OTEL_EXPORTER_SERVICE_NAME: `n8n-bench-${isolation}`,
+			N8N_OTEL_TRACES_INCLUDE_NODE_SPANS: 'true',
+		}),
+		...opts.env,
+		TEST_ISOLATION: `bench-${isolation}`,
+	};
+	if ((opts.mains ?? 1) > 1) env.N8N_MULTI_MAIN_SETUP_ENABLED = 'true';
```
Matsuuu, line 140: According to [this](https://github.com/n8n-io/n8n/pull/29024/changes#diff-48c3d5a87c84a4126af39ab6fc890c624cd9777381c4d00d572330678f0b39fcR17) , multimain requires the License to be set. Should we verify it's present when `N8N_MULTI_MAIN_SETUP_ENABLED === true`?

Matsuuu (commented):
A couple of comments to resolve before going through

packages/testing/containers/services/postgres.ts
shortstacked: Yes the description is out of date. Initially that was the case, but I measured and it doesn't actually impact that much. I'll update.

packages/testing/playwright/playwright-projects.ts
shortstacked: This is enforced at the n8n service layer — `packages/testing/containers/services/n8n.ts:89-93` throws if no `N8N_LICENSE_ACTIVATION_KEY`/`N8N_LICENSE_CERT` is set when `mains > 1`. Container fails to start before the env var has any effect, so adding a duplicate guard in `benchConfig()` would just be a fail-fast.

Matsuuu (approved):
🚀

PR #29028, opened 2026-04-23
fix(core): Make instance AI prefer newer AI models (no-changelog)

Cadiac:
## Summary

Making the instance AI builder & orchestrator to prefer using the latest models instead of the agent's training knowledge, by...

- Providing `builderHint`'s on all model nodes
  - image
- Instructing agents to use `explore-resources` to find available models
- Having `explore-resources` errors explain how to recover from errors
  - these made a big difference, now the agent recovers on failing to use them 
  - image
- Having `explore-resources` also return related `builderHints` so that the agent sees them if it skips reading `node-definitions`
  - image
- Removed mentions of the old models in the prompts

With this, we seem to much more consistently use the new models - but we'll likely have to adjust these in the future, and adjust these for gpt-5.5 soon once it becomes available on the API, but I feel like I've wasted enough time on this now. 😢 At least this should take us the right direciton I think.

https://linear.app/n8n/issue/INS-112/builder-doesnt-know-what-the-current-up-to-date-models-are

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/instance-ai/src/tools/nodes.tool.ts
```diff
@@ -344,15 +348,27 @@ export function createNodesTool(
...
-				'with your credentials.',
+				"Read node type definitions or query real resources for a node's RLC parameters " +
+				'(e.g. list Google Sheets, OpenAI models, Slack channels). Use `type-definition` ' +
+				'first to read `@searchListMethod` / `@loadOptionsMethod` annotations, then ' +
+				'`explore-resources` with the real method name and a credential.',
 			inputSchema: orchestratorInputSchema,
-			execute: async (input: z.infer<typeof orchestratorInputSchema>) => {
-				return await handleExploreResources(context, input);
+			execute: async (input: OrchestratorInput) => {
+				switch (input.action) {
```
Cadiac, line 367: These type definitions are needed by the orchestrator when it does things without the builder, like just updates a workflow with the update-file tool instead of going through the builder - it should understand how the nodes work in that case too.

Annoyingly it still sometimes skips checking these.

packages/@n8n/nodes-langchain/nodes/llms/LMChatAnthropic/LmChatAnthropic.node.ts
```diff
@@ -19,6 +19,11 @@ import {
 
 import { searchModels } from './methods/searchModels';
```
Cadiac, line 21: I went back and forth with these a lot, but considering MCP and the old ai builder also use these I ended up not making these too specific - for instance we shouldn't mention tools here by name, as each of these points checks these builderHints but works differently.

aalises (dismissed):

OlegIvaniv (approved):
This is great! IMO the manual maintenance of updating model references is a worthwhile tradeoff.

PR #29047, opened 2026-04-24
feat(agents): Polish agent builder UI and tighten builder/chat reliability

bjorger:
## Summary

Redesigned the agent page from:
image

To:
image

### UI / UX
- Three-column builder layout with config tree, section editor, and chat — replacing the previous two-pane shell.
- Chat session picker dropdown with relative timestamps, truncated labels, and an inline new-chat button.
- Session id persisted in the URL so refreshes and shared links land on the same chat; Test tab opens automatically when a session is in the URL.
- Unpublished-changes prompt now also fires when switching agents in-place via the header switcher.
- Trigger panel and add-trigger modal share live integration status, so connecting in one surface reflects immediately in the other.
- Test chat quick-actions row replaces the Add tool / trigger buttons with new-chat + history; new-chat icon hides until the session has messages.
- Config panels disable while the builder chat is streaming, with shared scrollbar + disabled-overlay styles.
- Hardcoded UI strings wrapped in i18n; dead translation keys dropped.

### Reliability fixes
- Build progress overlay now consumes the typed `AgentSseEvent` stream (was reading a legacy shape and dropping every event).
- Autosave snapshots `{projectId, agentId, config}` at schedule-time and gates response writes on the still-current agent, preventing cross-agent state bleed on quick switches.
- `useAgentConfig` and `agentSessions` store both token-guard their fetches so older responses can't overwrite the newer agent's data.
- Optimistic resume of interactive tool cards rolls back on transport or SSE error, so a failed resume no longer leaves the card permanently disabled.
- Stale `continueSessionId` deep links rebind to a real session instead of leaving the Test pane blank.
- `AgentMiniEditor`'s CodeMirror readOnly toggle made reactive via Compartment.

### Refactors (no behavior change)
- Five composables extracted from `AgentBuilderView` (layout, session, chat-mode, autosave, section-nav).
- Three CodeMirror editor instances collapsed into a single `useCodeMirrorEditor`.
- Handrolled buttons / banners / tools-tree / switch replaced with `@n8n/design-system` primitives.
- Model-string parsing centralised; magic strings extracted to typed constants.

https://linear.app/n8n/issue/AGENT-32/feature-redesign-agent-builder-with-three-column-config-centric-shell
https://linear.app/n8n/issue/AGENT-33/feature-derive-config-tree-and-wire-per-section-editing-in-agent
https://linear.app/n8n/issue/AGENT-34/feature-add-collapsible-blocks-to-agent-config-section-editor
https://linear.app/n8n/issue/AGENT-35/feature-add-section-aware-semantic-highlighting-to-agent-config-editor

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/agents/builder/interactive/ask-credential.tool.ts
```diff
@@ -27,10 +32,16 @@ export function buildAskCredentialTool(): BuiltTool {
 					input: AskCredentialInput,
 					ctx: InterruptibleToolContext<AskCredentialInput, AskCredentialResume>,
 				) => {
-					if (ctx.resumeData === undefined) {
-						return await ctx.suspend(input);
+					if (ctx.resumeData !== undefined) return ctx.resumeData;
+					// If the user has exactly one credential of the requested type the
+					// picker has nothing to ask — auto-resolve so the LLM doesn't render
+					// a card the user can only confirm.
+					const all = await deps.credentialProvider.list();
```
yehorkardash, line 39: Not sure if that's the best ux, we still want user to understand which credentials are being used and optionally be able to create new creds

packages/cli/src/modules/agents/builder/interactive/ask-llm.tool.ts
```diff
@@ -8,24 +8,45 @@ import {
...
-				if (ctx.resumeData === undefined) {
-					return await ctx.suspend(input);
+				if (ctx.resumeData !== undefined) return ctx.resumeData;
+				const all = await deps.credentialProvider.list();
+				// Keep only credentials whose type maps to a known LLM provider.
+				const llmCreds = all.filter((c) => LLM_PROVIDER_DEFAULTS[c.type]);
+				// Auto-resolve only when there's exactly one — if the user has
+				// multiple LLM credentials (or two of the same type), the choice is
+				// real and the card must render.
+				if (llmCreds.length === 1) {
```
yehorkardash, line 39: same here, user might want to create new creds and the first found might not be suitable

packages/cli/src/modules/agents/builder/builder-tool-names.ts
```diff
@@ -0,0 +1,17 @@
...
+	WRITE_CONFIG: 'write_config',
+	PATCH_CONFIG: 'patch_config',
+	BUILD_CUSTOM_TOOL: 'build_custom_tool',
+} as const;
+
+export type BuilderToolName = (typeof BUILDER_TOOLS)[keyof typeof BUILDER_TOOLS];
+
+/** Thread-id prefixes scoping different chat surfaces of the same agent. */
+export const AGENT_THREAD_PREFIX = {
+	TEST: 'test-',
```
yehorkardash, line 15: should we use same `:` format?

packages/cli/src/modules/agents/builder/interactive/llm-provider-defaults.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Azure variants), omit the entry so the tool falls through to suspending and
+ * lets the user pick explicitly.
+ */
+export interface LlmProviderDefault {
+	provider: string;
+	defaultModel: string;
+}
+
+export const LLM_PROVIDER_DEFAULTS: Record<string, LlmProviderDefault> = {
+	anthropicApi: { provider: 'anthropic', defaultModel: 'claude-sonnet-4-6' },
```
mike12345567, line 20: Some of these models are quite out of date, my suggestion for some defaults would be:
```
export const LLM_PROVIDER_DEFAULTS: Record<string, LlmProviderDefault> = {
  anthropicApi: { provider: 'anthropic', defaultModel: 'claude-sonnet-4-6' },
  openAiApi: { provider: 'openai', defaultModel: 'gpt-5' },
  googlePalmApi: { provider: 'google', defaultModel: 'gemini-2.5-pro' },
  xAiApi: { provider: 'xai', defaultModel: 'grok-4' },
  groqApi: { provider: 'groq', defaultModel: 'llama-3.1-70b-versatile' },
  mistralCloudApi: { provider: 'mistral', defaultModel: 'mistral-large-latest' },
  deepSeekApi: { provider: 'deepseek', defaultModel: 'deepseek-chat' },
  cohereApi: { provider: 'cohere', defaultModel: 'command-r-plus' },
};
```
Not for this PR necessarily but something we need to remember to do we - we should add support for OpenRouter https://openrouter.ai/docs/guides/community/vercel-ai-sdk

packages/cli/src/modules/agents/integrations/platforms/linear-integration.ts
```diff
@@ -24,6 +24,16 @@ export class LinearIntegration extends AgentChatIntegration {
 
 	readonly credentialTypes = ['linearApi', 'linearOAuth2Api'];
 
+	readonly displayLabel = 'Linear';
+
+	readonly displayIcon = 'list-checks';
+
+	readonly displayHelpText =
```
mike12345567: How are all of these strings surfaced? Just wondering about i18n, if these are displayed in the UI could be an issue.

packages/frontend/editor-ui/src/features/agents/composables/useAgentConfigAutosave.ts
```diff
@@ -0,0 +1,99 @@
...
+				// otherwise leave the newer promise in place so `settleAutosave`
+				// awaits the full tail of pending work.
+				if (autosaveInFlight === slot) autosaveInFlight = null;
+			});
+		}, getDebounceTime(debounceMs));
+	}
+
+	async function settleAutosave() {
+		if (autosaveTimer !== null) {
+			clearTimeout(autosaveTimer);
```
mike12345567, line 92: Should we clear `autoSaveInFlight` when clearing the timeout?

packages/cli/src/modules/agents/builder/interactive/llm-provider-defaults.ts
bjorger: I'm going to create a linear ticket for that

packages/frontend/editor-ui/src/features/agents/composables/useAgentConfigAutosave.ts
bjorger: nope - not necessary here. that would break the actual functionality of the `settleAutosave`.

The idea of the `settleAutosave` is to make sure that the user does not navigate away while a save is mid-flight. If `autosaveInFlight` is cleared here then the await `autosaveInFlight` on line 95 becomes a no-op against null, the route-leave guard returns immediately, and the in-flight POST keeps racing the navigation

mike12345567 (approved):
LGTM - thanks for the changes!

PR #29048, opened 2026-04-24
fix(core): Pin sandbox workflow-sdk to host version

OlegIvaniv:
## Summary

The Instance AI sandbox declared `@n8n/workflow-sdk: '*'` in its `PACKAGE_JSON`, which resolves to the npm `latest` dist-tag. That tag sits several versions behind what the CLI host is built against, so every remote sandbox has been running an older SDK than the code orchestrating it. Sandbox images are content-addressed by Dockerfile bytes, so floating `*` never invalidates the image cache on release.

This PR pins the version and adds a dev-only escape hatch for iterating on SDK changes end-to-end.

## Changes

**Pin sandbox deps to host-installed versions** (`sandbox-setup.ts`)

`resolveHostDepVersion(name)` reads `<host>/node_modules/<name>/package.json` at module load and returns the installed version. `PACKAGE_JSON` uses these pins for `@n8n/workflow-sdk` and `tsx` instead of `*`. On a CLI release where the SDK bumps, `PACKAGE_JSON` bytes change → Dockerfile bytes change → sandbox images invalidate.

**Opt-in workspace-SDK link for local dev** (new `pack-workspace-sdk.ts`, wired in `builder-sandbox-factory.ts`)

Gated by `N8N_INSTANCE_AI_SANDBOX_LINK_SDK=1`. When set:

- `pnpm pack` the host's workspace SDK (not `npm pack` — pnpm rewrites `workspace:*` / `catalog:` specifiers).
- Upload the tarball to the sandbox via `workspace.filesystem.writeFile`.
- `npm install <tarball> --no-save --ignore-scripts --force` to replace the registry copy.

Fails fast with a clear log if `<sdk>/dist` is missing so the dev knows to run `pnpm build`. No-op when unset — production sandboxes are unaffected.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
+++ b/packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
@@ -11,6 +11,7 @@ import { Daytona } from '@daytonaio/sdk';
 import { Workspace, LocalFilesystem, LocalSandbox } from '@mastra/core/workspace';
 import { DaytonaSandbox } from '@mastra/daytona';
 import assert from 'node:assert/strict';
+import { join as posixJoin } from 'node:path/posix';
 
 import type { Logger } from '../logger';
 import type { SandboxConfig } from './create-workspace';
@@ -86,7 +87,7 @@ export class BuilderSandboxFactory {
 		const packed = await this.sdkTarballPromise;
 		if (!packed) return;
 
-		const remotePath = `${root}/${packed.filename}`;
+		const remotePath = posixJoin(root, packed.filename);
 		if (workspace.filesystem) {
 			await workspace.filesystem.writeFile(remotePath, packed.tarball);
 		} else {
--- a/packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
+++ b/packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
@@ -12,15 +12,24 @@ import { Workspace, LocalFilesystem, LocalSandbox } from '@mastra/core/workspace
 import { DaytonaSandbox } from '@mastra/daytona';
 import assert from 'node:assert/strict';
 
+import type { Logger } from '../logger';
 import type { SandboxConfig } from './create-workspace';
 import { DaytonaFilesystem } from './daytona-filesystem';
 import { N8nSandboxFilesystem } from './n8n-sandbox-filesystem';
 import { N8nSandboxImageManager } from './n8n-sandbox-image-manager';
 import { N8nSandboxServiceSandbox } from './n8n-sandbox-sandbox';
-import { writeFileViaSandbox } from './sandbox-fs';
+import { packWorkspaceSdk, type WorkspaceSdkTarball } from './pack-workspace-sdk';
+import { runInSandbox, writeFileViaSandbox } from './sandbox-fs';
 import type { SnapshotManager } from './snapshot-manager';
 import type { InstanceAiContext } from '../types';
 import { formatNodeCatalogLine, getWorkspaceRoot, setupSandboxWorkspace } from './sandbox-setup';
+
+const NOOP_LOGGER: Logger = {
+	info: () => {},
+	warn: () => {},
+	error: () => {},
+	debug: () => {},
+};
 
 export interface BuilderWorkspace {
 	workspace: Workspace;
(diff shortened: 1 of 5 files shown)


Cadiac:
And the idea was to keep `N8N_INSTANCE_AI_SANDBOX_LINK_SDK` unset in production?

packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
```diff
@@ -61,8 +70,48 @@ export class BuilderSandboxFactory {
 	constructor(
 		private readonly config: SandboxConfig,
 		private readonly imageManager?: SnapshotManager,
+		private readonly logger: Logger = NOOP_LOGGER,
 	) {}
 
+	/** Cached workspace-SDK tarball promise (one pack per process). */
+	private sdkTarballPromise: Promise<WorkspaceSdkTarball | null> | null = null;
```
Cadiac, line 78: I guess keeping this in memory might consume a fair amount of it - but if this mechanism is only to be used in dev that shouldn't matter?

OlegIvaniv: Yes, this is for local development only when you need to have latest build of workflow SDK

OlegIvaniv:
> And the idea was to keep `N8N_INSTANCE_AI_SANDBOX_LINK_SDK` unset in production?

Yes. In production it should still use npm registry version, just making sure they're synced.

packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
```diff
@@ -61,8 +70,48 @@ export class BuilderSandboxFactory {
...
+	 * Pack and install the host's workspace `@n8n/workflow-sdk` into the remote
+	 * sandbox, overriding the registry version baked in at image build time.
+	 * No-op unless `N8N_INSTANCE_AI_SANDBOX_LINK_SDK=1` is set.
+	 */
+	private async linkWorkspaceSdkIfEnabled(workspace: Workspace, root: string): Promise<void> {
+		this.sdkTarballPromise ??= packWorkspaceSdk(this.logger);
+		const packed = await this.sdkTarballPromise;
+		if (!packed) return;
+
+		const remotePath = `${root}/${packed.filename}`;
```
Cadiac: this should probably use path.join for consistency, even though our windows support here is questionable anyways..

packages/@n8n/instance-ai/src/workspace/builder-sandbox-factory.ts
```diff
@@ -61,8 +70,48 @@ export class BuilderSandboxFactory {
...
+		if (workspace.filesystem) {
+			await workspace.filesystem.writeFile(remotePath, packed.tarball);
+		} else {
+			await writeFileViaSandbox(workspace, remotePath, packed.tarball);
+		}
+
+		const install = await runInSandbox(
+			workspace,
+			`npm install ${remotePath} --no-save --ignore-scripts --force`,
```
Cadiac, line 99: This is a bit sketchy, if someone found a way to call this function with malicious `root` parameter this would make it possible to run arbitrary commands on the sandbox, but since it's the sandbox I don't think we really care? (and if we switch the code above to path.join I guess it will also prevent some abuse)

packages/@n8n/instance-ai/src/workspace/sandbox-setup.ts
```diff
@@ -35,21 +38,87 @@ export const N8N_SANDBOX_HOME = '/home/user';
...
+	return JSON.stringify(
+		{
+			name: 'sandbox-workspace',
+			private: true,
+			dependencies: {
+				'@n8n/workflow-sdk': sdkSpecifier,
+				tsx: SANDBOX_TSX_VERSION,
+			},
+			devDependencies: {
+				'@types/node': '*',
```
Cadiac: should we just pin this to something hard coded for now?

packages/@n8n/instance-ai/src/workspace/sandbox-fs.ts
```diff
@@ -56,7 +56,10 @@ export async function writeFileViaSandbox(
 	}
 
 	// Encode content as base64 and decode in the sandbox
-	const b64 = Buffer.from(content, 'utf-8').toString('base64');
+	const b64 =
+		typeof content === 'string'
+			? Buffer.from(content, 'utf-8').toString('base64')
+			: content.toString('base64');
 	const cmd = `echo '${b64}' | base64 -d > '${escapeSingleQuotes(filePath)}'`;
```
Cadiac, line 63: out of scope of this PR, but I wonder if this mechanism really is the best way we can write files into the sandbox 😬

Cadiac (commented):
Nice find, left some minor comments - maybe fix the @types/node version to something specific and switch to path.join on builder-sandbox-factory.ts, but I'm happy with this otherwise!

Cadiac (approved):

PR #29051, opened 2026-04-24
fix(Discord Node): Correctly handle API rate limits

BerniWittmann:
## Summary

Discord's `x-ratelimit-reset-after` header is in seconds, but the value was passed directly to `sleep()` which expects milliseconds — so the client barely waited at all when a per-route bucket was exhausted and was likely to hit further limits. This PR converts the header from seconds to milliseconds, caps the wait at 60s, and adds a 429-aware retry loop in `requestApi` that honors `Retry-After` (with a 1s fallback and up to 3 retries). Rate-limit handling is extracted into a dedicated `handleRateLimitHeaders` helper and covered by new unit tests for both success-path header handling and the retry path.

**How to test:** Run the new unit tests:

```bash
cd packages/nodes-base && pnpm test discord.api.test
```

https://linear.app/n8n/issue/NODE-4839

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/nodes-base/nodes/Discord/v2/transport/helpers.ts
+++ b/packages/nodes-base/nodes/Discord/v2/transport/helpers.ts
@@ -6,6 +6,12 @@ import type {
 	ILoadOptionsFunctions,
 	IRequestOptions,
 } from 'n8n-workflow';
+import { sleep } from 'n8n-workflow';
+
+const MAX_RATE_LIMIT_RETRIES = 3;
+const MAX_RATE_LIMIT_SLEEP_MS = 60_000;
+const FALLBACK_RETRY_AFTER_MS = 1_000;
+const GLOBAL_RATE_LIMIT_SLEEP_MS = 20;
 
 export const getCredentialsType = (authentication: string) => {
 	let credentialType = '';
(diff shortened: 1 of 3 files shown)


packages/nodes-base/nodes/Discord/v2/transport/helpers.ts
```diff
@@ -25,22 +30,62 @@ export const getCredentialsType = (authentication: string) => {
 	return credentialType;
 };
 
+const secondsToMs = (value: unknown): number => {
+	const seconds = Number(value);
+	if (!Number.isFinite(seconds) || seconds <= 0) return 0;
+	return Math.min(seconds * 1000, MAX_RATE_LIMIT_SLEEP_MS);
+};
+
+export async function handleRateLimitHeaders(headers: IDataObject | undefined) {
```
dlavrenuek, line 40: The old implementation [has a hardcoded 20ms wait](https://github.com/n8n-io/n8n/pull/29051/changes#diff-0383912767db3e611d41b2d43b8a8db30ce23a0931b664c24266737f083a8193L93) to not exceed 50 requests per second. Although it's not reliable since multiple parallel executions do not track how many requests were actually sent I also don't see an obvious reason to remove it, or?

> 			await sleep(20); //prevent exceeding global rate limit of 50 requests per second

packages/nodes-base/nodes/Discord/v2/transport/helpers.ts
```diff
@@ -25,22 +30,62 @@ export const getCredentialsType = (authentication: string) => {
...
+	if (!headers) return;
+	const remaining = Number(headers['x-ratelimit-remaining']);
+	if (remaining !== 0) return;
+	const waitMs = secondsToMs(headers['x-ratelimit-reset-after']);
+	if (waitMs > 0) {
+		await sleep(waitMs);
+	}
```
dlavrenuek: The logic is a bit hard to understand due to multiple conditions. It's more or less:

> If headers are empty do nothing, else if remaining header not equal 0 do nothing, else if wait header greater 0 wait.

In general I find "positive" conditions easier to understand than "negated", but this might be my own preference. What do you think about something like


```suggestion
	if (Number(headers?.['x-ratelimit-remaining']) !== 0) {
		await sleep(secondsToMs(headers?.['x-ratelimit-reset-after']));
	}
```

Technically the difference with your implementation would be that if `x-ratelimit-remaining` header is set and > 0 but `x-ratelimit-reset-after` is 0 the execution will be moved to next node engine execution tick, but it should be irrelevant for performance anyway and this case should also not happen because it makes no sense

packages/nodes-base/nodes/Discord/v2/transport/helpers.ts
BerniWittmann: readded it, which also simplified the logic

dlavrenuek (approved):
Looks good, the addition of 429 error handling is also a nice improvement 🚀

PR #29056, opened 2026-04-24
fix(editor): Keep opening executions when credential prefetch fails

fuyua9:
## Summary

- keep `openExecution` flowing when workflow credential prefetch rejects
- still set the execution document state, update node issues, stop canvas loading, fit the view, and emit `open:execution`
- add a regression test for a rejected `fetchAllCredentialsForWorkflow()` call

## Tests

- `git diff --check`
- Attempted: `pnpm --filter n8n-editor-ui test -- usePostMessageHandler.test.ts --runInBand` (local checkout has no installed `node_modules`, so `vitest` is unavailable)

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/usePostMessageHandler.test.ts
+++ b/packages/frontend/editor-ui/src/app/composables/usePostMessageHandler.test.ts
@@ -380,7 +380,9 @@ describe('usePostMessageHandler', () => {
 				mode: 'trigger',
 				finished: true,
 			});
-			mockFetchAllCredentialsForWorkflow.mockRejectedValueOnce(new Error('credential prefetch failed'));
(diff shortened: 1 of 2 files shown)


fuyua9:
I pushed `5f3e0ee` to address the failing checks:

- Ran Biome formatting on the touched editor files and fixed the long test line.
- Added a comment to the best-effort credential prefetch `catch` block instead of leaving it empty.
- Updated the PR title to follow the n8n title convention with a capitalized imperative summary.

Local validation:

```bash
npx --yes @biomejs/biome@1.9.4 format --write packages/frontend/editor-ui/src/app/composables/usePostMessageHandler.ts packages/frontend/editor-ui/src/app/composables/usePostMessageHandler.test.ts
# Formatted 2 files; fixed 1 file before the commit
```

PR #29059, opened 2026-04-24
fix(instance-ai): scope submit workflow identity by project

fuyua9:
## Summary

- include `projectId` in the submit-workflow identity key, alongside the resolved workflow `filePath`
- keep same-file/same-project submissions deduped into updates
- add a regression test showing same file path submissions to different projects get independent workflow IDs

## Tests

- `git diff --check`
- Attempted: `pnpm --filter @n8n/instance-ai test -- submit-workflow-identity.test.ts --runInBand` (local checkout has no installed `node_modules`, so `jest` is unavailable)

### DIFF
--- a/packages/@n8n/instance-ai/src/tools/workflows/__tests__/submit-workflow-identity.test.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/__tests__/submit-workflow-identity.test.ts
@@ -152,6 +152,27 @@ describe('wrapSubmitExecuteWithIdentity', () => {
 		expect(calls[2].workflowId).toBe('wf_1');
 	});
 
+	it('treats omitted projectId and its resolved default projectId as the same identity', async () => {
+		const { execute, calls } = makeUnderlying();
+		const wrapped = wrapSubmitExecuteWithIdentity(
+			execute,
+			resolvePath,
+			(projectId) => projectId ?? 'personal-project',
+		);
+
+		const implicitPersonal = await wrapped({ filePath: MAIN_PATH });
+		const explicitPersonal = await wrapped({
+			filePath: MAIN_PATH,
+			projectId: 'personal-project',
+		});
+
+		expect(implicitPersonal.workflowId).toBe('wf_1');
+		expect(explicitPersonal.workflowId).toBe('wf_1');
+		expect(calls).toHaveLength(2);
+		expect(calls[0].workflowId).toBeUndefined();
+		expect(calls[1].workflowId).toBe('wf_1');
+	});
+
 	it('resolves differently-spelled paths to the same identity', async () => {
 		const { execute } = makeUnderlying();
 		const wrapped = wrapSubmitExecuteWithIdentity(execute, resolvePath);
--- a/packages/@n8n/instance-ai/src/tools/workflows/__tests__/submit-workflow-identity.test.ts
+++ b/packages/@n8n/instance-ai/src/tools/workflows/__tests__/submit-workflow-identity.test.ts
@@ -134,6 +134,24 @@ describe('wrapSubmitExecuteWithIdentity', () => {
 		expect(chunkAgain.workflowId).toBe(chunkResult.workflowId);
 	});
 
+	it('does not reuse workflow identity across different projectIds for the same filePath', async () => {
+		const { execute, calls } = makeUnderlying();
+		const wrapped = wrapSubmitExecuteWithIdentity(execute, resolvePath);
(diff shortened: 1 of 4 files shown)


fuyua9:
Thanks for the review. I pushed `99c3eb3` to normalize the identity key through the same project resolution used by workflow creation:

- `wrapSubmitExecuteWithIdentity()` now accepts an async project-id resolver and keys pending workflow identity on the resolved project id.
- The real adapter exposes `resolveCreateProjectId()` so omitted `projectId` and the user's explicit personal project id collapse to the same key.
- Added a regression test for omitted vs explicit default project id.

Local validation:

```bash
git diff --check
# passed
```

I also attempted the focused Jest test (`pnpm --filter @n8n/instance-ai test -- submit-workflow-identity.test.ts --runInBand`), but this local worktree does not have `node_modules` installed, so `jest` was unavailable. The pushed commit should let CI run the full project validation.

fuyua9:
Thanks — I refreshed this branch against the current `master` and pushed `fd562192`.

What changed in this follow-up:
- Kept the submit-workflow identity guard scoped by resolved `projectId` + file path.
- Preserved the newer upstream submit guard / remediation budget behavior while resolving the merge conflicts.
- Kept the PR scoped to the same four Instance AI files.

Validation / status:
- `git diff --check` — passed locally.
- Conflict-marker scan on the four changed files — clean locally.
- Focused test command could not run in this fresh checkout because `node_modules` is not installed (`jest: command not found`).
- GitHub now reports the PR branch as `MERGEABLE`; checks are running.

fuyua9:
Updated this branch to clear the current merge conflict with latest `master`.

What changed:
- rebased the submit-workflow project identity fix onto current upstream `master`
- kept the upstream pre-save submit budget/remediation guard behavior
- preserved project-aware identity dedupe by using the resolved project id in the pending-map key
- kept failure cleanup on the project-aware `identityKey`

Validation:
- `git diff --check upstream/master...HEAD` passed
- conflict marker scan on touched files passed
- `pnpm --filter @n8n/instance-ai test -- submit-workflow-identity.test.ts` was attempted, but this checkout has no `node_modules`; local `jest` is unavailable

Main alignment: current head is based on upstream `master` and preserves the newer remediation guard path.

fuyua9:
Thanks for the follow-up review. Cubic identified two remaining issues, and this update addresses both:

- project resolution failures now return the structured blocked remediation instead of escaping before submit error handling;
- updates with an explicit `workflowId` no longer require `workflow:create` project access.

Added regression coverage for both paths. Local validation: `git diff --check` passed. The focused Jest command was attempted, but this checkout has no `node_modules`, so `jest` is unavailable locally; CI is now running on `cc4e7ce535b`.

suyua9:
Follow-up pushed in `2809372a303`.

Cubic identified the current-head issue: when `workflowId` is supplied without `projectId`, the wrapper used the literal `personal` key while create-path resolution uses the personal project UUID. That could bypass identity deduplication and create a duplicate workflow.

Fixed by resolving the existing workflow's owning project through the workflow service and using that canonical project id in the identity key, without requiring `workflow:create`. Added a regression test covering an update that omits `projectId` and shares the key with the resolved default project.

Validation:
- `git diff --check` passed.
- Focused Jest test attempted but this checkout has no `node_modules` (`jest: command not found`).
- Package `typecheck` attempted but dependencies/shared TypeScript configs are absent in this checkout.

Source attribution: this follow-up addresses the issue identified by Cubic.

suyua9:
Follow-up for the current Cubic review finding (identified by Cubic):

- The optional `resolveWorkflowProjectId` callback is now installed only when the workflow service actually provides it.
- Service implementations without that optional method keep the existing `personal` fallback for explicit workflow updates without `projectId`, instead of returning a project-resolution blocker.
- Added a regression test for that no-resolver path.

Pushed as `ed14f2779f2bc62692050f9856a362efe15f5d3e`.

Validation: `git diff --check` passed. The targeted Jest test was attempted in an isolated checkout but could not run because the checkout has no `node_modules` (`jest: command not found`).

Cadiac:
Since this PR was opened it has grown very stale - it targeted a very early version of AI assistant, and its implementation doesn't make much sense today, AI Assistant has since become projects aware and does project scoping already. `submit-workflow-tool.ts` and the original `submit-workflow-identity.ts` have been removed back in June.

Closing thi PR as stale - if you find issues regarding project scoping in today's implementation please raise issues about them.

PR #29066, opened 2026-04-24
fix(ai-builder): Hide and reap intermediate AI-created workflows

aalises:
## Summary

The AI builder leaves intermediate workflows behind in two scenarios ([INS-24](https://linear.app/n8n/issue/INS-24)):

1. **Orphans from failed builds** — a task partially saves a workflow then fails or is cancelled.
2. **Stepping-stones** — chunks, sub-workflows, scratch attempts, or workflows obsoleted by a replan that the agent never references again.

Both surface in the user's workflows list as half-built clutter the user didn't ask for.

## How it works

- `meta.aiTemporary: true` is stamped on every workflow the AI builder creates (`submit-workflow` + `build-workflow`, only on create — never on update so user-authored workflows are untouched). No DB migration: `meta` is already a nullable JSON column.
- The default workflows list query hides stamped rows (postgres + sqlite). The user never sees an in-progress stepping-stone.
- The orchestrator clears the stamp on `mainWorkflowAttempt.workflowId` at the end of every successful build — the deliverable is now indistinguishable from a user workflow.
- At every run-finish (success / cancel / error), the platform iterates the run's in-memory created-set and calls `archiveIfAiTemporary` on each id. Atomic check-and-archive: anything still stamped (chunks, scratch, obsoleted-by-replan) gets soft-deleted; the main survives because its stamp was cleared.

## Artifacts panel UX

The artifacts panel keeps stepping-stones visible in the conversation history after the run ends — but they've been archived, so clicking them would lead to a dead preview. To close that loop:

- `run-finish` events carry `archivedWorkflowIds: string[]` — the ids the reap just soft-deleted.
- The frontend store accumulates them per thread and tags matching entries in the resource registry as `archived: true`.
- The artifacts panel rows, and the `ArtifactCard` used by the activity tree and timeline, render archived entries dimmed with a strikethrough name and an "Archived" pill.

## Why this shape

- Resume paths, crashes, abandoned sessions, direct `build-workflow-with-agent` calls — every code path that could leak a stepping-stone reduces to a single SQL predicate.
- The agent's prompt carries no cleanup instructions — the platform owns the lifecycle, not the model.
- Soft delete (archive) — a mistaken reap is recoverable from the archive view.

Crash leaks (process dies mid-run before run-finish fires) are accepted as known: invisible to the user (list filter) and rare. Easy to add a one-off admin sweep later if it ever materializes.

https://linear.app/n8n/issue/INS-24

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/instance-ai/src/tools/workflows/build-workflow.tool.ts
```diff
@@ -181,6 +193,8 @@ export function createBuildWorkflowTool(context: InstanceAiContext) {
 					};
 				} else {
 					const created = await context.workflowService.createFromWorkflowJSON(json, opts);
+					(context.aiCreatedWorkflowIds ??= new Set<string>()).add(created.id);
+					await options.onWorkflowCreated?.(created.id);
```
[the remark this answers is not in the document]
aalises: Fixed at 025ee6e...HEAD in build-workflow.tool.ts: the post-create `onWorkflowCreated` callback is now wrapped in its own try/catch. A callback failure no longer flips the create into a `success: false` response (which would have risked a duplicate on retry); the failure is logged via `context.logger.warn` instead of being silently swallowed.

packages/@n8n/instance-ai/src/types.ts
```diff
@@ -573,6 +584,14 @@ export interface PlannedTaskRecord extends PlannedTask {
 	outcome?: Record<string, unknown>;
 	startedAt?: number;
 	finishedAt?: number;
+	/**
+	 * Workflow IDs the task saved while running. Written every time
+	 * `submit-workflow` reports a successful save; read at run-end so the
+	 * service can hard-delete the ones that belong to failed or cancelled
```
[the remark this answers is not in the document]
aalises: Fixed: updated the `createdWorkflowIds` docstring in types.ts to say "archive (soft-delete)" so it matches the archive/restore semantics used by the reap pipeline.

packages/@n8n/db/src/repositories/workflow.repository.ts
```diff
@@ -882,6 +882,26 @@ export class WorkflowRepository extends Repository<WorkflowEntity> {
...
+	}
+
+	/**
+	 * Hide workflows the AI builder created and hasn't yet promoted to the
+	 * main deliverable. The orchestrator clears the stamp on the main at
+	 * build-time and reaps the rest at run-finish — but in the window between
+	 * create and reap, stamped rows must not surface in the workflows list.
+	 */
+	private applyAiTemporaryFilter(qb: SelectQueryBuilder<WorkflowEntity>): void {
+		const dbType = this.globalConfig.database.type;
```
Cadiac: This is a lot more robust than what we had before, but I'm worried that this might be a pretty inefficient lookup? Should we just turn this into a real `isTemporary` field or similar on the workflow, instead of having to dig up data from this field - I fear that on large instances this could be slow?

Although it seems like `applyAvailableInMCPFilter` is also looking up data inside workflow.settings in similar fashion.

I'll ask someone from cats for opinion on us introducing more json lookups on these queries...

packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
```diff
@@ -323,6 +319,27 @@ export class InstanceAiAdapterService {
 				await workflowService.delete(user, workflowId);
 			},
 
+			async clearAiTemporary(workflowId: string) {
+				const workflow = await workflowFinderService.findWorkflowForUser(workflowId, user, [
```
Cadiac, line 333: this doesn't check `assertNotReadOnly` while `archiveIfAiTemporary` does - do we consider clearing the `aiTemporary` field something that can be done on read-only instances - probably not?

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -2537,21 +2511,59 @@ export class InstanceAiService {
...
+	 */
+	private async reapAiTemporaryFromRun(
+		threadId: string,
+		user: User,
+		createdWorkflowIds: Set<string> | undefined,
+	): Promise<string[]> {
+		if (!createdWorkflowIds || createdWorkflowIds.size === 0) return [];
+		const adapter = this.adapterService.createContext(user, { threadId });
+		const archived: string[] = [];
+		for (const workflowId of createdWorkflowIds) {
```
Cadiac: These could probably be done in `Promise.all` concurrently to make it slightly faster

packages/@n8n/db/src/repositories/workflow.repository.ts
tomi: Jaakko is right. This will end up being a sequential scan on the entire table, which is very slow on larger instances. So let's come up with some other approach to this one.

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAi.store.ts
```diff
@@ -349,6 +351,14 @@ export const useInstanceAiStore = defineStore('instanceAi', () => {
 					thread.title = parsed.data.payload.title;
 				}
 			}
+			if (parsed.data.type === 'run-finish') {
+				const ids = parsed.data.payload.archivedWorkflowIds;
+				if (ids && ids.length > 0) {
+					const next = new Set(archivedWorkflowIds.value);
+					for (const id of ids) next.add(id);
+					archivedWorkflowIds.value = next;
```
Cadiac, line 360: Is recreating this some vue reactivity hack or...?

packages/frontend/editor-ui/src/features/ai/instanceAi/useResourceRegistry.ts
```diff
@@ -12,6 +12,13 @@ export interface ResourceEntry {
 	createdAt?: string;
 	updatedAt?: string;
 	projectId?: string;
+	/**
+	 * Set to true when the run-finish reap archived this workflow — a
+	 * stepping-stone the agent created but never promoted to the main
+	 * deliverable. The artifacts panel renders these as dimmed with an
+	 * "Archived" label.
+	 */
+	archived?: boolean;
```
Cadiac, line 21: minor: I'm slightly worried of these getting out-of-sync, but I can't prove it happening - it feels a bit odd that we're tracking these through `archivedWorkflowIds` instead of just observing the real workflow's `isArchived` field 🤔 But I suppose we need to know the status here in context that doesn't have the up-to-date workflow definition available.

packages/@n8n/db/src/repositories/workflow.repository.ts
aalises: Thanks for the challenge!, I tried not to create a migration, but it's true that this approach will add a lot of overhead... What about a marker/ledge table? something like:

- create an `ai_builder_temporary_workflow` table with `workflowId` as PK/FK to `workflow_entity` (only has workflow ID, thread ID and the created_at, updated_at)
- insert a row when the AI builder creates an intermediate workflow
- hide intermediates with an indexed `NOT EXISTS` / anti-join instead of the heavy JSON extraction
- promote the final workflow by deleting its marker row
- reap leftovers by joining against that marker table and archiving those workflow IDs

@tomi @Cadiac

tomi: I think the simplest option would be to introduce a `isDraft` column (similar to `isArchived`) and use that

tomi: I think a marker row would also work. I think it boils down to whether we want to make this a more "core" feature part of the workflows themselves, or keep it AI builder specific. The `isDraft` is also a bit overloaded term since workflow publish uses that term for workflows that haven't been published yet. So I'm fine with marker table / row approach

packages/frontend/editor-ui/src/features/ai/instanceAi/instanceAi.store.ts
aalises: Yep, Set.add() on a ref-wrapped Set doesn't trigger reactivity in Vue 3 so the app won't re-evaluate unless we reassign with a new Set instance #hax

packages/cli/src/modules/instance-ai/instance-ai.service.ts
aalises: Fair enough, we could use `Promise.allSettled` instead as with the `all`, if one `archiveIfAiTemporary()` rejects, Promise.all rejects early and the whole reap path would fail, which breaks the current best-effort behavior.

I wouldn't expect the performance gains to be massive as reap per session would be tiny (in favour of some deterministic logging) but I can change it if you prefer.

packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
aalises: good catch! fixed

packages/@n8n/db/src/repositories/workflow.repository.ts
aalises: Agree @tomi , adding a field seems a bit loaded. Marker row that `AI builder currently owns cleanup for this workflow` vs. the field which implies that `temporary is a native workflow property.`, more weight.

I'll do the marker row approach then cc @Cadiac

packages/cli/src/modules/instance-ai/instance-ai.service.ts
Cadiac: this is not critical to address now I think

packages/frontend/editor-ui/src/features/ai/instanceAi/useResourceRegistry.ts
aalises: Keeping the Set for now.. I think deriving from the workflows store would probably be cleaner but AI-built workflows aren't guaranteed to be loaded there :'(. Worth a follow-up once that changes

aalises:
@tomi @Cadiac https://github.com/n8n-io/n8n/pull/29066/changes/b43b32b063f0ada1f7003ca761bb5d0567dc23f6 here's implementing the ledger table + backfill

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -14,7 +13,7 @@ import { Logger } from '@n8n/backend-common';
 import { GlobalConfig } from '@n8n/config';
 import { Time } from '@n8n/constants';
```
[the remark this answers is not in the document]
aalises: Done in 5f92ea910c — wrapped `findByThread` in try/catch so a marker-table read failure logs and continues instead of bubbling up. The downstream archive loop already handles its own errors, so the reap path is best-effort end-to-end now.

packages/@n8n/db/src/migrations/common/1777281990044-BackfillAiBuilderTemporaryWorkflows.ts
```diff
@@ -0,0 +1,76 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class BackfillAiBuilderTemporaryWorkflows1777281990044 implements ReversibleMigration {
```
tomi: I don't think we need this if the `meta` based marker hasn't been released to anyone yet

aalises: done in 068d9b9905. Dropped the backfill migration, the dual-read JSON predicate in `applyAiBuilderTemporaryFilter`, the `meta.aiTemporary` fallback in `clearAiTemporary` / `archiveIfAiTemporary`, and the `WorkflowFEMeta.aiTemporary` field. Sooo the Marker table is now the single source of truth.

packages/@n8n/db/src/migrations/common/1777281990043-CreateAiBuilderTemporaryWorkflowTable.ts
```diff
@@ -0,0 +1,18 @@
...
+export class CreateAiBuilderTemporaryWorkflowTable1777281990043 implements ReversibleMigration {
+	async up({ schemaBuilder: { createTable, column } }: MigrationContext) {
+		await createTable('ai_builder_temporary_workflow')
+			.withColumns(column('workflowId').varchar(36).primary.notNull, column('threadId').uuid)
+			.withForeignKey('workflowId', {
+				tableName: 'workflow_entity',
+				columnName: 'id',
+				onDelete: 'CASCADE',
+			})
+			.withIndexOn('threadId').withTimestamps;
```
tomi, line 20: Should the `threadId` also have a FK to `instance_ai_threads` table?

aalises: true, forgot about it, added!

tomi (dismissed):
Migrations LGTM 👌

Cadiac (dismissed):

tomi (approved):

PR #29071, opened 2026-04-24
feat(core): Add instance-level JWE key infrastructure (no-changelog)

guillaumejacquart:
## Summary

Foundational infrastructure for JWE (JSON Web Encryption) token decryption in OAuth2 flows. A new `jwe` backend module generates an instance-level RSA-OAEP-256 key pair on boot and persists it in the existing `deployment_key` table — private JWK encrypted at rest via `Cipher.encryptWithInstanceKey`, public JWK as plaintext. The key pair is served via `CacheService` so mains and workers share a single view under queue mode and multi-main, and the DB's partial unique indexes serialise concurrent generation attempts so losers re-read the winner's pair. Also adds pure decryption utilities (`isJweToken`, `decryptJweToken`, `decryptJweTokenData`) for downstream OAuth2 consumers.

**How to test:**
- Start n8n on a fresh DB → observe `Generated new instance JWE key pair` log entry (scope `jwe`) and one rows in `deployment_key` with type `jwe.private-key`.
- Restart n8n → no new generation; same `kid` is served.
- Run multiple mains concurrently against the same DB → only one pair persists.

https://linear.app/n8n/issue/IAM-476

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/modules/oauth-jwe/oauth-jwe-key.service.ts
```diff
@@ -0,0 +1,211 @@
...
+		const winner = await this.readActivePairData();
+		if (!winner) {
+			throw new UnexpectedError('OAuth JWE key pair not found after generation');
+		}
+		return winner;
+	}
+
+	private async readActivePairData(): Promise<OAuthJweKeyPairData | null> {
+		const [publicRow, privateRow] = await Promise.all([
+			this.deploymentKeyRepository.findActiveByType(JWE_PUBLIC_KEY_TYPE),
```
afitzek: I think we don't need to store the public key, because a public key can afaik always be calculated from the private key. So loading the private key in a crypto primitive, will allow you to extract the public key as well.

packages/@n8n/db/src/migrations/common/1777023444000-AddJweKeyIndexesToDeploymentKey.ts
```diff
@@ -0,0 +1,33 @@
...
+			'IDX_deployment_key_jwe_public_key_active',
+			"status = 'active' AND type = 'jwe.public-key'",
+		);
+
+		await createIndex(
+			'deployment_key',
+			['type'],
+			true,
+			'IDX_deployment_key_jwe_private_key_active',
+			"status = 'active' AND type = 'jwe.private-key'",
```
afitzek, line 14: Why do we need a unique constraint on the active status for this key type? I thought we are fine to have multiple keys of different algorithms for encryption?

packages/@n8n/db/src/migrations/common/1777023444000-AddJweKeyIndexesToDeploymentKey.ts
```diff
@@ -0,0 +1,33 @@
+import type { MigrationContext, ReversibleMigration } from '../migration-types';
+
+export class AddJweKeyIndexesToDeploymentKey1777023444000 implements ReversibleMigration {
+	async up({ schemaBuilder: { createIndex } }: MigrationContext) {
+		await createIndex(
+			'deployment_key',
+			['type'],
+			true,
+			'IDX_deployment_key_jwe_public_key_active',
+			"status = 'active' AND type = 'jwe.public-key'",
```
afitzek: I don't think we should have the public key.

packages/cli/src/modules/oauth-jwe/oauth-jwe-key.service.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+function isUniqueConstraintViolation(error: unknown): boolean {
+	if (!(error instanceof QueryFailedError)) return false;
+	const driverError = error.driverError as { code?: string };
+	const code = driverError?.code;
+	return (
+		code === '23505' || // postgres
+		code === 'SQLITE_CONSTRAINT_UNIQUE' ||
+		code === 'ER_DUP_ENTRY' // mysql / mariadb
```
afitzek: There is no mariadb or mysql anymore.

packages/cli/src/modules/oauth-jwe/oauth-jwe.constants.ts
```diff
@@ -0,0 +1,5 @@
+export const JWE_PUBLIC_KEY_TYPE = 'jwe.public-key';
+export const JWE_PRIVATE_KEY_TYPE = 'jwe.private-key';
+export const JWE_KEY_ALGORITHM = 'RSA-OAEP-256';
+export const JWE_KEY_USE = 'enc';
+export const JWE_KEY_CACHE_KEY = 'jwe:key-pair';
```
afitzek, line 12: This probably suggest that we also have just a single key pair.

packages/cli/src/modules/oauth-jwe/oauth-jwe-key.service.ts
guillaumejacquart: good point

packages/@n8n/db/src/migrations/common/1777023444000-AddJweKeyIndexesToDeploymentKey.ts
guillaumejacquart: hum good point. I added the algorithm as part of the unique index, what do you think ?

packages/cli/src/modules/oauth-jwe/oauth-jwe.constants.ts
guillaumejacquart: Yes, good point. I changed for 'JWE_KEY_ALGORITHM**S**' to enable multiple algorithms at some point (one for now)

packages/cli/src/modules/oauth-jwe/oauth-jwe-key.service.ts
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+}
+
+/**
+ * Strips the private-only fields from an RSA private JWK to produce its
+ * public counterpart.
+ */
+function toPublicJwk(privateJwk: JWK): JWK {
+	const { d, p, q, dp, dq, qi, ...publicJwk } = privateJwk;
```
afitzek: I think it would be safer to whitelist fields that we want to expose and make this whitelist algorithm specific.

afitzek (dismissed):
One small comment, but we can change this later as well.

packages/cli/src/modules/oauth-jwe/oauth-jwe-key.service.ts
guillaumejacquart: good point, I added whitelist fields extraction by algorithm

afitzek (approved):
Nice work, thanks for addressing all comments. 🙏

tomi (approved):
Migration LGTM 👍

PR #29076, opened 2026-04-24
fix(editor): Align Model and Sandbox visibility with edit permissions

aalises:
## Summary

In Settings → AI Assistant, the Model and Sandbox cards were visible and appeared editable on cloud-managed instances, but every save was rejected by the backend with `Cannot update cloud-managed fields: …`. The UI visibility gate only checked `isProxyEnabled`, while the server's edit rejection also covers `isCloudManaged` (see `CLOUD_MANAGED_ADMIN_FIELDS` / `CLOUD_MANAGED_PREFERENCE_FIELDS` in `instance-ai-settings.service.ts`).

This aligns the visibility gate for both sections with `MemorySection` / `AdvancedSection`, which already check both flags. After the fix, the sections render in the exact modes where the server will actually accept edits:

| Mode | `isProxyEnabled` | `isCloudManaged` | Model / Sandbox cards |
|---|---|---|---|
| Self-hosted | false | false | Shown, editable |
| Self-hosted + proxy | true | false | Hidden |
| n8n Cloud | false | true | Hidden *(was incorrectly shown)* |
| Cloud + proxy | true | true | Hidden |

https://linear.app/n8n/issue/INS-118

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/SettingsInstanceAiView.test.ts
+++ b/packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/SettingsInstanceAiView.test.ts
@@ -176,6 +176,36 @@ describe('SettingsInstanceAiView', () => {
 		});
 	});
 
+	describe('section visibility — cloud managed (proxy disabled)', () => {
+		beforeEach(() => {
+			setModuleSettings(settingsStore, {
+				...defaultModuleSettings,
+				proxyEnabled: false,
+				cloudManaged: true,
+			});
+		});
+
+		it('hides Model section', () => {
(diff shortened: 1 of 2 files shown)


Cadiac (approved):
Like mentioned on slack I don't think this was a real issue on current cloud (the proxyEnabled should already hide these sections, at least I don't see these on my cloud instance) - do we really need this? We can merge and undo this later, but I just don't think the bug report is accurate.

Backend has concepts
`PROXY_MANAGED_PREFERENCE_FIELDS` and `CLOUD_MANAGED_PREFERENCE_FIELDS`, they both have the same value.

PR #29087, opened 2026-04-24
fix(core): Surface credential and resource choice when more than one match exists

aalises:
## Summary

Instance AI silently picked the first credential and guessed list-based resource IDs when more than one candidate was available. Users observed this as "it just assumed the first calendar for the credential" and had no way to correct the choice.

- `setup-workflow.service`: only auto-apply a credential when exactly one candidate exists. With multiple, leave the node's credential unset and rely on `needsAction` + `existingCredentials` so the setup wizard lets the user pick.
- Orchestrator (`system-prompt`) and planner (`plan-agent-prompt`): instructed to `ask-user` once when a requested service has more than one credential of the required type, and to pass the choice through `conversationContext` / `assumptions`.
- Builder (`build-workflow-agent.prompt`) and shared `PLACEHOLDERS_RULE`: when `explore-resources` returns multiple matches and the user did not name one, wrap the parameter with `placeholder('Select <resource>')` instead of guessing. The setup wizard already surfaces placeholder values post-build.
- Unit tests cover single-credential auto-apply and multi-credential ambiguity (no auto-apply, no credential test, `needsAction=true`).

https://linear.app/n8n/issue/INS-103

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/instance-ai/src/agent/system-prompt.ts
+++ b/packages/@n8n/instance-ai/src/agent/system-prompt.ts
@@ -203,6 +203,8 @@ Always pass \`conversationContext\` when spawning background agents (\`build-wor
 
 **Credentials**: Call \`credentials(action="list")\` first to know what's available. Build the workflow immediately — the builder auto-resolves available credentials and auto-mocks missing ones. Planned builder tasks handle their own verification and credential finalization flow.
 
+**Ask once when a service has multiple credentials of the same type.** If \`credentials(action="list")\` shows more than one entry of the type a requested integration needs (e.g. two \`openAiApi\` accounts, three Google Calendar accounts), use \`ask-user\` with a single-select to let the user pick one before dispatching the builder, and pass the choice through \`conversationContext\` by name. Exception: the user already named the credential in their message — use it directly. With a single candidate, auto-apply and do not ask.
+
 ${SECRET_ASK_GUARDRAIL}
 
 **Post-build flow** (for direct builds via \`build-workflow-with-agent\`):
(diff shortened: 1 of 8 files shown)


packages/@n8n/instance-ai/src/agent/system-prompt.ts
```diff
@@ -205,6 +205,8 @@ Always pass \`conversationContext\` when spawning background agents (\`build-wor
 
 **Credentials**: Call \`credentials(action="list")\` first to know what's available. Build the workflow immediately — the builder auto-resolves available credentials and auto-mocks missing ones. Planned builder tasks handle their own verification and credential finalization flow.
 
+**Ask once when a service has multiple credentials of the same type.** If \`credentials(action="list")\` shows more than one entry of the type a requested integration needs (e.g. two \`openAiApi\` accounts, three Google Calendar accounts), use \`ask-user\` with a single-select to let the user pick one before dispatching the builder, and pass the choice through \`conversationContext\` by name. Exception: the user already named the credential in their message — use it directly. With a single candidate, auto-apply and do not ask.
```
r00gm, line 208: something i've seen, given that we paginate if the user has more than X credentials, as in, the credential is not on the first page, the builder says, it doesn't exist.

Happy to approve, just don't know if that issue should also be tackled here?

aalises: [I'd do that as a follow-up (ticket here)](https://linear.app/n8n/issue/INS-137/ai-reports-credential-doesnt-exist-when-its-on-a-later-page). I see that `credentials(action="list")` defaults to limit=50 and has no name filter. The handler returns total, but seems like the system prompt doesn't tell the agent to compare total vs returned length, pre-filter by type, or paginate before declaring a credential missing.....                  


Follow-up here. https://github.com/n8n-io/n8n/pull/29098

r00gm (approved):

PR #29107, opened 2026-04-24
fix(Salesforce Node): Fix trigger not firing on repeated record updates

### LINKED ISSUE #21028: Salesforce Trigger fires only once for repeated Account updates
### Bug Description

When using the Salesforce Trigger node with the "Account updated" event, the trigger runs only once per account.

For example:
- Updating an account in Salesforce the first time correctly triggers the workflow.
- Subsequent updates to the same account do not trigger the node again.

From reviewing the following lines in the code, it appears that IDs are filtered and stored in the workflow’s static data to persist between runs:

https://github.com/n8n-io/n8n/blob/b2e0f8cdc0058fb764c37964864858327405f2d1/packages/nodes-base/nodes/Salesforce/SalesforceTrigger.node.ts#L273-L279

In my opinion, this behavior is a bug. When listening for changes on an Account, it’s reasonable to expect that multiple updates to the same record would each trigger the workflow.

### To Reproduce

1. Add a Salesforce Trigger node with the "Account updated" option selected.
2. Update an account in Salesforce and confirm that the workflow runs.
3. Wait at least the poll interval.
4. Update the same account again.
5. Observe that the workflow does not start.

### Expected behavior

The trigger should fire every time an account is updated, even if it’s the same record.

### Debug Info

<details><summary>Debug Info</summary>
<p>

## core

- n8nVersion: 1.116.2
- nodeJsVersion: 22.18.0
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- consumerId: 14352590-bf28-49aa-9666-b0c150cd4653

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/141.0.0.0 safari/537.36
- isTouchDevice: false

Generated at: 2025-10-21T19:47:07.371Z

</p>
</details> 

### Operating System

Debian 13 Trixie

### n8n Version

1.116.2

### Node.js Version

22.18.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

Joffcom:
## Summary

The Salesforce Trigger node was only firing once per record, even when the same account, lead, contact, or other sobject was updated multiple times. Subsequent updates to any record already seen by the trigger were silently dropped.

**Root cause:** `filterAndManageProcessedItems` in `GenericFunctions.ts` stored seen record `Id` values in static workflow data and used them as dedup keys. Because every update to the same Salesforce record shares the same `Id`, subsequent updates were filtered as duplicates.

**Fix:** Changed the dedup key from `Id` to `Id_LastModifiedDate`. Since Salesforce updates `LastModifiedDate` on every write, each genuine update produces a unique composite key. The 15-minute overlap window still correctly deduplicates the same record version appearing in back-to-back polls (same `Id_date` → filtered), while new updates (new `Id_date`) fire through.

`LastModifiedDate` is also added to the default field list for all 8 sobject types and the custom-object fallback, so it is always fetched by the poll query.

**Upgrade note:** On first poll after upgrade, records within the 15-minute overlap window may trigger once more (old `Id`-only keys don't match new composite keys). This is the minimum unavoidable disruption given the change in key format.

**How to test:**
1. Set up a Salesforce Trigger node with "Account updated"
2. Activate the workflow and update an account — workflow runs ✓
3. Wait at least the poll interval, update the same account again — workflow runs again ✓ (was broken before)

https://linear.app/n8n/issue/NODE-4428

fixes #21028

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

---
> When testing manually use the n8n2 credential, I tested with an older n8n version running then used this PR once I confirmed it wasn't working.


michael-radency (dismissed):
LGTM

michael-radency (approved):

PR #29135, opened 2026-04-25
fix(core): preserve bracket secret provider keys

fuyua9:
## Summary
- keep dot-notation external secret provider extraction restricted to JS-safe provider keys
- allow bracket-notation provider references to preserve existing hyphenated provider keys
- add regression coverage for `$secrets['aws-secrets-manager']['token']`

## Tests
- `node` smoke check of the updated extraction regex for hyphenated bracket provider keys, malformed dot notation, and existing bracket provider access
- `git diff --check`

> I did not run the repository Jest target locally because this checkout does not have `node_modules` installed.

### DIFF
--- a/packages/cli/src/credentials/__tests__/external-secrets.utils.test.ts
+++ b/packages/cli/src/credentials/__tests__/external-secrets.utils.test.ts
@@ -28,6 +28,14 @@ describe('External secrets utils', () => {
 			).toEqual(['awsSecretsManager']);
 		});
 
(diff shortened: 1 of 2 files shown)


fuyua9:
Rebased the fix onto current n8n master without carrying over the stale snapshot history, then pushed 866b943. The patch changes only the bracket-provider regex and adds coverage for hyphenated keys such as aws-secrets-manager. Static validation passed with git diff --check. The targeted Vitest could not run locally because this worktree has no node_modules/vitest installed.

PR #29137, opened 2026-04-25
ci: support list-style ci-filter patterns

haimingZZ:
## Summary
- support `- pattern` list-style lines in the multiline `ci-filter` DSL
- keep the existing plain indented pattern format working
- add regressions for the `util-data-tooling` DB filter shape so DB/package changes do not skip the tooling workflow

## Tests
- `node --test .github/actions/ci-filter/__tests__/ci-filter.test.ts`
- `git diff --check`


shortstacked (approved):
Thanks for this!

PR #29150, opened 2026-04-27
fix(cli): use valid grant type fallback when authorization server omits grant_types_supported

### LINKED ISSUE #29147: OAuth discovery fails when authorization server omits optional grant_types_supported
## Summary

When using dynamic client registration for MCP/OAuth, n8n can throw:
No supported grant type and authentication method found
even when OAuth discovery succeeds and the authorization server advertises usable auth methods and PKCE support.
The issue is in the fallback used when authorization server metadata omits grant_types_supported. In commit 35449aaa7b1ca016c919e555f9ba86ee0ea1db8f, n8n falls back to: `['authorization_code', 'implicit']` but `selectGrantTypeAndAuthenticationMethod()` requires both `authorization_code` and `refresh_token` for auth-code / PKCE flow. That makes the fallback internally inconsistent and causes valid servers to fail.

## Affected code

`packages/cli/src/oauth/oauth.service.ts`

Current logic:

```
const { grantType, authentication } = this.selectGrantTypeAndAuthenticationMethod(
	metadataValidation.data.grant_types_supported ?? ['authorization_code', 'implicit'],
	metadataValidation.data.token_endpoint_auth_methods_supported ?? ['client_secret_basic'],
	metadataValidation.data.code_challenge_methods_supported ?? [],
);
```
and:
```
if (grantTypes.includes('authorization_code') && grantTypes.includes('refresh_token')) {
	if (codeChallengeMethods.includes('S256')) {
		return { grantType: 'pkce' };
	}
	...
}
```

## Why this is a bug

`grant_types_supported is optional in authorization server metadata. A compliant authorization server may omit it.

In that case, n8n currently assumes: `['authorization_code', 'implicit']`
but then refuses to use auth-code/PKCE unless refresh_token is also present. So discovery succeeds, metadata is valid, PKCE may be supported, token auth methods may be supported, and n8n still throws.

## Suggested fix

Either:
### Option A: Fix the fallback

Change fallback from:
`['authorization_code', 'implicit']`
to something that matches supported flows, for example:
`['authorization_code', 'refresh_token', 'client_credentials']`

### Option B: Fix the selector

Do not require `refresh_token` to be present before using authorization_code / PKCE.
For example, treat authorization_code as sufficient for auth-code flow, and treat refresh-token support as optional.

Why Option B seems safer `refresh_token` is not the capability that determines whether auth-code / PKCE can start. Requiring it up front is overly strict and blocks otherwise valid OAuth servers.

## Impact
This can break MCP/OAuth integrations even when:
- protected resource discovery works
- authorization server discovery works
- PKCE is supported
- token endpoint auth methods are supported

### To Reproduce

Observed against an Auth0 authorization server discovered via MCP protected resource metadata.

Auth server metadata includes:
- authorization_endpoint
- token_endpoint
- registration_endpoint
- token_endpoint_auth_methods_supported
- code_challenge_methods_supported: ["S256", "plain"]

but omits:
- grant_types_supported
Example effect at runtime:
- grantTypes = ['authorization_code', 'implicit']
- tokenEndpointAuthMethods = ['client_secret_basic', 'client_secret_post', 'private_key_jwt']
- codeChallengeMethods = ['S256', 'plain']

Result:
- auth code branch fails because refresh_token is missing
- client credentials branch fails because client_credentials is missing
- n8n throws No supported grant type and authentication method found

### Expected behavior

If `grant_types_supported` is missing, n8n should use a fallback that is compatible with its own selector and with common real-world OAuth servers.
At minimum, PKCE/auth-code should not fail purely because the metadata omitted `grant_types_supported`.

### Debug Info

# Debug info

## core

- n8nVersion: 2.17.7
- nodeJsVersion: 24.14.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- isTouchDevice: false

Generated at: 2026-04-26T21:19:54.659Z

### Operating System

Arch Linux

### n8n Version

2.17.7

### Node.js Version

24.14.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

octo-patch:
Fixes #29147

## Problem

When an authorization server omits the optional `grant_types_supported` field from its OAuth discovery metadata (which is valid per [RFC 8414](https://www.rfc-editor.org/rfc/rfc8414)), n8n fell back to `['authorization_code', 'implicit']`. However, `selectGrantTypeAndAuthenticationMethod()` requires **both** `authorization_code` **and** `refresh_token` to be present before selecting the auth-code or PKCE flow. Since `implicit` ≠ `refresh_token`, neither the PKCE branch nor the auth-code branch matched, causing:

```
BadRequestError: No supported grant type and authentication method found
```

This broke MCP/OAuth integrations (e.g. Auth0) that:
- Support PKCE (`code_challenge_methods_supported: ['S256', 'plain']`)
- Advertise token auth methods
- But omit `grant_types_supported` from their metadata

## Solution

Change the fallback value for `grant_types_supported` from `['authorization_code', 'implicit']` to `['authorization_code', 'refresh_token']`. This aligns the fallback with what `selectGrantTypeAndAuthenticationMethod()` expects, so PKCE or auth-code flow is correctly selected when the server omits this optional field.

## Testing

Added a unit test in `oauth.service.test.ts` that simulates an authorization server omitting `grant_types_supported` but advertising `S256` PKCE support. The test verifies that `grantType` is correctly set to `'pkce'` without throwing.

### DIFF
--- a/packages/cli/src/oauth/__tests__/oauth.service.test.ts
+++ b/packages/cli/src/oauth/__tests__/oauth.service.test.ts
@@ -1512,6 +1512,62 @@ describe('OauthService', () => {
 			expect(callArgs[2] || []).toEqual([]);
 		});
 
+		it('should use pkce when server omits grant_types_supported but supports S256 and PKCE', async () => {
+			const axios = require('axios');
+			const { ClientOAuth2 } = await import('@n8n/client-oauth2');
+			const mockGetUri = jest.fn().mockReturnValue({
+				toString: () =>
+					'https://example.domain/oauth2/auth?client_id=registered_client_id&redirect_uri=http://localhost:5678/rest/oauth2-credential/callback&response_type=code&state=state',
+			});
+			jest.mocked(ClientOAuth2).mockImplementation(
+				() =>
+					({
+						code: {
+							getUri: mockGetUri,
+						},
+					}) as any,
+			);
+
+			const credential = mock<CredentialsEntity>({ id: '1', type: 'googleOAuth2Api' });
+			const oauthCredentials = {
+				serverUrl: 'https://example.domain',
+				useDynamicClientRegistration: true,
+			} as OAuth2CredentialData;
+
+			jest.spyOn(service, 'getOAuthCredentials').mockResolvedValue(oauthCredentials);
+			// Simulate an authorization server that omits grant_types_supported (valid per RFC 8414)
+			// but advertises S256 PKCE support and standard token auth methods.
+			jest.mocked(axios.get).mockResolvedValue({
+				data: {
+					authorization_endpoint: 'https://example.domain/oauth2/auth',
+					token_endpoint: 'https://example.domain/oauth2/token',
+					registration_endpoint: 'https://example.domain/oauth2/register',
+					// grant_types_supported intentionally omitted
+					token_endpoint_auth_methods_supported: ['client_secret_basic', 'client_secret_post'],
+					code_challenge_methods_supported: ['S256', 'plain'],
+				},
+			} as any);
+
+			jest.mocked(axios.post).mockResolvedValue({
+				data: {
+					client_id: 'registered_client_id',
+					client_secret: 'registered_client_secret',
+				},
+			} as any);
+
+			jest.spyOn(service, 'encryptAndSaveData').mockResolvedValue(undefined);
+
+			const authUri = await service.generateAOauth2AuthUri(credential, {
+				cid: credential.id,
+				origin: 'static-credential',
+				userId: 'user-id',
+			});
+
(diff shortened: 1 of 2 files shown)


vcastellm:
@octo-patch you're missing to sign the CLA

vcastellm:
@octo-patch this fix doesn't look appropriate because it changes the RFC standard, better take Option B

PR #29151, opened 2026-04-27
fix(core): Require auth token to mount credentials overwrite endpoint

etairl:
## Summary

Refuses to register the credentials overwrite POST route unless `CREDENTIALS_OVERWRITE_ENDPOINT_AUTH_TOKEN` is configured. If `CREDENTIALS_OVERWRITE_ENDPOINT` is set without the token, the server now throws an `UnexpectedError` at startup instead of mounting an unauthenticated handler. The same guard is applied to the worker server's overwrite route in `packages/cli/src/scaling/worker-server.ts`.

The dual-mode handler in `packages/cli/src/server.ts` is also removed: with auth always required on this route, the single-write fallback (`presetCredentialsLoaded` + `authenticationEnforced`) is no longer needed and the field is dropped.

### How to test

- Set `CREDENTIALS_OVERWRITE_ENDPOINT=/rest/presets` without `CREDENTIALS_OVERWRITE_ENDPOINT_AUTH_TOKEN` → server fails fast at startup with a clear error.
- Set both env vars → endpoint mounts as before; requests without a matching `Authorization: Bearer <token>` header get `401`.
- Run `pnpm --filter n8n test src/scaling/__tests__/worker-server.test.ts` — includes a new test asserting the fail-closed behavior.

_None — reported externally._

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/scaling/__tests__/worker-server.test.ts
+++ b/packages/cli/src/scaling/__tests__/worker-server.test.ts
@@ -121,6 +121,9 @@ describe('WorkerServer', () => {
 			const CREDENTIALS_OVERWRITE_ENDPOINT = 'credentials/overwrites';
 			globalConfig.credentials.overwrite.endpoint = CREDENTIALS_OVERWRITE_ENDPOINT;
 
+			const middleware = (_req: Request, _res: Response, next: NextFunction) => next();
+			credentialsOverwriteService.getOverwriteEndpointMiddleware.mockReturnValue(middleware);
+
 			await workerServer.init({ health: true, overwrites: true, metrics: true });
 
 			expect(app.get).toHaveBeenCalledWith('/internal/health', expect.any(Function));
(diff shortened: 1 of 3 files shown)


Joffcom:
@etairl any concerns about this being a breaking change?

etairl:
> @etairl any concerns about this being a breaking change?

Only deployments that set CREDENTIALS_OVERWRITE_ENDPOINT without CREDENTIALS_OVERWRITE_ENDPOINT_AUTH_TOKEN are affected, and those deployments are precisely the ones currently mounting an unauthenticated endpoint. The startup error message tells the operator exactly what env var to add.

I think marking this as a breaking change (fix(core)!: + BREAKING CHANGE: footer + BREAKING-CHANGES.md entry) is the right call given the prior behavior was unsafe and the fix is a one-env-var change. Happy to go with a softer warn-and-skip approach if you'd prefer not to break startup. Let me know..

PR #29156, opened 2026-04-27
fix(ai-builder): Adjust chat trigger memory builderHints

Cadiac:
## Summary

Make the ai builder aware of how memory on Chat trigger node works, it was misconfiguring it when it was asked to create chat workflows with memory, not understanding what the connection is for & when to configure it, sometimes making workflows like this

image

Where it should've either connnected a memory to the chat trigger or to leave the property alone.

https://linear.app/n8n/issue/INS-136/builder-defines-chat-triggers-wrong-with-memory

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
```diff
@@ -141,7 +141,10 @@ const commonOptionsFields: INodeProperties[] = [
 		],
 		default: 'notSupported',
 		description: 'If loading messages of a previous session should be enabled',
-		builderHint: { message: "Set to 'memory' to persist conversation history across sessions" },
+		builderHint: {
+			message:
+				"This ONLY rehydrates the chat widget UI with prior messages when the user reopens it — it does NOT give the Agent memory. The Agent gets memory from its own ai_memory subnode regardless of this setting. Only set to 'memory' if the user wants the widget to restore visible history on reload; if so, you MUST also connect a memory subnode to this trigger's ai_memory input (use the same memory node as the Agent). Otherwise leave as 'notSupported'.",
```
mutdmour: `ai_memory` is not something explicitly explain the workflow sdk. goal of the sdk is to abstract away the technical details of n8n as much as possible.

packages/@n8n/instance-ai/src/tools/nodes/suggested-nodes-data.ts
```diff
@@ -22,7 +22,7 @@ export const suggestedNodesData: Record<string, CategoryData> = {
 		nodes: [
 			{
 				name: '@n8n/n8n-nodes-langchain.chatTrigger',
-				note: 'When loadPreviousSession is set to memory, the downstream Agent must also have its own memory subnode to maintain conversation context during processing',
+				note: "options.loadPreviousSession is ONLY about rehydrating the chat widget UI when a user reopens it — it does NOT give the Agent memory. The Agent always gets memory via its own ai_memory subnode, independent of this setting. Only set loadPreviousSession to 'memory' if you want the widget to restore visible history on reload; if you do, the Chat Trigger itself requires a memory subnode on its ai_memory input (point it at the same memory node as the Agent, using the same session key, so widget history matches what the Agent remembers). Otherwise leave loadPreviousSession unset/'notSupported' — the Agent will still have full memory",
```
mutdmour: maybe we should move this entire note to the node, instead of expanding it here, where it's more likely to have side effects.

packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
Cadiac: Good point, adjusted the language to match the SDK terms.

We seem to have some mentions of `ai_memory` on these best practices prompts https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/workflow-sdk/src/prompts/best-practices/guides/chatbot.ts#L19-L20 but I don't think we use them at the moment at instance AI?

packages/@n8n/instance-ai/src/tools/nodes/suggested-nodes-data.ts
Cadiac: maybe, I left a short message here anyways for now

packages/@n8n/nodes-langchain/nodes/trigger/ChatTrigger/ChatTrigger.node.ts
mutdmour: they might be getting imported as part of the tools template. so yes this would just be confusing the builder.

packages/@n8n/instance-ai/src/tools/nodes/suggested-nodes-data.ts
mutdmour: without evaluations it's hard to feel confident. Especially since this gets imported into a lot of flows, when the the suggested nodes tool is called

mutdmour (approved):

packages/@n8n/instance-ai/src/tools/nodes/suggested-nodes-data.ts
Cadiac: yeah.. lets revisit these when we can more accurately measure effect of changes like this, I think your hunch of these potentially just being harmful might be right

PR #29161, opened 2026-04-27
feat(core): Add MCP tool search executions

dariacodes:
## Summary

- Added a new MCP tool `search_executions`
- Moved the code for adding a filter by a boolean settings value from workflow repository to a shared util `applyWorkflowBooleanSettingFilter`
- Added new filters to execution query builder

Closes [ADO-5020](https://linear.app/n8n/issue/ADO-5020/feature-tool-to-get-all-executions-for-a-workflow)

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/mcp/tools/search-executions.tool.ts
```diff
@@ -0,0 +1,151 @@
...
+import type { WorkflowFinderService } from '@/workflows/workflow-finder.service';
+
+import { USER_CALLED_MCP_TOOL_EVENT } from '../mcp.constants';
+import type { ToolDefinition, UserCalledMCPToolEventPayload } from '../mcp.types';
+import { createLimitSchema } from './schemas';
+import { getMcpWorkflow } from './workflow-validation.utils';
+
+const MAX_RESULTS = 200;
+
+const inputSchema = {
```
MiloradFilipovic, line 17: I as expecting a time range filter (first thing I asked CC was to find failed executions in past week). Should we add it? Seems to be supported by `RangeQuery` already so hopefully it's straightforward to add.

packages/cli/src/modules/mcp/tools/search-executions.tool.ts
```diff
@@ -0,0 +1,151 @@
...
+		} catch (error) {
+			const errorMessage = error instanceof Error ? error.message : String(error);
+
+			telemetryPayload.results = {
+				success: false,
+				error: errorMessage,
+			};
+			telemetry.track(USER_CALLED_MCP_TOOL_EVENT, telemetryPayload);
+
+			const output = { data: [], count: 0, estimated: false, error: errorMessage };
```
MiloradFilipovic: We are adding `error` field here, but it's not declared in the `outputSchema` which will not work in clients that respect `structuredOutput`. Lets add it to the schema as optional prop.

packages/cli/src/modules/mcp/tools/search-executions.tool.ts
```diff
@@ -0,0 +1,151 @@
...
+				startedAt: z.string().nullable().describe('ISO timestamp when the execution started'),
+				stoppedAt: z.string().nullable().describe('ISO timestamp when the execution stopped'),
+				waitTill: z
+					.string()
+					.nullable()
+					.describe('ISO timestamp until when the execution is waiting'),
+			}),
+		)
+		.describe('List of executions matching the query'),
+	count: z.number().int().min(0).describe('Total number of executions matching the filters'),
```
MiloradFilipovic: Looking at the executions service, `getCountForQuery` can actually return `-1` so count is not always `min(0)`. This could cause issues with mcp clients if the response doesn't match the schema. Just changing to `min(-1)` would do it but if we want to be really descriptive (usually with LLMs you want it) we can go with:
```suggestion
	count: z
	  .union([z.literal(-1), z.number().int().min(0)])
	  .describe('Total matching executions, or -1 if the count is unavailable'),
```

packages/cli/src/modules/mcp/tools/search-executions.tool.ts
```diff
@@ -0,0 +1,151 @@
...
+			telemetry.track(USER_CALLED_MCP_TOOL_EVENT, telemetryPayload);
+
+			return {
+				structuredContent: payload,
+				content: [{ type: 'text', text: JSON.stringify(payload) }],
+			};
+		} catch (error) {
+			const errorMessage = error instanceof Error ? error.message : String(error);
+
+			telemetryPayload.results = {
```
MiloradFilipovic, line 160: Should we add `error_reason` ([example](https://github.com/n8n-io/n8n/blob/90d875ce3e5a2a004a5a3d8f28ac4e9820b109f4/packages/cli/src/modules/mcp/tools/get-execution.tool.ts#L184)) here? It's been useful for us in the past for debugging issues with mcp.

packages/@n8n/db/src/repositories/execution.repository.ts
```diff
@@ -1092,19 +1108,31 @@ export class ExecutionRepository extends Repository<ExecutionEntity> {
 		return qb;
 	}
 
+	private toMcpQueryBuilder(query: ExecutionSummaries.Query) {
```
MiloradFilipovic: Ideally, we don't want any modules-related code in core. Not sure how to work around this so happy to leave it like this also.

MiloradFilipovic (dismissed):
Works nicely, thanks 🙌
I left a few non-blocking suggestions for improvements so take a look at what makes sense.

packages/cli/src/modules/mcp/tools/search-executions.tool.ts
dariacodes: Good catch!

packages/cli/src/modules/mcp/tools/search-executions.tool.ts
```diff
@@ -0,0 +1,173 @@
...
+					limit: safeLimit,
+					...(lastId ? { lastId } : {}),
+				},
+				order: { startedAt: 'DESC' as const },
+				...(workflowId ? { workflowId } : {}),
+				...(status?.length ? { status } : {}),
+				...(startedAfter ? { startedAfter } : {}),
+				...(startedBefore ? { startedBefore } : {}),
+				isArchived: false,
+				workflowBooleanSettings: [{ key: 'availableInMCP', value: true }],
```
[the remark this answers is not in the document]
dariacodes: Where is this coming from? 😆 We removed the requirement to show only active workflows in MCP quite a while ago.

packages/@n8n/db/src/repositories/execution.repository.ts
dariacodes: Very good point. I simplified everything by just adding a possibility to add filters that we need. I don't like the name `workflowBooleanSettings` that much, but I couldn't come up with a better one, as it is DB dependent. How do you like this solution?

packages/cli/src/modules/mcp/tools/search-executions.tool.ts
```diff
@@ -0,0 +1,173 @@
...
+const MAX_RESULTS = 200;
+
+const inputSchema = {
+	workflowId: z.string().optional().describe('Filter executions by workflow ID'),
+	status: z
+		.array(z.enum(ExecutionStatusList))
+		.optional()
+		.describe('Filter by execution status(es)'),
+	startedAfter: z
+		.string()
```
MiloradFilipovic, line 24: Should we use Zod's [ISO datetime](https://zod.dev/api?id=iso-datetimes#iso-datetimes) for this?

MiloradFilipovic (dismissed):
Thanks for addressing left one more minor comment

packages/@n8n/db/src/repositories/execution.repository.ts
MiloradFilipovic: Yeah, this does look better. Don't have better suggestion for the name so let's keep it.

MiloradFilipovic (approved):

PR #29207, opened 2026-04-27
fix(instance-ai): Make SANDBOX_LINK_SDK flag bypass Dockerfile SDK install (no-changelog)

OlegIvaniv:
## Summary
Fixes sandbox creation in `@n8n/instance-ai` when the host `@n8n/workflow-sdk` version is ahead of what's published on npm (the typical state on `master` between releases).

**Symptom:** sandbox creation fails with `Sandbox failed to become ready within the timeout period` and a `502 Bad Gateway` from the sandbox service.

**Root cause:** the Dockerfile baked by `N8nSandboxImageManager` runs `npm install` against `PACKAGE_JSON`, which pins `@n8n/workflow-sdk` to the host version (e.g. `0.11.4`). When that version isn't on the registry, image build hits a 404 inside the sandbox service — the sandbox never becomes ready.

`N8N_INSTANCE_AI_SANDBOX_LINK_SDK=1` was supposed to be the escape hatch, but it ran *after* sandbox creation, so the failed image build killed it before it could help.

**Fix:** under the flag, bake an SDK-less `PACKAGE_JSON` (only `tsx` + `@types/node`). The post-creation step then `npm install`s the packed workspace tarball as before. If packing fails under the flag, we now throw with a clear `pnpm build` hint instead of silently degrading to a broken sandbox.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/@n8n/instance-ai/src/workspace/__tests__/sandbox-setup.test.ts
+++ b/packages/@n8n/instance-ai/src/workspace/__tests__/sandbox-setup.test.ts
@@ -1,5 +1,57 @@
+import { jsonParse } from 'n8n-workflow';
+
 import type { SearchableNodeDescription } from '../../types';
 import { formatNodeCatalogLine } from '../sandbox-setup';
+
+function loadSandboxPackageJson(linkSdk: boolean): {
+	dependencies: Record<string, string>;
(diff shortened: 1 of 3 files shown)


Cadiac (approved):
👍

PR #29211, opened 2026-04-27
feat(agents): Schedule trigger for agents

burivuhster:
## Summary

Adds a schedule trigger for agents so a published agent can run from a cron rule with a configurable wake-up prompt.

## Key design decisions

- Store the schedule trigger in the existing agent `integrations` JSON to keep this POC migration-free.
- Execute only the published agent snapshot so scheduled runs behave like production integrations, not draft/test chat.
- Keep scheduling local to the main process with an in-memory cron registry; multi-main leader coordination is intentionally out of scope for this POC.
- Use a project member identity for compilation/tool access, but a fresh per-run `threadId`/`resourceId` so scheduled runs do not share chat memory.

- https://linear.app/n8n/issue/AGENT-38/feature-run-agent-on-schedule

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
```diff
@@ -0,0 +1,375 @@
...
+import { ConflictError } from '@/errors/response-errors/conflict.error';
+
+import { AgentsCredentialProvider } from '../adapters/agents-credential-provider';
+import { AgentsService } from '../agents.service';
+import type { Agent } from '../entities/agent.entity';
+import { AgentRepository } from '../repositories/agent.repository';
+
+@Service()
+export class AgentScheduleService {
+	private readonly jobs = new Map<string, CronJob>();
```
mike12345567, line 28: I assume this is one of the things we'll need to sort out for multi-main, if we have multiple instances this could be an issue.

In the past I've used BullMQ backed by Redis: https://docs.bullmq.io/guide/job-schedulers - it can take a crontab and load jobs onto the queue to be picked up by one of the workers, could be an option, do you know how we handle the schedule trigger in workflows today for multi-main?

packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
```diff
@@ -0,0 +1,375 @@
...
+		if (userIds.length === 0) {
+			return undefined;
+		}
+
+		const publishedById = agent.publishedVersion?.publishedById;
+		if (publishedById && userIds.includes(publishedById)) {
+			return publishedById;
+		}
+
+		return userIds[0];
```
mike12345567: should this default like this, if there isn't a published by ID should we return undefined like there was no user IDs?

mike12345567 (approved):
Looks good, just two questions!

packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
burivuhster: I would rather go with a refactor of the existing ScheduledTaskService. It's now tied to workflows, but should be possible to untie. As far as I understand it sets cron jobs on leader instance only and just have a bit of logic to ensure no duplicate jobs triggered.

packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
burivuhster: Good catch, fixed

packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
```diff
@@ -0,0 +1,375 @@
...
+		if (userIds.length === 0) {
+			return undefined;
+		}
+
+		const publishedById = agent.publishedVersion?.publishedById;
+		if (publishedById && userIds.includes(publishedById)) {
+			return publishedById;
+		}
+
+		return undefined;
```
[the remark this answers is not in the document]
burivuhster: This edge case needs more thinking anyway, leave it out of scope for now

packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
mike12345567: Right ok, sounds a little flimsy, if the leader instance fails/goes down I can imagine jobs go missing/its not guaranteed the schedule is maintained - that BullMQ cron approach is actually quite solid for distributing scheduled tasks.

For this PR one thing I think it would be good to do - can we tag anywhere like this (essentially this file) with a `TODO` for multi-main support? I think its something we'll need to think about sooner rather than later now given we're planning to release to internal next week.

PR #29297, opened 2026-04-27
refactor(editor): Provide workflow id and document store at app root (no-changelog)

alexgrozav:
## Summary

Hoists workflow id and workflow document store provides from `WorkflowLayout` up to `App.vue` so they are available across the app, replaces `useProvideWorkflowId` with a leaner `useWorkflowId` composable, and renames the workflow route param from `:name` to `:workflowId` for clarity. Cascading updates apply the new param name across views, composables, stores, and tests.

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/app/composables/useWorkflowId.ts
```diff
@@ -0,0 +1,14 @@
+import { computed } from 'vue';
+import { useRoute } from 'vue-router';
+import { VIEWS } from '@/app/constants';
+
+export function useWorkflowId() {
+	const route = useRoute();
+
+	return computed(() => {
+		if (route.name === VIEWS.DEMO || route.name === VIEWS.DEMO_DIFF) return 'demo';
```
autologie, line 13: Would it make sense to prioritize the workflow ID from context if it's provided from a higher level?
At least that's my intuition. 😄

autologie (commented):
Nice refactor!

packages/frontend/editor-ui/src/app/composables/useWorkflowId.ts
alexgrozav: That's a very good point. Thanks for the suggestion!

autologie (approved):

PR #29328, opened 2026-04-27
fix(editor): Show error toasts in Instance AI executable canvas

aalises:
## Summary

Error toasts inside the Instance AI executable workflow preview were being swallowed alongside the routine success/warning notifications, masking workflow execution failures (most notably missing-credential errors). This PR restores error visibility while keeping the rest of the suppression in place.

**Approach**

- `useUIStore` gains a second flag: `allowErrorNotificationsWhenSuppressed`, and `setNotificationsSuppressed` now accepts `{ allowErrors }`. Turning suppression off always clears the allowance, so state can't leak across iframe re-mounts.
- `useToast.showMessage` lets `type: 'error'` messages through when both flags are on, including the existing error telemetry path.
- `WorkflowPreview` exposes a new `allowErrorNotifications` prop (default `false`) and forwards it in the `openWorkflow` postMessage payload. `usePostMessageHandler` writes both store flags from every `openWorkflow` message.
- `InstanceAiWorkflowPreview` opts in by passing `:allow-error-notifications="true"` alongside its existing `suppress-notifications` and `can-execute` flags.

**How to test**

1. Open an Instance AI thread, generate a workflow that uses a credentialed node, and run it without configuring the credential.
2. Confirm the error toast appears inside the embedded canvas and the node retains its execution-error state.
3. Run a workflow that succeeds — confirm no success/warning toasts appear inside the preview.
4. Open a regular workflow canvas and confirm normal toast behavior is unchanged.

https://linear.app/n8n/issue/INS-122

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/WorkflowPreview.test.ts
+++ b/packages/frontend/editor-ui/src/app/components/WorkflowPreview.test.ts
@@ -1,7 +1,7 @@
 import type { Mock, MockInstance } from 'vitest';
 import { createPinia, setActivePinia } from 'pinia';
 import { waitFor } from '@testing-library/vue';
-import type { ExecutionSummary } from 'n8n-workflow';
+import { jsonParse, type ExecutionSummary } from 'n8n-workflow';
 import { createComponentRenderer } from '@/__tests__/render';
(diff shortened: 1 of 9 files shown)


Joco-95 (approved):
Tested working as expected 👍

PR #29329, opened 2026-04-27
feat(core): Add --projectId filter to export:workflow and export:credentials commands

alielkhateeb:
## Summary

https://www.loom.com/share/d4687cc87f034dfa9b8fbbf3bbb223b1

Adds a `--projectId` flag to both `export:workflow` and `export:credentials` CLI commands, allowing users to export only resources belonging to a specific project without having to export everything and filter manually.

**New behaviour:**
- `--projectId=<id>` — export all resources owned/shared in that project
- `--all --projectId` — invalid, mutually exclusive
- `--id --projectId` — invalid, mutually exclusive
- Omitting `--projectId` keeps the existing behaviour (backward-compatible)

**How to test:**
1. Create a project and add some workflows/credentials to it
2. Run `n8n export:workflow --projectId=<id> --output=out.json` and verify only that project's workflows are exported
3. Run `n8n export:credentials --projectId=<id> --output=out.json` and verify only that project's credentials are exported
4. Run with `--all --projectId=<id>` and verify it returns an error

https://linear.app/n8n/issue/LIGO-481

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/commands/export/credentials.ts
+++ b/packages/cli/src/commands/export/credentials.ts
@@ -11,6 +11,7 @@ import z from 'zod';
 import type { ICredentialsDecryptedDb } from '@/interfaces';
 
 import { BaseCommand } from '../base-command';
+import '../../zod-alias-support';
 
 const flagsSchema = z.object({
 	all: z.boolean().describe('Export all credentials').optional(),
@@ -118,7 +119,7 @@ export class ExportCredentialsCommand extends BaseCommand<z.infer<typeof flagsSc
 
 		const credentials: ICredentialsDb[] = await Container.get(CredentialsRepository).find({
 			where: this.getWhereFilter(flags),
-			relations: { shared: true },
+			relations: ['shared.project'],
 		});
 
 		if (flags.decrypted) {
--- a/packages/cli/src/commands/export/credentials.ts
+++ b/packages/cli/src/commands/export/credentials.ts
@@ -21,6 +21,7 @@ const flagsSchema = z.object({
 		)
 		.optional(),
 	id: z.string().describe('The ID of the credential to export').optional(),
+	projectId: z.string().describe('Export all credentials in the specified project').optional(),
 	output: z
 		.string()
 		.alias('o')
@@ -46,6 +47,7 @@ const flagsSchema = z.object({
 	description: 'Export credentials',
 	examples: [
 		'--all',
+		'--projectId=Ox8O54VQrmBrb4qL',
 		'--id=5 --output=file.json',
 		'--all --output=backups/latest.json',
 		'--backup --output=backups/latest/',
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/export/credentials.ts
```diff
@@ -157,4 +162,16 @@ export class ExportCredentialsCommand extends BaseCommand<z.infer<typeof flagsSc
...
+		if (flags.id) {
+			return { id: flags.id };
+		}
+
+		if (flags.projectId) {
+			return { shared: { projectId: flags.projectId } };
+		}
+
+		return {};
+	}
```
alielkhateeb, line 177: I've thought about extracting this somewhere as it's a bit duplicated, then I chose to leave them as 2 private methods for now, and if we universalise these options we can pull it up to base command class.

packages/cli/src/commands/export/credentials.ts
```diff
@@ -112,9 +116,10 @@ export class ExportCredentialsCommand extends BaseCommand<z.infer<typeof flagsSc
 			}
 		}
 
-		const credentials: ICredentialsDb[] = await Container.get(CredentialsRepository).findBy(
-			flags.id ? { id: flags.id } : {},
-		);
+		const credentials: ICredentialsDb[] = await Container.get(CredentialsRepository).find({
+			where: this.getWhereFilter(flags),
+			relations: { shared: true },
```
konstantintieber: Nit: could we set the relations to just `shared.project` here, as we do for workflows?
Since we don't need the shared credential relationship or any other that may be added in the future for now.

packages/cli/test/integration/commands/export/credentials.test.ts
```diff
@@ -0,0 +1,69 @@
...
+});
+
+test('should export credentials by project with --projectId', async () => {
+	const projectA = await createTeamProject();
+	const projectB = await createTeamProject();
+	const credentialInProjectA = await createCredentials(
+		{ name: 'Project A credential', type: 'test', data: '' },
+		projectA,
+	);
+	await createCredentials({ name: 'Project B credential', type: 'test', data: '' }, projectB);
```
konstantintieber, line 59: I would've expected to see a `createCredentials`  in ` projectA`  here as well.
Otherwise the first assertion of this test will be true, even if both projects would've been queried.

konstantintieber: Wait it's just the line breaks in github that made me miss the first `createCredentials`.

konstantintieber (dismissed):
nice one!

packages/cli/src/commands/export/credentials.ts
konstantintieber: I like the duplication here. I don't think we need to couple export of credentials and worfklows. From a domain perspective, I could see their APIs go different ways/names for arguments in the future.

packages/cli/test/integration/commands/export/credentials.test.ts
alielkhateeb: It's right there above this line you commented on, here: https://github.com/n8n-io/n8n/blob/b6bb44fe067aac41689a40a60f2ad88781d17f89/packages/cli/test/integration/commands/export/credentials.test.ts#L55-L58

packages/cli/src/commands/export/credentials.ts
alielkhateeb: Absolutely, good catch. d55d527cf99dd1f129dbccf5f88ea1e7451f17d3

konstantintieber (approved):

geemanjs (approved):

PR #29340, opened 2026-04-27
feat(editor): Improve Instance AI workflow setup credential handling (no-changelog)

scdekov:
## Summary

Builds on `feature/instance-ai-workflow-setup-rewrite-pr1` with the next slice of the Instance AI workflow setup wizard rewrite, focused on credential handling and end-to-end coverage.

** Includes expectations/recordings for the e2e tests to be able to run without an anthropic api key
Run the e2e tests against the recordings:
```
pnpm build:docker
pnpm test:container:sqlite tests/e2e/instance-ai/instance-ai-workflow-setup.spec.ts --workers 1
```

https://linear.app/n8n/issue/ADO-5117/add-minimal-working-setup-for-credentials-only
## Review / Merge checklist

- [x] Tests included.
- [ ] PR Labeled with backport label (not applicable - feature branch)


packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -350,8 +350,10 @@ export async function buildSetupRequests(
 		let needsAction = false;
 		if (credentialType) {
 			const existingOnNode = node.credentials?.[credentialType];
+			const selectedCredentialId =
```
scdekov: Check if we can drop that as it is not directly related to the main change

packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
```diff
@@ -883,13 +883,15 @@ export class InstanceAiAdapterService {
...
 				});
 
-				return credentials.map(
-					(c): CredentialSummary => ({
-						id: c.id,
-						name: c.name,
-						type: c.type,
-					}),
-				);
+				return [...credentials]
```
scdekov: Shouldn't be needed for this PR

packages/frontend/editor-ui/src/features/ai/instanceAi/__tests__/WorkflowSetupWizard.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+		const { queryByTestId } = renderComponent();
+
+		expect(queryByTestId('instance-ai-workflow-setup-later')).not.toBeNull();
+
+		isComplete.value = true;
+		await nextTick();
+
+		expect(queryByTestId('instance-ai-workflow-setup-later')).toBeNull();
+		expect(queryByTestId('instance-ai-workflow-setup-apply')).not.toBeNull();
```
scdekov, line 122: Add more test coverage

packages/@n8n/instance-ai/docs/e2e-tests.md
```diff
@@ -217,10 +217,11 @@ expectations/instance-ai/should-send-message-and-receive-assistant-response/
 
 ### File Naming
 
-`<timestamp>-<host>-<method>-<path_slugified>-<8char_sha256>.json`
```
CharlieKolb, line 220: Should we contribute this straight to master?

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/components/WorkflowSetupCard.vue
```diff
@@ -0,0 +1,147 @@
...
+	const cred = credentialsStore.getCredentialById(selectedCredentialId.value);
+	return {
+		...card.value.node,
+		credentials: cred ? { [credentialType.value]: { id: cred.id, name: cred.name } } : {},
+	} as INodeUi;
+});
+
+function onCredentialSelected(update: INodeUpdatePropertiesInformation) {
+	const data = update.properties.credentials?.[credentialType.value];
+	if (!data) {
```
CharlieKolb: [nit] could do
```ts
ctx.setSelection(card.value.targetNodeName, credentialType.value, data?.id ?? null);
```

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/components/WorkflowSetupStatus.vue
```diff
@@ -0,0 +1,68 @@
...
+import { N8nIcon, N8nText } from '@n8n/design-system';
+import { i18n } from '@n8n/i18n';
+import type { TerminalState } from '../workflowSetup.types';
+
+defineProps<{ state: TerminalState }>();
+</script>
+
+<template>
+	<div :class="$style.status" :data-test-id="`instance-ai-workflow-setup-status-${state}`">
+		<template v-if="state === 'applying'">
```
CharlieKolb: [nit] Since these all have the same components can we avoid the if/else branches and set the values based on a computed property instead?

And do
```ts
:class=({ ['applying']: $style.iconLoading, ... })[state]
```
or similar

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/components/WorkflowSetupWizard.vue
```diff
@@ -0,0 +1,153 @@
...
+}
+
+.actions {
+	display: flex;
+	align-items: center;
+	gap: var(--spacing--2xs);
+}
+
+.actionButton {
+	--button--font-size: var(--font-size--2xs);
```
CharlieKolb: this seems a bit surprising?

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/composables/useWorkflowSetupActions.ts
```diff
@@ -0,0 +1,168 @@
...
+interface SelectionAccessors {
+	selections: Ref<Record<string, Record<string, string>>>;
+	skippedCardIds: Ref<Set<string>>;
+	isCardComplete: (card: WorkflowSetupCard) => boolean;
+	isCardSkipped: (card: WorkflowSetupCard) => boolean;
+	markCardSkipped: (card: WorkflowSetupCard) => void;
+	buildCompletedNodeCredentials: () => Record<string, Record<string, string>>;
+}
+
+interface ApplyMachine {
```
CharlieKolb, line 16: I would have expected this to reuse/implement an overall "Apply/Defer" machine beyond workflow setup for when the AI has clarification requests 🤔

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/composables/useWorkflowSetupContext.ts
```diff
@@ -0,0 +1,207 @@
...
+		skipCurrentCard: actions.skipCurrentCard,
+	};
+
+	provide(WorkflowSetupContextKey, context);
+
+	return context;
+}
+
+export function useWorkflowSetupContext(): WorkflowSetupContext {
+	const ctx = inject(WorkflowSetupContextKey);
```
CharlieKolb, line 200: I don't fully grasp this yet, but as a quick sanity check - using a store for shared state is not an option here?

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/InstanceAiWorkflowSetup.vue
```diff
@@ -0,0 +1,34 @@
...
+import { toRef } from 'vue';
+import type { InstanceAiCredentialFlow, InstanceAiWorkflowSetupNode } from '@n8n/api-types';
+import WorkflowSetupWizard from './components/WorkflowSetupWizard.vue';
+import WorkflowSetupStatus from './components/WorkflowSetupStatus.vue';
+import { provideWorkflowSetupContext } from './composables/useWorkflowSetupContext';
+
+const props = defineProps<{
+	requestId: string;
+	setupRequests: InstanceAiWorkflowSetupNode[];
+	message: string;
```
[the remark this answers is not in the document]
CharlieKolb: This is true

CharlieKolb (commented):
Some initial comments, nice job with the tests!

packages/@n8n/instance-ai/docs/e2e-tests.md
scdekov: Good point. Will split it

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/components/WorkflowSetupWizard.vue
scdekov: True, will double check

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/composables/useWorkflowSetupActions.ts
scdekov: Yep those seems to be overlapping a lot so I'll look into reusing the logic

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/composables/useWorkflowSetupContext.ts
scdekov: I was also considering a store but decided to go with context as it is simpler and it allows resetting the state eg on new setup flow shown where with a store that would have to be an explicit cleanup. I think both does work though. Any strong arguments for using a store instead of the context?

CharlieKolb: No I was just curious since this isn't a pattern we should default to. But I think this is reasonable 👍

packages/frontend/editor-ui/src/features/ai/instanceAi/workflowSetup/composables/useWorkflowSetupActions.ts
scdekov: Actually after looking at it a little more I think it may be better to keep those separated as for the other requests the pattern is fire and forget which does make sense eg if the user approves a command/domain or selecting a credential/parameter in the simpler UIs. 

With the workflow setup we're passing the apply request again through the `handleSetup` handler which could produce another suspend for follow-ups.

On all of those we're already using the instanceAI.store but the workflow setup apply machine has the extra tool response waiting mechanism so there may not be too much space for reuse here.

CharlieKolb (approved):

PR #29341, opened 2026-04-27
feat(core): Add flag to import workflow cli to activate workflow on import

konstantintieber:
## Summary

Demo video:
https://www.loom.com/share/8467b743450641b091d0bb6027af2fc6

Adds an `--activeState` flag to the `import:workflow` CLI so imported workflows can optionally keep the active state from their JSON, instead of always being force-deactivated.

The implementation is inspired by the `SourceControlImportService.importWorkflowFromWorkFolder` method.

Behavior

- `--activeState=false` (default) — current behavior: all imported workflows are set to `active=false` and must be activated manually later.
- `--activeState=fromJson` — Only supported in queue/multi-main mode. The workflow's active field from the JSON is parsed to decide whether to publish/unpublish a workflow. Workflows with active: true are activated post-import (registered in `ActiveWorkflowManager`, `activeVersionId` set, and a `workflow_history` event recorded); workflows with `active: false` are imported inactive.

How to test

1. Checkout the branch (it's a 1.x one, so make sure you reset your home dir or create a new one)
2. Start n8n in multi main mode, since activation will not work in regular mode. See the instructions in https://github.com/n8n-io/n8n/pull/29079
3. Create a workflow that's inactive with a schedule trigger that runs every 10 seconds
4. Run `n8n export:workflow --backup --output=backups/` inside the docker container of your main-1
5. Edit the generated workflow json and set `active: true`
6. Run `n8n import:workflow --separate --input=backups --activeState=fromJson` inside the docker container of your main-1
7. Observe that the workflow is now active and schedule triggers are creating new executions

https://linear.app/n8n/issue/LIGO-482/support-cli-import-without-deactivating-workflow

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/services/import.service.ts
```diff
@@ -190,6 +216,32 @@ export class ImportService {
...
+			await this.workflowRepository.update(
+				{ id: workflowId },
+				{ activeVersionId: versionIdToActivate },
+			);
+			await this.workflowRepository.updateActiveState(workflowId, true);
+			await this.activeWorkflowManager.add(workflowId, 'activate');
+			didActivate = true;
+		} catch (e) {
+			const error = ensureError(e);
+			this.logger.error(`Failed to activate workflow ${workflowId}`, { error });
```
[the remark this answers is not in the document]
konstantintieber: This is inspired by the SourceControlimportService implementation, where we also get by with just logging the error: https://github.com/n8n-io/n8n/blob/1.x/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts#L792

Let's not diverge from that here.

alielkhateeb:
Very irrelevant to this PR but it is a quick win since we're touching this file: 
https://github.com/n8n-io/n8n/blob/35b3e73d2ce6ae8fa8917cf517045e4d0e9d3a0e/packages/cli/src/commands/import/workflow.ts#L161
The error message here should be `The workflow ...` instead of `The credential...` - of course this is just if you feel like doing it but as I mentioned, completely irrelevant to this PR 😂

packages/cli/src/services/import.service.ts
```diff
@@ -108,6 +117,7 @@ export class ImportService {
 		}
 
 		const insertedWorkflows: IWorkflowBase[] = [];
+		const workflowsToActivate: Array<{ workflowId: string; versionId: string }> = [];
```
alielkhateeb, line 120: super minor nit: We could use a map with composite key (e.g. `${workflowId}:${versionId}`) to make sure we have unique pairs of `workflowId` and `versionId` - might be good on performance in case of duplicates

packages/cli/src/services/import.service.ts
```diff
@@ -190,6 +216,31 @@ export class ImportService {
...
 
+	private async activateWorkflow(workflowId: string, versionIdToActivate: string): Promise<void> {
+		let didActivate = false;
+		try {
+			await this.workflowRepository.update(
+				{ id: workflowId },
+				{ activeVersionId: versionIdToActivate },
+			);
+			await this.workflowRepository.updateActiveState(workflowId, true);
+			await this.activeWorkflowManager.add(workflowId, 'activate');
```
alielkhateeb, line 227: I have quite a few concerns here:
- TypeORM has a default connection pool of size 10, this is called in a for loop, if I import 100 workflows, even 1000 workflows, how will this work? Shouldn't we batch this maybe each 100 workflows in one query with values table?

- I think we can also batch the updateActiveState instead of running it independently per workflow entry, all we need is another updateActiveState that accepts a list of IDs instead of a single one and use `IN(workflowIds)` in the where clause, what do you think?

- Same comment goes to activeWorkflowManager, how would we scale this? Maybe another `addBulk` method or something? - Thinking out loud here

packages/cli/src/services/import.service.ts
```diff
@@ -116,11 +126,19 @@ export class ImportService {
...
+				const versionIdToActivate = workflow.versionId;
+
+				// Always upsert with active=false and activeVersionId=null.
+				// Activation happens post-transaction once the new workflow_history row exists
+				// (the activeVersionId FK references workflow_history.versionId).
+				if (
+					!shouldActivate &&
+					(oldActiveVersionId || workflow.activeVersionId || workflow.active)
+				) {
+					this.logger.info(`Deactivating workflow "${workflow.name}".`);
```
alielkhateeb, line 141: It was quite hard to understand this conditional logger.info tbh, but I fully understand the change in this PR xD

Does this say: If it will remain inactive && was previously active => log "Deactivating workflow"?

If so, I personally find naming those conditionals makes it much easier to understand, something like this:
```ts
const isCurrentlyActive =
	Boolean(oldActiveVersionId) ||
	Boolean(workflow.activeVersionId) ||
	workflow.active;
```
or a better name, as usual naming things is hard

Then the condition will read as: If `!shouldActivate && isCurrentlyActive` => log "Deactivating workflow"

What do you think of this?

packages/cli/src/services/import.service.ts
```diff
@@ -134,6 +152,10 @@ export class ImportService {
 					workflowsNeedingPublishHistory.push({ workflowId, versionId: oldActiveVersionId });
 				}
 
+				if (shouldActivate) {
```
alielkhateeb, line 155: for my own knowledge: Do we always "add" a new version upon import?
I am asking to learn if there is a case where we are importing the exact same workflow, making the import idempotent, maybe that will also reduce the performance load of the import operation?

packages/cli/test/integration/import.service.test.ts
```diff
@@ -331,4 +333,126 @@ describe('ImportService', () => {
...
+			workflowToImport.active = true;
+
+			await importService.importWorkflows([workflowToImport], ownerPersonalProject.id, {
+				activeState: 'fromJson',
+			});
+
+			expect(mockActiveWorkflowManager.remove).toHaveBeenCalledTimes(1);
+			expect(mockActiveWorkflowManager.remove).toHaveBeenCalledWith(existingWorkflow.id);
+			expect(mockActiveWorkflowManager.add).toHaveBeenCalledTimes(1);
+			expect(mockActiveWorkflowManager.add).toHaveBeenCalledWith(existingWorkflow.id, 'activate');
```
alielkhateeb, line 455: I feel that we should assert on the final state from the DB, if I go now and swap the add/remove in the logic, this test will pass because it will detect both methods to be called once, but the final result (workflow) is not asserted on here

alielkhateeb: Wait, is that exactly the same test case? https://github.com/n8n-io/n8n/blob/35b3e73d2ce6ae8fa8917cf517045e4d0e9d3a0e/packages/cli/test/integration/import.service.test.ts#L393

If so, then why not move those assertions on add/remove to that test case instead?

packages/cli/src/services/import.service.ts
konstantintieber: This implementation is identical to the environments import service.
I would rather not open the box now of optimizing this.
See https://github.com/n8n-io/n8n/blob/1.x/packages/cli/src/environments.ee/source-control/source-control-import.service.ee.ts#L768

alielkhateeb: @konstantintieber Sure, but can we create a ticket to revisit this later? I don't mind not optimizing now to deliver a feature but I strongly recommend having a ticket created for optimizing later, otherwise we'll never do it unless a bug is reported (Too late).

packages/cli/src/services/import.service.ts
konstantintieber: Yes we always add a new version.
I'm not aware of users reporting performance issues of the workflow import cli.
Without performance issues being known, I'd rather maintain less code and have redundant versions on import, than adding a json comparison of the current workflow and the workflow to import.
There can also be the product argument that I want to see a new version when I ran the import for a specific json file.

packages/cli/test/integration/import.service.test.ts
konstantintieber: > I feel that we should assert on the final state from the DB

The dbstate testing is handled in the `should activate imported workflow when JSON has active=true` and `should deactivate imported workflow that is updating existing one when JSON has active=false` cases.

For these tests here, we explicitly test the fact that the activeWorkflowManager is called with the correct parameters, because that one's state is only kept in memory, and in the case of the cli command it's the memory of another process.

packages/cli/src/services/import.service.ts
konstantintieber: I'm not against a ticket, but the environments logic has shown no signs of not working for activation.

alielkhateeb (approved):
🚀 looks good

PR #29346, opened 2026-04-27
feat(otel): propagate span context into workflow execution for log-trace correlation

ranimufid:
## Summary

When `N8N_OTEL_ENABLED=true`, the workflow span is created but never made the **active**
context around execution, so `trace.getActiveSpan()` returns `undefined` during the run and log
lines carry no `traceId`/`spanId` — making log-trace correlation impossible.

> **Note (rebased onto current `master`):** this PR was originally opened against an older OTEL
> module and has been re-implemented against the current `execution-level-tracer` /
> `otel-lifecycle-handler` architecture. Earlier merge conflicts are resolved.

### Fix

- **Context propagation bridge** (`packages/core/src/execution-engine/otel-execution-context.ts`):
  a lightweight register/consume bridge so `packages/cli` can propagate the active span context
  into the execution loop in `packages/core` **without** adding `@opentelemetry/api` as a core
  dependency. `ExecutionLevelTracer.startWorkflow` registers a wrapper
  (`context.with(trace.setSpan(context.active(), span), fn)`) keyed by execution id;
  `WorkflowExecute` consumes it and runs the execution inside it. All `await`-ed operations
  (log writes, node executions, hooks) then inherit the active span via AsyncLocalStorage.

- **Non-invasive loop wrapping:** the execution loop is unchanged — the existing execution IIFE
  is renamed to a named function and run through the wrapper
  (`otelWrapper ? otelWrapper(fn) : fn()`). No `continue executionLoop` statements are touched.

- **Logger trace enrichment:** `Logger.prependFormat()` lets `OtelService.injectTraceContextIntoLogs()`
  inject a format that reads `trace.getActiveSpan()?.spanContext()` and adds `traceId`/`spanId` to
  every record. `jsonConsoleFormat` keeps these fields at the JSON root (not nested under
  `metadata`) so log forwarders (e.g. Fluent Bit) surface them as top-level attributes.

- **Robust teardown:** `ExecutionLevelTracer.endWorkflow` always ends dangling node spans and drops
  the per-execution context wrapper, even when the workflow span is missing.

Discovered during self-hosted deployment with `N8N_OTEL_ENABLED=true` and a Fluent Bit log pipeline.

## Review / Merge checklist

- [x] Tests included.


Joffcom:
Hi @ranimufid — thanks for the depth here, log-trace correlation is a really valuable add. Two things from cubic-dev-ai's review need addressing before we can route this for review:

1. **P1 (high confidence):** The edits in `packages/core/src/execution-engine/workflow-execute.ts` removed the outer-loop label, which changed where `continue` jumps to inside the retry loop. The `engineRequest` and null-output paths used to continue the **outer execution loop** but now continue the **inner retry loop** — that's a real control-flow regression independent of the OTEL feature. Could you restore the outer-loop label (or refactor in a way that preserves the same `continue` target) and add a test that exercises the engine-request retry path so this can't regress again?

2. **P2:** In `workflow-end.handler.ts`, the early return when a workflow span is missing skips execution-level cleanup, which can leave the OTEL context / span registry in stale state across executions. Could you either move the early return after the cleanup, or guard the cleanup so it always runs even when the span is absent?

Once those are addressed the team will pick this up for review. Thanks!

ranimufid:
@Joffcom , I believe the requested changes have been made. Would be great if the team can pick this up for review please 🙏

DawkliCrypto:
Could we release that ?

Esaias-Boines:
Hey! 👋 This PR has been open with `triage:needs-info` for over 30 days. As part of our stale PR automation, we're flagging it as stale. If you can address the requested information, please do so — otherwise this PR may be closed in 14 days to keep the queue manageable. Closed PRs can always be reopened! 🙏

Esaias-Boines:
Hey! 👋 It's been **96 days** since this PR was opened with the `triage:needs-info` label and we haven't received a response.

If you can still provide the requested information, please add the details and remove the `triage:needs-info` label so the PR can move forward. Otherwise, this PR may be closed in 14 days to keep the issue queue manageable. Closed PRs can always be reopened! 🙏

*This comment was posted automatically by n8n-monitor.*

ranimufid:
> Hi @ranimufid — thanks for the depth here, log-trace correlation is a really valuable add. Two things from cubic-dev-ai's review need addressing before we can route this for review:
> 
> 1. **P1 (high confidence):** The edits in `packages/core/src/execution-engine/workflow-execute.ts` removed the outer-loop label, which changed where `continue` jumps to inside the retry loop. The `engineRequest` and null-output paths used to continue the **outer execution loop** but now continue the **inner retry loop** — that's a real control-flow regression independent of the OTEL feature. Could you restore the outer-loop label (or refactor in a way that preserves the same `continue` target) and add a test that exercises the engine-request retry path so this can't regress again?
> 2. **P2:** In `workflow-end.handler.ts`, the early return when a workflow span is missing skips execution-level cleanup, which can leave the OTEL context / span registry in stale state across executions. Could you either move the early return after the cleanup, or guard the cleanup so it always runs even when the span is absent?
> 
> Once those are addressed the team will pick this up for review. Thanks!

@Joffcom , the required changes have already been made, but due to the late review, additional merge conflicts have showed up. How do you propose we proceed at this point to get this PR processed?

PR #29351, opened 2026-04-27
feat(ai-builder): Add three new workflow eval test cases (no-changelog)

JoseBra:
## Summary

Picks up the three new test cases from @sirdavidoff's [#29113](https://github.com/n8n-io/n8n/pull/29113) so they can land while that PR is parked. Two of David's prompts triggered the agent's Q&A mode in local runs; this PR keeps the test cases identical except for minimal anti-Q&A cleanups.

| Test case | Status | Change vs David's branch |
|---|---|---|
| `weather-alert.json` | as-is | none — passed first try |
| `linear-bq-leaderboard.json` | as-is + trailer | append `"Configure all nodes as completely as possible and don't ask me for credentials, I'll set them up later."` |
| `workflow-data-table.json` | edits + trailer | drop *"Make sure the workflow is deterministic"* (vague), drop *"and I can give you an API key for it"* (reliable Q&A trigger), append the same trailer |

No structural prescriptions added — node choices, implementation strategy, etc. all remain the agent's decision.

## Why these specific edits

In smoke runs with the unmodified prompts:

- `workflow-data-table` consistently looped on `"A few quick questions before I build this:"` instead of building, cued by *"I can give you an API key for it"*.
- `linear-bq-leaderboard` occasionally stalled with `"I don't see any Linear, BigQuery, or Slack credentials yet. Let me ask a few quick questions..."` — listing three services without a "don't ask for credentials" guard primes the agent toward Q&A mode.

After the cleanups, both built reliably and surfaced real `builder_issue` failures (BQ project param missing, branch synchronization missing, etc.) — exactly the agent-quality signal the eval framework is designed to expose.

## Caveat: harness timeout

`linear-bq-leaderboard` typically takes 215–276s to build. The harness's old 240s build cap will silently kill most attempts. Reliable builds need [#29219](https://github.com/n8n-io/n8n/pull/29219) (TRUST-64, honor `--timeout-ms` across the harness) to land first. Until then, this test case will look flaky in CI.

## Test plan

- [x] All three JSON files parse cleanly (`json.load` validates).
- [x] `workflow-data-table` builds and passes scenario when run with the cleaned prompt locally.
- [x] `weather-alert` builds and passes both scenarios locally.
- [x] `linear-bq-leaderboard` builds when the harness `--timeout-ms` is honored (#29219); surfaces real builder issues afterwards.
- [x] No structural prescriptions added to any prompt — verified by diffing against David's branch.


packages/@n8n/instance-ai/evaluations/data/workflows/linear-bq-leaderboard.json
```diff
@@ -0,0 +1,14 @@
...
...   select\n        timestamp,\n        timestamp_trunc(timestamp, hour) as timestamp_hour,\n        name,\n        f.user_id,\n        f.instance_id,\n        status,\n    from rudder_schema.manual_node_exec_finished f\n    inner join user_accounts a on a.user_id = f.user_id\n    cross join settings\n    where f.timestamp between start_cutoff and end_cutoff\n),\nworkflow_exec as (\n    select\n        timestamp,\n        timestamp_trunc(timestamp, hour) as timestamp_hour,\n        name,\n        f.user_id,\n        f.instance_id,\n        status,\n    from rudder_schema.manual_workflow_exec_finished f\n    inner join user_accounts a on a.user_id = f.user_id\n    cross join settings\n    where f.timestamp between start_cutoff and end_cutoff\n),\nexec as (\n    select * from node_exec\n    union all\n    select * from workflow_exec\n),\nexec_summary as (\n    select\n        name,\n        count(distinct instance_id) as instances,\n        count(distinct timestamp_hour) as hours,\n        count(*) as manual_execs,\n    from exec\n    group by 1\n)\n\nselect * from exec_summary\n\nConfigure all nodes as completely as possible and don't ask me for credentials, I'll set them up later.",
```
lucamattiazzi, line 2: is it ok that the list of names in the prompt and the list of names in the example are different? could it be confusing, or can it be left as it is?

JoseBra: I think it's a fair exercise for the instanceAI too to see if it can understand it, it did not seem to be an issue when running the test in local

lucamattiazzi (approved):

PR #29357, opened 2026-04-28
fix(core): Make task runner grant token TTL configurable

### LINKED ISSUE #28638: Task runner grant token TTL (15s) too short for low-power hardware — causes 100%+ CPU retry loop
### Bug Description

On low-power hardware (Intel N150, 4 cores), the external task runner consistently fails to authenticate with the task broker because the grant token expires before the runner process can spawn and connect.

The grant token TTL is hardcoded to 15 seconds in `task-broker-auth.service.js`:

```js
const GRANT_TOKEN_TTL = 15 * constants_1.Time.seconds.toMilliseconds;
```

**There is no environment variable or configuration option to change this value.**

On slower hardware, the runner takes >15s to:
1. Receive the grant token from the broker
2. Spawn the Node.js/Python process
3. Establish the WebSocket upgrade using the token

This causes a tight retry loop:
- Broker issues grant token → token expires before runner connects → 403 → broker issues new token → repeat
- The retry loop consumes 100%+ CPU continuously, making the entire system unusable
- Both internal and external runners are affected identically

### To Reproduce

1. Deploy n8n v2.14.2 with an external task runner on low-power hardware (Intel N150, Celeron, Atom, or similar)
2. Trigger any workflow that uses a Code node
3. Observe the runner failing with 403 errors in a tight loop
4. CPU usage spikes to 100%+ and stays there

### Expected behavior

- The grant token TTL should be configurable via an environment variable (e.g., `N8N_RUNNERS_GRANT_TOKEN_TTL=120`)
- The default of 15 seconds is reasonable for fast hardware, but there should be a way to increase it for slower systems
- Alternatively, the runner could use a backoff strategy instead of tight retries on 403

### Current workaround

Volume-mount a patched `task-broker-auth.service.js` with a higher TTL:

```yaml
volumes:
  - ./patches/task-broker-auth.service.js:/usr/local/lib/node_modules/n8n/dist/task-runners/task-broker/auth/task-broker-auth.service.js:ro
```

This is fragile and breaks on version updates.

### Debug Info

- n8nVersion: 2.14.2
- database: PostgreSQL
- executionMode: main (default)
- hardware: Intel N150 (4 cores, 12GB RAM)
- runner mode: external (`n8nio/runners:2.14.2`)

### Operating System

Linux (Debian 12)

### n8n Version

2.14.2

### Node.js Version

22.x

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

tomi:
## Summary

The task runner grant token TTL was hardcoded to 15 seconds. On low-power hardware (Intel N150, Celeron, etc.), the runner process can take longer than 15s to spawn and establish a WebSocket connection, causing the grant token to expire. This results in a tight 403 retry loop that pegs CPU at 100%+.

This PR adds a new `N8N_RUNNERS_GRANT_TOKEN_TTL` environment variable (in seconds, default: 15) so operators on slower hardware can increase the TTL. The `TaskBrokerAuthService` now reads from config instead of a hardcoded constant, and injects `TaskRunnersConfig` directly rather than the full `GlobalConfig`.

**How to test:** Set `N8N_RUNNERS_GRANT_TOKEN_TTL=60` and verify the task runner connects successfully on slow hardware, or confirm the default 15s behavior is unchanged without the env var.

https://linear.app/n8n/issue/CAT-2826
closes #28638
https://community.n8n.io/t/code-nodes-are-broken-by-task-runner-connection-attempt-failed-with-status-code-403/236238/7

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI


packages/@n8n/config/src/configs/runners.config.ts
```diff
@@ -64,6 +64,13 @@ export class TaskRunnersConfig {
 	@Env('N8N_RUNNERS_HEARTBEAT_INTERVAL')
 	heartbeatInterval: number = 30;
 
+	/**
+	 * How long (in seconds) a grant token is valid for runner authentication.
+	 * Increase on slow hardware where the runner needs more time to start.
+	 */
+	@Env('N8N_RUNNERS_GRANT_TOKEN_TTL')
+	grantTokenTtl: number = 15;
```
ivov: Accept only positive integers?

packages/cli/src/task-runners/task-broker/auth/__tests__/task-broker-auth.service.test.ts
```diff
@@ -1,5 +1,6 @@
 import { mockInstance } from '@n8n/backend-test-utils';
 import { GlobalConfig } from '@n8n/config';
+import { Time } from '@n8n/constants/src';
```
ivov: We don't typically import like this?

ivov (changes requested):
Nice fix! But how do we prevent this from coming back to us again? 

It'd be good to help the user and support to diagnose this by themselves, or if possible straight mitigate this on our end. Assuming we can reliably detect this case, we could...

- show a warning log?
- increase the TTL automatically?
- retry with exponential backoff?

packages/@n8n/config/src/configs/runners.config.ts
tomi: Added `z.number({ coerce: true }).int().positive()` validation — matches the pattern used in other config fields like `sentry.config.ts`.

packages/cli/src/task-runners/task-broker/auth/__tests__/task-broker-auth.service.test.ts
tomi: Good catch — fixed to import from `'@n8n/constants'` like everywhere else in the codebase.

tomi:
> * show a warning log?

Added a hint to the failure log

> * retry with exponential backoff?

Can't retry, since these are single use tokens. If we want exponential backoff it should happen in the launcher

ivov (approved):

PR #29363, opened 2026-04-28
feat: Validate workflow-sdk output topology against mode

mutdmour:
## Summary
- Adds declarative `builderHint.outputs` to node descriptions, mirroring the existing `builderHint.inputs` shape but keyed on `NodeConnectionType` so `main` is allowed.
- Adds `INVALID_OUTPUT_FOR_MODE` validator (severity `major`) that flags when a node has connections of a type its current parameters disable.
- Authors `outputs` for all 14 vector stores via `createVectorStoreNode` — `main` for insert/load/update, `ai_vectorStore` for retrieve, `ai_tool` for retrieve-as-tool.
- Surfaces `builderHint.outputs` through the Instance AI adapter so the agent sees the topology rules at prompt time.

## Why

A real AI-builder eval produced a workflow with `vectorStoreInMemory` in `mode: 'retrieve'` wired via `main` edges. In `retrieve` mode the node's only output port is `ai_vectorStore`; the `main` edges referenced ports that don't exist, so the downstream chain silently received no data. The runtime `outputs:` expression on the node already encoded the rule — it just wasn't declarative metadata anything could validate against.

This PR makes the rule first-class:
1. **Declarative** — `builderHint.outputs.main.displayOptions.show.mode` says "main is exposed only in insert/load/update".
2. **Validated** — the new validator walks each node's outgoing edges and emits a warning when the source mode disables that output.
3. **Visible to the agent** — the adapter passes the metadata through `SearchableNodeDescription`, alongside the existing inputs hint.

The validator reuses the existing `matchesDisplayOptions` + `buildConditionSummary` helpers — no new infrastructure.

## Test plan
- [x] 7 new tests in `validation.test.ts` covering retrieve / retrieve-as-tool / load / no-provider / expression / regression-fixture cases
- [x] Committed regression fixture `vector-store-retrieve-with-main.json` taken from a real failing eval workflow
- [x] Existing `createVectorStoreNode` snapshot updated (additive — only new `outputs` block)
- [x] Full `workflow-sdk` jest suite green (9531+ tests), full `ai-utilities` suite green (418 tests)
- [x] `pnpm typecheck` clean in `workflow`, `workflow-sdk`, `ai-utilities`, `instance-ai`


Cadiac (approved):
👍 seems like a smart addition, and implementation & tests seem solid

PR #29364, opened 2026-04-28
feat(core): Add --include and --exclude flags to import:credentials command

alielkhateeb:
## Summary

https://www.loom.com/share/bdc61e845f37490795bd25d5bb439267

Adds `--include` and `--exclude` flags to the `n8n import:credentials` CLI command, enabling field-level control during credential imports.

- `--include=<props>`: import only the listed comma-separated credential properties
- `--exclude=<props>`: import all properties except the listed ones
- The two flags are mutually exclusive — using both together throws an error
- Unknown property names in either flag emit a warning and are silently ignored
- An empty or whitespace-only list throws an error

**How to test:**
1. Export credentials to a JSON file
2. Run `n8n import:credentials --input=credentials.json --include=id,name,type,data` and verify `createdAt`/`updatedAt` are not imported
3. Run `n8n import:credentials --input=credentials.json --exclude=createdAt,updatedAt` and verify same behaviour
4. Run with both `--include` and `--exclude` and verify an error is thrown

https://linear.app/n8n/issue/LIGO-483

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/commands/import/credentials.ts
```diff
@@ -13,14 +13,41 @@ import { PROJECT_OWNER_ROLE_SLUG } from '@n8n/permissions';
...
+
 import { BaseCommand } from '../base-command';
 
-import { UM_FIX_INSTRUCTION } from '@/constants';
+type AllowedCredentialProperty = Exclude<
+	Extract<keyof CredentialsEntity, string>,
+	'shared' | 'toJSON' | 'generateId' | 'setUpdateDate'
+>;
+
+const allowedCredentialProperties: Record<AllowedCredentialProperty, true> = {
```
ireneea: This is only used in `isCredentialPropertyAllowed` is it worth moving it there rather than declaring it at the top

packages/cli/src/commands/import/credentials.ts
```diff
@@ -220,18 +275,82 @@ export class ImportCredentialsCommand extends BaseCommand<z.infer<typeof flagsSc
...
+				this.isCredentialPropertyAllowed(property),
+			);
+			return omit(credential, excludeProperties);
+		}
+
+		return credential;
+	}
+
+	private isCredentialPropertyAllowed(property: string): property is AllowedCredentialProperty {
+		return property in allowedCredentialProperties;
```
ireneea: [nit] Would this allow js object properties names such as `toString`. It would quite strange to have the json contains those js props, unless something goes wrong in the export 😅

alielkhateeb: This would only allow properties described here https://github.com/n8n-io/n8n/blob/5b466bf6d29b8a89dc36de32e4479d041f2fd5fc/packages/cli/src/commands/import/credentials.ts#L31-L43

And those are derived from the `CredentialsEntity` type here https://github.com/n8n-io/n8n/blob/5b466bf6d29b8a89dc36de32e4479d041f2fd5fc/packages/cli/src/commands/import/credentials.ts#L26-L29

alielkhateeb: And I just got what you mean 😂 interesting question, let me check and get back to you ...

alielkhateeb: @ireneea that is absolutely true, it would allow them.
Lets think about the problem we could face here, I tested with a JSON that has `"toString"` as a key in the end this is an import, so it ignores that key because the goal is to insert to credential entity, so if I import with `--include=toString` or with `exclude=--toString` it doesn't have an impact on the actual imported credential.

Do you see any scenario where someone might misuse this?

ireneea: Good point, in this context it doesn't have any effect

packages/cli/src/commands/import/credentials.ts
alielkhateeb: Yes good point, makes me wonder if it's even the best implementation, I'll try to think of a better way to encapsulate or define this

alielkhateeb: I also renamed "allowed" to "importable" IMO it's more explicit
ea70a17a22392f32d224521ecd86aa55f992fa0b

packages/cli/src/commands/import/credentials.ts
```diff
@@ -220,18 +261,98 @@ export class ImportCredentialsCommand extends BaseCommand<z.infer<typeof flagsSc
...
+				this.isCredentialPropertyImportable(property),
+			);
+			return pick(credential, includeProperties);
+		}
+
+		if (exclude?.length) {
+			const excludeProperties = exclude.filter((property) =>
+				this.isCredentialPropertyImportable(property),
+			);
+			return omit(credential, excludeProperties);
```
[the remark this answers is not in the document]
alielkhateeb: This is unnecessary because it means excluding non importable properties, which means import all importable properties :)
This is harmless

packages/cli/src/commands/import/credentials.ts
```diff
@@ -220,18 +261,98 @@ export class ImportCredentialsCommand extends BaseCommand<z.infer<typeof flagsSc
...
+	private filterCredentialProperties(
+		credential: Partial<CredentialsEntity>,
+		include?: string[],
+		exclude?: string[],
+	): Partial<CredentialsEntity> {
+		if (include?.length) {
+			const includeProperties = include.filter((property) =>
+				this.isCredentialPropertyImportable(property),
+			);
+			return pick(credential, includeProperties);
```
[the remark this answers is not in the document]
alielkhateeb: This is reasonable, better UX than NOT NULL constraint
Fixed in 98ee928287bd85ed4075602aec49c2f3ecdc7f1c

ireneea (approved):

PR #29371, opened 2026-04-28
fix(core): Make instance AI see workflow runtime error messages correctly (no-changelog)

Cadiac:
## Summary

We weren't stringifying errors correctly, which mean that the LLM would only see `[object Object]` from the actual error reasons like this:

image

and this would often lead to the death loop where the agent can't recover from when it tries to guess what the issue is.

This PR makes us stringify the errors so that the LLM sees the actual error message and can recover from them correctly:

image

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
```diff
@@ -2320,6 +2322,34 @@ export async function extractExecutionResult(
...
+ * prototypes are gone and `instanceof Error` would be false.
+ */
+export function formatExecutionError(error: ExecutionError): string {
+	const parts: string[] = [];
+	if (error.message) parts.push(error.message);
+	if (error.description && error.description !== error.message) {
+		parts.push(error.description);
+	}
+	if ('messages' in error && error.messages.length > 0) {
+		parts.push(`Details: ${error.messages.join(' | ')}`);
```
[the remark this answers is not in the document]
Cadiac: Are we worried about this? (looking for human opinion, I'm aware that these *could* contain something but we show these very same errors on UI anyways already)

aalises: `formatExecutionError()` now forwards `ExecutionError.messages` to the LLM even when `allowSendingParameterValues` is false. 

Those messages can maybe contain raw upstream API response bodies, so this bypasses the setting that suppresses execution data 😓 . 

We can gate `messages` behind `allowSendingParameterValues` or redact them before returning the error text?

not sure there, as we'd be losing some valuable data for the LLM?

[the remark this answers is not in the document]
aalises: Ah I posted something similar... I would definitely err on the side of caution here if we're to store this somewhere more permanent than a UX toast... but if it's for clarity of the LLM maybe I'd keep it? your call

Cadiac: 🤷 we can make it respect `allowSendingParameterValues`, but it will quite likely lead into unrecoverable errors if the agent can't see what's up

aalises (dismissed):

packages/cli/src/modules/instance-ai/instance-ai.adapter.service.ts
Cadiac: @aalises, I gated it behind `allowSendingParameterValues` after all at 28464758c53b08849d2d9320cbfebac9d91ebc48, not passing the details to LLM if set

aalises (approved):

PR #29377, opened 2026-04-28
fix(core): Reset Redis retry counter on successful reconnect

ivov:
## Summary

The Redis retry strategy state persisted across disconnect windows, so a disconnect was leaving behind `lastAttemptTs` and `cumulativeTimeout`, and the next disconnect within 30s was absorbing the entire intervening healthy interval as accumulated retry time, leading to tripping the threshold and exiting the process.

This PR resets each client's retry state on its own `'ready'` event. The reset is per-client, so one client recovering doesn't wipe another client's budget. It is also unconditional on the aggregate `lostConnection` flag, so a second client reconnecting after the first still resets its own state.

https://n8nio.slack.com/archives/C0789EN39RC/p1777363701100569

## Review / Merge checklist

- [x] Tests included.
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1`

### DIFF
--- a/packages/cli/src/services/__tests__/redis-client.service.test.ts
+++ b/packages/cli/src/services/__tests__/redis-client.service.test.ts
@@ -122,5 +122,130 @@ describe('RedisClientService', () => {
 			handlers.get('ready')!();
 			expect(service.isConnected()).toBe(true);
 		});
+
+		it("should not reset another client's retry state when one client recovers", () => {
+			const T0 = 1_700_000_000_000;
+			const dateNowSpy = jest.spyOn(Date, 'now').mockReturnValue(T0);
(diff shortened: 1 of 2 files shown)


Flexicon (approved):

mfsiega (commented):
I don't feel super strongly, but might it be worthwhile to introduce a small class to wrap the retry strategy? It feels like there's a bunch of data and methods getting passed around, it's a bit hard to follow.

ivov:
> I don't feel super strongly, but might it be worthwhile to introduce a small class to wrap the retry strategy? It feels like there's a bunch of data and methods getting passed around, it's a bit hard to follow.

Agree, I was on the fence. Follow-up: https://linear.app/n8n/issue/CAT-2979

PR #29383, opened 2026-04-28
chore(Brevo Node): Fix spelling

revolunet:
## Summary

just a spelling correction

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Brevo/EmailDescription.ts
```diff
@@ -126,8 +126,8 @@ const sendHtmlEmailFields: INodeProperties[] = [
 		},
 	},
 	{
-		displayName: 'Receipients',
-		name: 'receipients',
+		displayName: 'Recipients',
+		name: 'recipients',
```
DawidMyslak: Thanks for fixing the typo! However, changing the `name` property (not just `displayName`) will break existing saved workflows. The `name` is the key used to persist parameter values in workflow JSON. Existing workflows still store `"receipients"` as the key, so after this change they'd silently lose their recipient configuration. For v1, we can safely fix the `displayName` only and keep the `name` as is. If you'd like to also fix the `name` properties, we can introduce a v2 of the node with the corrected names, as long as the original v1 logic stays in place so existing workflows keep working.

packages/nodes-base/nodes/Brevo/EmailDescription.ts
```diff
@@ -186,18 +186,18 @@ const sendHtmlEmailFields: INodeProperties[] = [
 				},
 			},
 			{
-				displayName: 'Receipients BCC',
-				name: 'receipientsBCC',
+				displayName: 'Recipients BCC',
+				name: 'recipientsBCC',
```
DawidMyslak: Same concern here, renaming `name: 'receipientsBCC'` will break existing workflows that have BCC configured. The `displayName` fix is okay though.

packages/nodes-base/nodes/Brevo/EmailDescription.ts
```diff
@@ -212,18 +212,18 @@ const sendHtmlEmailFields: INodeProperties[] = [
 				},
 			},
 			{
-				displayName: 'Receipients CC',
-				name: 'receipientsCC',
+				displayName: 'Recipients CC',
+				name: 'recipientsCC',
```
DawidMyslak: And same for `receipientsCC` to `recipientsCC`. The `displayName` change is safe, but the `name` change would need a version bump.

packages/nodes-base/nodes/Brevo/EmailDescription.ts
```diff
@@ -327,8 +327,8 @@ const sendHtmlTemplateEmailFields: INodeProperties[] = [
 		},
 	},
 	{
-		displayName: 'Receipients',
-		name: 'receipients',
+		displayName: 'Recipients',
+		name: 'recipients',
```
DawidMyslak: This is the template email variant, same issue applies. Saved workflows using "Send HTML Template Email" would lose their recipients too.

packages/nodes-base/nodes/Brevo/GenericFunctions.ts
```diff
@@ -189,7 +189,7 @@ export namespace BrevoNode {
 			requestOptions: IHttpRequestOptions,
 		): Promise<IHttpRequestOptions> {
 			const ccData = this.getNodeParameter(
-				'additionalFields.receipientsCC.receipientCc',
+				'additionalFields.recipientsCC.recipientCc',
```
DawidMyslak: These `getNodeParameter` paths need to match what's stored in the workflow JSON, so changing them without a version bump would cause runtime errors for existing workflows.

DawidMyslak (changes requested):
Hey @revolunet, thanks for the contribution and for catching these typos! The spelling fix is definitely welcome. The main concern is that renaming the `name` fields (as opposed to `displayName`) is a breaking change for anyone with existing Brevo email workflows. The simplest path forward would be to only fix the `displayName` values in this PR, which gives users the corrected labels without any risk. If you'd also like to correct the `name` properties, we'd need to introduce a v2 of the node while keeping v1 intact. Happy to help guide you through that if you're interested!

Please take a look on our **[node versioning guidelines](https://docs.n8n.io/integrations/creating-nodes/build/reference/node-versioning/)**.

revolunet:
Ok thanks; i restored the `name` fields and only changed the `displayName` and validate function name

packages/nodes-base/nodes/Brevo/GenericFunctions.ts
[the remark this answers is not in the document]
revolunet: this looks outdated ?

DawidMyslak (approved):
Thanks for your contribution 👍

PR #29392, opened 2026-04-28
refactor(editor): Scope NDV store per workflow document (no-changelog)

alexgrozav:
## Summary

Convert the NDV store from a singleton into a per-workflow-document instance keyed by workflow document ID, mirroring the existing workflow document store pattern. The current NDV store is provided via injection (`NDVStoreKey`) and disposed alongside the workflow document store on cleanup, ensuring NDV state (active node, panel dimensions, drag/hover state, etc.) is isolated between workflows opened in the same session.

To test: open a workflow, interact with NDV (open a node, switch panels), navigate to another workflow and confirm NDV state does not leak between them. Verify demo and embedded (postMessage) flows still initialize and dispose correctly.

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
@@ -86,9 +86,8 @@ export function useWorkflowInitialization(workflowState: WorkflowState) {
 			disposeNDVStore(storeId);
 			disposeWorkflowDocumentStore(storeId);
 			currentWorkflowDocumentStore.value = null;
+			currentNDVStore.value = null;
 		}
-
-		currentNDVStore.value = null;
 	}
 
 	const isNewWorkflowRoute = computed(() => route.query.new === 'true');
--- a/packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
@@ -1,4 +1,4 @@
-import { ref, computed } from 'vue';
+import { ref, computed, shallowRef } from 'vue';
 import { type RouteRecordNameGeneric, useRoute, useRouter } from 'vue-router';
 import { useI18n } from '@n8n/i18n';
 import { useToast } from '@/app/composables/useToast';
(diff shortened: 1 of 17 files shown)


autologie:
This PR struck me with a bit of surprise 😄

I'm not sure if I understand the intent behind this – do we ever want multiple NDVs at the same time in UI (*)? If

> disposed alongside the workflow document store on cleanup, ensuring NDV state (active node, panel dimensions, drag/hover state, etc.) is isolated between workflows opened in the same session.

is the goal, we can also call `ndvStore.$reset()` on the global singleton NDV store when disposing workflow document store.

----

*I had this use case in the canvas experiment project, but that was more like "one workflow, multiple NDVs"

packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
```diff
@@ -81,8 +83,10 @@ export function useWorkflowInitialization(workflowState: WorkflowState) {
 				currentWorkflowDocumentStore.value.workflowId,
 				currentWorkflowDocumentStore.value.workflowVersion,
 			);
+			disposeNDVStore(storeId);
```
autologie: Can we call `currentNDVStore.value.$dispose()` directly here, so we don't need to rely on Pinia's internal API?

packages/frontend/editor-ui/src/features/ndv/shared/ndv.store.ts
```diff
@@ -18,436 +18,484 @@ import {
...
+}
+
+type Writable<T> = { -readonly [Key in keyof T]: T[Key] };
+
+export type NDVStore = Writable<ReturnType<ReturnType<typeof defineNDVStore>>>;
+
+export function useNDVStore(id: NDVStoreId): NDVStore;
+export function useNDVStore(pinia: Pinia): NDVStore;
+export function useNDVStore(): NDVStore;
+export function useNDVStore(idOrPinia?: NDVStoreId | Pinia): NDVStore {
```
autologie, line 475: Smart approach for a step-by-step migration :)

packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
alexgrozav: Updated the dispose function so that it takes in the store itself. We still need to remove the store from Pinia's store list, but now the API is cleaner.

autologie (approved):

PR #29396, opened 2026-04-28
fix(editor): Keep publish actions menu enabled for published workflows

heymynameisrob:
### Summary

The version dropdown was being disabled when all menu items were disabled, which caused the `Unpublish` option to be greyed out for published workflows with no pending changes.

- Update `shouldDisableActionDropdown` in `WorkflowHeaderDraftPublishActions.vue` to return `false` when an `activeVersion` exists so the chevron activator remains enabled.
- Add a regression test in `WorkflowHeaderDraftPublishActions.test.ts` that asserts the version menu stays enabled for a published workflow with no changes when `workflowPermissions.unpublish` is `false`.
- Ensure the `Unpublish` menu item remains present and disabled while the dropdown activator itself stays enabled.

| Before | After |
|:---:|:---:|
|  |  |

[ADO-5144](https://linear.app/n8n/issue/ADO-5144/impossible-to-unpublish-a-workflow-unless-you-change-it)

## Review / Merge checklist

- [ ] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.test.ts
+++ b/packages/frontend/editor-ui/src/app/components/MainHeader/WorkflowHeaderDraftPublishActions.test.ts
@@ -509,6 +509,29 @@ describe('WorkflowHeaderDraftPublishActions', () => {
 			expect(getByTestId('version-menu-button')).not.toBeDisabled();
 		});
 
+		it('should keep the version menu enabled when workflow is published with no changes and unpublish is unavailable', () => {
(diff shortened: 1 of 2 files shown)


konstantintieber (approved):

CharlieKolb (approved):
I'd also be up for always keeping the dropdown enabled if you prefer that

heymynameisrob:
> I'd also be up for always keeping the dropdown enabled if you prefer that

There's a small edge case where you can't unpublish OR publish and in that case it should be disabled as both it's options are disabled.

PR #29403, opened 2026-04-28
feat(core): Add configurable event log path per process

afitzek:
## Summary

Adds a new opt-in env var `N8N_EVENTBUS_LOGWRITER_LOGFULLPATH` that lets operators route the event log to a specific absolute path — used verbatim, with no process-type suffix appended. When the env is unset, behavior is byte-identical to today (default path under `${N8N_USER_FOLDER}` with the existing `-worker` / `-webhook-processor` / no-suffix derivation).

## Why

In multi-pod deployments backed by a shared RWX volume, every worker pod resolves to the same `${N8N_USER_FOLDER}/n8nEventLog-worker.log` because the `-worker` suffix discards the per-pod `hostId`. Concurrent unguarded `appendFileSync` writes from N pods to that single file produce interleaved bytes and rotation races, which then destabilizes startup recovery.

The companion read-side fix (bounded skip-and-warn parser + total-messages guard) shipped earlier. This PR addresses the write-side root cause: give operators a knob to put each pod on its own durable per-pod path (typically templated via the Kubernetes Downward API or a per-pod PVC). The operator owns per-pod uniqueness — n8n does not auto-suffix when the env is set.

## What changed

- New config field `LogWriterConfig.logFullPath` (env: `N8N_EVENTBUS_LOGWRITER_LOGFULLPATH`), default empty
- New pure helper `resolveEventLogPath()` co-located with the writer — handles validation, parent-dir auto-create, and legacy fallback in one place
- `MessageEventBusLogWriter.getInstance` accepts a new `resolvedPath` option that takes precedence over the legacy `logBasePath`/`logBaseName` derivation
- `MessageEventBus.initialize` calls the helper, threads the resolved path into the writer, and emits an advisory `logger.warn` when the operator opts into a configured path while files at the default location still exist (non-blocking)
- Worker thread (`message-event-bus-log-writer-worker.ts`) untouched — it already derives all rotation siblings (`.log-N`) and the `.recoveryInProgress` marker purely from `logFullBasePath`

## How to test manually

1. **Configured path is honored**
   ```bash
   N8N_EVENTBUS_LOGWRITER_LOGFULLPATH=/tmp/n8n-events/custom.log n8n start
   ```
   Verify writes land at `/tmp/n8n-events/custom.log`. Trigger a few rotations (lower `N8N_EVENTBUS_LOGWRITER_MAXFILESIZEINKB` to e.g. `1`) and confirm siblings appear at `/tmp/n8n-events/custom-1.log`, `custom-2.log`, …, with no `-worker` suffix anywhere.

2. **Default behavior unchanged when env unset**
   Start a worker without the env. Confirm it still writes to `${N8N_USER_FOLDER}/n8nEventLog-worker.log` (existing behavior).

3. **Validation**
   - `N8N_EVENTBUS_LOGWRITER_LOGFULLPATH=relative/path.log` → startup fails with a clear message about the path needing to be absolute
   - `N8N_EVENTBUS_LOGWRITER_LOGFULLPATH=/tmp/foo.txt` → startup fails about needing `.log` extension

4. **Parent dir auto-create**
   Point the env at a path whose parent does not exist (e.g. `/tmp/does/not/exist/yet/events.log`). Start a worker; confirm the parent dir is created and the file is written.

5. **Default-location coexistence advisory**
   Pre-create `${N8N_USER_FOLDER}/n8nEventLog-worker.log` (e.g. `touch`), then start a worker with `N8N_EVENTBUS_LOGWRITER_LOGFULLPATH=/tmp/elsewhere/events.log`. Confirm a single `WARN` log lists the stale default-location file.

## Key implementation decisions

- **Pure helper, service-layer validation.** `resolveEventLogPath` is a synchronous pure function with one filesystem side-effect (`mkdirSync({ recursive: true })`). Validation throws `UserError` from inside the helper rather than from a config decorator, matching the n8n convention (config decorators stay declarative; service `initialize()` is the validation point).
- **Verbatim when set, no auto-suffix.** When the env is set, the configured path is used as-is for main, worker, and webhook processes alike. The operator templates uniqueness (Downward API `$(POD_NAME)`, per-pod PVC `subPathExpr`, etc.). Auto-suffixing would fight that templating.
- **Worker thread untouched.** The corruption isn't in the worker code — it's in the path collision. Threading a unique `logFullBasePath` through to the worker resolves it without modifying any of the rotation / append / recovery-marker logic.
- **Default-location coexistence is advisory, not destructive.** Lists stale files in a single `WARN` log; never deletes anything. Operators handle drainage as part of their migration.

## Related Links

- Linear: https://linear.app/n8n/issue/IAM-579

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/@n8n/config/src/configs/event-bus.config.ts
+++ b/packages/@n8n/config/src/configs/event-bus.config.ts
@@ -20,8 +20,9 @@ class LogWriterConfig {
 	 * Absolute path to the primary event log file (must end in `.log`). When set,
 	 * used verbatim with no process-type suffix; the operator owns per-pod
 	 * uniqueness (e.g. via Kubernetes Downward API or per-pod PVC). Parent
-	 * directory is auto-created. Rotation siblings (`.log-1`, `.log-2`, …) and
-	 * the `.recoveryInProgress` marker colocate with this path. Empty (default)
+	 * directory is auto-created. Rotation siblings (e.g. `myEventLog-1.log`,
+	 * `myEventLog-2.log`, …, derived from the configured `logFullPath`) and the
+	 * `.recoveryInProgress` marker colocate with this path. Empty (default)
 	 * preserves the legacy `${N8N_USER_FOLDER}/n8nEventLog[-worker|-webhook-processor]`
 	 * behavior.
 	 */
--- a/packages/@n8n/config/src/configs/event-bus.config.ts
+++ b/packages/@n8n/config/src/configs/event-bus.config.ts
@@ -17,6 +17,18 @@ class LogWriterConfig {
 	logBaseName: string = 'n8nEventLog';
 
 	/**
+	 * Absolute path to the primary event log file (must end in `.log`). When set,
+	 * used verbatim with no process-type suffix; the operator owns per-pod
+	 * uniqueness (e.g. via Kubernetes Downward API or per-pod PVC). Parent
+	 * directory is auto-created. Rotation siblings (`.log-1`, `.log-2`, …) and
+	 * the `.recoveryInProgress` marker colocate with this path. Empty (default)
+	 * preserves the legacy `${N8N_USER_FOLDER}/n8nEventLog[-worker|-webhook-processor]`
+	 * behavior.
+	 */
+	@Env('N8N_EVENTBUS_LOGWRITER_LOGFULLPATH')
+	logFullPath: string = '';
+
+	/**
 	 * Safety tripwire: per-file cap on concurrently unconfirmed messages held in memory
 	 * during startup log parsing. Aborts the file if exceeded, to prevent OOM on legacy
 	 * logs with many orphaned messages. Tune up if healthy workloads hit false positives.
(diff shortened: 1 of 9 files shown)


packages/@n8n/config/src/configs/event-bus.config.ts
```diff
@@ -16,6 +16,18 @@ class LogWriterConfig {
 	@Env('N8N_EVENTBUS_LOGWRITER_LOGBASENAME')
 	logBaseName: string = 'n8nEventLog';
 
+	/**
+	 * Absolute path to the primary event log file (must end in `.log`). When set,
+	 * used verbatim with no process-type suffix; the operator owns per-pod
+	 * uniqueness (e.g. via Kubernetes Downward API or per-pod PVC). Parent
+	 * directory is auto-created. Rotation siblings (`.log-1`, `.log-2`, …) and
```
guillaumejacquart: i think the rotation sibling pattern is wrong. Should be "Rotation siblings (n8nEventLog-1.log, n8nEventLog-2.log, …)"

packages/cli/src/eventbus/message-event-bus/message-event-bus.ts
```diff
@@ -200,6 +215,34 @@ export class MessageEventBus extends EventEmitter {
 		this.isInitialized = true;
 	}
 
+	private warnIfDefaultLocationEventLogsCoexist(processType: EventLogProcessType): void {
+		const suffix =
```
guillaumejacquart: This suffix computing is duplicated in resolve-event-log-path.ts
Maybe we could call `resolveEventLogPath ` here to get the defaultBase ?

phyllis-noester (approved):

PR #29405, opened 2026-04-28
fix(core): Stop logging password reset token values

etairl:
## Summary

Drop the raw JWT from debug log entries in `AuthService.resolvePasswordResetToken`. Previously, the token was attached to the log context in three branches (expired, malformed, user-not-found), which surfaced live reset credentials to anyone with read access to the log pipeline (SRE, log shipping, SIEM, support tooling) for the duration of the token's TTL (default 20m).

The fix removes `token` from the structured log context in those three call sites. The `userId` is still logged for the user-not-found branch so the entry remains actionable for debugging.

### Changes

- `packages/cli/src/auth/auth.service.ts` — drop `token` from three `logger.debug` calls in `resolvePasswordResetToken`.
- `packages/cli/src/auth/__tests__/auth.service.test.ts` — add three regression tests, one per branch, asserting that no captured log call contains the raw token.

N/A — reported via independent review.

## Review / Merge checklist

- [x] Tests included.


garritfra (approved):
Good catch. Thanks @etairl!

PR #29408, opened 2026-04-28
fix(editor): Stabilize Instance AI workflow preview rendering (no-changelog)

r00gm:
## Summary

Reworks the Instance AI workflow preview to paint live execution state on the canvas using the SDK push events that already arrive via the n8n WebSocket — no more `openExecution` fetch, no more loader, and the build agent's verification runs are now visible too.

**The original problem:** the iframe was fetching execution data via `/rest/executions/{id}` after the agent's run-workflow tool completed. That produced a visible loader, and on fast runs the loader showed up after the execution had already finished. Worse, the build agent's internal verification runs never reached the canvas at all because the backend wasn't routing their push events to the user's session.

**What changed**

- **Frontend simplification.** The iframe stays in workflow mode forever. `useExecutionPushEvents` captures `executionStarted` / `nodeExecuteAfter` / `executionFinished` from the existing push connection; `useEventRelay` forwards them to the iframe via `relayPushEvent`, which paints node statuses on the canvas as the run progresses. Removed `activeExecutionId`, `latestExecution` watcher, historical-execution restore, polling in `InstanceAiWorkflowPreview`, and the `executionStatus` field on tabs.
- **Build-phase visibility.** Build agent runs the workflow before its tab is active. `useEventRelay` now keeps the buffered event log alive when an inactive workflow's run finishes, and replays it (plus a synthetic `executionFinished`, gated by `finishedSynthSent` so re-switching doesn't double-fire) when the user activates the tab.
- **Backend pushRef wiring (`packages/cli/src/modules/instance-ai/instance-ai.service.ts`).** Two fixes that were both needed for the build agent's runs to reach the user's session:
  - Pass `this.threadPushRef.get(threadId)` to `createExecutionEnvironment` in `doSchedulePlannedTasks` so build-agent and checkpoint executions are tagged with the user's iframe pushRef.
  - Stop deleting `threadPushRef` in the `finally` blocks of `executeRun` and `processResumedStream`. Planned tasks dispatch later in the same finally and depend on the entry being alive. `clearThreadState` still removes it on thread teardown, and `startRun` overwrites on each new chat send so a refreshed iframe pushRef is picked up immediately.

**Tests**

- New FE integration test (`bufferedEventsTabSwitch.test.ts`) drives the build-phase scenario end-to-end: events arrive while a different tab is active → user switches → events replay onto the iframe with the synthetic `executionFinished`.
- New BE regression test (`instance-ai.service.threadPushRef.test.ts`) source-asserts that neither `executeRun` nor `processResumedStream` re-introduces `threadPushRef.delete`, and verifies `clearThreadState` still does the legitimate cleanup.
- Pruned tests targeting the deleted execution-fetch flows in `useCanvasPreview.test.ts` and `InstanceAiWorkflowPreview.test.ts`. Removed the integration test file that exercised the same removed flows.

**How to test**

1. Open Instance AI, send a build prompt (e.g., \`Build a workflow that fetches the weather and sends a Slack message\`). Approve the plan.
2. While the build agent works, watch the canvas. As the verification run executes, node statuses should paint live (running → success/error per node) without any loader appearing.
3. Open another artifact tab during the build, then come back to the workflow tab — the run that happened while you were away should be reflected on the canvas.
4. Tell the chat to run the workflow again. The iframe should switch back to workflow mode with live painting; no openExecution fetch, no loader.

## Review / Merge checklist

- [ ] Tests included.


packages/frontend/editor-ui/src/features/ai/instanceAi/useCanvasPreview.ts
```diff
@@ -344,7 +344,17 @@ export function useCanvasPreview({ store, route, workflowExecutions }: UseCanvas
...
-			activeExecutionId.value = exec.executionId;
+			// While push events are still streaming this execution, leave the iframe
+			// in workflow mode so `relayPushEvent` keeps painting the live state —
+			// flipping to execution mode would refetch and discard it. Once the
+			// execution finishes (status leaves 'running'), set activeExecutionId so
+			// the iframe loads the persisted final data.
+			const liveExec = workflowExecutions?.value.get(exec.workflowId);
+			const isLiveRunning =
+				liveExec?.executionId === exec.executionId && liveExec?.status === 'running';
+			if (!isLiveRunning) {
```
aalises: This only runs when `executionId` changes iirc. If we skip setting `activeExecutionId` while the live execution is `running`, the later `running -> success/error` update will not rerun this code, so the iframe can stay in live workflow mode instead of showing final execution results 😢

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiWorkflowPreview.vue
```diff
@@ -135,13 +133,15 @@ async function pollExecutionUntilDone(executionId: string) {
 
 		const isFinished = execution?.finished === true;
 		if (isFinished) {
-			// Tell the iframe to re-load the now-complete execution data
-			const preview = previewRef.value as
-				| { reloadExecution?: () => void; iframeRef?: HTMLIFrameElement | null }
-				| undefined;
-			preview?.reloadExecution?.();
+			if (observedRunning) {
```
aalises: This can miss fast executions. `openExecution` may load while the run is still in progress, then the first poll sees `finished=true` and skips reload because `observedRunning` is false. The iframe can be left showing stale running state.

packages/frontend/editor-ui/src/features/ai/instanceAi/useEventRelay.ts
```diff
@@ -108,17 +113,56 @@ export function useEventRelay({
...
+		relayedCount.set(wfId, log.length);
+
+		// If the run already finished while inactive, the eventLog has the
+		// per-node events but useExecutionPushEvents doesn't store
+		// `executionFinished` in the log — it just updates status. Send a
+		// synthetic one so the iframe clears its executing-node queue.
+		// Gated by `finishedSynthSent` so re-loading the same workflow
+		// doesn't double-fire it.
+		if (current.status !== 'running' && !finishedSynthSent.has(current.executionId)) {
+			finishedSynthSent.add(current.executionId);
```
aalises, line 155: `handleWorkflowLoaded()` marks the synthetic `executionFinished` as sent before confirming the iframe is ready. If that postMessage is dropped during cold boot, the execution queue may never clear. Maybe can gate this on iframe readiness or re-synthesize on `n8nReady`?

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiWorkflowPreview.vue
```diff
@@ -72,6 +69,13 @@ async function fetchWorkflow(id: string) {
 		const result = await workflowsListStore.fetchWorkflow(id);
 		if (generation !== fetchGeneration) return;
 		workflow.value = result;
+		// Wait for Vue to propagate the new workflow to <WorkflowPreview>'s
+		// reactive watcher, which posts `openWorkflow` to the iframe. Emitting
+		// after this tick lets parents replay buffered execution events
+		// without racing the workflow load.
+		await nextTick();
+		if (generation !== fetchGeneration) return;
+		emit('workflow-loaded', id);
```
aalises, line 78: `workflow-loaded` currently fires after the parent posts `openWorkflow`, but before the iframe has necessarily finished importing it. Can we move this to an iframe-side “workflow imported” ack so buffered execution events replay only after the canvas is ready? not sure how big of a problem that'd be tho

packages/frontend/editor-ui/src/features/ai/instanceAi/useEventRelay.ts
r00gm: Valid race in theory. Defending the current shape for now: the iframe is mounted eagerly with InstanceAiView, so by the time push events arrive for an active workflow it’s effectively always past n8nReady. Adding an iframeReady gate here works but bakes defensive plumbing into the relay for a window we don’t see in practice.

If we do see “spinner stuck on the canvas after a finished run” reports, the cleaner fix is on the backend side: gate the agent’s execute-workflow tool on a frontend→backend iframe-ready signal so we never dispatch into a not-yet-ready canvas. Happy to revisit then.

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiWorkflowPreview.vue
r00gm: Same shape of concern as the other thread, and the same answer: in practice the iframe finishes importing the workflow before any execution events for it land, because openWorkflow and the build-agent’s verify run are decoupled in time (user reads the artifact, then asks to run). An iframe-side “workflow imported” ack would be the right contract long-term, but it’s a WorkflowPreview change that ripples beyond this PR.

Leaving as-is and will reach for the ack (or the backend-gates-on-ready approach) if we see drops.

aalises (dismissed):

aalises (approved):

PR #29418, opened 2026-04-28
feat(core): Add preAuthentication support to requestOAuth2 pipeline

guillaumejacquart:
## Summary

The `requestOAuth2` pipeline previously bypassed `preAuthentication` entirely, which prevented custom credential types extending `oAuth2Api` (e.g. for OIDC/JWE/JWT-claim extraction) from transforming token data before signing. This PR wires `preAuthentication` into both the initial request path and the centralized `refreshOrFetchToken` helper, so the retry-with-new-token path and the standalone `refreshOAuth2Token` helper inherit the transformation. `credentialsExpired=true` is passed to bypass the expirable-property guard, since OAuth2 credentials need transformation on every request. Token construction is extracted into a `buildSigningToken` helper to remove duplication across the three call sites. Behavior is unchanged for credentials that do not declare `preAuthentication`.

## How to test

1. Unit tests: `pnpm --filter n8n-core test request-helper-functions.test.ts` — covers initial path, refresh-via-retry, standalone `refreshOAuth2Token`, and the no-op path when `preAuthentication` returns `undefined`.

2. Manual: create a custom OAuth2 credential extending `oAuth2Api` with a `preAuthentication` hook so you can observe the transformation end-to-end:

   ```ts
   import type {
     ICredentialType,
     INodeProperties,
     IAuthenticateGeneric,
     ICredentialDataDecryptedObject,
     IHttpRequestHelper,
   } from 'n8n-workflow';

   export class TestPreAuthOAuth2Api implements ICredentialType {
     name = 'testPreAuthOAuth2Api';
     displayName = 'Test PreAuth OAuth2 API';
     extends = ['oAuth2Api'];

     properties: INodeProperties[] = [
       {
         displayName: 'Grant Type',
         name: 'grantType',
         type: 'hidden',
         default: 'authorizationCode',
       },
       // Required: an expirable hidden property is what enables preAuthentication
       {
         displayName: 'Session Token',
         name: 'sessionToken',
         type: 'hidden',
         typeOptions: { expirable: true },
         default: '',
       },
     ];

     async preAuthentication(this: IHttpRequestHelper, credentials: ICredentialDataDecryptedObject) {
       const tokenData = (credentials.oauthTokenData ?? {}) as { access_token?: string };
       // Transform the token however you need (e.g. decrypt JWE, extract a claim, etc.)
       const transformed = `transformed-${tokenData.access_token ?? ''}`;
       return {
         sessionToken: transformed,
         oauthTokenData: { ...tokenData, access_token: transformed },
       };
     }
   }
   ```

   Connect the credential, then run an HTTP Request node (or any node using `requestOAuth2`) and verify the outgoing `Authorization` header carries the transformed token. Force a refresh (expire the token) and confirm the retried request also uses the transformed value and that the persisted credential reflects it.

https://linear.app/n8n/issue/IAM-570

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
+++ b/packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
@@ -729,6 +729,23 @@ function createOAuth2Client(credentials: OAuth2CredentialData): ClientOAuth2 {
 	});
 }
 
+function buildSigningToken(
+	client: ClientOAuth2,
+	tokenData: ClientOAuth2TokenData,
+	oAuth2Options?: IOAuth2Options,
+): ClientOAuth2Token {
+	const accessToken = get(tokenData, oAuth2Options?.property as string) || tokenData.accessToken;
+	const refreshToken = tokenData.refreshToken;
+	return client.createToken(
+		{
+			...tokenData,
+			...(accessToken ? { access_token: accessToken } : {}),
+			...(refreshToken ? { refresh_token: refreshToken } : {}),
+		},
+		oAuth2Options?.tokenType || tokenData.tokenType,
+	);
+}
+
 interface RefreshOAuth2TokenContext {
 	credentials: OAuth2CredentialData;
 	token: ClientOAuth2Token;
@@ -737,10 +754,20 @@ interface RefreshOAuth2TokenContext {
 	additionalData: IWorkflowExecuteAdditionalData;
 	oAuth2Options?: IOAuth2Options;
 	logger: WorkflowLogger;
+	helpers: IAllExecuteFunctions['helpers'];
 }
 
 async function refreshOrFetchToken(ctx: RefreshOAuth2TokenContext): Promise<ClientOAuth2Token> {
-	const { credentials, token, credentialsType, node, additionalData, oAuth2Options, logger } = ctx;
+	const {
+		credentials,
+		token,
+		credentialsType,
+		node,
+		additionalData,
+		oAuth2Options,
+		logger,
+		helpers,
+	} = ctx;
 	const tokenRefreshOptions: IDataObject = {};
 	if (oAuth2Options?.includeCredentialsOnRefreshOnBody) {
 		const body: IDataObject = {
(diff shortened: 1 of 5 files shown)


packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
```diff
@@ -769,6 +796,28 @@ async function refreshOrFetchToken(ctx: RefreshOAuth2TokenContext): Promise<Clie
 	);
 
 	credentials.oauthTokenData = newToken.data;
+
+	// Re-run preAuthentication so custom credential types extending oAuth2Api can transform
+	// refreshed token data (e.g. extracting a claim from a decrypted JWE/JWT) before signing.
+	// credentialsExpired=true bypasses the expirable-property guard since the transformation
+	// must run on every refresh, not only when the cached value is empty.
+	const preAuthData = await additionalData.credentialsHelper.preAuthentication(
```
afitzek: Would this overwrite credential data on the actual credential, when a dynamic credential is updated? I think there is an update function of the credential in the database in the preAuthentication function.

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
```diff
@@ -856,17 +905,23 @@ export async function requestOAuth2(
...
-			...oauthTokenData,
-			...(accessToken ? { access_token: accessToken } : {}),
-			...(refreshToken ? { refresh_token: refreshToken } : {}),
-		},
-		oAuth2Options?.tokenType || oauthTokenData.tokenType,
+	// Support preAuthentication for custom OAuth2 credential types extending oAuth2Api.
+	// Enables transforming token data on every request (e.g. extracting a claim from a
+	// decrypted JWE/JWT). credentialsExpired=true bypasses the expirable-property guard,
+	// which would otherwise skip the call once the cached value is populated.
+	const preAuthData = await additionalData.credentialsHelper.preAuthentication(
```
afitzek: Would this result in a potential DB write on every web request that is done with this credential?

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
guillaumejacquart: Good point. I introduced a new method on the helper to just invoke the preAuthenticate hook

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
guillaumejacquart: good point. same fix as previous comment

afitzek (approved):
Looks good 👍

PR #29421, opened 2026-04-28
feat(agents): Rework execution logging UX for agent sessions

mike12345567:
## Summary

Rebuild of the agent session view covering the timeline, filter controls, detail panel, and end-to-end tagging of tool calls.

- **Timeline chart** — proportional bar at the top of the session with fixed-width idle blocks for gaps over 10 min. Hover popovers, click to select and populate the detail panel.
- **Sub-header** — trigger source (slack/chat icon), events filter (multi-select dropdown), free-text search.
- **Timeline table** — pill/info/time rows with idle rows interleaved chronologically.
- **Right detail panel** — only opens on selection; full-width table when closed. "Open full execution" button moved to the top of the panel.
- **Workflow log viewer** for workflow tool calls — reuses `LogsOverviewRow` for the node list and `RunData`/`NodeErrorView` for I/O.
- **Node tools** — `node-tool-factory` now tags built tools with `metadata: { kind: 'node', nodeType, nodeTypeVersion }`, propagated through `tool-registry`, `ExecutionRecorder`'s `TimelineEvent`, and frontend types. Renders with a "Node" pill and synthesized single-node `RunData` for I/O via the new `ToolIoView` component.
- **Custom/built-in tools** — pretty JSON view for non-node payloads. `rich_interaction` still uses `RichInteractionCard`; suspension rows show "Waiting for user response".

## Related

https://linear.app/n8n/issue/AGENT-36

## Screenshots
image

## Review / Merge checklist

- [x] Tests included.


packages/frontend/editor-ui/src/features/agents/session-timeline.utils.ts
```diff
@@ -0,0 +1,213 @@
...
+		if (userMsg) {
+			items.push({
+				kind: 'user',
+				executionId: exec.id,
+				content: userMsg,
+				timestamp: exec.startedAt ? new Date(exec.startedAt).getTime() : 0,
+			});
+		}
+
+		for (const event of parseTimeline(exec)) {
```
bjorger, line 158: Resumed HITL/approval tool executions can disappear here. On resume, the SDK emits a `tool-result` for the pending call before any matching `tool-call`; `ExecutionRecorder` stores that fallback only in flat `toolCalls`, not in `timeline`. Since this new timeline flattener only reads `parseTimeline(exec)`, approved workflow tools will not render a workflow row and the `workflowExecutionId` log link is lost. Please add a fallback from `toolCalls` when timeline has no corresponding tool-call row.

packages/frontend/editor-ui/src/features/agents/components/SessionDetailPanel.vue
```diff
@@ -0,0 +1,323 @@
...
+						size="small"
+						:label="i18n.baseText('agentSessions.workflowLog.openFull')"
+						data-test-id="open-full-execution"
+						@click="openFullExecution"
+					/>
+				</div>
+			</div>
+
+			<template v-if="item.kind === 'workflow'">
+				<WorkflowExecutionLogViewer
```
bjorger, line 162: This component is reused across workflow row selections without a `:key`, while `WorkflowExecutionLogViewer` fetches only on mount. Selecting a second workflow row can keep showing the first execution log. Please either key this by `item.workflowExecutionId` / `item.workflowId`, or make the child watch prop changes and refetch/reset state.

packages/frontend/editor-ui/src/features/agents/components/WorkflowExecutionLogViewer.vue
```diff
@@ -0,0 +1,356 @@
...
+		const result = await executionsStore.fetchExecution(props.workflowExecutionId);
+		if (!result) {
+			errorMessage.value = i18n.baseText('agentSessions.workflowLog.unavailable');
+		} else {
+			execution.value = result;
+			// Populate the workflow execution data into the store. RunDataTable,
+			// pairedItemMappings, and various NodeErrorView code paths read from the
+			// store rather than the prop, so the prop alone isn't enough to make the
+			// table/JSON views render correctly for non-trivial nodes.
+			workflowState.setWorkflowExecutionData(result);
```
bjorger, line 177: This async mount flow can still run after the component unmounts. If the user closes the panel or selects another row while `loadNodeTypesIfNotLoaded()` / `fetchExecution()` is pending, `onBeforeUnmount()` restores the previous execution data, then this line can overwrite the global workflow execution state with stale data. Please add an unmounted/cancel guard before setting local state and calling `setWorkflowExecutionData`, similar to `ToolIoView`.

bjorger (changes requested):
## Execution log:
In the execution timeline, the rendered step looks a bit funny:

Image

## Filtering
I like the filtering here, but I'm missing a "Tool" filter, that only shows tool calls:
Image

Right now I can see each individual tool

## Executions Table
Nitpicky probably but, it would be cool if the executions table can automatically refresh:

Image

Right now the user is forced to reload the page.

packages/frontend/editor-ui/src/features/agents/agentSessions.store.ts
```diff
@@ -54,6 +54,44 @@ export const useAgentSessionsStore = defineStore('agentSessions', () => {
...
+			);
+			if (latestKey !== key) return;
+			const seen = new Set(page.threads.map((t) => t.id));
+			const tail = threads.value.filter((t) => !seen.has(t.id));
+			threads.value = [...page.threads, ...tail];
+			// Only adopt the new cursor if we hadn't paginated yet — otherwise
+			// the existing cursor already points past what we've loaded.
+			if (tail.length === 0) {
+				nextCursor.value = page.nextCursor;
+			}
```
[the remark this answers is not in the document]
mike12345567: This is fine, the user has loaded the sessions, when they refresh the page any stale sessions will be removed, the refresh is just to check for new sessions.

bjorger (approved):

PR #29429, opened 2026-04-28
fix(core): Make MCP client registration cap tunable and surface a proper limit error

RicardoE105:
## Summary

This PR addresses ADO-5107 with a hotfix-style approach: instead of restructuring the quota to be per-user, the existing instance-wide cap is made configurable and the limit error is made meaningful. The original architectural fix (per-user list scoping + per-user quota) is split out for a follow-up PR.

The Connected Clients list under MCP settings was instance-wide with a hardcoded 200-client cap. Customers (e.g. Canva) hit it organically because Claude Code/Desktop register a new OAuth `client_id` on every reauth via DCR, and accumulated registrations across many users exhausted the shared pool. The DCR endpoint then returned a generic `500 server_error / Internal Server Error`, which made the failure look like an internal bug to third-party support.

This PR:

- Raises the default for `N8N_MCP_MAX_REGISTERED_CLIENTS` from 200 to 10000 and keeps it as a tunable env var, so admins can adjust to their fleet size.
- Returns a structured `503` with a descriptive `error_description` when the cap is hit (instead of the SDK's opaque `500`), and adds server-side warn logging so admins can see when this triggers.
- Fixes the `expires_in` field on the `authorization_code` token response to read from the configured access token lifetime instead of being hardcoded to `3600`. Latent bug; harmless today because both happen to equal 1h, but a footgun the moment the constant moves.
- Paginates the Connected Clients table client-side via `N8nDataTableServer` (related to ADO-5101).

- https://linear.app/n8n/issue/ADO-5107
- Triggering incident: https://linear.app/n8n/issue/ADO-5099
- Related (pagination): https://linear.app/n8n/issue/ADO-5101

## Review / Merge checklist

- [x] Tests included.
- [ ] PR labeled with \`Backport to Beta\`, \`Backport to Stable\`, or \`Backport to v1\` (if applicable).


packages/@n8n/config/src/configs/endpoints.config.ts
```diff
@@ -134,9 +134,17 @@ export class EndpointsConfig {
 	@Env('N8N_MCP_BUILDER_ENABLED')
 	mcpBuilderEnabled: boolean = true;
 
-	/** Maximum number of OAuth clients that can be registered for MCP. */
+	/**
+	 * Instance-wide ceiling on registered OAuth clients for MCP. This is an
+	 * anti-abuse safety net for the unauthenticated dynamic client registration
+	 * endpoint — the user-facing quota is `mcpMaxClientsPerUser`.
+	 */
 	@Env('N8N_MCP_MAX_REGISTERED_CLIENTS')
```
MiloradFilipovic, line 138: It feels a bit redundant to have both of these. If there are not technical blockers, I would suggest keeping only the old one (`N8N_MCP_MAX_REGISTERED_CLIENTS`) and modifying it so it applies per user.
I don't think it's a big hard since we are not documenting this either way, so if we update it and add it to the docs I think that's acceptable.

packages/cli/src/modules/mcp/mcp.auth.consent.controller.ts
```diff
@@ -66,6 +67,21 @@ export class McpConsentController {
 				},
 			});
 		} catch (error) {
+			if (error instanceof McpClientLimitReachedError) {
+				this.logger.info('Consent rejected: per-user MCP client limit reached', {
+					limit: error.limit,
+				});
+				res.status(409).json({
```
MiloradFilipovic: Maybe `403` (Forbidden) or `429` (Too Many Requests) are better options here?

packages/frontend/editor-ui/src/app/views/OAuthConsentView.vue
```diff
@@ -28,6 +29,13 @@ const handleAllow = async () => {
 		waitingForRedirect.value = true;
 		window.location.href = response.redirectUrl;
 	} catch (err) {
+		if (err instanceof ResponseError && (err.meta?.code as string) === 'mcp_client_limit_reached') {
```
MiloradFilipovic: Is the cast here necessary? 
```suggestion
		if (err instanceof ResponseError && err.meta?.code === 'mcp_client_limit_reached') {
```

MiloradFilipovic (commented):
Works nicely, I left a few comments so take a look at what makes sense.
Also, I was expecting a warning here in the UI at least (if not a banner) when quota is exhausted:

Image

packages/@n8n/config/src/configs/endpoints.config.ts
RicardoE105: Keeping both is intentional: the two limits guard different things and removing the instance-wide one would expose the unauthenticated DCR endpoint.

`N8N_MCP_MAX_REGISTERED_CLIENTS` caps the total `oauth_clients` rows. That limit runs at `POST /mcp-oauth/register`, which is the OAuth Dynamic Client Registration endpoint and is unauthenticated by design (no user identity is available there yet — the user only enters the picture later, at consent time). Without it, anyone reachable on that endpoint could spam-register clients and bloat the DB until the instance is unusable. That's why I bumped its default to 10000 and reframed it as an anti-abuse safety net.

`N8N_MCP_MAX_CLIENTS_PER_USER` is the user-facing quota, enforced at consent (where we have a user). That's the one whose number a user sees and whose 200 default we actually want to expose in docs.

So the two vars serve distinct roles: instance-wide DoS protection vs per-user quota. If we collapse them into one, either we lose the unauthenticated-DCR safety net, or we cap each user at 10k by default — both regressions. Happy to add docs for both.

RicardoE105:
@MiloradFilipovic 

CleanShot 2026-04-29 at 07 54 15@2x

Qodo-Free-For-OSS:
Hi, Concurrent DCR requests can pass the new pre-check guard and then hit enforceClientLimit(), which throws a generic Error; this can still surface as the SDK/controller returning a generic 500 instead of the intended structured 503.

**How to fix:** Handle limit overflow as 503

> ### Issue description
> The new `mcpClientLimitGuard` returns a structured 503 only for the pre-check path. Under concurrency, requests can still exceed the cap after insertion and hit `enforceClientLimit()`, which throws a generic `Error`. This likely bubbles up as a generic 500 from the MCP SDK/Express stack, defeating the purpose of surfacing a meaningful limit error.
> 
> ### Issue Context
> The service explicitly documents checking after insert to avoid the count/insert race. So the post-insert overflow path is expected and must return the same structured 503 response shape as the guard.
> 
> ### Fix Focus Areas
> - packages/cli/src/modules/mcp/mcp-oauth-service.ts[100-115]
> - packages/cli/src/modules/mcp/mcp.oauth.controller.ts[42-53]
> 
> ### Implementation directions
> - Introduce a dedicated error type (e.g. `McpClientLimitReachedError` carrying `limit` and maybe `clientCount`) thrown from `enforceClientLimit()`.
> - Ensure the DCR endpoint translates that error into the same `503 { error: 'server_error', error_description: ... }` response.
>   - Option A: Wrap `clientsStore.registerClient` (or the router handler) with try/catch and map only this error to 503.
>   - Option B: Move the structured response responsibility into a custom register route/controller instead of using the SDK handler directly, so you control error serialization.
> - Add an integration test that simulates the post-insert overflow path (e.g. stub `isClientLimitReached()` to return false while `enforceClientLimit()` triggers overflow) to ensure 503 is returned even when the guard doesn’t catch it.

---
*Spotted by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source projects.*

packages/@n8n/config/src/configs/endpoints.config.ts
```diff
@@ -136,7 +136,7 @@ export class EndpointsConfig {
 
 	/** Maximum number of OAuth clients that can be registered for MCP. */
 	@Env('N8N_MCP_MAX_REGISTERED_CLIENTS')
-	mcpMaxRegisteredClients: number = 200;
+	mcpMaxRegisteredClients: number = 5000;
```
MiloradFilipovic, line 139: PR description says this is `10000`. I actually would vote for less, so this is ok to leave but wanted to check if it makes sense to reduce to 1000?
Trying to figure out what's the best balance of keeping DDoS at bay and having a generous limit

packages/cli/src/modules/mcp/mcp-oauth-service.ts
```diff
@@ -91,17 +91,26 @@ export class McpOAuthService implements OAuthServerProvider {
 		};
 	}
 
+	/** Returns true when the instance is already at or above the registered-client cap. */
+	async isClientLimitReached(): Promise<boolean> {
+		const clientCount = await this.oauthClientRepository.count();
+		return clientCount >= this.globalConfig.endpoints.mcpMaxRegisteredClients;
```
MiloradFilipovic, line 98: Any reason this is using `>=` and `enforceClientLimit` is using `>` to check client limit?

MiloradFilipovic (dismissed):
Looks good, thanks for splitting these up 🙌
Left a few comments so take a look at what makes sense.
Also, does it makes sense to return warning notice in settings page (maybe only for admins) if limit is reached?

packages/@n8n/config/src/configs/endpoints.config.ts
RicardoE105: considering that i did not fix the issue with the reconnections create a new client ID, wanted a bit more than 1000. Also, creating 5000 records in the db when trying to DDoS should not be a problem

packages/cli/src/modules/mcp/mcp-oauth-service.ts
RicardoE105: - isClientLimitReached (pre-check): runs before insert. count == limit already means we're at capacity,
   so admitting one more would push us over. Use >= to reject at the boundary.                           
  - enforceClientLimit (post-insert): runs after the row is inserted, so count already includes this     
  request's row. The first legitimate row at the boundary should pass, only excess rows over the limit   
  should roll back. Use >.

RicardoE105:
> Also, does it makes sense to return warning notice in settings page (maybe only for admins) if limit is reached?

Thought about it but If I do it in the MCP connected client list it will confuse the user as they see a list with clients that does not correlate with the max limit reached. Maybe we do it on the MCP tab

packages/cli/src/modules/mcp/mcp-oauth-service.ts
RicardoE105: Intentional — the two checks observe different counts:

- `isClientLimitReached` runs **before** insert, so `count == limit` already means we are at capacity and admitting one more would push us over → `>=`.
- `enforceClientLimit` runs **after** insert (post-rollback path for the race), so `count` already includes this request's just-inserted row → `>`. Only excess rows over the limit should roll back; the first legitimate row at the boundary should pass.

Worked example with `limit = 1`:

| Step | Count in DB | Check | Outcome |
|---|---|---|---|
| Pre-check at start | 0 | `0 >= 1` → false | Pass |
| Insert | 1 | — | — |
| Post-insert check | 1 | `1 > 1` → false | Keep ✅ |
| Second request hits pre-check | 1 | `1 >= 1` → true | Reject 503 |

If we used `>=` on the post-check too, that first legitimate insert would roll itself back (since `1 >= 1` after it inserts). Pushed a doc comment in `mcp-oauth-service.ts` to capture this so it's not tripped over again.

MiloradFilipovic:
> > Also, does it makes sense to return warning notice in settings page (maybe only for admins) if limit is reached?
> 
> Thought about it but If I do it in the MCP connected client list it will confuse the user as they see a list with clients that does not correlate with the max limit reached. Maybe we do it on the MCP tab

Ah true, even admins will not be able to see all. Ok, let's keep it without it

MiloradFilipovic (dismissed):

MiloradFilipovic (approved):

PR #29438, opened 2026-04-29
feat(core): Add MCP tool to list credentials

RicardoE105:
## Summary

Adds a new read-only MCP tool, `list_credentials`, that returns credentials the current user can access. The tool never decrypts or returns secret data — only metadata needed to identify and pick a credential (id, name, type, scopes, isManaged, isGlobal, homeProject).

Filters: `query` (name partial match), `type`, `projectId`, `onlySharedWithMe`, `limit` (max 200). Permissions are enforced by `CredentialsService.getMany()` (the same path used by the REST list endpoint), so users only see credentials they're entitled to.

This unblocks MCP clients from listing existing credentials so they can be referenced from a workflow node by id.

https://linear.app/n8n/issue/ADO-5125

## Review / Merge checklist

- [x] Tests included.


packages/cli/src/modules/mcp/tools/list-credentials.tool.ts
```diff
@@ -0,0 +1,204 @@
...
+		includeScopes: true,
+		includeData: false,
+		includeGlobal: !onlySharedWithMe,
+		onlySharedWithMe,
+	});
+
+	const data: ListCredentialsItem[] = credentials.map((credential) => {
+		const c = credential as CredentialsEntity & {
+			scopes?: string[];
+			homeProject?: { id: string; name: string; type: string } | null;
```
MiloradFilipovic: Using this tool from MCP clients that validate output schemas (like OpenClaude) fails:

Image

My hunch is it's because `homeProject` contains `icon` which is not declared in the output schema and casting `credential as CredentialsEntity` just hides it from the compiler but it's not really removing it. We should use proper type or make sure that returned `homeObject` includes only `id`, `name` and `type` if that's what we want to serve.

packages/cli/src/modules/mcp/tools/list-credentials.tool.ts
```diff
@@ -0,0 +1,204 @@
...
+import { createLimitSchema } from './schemas';
+
+const MAX_RESULTS = 200;
+
+const inputSchema = {
+	limit: createLimitSchema(MAX_RESULTS),
+	query: z
+		.string()
+		.optional()
+		.describe('Filter credentials by name (case-insensitive partial match)'),
```
MiloradFilipovic: I think the `case-insensitive` part is not true for postgress (`LIKE` is [case-sensitive](https://www.postgresql.org/docs/7.3/functions-matching.html) there). I would vote for just removing this part of the description

packages/cli/src/modules/mcp/tools/list-credentials.tool.ts
```diff
@@ -0,0 +1,204 @@
...
+			telemetryPayload.results = {
+				success: false,
+				error: errorMessage,
+			};
+			telemetry.track(USER_CALLED_MCP_TOOL_EVENT, telemetryPayload);
+
+			const output: ListCredentialsResult & { error: string } = {
+				data: [],
+				count: 0,
+				error: errorMessage,
```
MiloradFilipovic, line 152: `error` is not declared in the output schema so this will aslo fail in clients like OpenCode that use strict output validation

packages/cli/src/modules/mcp/tools/list-credentials.tool.ts
```diff
@@ -0,0 +1,204 @@
...
+		.boolean()
+		.optional()
+		.describe('Only return credentials shared directly with the current user'),
+} satisfies z.ZodRawShape;
+
+const homeProjectSchema = z
+	.object({
+		id: z.string(),
+		name: z.string(),
+		type: z.string(),
```
MiloradFilipovic: Should we add some descriptions here (especially for `type`)?

MiloradFilipovic (commented):
Works well 🙌. I left a couple of comments that we should address so this works with with strict clients.

packages/cli/src/modules/mcp/tools/list-credentials.tool.ts
```diff
@@ -33,9 +30,9 @@ const inputSchema = {
 
 const homeProjectSchema = z
 	.object({
-		id: z.string(),
-		name: z.string(),
-		type: z.string(),
+		id: z.string().describe('The unique identifier of the project'),
+		name: z.string().describe('The name of the project'),
+		type: z.string().describe("The project type, e.g. 'personal' or 'team'"),
```
MiloradFilipovic: Nit: We should elaborate that 'team' means 'shared' in n8n. It's maybe obvious but good to mention so LLMs know what they are looking for.

MiloradFilipovic (dismissed):
Great! Thanks for addressing the feedback. I left a small nit but good to go without it also.

MiloradFilipovic (approved):
Thanks ❤️

PR #29444, opened 2026-04-29
fix(Google Sheets Node): Reduce duplicate API calls in append operation to avoid quota limits

michael-radency:
## Summary
The append operation was making 3 redundant GET requests per execution cycle (fetching the header row in `getData`, `autoMapInputData`, and` convertObjectArrayToSheetDataArray)`. With the 60 reads/minute quota, this caused 429 Too Many Requests errors after ~20 loop iterations.

Changes:

- Added `columnNamesHint` side-channel on `GoogleSheet` — set once by the caller, consumed and cleared by `convertObjectArrayToSheetDataArray` to skip its own fetch
- append.operation.ts now extracts the header row from the already-fetched `sheetData` and passes it down, reducing reads per append from 3 to 1
- `autoMapInputData` accepts an optional `prefetchedColumnNames` param to skip its internal `getData` call and stores the - final column list (including newly-added columns) via `setColumnNamesHint`
- Fixed a bug where newly-discovered columns were added to the sheet header but their values were silently dropped from the data row

https://linear.app/n8n/issue/NODE-1960/google-sheets-getting-too-many-requests-error-when-appending-to-a

## Review / Merge checklist

- [ ] Tests included. 


alexander-gekov (approved):
Tested and base case looks good.

Claude did find this:
`Stale hint when appendSheetData is called multiple times in one execution
In GoogleSheet.ts, the columnNamesHint is cleared after the first consumption by convertObjectArrayToSheetDataArray. However, in append.operation.ts the hint is set before calling appendSheetData, which iterates over data in batches. If appendSheetData is called more than once per execution (e.g. chunked batches), the second call won't have the hint and will fall back to an API call — partially defeating the optimization.`

But it seems to be an edge case

PR #29445, opened 2026-04-29
feat: Add deeplinkpairing and connection handling for native computer use (no-changelog)

BerniWittmann:
## Summary

This branch delivers native **computer-use ↔ Local Gateway** connectivity end-to-end, then tightens correctness and pairing UX:

- **Gateway session model:** Replace the legacy local daemon wiring with **`GatewayClient` / gateway session** in `@n8n/computer-use`, expose settings helpers (`settings-store`), **`getCurrentGatewayKey`**, and **`ensureInitialized`** instead of separate file bootstrap; restore local-gateway **`build` / `lint` / `typecheck`** scripts and refresh README basics.
- **Electron app (Local Gateway):** Settings UI, tray flow, preload/IPC plumbing, **`n8n-computer-use`** custom protocol, **`open-url`** / **`second-instance`** handling (single-instance + minimal **`setAsDefaultProtocolClient`** on startup), **`connect-payload`** extraction with tests for `n8n-computer-use://` parsing.
- **Deeplink-only pairing:** Pairing happens **only** through the OS deeplink (`connect?url=…&token=…`); manual instance URL/token fields and renderer-driven **Connect** are removed—**Disconnect** and **Allowed origins** stay. Lifecycle is hardened (**teardown/`finally`** so gateway status resets reliably; **`disconnected`** wording instead of idle where appropriate).
- **Security UX:** **`connect-origin`** / **`assertConnectOriginAllowed`** so connects are rejected unless the deeplink hostname is allowed (**allowed-origin validation** expanded and tested).

**computer-use CLI:** callers use **`SettingsStore.ensureInitialized`**; **`ensureSettingsFile`** path removed/cleaned in startup config/tests.

#### How reviewers can validate

- `pnpm typecheck` and `pnpm test` from **`packages/@n8n/local-gateway`** (same for **`packages/@n8n/computer-use`** if you touch that package).
- **Real OS deeplinks (macOS):** `pnpm --filter=@n8n/local-gateway dist:mac`, run the app from **`packages/@n8n/local-gateway/out/`**, then complete pairing from the Instance AI computer-use flow and confirm connect/disconnect and origin rules.

https://linear.app/n8n/issue/NODE-4899

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/computer-use/src/gateway-client.ts
```diff
@@ -101,6 +101,11 @@ export class GatewayClient {
 
 	constructor(private readonly options: GatewayClientOptions) {}
 
+	/** Returns the currently active gateway key (session key when available). */
+	getCurrentGatewayKey(): string {
```
dlavrenuek: This does the same as making `private get apiKey()` public.

packages/@n8n/computer-use/src/startup-config-cli.ts
```diff
@@ -85,25 +85,5 @@ export function isAllDeny(permissions: Record<ToolGroup, PermissionMode>): boole
  * filesystemDir is left empty. Does NOT prompt. Safe to call on every startup.
  */
 export async function ensureSettingsFile(config: GatewayConfig): Promise<void> {
```
dlavrenuek: This method is not used any more and can be removed, we should probably just move the tests to settings store tests

packages/@n8n/local-gateway/src/main/connect-payload.ts
```diff
@@ -0,0 +1,40 @@
+import type { ConnectPayload } from '../shared/types';
+
+/** Custom URI scheme registered with the OS for deep links (`n8n-gateway://…`). */
+export const DEEP_LINK_PROTOCOL = 'n8n-gateway';
```
dlavrenuek: Technically we still call it "gateway" in multiple places in the code but for users it should be `Computer Use`. Shall we also go for `n8n-computer-use` here?

```suggestion
export const DEEP_LINK_PROTOCOL = 'n8n-computer-use';
```

packages/@n8n/computer-use/src/gateway-client.ts
dlavrenuek: I'm not sure that we need to expose the api key, at least right now to make the app work. Will ellaborate in more details further down (probably daemon controller or settings store)

packages/@n8n/local-gateway/src/main/daemon-controller.ts
```diff
@@ -15,108 +14,127 @@ export interface DaemonControllerEvents {
...
-	private _port: number | null = null;
-	private _status: DaemonStatus = 'stopped';
+	private client: GatewayClient | null = null;
+	private session: GatewaySession | null = null;
+	private settingsStore: SettingsStore | null = null;
+	private _status: DaemonStatus = 'idle';
 	private _connectedUrl: string | null = null;
 	private _connectedAt: string | null = null;
+	private _lastError: string | null = null;
+	private lastCredentials: { url: string; apiKey: string } | null = null;
```
dlavrenuek: The credentials are only used to reconnect when settings change if I see it correctly. Originally the permissions were always persisted, so a change in permissions would mean that the client should restart. Then we changed the permissions to be session based, so when you connect to an instance you can specify which tools are enabled in which modes. To make the app work right now, I would simply remove the reconnect functionality because the persisted settings are now separate from session based settings.

There is another issue to change tools/permissions for an active connection without a reconnect, so I would handle any related functionality there https://linear.app/n8n/issue/NODE-4903/in-the-client-for-an-active-connection-user-can-change-what-tools-are (or NODE-4803), a very outdated WIP version ca be found at https://github.com/n8n-io/n8n/compare/NODE-4803/add-tool-to-update-available-tool-groups?expand=1

packages/@n8n/local-gateway/src/main/index.ts
```diff
@@ -80,26 +121,49 @@ app
...
+		}
+
+		app.on('open-url', (event, url) => {
+			event.preventDefault();
+			const payload = parseConnectPayload(url);
+			if (!payload) return;
+			handleConnectPayload(payload);
+		});
+
+		app.on('second-instance', (_event, argv) => {
```
dlavrenuek: When is this event fired?

packages/@n8n/local-gateway/src/main/ipc-handlers.ts
```diff
@@ -42,15 +45,23 @@ export function registerIpcHandlers(
...
+		logger.debug('IPC gateway:connect', { url: payload.url });
+		try {
+			await connect(payload);
+			return { ok: true };
+		} catch (error) {
+			const snapshot = controller.getSnapshot();
+			const message =
+				snapshot.lastError ?? (error instanceof Error ? error.message : String(error));
+			logger.error('IPC gateway:connect failed', { error: message });
+			return { ok: false, error: message };
```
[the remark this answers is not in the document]
BerniWittmann: we anyhow log this to the file, will create a follow up ticket

packages/@n8n/local-gateway/src/main/index.ts
```diff
@@ -39,38 +50,68 @@ app
...
 
-		function confirmConnect(url: string, _session: GatewaySession): boolean {
-			const lastUrl = settingsStore.getLastConnectedUrl();
-			if (lastUrl !== null && lastUrl === url) {
-				logger.info('Auto-approving connection from known URL', { url });
+		async function connect(payload: ConnectPayload): Promise<void> {
+			const config = settingsStore.toGatewayConfig();
+			const inlineKey = payload.apiKey?.trim();
+			const reconnectKey = settingsStore.getReconnectKey() ?? undefined;
+			const effectiveKey = inlineKey && inlineKey.length > 0 ? inlineKey : reconnectKey;
```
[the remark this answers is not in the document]
BerniWittmann: fixed

packages/@n8n/local-gateway/src/main/settings-store.ts
```diff
@@ -68,12 +67,37 @@ export class SettingsStore {
 		logger.debug('Last connected URL updated', { url });
 	}
 
+	getReconnectKey(): string | null {
+		return this.store.get('reconnectKey');
+	}
+
+	setReconnectKey(key: string | null): void {
+		this.store.set('reconnectKey', key);
```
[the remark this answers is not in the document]
BerniWittmann: creating a follow up for that

packages/@n8n/local-gateway/src/main/daemon-controller.ts
```diff
@@ -15,108 +14,127 @@ export interface DaemonControllerEvents {
...
+			apiKey,
+			config,
+			session,
+			confirmResourceAccess: () => 'denyOnce',
+			onPersistentFailure: () => {
+				this._lastError = 'Gateway authentication failed repeatedly';
+				this.clearConnectionState('error');
+			},
+			onDisconnected: () => {
+				this.clearConnectionState('disconnected');
```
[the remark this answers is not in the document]
BerniWittmann: should be fixed

packages/@n8n/local-gateway/src/main/index.ts
```diff
@@ -80,26 +121,49 @@ app
...
+					return;
+				}
+				void reconnectStoredCredentials().then((connected) => {
+					if (!connected) openSettingsForTokenEntry();
+				});
+			},
 			() => {
 				logger.info('n8n Gateway quitting');
-				void controller.stop().then(() => {
+				void controller.disconnect().then(() => {
```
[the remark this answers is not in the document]
BerniWittmann: should be fixed now

packages/@n8n/computer-use/src/settings-store.ts
```diff
@@ -82,6 +83,33 @@ export class SettingsStore {
...
+		const template = getTemplate('default');
+		const permissions = { ...template.permissions, ...config.permissions };
+		const initialSettings: PersistentSettings = {
+			permissions,
+			filesystemDir: '',
+			resourcePermissions: {},
+		};
+
+		const dir = path.dirname(filePath);
+		await fs.mkdir(dir, { recursive: true, mode: 0o700 });
```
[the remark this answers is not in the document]
BerniWittmann: should be fixed now

packages/@n8n/local-gateway/src/main/daemon-controller.ts
```diff
@@ -15,108 +14,127 @@ export interface DaemonControllerEvents {
...
+		const session = new GatewaySession(defaults, store);
+		const client = new GatewayClient({
+			url: normalizedUrl,
+			apiKey,
+			config,
+			session,
+			confirmResourceAccess: () => 'denyOnce',
+			onPersistentFailure: () => {
+				this._lastError = 'Gateway authentication failed repeatedly';
+				this.clearConnectionState('error');
```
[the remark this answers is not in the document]
BerniWittmann: should be fixed now

packages/@n8n/local-gateway/src/main/settings-store.ts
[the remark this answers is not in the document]
BerniWittmann: https://linear.app/n8n/issue/NODE-4946

packages/@n8n/local-gateway/src/main/settings-store.ts
```diff
@@ -10,20 +10,20 @@ import type { AppSettings } from '../shared/types';
 export type { AppSettings };
 
 const DEFAULTS: AppSettings = {
-	port: 7655,
+	instanceUrl: '',
```
dlavrenuek: I would not store the instance url in the settings to have the settings separated from connections. This is the logic we currently have in computer-use package (or don't have depending how you see it) - the rationale behind it is that the settings are global for the app and it could be possible to have multiple concurrent connections at the same time (currently you can start multiple computer-use commands), so storing the url in the global settings breaks this option.

packages/@n8n/local-gateway/src/main/settings-store.ts
```diff
@@ -68,12 +67,37 @@ export class SettingsStore {
 		logger.debug('Last connected URL updated', { url });
 	}
 
+	getReconnectKey(): string | null {
```
dlavrenuek: We have a separate issue for reconnecting on app start, but it's not specified very well (that's on me). So far we did not have this functionality due to technical limitations of current token handling (only short lived session tokens). To have proper reconnect we should see how the computer use tokens are generated and if it should be adjusted. So for now I would remove the reconnect functionality from the app

packages/@n8n/local-gateway/src/main/daemon-controller.ts
```diff
@@ -15,110 +14,141 @@ export interface DaemonControllerEvents {
...
+	async disconnect(): Promise<void> {
+		await this.closeCurrentConnection({ preserveServerSession: false });
+		this.setStatus('disconnected');
+	}
+
+	private afterGatewayPersistentFailure(): void {
+		this._lastError = 'Gateway authentication failed repeatedly';
+		void this.closeCurrentConnection({ preserveServerSession: true }).then(() => {
+			this.setStatus('error');
+		});
```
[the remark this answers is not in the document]
BerniWittmann: fixed

packages/@n8n/local-gateway/src/main/settings-store.ts
```diff
@@ -68,12 +67,37 @@ export class SettingsStore {
...
+	}
+
+	setReconnectKey(key: string | null): void {
+		this.store.set('reconnectKey', key);
+	}
+
+	/** Origin derived from the configured n8n URL for `GatewayConfig.allowedOrigins`. */
+	private allowedOriginsForN8nInstance(settings: AppSettings): string[] {
+		const fromSettings = settings.instanceUrl.trim();
+		const fromLast = this.getLastConnectedUrl()?.trim() ?? '';
```
dlavrenuek: We must not use the connection url as allowed origin, these are separate "logical entities". Allowed origins is part of the global config (stored via settings.json file, should be reflected in app settings view). The user must actively set/configure the allowed origins to anything else than the n8n cloud. The url to connect to must be treated as external user input. Let's imagine that some malicious webpage includes a deeplink `n8n-gateway://connect...` linking to an external server, the app must block this request.

Looking at the current implementation we don't store the origins int the settings file, which we actually should instead of only parsing them from args in cli

packages/@n8n/local-gateway/src/main/tray.ts
```diff
@@ -7,23 +7,26 @@ import type { DaemonController, DaemonStatus, StatusSnapshot } from './daemon-co
 
 const STATUS_LABELS: Record<DaemonStatus, string> = {
 	connected: '', // set dynamically with URL
-	waiting: '◌  Waiting for connection',
-	starting: '○  Starting...',
-	disconnected: '✕  Disconnected — reconnecting',
-	stopped: '■  Stopped',
+	connecting: '○  Connecting...',
+	disconnected: '✕  Disconnected',
+	idle: '■  Idle',
```
dlavrenuek: I think there is no more idle state since we don't have any more server running, or?

packages/@n8n/local-gateway/src/renderer/index.html
```diff
@@ -31,28 +31,25 @@
 					<div class="form-section-title">Connection</div>
 
 					<div class="form-row">
-						<label class="form-label" for="port">Port</label>
+						<label class="form-label" for="instanceUrl">n8n URL</label>
```
dlavrenuek: I would not store the instance url in the settings but have a separate view for that somewhere. For now it should be fine to just have the deeplink, so the app is functional and add any changes afterwards

packages/@n8n/local-gateway/src/renderer/index.html
```diff
@@ -31,28 +31,25 @@
...
+						<input class="form-input" type="text" id="instanceUrl" name="instanceUrl" />
+					</div>
+
+					<div class="form-row" id="gatewayTokenRow">
+						<label class="form-label" for="gatewayKey">Gateway token</label>
```
dlavrenuek: Also the token should not be part of the settings

dlavrenuek (changes requested):
Makes the app work again, very nice! The deeplink functionality is very cool and provides a nice user experience. There are a lot of changes since the app is very outdated, I've added comments and suggestions

packages/@n8n/local-gateway/src/main/index.ts
BerniWittmann: Apparently this fires on windows/linux on deep link while app is already running. Will add a code comment

packages/@n8n/local-gateway/src/main/settings-store.ts
BerniWittmann: Ditched all the reconnection parts

packages/@n8n/local-gateway/src/main/tray.ts
BerniWittmann: yes true. I had it in there intially for the state: app running but never had any connection. But ditchd it and shows disconnected now as well. Which is correct, but might be something we examine later for better user understanding

packages/@n8n/local-gateway/src/main/index.ts
```diff
@@ -1,107 +1,130 @@
...
-				.then(() => {
-					controller.start(config, confirmConnect);
-				})
-				.catch((e: unknown) => {
-					logger.error('Failed to restart daemon', { e: String(e) });
-				});
-		}
 
-		registerIpcHandlers(controller, settingsStore, restartDaemon);
+			app.setAsDefaultProtocolClient(DEEP_LINK_PROTOCOL);
```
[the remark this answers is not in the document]
BerniWittmann: I would defer that until we set up automated building & signing process

packages/@n8n/local-gateway/src/main/settings-store.ts
BerniWittmann: Done now

packages/@n8n/local-gateway/src/main/index.ts
```diff
@@ -1,107 +1,130 @@
...
+				if (!token || token.length === 0) {
+					throw new Error(
+						'Missing gateway token in deeplink. Connect from n8n using the computer-use link.',
+					);
+				}
+				await controller.connect(config, payload.url, token);
+			}
+
+			async function disconnectGateway(): Promise<void> {
+				await controller.disconnect();
```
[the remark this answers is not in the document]
BerniWittmann: We dont have to worry about that, because this is a MVP right now, so dont have to thing about migrating legacy things

BerniWittmann:
@dlavrenuek As your comments made some good points around the "architecture" / direction of this I made some more changes (sorry for having to review this).
Essentially I ditched reconnection logic, manually entering token & url (only deeplink) and decoupled connection stuff from the settings

dlavrenuek (approved):
Build works as expected. I did not test the full functionality of the native app since the deep link is not registered (or displayed) but we can address it in the future as needed

PR #29448, opened 2026-04-29
fix(agents): create agents in the selected project

bjorger:
## Summary

- Pass the active `projectId` from `ProjectHeader` into the new-agent route
  so agents created from a project header land in that project instead of
  the user's personal project.
- Keep the sidebar's universal-add menu above chat content by teleporting
  the submenu and giving the sidebar its own stacking context.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
@@ -100,6 +100,7 @@ const {
 			ref="createBtn"
 			data-test-id="universal-add"
 			:menu="menu"
+			:teleport="true"
 			@select="handleMenuSelect"
 		>
 			<N8nIconButton
--- a/packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
@@ -100,8 +100,6 @@ const {
 			ref="createBtn"
 			data-test-id="universal-add"
 			:menu="menu"
-			:teleport="true"
-			:submenu-class="$style.createMenu"
 			@select="handleMenuSelect"
 		>
 			<N8nIconButton
@@ -243,9 +241,5 @@ const {
 
 .upgradeButton {
 	margin-left: auto;
-}
-
-.createMenu {
-	z-index: var(--floating-ui--z) !important;
 }
 </style>
--- a/packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
+++ b/packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
@@ -100,6 +100,8 @@ const {
 			ref="createBtn"
 			data-test-id="universal-add"
 			:menu="menu"
+			:teleport="true"
+			:submenu-class="$style.createMenu"
 			@select="handleMenuSelect"
 		>
 			<N8nIconButton
@@ -241,5 +243,9 @@ const {
 
 .upgradeButton {
 	margin-left: auto;
+}
+
+.createMenu {
+	z-index: var(--floating-ui--z) !important;
 }
 </style>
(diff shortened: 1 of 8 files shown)


packages/frontend/editor-ui/src/app/components/MainSidebarHeader.vue
```diff
@@ -100,6 +100,8 @@ const {
 			ref="createBtn"
 			data-test-id="universal-add"
 			:menu="menu"
+			:teleport="true"
```
yehorkardash, line 103: Can we change something on our page rather than change main sidebar and main layout? It also seems to break some e2e tests

yehorkardash (changes requested):

packages/frontend/editor-ui/src/features/agents/views/AgentView.vue
```diff
@@ -1,17 +1,22 @@
...
+<style lang="scss" module>
 .agentView {
 	display: flex;
 	flex-direction: column;
 	height: 100%;
 	width: 100%;
 	min-height: 0;
 }
+
+:global(.app-grid:has(.agent-view-page) > #sidebar) {
```
yehorkardash: It feels a bit hacky imo that we change sidebar style if we are on agents page. Checked the code and it seems to be caused by us using N8NResizewrapper with custom z-index.
Your initial fix where you set `teleport="true"` seems to be correct. Why'd you added `.createMenu` and changed `sidebar` style in the first commit, though?

We could potentially go back to `teleport="true"`, just need to make sure that e2e test catch any issues

packages/frontend/editor-ui/src/features/agents/views/NewAgentView.vue
```diff
@@ -12,12 +12,17 @@ import { AGENT_BUILDER_VIEW } from '../constants';
...
 
 const router = useRouter();
+const route = useRoute();
 const rootStore = useRootStore();
 const usersStore = useUsersStore();
 const projectsStore = useProjectsStore();
 const telemetry = useTelemetry();
 
-const projectId = computed(() => projectsStore.personalProject?.id ?? '');
+const projectId = computed(() =>
```
yehorkardash, line 28: Clicking back button returns to home page rather than previous project page

Image

packages/frontend/editor-ui/src/features/agents/views/AgentView.vue
bjorger: That is an excellent point to be honest. I will go back to the teleport="true" and make sure that the e2e tests pass

packages/testing/playwright/pages/SidebarPage.ts
```diff
@@ -27,10 +27,8 @@ export class SidebarPage {
 		return this.visibleNavigationMenu.getByTestId('navigation-submenu').filter({ hasText: label });
 	}
 
-	getVisibleNavigationSubmenuItem(label: string): Locator {
-		return this.visibleNavigationMenu
-			.getByTestId('navigation-submenu-item')
-			.filter({ hasText: label });
+	getVisibleNavigationLink(label: string, hrefPart: string): Locator {
+		return this.visibleNavigationMenu.locator(`a[href*="${hrefPart}"]`).filter({ hasText: label });
```
yehorkardash, line 31: Did you figure out why "this.visibleNavigationMenu.getByTestId('navigation-submenu-item')" start to fail with teleport enabled?

yehorkardash (commented):
LGTM 🚀
Just one question

packages/testing/playwright/pages/SidebarPage.ts
bjorger: With teleport="true", the dropdown menu is no longer rendered inside the universal-add element. Element Plus moves it into floating poppers elsewhere in the DOM.

yehorkardash (approved):

PR #29450, opened 2026-04-29
feat: Improve computer-use prompt in Instance AI (no-changelog)

dlavrenuek:
## Summary

- Improves the code structure around Computer Use prompt generation within AI Assistant
  - extract to separate file
  - remove separate filesystem and browser prompt generation methods, use one computer use prompt generation method
- Removes redundant state variables like if filesystem is available (`browserAvailable` is still in place but could probably also be removed and replaces by check if capabilities include browser)
- Includes a list of basic capabilities in system prompt unconditionally, previously these were condition based on the connect status
  - Allows AI to suggest Computer Use when user describes a use case suitable for it
  - Allows AI to suggest enabling additional tools if these are disabled and user asks for a use case where these are needed
- Makes the Computer Use connection status more straightforward, also add `disabledGlobally` status to have different handling than disabled for user (suggest user to enable Computer Use if they need it)

https://linear.app/n8n/issue/NODE-4932

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/instance-ai/src/agent/__tests__/computer-use-prompt.test.ts
+++ b/packages/@n8n/instance-ai/src/agent/__tests__/computer-use-prompt.test.ts
@@ -24,7 +24,7 @@ describe('getComputerUsePrompt', () => {
 		it('includes the Computer Use intro section', () => {
 			const result = getComputerUsePrompt({
 				browserAvailable: undefined,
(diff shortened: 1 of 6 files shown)


BerniWittmann (approved):
LGTM, without being able to assess the performance

PR #29453, opened 2026-04-29
fix: Skip AI tool generation for community trigger nodes

michael-radency:
## Summary

Fixes a bug where community nodes that are triggers were incorrectly having AI tool variants generated for them. Trigger nodes should not be used as AI tools.

The fix filters out nodes from AI tool generation if they either:
- Have trigger in their group array
- Have trigger (case-insensitive) anywhere in their node type name

## Review / Merge checklist

- [ ] Tests included. 


packages/cli/src/modules/community-packages/community-node-types.service.ts
```diff
@@ -148,7 +148,11 @@ export class CommunityNodeTypesService {
 
 	private createAiTools() {
 		const usableAsTools = Array.from(this.communityNodeTypes.values()).filter(
-			(nodeType) => nodeType.nodeDescription.usableAsTool && !isToolType(nodeType.name),
+			(nodeType) =>
+				nodeType.nodeDescription.usableAsTool &&
+				!isToolType(nodeType.name) &&
+				!nodeType.nodeDescription.group?.includes('trigger') &&
+				!nodeType.nodeDescription.name.toLowerCase().includes('trigger'),
```
elsmr: Do we need this name check? I could see edge cases, like if someone were to write a node for `trigger.dev`

The frontend also uses the `group` check everywhere, so without that I don't think a trigger node will even work correctly.

elsmr (approved):

PR #29454, opened 2026-04-29
fix(core): Recreate data table backing tables on entity import

CharlieKolb:
## Summary

`n8n export:entities` captures rows in `data_table` and `data_table_column`, but the backing `data_table_user_<id>` tables are pure runtime DDL with no TypeORM entity, so they aren't recreated on the destination during `import:entities`. After import the data tables appear in the UI but any read or write fails because the backing tables don't exist on the destination.

This PR fixes the import side only — it does not change export. After import, data tables are visible and usable, but contain zero rows. [A follow-up PR](https://github.com/n8n-io/n8n/pull/29455) extends this with row export/import for full round-trip backups.

### What it does
- Injects `DataTableDDLService` into `ImportService`
- Adds `dropExistingDataTableUserTables` — runs before `--truncateTables` so dynamic backing tables don't end up orphaned once the registry is wiped
- Adds `recreateDataTableUserTablesFromRegistry` — runs after the entity import phase, drops any stale backing table, then creates an empty one with the imported column metadata
- Tests cover: drop on existing registry, prefix handling, missing-registry skip, columns sorted by index, drop-then-create order, registry-empty no-op

### How to test
1. On a source instance, create a few data tables with columns and rows in the UI.
2. Run `n8n export:entities --outputDir=./out`.
3. On a fresh destination, run `n8n import:entities --inputDir=./out --truncateTables`.
4. Open the destination — the data tables appear with their columns intact (rows will be empty, that's expected for this PR).
5. Verify you can add rows, edit columns, and delete the tables — all operations now work.

https://linear.app/n8n/issue/ADO-5143

Related: ADO-5032 (Personio incident — same root cause)

## Review / Merge checklist

  - Tested both sqlite and postgres import
- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/services/__tests__/import.service.test.ts
+++ b/packages/cli/src/services/__tests__/import.service.test.ts
@@ -6,6 +6,7 @@ import { mock } from 'jest-mock-extended';
 import type { Cipher } from 'n8n-core';
 
 import type { ActiveWorkflowManager } from '@/active-workflow-manager';
+import type { DataTableDDLService } from '@/modules/data-table/data-table-ddl.service';
 import type { WorkflowIndexService } from '@/modules/workflow-index/workflow-index.service';
 
 import { ImportService } from '../import.service';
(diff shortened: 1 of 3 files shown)


RicardoE105 (approved):
Code looks good. Tested with postgres and worked great! Nice work 🥇

PR #29455, opened 2026-04-29
feat(core): Export and import data table user-row contents

CharlieKolb:
## Summary

The previous PR (#29454) recreates empty backing tables on import, restoring basic functionality. This PR completes the round trip by exporting the rows themselves so a true backup/restore of data tables is possible.

### Export
- Adds \`exportDataTableUserTables\` after the entity-metadata loop. Uses keyset pagination on \`id\` (auto-increment) instead of \`OFFSET\` so large tables don't degrade on Postgres.
- Encrypts to \`data_table_user_<id>.jsonl\` (paginated 500 rows/file) using the same cipher and zip path as the rest of the archive.
- Adds \`--includeDataTableRows\` flag (default \`true\`) so users can opt out of row contents and produce schema-only archives.

### Import
- Routes \`data_table_user_*.jsonl\` files in \`getImportMetadata\` into a separate bucket keyed by data table id. Old archives (or schema-only ones) without those files still import cleanly — the backing tables are recreated empty.
- Extends \`recreateDataTableUserTablesFromRegistry\` to load and insert rows from the bucket, preserving \`id\`, \`createdAt\`, and \`updatedAt\` on the destination.
- Normalises booleans (sqlite stores 0/1 INTEGER, Postgres BOOLEAN) and dates (sqlite TEXT lacks timezone info, treated as UTC) so cross-DB exports round-trip correctly.
- Resets the Postgres IDENTITY sequence to \`MAX(id)\` after inserts since explicit-id inserts don't auto-advance it.

### How to test

**Same-DB round trip (sqlite → sqlite or Postgres → Postgres):**
1. On a source, create a data table with columns of every type (string, number, boolean, date) and add a few rows.
2. \`n8n export:entities --outputDir=./out\`
3. On a fresh destination, \`n8n import:entities --inputDir=./out --truncateTables\`
4. Verify the data table on the destination has the same rows with the same ids and timestamps.
5. Add a new row via the UI on the destination — confirm the auto-id continues from \`MAX(id)+1\` (no collision).

**Cross-DB round trip (sqlite → Postgres):** repeat with two instances on different DB engines and confirm boolean and date values survive correctly.

**Schema-only export:** \`n8n export:entities --includeDataTableRows=false\` then import. Backing tables are created empty on the destination.

https://linear.app/n8n/issue/ADO-5143

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/modules/data-table/utils/sql-utils.ts
```diff
@@ -296,6 +296,35 @@ export function normalizeValueForDatabase(
 	return value;
 }
 
+/**
+ * Normalize a raw value coming from a cross-database data-table export so it can
+ * be inserted into the destination database. Differs from `normalizeValueForDatabase`
+ * by also handling booleans (SQLite stores them as 0/1 INTEGER, Postgres as BOOLEAN)
+ * and SQLite-style date strings without timezone info (treated as UTC).
+ */
+export function normalizeUserRowValueForDatabase(
```
RicardoE105, line 306: can we reuse anything from normalizeRows in the sql-utils.ts?

packages/cli/src/services/export.service.ts
```diff
@@ -90,11 +93,163 @@ export class ExportService {
 		return systemTablesExported;
 	}
 
+	private async exportDataTableUserTables(
```
RicardoE105: Should we have smaller functions inside? quite a long function to read?

packages/cli/src/services/import.service.ts
```diff
@@ -755,22 +793,77 @@ export class ImportService {
 		const columnsByDataTableId = new Map<string, DataTableColumn[]>();
 		for (const row of columnRows) {
 			const list = columnsByDataTableId.get(row.dataTableId) ?? [];
-			list.push(transactionManager.create(DataTableColumn, row));
+			list.push(row as unknown as DataTableColumn);
```
RicardoE105: can we avoid the type assertion ?

packages/cli/src/services/__tests__/export.service.test.ts
```diff
@@ -338,6 +370,103 @@ describe('ExportService', () => {
...
+	describe('exportDataTableUserTables', () => {
+		it('should silently skip when the data_table registry is missing', async () => {
+			jest.mocked(mockDataSource.query).mockImplementation(async (query: string) => {
+				if (query.includes('data_table') && !query.includes('data_table_column')) {
+					throw new Error('relation does not exist');
+				}
+				return [];
+			});
+
+			// @ts-expect-error accessing private method for testing
```
RicardoE105, line 382: can we avoid calling this private method?

packages/cli/src/services/export.service.ts
RicardoE105: maybe running the simplify skill in this PR?

packages/cli/src/services/export.service.ts
```diff
@@ -90,11 +93,163 @@ export class ExportService {
...
+			let fileIndex = 1;
+			let currentFileEntityCount = 0;
+			let totalEntityCount = 0;
+			let hasNextPage = true;
+
+			do {
+				let pageRows: Array<Record<string, unknown>>;
+				try {
+					pageRows = await this.dataSource.query(
+						`SELECT * FROM ${escapedUserTable} WHERE "id" > ${lastId} ORDER BY "id" LIMIT ${pageSize}`,
```
RicardoE105: should we make sure the ID quotation is correct?

`  const idCol = quoteIdentifier('id', dbType);`

packages/cli/src/services/import.service.ts
```diff
@@ -755,22 +793,77 @@ export class ImportService {
...
+					rowsForTable++;
+				}
+			}
+
+			// Postgres IDENTITY columns don't auto-advance when explicit ids are
+			// inserted, so the next implicit insert would collide. Reset to MAX(id).
+			// pg_get_serial_sequence's first arg goes through regclass and folds
+			// unquoted identifiers to lowercase, so pass the already-quoted form.
+			if (rowsForTable > 0 && dbType === 'postgres') {
+				await transactionManager.query(
```
RicardoE105: can we reuse the query we have in the `advanceIdentitySequences` method. Seems the same

packages/cli/src/services/__tests__/export.service.test.ts
CharlieKolb: This is what all the existing tests did 😅 Can refactor that and make this public if you prefer

packages/cli/src/modules/data-table/utils/sql-utils.ts
CharlieKolb: I refactored a `normalizeBoolean` function, I don't think it's worth generalizing `normalizeRows` beyond that tho. Can try if you disagree

packages/cli/src/services/__tests__/export.service.test.ts
RicardoE105: ahh if this is what all other tests had then no need to address in this PR

RicardoE105 (approved):

PR #29459, opened 2026-04-29
feat(core): Broadcast workflow settings updates

MiloradFilipovic:
## Summary
Fixes a bug where a workflow's `checksum` would be out of sync (causing autosave to break) if mcp access is toggle from another browser tab/window.
As a fix, we are broadcasting `workflowSettingsUpdated` event for all open workflows. 

Fixes ADO-5111

## How to test
Check [steps to reproduce](https://linear.app/n8n/issue/ADO-5111/bug-workflow-autosave-breaks-if-mcp-toggle-is-updated-from-another) in the linear ticket and make sure autosave still works.

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/collaboration/collaboration.service.ts
+++ b/packages/cli/src/collaboration/collaboration.service.ts
@@ -266,16 +266,29 @@ export class CollaborationService {
 
 		for (let start = 0; start < uniqueWorkflowIds.length; start += OPEN_WORKFLOW_CHECK_BATCH_SIZE) {
 			const chunk = uniqueWorkflowIds.slice(start, start + OPEN_WORKFLOW_CHECK_BATCH_SIZE);
-			const openIdsInChunk = await Promise.all(
+			const collaboratorLookups = await Promise.allSettled(
 				chunk.map(async (workflowId) => {
 					const collaborators = await this.state.getCollaborators(workflowId);
-					return collaborators.length > 0 ? workflowId : undefined;
+					return { workflowId, isOpen: collaborators.length > 0 };
 				}),
 			);
+			const failedWorkflowIds: Array<Workflow['id']> = [];
 
-			openWorkflowIds.push(
-				...openIdsInChunk.filter((id): id is Workflow['id'] => id !== undefined),
-			);
+			for (const [index, result] of collaboratorLookups.entries()) {
+				if (result.status === 'fulfilled') {
+					if (result.value.isOpen) openWorkflowIds.push(result.value.workflowId);
+				} else {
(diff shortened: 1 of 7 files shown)


packages/cli/src/collaboration/collaboration.service.ts
```diff
@@ -258,6 +260,27 @@ export class CollaborationService {
 		this.push.sendToUsers({ type: 'workflowUpdated', data: msgData }, userIds);
 	}
 
+	async filterOpenWorkflowIds(workflowIds: Array<Workflow['id']>): Promise<Array<Workflow['id']>> {
+		const uniqueWorkflowIds = [...new Set(workflowIds)];
+		const openWorkflowIds: Array<Workflow['id']> = [];
+
+		for (let start = 0; start < uniqueWorkflowIds.length; start += OPEN_WORKFLOW_CHECK_BATCH_SIZE) {
+			const chunk = uniqueWorkflowIds.slice(start, start + OPEN_WORKFLOW_CHECK_BATCH_SIZE);
+			const openIdsInChunk = await Promise.all(
```
dariacodes: Should we switch to `Promise.allSettled` so a failure in `getCollaborators` doesn’t cause the entire batch to fail?

packages/cli/src/modules/mcp/mcp.settings.service.ts
```diff
@@ -16,13 +18,35 @@ const KEY = 'mcp.access.enabled';
...
+	'name',
+	'description',
+	'nodes',
+	'connections',
+	'settings',
+	'meta',
+	'pinData',
+	'isArchived',
+	'activeVersionId',
+];
```
dariacodes: Could we remove this duplication by exporting `CHECKSUM_FIELDS` from `workflow-checksum.ts`?
And `id` will be selected anyway as it's the primary key

dariacodes (commented):
I tested it on 7 different workflows - works well, and the events are sent only for those that are open in the browser. Nice work! 🚀 
I left two suggestions that can be addressed very quickly!

packages/cli/src/collaboration/collaboration.service.ts
MiloradFilipovic: Yup, will update

packages/cli/src/modules/mcp/mcp.settings.service.ts
MiloradFilipovic: Good catch 👍

dariacodes (approved):
🚢

PR #29469, opened 2026-04-29
fix: Handle IMAP fetch errors to prevent instance crash and stuck workflows

michael-radency:
## Summary

Fixes two bugs in `@n8n/imap` that caused the IMAP trigger to either crash the n8n instance or leave workflows permanently stuck.

In `imap-simple.ts`, removing the error listener from the fetch object in `fetchOnEnd` left a window where a late 'error' event had no handler, producing an uncaught exception that killed the process. In `get-message.ts`, the message parsing promise had no reject path, so a stream error during fetch caused it to hang forever — blocking the `onMail` callback indefinitely.

https://linear.app/n8n/issue/NODE-2523/imap-trigger-crashes-instance-when-it-cannot-connect-to-the-inbox

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/imap/src/helpers/get-message.ts
```diff
@@ -35,6 +35,7 @@ export async function getMessage(
 					body: /^HEADER/g.test(info.which) ? parseHeader(body) : body,
 				});
 			});
+			stream.once('error', reject);
```
dlavrenuek: The stream could fire multiple error events. If I understand the `once` functionality correctly it would first reject but the next error event would then pop up and potentially crash the process?

Afaik it is safe to call reject multiple times, so `stream.on('error', reject);` should technically work. Otherwise you could also call `stream.end(); reject();` from within the error handler

dlavrenuek (approved):

PR #29473, opened 2026-04-29
fix: Use explicit node references for AI memory session keys

aalises:
## Summary

- Add `nodeJson(node, path)` to `@n8n/workflow-sdk` so generated workflows can create explicit source-node JSON references without hand-writing `$(...)` expressions.
- Add a focused workflow-sdk validator for custom AI memory session keys that use `$json` in memory subnodes.
- Update Instance AI/workflow-sdk prompt guidance to teach explicit references for AI subnodes and chatbot memory session keys.
- Add Instance AI eval coverage: a deterministic memory-session-key check plus a Telegram chatbot memory workflow case.

https://linear.app/n8n/issue/INS-178

## Review / Merge checklist


packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
```diff
@@ -108,6 +118,19 @@ const parser = outputParser({
   }
 });
 
+const memoryNode = memory({
+  type: '@n8n/n8n-nodes-langchain.memoryBufferWindow',
+  version: 1.3,
+  config: {
+    name: 'Conversation Memory',
+    parameters: {
+      sessionIdType: 'customKey',
```
Cadiac, line 127: This prompt will probably guide the AI away from only using the customKey mechanism when it makes sense instead of the automatic "connected chat trigger" mode that would work just fine here - but I suppose we don't really mind if this ends up preferring the customKey way of connecting the chat trigger manually even when it doesn't have to?

packages/@n8n/workflow-sdk/src/prompts/best-practices/guides/chatbot.ts
```diff
@@ -33,6 +33,17 @@ Include information with the user prompt such as timestamp, user ID, or session
 If there are other agents involved in the workflow you should share memory between the chatbot and those other agents where it makes sense.
 Connect the same memory node to multiple agents to enable data sharing and context continuity.
 
+### Memory Session Keys
```
Cadiac, line 36: I wonder if builderHint on the memory nodes sessionKey property would be more effective, but if this seems to do the trick I'm not opposed to mentioning this here either

Cadiac (approved):
🚀 if this fixes the cases we saw

PR #29474, opened 2026-04-29
fix(core): Reject empty webhookMethods in community lint rule

haimingZZ:
## Summary
- Treat an empty top-level `webhookMethods = {}` object as an incomplete webhook lifecycle
- Report the same missing lifecycle methods as an empty `default` lifecycle group
- Add a focused RuleTester invalid case for webhook trigger nodes with no lifecycle groups

## Review notes
The rule already catches missing `webhookMethods` and empty nested lifecycle groups. This closes the adjacent false negative where the property exists but has no groups at all.

## Validation
- `git diff --check`

Not run locally:
- `pnpm --filter @n8n/eslint-plugin-community-nodes test -- webhook-lifecycle-complete.test.ts` (local dependency install did not complete; `vitest` is unavailable in this checkout)

### DIFF
--- a/packages/@n8n/eslint-plugin-community-nodes/src/rules/webhook-lifecycle-complete.ts
+++ b/packages/@n8n/eslint-plugin-community-nodes/src/rules/webhook-lifecycle-complete.ts
@@ -91,6 +91,18 @@ export const WebhookLifecycleCompleteRule = createRule({
 					return;
 				}
 
+				if (webhookMethodsProperty.value.properties.length === 0) {
+					context.report({
+						node: webhookMethodsProperty.value,
+						messageId: 'missingLifecycleMethod',
+						data: {
+							group: 'default',
+							missing: REQUIRED_METHODS.map((m) => `\`${m}\``).join(', '),
+						},
+					});
+					return;
+				}
+
 				for (const groupProperty of webhookMethodsProperty.value.properties) {
 					if (groupProperty.type !== AST_NODE_TYPES.Property) continue;
 					if (groupProperty.value.type !== AST_NODE_TYPES.ObjectExpression) continue;
--- a/packages/@n8n/eslint-plugin-community-nodes/src/rules/webhook-lifecycle-complete.ts
+++ b/packages/@n8n/eslint-plugin-community-nodes/src/rules/webhook-lifecycle-complete.ts
@@ -56,6 +56,8 @@ export const WebhookLifecycleCompleteRule = createRule({
 		messages: {
 			missingWebhookMethods:
 				'Webhook trigger node is missing the `webhookMethods` property. Implement `checkExists`, `create`, and `delete` to register, verify, and clean up the webhook on the third-party service.',
+			emptyWebhookMethods:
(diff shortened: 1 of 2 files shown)


packages/@n8n/eslint-plugin-community-nodes/src/rules/webhook-lifecycle-complete.ts
```diff
@@ -91,6 +91,18 @@ export const WebhookLifecycleCompleteRule = createRule({
 					return;
 				}
 
+				if (webhookMethodsProperty.value.properties.length === 0) {
+					context.report({
+						node: webhookMethodsProperty.value,
```
garritfra: The existing loop reports on `groupProperty.key` — for consistency this should probably be `webhookMethodsProperty.key` instead.

packages/@n8n/eslint-plugin-community-nodes/src/rules/webhook-lifecycle-complete.ts
```diff
@@ -91,6 +91,18 @@ export const WebhookLifecycleCompleteRule = createRule({
...
+						messageId: 'missingLifecycleMethod',
+						data: {
+							group: 'default',
```
garritfra: This feels off — the message will say "`webhookMethods.default` is missing…" but there is no `default` key anywhere in the source. Would `missingWebhookMethods` be more accurate, or a dedicated message?

garritfra (changes requested):

haimingZZ:
Addressed the review comments in 2e3773a:\n\n- report the empty object on the  property key for consistency with the existing lifecycle group reports\n- replaced the synthetic  message with a dedicated  diagnostic, since an empty object has no  key in the source\n- updated the rule test to assert the new diagnostic for the empty-object case\n\nValidation note:  and the delivery conflict-marker scan are clean. I attempted to run the focused package test (), but this local worktree does not currently have package dependencies installed (), and a  ERR_PNPM_NO_PKG_MANIFEST  No package.json found in /Users/ming attempt was stopped after repeated registry  retries. GitHub CI has been triggered on the new commit for the authoritative package validation.

haimingZZ:
Addressed the review comments in 2e3773a:

- report the empty object on the `webhookMethods` property key for consistency with the existing lifecycle group reports
- replaced the synthetic `webhookMethods.default is missing...` message with a dedicated `emptyWebhookMethods` diagnostic, since an empty object has no `default` key in the source
- updated the rule test to assert the new diagnostic for the empty-object case

Validation note: `git diff --check` and the delivery conflict-marker scan are clean. I attempted to run the focused package test (`pnpm --filter @n8n/eslint-plugin-community-nodes test -- webhook-lifecycle-complete.test.ts`), but this local worktree does not currently have package dependencies installed (`vitest: command not found`), and a `pnpm install --frozen-lockfile` attempt was stopped after repeated registry `ECONNRESET` retries. GitHub CI has been triggered on the new commit for the authoritative package validation.

garritfra:
Thank you!

garritfra (approved):

PR #29476, opened 2026-04-29
fix: Merge tool-result and tool-call content blocks into one

yehorkardash:
## Summary

This PR makes a breaking but needed change in @n8n/agents package. Now tool-call content blocks also contain tool results. Previous implementation stored tool result blocks as a separate content block, but it leads to some problems because AI SDK expects tool-result to always go after tool-call block. This change guarantees that this will always be the case

https://linear.app/n8n/issue/AGENT-41/bug-tool-result-is-missing-for-tool-call-when-tool-result-arrives

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/agents/src/runtime/message-list.ts
+++ b/packages/@n8n/agents/src/runtime/message-list.ts
@@ -2,11 +2,13 @@ import type { ProviderOptions } from '@ai-sdk/provider-utils';
 import type { ModelMessage } from 'ai';
 
 import { toAiMessages } from './messages';
+import { stringifyError } from './runtime-helpers';
 import { stripOrphanedToolMessages } from './strip-orphaned-tool-messages';
 import { buildWorkingMemoryInstruction } from './working-memory';
 import { filterLlmMessages, getCreatedAt } from '../sdk/message';
 import type { SerializedMessageList } from '../types/runtime/message-list';
-import type { AgentDbMessage, AgentMessage } from '../types/sdk/message';
+import type { AgentDbMessage, AgentMessage, ContentToolCall } from '../types/sdk/message';
+import type { JSONValue } from '../types/utils/json';
 
 export type { SerializedMessageList };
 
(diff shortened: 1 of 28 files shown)


packages/@n8n/agents/src/runtime/strip-orphaned-tool-messages.ts
```diff
@@ -48,14 +22,7 @@ export function stripOrphanedToolMessages<T extends AgentMessage>(messages: T[])
...
 		const filtered = msg.content.filter((block: MessageContent) => {
-			if (block.type === 'tool-call' && block.toolCallId && orphanedCallIds.has(block.toolCallId)) {
-				return false;
-			}
-			if (
-				block.type === 'tool-result' &&
-				block.toolCallId &&
-				orphanedResultIds.has(block.toolCallId)
-			) {
+			if (block.type === 'tool-call' && block.state === 'pending') {
```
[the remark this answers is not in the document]
yehorkardash: This is a conscious change

packages/@n8n/agents/src/runtime/message-list.ts
```diff
@@ -134,6 +136,70 @@ export class AgentMessageList {
...
+		if ('output' in mutableBlock) {
+			delete (mutableBlock as { output: unknown }).output;
+		}
+
+		this.responseSet.add(host);
+		return host;
+	}
+
+	private findToolCallHost(toolCallId: string): AgentDbMessage | undefined {
+		return this.all.find(
```
bjorger: `findToolCallHost()` searches `this.all` from oldest to newest, so if a `toolCallId` is ever reused across loaded history and the current response, `setToolCallResult()` / `setToolCallError()` can mutate the historical block instead of the pending block from this turn. That leaves the current pending block unresolved, and the next LLM call strips it while replaying the stale historical result.

Could we prefer the newest pending tool-call block, ideally from response messages, or pass the host message/block reference through from the AI response conversion?

packages/@n8n/agents/src/runtime/messages.ts
```diff
@@ -201,82 +177,172 @@ function fromAiContent(part: AiContentPart): MessageContent | undefined {
...
+
+				const { output } = part;
+				if (output.type === 'json' || output.type === 'text') {
+					const mutableBlock = block as Extract<ContentToolCall, { state: 'resolved' }>;
+					mutableBlock.state = 'resolved';
+					mutableBlock.output = output.value as JSONValue;
+				} else if (output.type === 'error-json') {
+					const mutableBlock = block as Extract<ContentToolCall, { state: 'rejected' }>;
+					mutableBlock.state = 'rejected';
+					mutableBlock.error = stringifyError(output.value);
```
bjorger: For `error-json`, this goes through `stringifyError()`, which uses `String(error)` for non-`Error` values. If the provider gives us a structured JSON error object, this becomes `[object Object]`, so the stored rejected tool-call and future LLM context lose the useful error details.

Could we stringify object values with `JSON.stringify()` here, with a fallback to `String()` for non-serializable values?

bjorger (changes requested):
LGTM in overall - 2 small findings.

packages/@n8n/agents/src/runtime/message-list.ts
yehorkardash: It is possible for tool call to be in the previous turn and tool call ids shouldn't repeat, but it still makes sense to search from last 👍

bjorger (approved):

PR #29477, opened 2026-04-29
fix(Postgres Node): Output Large-Format Numbers As option ignored after pool is cached

michael-radency:
## Summary

Fixes the **Output Large-Format Numbers As** option being ignored, always outputting text regardless of the selected value.

Two bugs: the connection pool cache key didn't include `largeNumbersOutput` (so changing the setting reused the old pool), and `pg.types.setTypeParser` mutates global state (so concurrent nodes with different settings overwrote each other's parsers).

Fix: added `poolKeyExtras` to `ConnectionPoolManager` to separate pools by setting, and replaced `setTypeParser` with `pg-promise`'s per-instance receive event.

closes https://github.com/n8n-io/n8n/issues/18470
https://linear.app/n8n/issue/NODE-4619/community-issue-postrgres-node-output-large-format-numbers-as-option

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Postgres/transport/index.ts
```diff
@@ -89,6 +89,22 @@ export async function configurePostgres(
 			// prevent spam in console "WARNING: Creating a duplicate database object for the same connection."
 			// duplicate connections created when auto loading parameters, they are closed immediately after, but several could be open at the same time
 			noWarnings: true,
+			// Use per-instance receive event instead of pgp.pg.types.setTypeParser, which mutates
+			// global pg state and would affect all pools regardless of their largeNumbersOutput setting
+			receive(e) {
```
BerniWittmann, line 116: I feel like this function could deserve some tests (especially with all the nested ifs etc)
The current test only tackle the connection pool changes

BerniWittmann (changes requested):
Looking good, works well. 

I just have a concern about test coverage of the `receive` function, maybe we could add some tests there?

BerniWittmann (approved):
Thanks, looking good 🙌

PR #29488, opened 2026-04-29
chore: Jotform text updated

atakandemircioglu:
## Summary

None name changed to "Jotform".

### DIFF
--- a/packages/nodes-base/nodes/JotForm/JotFormTrigger.node.ts
+++ b/packages/nodes-base/nodes/JotForm/JotFormTrigger.node.ts
@@ -20,14 +20,14 @@ interface IQuestionData {
 
 export class JotFormTrigger implements INodeType {
 	description: INodeTypeDescription = {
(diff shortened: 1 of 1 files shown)


Joffcom:
Hi @atakandemircioglu — thanks for taking a look at this! Going through the diff though, this PR drops the word "Trigger" from the JotForm Trigger node's display name (changes `"Jotform Trigger"` → `"Jotform"`).

We'd like to keep "Trigger" in trigger-node names because it's an n8n convention — every trigger node has "Trigger" in its displayName so users can immediately tell trigger nodes apart from action nodes (e.g. "Slack" vs "Slack Trigger", "Pipedrive" vs "Pipedrive Trigger"). Even though the JotForm Trigger is currently the only JotForm node in core, we'd want to keep the naming pattern consistent across the catalog.

The casing of "Jotform" itself is already correct in the codebase. Closing this PR — if you'd like to propose a different display-name improvement we'd be happy to look. 🙏

PR #29493, opened 2026-04-29
feat(core): Transform MCP server configs into dedicated MCP tools

elsmr:
## Summary

Adds a new `mcp-registry` backend module that exposes registered MCP servers as dedicated nodes in the panel, each backed by a shared `McpRegistryClientTool` runtime.

This PR also reworks how modules contribute nodes: the `loadDir` hook on `BackendModule` is replaced with a `nodeLoaders()` contract that returns `NodeLoader` classes. `LoadNodesAndCredentials` now collects loaders from registered modules and runs them alongside the package loaders.

### How to test

1. Start n8n with `N8N_ENABLED_MODULES=mcp-registry`
3. Add it to a workflow connected to an AI Agent and configure the OAuth2 credential. The transport and endpoint are already set
4. Run the workflow and verify tool listing/execution works the same as the existing `MCP Client Tool` node.
5. Confirm the standard `MCP Client Tool` node still works.

https://linear.app/n8n/issue/NODE-4924

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI


packages/workflow/src/interfaces.ts
```diff
@@ -2730,6 +2730,31 @@ type LoadedData<T> = Record<string, LoadedClass<T>>;
...
+/**
+ * Contract that the runtime consumes from each node source. Implemented by the
+ * filesystem-backed `DirectoryLoader` in `n8n-core` and in modules
+ */
+export interface NodeLoader {
+	packageName: string;
+
+	known: KnownNodesAndCredentials;
+	types: NodeLoaderTypes;
+	credentialTypes: ICredentialTypeData;
```
RomanDavydchuk: Not sure we need this, maybe remove this just like `nodeTypes`?

packages/cli/src/modules/mcp-registry/mcp-registry-node-loader.ts
```diff
@@ -0,0 +1,132 @@
...
+ * routed to the `McpRegistryClientTool` runtime class
+ */
+export class McpRegistryNodeLoader implements NodeLoader {
+	packageName = MCP_REGISTRY_PACKAGE_NAME;
+
+	known: KnownNodesAndCredentials = { nodes: {}, credentials: {} };
+
+	types: NodeLoaderTypes = { nodes: [], credentials: [] };
+
+	nodeTypes: INodeTypeData = {};
```
RomanDavydchuk: I think this is no longer needed (was removed from the interface)

packages/cli/src/modules/mcp-registry/mcp-registry-node-loader.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	credentialTypes: ICredentialTypeData = {};
+
+	loadedNodes: INodeTypeNameVersion[] = [];
```
RomanDavydchuk: Same?

packages/cli/src/modules/mcp-registry/node-description-transform.ts
```diff
@@ -0,0 +1,77 @@
...
+): INodeTypeDescription | null {
+	if (server.authType !== 'oauth2') return null;
+
+	const remote = pickRemote(server);
+	if (!remote) return null;
+
+	const displayName = `${server.title} MCP`;
+	const description = structuredClone(baseDescription);
+
+	delete description.hidden;
```
RomanDavydchuk: Maybe we set `hidden` to `true` if the server's `status` is `deprecated`? That way, we generate still generate node types for deprecated servers, as to not break existing workflows, but they won't be available in the Node Creator

packages/cli/src/modules/mcp-registry/registry/mcp-registry.types.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type McpRegistryToolAnnotations = {
+	readOnlyHint?: boolean;
+	destructiveHint?: boolean;
+};
+
+export type McpRegistryTool = {
+	name: string;
+	title?: string;
+	description: string;
```
RomanDavydchuk: Do we keep these two fields or remove them?

packages/cli/src/modules/mcp-registry/registry/mcp-registry.types.ts
```diff
@@ -0,0 +1,46 @@
...
+export type McpRegistryRemoteType = 'streamable-http' | 'sse';
+
+export type McpRegistryRemote = {
+	type: McpRegistryRemoteType;
+	url: string;
+};
+
+export type McpRegistryToolAnnotations = {
+	readOnlyHint?: boolean;
+	destructiveHint?: boolean;
```
RomanDavydchuk: I guess we also don't plan on using the `destructiveHint` in the future, so not sure if we need to keep it

packages/cli/src/modules/mcp-registry/registry/mcp-registry.types.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * The shape of an entry returned by the MCP server registry.
+ */
+export type McpRegistryServer = {
```
RomanDavydchuk, line 4: I think we'd need a field like `updatedAt` - would be useful for determining which configs to load from Strapi when loading them + I think some designs had a field like `Last updated` in the UI

packages/cli/src/modules/mcp-registry/node-description-transform.ts
```diff
@@ -0,0 +1,77 @@
...
+	description.displayName = displayName;
+	description.name = camelCase(server.slug);
+	description.iconUrl = server.icons[0]?.src;
```
RomanDavydchuk: Should we set the `icon` property (with `light` and `dark` fields) instead of the `iconUrl`? Looking at the types we might have this data for some servers

packages/cli/src/modules/mcp-registry/registry/mcp-registry.types.ts
```diff
@@ -0,0 +1,46 @@
...
+	isOfficial: boolean;
+	origin: 'registry';
+	status: 'active' | 'deprecated';
+	tags?: string[];
+};
+
+export type McpRegistryIcon = {
+	src: string;
+	mimeType?: 'image/png' | 'image/jpeg' | 'image/jpg' | 'image/svg+xml' | 'image/webp';
+	sizes?: string[];
```
RomanDavydchuk: I don't think we use these either, but they shouldn't take too much memory, so not sure if we should keep or remove them

packages/cli/src/modules/mcp-registry/registry/mcp-registry.service.ts
```diff
@@ -0,0 +1,20 @@
...
+@Service()
+export class McpRegistryService {
+	// TODO: Implement actual registry fetching and caching
+	private readonly servers = new Map<string, McpRegistryServer>([
+		[notionMockServer.slug, notionMockServer],
+	]);
+
+	getAll(): McpRegistryServer[] {
+		return Array.from(this.servers.values());
+	}
```
RomanDavydchuk: Maybe we can only return servers whose `status` is `active` by default (add a flag like `includeDeprecated = false`)? Just thinking a bit ahead: if we want to expose these servers to the Instance AI, it'd probably use this service to get all available servers, but `deprecated` are probably not going to be useful

RomanDavydchuk (changes requested):
Works as expected and the code looks good mostly! Left a couple of comments, I'd say none of them are critical, just cleanup and potential improvements that are not required to implement, we can discuss this further

packages/cli/src/modules/mcp-registry/registry/mcp-registry.types.ts
elsmr: Removed them, I did keep mimeType, I think it could be useful

packages/cli/src/modules/mcp-registry/registry/mcp-registry.types.ts
elsmr: removed them

packages/@n8n/nodes-langchain/nodes/mcp/McpRegistryClientTool/McpRegistryClientTool.node.ts
```diff
@@ -0,0 +1,182 @@
+import {
```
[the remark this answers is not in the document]
elsmr: This will be handled in another PR

RomanDavydchuk (dismissed):
LGTM

RomanDavydchuk:
@elsmr look like some tests are failing ATM

elsmr:
@RomanDavydchuk pushed fix!

RomanDavydchuk (approved):

PR #29497, opened 2026-04-29
feat(core): Add JWE decryption to OAuth2 credential flow

guillaumejacquart:
## Summary

Adds JWE decryption to the OAuth2 credential flow on both the authorization-code callback (`OAuth2CredentialController.handleCallback()`) and the refresh path (`refreshOrFetchToken` in the execution engine, plus the `clientCredentials` first-time acquisition). Decryption uses the instance JWE key from `OAuthJweKeyService` (shipped in IAM-476) — no per-credential keypairs.

A new `OAuthJweServiceProxy` lives in `packages/cli/src/oauth/`, outside the module; the `oauth-jwe` module wires the real `OAuthJweDecryptService` at `init()` time via a setter (mirroring the `ExecutionRedactionServiceProxy` pattern). The same proxy is exposed to `packages/core` via `additionalData['oauth-jwe']`. Plaintext rejection is response-level: the response must contain at least one JWE-shaped token (`access_token` or `id_token`); fully plaintext responses are rejected. This permits the common OIDC shape where `access_token` is opaque while `id_token` is JWE-wrapped.

Behaviour is gated behind `N8N_ENV_FEAT_OAUTH2_JWE` and a per-credential `jweEnabled` toggle on `OAuth2Api`. A temporary unauthenticated `GET /rest/oauth-jwe/jwks` endpoint is included for local IdP testing — to be replaced when IAM-478 ships the proper JWKS endpoint.

**Testing:** unit tests cover the proxy (flag-off/no-handler short-circuits, delegation, error propagation), the decrypt service (happy path, mixed JWE+plaintext acceptance, all-plaintext rejection, malformed JWE), the callback path (decrypted persist, plaintext rejection rendering, no-op passthrough), and the refresh path (proxy decrypts via `additionalData`, absent-provider no-op, plaintext rejection fails the refresh).

https://linear.app/n8n/issue/IAM-569

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
```diff
@@ -757,6 +757,17 @@ interface RefreshOAuth2TokenContext {
 	helpers: IAllExecuteFunctions['helpers'];
 }
 
+async function maybeDecryptOAuth2TokenData<T extends Record<string, unknown> | undefined>(
```
BGZStephen: Would be nice to avoid the `as unknown as IDataObject` cast, something like this might work:

```
async function maybeDecryptOAuth2TokenData<T extends IDataObject | undefined>(
      additionalData: IWorkflowExecuteAdditionalData,                                               
      tokenData: T,                                      
      jweEnabled: boolean,                                                                          
  ): Promise<T | IDataObject> {                          
      if (!jweEnabled) return tokenData;
      const proxy = additionalData['oauth-jwe']?.oauthJweProxyProvider;                             
      if (!proxy || !tokenData) return tokenData;
      return await proxy.decryptOAuth2TokenData(tokenData);                                         
  }
```

BGZStephen: Going to sound picky: `maybeDecryptOAuth2TokenData` -> `decryptOAuth2TokenDataIfConfigured`

Qodo-Free-For-OSS:
Hi, `OAuthJweModule.init()` calls `OAuthJweKeyService.initialize()` unconditionally and the module is no longer limited to `instanceTypes: ['main']`, so workers will also perform key initialization even when the JWE feature flag is off. This can trigger unnecessary DB/cache work and potentially expensive key generation/insert races during startup.

**Severity:** remediation recommended | **Category:** performance

**How to fix:** Gate init by feature

> ### Issue description
> `OAuthJweModule.init()` runs `OAuthJweKeyService.initialize()` on all instance types and regardless of the `N8N_ENV_FEAT_OAUTH2_JWE` feature flag. This can cause unnecessary startup work and potential key-generation races on workers even when JWE is disabled.
> 
> ### Issue Context
> - ModuleRegistry will call `init()` for modules without `instanceTypes` on both main and worker.
> - The decrypt handler wiring is already gated behind `N8N_ENV_FEAT_OAUTH2_JWE`, but key initialization is not.
> 
> ### Fix Focus Areas
> - packages/cli/src/modules/oauth-jwe/oauth-jwe.module.ts[12-27]
> 
> ### Implementation notes
> Choose one:
> - Move `OAuthJweKeyService.initialize()` inside the `if (isFeatureFlagEnabled()) { ... }` block, if keys are only needed when the feature is on.
> - Or, if keys must exist even when feature is off, restore `instanceTypes` and/or add an `InstanceSettings.instanceType` guard to avoid worker-side key generation and restrict it to main only.

---

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
guillaumejacquart: cool ! thanks

packages/cli/src/oauth/oauth-jwe-service.proxy.ts
```diff
@@ -0,0 +1,31 @@
...
+@Service()
+export class OAuthJweServiceProxy implements OauthJweProxyProvider {
+	private handler?: OAuthJweDecryptHandler;
+
+	setHandler(handler: OAuthJweDecryptHandler) {
+		this.handler = handler;
+	}
+
+	async decryptOAuth2TokenData(tokenData: IDataObject): Promise<IDataObject> {
+		if (!this.handler) return tokenData;
```
afitzek: Should the default case be an error in case we request a decryption but not decryption module is set up?

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
```diff
@@ -757,6 +757,17 @@ interface RefreshOAuth2TokenContext {
 	helpers: IAllExecuteFunctions['helpers'];
 }
 
+async function decryptOAuth2TokenDataIfConfigured<T extends IDataObject | undefined>(
+	additionalData: IWorkflowExecuteAdditionalData,
+	tokenData: T,
+	jweEnabled: boolean,
+): Promise<T | IDataObject> {
+	if (!jweEnabled) return tokenData;
+	const proxy = additionalData['oauth-jwe']?.oauthJweProxyProvider;
```
afitzek, line 766: Is the DI container here not available? I'm not sure why we need the proxy in the additional data.

packages/cli/src/modules/oauth-jwe/oauth-jwe.module.ts
```diff
@@ -1,19 +1,32 @@
...
-@BackendModule({ name: 'oauth-jwe', instanceTypes: ['main'] })
+@BackendModule({ name: 'oauth-jwe' })
 export class OAuthJweModule implements ModuleInterface {
 	async init() {
-		const { OAuthJweKeyService } = await import('./oauth-jwe-key.service');
-		await Container.get(OAuthJweKeyService).initialize();
+		if (!isFeatureFlagEnabled()) return;
+
+		const { OAuthJweDecryptService } = await import('./oauth-jwe-decrypt.service');
+		Container.get(OAuthJweServiceProxy).setHandler(Container.get(OAuthJweDecryptService));
```
afitzek, line 18: NIT: I think a worker might be generating a key because of a race condition: If the worker needs to run a queued workflow in queue mode, it might call getKeyPair and might end up with a cache miss. It shouldn't really matter, but the comment below claims that only mains generate key data.

packages/core/src/execution-engine/node-execution-context/utils/request-helper-functions.ts
guillaumejacquart: Indeed, but the service `OAuthJweServiceProxy` lives in the `cli` package, which is not imported by the core package. that's why we have this pattern:
image

packages/cli/src/modules/oauth-jwe/oauth-jwe.module.ts
guillaumejacquart: good catch. I updated the comment for now

packages/cli/src/oauth/oauth-jwe-service.proxy.ts
guillaumejacquart: good catch

afitzek (approved):

PR #29498, opened 2026-04-29
feat: Add instance-level JWKS URI endpoint for JWE public key distribution

afitzek:
## Summary

Adds a new instance-level JWKS endpoint at `GET /rest/.well-known/jwks.json` that serves all active OAuth JWE public keys from the `deployment_key` table, along with a JWKS URI display in the OAuth2 credential UI so admins can copy it into their IdP configuration.

In enterprise OIDC setups, n8n is typically registered as a single OAuth client whose IdP needs to fetch a JWKS document to encrypt tokens (JWE). One instance-level endpoint matches the IdP's single `jwks_uri` field per client and is configured once for all JWE-enabled credentials. The endpoint is feature-flagged, so instances that have not opted in expose zero new public surface.

## Why

Part of the OAuth JWE credential support initiative ([Notion PoC](https://www.notion.so/n8n/PoC-Meta-OIDC-Custom-Credential-with-JWE-Decryption-3415b6e0c94f80c0bb89d6f66fabac30), [Approach](https://www.notion.so/n8n/3425b6e0c94f80d8b8ecc2178d534a37)). The underlying key generation/storage already exists in `OAuthJweKeyService` (one active private JWK per algorithm in `deployment_key`, with `kid` and a public-field allow-list). This PR wires up the HTTP surface and the credential UI display — no new domain logic.

Key design decisions:
- **Instance-level, not per-credential** — one IdP client → one `jwks_uri`; per-credential endpoints would also leak credential IDs.
- **Feature-flagged at module init** — when `N8N_ENV_FEAT_OAUTH2_JWE` is off, the controller is never imported, so the route is never registered (genuine 404, not a runtime check). Matches the `token-exchange` module pattern.
- **Defense-in-depth Zod schema** — every JWK is validated against a strict (`.strict()`) public-only schema before it reaches the response, on top of the service-side allow-list. Malformed JWKs are dropped with a `logger.warn` so a bad row never breaks the endpoint for the rest of the keys.
- **Forward-looking schemas** — discriminated union on `kty` with both RSA and EC variants, even though only RSA-OAEP-256 is generated today. Adding EC support is a service-side change only.
- **Server-supplied absolute URL for the UI** — `jwksUri` is constructed in `frontend.service.ts` using `urlService.getInstanceBaseUrl()`, mirroring how `oauthCallbackUrls` flow. This honours the configured public host (e.g. behind a reverse proxy) instead of constructing a relative path on the frontend.
- **Path: `/rest/.well-known/jwks.json`** — closer to OIDC convention than `/rest/oauth2/jwks.json`. Still under `/rest`, so a strict OIDC discovery client must use it via explicit `jwks_uri` configuration (which is the intended flow).

## How to test manually

Prerequisites: a fresh n8n main with the JWE module initialised (it auto-generates an `RSA-OAEP-256` key in `deployment_key` on first boot).

**Flag OFF (default):**
```bash
pnpm start
curl -i http://localhost:5678/rest/.well-known/jwks.json
# Expected: 404 Not Found — route is not registered
```

**Flag ON:**
```bash
N8N_ENV_FEAT_OAUTH2_JWE=true pnpm start
curl -i http://localhost:5678/rest/.well-known/jwks.json
# Expected: 200 OK with:
#   Cache-Control: public, max-age=3600, must-revalidate
#   Content-Type: application/json; charset=utf-8
# Body:
#   { "keys": [ { "kty": "RSA", "kid": "...", "use": "enc", "alg": "RSA-OAEP-256", "n": "...", "e": "AQAB" } ] }
```

**Rate limit (optional):**
- Hit the endpoint > 60 times/minute from the same IP. The 61st call should be rejected with 429. Tunable via `N8N_OAUTH_JWE_JWKS_PER_MINUTE`.

**UI verification:**
1. With `N8N_ENV_FEAT_OAUTH2_JWE=true`, open the editor, create or open an **OAuth2 API** credential.
2. Confirm a new copyable field **OAuth2 JWKS URI** appears below the existing **OAuth Redirect URL**.
3. Confirm the value is the **absolute** URL (e.g. `http://localhost:5678/rest/.well-known/jwks.json`), not a relative path. Click → toast confirms copy.
4. Restart with the flag unset; the JWKS URI field should not render.

**Security:**
- Decode the JWK from the response and confirm only public RSA fields are present (`kty, kid, use, alg, n, e`). No `d`, `p`, `q`, `dp`, `dq`, `qi`. The Zod schema (`.strict()`) drops anything else, and a unit test asserts this on a JWK that contains a leaked `d` field.

## Related Links

- closes https://linear.app/n8n/issue/IAM-478

## Notes for reviewers

- The spec example uses `"alg": "RSA-OAEP"`; the implementation actually serves whatever `OAuthJweKeyService` generates, which is `RSA-OAEP-256` (the more secure AES-key-wrap variant). The endpoint is intentionally a passthrough of what the service produces.

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/credentials/components/CredentialEdit/CredentialConfig.vue
```diff
@@ -447,6 +450,17 @@ watch(showOAuthSuccessBanner, (newValue, oldValue) => {
 						"
 						:redact-value="true"
 					/>
+
+					<EnvFeatureFlag name="OAUTH2_JWE">
```
guillaumejacquart, line 454: That's fine I guess for now, but eventually we will need to add a condition to display this only if "jwe encryted" toggle is enabled

guillaumejacquart (approved):
Nice ! one nit comment

PR #29500, opened 2026-04-29
fix(ai-builder): Handle properties with contradicting displayOptions as OR alternatives instead of AND

Cadiac:
## Summary

Properties declared multiple times with the same name are OR alternatives - the BigQuery `sqlQuery` field shown for useLegacySql SQL is the canonical example, declared in [`packages/nodes-base/nodes/Google/BigQuery/v2/actions/database/executeQuery.operation.ts`](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Google/BigQuery/v2/actions/database/executeQuery.operation.ts) as:

```ts
{
    displayName: 'SQL Query',
    name: 'sqlQuery',
    type: 'string',
    displayOptions: {
        hide: { '/options.useLegacySql': [true] },
    },
    placeholder: 'SELECT * FROM dataset.table LIMIT 100',
    description: 'SQL query to execute … Standard SQL syntax used by default …',
},
{
    displayName: 'SQL Query',
    name: 'sqlQuery',
    type: 'string',
    displayOptions: {
        show: { '/options.useLegacySql': [true] },
    },
    placeholder: 'SELECT * FROM [project:dataset.table] LIMIT 100;',
    hint: 'Legacy SQL syntax',
    description: 'SQL query to execute … about Legacy SQL syntax …',
},
```

The schema generator was naively union-merging the variants' `show`/`hide` independently, producing a self-contradicting predicate that rejected every valid configuration. The original repro trace showed instance AI burning 4 `submit-workflow` retries on `[INVALID_PARAMETER] sqlQuery: only allowed when /options.useLegacySql=true` before giving up and setting `useLegacySql: true` as an incorrect workaround (the query was standard SQL).

The agent was also reading a `.d.ts` whose JSDoc said:

```
@displayOptions.show { resource: ["database"], operation: ["executeQuery"], /options.useLegacySql: [true] }
@displayOptions.hide { /options.useLegacySql: [true] }
```

- directly self-contradictory, reinforcing the wrong mental model.

## What this PR fixes

This PR fixes both layers:

- **Runtime validator**: new `resolveOneOfSchemas` helper picks the first variant whose displayOptions match at evaluation time, with full DSL support (`@version`, `@tool`, `_cnd` operators, root paths) and correct required-when-visible enforcement (a naive `z.union` of per-variant `resolveSchema` calls would silently mask required because each inactive variant's "must-be-undefined" branch matches `undefined`).
- **JSDoc emitter**: `tryMergeUxForkVariants` collapses two-variant partitions into a single honest predicate so the `.d.ts` agents read no longer contradicts the validator. Handles the clean-partition case (BigQuery: same values on hide and show ⇒ drop the key) and the subset case (Pushbullet: `hide: K=[a,b]` + `show: K=[a]` ⇒ `hide: K=[b]`). Multi-key forks fall through and keep variant 1; the runtime path still uses `resolveOneOfSchemas` so validation stays correct.
- **String-operator hardening**: `_cnd.includes` / `startsWith` / `endsWith` / `regex` now return `false` instead of crashing when the referenced parameter isn't a string. Split variants now reach these operators where the old AND-merge happened to short-circuit past them (e.g. evaluating OpenAI image-edit's `/modelId: { _cnd: { includes: 'gpt-image' } }` when `modelId` is unset).

After the fix, the same BigQuery prompt produces zero `INVALID_PARAMETER sqlQuery` errors and the agent correctly leaves `useLegacySql` unset.

The simplified JSDoc reads:

```
@displayOptions.show { resource: ["database"], operation: ["executeQuery"] }
```

## Coverage across the 12 originally-flagged nodes

- **Simplifier kicks in** (clean .d.ts): BigQuery v2/v2.1 (`sqlQuery`), Git v1/v1.1 (`repositoryPath`), Firebase RealtimeDB v1 (`path`), Summarize v1/v1.1 (`fieldsToSplitBy`), Pushbullet (`value`).
- **False positives in original scan** (single declarations with deliberate symmetric show/hide, no fix needed): CoinGecko, Flow Trigger.
- **Falls through JSDoc simplifier** (runtime still correct via `resolveOneOfSchemas`): Cortex (`parameters`) — 5-variant `entityType` fork with extra `responder` constraints. The pairwise simplifier bails on multi-key forks; documented limitation.

https://linear.app/n8n/issue/INS-145

## Test plan

- [x] Unit tests on `tryMergeUxForkVariants` covering clean partition, subset partition, shared+fork, multi-key fork (returns null), and missing-displayOptions cases.
- [x] Unit tests on `resolveOneOfSchemas` covering 0/1/many visible variants, required-when-visible, and the combined "A OR B …" error message.
- [x] Integration tests via `generateNodeTypeFile` exercising both the single-version codegen path (BigQuery shape) and the discriminator-split path (Pushbullet shape).
- [x] Verified end-to-end against real instance AI traces (before/after on the same BigQuery prompt).

## Review / Merge checklist

- [x] Tests included.


mutdmour (approved):
LGTM

PR #29506, opened 2026-04-29
feat(core): Add built-in cluster health checks (no-changelog)

afitzek:
## Summary

Adds four built-in cluster health checks on top of the reconciliation infrastructure shipped in #28936. Each check runs every 3 minutes on the leader, evaluates a snapshot of the cluster state, and emits warnings (surfaced via the `/instance-registry` REST endpoint) and audit events (delivered to log-streaming destinations) when issues are detected.

- **Version Mismatch** — instances running different n8n versions (typical during rolling upgrades).
- **HostId Clash** — multiple instances sharing the same `hostId` (cloned VMs or misconfiguration).
- **Split-Brain** — multiple instances claiming the `leader` role (data-corruption risk).
- **Lifecycle** — audit trail of instances joining or leaving the cluster.

### Why

This is the second slice of the cluster-visibility initiative. #28936 added the reconciliation loop and the `@ClusterCheck()` decorator infrastructure, plus a stub `VersionMismatchCheck` so the loop had something to run. This PR replaces that stub with the spec'd version (audit events + dedup) and adds the remaining three checks. After this PR, operators get out-of-the-box detection of the most common distributed-deployment failure modes.

**Stateless deduplication via `previousState` fingerprinting.** The three stateful checks (version-mismatch, hostid-clash, split-brain) compute a deterministic fingerprint of their finding from `context.currentState` and compare it to one derived from `context.previousState`. Same fingerprint → warning only (so the REST `checks` summary stays accurate); changed → fresh audit event; cleared → one-shot `*.resolved` audit event; then silence. No `CacheService`, no in-memory state on the check — survives leader handoffs cleanly because `previousState` is already persisted by `InstanceRegistryService.saveLastKnownState`.

**Lifecycle check uses `diff.added` / `diff.removed` directly.** No fingerprinting — the diff is already one-shot per cycle.

**Direct audit emission, no `EventService` relay.** Checks declaratively describe their audit events via `ClusterCheckResult.auditEvents`; `CheckService` forwards them to `MessageEventBus.sendAuditEvent()` directly. This diverges from the typical `EventService.emit() → @Redactable() relay → sendAuditEvent()` pattern but is appropriate here because checks are pluggable and own their audit semantics. Cluster payloads contain no user PII (instance keys, hostIds, versions, leader info), so redaction isn't needed. The 8 cluster audit names are registered in `eventNamesAudit`; from there they reach log-streaming destinations via prefix match (`hasSubscribedToEvent` in `message-event-bus-destination.ee.ts`) — no relay map entry required.

**No push notifications.** Earlier draft of this PR added a `ClusterCheckPushMessage` push type for real-time UI updates; removed after team discussion. The REST endpoint already exposes `checks` summaries, and audit events cover the long-lived observability story.

closes  https://linear.app/n8n/issue/IAM-333

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/modules/instance-registry/checks/version-mismatch.check.ts
```diff
@@ -1,39 +1,59 @@
...
+const AUDIT_RESOLVED = 'n8n.audit.cluster.version-mismatch.resolved';
+
+/**
+ * Returns the set of distinct versions running in the cluster as a
+ * deterministic fingerprint (sorted, pipe-joined). Returns an empty string
+ * when there is only a single version, indicating "no mismatch".
+ */
+function computeFingerprint(instances: Iterable<InstanceRegistration>): string {
+	const versions = [...new Set([...instances].map((i) => i.version))].sort();
+	return versions.length > 1 ? versions.join('|') : '';
```
guillaumejacquart, line 20: I don't see the value of returning a string here 🤔 . Is it just to follow the same fingerprint pattern ?

packages/cli/src/modules/instance-registry/checks/split-brain.check.ts
```diff
@@ -0,0 +1,72 @@
...
+	checkDescription = {
+		name: 'split-brain',
+		displayName: 'Split-brain',
+	};
+
+	async run(context: ClusterCheckContext): Promise<ClusterCheckResult> {
+		const current = computeFingerprint(context.currentState.values());
+		const previous = computeFingerprint(context.previousState.values());
+
+		if (current.fingerprint === '') {
```
guillaumejacquart: I think this empty string comparison hides a boolean behavior that makes it harder to read. Wouldn't it be simpler if the computeFingerprint returned another field, a boolean if split brain is detected ?

packages/cli/src/modules/instance-registry/checks/lifecycle.check.ts
```diff
@@ -0,0 +1,58 @@
...
+export class LifecycleCheck implements IClusterCheck {
+	checkDescription = {
+		name: 'lifecycle',
+		displayName: 'Cluster membership',
+	};
+
+	async run(context: ClusterCheckContext): Promise<ClusterCheckResult> {
+		const auditEvents: ClusterCheckAuditEvent[] = [];
+
+		for (const joined of context.diff.added) {
```
guillaumejacquart, line 40: i'm fine with that, but i'm just wondering if it makes sense for this to be a check. Another way would be to send the event where the diff is computed, no ?

packages/cli/src/modules/instance-registry/checks/hostid-clash.check.ts
```diff
@@ -0,0 +1,82 @@
...
+} {
+	const byHost = new Map<string, string[]>();
+	for (const instance of instances) {
+		const keys = byHost.get(instance.hostId) ?? [];
+		keys.push(instance.instanceKey);
+		byHost.set(instance.hostId, keys);
+	}
+
+	const clashing: Array<{ hostId: string; instanceKeys: string[] }> = [];
+	for (const [hostId, instanceKeys] of byHost) {
```
guillaumejacquart: I've seen look doing more functional style on other checks.
What about:
```ts
const clashing = Array.from(byHost.entries())
        .filter(([_, keys]) => keys.length > 1)
        .map(([hostId, keys]) => ({
            hostId,
            instanceKeys: [...keys].sort()
        }))
        .sort((a, b) => a.hostId.localeCompare(b.hostId));
```

guillaumejacquart (commented):
some minor comments and question. Looks good though

packages/cli/src/modules/instance-registry/checks/version-mismatch.check.ts
afitzek: I wanted to follow the same fingerprint pattern in all checks yes.

packages/cli/src/modules/instance-registry/checks/lifecycle.check.ts
afitzek: I think this is separation of concerns, if we send the event when the diff is computed, than this check leaks into the reconciliation loop.

packages/cli/src/modules/instance-registry/checks/version-mismatch.check.ts
afitzek: But if the versions jump from 2 different versions to 3 different version for example the fingerprints would also mismatch.

packages/cli/src/modules/instance-registry/checks/lifecycle.check.ts
guillaumejacquart: I see, all right

BGZStephen (approved):

PR #29527, opened 2026-04-29
feat(core): Decouple insights pruning max age from license

sandra0503:
- Default N8N_INSIGHTS_MAX_AGE_DAYS is 365.
- Caps max age at 730 days (INSIGHTS_MAX_AGE_DAYS_CAP).
- Legacy -1 maps to 730.
- Invalid / non-finite / < 1 (except -1) fall back to the default.

## Summary

Deletion of self-hosted insights data is now controlled only by the environment variable (`N8N_INSIGHTS_MAX_AGE_DAYS`). License max age is no longer combined with the env var. The default retention window is 365 days, with a maximum of 730 days. The legacy -1 value is treated as maximum, so behavior stays consistent and there is no separate “unlimited” mode via this variable.

### Context
Previously, effective pruning used `min(license quota: insights retention max age, env)`, with a common license default of 180 days when the env default was -1. That coupling is removed.

### Impact

| What the customer has today | Parsed config | Effective days kept before prune | Notes |
|------------------------------|----------------|-----------------------------------|--------|
| **Unset** (never defined) | default `365` | **365** | Default is **365**; not license-derived. Customers who previously had no env and saw ~180 from license logic will retain **more** data unless they set `180` explicitly. |
| **Positive integer** `1`–`730` | same as env | **Same as set** | Explicit value is used as-is. |
| **Integer > 730** (e.g. `9999`) | number | **730** | Capped at the maximum allowed retention. |
| **`-1`** (legacy “max” / unlimited style) | `-1` | **730** | Mapped to the cap (same as maximum). |
| **`0` or other < 1** (not `-1`) | e.g. `0` | **365** | Falls back to the default. |
| **Not a finite number** (invalid / bad coercion) | non-finite | **365** | Invalid values use the default. |
## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/insights/insights-pruning.service.ts
+++ b/packages/cli/src/modules/insights/insights-pruning.service.ts
@@ -1,10 +1,11 @@
-import { LicenseState, Logger } from '@n8n/backend-common';
+import { Logger } from '@n8n/backend-common';
 import { Time } from '@n8n/constants';
 import { Service } from '@n8n/di';
 import { strict } from 'assert';
 
 import { InsightsByPeriodRepository } from './database/repositories/insights-by-period.repository';
 import { InsightsConfig } from './insights.config';
+import { INSIGHTS_MAX_AGE_DAYS_CAP, INSIGHTS_MAX_AGE_DAYS_DEFAULT } from './insights.constants';
 
 @Service()
 export class InsightsPruningService {
@@ -17,23 +18,26 @@ export class InsightsPruningService {
 	constructor(
 		private readonly insightsByPeriodRepository: InsightsByPeriodRepository,
 		private readonly config: InsightsConfig,
-		private readonly licenseState: LicenseState,
 		private readonly logger: Logger,
 	) {
 		this.logger = this.logger.scoped('insights');
 	}
 
-	get isPruningEnabled() {
-		return this.licenseState.getInsightsRetentionMaxAge() > -1 || this.config.maxAgeDays > -1;
-	}
-
 	get pruningMaxAgeInDays() {
-		const toMaxSafeIfUnlimited = (days: number) => (days === -1 ? Number.MAX_SAFE_INTEGER : days);
+		const configuredMaxAgeDays = this.config.maxAgeDays;
+		if (typeof configuredMaxAgeDays !== 'number' || !Number.isFinite(configuredMaxAgeDays)) {
+			return INSIGHTS_MAX_AGE_DAYS_DEFAULT;
+		}
 
-		const licenseMaxAge = toMaxSafeIfUnlimited(this.licenseState.getInsightsRetentionMaxAge());
-		const configMaxAge = toMaxSafeIfUnlimited(this.config.maxAgeDays);
+		if (configuredMaxAgeDays === -1) {
+			return INSIGHTS_MAX_AGE_DAYS_CAP;
+		}
 
-		return Math.min(licenseMaxAge, configMaxAge);
+		if (configuredMaxAgeDays < 1) {
+			return INSIGHTS_MAX_AGE_DAYS_DEFAULT;
+		}
+
+		return Math.min(configuredMaxAgeDays, INSIGHTS_MAX_AGE_DAYS_CAP);
 	}
 
 	startPruningTimer() {
--- a/packages/cli/src/modules/insights/insights.config.ts
+++ b/packages/cli/src/modules/insights/insights.config.ts
@@ -1,5 +1,7 @@
 import { Config, Env } from '@n8n/config';
 
+import { INSIGHTS_MAX_AGE_DAYS_DEFAULT } from './insights.constants';
+
 @Config
 export class InsightsConfig {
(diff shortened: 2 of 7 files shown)


packages/cli/src/modules/insights/insights-pruning.service.ts
```diff
@@ -17,23 +18,26 @@ export class InsightsPruningService {
...
 	get pruningMaxAgeInDays() {
-		const toMaxSafeIfUnlimited = (days: number) => (days === -1 ? Number.MAX_SAFE_INTEGER : days);
+		const configuredMaxAgeDays = this.config.maxAgeDays;
+		if (typeof configuredMaxAgeDays !== 'number' || !Number.isFinite(configuredMaxAgeDays)) {
+			return INSIGHTS_MAX_AGE_DAYS_DEFAULT;
+		}
 
-		const licenseMaxAge = toMaxSafeIfUnlimited(this.licenseState.getInsightsRetentionMaxAge());
-		const configMaxAge = toMaxSafeIfUnlimited(this.config.maxAgeDays);
+		if (configuredMaxAgeDays === -1) {
```
konstantintieber, line 32: I'd say let's rather require max age days to be a positive number.
Disabling pruning has been effectively not supported because our [LicenseState](https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/backend-common/src/license-state.ts#L226) service implementation would always return 180 days as fallback if no value was configured on the license.

In the spirit of shipping features incrementally, I'd argue we should only add support for configuring a positive number for max retention.
Personally I don't expect we'll see customers asking for a `-1` option in addition to that.

konstantintieber: From slack: @sandra0503 made the point that users who have set `N8N_INSIGHTS_MAX_AGE_DAYS` to -1 will effectively disable pruning.
So I guess we need to keep supporting it.
As of today, setting that env var to `-1` means a max retention of `Number.MAX_SAFE_INTEGER`.
This PR changes this to 730.
Why do we decrease it from `Number.MAX_SAFE_INTEGER` to `730`?

sandra0503: We could also treat -1 like other unsupported values and fall back to default; we only need -1 → 730 if we explicitly want to support ‘as long as allowed’ from older settings.

Should we set a maximum for all users or do we want to allow 'disable pruning'? This ticket suggested that we want to cap it: https://linear.app/n8n/issue/LIGO-494/adjust-the-max-time-range-limit-on-the-instance-metric-environment

I suggested 730 days based on the spike findings

konstantintieber: And changing the effect of `-1` from basically infinite retention, since it's javascript's max integer, to 730 is not a breaking change?
EDIT: I guess one could say it's not, since insights hasn't been around more than two years :D

sandra0503: Until now the value from the license key took precedence: `min(license quota: insights retention max age, env)`. So it was not possible to have infinite retention

sandra0503: So I would say it's not a breaking change

konstantintieber: > Until now the value from the license key took precedence: min(license quota: insights retention max age, env). So it was not possible to have infinite retention

But if the env value would be -1, than that one would be applied, because it was lower than the license value, no?

sandra0503: Actually what happened is that -1 gets converted to `MAX_SAFE_INTEGER` and then 180 is lower

konstantintieber: Ooooh yes, my bad.

packages/cli/src/modules/insights/insights.config.ts
```diff
@@ -45,11 +47,11 @@ export class InsightsConfig {
 	flushIntervalSeconds: number = 30;
 
 	/**
-	 * How old (days) insights data must be to qualify for regular deletion
-	 * Default: -1 (no pruning)
+	 * How old (days) insights data must be to qualify for regular deletion.
+	 * Default: 365. Values are capped at 730 (two years). Legacy `-1` uses that maximum. Values below 1 (other than `-1`) use the default.
```
konstantintieber: This "Legacy `-1` uses that maximum." text in this doc is hard to understand if you don't know the context of the old implementation.
Should we just remove it?

sandra0503: Yes, let's remove that!

sandra0503: Removed, we can add more documentation in n8n-docs

konstantintieber (approved):
LGTM
These docs are affected by this: https://docs.n8n.io/insights/#disable-or-configure-insights-metrics-collection

PR #29536, opened 2026-04-29
feat(agents): Expose triggers configuration in agent JSON config

burivuhster:
## Summary

Exposes scheduled and chat integration triggers through `AgentJsonConfig.integrations` so the builder agent can read and modify them via `write_config` / `patch_config`, the same way it manages `tools` today.

- **API types & validation**: new `integrations` discriminated union (`schedule` + chat platforms) added to `AgentJsonConfig` with Zod schema. Chat entries now carry `credentialName` next to `credentialId`. The legacy "Connect" controller endpoint resolves and stores `credentialName` so existing UI flows match the new schema shape.
- **Storage stays put**: `agent.integrations` column remains the source of truth — no DB migration. `composeJsonConfig` / `decomposeJsonConfig` merge it with `agent.schema` on read and split it back on write.
- **Lifecycle sync**: `updateConfig` now diffs old vs new integrations and drives `AgentScheduleService.applyConfig` (cron registration) and `ChatIntegrationService.syncToConfig` (connect/disconnect chat platforms) so the runtime tracks the persisted config without requiring a publish.
- **Builder integration**: new `list_integration_types` tool for discovering available chat platforms plus an `INTEGRATIONS_SECTION` in the system prompt explaining the workflow (`list_integration_types` → `ask_credential` → `patch_config`).

**How to test:**
1. Open an agent in the builder and ask "What triggers does my agent have?" — the LLM should report the actual integrations.
2. On an **unpublished** agent, ask "Connect this agent to my Slack workspace" — the LLM calls `list_integration_types`, then `ask_credential`, then `patch_config`. The integration is persisted but not connected; check the integrations status endpoint to confirm.
3. Publish the agent — the Slack connection now activates (status flips to `connected`, webhook live).
4. Ask the LLM to set the cron to `not-a-cron` — `patch_config` fails at the schema stage with an "Invalid cron expression" error.
5. Ask the LLM to set `active: true` on an unpublished agent — `patch_config` is rejected with a publish-gate error.

## Out of scope / follow-ups

- **`AgentPublishedVersion` snapshot** — integrations still live on the live entity, not the published version. Existing schedule activation requires a published agent; that contract is unchanged.
- **`AgentScheduleTriggerCard` local-state dance** — the card maintains `lastSavedCronExpression` / `hydrating` flags to coordinate its debounced save against server fetches. A Pinia-backed store would let cards subscribe to a single source of truth and remove that local plumbing entirely.
- **`AgentIntegrationsPanel.vue` cleanup** — the component is no longer mounted in production after the `AgentBuilderEditorColumn` restructure on `n8n-agents` (only test mocks reference it). Removal is out of scope for this PR.

https://linear.app/n8n/issue/AGENT-60

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/modules/agents/agents.service.ts
```diff
@@ -46,8 +46,11 @@ import { AgentsToolsService } from './agents-tools.service';
 import { Agent } from './entities/agent.entity';
 import { ExecutionRecorder } from './execution-recorder';
 import { ChatIntegrationRegistry } from './integrations/agent-chat-integration';
+// eslint-disable-next-line import-x/no-cycle
```
bjorger: Is that disable necessary?

.gitignore
```diff
@@ -59,6 +59,7 @@ packages/cli/src/commands/export/outputs
 .data/
 .claude/settings.local.json
 .claude/plans/
+.claude/specs/agent-60-integrations-as-json-config.md
```
bjorger: This should probably be a wildcard / catch all rule and not just for one spec :)

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
```diff
@@ -195,6 +200,79 @@ export class ChatIntegrationService {
...
+	/**
+	 * Diff the previous and next chat integrations of an agent and reconcile
+	 * runtime connections accordingly. Used by `AgentsService.updateConfig`
+	 * after the builder writes a new integrations array.
+	 *
+	 * Connection failures are logged at the call site — this method propagates
+	 * errors from disconnect but swallows connect errors per integration so a
+	 * single bad credential doesn't block the others.
+	 */
+	async syncToConfig(
```
bjorger, line 221: syncToConfig() connects newly added chat integrations without checking whether the agent is published. The existing connect endpoint rejects unpublished agents, and startup reconnect only restores published agents, so this builder JSON path can open a live chat/webhook connection for an unpublished agent. If that publish gate is still intended, can we skip runtime connects here when !agent.publishedVersion?

packages/cli/src/modules/agents/json-config/integration-config.ts
```diff
@@ -0,0 +1,32 @@
+import { z } from 'zod';
+
+export const AgentScheduleIntegrationSchema = z
+	.object({
+		type: z.literal('schedule'),
+		active: z.boolean(),
+		cronExpression: z.string().min(1, 'cronExpression is required'),
+		wakeUpPrompt: z.string().min(1, 'wakeUpPrompt is required'),
```
bjorger, line 13: The JSON config path allows schedule integrations to set active: true directly, bypassing the schedule endpoint’s activation rules. It also only checks cronExpression is non-empty rather than validating cron syntax. That can leave persisted schedule config saying active even if runtime registration refuses or fails. Can we share the schedule endpoint validation/invariants here?

packages/cli/src/modules/agents/builder/agents-builder-tools.service.ts
```diff
@@ -165,9 +165,27 @@ export class AgentsBuilderToolsService {
 			})
 			.build();
 
+		const listIntegrationTypesTool = new Tool(BUILDER_TOOLS.LIST_INTEGRATION_TYPES)
+			.description(
+				"List trigger / integration types that can be added to the agent's `integrations` array. " +
+					'Returns the schedule trigger plus every connected chat platform with its credential type(s). ' +
+					'Call this BEFORE asking the user for a credential — pass the returned `credentialTypes` ' +
```
bjorger: This wording can make the LLM call ask_credential with the wrong shape. list_integration_types returns credentialTypes: string[], but ask_credential accepts singular credentialType: string. For Slack/Linear, multiple credential types may be returned. Can we change this to say to pick one returned credential type and pass it as credentialType?

packages/cli/src/modules/agents/integrations/chat-integration.service.ts
burivuhster: Good catch! Added debug warning and early return

packages/cli/src/modules/agents/integrations/agent-schedule.service.ts
```diff
@@ -159,6 +160,20 @@ export class AgentScheduleService {
...
 
+	/**
+	 * Idempotent apply: read the schedule integration off the agent and reconcile
+	 * the cron job. Used when the JSON config writes a new schedule via the
+	 * builder so the live runtime tracks the persisted config.
+	 */
+	async applyConfig(agent: Agent): Promise<void> {
+		const schedule = this.getScheduleIntegration(agent);
+		if (schedule?.active) {
+			await this.registerOrRefresh(agent);
```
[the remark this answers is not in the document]
burivuhster: Multi-main will be addressed in a separate PR

bjorger (approved):
LGTM 🚀

PR #29537, opened 2026-04-29
fix(core): Restore peer project discovery in share dropdowns

### LINKED ISSUE #29002: "No matching user" when trying to share a personal workflow since updating to 2.17.5/6
### Bug Description

The non-admin users from our organization are experiencing issues when trying to share their personal workflows. When having to select a user to share the workflow with, none of the existing users are being listed or found by writing their user names.
When accessing a workflow, I can see the following error in the console:
```
ResponseError: Listing all users is limited to instance administrators and project admins. Filter by project to list project members.
```
Image
Image

### To Reproduce

1. Go to a personal workflow.
2. Click on Share
3. Look for another user in the organization.

### Expected behavior

Get the list of users when trying to share a personal workflow.

### Debug Info

# Debug info

## core

- n8nVersion: 2.17.6
- platform: docker (cloud)
- nodeJsVersion: 24.14.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: 50

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 720 hours
- maxCount: 25000 executions

## client

- isTouchDevice: false

Generated at: 2026-04-23T14:26:52.445Z

### Operating System

mac os x 10_15_7

### n8n Version

2.17.6

### Node.js Version

24.14.1

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

afitzek:
## Summary

`global:member` users could not share personal workflows or credentials: the
"add users" field in the share modal was empty, regardless of the Personal
Space Sharing policy. This restores peer-personal-project discovery for the
share dropdowns by routing them through a new dedicated endpoint, while
keeping the existing `GET /rest/projects` listing intentionally strict.

### Why

The share modal is wired to a server-paginated project search introduced in
#27036, which originally returned every personal project plus the caller's
own relations — exactly what the dropdown needs to surface other users as
share targets.

#27206 (Instance AI module) bundled a change to
`ProjectRepository.getAccessibleProjectsAndCount` that tightened the query
to `INNER JOIN projectRelations WHERE userId = :userId`. The same tightening
was also applied to `getAccessibleProjects` (non-paginated), where it was
correct: that variant is consumed by the Instance AI agent's `listProjects`
tool and the favorites filter set, both of which want only projects the
caller can manipulate. The paginated variant, however, backs the share
dropdowns — conflating the two semantics broke peer discovery for members.

This PR makes the two semantics first-class so they can't be silently
conflated again:

- `getAccessibleProjectsAndCount` keeps its post-#27206 strict semantics
  ("projects I can manipulate"). Backs `GET /rest/projects`.
- `getShareableProjectsAndCount` (new) restores the wide semantics
  ("projects I can pick as share targets, including peer personal
  projects"). Backs `GET /rest/projects/sharing-candidates`.
- The share / transfer / user-deletion modal composable
  (`useRemoteProjectSearch`) is switched to call the new endpoint.

The new endpoint is gated by `@GlobalScope('user:list')` — the same boundary
`GET /rest/users` enforces, since the dropdown surfaces user identity
(`First Last <email>` is the personal project name). Members already carry
this scope; restricted roles like chat-only users do not and receive 403.

### How to test manually

1. Run an Enterprise build (license with Personal Space Sharing).
2. Create one instance owner and at least two members (e.g. Alice, Bob);
   complete signup so both members have passwords.
3. Sign in as Alice → open any of Alice's personal workflows → Share.
4. Click the "add users…" field. **Expected:** Bob's personal project (and
   any other peer personal projects) appears in the dropdown.
5. Type `bob` in the field. **Expected:** the dropdown narrows server-side
   and Bob remains visible.
6. Repeat steps 3–4 from a credential's Sharing tab. Same expected
   behaviour.
7. Sign in as the instance owner → Share modal → confirm the full project
   list still appears (no admin regression).
8. In DevTools, confirm the share dropdown issues
   `GET /rest/projects/sharing-candidates?...take=50&...` (200) and not
   `GET /rest/projects`. Confirm the response `data[]` contains every
   member's personal project.
9. Optional negative check: as a chat-only user, request
   `GET /rest/projects/sharing-candidates` directly. **Expected:** 403.

- **Dedicated endpoint instead of a query-param flag.** The two list
  semantics are genuinely different, so they get distinct names at every
  layer (route, service, repository).
- **Don't touch `getAccessibleProjects` (non-paginated).** Its tightening
  in #27206 is correct and load-bearing for Instance AI and favorites.
- **Don't gate the listing on the Personal Space Sharing license/policy.**
  That license controls whether a user can be the *source* of sharing from
  their personal space; it's enforced at write-time on the share endpoints.
  Visibility of peer personal projects in the listing is an orthogonal
  concern, consistent with the design from #27036.

## Related Links

- closes https://linear.app/n8n/issue/IAM-591
- Closes #29002
- Regressing commit: `629826ca1d` (PR #27206)
- Original correct design: `3b00e1bafe` (PR #27036)

## Review / Merge checklist

- [X] Tests included. 

### DIFF
--- a/packages/@n8n/db/src/repositories/project.repository.ts
+++ b/packages/@n8n/db/src/repositories/project.repository.ts
@@ -53,6 +53,11 @@ export class ProjectRepository extends Repository<Project> {
 		return await query.getManyAndCount();
 	}
 
+	// Strict semantics: returns only projects the user has a relation to
+	// (their personal project + projects they are explicitly a member of).
+	// Do not broaden — peer-personal-project discovery for the share modal lives
+	// in `getShareableProjectsAndCount` below; conflating the two has regressed
+	// the share dropdown before (see IAM-591).
 	async getAccessibleProjectsAndCount(
 		userId: string,
 		options: ProjectListOptions,
@@ -62,6 +67,40 @@ export class ProjectRepository extends Repository<Project> {
 			.innerJoin('p.projectRelations', 'pr')
 			.where('pr.userId = :userId', { userId });
 
+		this.applyIdsQueryFilters(idsQuery, options);
+		return await this.runProjectListByIdsQuery(idsQuery, options);
+	}
+
+	// Wide semantics: returns peer personal projects in addition to projects
+	// the user has a relation to. Used only by the sharing-discovery endpoint
+	// (`GET /rest/projects/sharing-candidates`) so the workflow / credential
+	// share dropdowns can list other users as share targets.
+	async getShareableProjectsAndCount(
+		userId: string,
+		options: ProjectListOptions,
+	): Promise<[Project[], number]> {
+		// DISTINCT + LEFT JOIN avoids duplicate rows from the relation join
+		// while still allowing personal projects with no caller relation to match.
+		const idsQuery = this.createQueryBuilder('p')
+			.select('DISTINCT p.id', 'id')
+			.leftJoin('p.projectRelations', 'pr')
+			.where(
+				new Brackets((qb) => {
+					qb.where('p.type = :personalType', { personalType: 'personal' }).orWhere(
+						'pr.userId = :userId',
+						{ userId },
+					);
+				}),
+			);
+
+		this.applyIdsQueryFilters(idsQuery, options);
+		return await this.runProjectListByIdsQuery(idsQuery, options);
+	}
+
+	private applyIdsQueryFilters(
+		idsQuery: SelectQueryBuilder<Project>,
+		options: ProjectListOptions,
+	): void {
 		if (options.search) {
 			idsQuery.andWhere('LOWER(p.name) LIKE LOWER(:search)', {
 				search: `%${options.search}%`,
@@ -81,7 +120,12 @@ export class ProjectRepository extends Repository<Project> {
 				}),
 			);
 		}
+	}
 
+	private async runProjectListByIdsQuery(
+		idsQuery: SelectQueryBuilder<Project>,
+		options: ProjectListOptions,
+	): Promise<[Project[], number]> {
 		const query = this.createQueryBuilder('project')
 			.leftJoin('project.creator', 'creator')
 			.where(`project.id IN (${idsQuery.getQuery()})`);
--- a/packages/cli/src/controllers/project.controller.ts
+++ b/packages/cli/src/controllers/project.controller.ts
@@ -75,6 +75,26 @@ export class ProjectController {
 		return await this.projectsService.getProjectCounts();
 	}
 
+	// Lists projects a caller can pick as share targets, including peer
+	// personal projects so the workflow / credential share dropdowns can
+	// surface other users. Gated on `user:list` (the same boundary that
+	// `GET /rest/users` enforces) — restricted roles without that scope
+	// (e.g. chat-only users) cannot enumerate peer personal projects here.
+	@Get('/sharing-candidates')
+	@GlobalScope('user:list')
+	async getSharingCandidates(
+		req: AuthenticatedRequest,
+		res: Response,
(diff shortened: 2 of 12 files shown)


packages/@n8n/db/src/repositories/project.repository.ts
```diff
@@ -62,6 +67,40 @@ export class ProjectRepository extends Repository<Project> {
...
+		options: ProjectListOptions,
+	): Promise<[Project[], number]> {
+		// DISTINCT + LEFT JOIN avoids duplicate rows from the relation join
+		// while still allowing personal projects with no caller relation to match.
+		const idsQuery = this.createQueryBuilder('p')
+			.select('DISTINCT p.id', 'id')
+			.leftJoin('p.projectRelations', 'pr')
+			.where(
+				new Brackets((qb) => {
+					qb.where('p.type = :personalType', { personalType: 'personal' }).orWhere(
```
phyllis-noester, line 89: for my understanding: 
`'p.type = :personalType', { personalType: 'personal' }` these are all personal projects
`'pr.userId = :userId'` are all projects shared with me, e.g. non personal projects that i just have access to, correct?

packages/@n8n/db/src/repositories/project.repository.ts
```diff
@@ -62,6 +67,40 @@ export class ProjectRepository extends Repository<Project> {
...
+	}
+
+	// Wide semantics: returns peer personal projects in addition to projects
+	// the user has a relation to. Used only by the sharing-discovery endpoint
+	// (`GET /rest/projects/sharing-candidates`) so the workflow / credential
+	// share dropdowns can list other users as share targets.
+	async getShareableProjectsAndCount(
+		userId: string,
+		options: ProjectListOptions,
+	): Promise<[Project[], number]> {
```
phyllis-noester, line 81: do we really want to share the whole project? doesn't that also include variables etc that another user should not care about?

packages/cli/src/controllers/project.controller.ts
```diff
@@ -75,6 +75,26 @@ export class ProjectController {
 		return await this.projectsService.getProjectCounts();
 	}
 
+	// Lists projects a caller can pick as share targets, including peer
+	// personal projects so the workflow / credential share dropdowns can
+	// surface other users. Gated on `user:list` (the same boundary that
+	// `GET /rest/users` enforces) — restricted roles without that scope
+	// (e.g. chat-only users) cannot enumerate peer personal projects here.
+	@Get('/sharing-candidates')
+	@GlobalScope('user:list')
```
phyllis-noester, line 84: why not also list:projects? since we're also listing projects the user has been added to

afitzek: We do mostly list personal projects, which are basically users, the projects that a user is added to, would not really need a project:list scope, but a project:read scope for each of the projects individually.

packages/@n8n/db/src/repositories/project.repository.ts
afitzek: On the repository level, I wanted to keep this function as close as possible to the original function, which semantics where changed.

packages/@n8n/db/src/repositories/project.repository.ts
afitzek: Yes exactly. The regression is, that the original function, was showing all projects the user was added to and all personal projects. This was changed in the instance AI PR, so that the other personal projects are not returned anymore. So to not break the instance AI, we introduce a new function here, that lists all personal and projects that the user is a member of.

phyllis-noester (approved):

PR #29530, opened 2026-04-29
fix(core): Add missing @n8n/tournament alias to Vite config (no-changelog)

Flexicon:
## Summary

After `@n8n/tournament` was moved into the monorepo (#29358), its Vite alias was never added to `editor-ui/vite.config.mts`. All other monorepo packages have an alias redirecting bare imports to their `src/` directory so Vite serves TypeScript directly in dev mode. Without this alias, Vite resolved `@n8n/tournament` via the `exports` map, which points both `import` and `require` conditions to the CJS `dist/index.js`. CJS re-exports of `recast` named exports (like `astVisit`) don't surface as ESM named imports, causing the dev server crash:

```
Uncaught SyntaxError: The requested module '/@fs/.../packages/@n8n/tournament/dist/index.js'
does not provide an export named 'astVisit'
```

**Fix:** Add one alias entry pointing `@n8n/tournament` → `src/`, consistent with every other monorepo package alias in the file.

**To test:** Run `pnpm dev:fe` -> browser should load without the SyntaxError.

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/CAT-2988

## Review / Merge checklist

- [x] I have seen this code, I have run this code, and I take responsibility for this code.
- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md))
- [x] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [x] Tests included.
- [x] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if the PR is an urgent fix that needs to be backported)

🤖 PR Summary generated by AI

PR #29358, opened 2026-04-28
chore: Move @n8n/tournament to monorepo

Matsuuu:
## Summary

Move @n8n/tournament to monorepo for easier updating and releases.

Also migrates the library from Jest to Vitest

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/CAT-2868/move-n8ntournament-to-the-monorepo

## Review / Merge checklist

- [x] I have seen this code, I have run this code, and I take responsibility for this code.
- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [ ] Tests included. 
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if the PR is an urgent fix that needs to be backported)

PR #29539, opened 2026-04-29
test: Add Playwright smoke spec for Vite dev-server boot

cstuncsik:
## Summary

Adds a Playwright smoke spec that boots the editor against the **Vite dev server** (not the production bundle) and fails on any error-level console message or uncaught page error during load. Wires it into CI on PRs touching files known to influence dev-server module resolution.

The existing E2E suite runs against the production build, which bundles around dev-only resolution problems — for example, [#29358](https://github.com/n8n-io/n8n/pull/29358) moved `@n8n/tournament` into the monorepo without adding a Vite alias, and every developer pulling master hit `SyntaxError` on dev-server boot until [#29530](https://github.com/n8n-io/n8n/pull/29530) fixed it. Nothing pre-merge caught it. This spec does.

### What's in the diff

- **New spec** `packages/testing/playwright/tests/dev-server-smoke/dev-server-boots.spec.ts` — covers home, blank canvas, credentials. Fails on `console.error` and `pageerror`. Manual capture (rather than the existing context-scoped `consoleErrorFixtures`) because that fixture only attaches diagnostics, never fails the test, and doesn't observe `pageerror`.
- **New Playwright project** `dev-server-smoke` registered in `playwright-projects.ts` with its own `testDir` so it's not picked up by the container e2e modes.
- **New npm script** `test:dev-server-smoke` that sets `N8N_EDITOR_URL=http://localhost:8080` (which flips Playwright into `IS_DEV` mode and auto-launches Vite via `pnpm dev:fe:editor`).
- **New reusable workflow** `.github/workflows/test-dev-server-smoke-reusable.yml`.
- **CI wiring** in `ci-pull-requests.yml`: new `dev-server-smoke` filter (gated paths from the ticket), new conditional job, added to `required-checks.needs`. Skipped jobs report as pass, so non-matching PRs are not gated.

### Verification

Locally, all 3 tests pass against master in ~1.1 min (cold-start home is 17s; subsequent navigations sub-10s).

To verify the regression catch, removing the `@n8n/tournament` alias from `vite.config.mts` (line 38) makes all 3 tests fail with the exact original-failure diagnostic:

> `SyntaxError: The requested module '/@fs/.../@n8n/tournament/dist/index.js' does not provide an export named 'astVisit'`

This satisfies AC #2 (readable diagnostic pointing at the offending module URL).

### Open question — runtime budget

Acceptance criterion says "Adds < 60 s to the gated workflow." Realistic CI wall-clock with warm caches is ~90–120 s (install ~30 s + targeted build ~30 s + browsers ~10 s + n8n+Vite boot ~25 s + 3 sequential tests ~25 s). Tightening to <60 s requires reusing pre-built artifacts from `install-and-build` (similar to the `prepare-docker` cache pattern) — happy to do that as a follow-up if the budget is hard.

https://linear.app/n8n/issue/IAM-600

Related PRs: #29358 (cause), #29530 (fix).

## Review / Merge checklist

- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created. _(N/A — internal test infra)_
- [x] Tests included.
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1`. _(N/A — test infra, no runtime change)_


Matsuuu (approved):
Awesome! Thank you for this!

Flexicon (approved):
🚀

packages/testing/playwright/tests/dev-server-smoke/dev-server-boots.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * These tests visit a small set of representative routes and fail the suite if
+ * any error-level console message or uncaught page error is observed during the
+ * load. They are intentionally light — UI behaviour is covered elsewhere.
+ *
+ * Must run against the Vite dev server (`N8N_EDITOR_URL` set), which is what the
+ * `test:dev-server-smoke` script wires up.
+ */
+
+const BENIGN_PATTERNS: Array<{ messageRe: RegExp; reason: string }> = [
```
shortstacked, line 18: This pattern we should be wary of. This usually scope creeps once people see errors. Do we know the root cause for the server connection lost issue. There's been an increase in bug reports related to server connections lately and this could be part of it.

PR #29546, opened 2026-04-29
feat(agents): Let agents post images and rich cards to chat 

mike12345567:
## Summary

Extends `rich_interaction` to support display-only content (images, gifs, info cards, summary blocks) alongside its existing interactive behaviour.

Essentially this allows the rich interaction to be used for more than just getting user feedback, it allows the LLM to generate any kind of rich UI to show whatever it needs to the user.

### What changed

- **Display vs interactive — auto-detected.** If the `components` input contains any `button`/`select`/`radio_select`, the tool suspends as before. Otherwise it renders the card and the agent continues without halting — no new input fields, no DB checkpoint on the display-only path.
- **New `tool-card-display` stream chunk.** Tools call `ctx.display(payload)` to fire a render side-effect without halting. The runtime emits the chunk, `agent-chat-bridge` renders it, and `ExecutionRecorder` captures it as a closed tool-call timeline event.
- **Generic, platform-agnostic tool description.** Per-integration `description` overrides on `SlackIntegration` / `TelegramIntegration` removed. The tool factory now composes a single description from `supportedComponents`, inherited automatically by any future integration.
- **`Tool.systemInstruction(text)` framework feature.** Tools can now ship a behavioural directive that the runtime collects and prepends to the agent's system prompt under a `<built_in_rules>` block. Used by `rich_interaction` to nudge the LLM toward posting images/gifs as cards rather than plain text URLs.

### Screenshots

Before:
image

Now:
image

## Review / Merge checklist


packages/cli/src/node-execution/ephemeral-node-executor.ts
```diff
@@ -298,11 +298,30 @@ export class EphemeralNodeExecutor {
...
+			const message = error instanceof Error ? error.message : String(error);
+			this.logger.debug('Node execution validation failed', {
+				nodeType: request.nodeType,
+				error: message,
+			});
+			return {
+				status: 'error',
+				data: [],
+				error: `Cannot execute node "${request.nodeType}": ${message}`,
+			};
```
bjorger, line 323: Returning `{ status: 'error' }` here makes validation failures look like successful tool executions downstream. `resolveNodeTool()` returns this object normally from the handler, and the agent runtime only sets `isError: true` when the handler throws, so the execution recorder marks the timeline entry as successful. Could we either throw for these validation failures in the node tool handler, or add a typed result contract that the runtime/recorder treats as a failed tool call?
text

packages/@n8n/agents/src/sdk/tool.ts
```diff
@@ -139,6 +141,20 @@ export class Tool<
...
+	/**
+	 * Attach a behavioural directive to this tool. When the tool is registered
+	 * with an agent, the runtime injects this text into the agent's system
+	 * prompt under a `<built_in_rules>` block, where the LLM weighs it heavily
+	 * for "should I call this tool?" decisions.
+	 *
+	 * Use sparingly — only for guidance the description alone doesn't reliably
+	 * convey (e.g. "prefer this tool over plain text when X").
+	 */
+	systemInstruction(text: string): this {
```
bjorger, line 153: This documents `systemInstruction()` as a general Tool builder feature, but it does not round-trip through the custom-tool descriptor path. `build()` puts it on `BuiltTool`, while `describe()`/`ToolDescriptor` omit it, and JSON-config reconstruction builds custom tools only from the descriptor. So a custom tool using `.systemInstruction(...)` will lose the directive after save/publish/reconstruct. Could we either persist this field through `ToolDescriptor`, or make the API/comment explicit that this only works for in-memory/runtime-injected tools?

bjorger (approved):
LGTM 🚀

PR #29553, opened 2026-04-30
fix(core): Round fractional time saved values before inserting into insights BIGINT column

alielkhateeb:
## Summary

https://www.loom.com/share/b0bf0894234949a993b09a115787c23b

Fixes a `QueryFailedError: invalid input syntax for type bigint` crash that occurred when dynamic time saved workflows produced fractional minute values (e.g. `9.799999999999999`). PostgreSQL's `BIGINT` column rejects non-integer JavaScript numbers.

**Two-layer fix:**
- **Backend:** Added `integerValueForInsightsRaw()` which rounds values with `Math.round()` before inserting into `insights_raw.value`. Also handles non-finite values (returns 0).
- **Frontend/Node:** Added `precision: 0` / `:precision="0"` to the time saved inputs in `WorkflowSettings.vue` and `TimeSaved.node.ts` so the UI prevents fractional input at the source.

**How to test:**
Note: This should be tested with n8n in docker with postgres
1. Create a workflow with a Time Saved node set to dynamic mode
2. Set a decimal value (e.g. `5.4` minutes) via expression
3. Execute the workflow and wait for the insights flush cycle
4. Verify no `QueryFailedError` appears in logs and insights are recorded correctly

https://linear.app/n8n/issue/LIGO-420

fixes #27607

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInput.vue
+++ b/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInput.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { computed, inject, nextTick, onBeforeUnmount, onMounted, onUpdated, ref, watch } from 'vue';
-import { computedAsync, useDebounceFn } from '@vueuse/core';
+import { computedAsync, useDebounceFn, useElementSize } from '@vueuse/core';
 
 import get from 'lodash/get';
 
@@ -81,7 +81,6 @@ import { useWorkflowsListStore } from '@/app/stores/workflowsList.store';
 import { useUIStore } from '@/app/stores/ui.store';
 import type { EventBus } from '@n8n/utils/event-bus';
 import { createEventBus } from '@n8n/utils/event-bus';
-import { useElementSize } from '@vueuse/core';
 import { captureMessage } from '@sentry/vue';
 import { isCredentialOnlyNodeType } from '@/app/utils/credentialOnlyNodes';
 import {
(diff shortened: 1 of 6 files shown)


packages/nodes-base/nodes/TimeSaved/tests/TimeSaved.node.test.ts
alielkhateeb, line 1: Refactored this test file, didn't like that we manually mock and type cast in each test case, when we can properly mock in a helper function

packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInput.vue
alielkhateeb, line 1: Also while navigating the codebase found those unnecessary eslint-disable comments and some minor refactoring opportunities

packages/nodes-base/nodes/TimeSaved/tests/TimeSaved.node.test.ts
```diff
@@ -1,28 +1,58 @@
-import type { IExecuteFunctions } from 'n8n-workflow';
+import { mock } from 'jest-mock-extended';
+import type { IExecuteFunctions, INodeExecutionData } from 'n8n-workflow';
+
 import { TimeSaved } from '../TimeSaved.node';
 
 describe('TimeSaved node', () => {
+	function createExecuteFunctionsMock(options?: {
```
sandra0503, line 7: Is it something we need more often in unit tests? Then it could be worth moving to centralised mocks :)

sandra0503:
Nice!

sandra0503 (approved):

packages/nodes-base/nodes/TimeSaved/tests/TimeSaved.node.test.ts
alielkhateeb: Most likely yes, but we will tackle this as part of https://linear.app/n8n/issue/LIGO-489/refactor-duplicate-mocks-in-tests

PR #29556, opened 2026-04-30
fix(Airtable Node): Fix typecast option dropping attachment field updates

### LINKED ISSUE #22999: Airtable Node: Typecast option doesn't allow attachment upload
### Bug Description

When enabling the typecast option inside of the Airtable node. This allows me to create new single-select options but when I simultaneously try to upload an attachment, the attachment would not get updated which forces me to divide the action into two steps:
1. Uploading the attachment without the typecast option
2. Creating the single-select options using the typecast option
This seems to be a bug for me. Feel free to reproduce it on your end

### To Reproduce

Go into the airtable module, try to enable typecast and upload an attachment as well as creating a new dropdown option afterwards check in Airtable and you will see that only the single select option has been created but not the attachment has been uploaded 

### Expected behavior

I would expect that both actions happen inside of one node step 

### Debug Info

I am using the paid cloud version and also using a personal access token for Airtable 

# Debug info

## core

- n8nVersion: 1.122.4
- platform: docker (cloud)
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: 20
- license: community

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 720 hours
- maxCount: 25000 executions

## client

- isTouchDevice: false

Generated at: 2025-12-10T00:08:16.097Z

### Operating System

docker (cloud)

### n8n Version

1.122.4

### Node.js Version

22.21.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

n8n cloud

michael-radency:
## Summary

When the Typecast option is enabled on the **Airtable** create/update/upsert operations,
`skipValidation: true` is passed to `getNodeParameter`, which bypasses the type
coercion step that normally converts JSON strings to arrays for `array`-type fields
(multipleAttachments). As a result, attachment field values entered as JSON strings
in the resource mapper were sent to the Airtable API as plain strings, which the API
silently ignored — forcing users to split the operation into two separate steps.

This fix adds a post-processing step that uses the resource mapper schema to identify
`array`-type fields and parses any string values as JSON arrays before the API request
is made. Also fixes upsert, which was missing `skipValidation: true` entirely, causing
typecast to fail with a validation error when entering new select option values.

closes https://github.com/n8n-io/n8n/issues/22999
https://linear.app/n8n/issue/NODE-4584/community-issue-airtable-node-typecast-option-doesnt-allow-attachment

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Airtable/v2/helpers/utils.ts
```diff
@@ -91,3 +96,35 @@ export function legacyFlattenOutput(record: IDataObject, nodeVersion: number): I
...
+	for (const schemaField of columnsParam.schema) {
+		if (schemaField.type !== 'array') continue;
+		const value = fields[schemaField.id];
+		if (typeof value !== 'string') continue;
+		try {
+			let parsed: unknown;
+			try {
+				parsed = JSON.parse(value);
+			} catch {
+				parsed = JSON.parse(value.replace(/'/g, '"'));
```
DawidMyslak: Do we need the `value.replace(/'/g, '"')` fallback for sure? It could corrupt data with legitimate apostrophes - e.g. a filename like `O'Brien.pdf` in an attachment would get mangled. Standard JSON uses double quotes and that's what the resource mapper produces.

packages/nodes-base/nodes/Airtable/v2/helpers/utils.ts
```diff
@@ -91,3 +96,35 @@ export function legacyFlattenOutput(record: IDataObject, nodeVersion: number): I
...
+			}
+			if (isIDataObjectArray(parsed)) {
```
DawidMyslak, line 118: I wonder, if a user types `{"url":"..."}` (object instead of array), the parse succeeds but `isIDataObjectArray` returns false, so the string is quietly kept as-is?

packages/nodes-base/nodes/Airtable/v2/helpers/utils.ts
```diff
@@ -91,3 +96,35 @@ export function legacyFlattenOutput(record: IDataObject, nodeVersion: number): I
...
+	return Array.isArray(value);
+}
+
+/**
+ * When typecast is enabled, `skipValidation: true` is passed to `getNodeParameter`, which bypasses
+ * the type coercion step in `validateResourceMapperValue`. For `array`-type fields (e.g.
+ * multipleAttachments), values may be stored as JSON strings and need to be parsed into actual
+ * arrays before being sent to the Airtable API.
+ */
+export function coerceArrayTypeFields(fields: IDataObject, columnsParam: unknown): void {
```
DawidMyslak, line 110: Worth maybe adding some test for this new helper?

DawidMyslak (dismissed):
No real blockers from my end @michael-radency , left some minor hints/questions.

michael-radency:
thanks @DawidMyslak , updated as suggested

DawidMyslak (approved):

PR #29568, opened 2026-04-30
fix(core): Harden Set node workflow SDK contract

aalises:
## Summary

Hardens the Set/Edit Fields contract in `@n8n/workflow-sdk`, which is consumed by the Instance AI workflow builder.

This PR addresses repeated builder failures where generated workflows used invalid or incomplete Set node shapes, especially:

- `mode: "keepAllExistingFields"` instead of valid Set modes
- missing v3.4 assignment `id` fields
- incomplete manual assignments
- empty Set examples in SDK guidance
- simple reshaping being fragile enough to fall back to Code nodes

The fix adds workflow-sdk validation for invalid Set modes and malformed assignment entries, updates the Set/Edit Fields prompt guide with one canonical v3.4 manual mapping shape, replaces empty Set examples in workflow patterns, and adds an execution eval for the preserve-fields contract.

## Workflow eval evidence

LangSmith experiment: https://smith.langchain.com/o/a7b41b9d-bee0-4cf0-9786-f4600380f803/datasets/e4a43661-01a8-4a10-a4dd-ba346390f271/compare?selectedSessions=f6f19782-23b4-4106-ab26-2ca8018af27d

Command:

```bash
LANGSMITH_API_KEY="$LANGSMITH_API_KEY" pnpm eval:instance-ai --filter set-edit-fields-contract --iterations 10 --verbose
```

Summary:

| Metric | Result |
| --- | --- |
| Fresh workflow builds | 10 |
| Build success | 10/10 |
| Execution scenario | `preserve-fields` |
| Scenario pass count | 10/10 |
| pass@10 | 100% |
| pass^10 | 100% |
| Generated workflows kept | No (`--keep-workflows` was not used) |

Per-build evidence:

| Run | Generated workflow nodes | Set/Edit Fields node | Contract verified | Result |
| --- | --- | --- | --- | --- |
| 1 | `Daily at 09:00 -> Fetch Post -> Add Caption and Source` | `Add Caption and Source` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 2 | `Daily Schedule -> Fetch Post -> Add Fields` | `Add Fields` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 3 | `Daily at 09:00 -> Fetch Post -> Transform Data` | `Transform Data` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 4 | `Daily Schedule -> Fetch Post -> Transform Data` | `Transform Data` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 5 | `Daily at 09:00 ET -> Fetch Post -> Add Fields` | `Add Fields` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 6 | `Daily at 09:00 -> Fetch Post -> Add Caption and Source` | `Add Caption and Source` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 7 | `Daily at 09:00 -> Fetch Post -> Add Caption and Source` | `Add Caption and Source` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 8 | `Daily at 09:00 -> Fetch Post -> Add Fields` | `Add Fields` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 9 | `Daily at 09:00 ET -> Fetch Post -> Add Fields` | `Add Fields` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |
| 10 | `Daily at 09:00 -> Fetch Post -> Transform Data` | `Transform Data` | Manual Set mapping preserved `id`, `userId`, `title`, `body` and added `caption`, `source` | Pass |

Across all 10 runs, the verifier confirmed the workflow used Set/Edit Fields for reshaping, preserved existing fields via `mode: manual` + `includeOtherFields: true`, and did not use `keepAllExistingFields` as a mode.

https://linear.app/n8n/issue/INS-183

## Review / Merge checklist


packages/@n8n/workflow-sdk/src/prompts/node-guidance/parameter-guides/set-node.ts
```diff
@@ -7,6 +7,40 @@ export const SET_NODE_GUIDE: NodeTypeGuide = {
...
+}
+\`\`\`
+
+- Use \`mode: "manual"\` for normal field mapping.
+- Use \`includeOtherFields: true\` when the user asks to keep existing/input fields.
+- Do NOT use \`mode: "keepAllExistingFields"\`; that is not a valid Set/Edit Fields mode.
+- Expression values must start with \`=\`, for example \`"={{ $json.title }}"\`. Raw \`"{{ $json.title }}"\` is a literal string, not an expression.
+- Each v3.4 assignment must include \`id\`, \`name\`, \`value\`, and \`type\`.
+- Use \`mode: "raw"\` only when the user explicitly needs custom JSON output.
+
```
mutdmour: why do we still have guides per node? where is this used? We should delete these parameter guides.  That's what builder hints is supposed to solve.

packages/@n8n/workflow-sdk/src/workflow-builder/plugins/validators/set-node-validator.ts
```diff
@@ -47,16 +69,102 @@ export const setNodeValidator: ValidatorPlugin = {
...
+				nodeName: displayName,
+				parameterPath: 'parameters.assignments.assignments',
+				originalName: origForWarning,
+			});
 			return issues;
 		}
 
-		for (const assignment of assignments.assignments) {
-			if (assignment.name && isCredentialFieldName(assignment.name)) {
+		for (const [index, assignment] of assignmentItems.entries()) {
```
mutdmour, line 143: can you add some workflows to committed-workflows with expected validation warnings to the manifest? those are basically e2e tests for the workflow sdk

packages/@n8n/workflow-sdk/src/workflow-builder/plugins/validators/set-node-validator.ts
mutdmour, line 59: I don't see a check for node version here. there was a lot of Set node versions, that likely have different implementations. Can you make sure this applies only for related versions?

aalises: v1/v2 use parameters.values.* AFAIK v3.0–3.2 use parameters.fields.values[], and only v3.3+ uses parameters.assignments.assignments[]. The validator early-returns when params.assignments is undefined, so older versions never hit it. But the implicit gate is fragile... will add it so the contract is self-documenting

packages/@n8n/workflow-sdk/src/prompts/node-guidance/parameter-guides/set-node.ts
aalises: It's used in the `build-workflow-agent.prompt.ts` ... why are we still keeping those in the system prompts instead of them being discoverable using the hints? not 100% sure... perhaps we should try removing them and see how they affect evals @mutdmour , I'd rather not bloat the system prompt

packages/@n8n/workflow-sdk/src/workflow-builder/plugins/validators/set-node-validator.ts
aalises: True! added some execution eval but a focused workflow-sdk test would also be 💯 , added 7.json and 8.json

packages/@n8n/workflow-sdk/src/prompts/node-guidance/parameter-guides/set-node.ts
mutdmour: then maybe we don't add this here?

packages/@n8n/workflow-sdk/src/workflow-builder/plugins/validators/set-node-validator.ts
mutdmour: I think we have enough safety mechanisms around this. we test around 2000 templates in the workflow-sdk tracking any new validation issues. it should throw an error for old nodes if a new validation is added that does not match.

mutdmour (approved):

PR #29573, opened 2026-04-30
fix: Drop template search tools from builder

mutdmour:
Removes the `search-structures` and `search-parameters` actions from the `templates` tool in `@n8n/instance-ai`. Both actions hit the public template API, returned large amounts of context-light example data, and have not paid off in practice — the planner and builder do better with curated best-practices documentation than with raw example workflows.

The `templates` tool keeps its name and its `{ action: 'best-practices', technique }` shape, so the planner prompt (`plan-agent-prompt.ts`) and builder prompt (`build-workflow-agent.prompt.ts`) — which already only reference `templates(action="best-practices", ...)` — remain valid verbatim. Tool registration in `tools/index.ts`, `plan-with-agent.tool.ts`, and `build-workflow-agent.tool.ts` is unchanged.

Net diff: **+15 / −1,935 lines** across 8 files. The two heavy support files this drags out (Mermaid generation and node-config extraction) had no other callers and are deleted.

https://linear.app/n8n/issue/INS-184
---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Claude Code-D97757?logo=claude&logoColor=white)


OlegIvaniv (approved):
🧹🧹

PR #29582, opened 2026-04-30
refactor(core): Remove global builder node guides

aalises:
## Summary

Removes the seven always-on node-specific guides from the Instance AI builder system prompt and relies on generated type definitions, `@builderHint`, defaults, and live `explore-resources` lookups as the source of truth.

## Prompt Budget Impact

Token counts are estimated with `chars / 4`.

| Prompt | Before | After | Saved |
| --- | ---: | ---: | ---: |
| Builder tool prompt | ~14,596 tokens | ~9,148 tokens | ~5,448 tokens (37.3%) |
| Sandbox builder prompt | ~18,895 tokens | ~13,448 tokens | ~5,448 tokens (28.8%) |

The removed static guide block is `21,791` characters, or ~`5,448` estimated tokens, per builder run.

## What changed

- Removed inline `IF_NODE_GUIDE`, `SWITCH_NODE_GUIDE`, `SET_NODE_GUIDE`, `HTTP_REQUEST_GUIDE`, `TOOL_NODES_GUIDE`, `EMBEDDING_NODES_GUIDE`, and `RESOURCE_LOCATOR_GUIDE` from `composeSdkRulesAndPatterns()`.
- Did not add an on-demand `nodes(action="guide")` fallback. The builder should not keep static guide content as a secondary source of truth.
- Kept a compact `NODE_CONFIGURATION_SAFETY_RULES` block that points the builder to type definitions, `@builderHint`, defaults, and `explore-resources`, and tells it not to guess from static node-shape memory.
- Surfaced node-level `builderHint.message` through `nodes(action="type-definition")`, so direct utility-node lookups retain node-level guidance even when the builder skips search.
- Fixed stale Set-node guide content that conflicted with the v3.4 generated types: arrays/objects no longer require `JSON.stringify()`.

## Why

The previous static guides were always appended to the builder prompt even when the workflow did not use those node types. That increased context cost and latency, and it created a stale-instruction risk: at least the Set guide had drifted from generated definitions.

Generated node definitions and `@builderHint` are the fresher source of truth. This change removes the builder-facing static guide path and fixes the missing node-level hint bridge in type-definition responses.

https://linear.app/n8n/issue/INS-186

## Review / Merge checklist

- [x] Tests included.


aalises:
Ran `eval:pairwise` against `notion-pairwise-workflows` head-to-head (77 examples, 3 judges, iter=1):

- primary pass: master `dad423155f` **33.8%** (26/77) → branch `5aef3d34a2` **32.5%** (25/77)
- diagnostic: **67.4% → 67.7%** (basically flat)
- 7 prompts regressed, 6 improved — within iter=1 judge wobble

No degradation signal apparent

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
```diff
@@ -62,8 +53,14 @@ const SDK_CODE_RULES = `## SDK Code Rules
...
 - Use string values directly for discriminator fields like \`resource\` and \`operation\` (e.g., \`resource: 'message'\` not \`resource: 'message' as const\`).
 - When editing a pre-loaded workflow, **remove \`position\` arrays** from node configs — they are auto-calculated.
-- **No em-dash (\`—\`) or other special Unicode characters in node names or string values.** Use plain hyphen (\`-\`) instead. The SDK parser cannot handle em-dashes.
-- **IF node combinator** must be \`'and'\` or \`'or'\` (not \`'any'\` or \`'all'\`).`;
+- **No em-dash (\`—\`) or other special Unicode characters in node names or string values.** Use plain hyphen (\`-\`) instead. The SDK parser cannot handle em-dashes.`;
+
+const NODE_CONFIGURATION_SAFETY_RULES = `## Node Configuration Safety Rules
+
+- Fetch \`nodes(action="type-definition")\` before configuring nodes. Generated definitions and \`@builderHint\` annotations are the source of truth.
+- Prefer generated definitions, \`@builderHint\`, \`@default\`, \`@searchListMethod\`, and \`@loadOptionsMethod\` over memory or examples in this prompt.
```
mutdmour: That seems confusing for an LLM. Is this absolutely necessary? Could we flag and fix the conflict instead? Like the tool you had before for the builder

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
```diff
@@ -62,8 +53,14 @@ const SDK_CODE_RULES = `## SDK Code Rules
 - Do NOT use \`as const\` assertions — the workflow parser only supports JavaScript syntax, not TypeScript-only features. Just use plain string literals.
 - Use string values directly for discriminator fields like \`resource\` and \`operation\` (e.g., \`resource: 'message'\` not \`resource: 'message' as const\`).
 - When editing a pre-loaded workflow, **remove \`position\` arrays** from node configs — they are auto-calculated.
-- **No em-dash (\`—\`) or other special Unicode characters in node names or string values.** Use plain hyphen (\`-\`) instead. The SDK parser cannot handle em-dashes.
-- **IF node combinator** must be \`'and'\` or \`'or'\` (not \`'any'\` or \`'all'\`).`;
+- **No em-dash (\`—\`) or other special Unicode characters in node names or string values.** Use plain hyphen (\`-\`) instead. The SDK parser cannot handle em-dashes.`;
```
mutdmour: We should fix this instead of having a rule for it. Though not related to this PR

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
```diff
@@ -333,14 +330,7 @@ function composeSdkRulesAndPatterns(mode: 'tool' | 'sandbox'): string {
 		'## SDK Patterns Reference\n\n' + WORKFLOW_SDK_PATTERNS,
 		'## Expression Reference\n\n' + EXPRESSION_REFERENCE,
 		'## Additional Functions\n\n' + ADDITIONAL_FUNCTIONS,
-		'## Node-Specific Configuration Guides',
-		IF_NODE_GUIDE.content,
-		SWITCH_NODE_GUIDE.content,
-		SET_NODE_GUIDE.content,
```
mutdmour, line 339: can we delete the related guides completely?

packages/@n8n/workflow-sdk/src/prompts/node-guidance/parameter-guides/set-node.ts
mutdmour, line 1: do we need this? can we delete it completely? the philosophy of the code builder is less prompts and less guidance. the types should drive everything

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
aalises: Still used by the old workflow builder. Would do when we remove it / deprecate it

packages/@n8n/workflow-sdk/src/prompts/node-guidance/parameter-guides/set-node.ts
aalises: Same as top (It's still used by the old workflow builder. Would do when we remove it / deprecate it)

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
aalises: fixedd

packages/@n8n/instance-ai/src/tools/orchestration/build-workflow-agent.prompt.ts
aalises: I've checked and the rule is stale... the SDK does handle em-dashes (and en-dashes, curly quotes, middle-dot, ellipsis) end-to-end through workflow → generateWorkflowCode → parseWorkflowCode. The rule has been there since the file was first added (#27206) and was never re-validated. Removed the rule and added a round-trip regression test in codegen-roundtrip.test.ts to lock the behavior in.

mutdmour (approved):

PR #29583, opened 2026-04-30
feat(editor): Surface cluster information in debug data (no-changelog)

afitzek:
## Summary

Removes the `Settings → Instance Registry` placeholder UI, instead surfacing the
cluster snapshot — instances and check results — through the existing debug-info
channel that the About modal copies to clipboard and "Report a bug" embeds into
the GitHub issue body. Drops the `N8N_ENV_FEAT_INSTANCE_REGISTRY` env flag and
the `orchestration:read` scope from the controller, so the registry is on by
default for all deployments and the cluster context flows into every bug report.

### Why

This is the final slice of the cluster-visibility initiative. #28936 added the
reconciliation loop and the `/instance-registry` endpoint; #29506 added four
built-in health checks (version mismatch, hostid clash, split brain, lifecycle).
The `SettingsInstanceRegistryView` shipped alongside that work was always a
placeholder behind the env flag — it printed the raw JSON and was never going to
be the long-term presentation. Rather than ship a UI that needs a design pass for
GA, we drop the surface entirely and route the same data into the channels
operators already reach for when filing a bug.

With the project landing in this PR, the env-flag wrapper around the module is
no longer useful. Keeping it would mean shipping a feature that's silently off
in every default deployment. The `orchestration:read` scope is dropped because
the cluster payload (instance keys, hostIds, versions, roles, check status,
warning codes) contains no user PII and has the same operator-debug character
as the existing `## core` block — n8n version, deployment type, license — that
every authenticated user already sees in the About modal.

Example debug data:

```
...

## cluster

- instanceCount: 1
- versions: 2.19.0
- instances:
  - instanceKey: 2792a4c2-64fe-4980-84a7-2ada52f2a241, hostId: main-V60CE4pd3EkxUDtA, instanceType: main, instanceRole: leader, version: 2.19.0
- checks:
  - check: hostid-clash, status: succeeded, warnings: -
  - check: lifecycle, status: succeeded, warnings: -
  - check: split-brain, status: succeeded, warnings: -
  - check: version-mismatch, status: succeeded, warnings: -

...
```

### How to test manually

1. Start n8n (no env vars needed any more — single instance is fine, queue-mode
   multi-main makes the cluster section more interesting).
2. Confirm the *removal*:
   - Open `Settings`. There should be **no "Instance Registry" entry** in the sidebar.
   - Visit `/settings/instance-registry` directly — the route should not resolve.
3. Confirm the *debug-data integration*:
   - Open the user menu → "About n8n" → click **Copy debug information**.
   - Paste anywhere. The output should include a `## cluster` section with
     `instanceCount`, `versions`, an `instances:` bullet list, and a `checks:`
     bullet list showing every check that ran with `status` and warning codes.
4. Confirm the *bug-report integration*:
   - Sidebar → Help → **Report a bug**. The GitHub issue page should open with
     the `## cluster` block already in the body, under `## Debug info`.
   - Right-click the menu item → "Copy link" should still produce a usable URL.
   - Open the command bar (Cmd-K / Ctrl-K) → "Report a bug". Same URL, same
     content.
5. Confirm *no scope barrier*:
   - Sign in as a regular member (not owner/admin). Repeat steps 3 and 4. The
     cluster section should appear for them too — `GET /rest/instance-registry`
     returns 200, not 403.
6. Confirm *queue-mode multi-main detection*:
   - Run `mains: 2, workers: 1` (the existing playwright fixture configuration).
     The `instances:` list should show all three processes; the `checks:` block
     should reflect their status.

- **Pinia store + synchronous reads, not `async generateDebugInfo`.** The fetch
  lives in a small `useInstanceRegistryStore`; `useDebugInfo` reads `clusterInfo`
  synchronously and omits the section when `null`. Keeping `getReportingURL()`
  synchronous preserves middle-click / "copy link" semantics on the help-menu
  item and avoids popup-blocker pitfalls in the command-bar `window.open` path.
  Fetches are triggered on natural hooks (sidebar mount, About-modal mount,
  "Report a bug" click).
- **All check results, not just failures.** The `checks:` rendering surfaces
  every check that ran (with `status: succeeded` and `warnings: -` where the
  check passed) so a reviewer can see what the cluster watchdog actually
  evaluated, not only what it complained about.
- **Per-instance fields are a deliberate subset.** `registeredAt` / `lastSeen`
  timestamps are omitted — they change every reconciliation cycle, add no
  debugging value, and would churn snapshot tests.
- **Errors are swallowed.** Transient network errors leave the previous snapshot
  in place and never break debug generation. The bug-reporting affordance must
  remain functional even if the endpoint is briefly unreachable.
- **Stacks on top of #29506.** Base branch is `iam-333-built-in-health-checks`.
  No shared files, but the cluster `checks` summary this PR renders is
  populated by the checks added there.

https://linear.app/n8n/issue/IAM-595

Builds on #29506.

## Review / Merge checklist

- [X] Tests included. 


packages/cli/src/modules/instance-registry/instance-registry.controller.ts
```diff
@@ -12,7 +12,6 @@ export class InstanceRegistryController {
 	) {}
 
 	@Get('/')
-	@GlobalScope('orchestration:read')
 	async getClusterInfo(): Promise<ClusterInfoResponse> {
```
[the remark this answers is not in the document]
afitzek: This is by design.

packages/frontend/editor-ui/src/app/components/MainSidebar.vue
```diff
@@ -235,6 +237,7 @@ watch(isCollapsed, () => {
 onMounted(() => {
 	basePath.value = rootStore.baseUrl;
 	if (isAiGatewayEnabled.value) void fetchWallet();
+	void instanceRegistryStore.fetchClusterInfo();
```
guillaumejacquart: Why fetch on the 2 places ? about modal and the main sidebar ?

guillaumejacquart (commented):
1 small question ?

packages/frontend/editor-ui/src/app/components/MainSidebar.vue
afitzek: My thoughts were that we try to have as actual data as possible, because in the copy and the open handler I can't fetch it, since these browser actions should be synchronous on the event of the user.

guillaumejacquart: but the onMounted of the about page is already the most up to date data, no ?

afitzek: Ah yes you are right. 👍

afitzek: I removed the call in the MainSidebar

guillaumejacquart (approved):

PR #29586, opened 2026-04-30
ci: Remove unused bin fields to fix pnpm install warnings (no-changelog)

shortstacked:
## Summary

Remove unused `bin` fields from three workspace packages that cause `pnpm install` to emit warnings (which CI treats as errors). The warnings occur because pnpm tries to symlink `dist/` files that don't exist until build time.

- **`@n8n/mcp-browser`** — `mcp-browser` bin removed (only consumed as a library import by `@n8n/computer-use`)
- **`@n8n/computer-use`** — `n8n-computer-use` bin removed (`local-gateway` uses programmatic imports, not CLI)
- **`@n8n/playwright-janitor`** — `playwright-janitor` bin removed (scripts use `tsx ../janitor/src/cli.ts` directly)

## Review / Merge checklist

- [ ] Tests included.


packages/@n8n/computer-use/package.json
```diff
@@ -2,9 +2,6 @@
   "name": "@n8n/computer-use",
   "version": "0.5.0",
   "description": "Local AI gateway for n8n AI Assistant — filesystem, shell, screenshots, mouse/keyboard, and browser automation",
-  "bin": {
-    "n8n-computer-use": "dist/cli.js"
```
dlavrenuek, line 6: This package does use the binary, for example when running `npx @n8n/computer-use`

dlavrenuek (changes requested):

packages/@n8n/computer-use/package.json
shortstacked: Moved them to the publishConfig section instead.

packages/@n8n/mcp-browser/package.json
```diff
@@ -4,8 +4,10 @@
   "description": "Browser automation MCP tools built on Playwright, WebDriver BiDi, and safaridriver",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
-  "bin": {
-    "mcp-browser": "dist/server.js"
+  "publishConfig": {
+    "bin": {
+      "mcp-browser": "dist/server.js"
+    }
   },
```
dlavrenuek: For `mcp-browser` package we can actually remove the binary, it's not used


```suggestion
```

dlavrenuek (changes requested):

dlavrenuek (approved):

PR #29591, opened 2026-04-30
fix(core): Guard leader-stepdown emission on vacant leadership check

tomi:
## Summary

In `MultiMainSetup.checkLeader()`, the `!leaderId` (vacant leadership) branch unconditionally emitted `leader-stepdown` and called `markAsFollower()` — even when the instance was already a follower. This caused phantom stepdown→takeover cycles where a leader that briefly couldn't see its own Redis key triggered a full workflow deactivation→reactivation for no reason, contributing to workflows failing to reactivate after leader failover.

This PR adds an `isLeader` guard (matching the pattern already used in the `leaderId !== hostId` branch) so that `leader-stepdown` is only emitted when actually transitioning from leader to follower. Tests updated to cover both the follower and leader cases for vacant leadership.

https://linear.app/n8n/issue/CAT-2762

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/scaling/multi-main-setup.ee.ts
```diff
@@ -116,9 +116,10 @@ export class MultiMainSetup extends TypedEmitter<MultiMainEvents> {
 				`[Instance ID ${hostId}] Leadership vacant, attempting to become leader...`,
 			);
 
-			this.instanceSettings.markAsFollower();
-
-			this.emit('leader-stepdown');
+			if (this.instanceSettings.isLeader) {
```
mfsiega, line 119: I'm a bit confused here. If this condition is false then we're already a follower, so marking as follower and stepping down is a no-op. If the condition is true, then we're back to the current situation where we might flap from leader -> follower -> back to leader and trigger an unnecessary activation cycle. So I'm not sure this change will improve the situation?

I thought about putting this _after_ `tryBecomeLeader`. In other words: we only step down if we _were_ previously the leader, leadership becomes vacant, and then we don't become the new leader. WDYT?

mfsiega (commented):
~~I'm a bit confused here. If the condition in~~ Moved this comment in-line.

packages/cli/src/scaling/multi-main-setup.ee.ts
tomi: @mfsiega Yeah I think that makes sense, but in that case we need to accommodate for the case where this host is the leader and the key has expired. If we move the stepdown after the attempt, we would trigger takeover even when we already were the leader. I'll accommodate that as well

tomi:
Superseded by #29696

PR #29598, opened 2026-04-30
feat(instance-ai): Reuse workflow builder sandboxes (no-changelog)

OlegIvaniv:
## Summary
This PR reduces latency and token growth for Instance AI workflow-builder follow-up turns while preserving sandbox continuity.

It keeps completed workflow-builder sandboxes warm for 10 minutes by default, keyed per thread/workflow, so simple follow-up edits can reuse the existing workspace, source files, installed dependencies, and dedicated builder memory thread. The builder memory thread is compacted after successful runs into a synthetic summary message that keeps the useful state (workflow/work item IDs, source path, node list, triggers, mocked credentials, verification state, last requested change, and final result) while deleting bulky raw tool transcripts from persisted memory.

It also compacts `verify-built-workflow` responses by default. The tool now returns concise evidence (`executionId`, `success`, `status`, executed nodes, and per-node previews), while preserving the full execution data internally for cleanup and workflow-task verification records. Full data remains available via `includeData: true`.

Additional changes:

- Reuse successful structured builder verification evidence in orchestrator post-build/checkpoint flows instead of running duplicate verification.
- Preserve builder stream resume options so suspended builder streams keep memory, save-per-step, temperature, and provider options.
- Repair missing workflow SDK installs in builder sandboxes, including the unpublished pinned-version fallback.
- Include current input tokens in orchestrator context-compaction decisions.
- Pass completed dependency context into planned sub-agent tasks.
- Remove temporary `[InstanceAI]...` debug console logs added during tuning.

## Before / after validation

LangSmith comparison on the meeting-booking workflow scenario (initial build, Gmail recipient follow-up, Slack DM follow-up) showed:

| Metric | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Wall time | 10m27.7s | 5m59.1s | -4m28.6s (-42.8%) |
| LLM calls | 64 | 54 | -10 (-15.6%) |
| Tool calls | 64 | 51 | -13 (-20.3%) |
| Reported total tokens | 530,677 | 434,031 | -96,646 (-18.2%) |
| Prompt tokens | 486,755 | 400,103 | -86,652 (-17.8%) |
| Completion tokens | 43,922 | 33,928 | -9,994 (-22.8%) |

Builder-specific changes:

| Builder run | Before | After | Finding |
| --- | ---: | ---: | --- |
| Initial build turns | 19 | 10 | Fewer builder turns |
| Initial build wall time | 3m08.8s | 1m57.9s | -70.9s (-37.6%) |
| Initial build with cache | 1,031,351 | 529,969 | -501,382 (-48.6%) |
| Gmail follow-up wall time | 35.8s | 28.4s | -20.7% |
| Slack follow-up wall time | 1m32.6s | 57.1s | -38.4% |

Verification behavior:

- Before: 7 `verify-built-workflow` calls, including 3 duplicate orchestrator-side verifications after the builder had already verified.
- After: 3 `verify-built-workflow` calls, all from the workflow-builder, with no duplicate orchestrator verification.
- Before: verify outputs returned full top-level `data` payloads.
- After: verify outputs return compact evidence (`executionId`, executed nodes, and `nodePreviews`) by default.

The comparison confirms the main intended behavior: warm builder continuity and structured verification reuse reduce end-to-end latency, tool calls, and token usage while preserving builder verification evidence.

## Review / Merge checklist

- [ ] Tests included. 


aalises:
awesome!! reminds me of the classic issue with lambda deployments where you had to keep'em warm or spinning them up again was so slow...

packages/@n8n/instance-ai/src/tools/orchestration/verify-built-workflow.tool.ts
```diff
@@ -359,6 +447,7 @@ export function createVerifyBuiltWorkflowTool(context: OrchestrationContext) {
 						failureSignature: success ? undefined : result.error,
 						evidence: {
 							nodesExecuted: nodesExecuted && nodesExecuted.length > 0 ? nodesExecuted : undefined,
+							producedOutputRows: countProducedOutputRows(result.data),
```
aalises, line 634: `workItemId` identifies an outcome, but this path still runs verification for the caller-supplied `workflowId`. 

With checkpoint reuse now trusting successful `outcome.verification`, a stale or wrong `workflowId` could verify one workflow and attach that evidence to another build, so we could derive it maybe from `buildOutcome.workflowId`, or reject mismatches before running/updating verification 💪

packages/@n8n/instance-ai/src/tools/orchestration/verify-built-workflow.tool.ts
```diff
@@ -95,6 +106,60 @@ function extractRowIdsFromNodeOutput(nodeOutput: unknown): number[] {
...
+	chars: number;
+}> {
+	if (!resultData) return [];
+
+	return Object.entries(resultData).map(([nodeName, nodeOutput]) => {
+		const serialized = stringifyForToolOutput(nodeOutput);
+		const preview = previewValue(nodeOutput, maxChars);
+		return {
+			nodeName,
+			itemCount: countOutputItems(nodeOutput),
```
aalises, line 187: The production execution adapter wraps each node output as an untrusted-data string, but these helpers count raw arrays. In production this means `itemCount` and `producedOutputRows` will usually become `1` per executed node, regardless of actual item count I thinkk. 

We could either parse/count before wrapping, or adjust tests to use the real adapter output shape.

packages/@n8n/instance-ai/src/runtime/builder-sandbox-session-registry.ts
```diff
@@ -0,0 +1,227 @@
...
+		session.cleanupTimer.unref();
+	}
+
+	private async cleanupSession(sessionId: string, _reason: string): Promise<void> {
+		const session = this.sessions.get(sessionId);
+		if (!session) return;
+
+		this.sessions.delete(session.sessionId);
+		if (session.workflowId) {
+			this.byThreadWorkflowId.delete(sessionKey(session.threadId, session.workflowId));
```
aalises: `cleanupSession()` deletes the workflow lookup unconditionally. If a newer session has already replaced the same `threadId:workflowId` mapping, an older session expiring can remove the newer alias and break warm reuse. Please guard the delete by checking the current mapped session ID still matches the session being cleaned up 😢

packages/@n8n/instance-ai/src/tools/orchestration/verify-built-workflow.tool.ts
OlegIvaniv: Fixed. The verifier now unwraps the production execution-output wrapper before inspecting node output, so arrays count as their actual item count again. It also reads the truncated-output metadata fields we already emit (`totalItems` and `_itemCount`) instead of treating every wrapped node output as one row.
I also made data-table insert cleanup read row IDs from wrapped output, and added tests for wrapped counts and wrapped insert cleanup.

packages/@n8n/instance-ai/src/runtime/builder-sandbox-session-registry.ts
OlegIvaniv: Fixed. Cleanup now only deletes the workflow alias if the current `threadId:workflowId` mapping still points to the session being cleaned up. That way an older expired session can't remove the alias for a newer retained session. Added a regression test for that replacement case.

packages/@n8n/instance-ai/src/tools/orchestration/verify-built-workflow.tool.ts
OlegIvaniv: Fixed this by making the build outcome's workflow ID the source of truth. The verify tool now rejects missing or mismatched workflow IDs before it runs anything, and then uses the outcome workflow ID for the execution, write-node lookup, cleanup logging, and terminal verdict/telemetry paths. I added coverage for both the mismatch case and the missing workflow ID case.

aalises (approved):

Issue #29434, opened 2026-04-28
LinkedIn node fails with 'refreshToken is required' for standard OAuth apps (no Marketing Developer Platform)


Mariano215:
## Summary

The LinkedIn node crashes with `NodeApiError: refreshToken is required` (`ERR_ASSERTION`) when the stored OAuth credential does not contain a refresh token. This affects **all standard LinkedIn apps** that have not been approved for LinkedIn's Marketing Developer Platform (MDP).

## Root Cause

LinkedIn only issues refresh tokens to apps approved for the **Marketing Developer Platform**. Standard apps using the "Share on LinkedIn" product receive a 60-day access token and **no refresh token**. This is documented LinkedIn behavior, not a user misconfiguration.

When the access token expires (or n8n attempts to refresh it), the call chain reaches:

```
client-oauth2-token.js → refresh() → a.ok(this.refreshToken, 'refreshToken is required')
```

This hard assertion throws an opaque `ERR_ASSERTION` error instead of a useful message.

Additionally, the `LinkedInOAuth2Api` credential definition does not include `offline_access` in its scope — and LinkedIn rejects `offline_access` for standard apps with `invalid_scope_error: Unknown scope "offline_access"` — so there is no way to obtain a refresh token through normal OAuth flow without MDP access.

## Steps to Reproduce

1. Create a standard LinkedIn app (no Marketing Developer Platform product)
2. Add LinkedIn OAuth2 credential in n8n and connect via OAuth
3. Run a workflow using the LinkedIn node
4. Wait for the 60-day access token to expire (or revoke and reconnect)
5. Run the workflow again → `NodeApiError: refreshToken is required`

## Expected Behavior

When no refresh token is present and the access token is expired, n8n should throw a clear, actionable error:

> "LinkedIn access token has expired. Please reconnect the credential."

Or better: attempt the request with the existing access token, and only surface the re-auth message on a 401 response.

## Actual Behavior

```
{
  "errorMessage": "refreshToken is required",
  "errorDetails": {
    "rawErrorMessage": ["refreshToken is required"],
    "httpCode": "ERR_ASSERTION"
  }
}
```

Stack:
```
NodeApiError: refreshToken is required
    at ExecuteContext.linkedInApiRequest (nodes/LinkedIn/GenericFunctions.ts:71)
    at @n8n/client-oauth2/src/client-oauth2-token.ts → refresh()
```

## Environment

- n8n version: 2.16.1 (self-hosted)
[...]

PR #29613, opened 2026-05-01
fix(LinkedIn Node): Improve refresh token expiry error messaging

rajats-dev:
## Summary

Improves the error shown when an OAuth access token expires but no refresh token exists. In @n8n/client-oauth2, now throws a clear actionable error instead of failing an assertion.

Screenshot 2026-05-01 at 2 30 42 PM

Added an UI notice to LinkedIn OAuth2 credentials model explaining refresh-token behavior. The LinkedIn credential now shows a notice about the standard LinkedIn app.

https://github.com/n8n-io/n8n/issues/29434

## Review / Merge checklist

- [x] Tests included. 


packages/nodes-base/credentials/LinkedInOAuth2Api.credentials.ts
```diff
@@ -10,6 +10,13 @@ export class LinkedInOAuth2Api implements ICredentialType {
 	documentationUrl = 'linkedin';
 
 	properties: INodeProperties[] = [
+		{
+			displayName:
+				'LinkedIn standard apps typically do not issue refresh tokens. Access tokens expire (often after ~60 days). When it expires, you’ll need to reconnect the credentials.',
```
alexander-gekov: The current copy tells the user *what* happens but not *why* or what the long-term fix is. Suggest naming the Marketing Developer Platform explicitly and linking to LinkedIn's docs, so users know whether reconnecting every 60 days is just a workaround or whether MDP access is the real solution:

```suggestion
			displayName:
				'Standard LinkedIn apps issue access tokens valid for 60 days and do not return a refresh token. You will need to reconnect this credential when the token expires. Automatic token refresh requires <a href="https://learn.microsoft.com/en-us/linkedin/marketing/" target="_blank">Marketing Developer Platform</a> access from LinkedIn.',
```

alexander-gekov (commented):
Hi @rajats-dev — thanks for picking this up! I was assigned the same Linear ticket (NODE-4943) and was about to open a duplicate PR before noticing yours, so I closed mine (#30103) in favor of this one.

Two suggestions to help get this over the line — neither blocking, both should make a reviewer more comfortable approving:

### 1. Add a unit test for the new error path

The "Tests included" checkbox is unblocked once we lock in the new behavior. `packages/@n8n/client-oauth2/test/code-flow.test.ts` already has a `#refresh` block where this fits naturally — suggested addition right after the existing `should make a request to get a new access token` test:

```ts
it('should throw a clear error when no refresh token is present', async () => {
  const token = githubAuth.createToken({
    access_token: config.accessToken,
    refresh_token: '',
  });

  await expect(token.refresh()).rejects.toThrow(
    'OAuth access token expired and no refresh token is available. Please reconnect the credentials.',
  );
});
```

### 2. Stronger notice copy on the LinkedIn credential

See inline suggestion below — the current copy is correct but doesn't tell the user *why* there's no refresh token. The root cause is that LinkedIn only issues refresh tokens to apps approved for the Marketing Developer Platform, so calling that out (with a link) gives users a long-term path forward, not just a workaround.

Thanks again for the fix!

packages/nodes-base/credentials/LinkedInOAuth2Api.credentials.ts
rajats-dev: Thanks @alexander-gekov for the code suggestion. I have updated the PR.

alexander-gekov (approved):
Looks good to me

alexander-gekov:
@rajats-dev Can you also remove full-stop from PR title for CI to pass

rajats-dev:
@alexander-gekov  Sure

PR #29616, opened 2026-05-01
fix(FTP Node): Queue concurrent handshakes per credential

krider2010:
## Summary

Workflows that fan out to multiple SFTP nodes against the same SSH user fail intermittently with `getConnection: Timed out while waiting for handshake`. Each FTP/SFTP node creates its own `ssh2-sftp-client` and calls `.connect()` independently, so the SSH server sees a burst of handshakes from one user and throttles them (`MaxStartups`). n8n's 10s `readyTimeout` surfaces the throttle as a hard failure.

This PR adds a per-credential semaphore around `connect()` keyed by `host:port:user:protocol`. Concurrent handshakes for the same destination queue instead of racing. Items inside one node already share one connection, so item-level behaviour is unchanged.

- New `packages/nodes-base/nodes/Ftp/HandshakeQueue.ts` — process-level `Map<credentialHash, Bucket>` with a tiny promise semaphore, `makeHandshakeKey()`, `withHandshakePermit()`. No new dependencies.
- `packages/nodes-base/nodes/Ftp/Ftp.node.ts` — wraps the 6 `connect()` call sites (FTP/SFTP credential test + execute path).
- Default cap is 5, configurable via `N8N_SFTP_HANDSHAKE_CONCURRENCY` (range 1–32, invalid values fall back to default).

### Why default 5?

The relevant ceiling is the server's `MaxStartups` (OpenSSH default 10 before drops). 5 sits below that, lets normal 3–4-node fan-outs run unqueued, and stops the burst pattern that triggered this bug. Stricter servers tune via the env var.

This PR is the safety net: it caps concurrency, but doesn't reduce handshake volume. A connection pool would do that — tracked as a follow-up.

### Test plan

- Unit tests for `HandshakeQueue`: ordering under load, permit release on `connect()` resolve and reject, isolation across credentials, env var parsing for valid/invalid/out-of-range values.
- Existing `Ftp.node.test.ts` continues to pass.
- Manual repro pending — to be validated by sharing this branch with the original reporter against their reproducer workflow.

```
pnpm --filter n8n-nodes-base test Ftp
# 17 passed, 17 total
pnpm --filter n8n-nodes-base typecheck
pnpm --filter n8n-nodes-base lint nodes/Ftp
```

## Review / Merge checklist

- [x] Tests included.


Joffcom:
@krider2010 I like the idea but instead of having it as an env option what if it was an option in the credential itself?

krider2010:
> @krider2010 I like the idea but instead of having it as an env option what if it was an option in the credential itself?

Definitely open to that. I'm working with a customer so can we see if this fixes it first (they look at the branch) and then talk refactoring where this goes?

JUVOJustin:
This already looks great and does solve my inital issue. 

One theoretical issue with how the key/buckets are generated: Using host:port:user:protocol as key can cause the same error again if more than 10 different ssh users executing in parallel. I think the solution is simply to strip the username.
```mermaid
sequenceDiagram
    autonumber
    participant Alice as Alice bucket
    participant Bob as Bob bucket
    participant Carol as Carol bucket
    participant Queue as Local queue
    participant Server as SFTP server
    participant Error as Failed connect

    Note over Alice: Key is sftp host 22 alice
    Note over Bob: Key is sftp host 22 bob
    Note over Carol: Key is sftp host 22 carol
    Note over Alice,Carol: Username is part of the key, so each user has an independent limit of 5

    par Alice requests
        Alice->>Server: A1 to A5 sent now
        Alice->>Queue: A6 queued because Alice bucket is full
    and Bob requests
        Bob->>Server: B1 to B3 sent now
    and Carol requests
        Carol->>Server: C1 to C3 sent now
    end

    Note over Server: Server receives 11 concurrent unauthenticated SSH handshakes
    Note over Queue: Only A6 is queued locally. Bob and Carol are not coordinated with Alice

    Note over Server: First 10 handshakes enter startup handling
    Server-->>Error: 11th handshake may be throttled by MaxStartups
    Error-->>Carol: connect waits until readyTimeout
    Error-->>Carol: Handshake timeout or reset

    Server-->>Alice: One Alice handshake completes
    Queue-->>Alice: A6 leaves the local queue
    Alice->>Server: A6 is sent after permit release

```

Something i have not tested but might be an issue: The Queue seems to be stored in memory. In a setup with multiple n8n workers the queue will not be shared. Assuming openSSHs default of 10 MaxStartups this error would occur again with 3 instances. Same would affect connection pooling i think

krider2010:
Yep - definitely not fully production ready yet! Thank you for verifying.

PR #29618, opened 2026-05-01
feat(core): Add NVIDIA Nemotron Models with cloud and self-hosted NIM support

krider2010:
## Summary

Adds NVIDIA Nemotron as a first-class chat model provider in n8n. Mirrors the DeepSeek / xAi pattern by reusing `ChatOpenAI` with a credential-supplied `baseURL` (NVIDIA's hosted API is OpenAI-spec compatible).

**What ships**

- New **NVIDIA Nemotron Chat Model** node (`lmChatNvidia`) selectable from the Nodes Panel and wireable into AI Agent / AI Chain.
- Two credential types so the same node covers both deployment modes:
  - **NVIDIA Nemotron** — cloud (build.nvidia.com), hidden default URL `https://integrate.api.nvidia.com/v1`.
  - **NVIDIA Nemotron (Self-Hosted NIM)** — visible required URL, optional API key for open NIM containers.
  - After a refactor, there is one type in the code, but both options are possible.
- Model dropdown fetches `/v1/models` live and filters to ids matching `/nemotron/i`. A curated fallback list of well-known Nemotron model ids is shown when the dropdown can't reach the endpoint.
- Full Chat Hub wiring so NVIDIA Nemotron appears in the unified `ModelSelector.vue` alongside other providers (api-types schema + map, backend fetcher, frontend display name).
- Unit tests for the new node (10 tests, including both credential paths and the Nemotron filter).

**How to test**

1. Open n8n, create a credential of type **NVIDIA Nemotron** with a key from build.nvidia.com — confirm the connection test passes.
2. In a new workflow, add an AI Agent and attach the **NVIDIA Nemotron Chat Model** node. Pick a Nemotron model from the dropdown and send a prompt — confirm a non-empty response.
3. Repeat with a **NVIDIA Nemotron (Self-Hosted NIM)** credential pointed at any OpenAI-spec NIM endpoint.
4. Open Chat Hub → confirm NVIDIA Nemotron appears in the model picker with the new icon and a working chat round-trip.

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/cli/src/modules/chat-hub/chat-hub-workflow.service.ts
```diff
@@ -895,6 +895,16 @@ ${this.getSystemMessageMetadata(timeZone) + artifactContext}`;
 					},
 				};
 			}
+			case 'nvidia': {
+				return {
+					...common,
+					parameters: {
+						authentication: 'cloud',
```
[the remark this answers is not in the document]
krider2010: Ignored as it follows existing patterns

packages/@n8n/nodes-langchain/credentials/NvidiaSelfHostedApi.credentials.ts
```diff
@@ -0,0 +1,52 @@
...
+			default: '',
+			description: 'Optional. Leave blank if your NIM container does not require authentication',
+		},
+	];
+
+	authenticate: IAuthenticateGeneric = {
+		type: 'generic',
+		properties: {
+			headers: {
+				Authorization: '=Bearer {{$credentials.apiKey}}',
```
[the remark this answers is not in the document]
krider2010: Fixing in next push

packages/@n8n/nodes-langchain/credentials/NvidiaSelfHostedApi.credentials.ts
Cadiac, line 1: I wonder if having a separate credential type for self-hosted here matches how we've been implementing other services like this - if the regular credential made base URL visible and the token optional it would work for self hosted too if user changes the URL to point to their local instance.

But this being a separate one has its benefits too, my gut feeling is just that we've been preferring having less but more generic credential types? But I'm not that up to date with our node credential best practices.

Cadiac: To be fair, now testing this on the UI the UX here is pretty nice here.

Image

So maybe disregard this comment.

packages/@n8n/nodes-langchain/nodes/llms/LmChatNvidia/LmChatNvidia.node.ts
```diff
@@ -0,0 +1,308 @@
...
+		codex: {
+			categories: ['AI'],
+			subcategories: {
+				AI: ['Language Models', 'Root Nodes'],
+				'Language Models': ['Chat Models (Recommended)'],
+			},
+			resources: {
+				primaryDocumentation: [
+					{
+						url: 'https://docs.n8n.io/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.lmchatnvidia/',
```
Cadiac, line 48: This docs page needs to be created on n8n-docs repo (after this is in).

packages/@n8n/nodes-langchain/credentials/NvidiaApi.credentials.ts
Cadiac, line 1: The automatic "how to setup this credential" link takes user to https://docs.n8n.io/integrations/builtin/credentials/nvidia/, which would render a page based on a markdown docs file at https://github.com/n8n-io/n8n-docs/tree/main/docs/integrations/builtin/credentials, we'll need to make those docs pages for these credentials later (or file docs tickets)

packages/@n8n/nodes-langchain/credentials/NvidiaSelfHostedApi.credentials.ts
Cadiac: Although one downside of having two credentials I realize now is that on chat hub one is limited to only using cloud models, and the self-hosted option isn't available, as the concept there was that provider<->credential link would be 1:1.

Image

We ran into this with Azure Api key / Entra ID credentials which is why it appears on the list twice as a workaround. To support local Nemotron models here we'd need to add another provider for the self-hosted version here too :l 

This is something we probably would've wanted to change on chat hub, but as priorities on it shifted it hasn't happened..

Cadiac (commented):
Looks well implemented, but I wasn't able to acquire an api token to test it locally (their portal seemed to be super slow today / not work). I'd be happy to trust that it works though.

I think the main thing to discuss is whether we want two or one credential type for this node, left some comments about that.

packages/@n8n/nodes-langchain/credentials/NvidiaSelfHostedApi.credentials.ts
krider2010: Really useful info. Thank you!

krider2010: Made the single credential change. I agree on UX but we are where we are!

Cadiac (approved):
🚀

krider2010:
Docs PR at https://github.com/n8n-io/n8n-docs/pull/4646

PR #29621, opened 2026-05-01
Ignore peer dependency catalog ranges in code health

suyua9:
## Summary

- Skip `peerDependencies` when enforcing pnpm catalog references in the code-health catalog rule
- Add a regression test for a peer range that is intentionally different from the cataloged dev dependency
- Remove the stale `ts-morph` peer-dependency entry from the code-health baseline

Peer dependency ranges describe compatibility with a host package, so they should not be forced to use `catalog:` the same way install-time dependencies are.

## Tests

- `pnpm --filter @n8n/code-health test -- catalog-violations.rule.test.ts`
- `pnpm --filter @n8n/code-health format:check`
- `pnpm --filter @n8n/code-health lint`
- `pnpm --filter @n8n/code-health typecheck`

### DIFF
--- a/packages/testing/code-health/src/rules/catalog-violations.rule.test.ts
+++ b/packages/testing/code-health/src/rules/catalog-violations.rule.test.ts
@@ -97,6 +97,36 @@ catalogs:
 		expect(violations[0].message).toContain('"catalog:frontend"');
 	});
 
+	it('ignores peer dependency ranges that are also available in the catalog', async () => {
+		writeFile(
+			tmpDir,
+			'pnpm-workspace.yaml',
+			`
+packages:
+  - packages/*
+catalog:
+  ts-morph: ^27.0.0
+`,
+		);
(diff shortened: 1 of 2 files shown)


suyua9:
Rebased/recreated this branch on current `master` to clear the merge conflict. The code change remains scoped to the code-health catalog rule/tests; the baseline removal from the old branch was already reflected in current `master`, so it was not reintroduced.

Local validation:
- `git diff --check` passed
- `pnpm install --frozen-lockfile --ignore-scripts --filter @n8n/code-health...`
- `pnpm --filter @n8n/vitest-config build`
- `pnpm --filter @n8n/rules-engine build`
- `pnpm --filter @n8n/code-health test -- src/rules/catalog-violations.rule.test.ts` → 52 tests passed.

shortstacked (approved):
Thanks for the PR!

PR #29625, opened 2026-05-01
fix(editor): Preserve Form Trigger custom path on import and duplication

### LINKED ISSUE #29596: Form Trigger: Custom "Form Path" resets to random UUID on JSON re-import
## Describe the problem/error/question

When exporting a workflow as JSON and re-importing it (e.g. as part of a local backup/edit workflow), the **Form Trigger node's custom "Form Path"** is not preserved. Instead, it gets reset to a randomly generated UUID-like value.

This is inconsistent with the behavior of **Webhook nodes**, where the custom path is correctly preserved after re-import. Since the Form Path determines the public-facing URL of the form, this forces manual reconfiguration after every re-import.

**Steps to reproduce:**
1. Create a workflow with a **n8n Form Trigger** node
2. Set a **custom Form Path** (e.g. `my-custom-form`)
3. Export the workflow as JSON
4. Delete or overwrite the workflow and re-import the JSON file
5. Open the Form Trigger node → **Form Path is now a random UUID** instead of `my-custom-form`

**Expected behavior:**
The custom Form Path defined in the JSON file should be preserved after re-import, just like Webhook paths are.

**Actual behavior:**
The Form Path is reset to a newly generated UUID on every re-import, breaking existing form URLs and requiring manual reconfiguration.

## What is the error message (if any)?

No error message — the import succeeds silently, but the Form Path value is overwritten with a new UUID.

### core

- n8nVersion: 2.17.3
- nodeJsVersion: 24.14.1
- nodeEnv: production
- database: sqlite
- executionMode: regular
- concurrency: -1

### storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

### pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

### client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/147.0.0.0 safari/537.36
- isTouchDevice: false

### security

- secureCookie: false

Generated at: 2026-04-30T15:11:49.685Z

Deborg201:
Previously, custom webhook paths in Form Trigger nodes (v2.2+) were overwritten by randomly generated UUIDs during workflow import, node duplication, or canvas initialization. This occurred because the editor-ui was unconditionally regenerating the path parameter for specific node types.

This change:
- Updates `importWorkflowData` to only regenerate the path if it matches the old `webhookId` (the default behavior).
- Updates `resolveNodeWebhook` to respect an existing `options.path` before falling back to a default `webhookId`.
- Updates and adds tests in `useCanvasOperations.test.ts` to ensure custom paths are preserved while default behavior is maintained.

##Tests
Added regression test covering Form Trigger custom formPath persistence during import/duplication.

Fixes #29596

### DIFF
--- a/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.test.ts
+++ b/packages/frontend/editor-ui/src/app/composables/useCanvasOperations.test.ts
@@ -4629,7 +4629,7 @@ describe('useCanvasOperations', () => {
 						position: [200, 200] as [number, number],
 						webhookId: 'first-webhook',
 						parameters: {
-							path: 'some-path',
+							path: 'first-webhook',
 						},
 					},
 					{
@@ -4641,7 +4641,7 @@ describe('useCanvasOperations', () => {
 						webhookId: 'second-webhook',
 						parameters: {
 							options: {
-								path: 'some-path',
+								path: 'second-webhook',
 							},
 						},
 					},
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, `importWorkflowData()` updates only one of `parameters.path` or `parameters.options.path` when regenerating a webhookId, so the other can remain equal to the old webhookId. If later parameter normalization drops the visible field due to version-gated `displayOptions`, the remaining stale field can become the effective webhook path, producing incorrect webhook URLs after paste/duplicate/import.

**How to fix:** Update both path locations

> ### Issue description
> `importWorkflowData()` regenerates `webhookId` for `UPDATE_WEBHOOK_ID_NODE_TYPES`, but only updates **one** of `parameters.path` or `parameters.options.path` when they match the old id. If both fields exist (possible across versions/migrations), the other field can remain stale and later become the effective path after parameter normalization.
> 
> ### Issue Context
> Form Trigger v2 webhook path resolution uses `path || options.path || $webhookId`, and Form Trigger v2.2+ version-gates where the path property lives (top-level vs `options`). The editor’s parameter resolution filters hidden fields, so leaving one stale can break pasted/duplicated/imported webhook URLs.
> 
> ### Fix Focus Areas
> - packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts[2640-2650]
> 
> ### Suggested change
> Replace the `if/else if` with two independent checks so **both** locations are updated when they equal `oldWebhookId`:
> 
> ```ts
> const oldWebhookId = node.webhookId;
> nodeHelpers.assignWebhookId(node);
> 
> if (node.parameters.path === oldWebhookId) {
>   node.parameters.path = node.webhookId;
> }
> 
> const options = node.parameters.options as IDataObject | undefined;
> if (options?.path === oldWebhookId) {
>   options.path = node.webhookId;
> }
> ```
> 
> (Keep the existing behavior of preserving truly custom paths.)

---

Deborg201:
Done. Now both locations are updated when they equal oldWebhookId.

Joffcom:
Hi @Deborg201 — thanks for tracking this down! Quick update: this exact bug was fixed in master by #30053 (merged 2026-05-14), which landed the same `pathMatchedWebhookId` / `optionsPathMatchedWebhookId` logic you're proposing here. The custom-path-preservation flow you describe is now the behaviour on current `master`, so this PR is effectively duplicated by an internal fix that crossed yours.

Closing this one in favour of the already-merged change — really appreciate the work though, and apologies that the team didn't realise yours was open before shipping #30053. If you spot any cases where the merged fix still doesn't preserve a custom path (since your test suite is more thorough), please open a fresh issue with the specific scenario and we'll happily look.

Thanks for the contribution! 🙏

PR #29628, opened 2026-05-01
chore(agents): Tidy up un-necessary migrations

mike12345567:
## Summary

Throughout the process of building the feature quite a few different migrations were created as we developed, but this is an un-released feature, so we don't need to worry about all the various stages of migration before we merge this part. I've compressed all the individual parts down into two migrations:

1. One for the `execution_entity` and the new `execution_threads` tables.
2. One to create the agents table.

Addresses: https://linear.app/n8n/issue/AGENT-49/cleanup-sql-migrations-before-merge-to-main

## Review / Merge checklist

- [ ] Tests included. 


Qodo-Free-For-OSS:
Hi, The `agent_published_version` table defines `createdAt`/`updatedAt` as NOT NULL without defaults, but the publish upsert only sets `updatedAt`, so the first publish insert will fail with a NOT NULL violation.

**How to fix:** Add defaults for timestamps

> ### Issue description
> `agent_published_version.createdAt/updatedAt` are `NOT NULL` but have no defaults. The repository upsert only supplies `updatedAt`, so the initial insert fails.
> 
> ### Issue Context
> - This migration is intended to replace unreleased migrations; it must work on fresh installs.
> 
> ### Fix Focus Areas
> - packages/@n8n/db/src/migrations/common/1783000000000-CreateAgentTables.ts[107-120]
>   - Prefer `.withTimestamps` on `agent_published_version`, OR
>   - Add `.default('NOW()')` (using existing DSL convention) for both columns.
> - (Optional hardening) packages/cli/src/modules/agents/repositories/agent-published-version.repository.ts[22-47]
>   - Include `createdAt: new Date()` on upsert insert payload if you don’t want DB defaults.

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

PR #29629, opened 2026-05-01
chore(agents): Whitelist AI provider nodes that are not usable as tools

mike12345567:
## Summary
Whitelisting some of the AI provider nodes that we have that are not usable as tools, also fixing an issue with the workflow tool I/O sizing, sometimes it would have far too much height and leave you scrolling the details panel that opens on the right.

This makes it easier to use multiple models/different capabilities behind the scenes with the agent for example:
image

## Review / Merge checklist

- [ ] Tests included. 


bjorger (approved):
LGTM

PR #29645, opened 2026-05-02
refactor(agents): Refactor styles on Executions Timeline

heymynameisrob:
## Summary
- Refresh the Executions Timeline UI to match the Agents experience, including the top bar, breadcrumbs, session switcher, spacing, and design-system-aligned styling.
- Add keyboard navigation for timeline items with `ArrowUp` / `ArrowDown`, boundary jumps with modifier keys, and `Esc` to clear the selection and close the detail panel.
- Improve performance for large sessions by virtualizing the timeline table with `N8nRecycleScroller`, adding scroll-to-selection behavior, and replacing per-segment tooltips with a shared chart popover.
- Standardize markdown rendering across chat and execution surfaces by extracting shared markdown styles into the design system and reusing them in both panels.
- Polish the detail experience with reusable event pills, better metadata display, copy-to-clipboard actions for JSON blocks, and cleaner row/chart selection states

CleanShot 2026-05-02 at 16 08 45@2x

[AGENT-95](https://linear.app/n8n/issue/AGENT-95/p05-style-executions-detail)

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/frontend/editor-ui/src/features/agents/components/SessionTimelineTable.vue
+++ b/packages/frontend/editor-ui/src/features/agents/components/SessionTimelineTable.vue
@@ -86,7 +86,11 @@ const rows = computed<Row[]>(() => {
 });
 
 function updateScrollMask() {
-	if (!scrollContainer) return;
+	if (!scrollContainer) {
+		canScrollUp.value = false;
+		canScrollDown.value = false;
+		return;
+	}
 	canScrollUp.value = scrollContainer.scrollTop > 0;
 	canScrollDown.value =
 		scrollContainer.scrollTop + scrollContainer.clientHeight < scrollContainer.scrollHeight - 1;
@@ -144,6 +148,7 @@ watch(
 		]"
 	>
 		<N8nRecycleScroller
+			v-if="rows.length > 0"
 			ref="scrollerRef"
 			:items="rows"
 			:item-size="ROW_HEIGHT"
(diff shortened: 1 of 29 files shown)


packages/frontend/editor-ui/src/features/agents/views/AgentSessionTimelineView.vue
```diff
@@ -137,12 +285,10 @@ function formatDuration(ms: number): string {
 	return `${(ms / 1000).toFixed(1)}s`;
 }
```
bjorger: The new session switcher changes the URL but does not reload the session detail. Data is fetched only in `onMounted`, while `onSessionSelect()` pushes the same route with a different threadId; Vue Router will reuse this component, so the header can show the new active session while thread, executions, selection, search, and filters stay from the previous session.

I would extract a loadThreadDetail() and watch projectId/agentId/threadId with reset state.

packages/frontend/editor-ui/src/features/agents/components/SessionTimelineTable.vue
```diff
@@ -17,117 +21,193 @@ const props = defineProps<{
...
+			canScrollDown && $style.canScrollDown,
+		]"
+	>
+		<N8nRecycleScroller
+			ref="scrollerRef"
+			:items="rows"
+			:item-size="ROW_HEIGHT"
+			:scroll-padding="SCROLL_PADDING"
+			item-key="id"
+		>
```
bjorger: This wraps the table in `N8nRecycleScroller`, but rows can be empty when a search/filter has no matching events. `N8nRecycleScroller` currently assumes at least one item, so this can throw at render time. Please render an empty state before mounting the scroller, or only pass it non-empty rows.

bjorger (changes requested):
Absolute banging UI changes - just two findings in the code.

bjorger (approved):

Issue #29619, opened 2026-05-01
[Bug] multipleValues fixedCollection iterates string characters when parameter value is an expression


SoapyRED:
## Describe the bug

When a node parameter is declared as a `fixedCollection` with `typeOptions: { multipleValues: true }`, and the user sets that parameter via an expression (e.g. `={{ $json.items }}`) instead of typing items literally into the UI's "Add Item" flow, `NodeHelpers.getNodeParameters` iterates the **expression string's characters** instead of the resolved array, producing N schema-default items where N is the string's character length.

The result is that nodes receive a phantom array of default-filled items, transparent to the node author. For an 18-character expression the node sees 18 default items instead of the 2-or-N real items the user expected.

## To Reproduce

Minimal reproduction in any community node:

1. Define a parameter:
```ts
   {
     displayName: 'Items',
     name: 'items',
     type: 'fixedCollection',
     typeOptions: { multipleValues: true },
     default: {},
     options: [
       {
         name: 'itemValues',
         displayName: 'Item',
         values: [
           { displayName: 'Foo', name: 'foo', type: 'string', default: 'default-foo' },
         ],
       },
     ],
   }
```

2. In a workflow, place a Set node producing `{ items: [{ foo: 'a' }, { foo: 'b' }] }`.

3. Wire its output to your custom node, and set `items.itemValues` to the expression `={{ $json.items }}`.

4. Inspect the parameter value the node receives via `getNodeParameter('items.itemValues', i)` or the resolved request body if using declarative routing.

**Expected:** node receives `[{ foo: 'a' }, { foo: 'b' }]`.

**Actual:** node receives 18 elements, each `{ foo: 'default-foo' }`. (18 = `'={{ $json.items }}'`.length)

## Root cause

`NodeHelpers.getNodeParameters` in `packages/workflow/src/NodeHelpers.ts:723-770` (master at time of filing) loops over the parameter value with `for (const nodeValue of propertyValues[itemName])`. When `propertyValues[itemName]` is a string (i.e. an unresolved expression that hasn't been evaluated to its array value yet), the for-of iterates the string's characters. Each character is then treated as a fixedCollection row, recursively resolved through `getNodeParameters`, and emitted with schema defaults.

By the time a node's `preSend` hook or operation handler runs, `node.parameters.<param>.<itemKey>` has already been mutated to that defaults-array. The original expression string is unrecoverable from inside the node.

## Impact

[...]

PR #29651, opened 2026-05-02
fix: prevent fixedCollection multipleValues from iterating expression string characters

mixelburg:
## Problem

When a node parameter is declared as a `fixedCollection` with `typeOptions: { multipleValues: true }`, and the user sets that parameter via an expression (e.g. `={{ $json.items }}`) instead of typing items literally into the UI, `NodeHelpers.getNodeParameters` iterates the **expression string's characters** instead of the resolved array, producing N schema-default items where N is the string's character length.

For example, setting `itemValues` to `={{ $json.items }}` (18 characters) produces 18 default-filled items instead of the expected real items.

Reported in [#29619](https://github.com/n8n-io/n8n/issues/29619).

## Root Cause

In `packages/workflow/src/node-helpers.ts` (line ~933), the code does:

```js
for (const nodeValue of (propertyValues as INodeParameters)[itemName] as INodeParameters[]) {
```

When `propertyValues[itemName]` is a string (an unresolved expression), `for...of` iterates the string's characters. Each character is treated as a fixedCollection row and processed with schema defaults.

## Fix

Added an `Array.isArray` check before the iteration loop to skip non-array values:

```js
const itemValues = (propertyValues as INodeParameters)[itemName];
if (!Array.isArray(itemValues)) {
    continue;  // skip unresolved expression strings
}
for (const nodeValue of itemValues as INodeParameters[]) {
```

This prevents the character-iteration bug while preserving the expression string in the raw parameters for later evaluation by the workflow runtime.

## Testing

1. Define a custom node with a `multipleValues: true` fixedCollection parameter
2. Set the parameter to an expression like `={{ $json.items }}`
3. Verify the node does not receive N default items (where N = expression string length)

### DIFF
--- a/packages/workflow/src/node-helpers.ts
+++ b/packages/workflow/src/node-helpers.ts
@@ -941,9 +941,11 @@ export function getNodeParameters(
 					}
 
 					// Iterate over all items as it contains multiple ones
-					for (const nodeValue of (propertyValues as INodeParameters)[
-						itemName
-					] as INodeParameters[]) {
+					const itemValues = (propertyValues as INodeParameters)[itemName];
+					if (!Array.isArray(itemValues)) {
+						continue;
+					}
+					for (const nodeValue of itemValues as INodeParameters[]) {
 						tempNodePropertiesArray = nodePropertyOptions.values!;
 						tempValue = getNodeParameters(
 							tempNodePropertiesArray,


Joffcom:
Hi @mixelburg — thanks for the contribution! Good catch on the character-iteration bug. A few things before we can review:

1. **Sign the CLA.** The cla-bot is still flagging this PR — please sign at https://cla-bot-prod.users.n8n.cloud/webhook/cla and comment `/cla-check` to re-run.
2. **Add a unit test.** Per our [contributing guide](../blob/master/CONTRIBUTING.md), bug fixes need a regression test. A test in `packages/workflow/test/node-helpers.test.ts` (or wherever `getNodeParameters` is exercised) that calls it with a `fixedCollection` set to an unresolved expression string and asserts the output doesn't produce N default rows would lock this in.
3. **Update the PR description to match our [template](../blob/master/.github/pull_request_template.md).** Your `## Problem` / `## Root Cause` / `## Fix` / `## Testing` sections have great content, but the template uses `## Summary`, `## Related Linear tickets, Github issues, and Community forum posts` (your reference to #29619 belongs there), and `## Review / Merge checklist`.
4. **Minor:** title scope — `fix(core):` rather than just `fix:` would help the changelog categorise this correctly.

Once those are sorted we'll route it for review. Thanks!

mixelburg:
Rebased on latest master ✅

The upstream restructuring of the  function caused a large conflict, so I recreated the fix on top of the current code. The actual change remains the same — adding an `Array.isArray` guard before the `for...of` loop on fixedCollection `multipleValues` items to prevent expression strings from being iterated character-by-character.

mixelburg:
Hi! Is there anything else needed from my side on this PR?

mixelburg:
Closing — no longer needed.

PR #29663, opened 2026-05-04
fix(Microsoft Outlook Trigger Node): Use per-folder endpoints for folder-scoped message polling

michael-radency:
## Summary

`GET /me/messages` in Microsoft Graph does not support `$filter` on `parentFolderId` — the filter is silently ignored, returning messages from all folders regardless of the "Folders to Include" setting. This caused the trigger to fire on emails from all folders in production, while appearing to work correctly in editor test mode (which uses `$top=1` with no date filter and happens to return a recent message).

**Fix:** Replace the single `/messages` request with per-folder `GET /me/mailFolders/{id}/messages` calls (one per selected folder), run in parallel via `Promise.all`, and merge the results. When no folders are specified, falls back to `/messages` as before.

**How to test:**
1. Create two folders in Outlook and add an email to each.
2. Set up a Microsoft Outlook Trigger with "Folders to Include" set to only one of the folders.
3. In editor test mode — should return an email from that folder.
4. Activate the trigger and send a new email to the included folder → workflow fires.
5. Send a new email to the excluded folder → workflow does not fire.

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/Microsoft/Outlook/trigger/GenericFunctions.ts
```diff
@@ -34,31 +35,38 @@ export async function getPollResponse(
...
+							...qs,
+						}),
+				),
 			);
+			responseData = results.flat();
 		} else {
 			qs.$top = 1;
-			responseData = await microsoftApiRequest.call(this, 'GET', endpoint, undefined, qs);
-			responseData = responseData.value;
+			const endpoint = folderIds.length > 0 ? `/mailFolders/${folderIds[0]}/messages` : '/messages';
```
BerniWittmann: I found the folloing case: 
2+ Mailbox folders
Folders to include filter: A, B
A has no emails
B has 1 email.

Running in manual mode would try to fetch messages only for folder A, returning no email.
Shouldnt we align this code branch to the "production" branch with fetching all browsers, so that in this example it would output the email

BerniWittmann (changes requested):
Tested it locally worked well in regards to fixing the bug at hand. 

But i think we now introduced two different behaviors between "manual" and "production" mode (see comment for an example case)

BerniWittmann (approved):
LGTM now :+1:

PR #29670, opened 2026-05-04
fix(core): Show AI Builder draft workflows in workflow list

aalises:
## Summary

When the AI Builder creates a workflow, it inserts a row into `ai_builder_temporary_workflow` to mark it as a draft. The orchestrator clears the marker on success (`clearAiTemporary`) or archives the workflow at run-finish (`reapAiTemporaryFromRun`). Both paths require the orchestrator to reach a terminal state.

If the run never finalizes (LangSmith proxy 401 cascade, OOM, kill, tampering throw mid-build, …), the marker stays in place and the workflows-listing filter — a `NOT EXISTS` subquery in `WorkflowRepository.applyFilters` — hides the workflow until the thread expires by TTL. On David's instance this left **11 workflows** invisible with no self-recovery path.

This PR removes the `applyAiBuilderTemporaryFilter` so the marker becomes pure metadata. Mark / promote / reap logic in `instance-ai` is **unchanged** — the table still drives `archiveIfAiTemporary` at run-finish — it just no longer hides the row from listings. Crashed-run workflows immediately become recoverable, and David's 11 zombie workflows reappear with no DB intervention after deploy.

The trade-off is that drafts are visible in the workflows list during a build. Worth a follow-up for a "draft" badge if it turns out to be confusing.

### Follow-ups (not in this PR)

- Unconditional startup sweep that archives orphaned markers whose threads have no active orchestrator (instead of waiting for `threadTtlDays`).
- Investigate the LangSmith proxy 401 cascade that brought the process down on David's instance (separate ticket).
- Audit `validateWorkflowCredentialUsage` scoping for AI-authored builds — the "Blocked workflow update due to tampering attempt" log line on a fresh build looks suspect.

https://linear.app/n8n/issue/INS-194

## Review / Merge checklist

### DIFF
--- a/packages/@n8n/db/src/repositories/__tests__/workflow.repository.test.ts
+++ b/packages/@n8n/db/src/repositories/__tests__/workflow.repository.test.ts
@@ -2,7 +2,7 @@ import { GlobalConfig } from '@n8n/config';
 import { In, type SelectQueryBuilder } from '@n8n/typeorm';
 import { mock } from 'jest-mock-extended';
 
-import { AiBuilderTemporaryWorkflow, WorkflowEntity } from '../../entities';
+import { WorkflowEntity } from '../../entities';
 import { mockEntityManager } from '../../utils/test-utils/mock-entity-manager';
 import { mockInstance } from '../../utils/test-utils/mock-instance';
 import { FolderRepository } from '../folder.repository';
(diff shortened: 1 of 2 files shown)


Cadiac (approved):
👍 I think this is better until we make the marker clearing logic more robust

PR #29673, opened 2026-05-04
fix(editor): Ignore paste events on read-only canvas

Flexicon:
## Summary

Pasting nodes (Ctrl/Cmd+V) on an archived workflow was bypassing the canvas read-only guard, triggering autosave attempts that failed with repeated 403/404 toast errors.

**Root cause:** `onClipboardPaste` in `NodeView.vue` guarded execution via `checkIfEditingIsAllowed()`, which does not account for the archived state. By contrast, `isCanvasReadOnly` (which already includes the archived check, added in #29559) was not consulted.

**Fix:** Add `isCanvasReadOnly.value` as the first condition in `onClipboardPaste`'s early-return guard — the same pattern already used by `onCutNodes`.

**How to test:**
1. Create and save a workflow with at least one node
2. Copy the node (Ctrl/Cmd+C)
3. Archive the workflow via the workflow menu
4. Open the archived workflow — canvas should be in read-only mode
5. Press Ctrl/Cmd+V — no nodes should appear, no autosave should fire, no error toasts should appear

https://linear.app/n8n/issue/CAT-2995

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/frontend/editor-ui/src/app/views/NodeView.vue
+++ b/packages/frontend/editor-ui/src/app/views/NodeView.vue
@@ -518,6 +518,7 @@ async function onCopyNodes(ids: string[]) {
 
 async function onClipboardPaste(plainTextData: string): Promise<void> {
 	if (
+		isCanvasReadOnly.value ||
 		getNodeViewTab(route) !== MAIN_HEADER_TABS.WORKFLOW ||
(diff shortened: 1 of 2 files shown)


r00gm (approved):
🚀

PR #29674, opened 2026-05-04
refactor(core): Use the common SSRF service on instance AI and harden web fetch

Cadiac:
## Summary

Drop the custom SSRF guard from instance AI and instead use the common `SsrfProtectionService` also at instance AI web fetch. Unlike elsewhere on n8n just force it enabled regardless of `N8N_SSRF_PROTECTION_ENABLED` - this is new functionality and we don't need to worry about backwards compatibility here (the checks were at caller sites instead of on the service). This refactor also means that these checks will respect `N8N_SSRF_ALLOWED_HOSTNAMES` in future.

I tested this locally a fair amount, seemed to cover the cases we want to protect against.

https://linear.app/n8n/issue/INS-189/improve-instance-ai-external-request-controls first issue

## Review / Merge checklist

- [x] Tests included. 


tomi (approved):
LGTM 🚀

PR #29676, opened 2026-05-04
fix(editor): Make textarea resize handle accessible in NDV

CharlieKolb:
## Summary

In the NDV, the textarea fullscreen ("open in modal") button was absolutely positioned in the bottom-right corner — directly on top of the browser's native textarea resize handle — so the resize affordance was unreachable. This PR moves the fullscreen button to the top-right corner of the plain textarea variant only (code editors that share the same class are unaffected since they have no native resize handle), switches the textarea from `resize: vertical` to `resize: both` with `max-width: 100%` so the resize indicator renders as a corner grip in the bottom-right rather than a horizontal grip in the bottom-center, and zeroes the input wrapper's flex `gap` for this variant so the corner grip sits flush with the visual corner instead of being inset by a spacing unit. Test by opening any string parameter with rows > 1 in NDV: the fullscreen button should sit in the top-right corner and the resize handle should be grabbable from the bottom-right corner.

https://linear.app/n8n/issue/ADO-5155

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInput.vue
+++ b/packages/frontend/editor-ui/src/features/ndv/parameters/components/ParameterInput.vue
@@ -2215,6 +2215,26 @@ onUpdated(async () => {
 	}
 }
 
+.input-with-opener .textarea-modal-opener {
+	top: 1px;
+	bottom: auto;
+	border-top: none;
+	border-bottom: var(--border);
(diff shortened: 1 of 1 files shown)


dariacodes (approved):
I tested it, looks good! 🚀

PR #29677, opened 2026-05-04
feat(agents): Tidy up trigger UI

bjorger:
## Summary

- **Schedule refresh:** Synced trigger status after schedule updates so newly added schedules appear immediately without a page refresh.
- **Schedule active state:** Replaced the old deactivate behavior with an Active toggle that keeps the schedule visible while switching enabled state.
- **Schedule actions:** Added Cancel and Save buttons matching the app’s modal footer conventions, with Save closing the modal.
- **Schedule disconnect:** Restored schedule removal using the same destructive Disconnect button pattern used by Slack, Telegram, Linear, and other integrations.
- **Integration modal layout:** Moved integration action buttons into the modal footer so Disconnect sits bottom-right consistently.
- **Test coverage:** Updated schedule trigger tests for save, cancel, active toggling, disconnect, and emitted status changes.

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/features/agents/components/AgentScheduleTriggerCard.vue
```diff
@@ -101,99 +112,106 @@ async function loadConfig() {
...
+				cronExpression: nextCronExpression,
 			},
 		);
-		applyConfig(config);
-		return true;
 	} catch (error) {
 		setSaveError(error, 'agents.schedule.saveError');
-		return false;
-	} finally {
-		saving.value = false;
```
burivuhster, line 132: Was removing `saving` flag update here intentional?

packages/frontend/editor-ui/src/features/agents/components/AgentScheduleTriggerCard.vue
```diff
@@ -101,99 +112,106 @@ async function loadConfig() {
...
+async function onDisconnect() {
+	if (!isConfigured.value || loading.value || saving.value) return;
+	saving.value = true;
+	clearErrors();
 
-watch(cronExpression, () => {
-	if (hydrating.value) {
-		return;
+	try {
+		if (lastSavedActive.value) {
```
burivuhster: Would it make sense to always send the request? It should be idempotent, so it'll make no harm if it was already disabled.

packages/frontend/editor-ui/src/features/agents/composables/useAgentBuilderTelemetry.ts
```diff
@@ -246,10 +248,29 @@ export function useAgentBuilderTelemetry(deps: AgentBuilderTelemetryDeps) {
 				deps.projectId.value,
 				deps.agentId.value,
 			);
```
burivuhster: Do we need this after I added integrations to the agent schema in [this PR](https://github.com/n8n-io/n8n/pull/29536)? Backend should put integrations into the agent json schema now, so it should be fetched already

burivuhster (commented):
Hey, can you please check if this needs to be changed after exposing integrations in the schema JSON?

packages/frontend/editor-ui/src/features/agents/components/AgentScheduleTriggerCard.vue
bjorger: Yep, that was intentional :)

I moved the “start saving / stop saving” state out of the small cron-save helper and into the actual button actions.

packages/frontend/editor-ui/src/features/agents/components/AgentScheduleTriggerCard.vue
bjorger: Good catch!

mike12345567 (approved):

PR #29679, opened 2026-05-04
fix(core): Add timeout to external secrets provider refresh

alielkhateeb:
## Summary

https://www.loom.com/share/50d1b99ee6fb464783fafa67dbe05ec3

An unresponsive external secrets provider (e.g. Vault, AWS Secrets Manager) could cause `provider.update()` to hang indefinitely, blocking instance startup and periodic refresh cycles.

This PR wraps each provider refresh in a `Promise.race` against a 20-second timeout. If the provider does not respond within that window, the refresh is aborted with an `OperationalError` and the error is logged — the instance continues normally instead of hanging.

**How to test:**
1. Configure an external secrets store that is unreachable or slow to respond.
2. Start the n8n instance and verify it starts up within the timeout window rather than hanging indefinitely.
3. Check logs for the timeout error message on the affected provider.

***OR test the failure/timeout with:***
1. Replicate a never-ending promise with `await new Promise(() => {});`
2. Start the n8n instance

https://linear.app/n8n/issue/LIGO-522

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/modules/external-secrets.ee/constants.ts
+++ b/packages/cli/src/modules/external-secrets.ee/constants.ts
@@ -3,8 +3,7 @@ import type { INodeProperties } from 'n8n-workflow';
 export const EXTERNAL_SECRETS_DB_KEY = 'feature.externalSecrets';
 export const EXTERNAL_SECRETS_INITIAL_BACKOFF = 10 * 1000;
 export const EXTERNAL_SECRETS_MAX_BACKOFF = 5 * 60 * 1000;
-/** Max wait for a single provider refresh (startup and periodic refresh) */
-export const EXTERNAL_SECRETS_REFRESH_TIMEOUT_MS = 120 * 1000;
+export const EXTERNAL_SECRETS_REFRESH_TIMEOUT_MS = 20_000;
 
 export const DOCS_HELP_NOTICE: INodeProperties = {
 	displayName:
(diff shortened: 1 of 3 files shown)


sandra0503 (approved):

packages/cli/src/modules/external-secrets.ee/constants.ts
```diff
@@ -3,6 +3,7 @@ import type { INodeProperties } from 'n8n-workflow';
 export const EXTERNAL_SECRETS_DB_KEY = 'feature.externalSecrets';
 export const EXTERNAL_SECRETS_INITIAL_BACKOFF = 10 * 1000;
 export const EXTERNAL_SECRETS_MAX_BACKOFF = 5 * 60 * 1000;
+export const EXTERNAL_SECRETS_REFRESH_TIMEOUT_MS = 20_000;
```
konstantintieber, line 6: What was the rationale for the 20 second value?
One thing to note about our hashicorp provider: that one creates hundreds of http requests on refresh (iirc it's one per secret).
I don't have any data on how long this takes.

alielkhateeb: I have created dummy secrets in vault (900+) and tested the timeout, vault secrets were all loaded and GCP (Hanging) kept trying for 20 seconds, I did 30 seconds first then I thought it's too long, because you have to wait until it's over before you can actually access your instance.

image

konstantintieber: Nice 👍

PR #29681, opened 2026-05-04
fix(MCP Client Node): Surface errors as node failures in new node versions

michael-radency:
## Summary

The MCP Client node and the MCP Client Tool node were ignoring the `isError: true` flag in MCP tool call responses. When an upstream MCP server returned a tool-level error per the MCP spec (`{ isError: true, content: [...] }`), n8n treated it as a successful execution — the node stayed green and the error content was silently passed through as output data.

**Fix:** Both nodes now check `result.isError` after each tool call and throw a `NodeOperationError` with the error message from the response content. This turns the node red with the error surfaced, and respects "Continue On Fail (using error output)" if enabled.

**Versioning:** The fix is applied only to new node versions to avoid breaking existing workflows:
- **MCP Client** node: fixed in v1.1 (new default), v1 keeps legacy behaviour
- **MCP Client Tool** node: fixed in v1.3 (new default), v1/1.1/1.2 keep legacy behaviour

**To test:**
1. Set up an MCP server that returns error
2. Drop a fresh **MCP Client** node (v1.1) or **MCP Client Tool** node (v1.3) → execute → node should turn red with "boom" as the error message
3. Enable "Continue On Fail (using error output)" → node should stay green and emit the error message as output
4. Downgrade to v1 / v1.2 → node should stay green (legacy behaviour preserved)

https://linear.app/n8n/issue/NODE-4952/mcp-node-mcp-client-node-ignores-iserror-true-from-mcp-server-treats

## Review / Merge checklist

- [ ] Tests included. 


ShireenMissi (dismissed):
The code looks good, I tested locally and it works as expected, the only test case that didn't work is when MCP client tool 1.3 errors and the agent is sat to continue on fail the workflow flow still errors 

Image

also there are some conflicts with mater to resolve before merging

michael-radency:
> The code looks good, I tested locally and it works as expected, the only test case that didn't work is when MCP client tool 1.3 errors and the agent is sat to continue on fail the workflow flow still errors
> Image
> 
> also there are some conflicts with mater to resolve before merging

I now somewhat confused, but this seems to be aligned with other tools
image
image
image

alexander-gekov (approved):
Works as expected! (Tool Continue on Fail support to be addressed in separate PR)

PR #29683, opened 2026-05-04
fix/docker permission error on community node installation due to missing tar flag

pReya:
## Summary

Busybox `tar` (used in the Alpine-based n8n Docker image) preserves UID/GID from archive headers even for non-root users. npm tarballs carry `uid=0/gid=0` headers, so extracted files end up owned by `root` when n8n runs as a custom UID — causing `EACCES` on the subsequent `package.json` read.

**Fix:** add `--no-same-owner` to the `tar` invocation in `downloadPackage`. Both GNU tar and busybox tar honour this flag, instructing tar to use the current process UID/GID instead of the archive's stored ownership.

## Root cause

The official n8n Docker image is Alpine-based, where `tar` is provided by busybox. Busybox `tar`, unlike GNU `tar`, by default preserves UID/GID metadata from the tarball even when extracting as a non-root user. npm-published tarballs typically contain `uid=0/gid=0` in their headers, so files end up owned by `root`. When n8n then tries to read `package.json` as the runtime UID, it gets `EACCES`.

This only surfaces when overriding the user via `docker run --user <UID>:<GID>`, which is the documented and recommended way to control file ownership on bind-mounted volumes.

## Changes

- `community-packages.service.ts` — adds `--no-same-owner` to the `tar` args
- `community-packages.service.test.ts` — updates the `execFile` assertion to match

No behaviour change for users running as `root` or the default `node` user (UID 1000).

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
+++ b/packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
@@ -451,7 +451,7 @@ describe('CommunityPackagesService', () => {
 			expect(execFile).toHaveBeenCalledTimes(1);
 			expect(execFile).toHaveBeenCalledWith(
 				'tar',
-				['-xzf', testBlockTarballName, '-C', testBlockPackageDir, '--strip-components=1'],
+				['-xzf', testBlockTarballName, '-C', testBlockPackageDir, '--strip-components=1', '--no-same-owner'],
 				{ cwd: testBlockDownloadDir },
 				expect.any(Function),
 			);
--- a/packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
+++ b/packages/cli/src/modules/community-packages/__tests__/community-packages.service.test.ts
@@ -451,7 +451,14 @@ describe('CommunityPackagesService', () => {
 			expect(execFile).toHaveBeenCalledTimes(1);
 			expect(execFile).toHaveBeenCalledWith(
 				'tar',
-				['-xzf', testBlockTarballName, '-C', testBlockPackageDir, '--strip-components=1', '--no-same-owner'],
+				[
+					'-xzf',
+					testBlockTarballName,
+					'-C',
+					testBlockPackageDir,
+					'--strip-components=1',
+					'--no-same-owner',
+				],
 				{ cwd: testBlockDownloadDir },
 				expect.any(Function),
 			);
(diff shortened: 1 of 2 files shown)


Joffcom:
Hey @pReya 

Can you provide an example of when you see this issue as it is not something I have seen when installing community nodes.

pReya:
> Hey @pReya
> 
> Can you provide an example of when you see this issue as it is not something I have seen when installing community nodes.

Sure. I've tried to come up with a reproduction with the help of Claude and successfully tried the Compose based reproduction on my local Mac system.

## Minimal reproduction (no n8n install required)

This isolates the exact `tar` invocation from `community-packages.service.ts`:

```sh
docker run --rm --user 3001:3001 --entrypoint sh docker.n8n.io/n8nio/n8n -c '
  cd /tmp
  wget -q https://registry.npmjs.org/n8n-nodes-imap/-/n8n-nodes-imap-2.17.0.tgz
  mkdir out
  tar -xzf n8n-nodes-imap-2.17.0.tgz -C out --strip-components=1
  echo "Running as UID $(id -u); extracted file owner:"
  ls -la out/package.json
'
```

**Actual output (bug):**
```
Running as UID 3001; extracted file owner:
-rw-r--r--    1 root     root  ...  out/package.json
```

**Expected output (matches GNU tar behavior):**
```
-rw-r--r--    1 3001     3001  ...  out/package.json
```

Adding `--no-same-owner` to the same `tar` invocation produces the expected
output — busybox tar supports this flag and it makes its behavior match GNU tar's
default for non-root extractors.

## End-to-end reproduction

### compose.yaml

```yaml
services:
  init:
    image: alpine
    command: chown -R 3001:3001 /data
    volumes:
      - n8n_data:/data
  n8n:
    image: docker.n8n.io/n8nio/n8n
    user: "3001:3001"
    depends_on:
      init:
        condition: service_completed_successfully
    ports:
      - "5678:5678"
    volumes:
      - n8n_data:/data
    environment:
      - N8N_USER_FOLDER=/data
      - N8N_COMMUNITY_PACKAGES_ENABLED=true
      - N8N_COMMUNITY_PACKAGES_ALLOW_TOOL_USAGE=true

volumes:
  n8n_data:
```

> A named volume + init container is used so the reproduction works on macOS
> Docker Desktop (where bind-mount UIDs are not preserved) as well as native Linux.
> A bind mount with `chown 3001:3001` on the host works equivalently on Linux.

### Steps

1. `docker compose up -d`
2. Open http://localhost:5678 and complete first-run setup.
3. **Settings → Community Nodes → Install** → enter `n8n-nodes-imap` → Install.
4. Observe the error toast:
   ```
   Error loading package "n8n-nodes-imap": EACCES: permission denied,
   open '/data/.n8n/nodes/node_modules/n8n-nodes-imap/package.json'
   ```
5. Confirm root cause:
   ```sh
   docker compose exec n8n sh -c \
     'id; ls -la /data/.n8n/nodes/node_modules/n8n-nodes-imap/'
   ```
   The container reports `uid=3001`, yet `package.json`, `LICENSE.md`, and
   `README.md` are owned by `root`.

## Environments tested

- macOS, Docker Desktop 4.x, image `docker.n8n.io/n8nio/n8n` (latest)
- Linux, native Docker, bind mount with host directory `chown 3001:3001`

Both reproduce the bug.

pReya:
@Joffcom Not sure if you saw my answer, since I did not mention you in the post. Can you give this another look?

Joffcom:
@pReya Looking at your reproduction you are changing the user which is not something we currently support so I would expect you to see a permission error. This makes sense now I read your initial description again, Keep in mind that the n8n application will run under the `node` user so just this one change may cause other issues.

I will pass this to one of the teams for review but it is possible that we may not accept this at the moment.

pReya:
> @pReya Looking at your reproduction you are changing the user which is not something we currently support so I would expect you to see a permission error. This makes sense now I read your initial description again, Keep in mind that the n8n application will run under the `node` user so just this one change may cause other issues.
> 
> I will pass this to one of the teams for review but it is possible that we may not accept this at the moment.

@Joffcom In this case, a documentation update would be nice to let users know, that n8n does not support running as a different user via docker `--user` flag.

Adding this tar flag still seems to be a good idea for consistent file permissions within the n8n file system – even without supporting the `--user` attribute. All files that are managed by n8n having the same owner/group seems to be a reasonable expectation and not having that might lead to more errors in the future. Especially since the behavior between a bare metal installation and a docker installation is different right now. It makes sense to keep the ownership consistent between the two installation methods (which this PR achieves).

P.S.: Sorry, I tried renaming the PR but GH decided to close the PR when I renamed the branch in my fork. Did not intend to close it.

PR #29688, opened 2026-05-04
fix(Notion Node): Serialize staticData as ISO string in NotionTrigger

alexander-gekov:
## Summary

The Notion Trigger node stored a live `moment.js` object in `webhookData.lastTimeChecked`. When serialized to JSON for persistence, this became a complex nested object (`{_isAMomentObject: true, _d: "..."}`). On workflow restart, `moment()` could not parse it back, causing activation errors.

**Fix:** Added `.utc().format()` to convert the moment object to a plain ISO string before storing in `staticData`, ensuring the value survives JSON round-trip serialization.

**How to test:**
1. Set up a Notion Trigger node pointing to any Notion database
2. Activate the workflow and let it poll at least once
3. Deactivate and reactivate the workflow or restart n8n
4. Before the fix: Error in logs, workflow non-affected After the fix: activates cleanly and no log error

https://linear.app/n8n/issue/NODE-4849

## Review / Merge checklist

- [x] Tests included.

---
🤖 PR Summary generated by AI


alexander-gekov:
Closing — the bug is not reproducible. moment objects serialize correctly via JSON.stringify() thanks to moment's .toJSON() method. No crash occurs on workflow restart.

alexander-gekov:
Reopening - on second look, the bug is reproducible. The error occurs because staticData is wrapped in an observable Proxy, and when TypeORM calls JSON.stringify on it, moment's .toJSON() runs with `this` bound to the Proxy instead of the real moment object, breaking Date.prototype.getTime.

However, the impact is limited to a TypeError in the logs ("this is not a Date object") when saving staticData. The workflow still activates and polls correctly despite the error.

Still worth fixing: explicitly casting to an ISO string avoids the error and removes any ambiguity about what gets stored in staticData.

DawidMyslak (dismissed):
LGTM!

packages/nodes-base/nodes/Notion/test/NotionTrigger.test.ts
```diff
@@ -0,0 +1,140 @@
...
+		},
+	};
+}
+
+describe('NotionTrigger', () => {
+	beforeEach(() => {
+		jest.clearAllMocks();
+	});
+
+	describe('staticData serialization (NODE-4849)', () => {
```
DawidMyslak: Not sure about leaving ticket IDs in tests?

DawidMyslak (dismissed):
Looks good, minor feedback 👍

packages/nodes-base/nodes/Notion/test/NotionTrigger.test.ts
alexander-gekov: removed

DawidMyslak (dismissed):

DawidMyslak (approved):

PR #29690, opened 2026-05-04
fix(Notion Node): Paginate Get Many operations beyond 100-item API cap

alexander-gekov:
## Summary

The Notion "Get Many" operations (database pages, databases, blocks) were capped at 100 results because the node made a single API call with `page_size` set to the user's limit. Since Notion's API caps `page_size` at 100, any limit > 100 was silently truncated.

**Fix:** Changed all 6 affected operations across V1 and V2 (`databasePage.getAll`, `database.getAll`, `block.getAll`) to use `notionApiRequestAllItems` with proper cursor-based pagination. Each request caps `page_size` at `Math.min(limit, 100)` and the final result is sliced to the exact user limit.

**How to test:**
1. Create a Notion database with more than 100 pages
2. Add a Notion node with "Get Many" operation, set limit to 150
3. Execute the workflow
4. Before the fix: returns exactly 100 items. After the fix: returns 150 items
5. Also test with `returnAll = true` to confirm existing pagination still works

https://linear.app/n8n/issue/NODE-4612

## Review / Merge checklist

- [x] Tests included.

---
🤖 PR Summary generated by AI


packages/nodes-base/nodes/Notion/v2/NotionV2.node.ts
```diff
@@ -484,15 +490,17 @@ export class NotionV2 implements INodeType {
...
 								this,
+								'results',
 								'POST',
 								`/databases/${databaseId}/query`,
 								body,
-								qs,
+								{ limit },
 							);
-							responseData = responseData.results;
+							responseData = responseData.slice(0, limit);
```
michael-radency: maybe` responseData.slice(0, limit);` should be part of `notionApiRequestAllItems`, we seems to be doing this after each request

packages/nodes-base/nodes/Notion/v1/NotionV1.node.ts
```diff
@@ -272,15 +272,16 @@ export class NotionV1 implements INodeType {
...
 							this,
+							'results',
 							'GET',
 							`/blocks/${blockId}/children`,
 							{},
-							qs,
+							{ page_size: Math.min(limit, 100), limit },
 						);
-						responseData = responseData.results;
+						responseData = responseData.slice(0, limit);
```
michael-radency: I would avoid modify v1, to avoid breaking changes and potentially introducing regressions

michael-radency (dismissed):
LGTM, added two comments, both non blocking

packages/nodes-base/nodes/Notion/v2/NotionV2.node.ts
alexander-gekov: Good call — moved into `notionApiRequestAllItems` in d6a2d105e2 so callers don't need to slice.

packages/nodes-base/nodes/Notion/v1/NotionV1.node.ts
alexander-gekov: Agreed — reverted V1 to master in d6a2d105e2. Only V2 gets the pagination fix; V1 stays untouched.

michael-radency (dismissed):

michael-radency (approved):

PR #29694, opened 2026-05-04
feat(core): Use McpManagerClient and enforce whether MCP server connections are allowed

Cadiac:
## Summary

We had `McpManagerClient` at instance-ai but it was actually just dead code - the actual MCP servers were handled via inline `buildMcpServers` logic and state on the agent. Encapsulate that logic into the `McpManagerClient` like we probably meant to do earlier but left that half finished, and enforce that our SSRF external connection rules are enforced also when connecting to external MCP servers through this mechanism when `N8N_SSRF_PROTECTION_ENABLED` is enabled (on cloud it is).

https://linear.app/n8n/issue/INS-189/improve-instance-ai-external-request-controls third issue

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/instance-ai/src/mcp/mcp-client-manager.ts
```diff
@@ -1,38 +1,135 @@
...
+			}
+
+			if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+				throw new UserError(
+					`MCP server "${server.name}": only http(s) URLs are allowed, got "${parsed.protocol}"`,
+				);
+			}
+
+			if (this.ssrfValidator) {
+				const result = await this.ssrfValidator.validateUrl(server.url);
```
aalises, line 156: This validates the URL before constructing `MCPClient`, but the actual transport still performs its own DNS lookup and may follow redirects. 

Perhaps this leaves a gap versus the shared SSRF path, which also enforces lookup/redirect checks at connection time. 

Perhaps we can wire those checks into the MCP transport path

packages/@n8n/instance-ai/src/mcp/mcp-client-manager.ts
```diff
@@ -1,38 +1,135 @@
...
+		if (configs.length === 0) return {};
+
+		const key = JSON.stringify(configs);
+		const cached = this.regularToolsByKey.get(key);
+		if (cached) return cached;
 
-		this.mcpClient = new MCPClient({ servers: serverMap });
-		return await this.mcpClient.listTools();
+		await this.validateConfigs(configs);
+		const tools = await this.connectAndListTools(`mcp-${nanoid(6)}`, configs, key);
```
aalises: If two runs request the same uncached config at once, both can create an `MCPClient`; the later one overwrites the tracked entry, leaving the earlier client undisconnectable. Some cache/share the in-flight creation promise by key would maybe do it

packages/@n8n/instance-ai/src/mcp/mcp-client-manager.ts
```diff
@@ -1,38 +1,135 @@
...
+		this.regularToolsByKey.set(key, tools);
```
aalises: When good ole' MCP admin settings change, old config keys remain cached until service shutdown. That means removed servers can stay connected even after they are no longer configured. We could disconnect/prune clients for configs that are no longer active, but not a bigige

packages/@n8n/instance-ai/src/mcp/mcp-client-manager.ts
Cadiac: I tried wiring those, but the end result was rather messy - do you think this is worth doing now with us potentially dropping Mastra and it's MCP handling in favor of our own agents SDK? Given that only admins can configure these the threatl doesn't seem that high right now.

packages/@n8n/instance-ai/src/mcp/mcp-client-manager.ts
Cadiac: Addressed this, changing the settings now emits an event that the service listens to (similar to how LDAP handles settings changes). Tested this by switching between two servers.

packages/@n8n/instance-ai/src/mcp/mcp-client-manager.ts
Cadiac: Addressed this, now there's a mechanism to handle concurrent pending connections

aalises (approved):

PR #29696, opened 2026-05-04
refactor(core): Extract leader election client and improve robustness (no-changelog)

tomi:
## Summary

Make the leader election more robust:
- Use a separate redis connection with tighter command timeout. The previous implementation didn’t have any timeouts. Extract the communication part to its own class
- Prevent overlapping leadership checks, which can cause odd behaviour
- Make the code easier to reason by splitting the paths based on whether we think we are the leader or not
- Make the leader check & TTL extension atomic using a single LUA script to prevent race conditions
- In case we are the leader and the leader key has expired, only step down if we are not able to reacquire the key. This prevents possible flip-flops

https://linear.app/n8n/issue/CAT-2762

## Review / Merge checklist

- [ ] Tests included.


packages/cli/src/scaling/multi-main-setup-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+		} else {
+			this.instanceSettings.markAsFollower();
+		}
+	}
+
+	async shutdown() {
+		const { isLeader } = this.instanceSettings;
+
+		if (isLeader) {
+			const result = await this.client.clearLeader();
```
mfsiega, line 43: The previous draft had a TODO to only clear the leader key if it's our own key. I do think we should do that, so maybe at least bring the TODO back? (Or implement already, if it's easy.)

packages/cli/src/scaling/multi-main-setup-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+			return;
+		}
+
+		const wasSet = result.result;
+		if (wasSet) {
+			this.becomeLeader();
+		}
+	}
+
+	private becomeLeader() {
```
mfsiega: Nit: since we use the terms "takeover" and "stepdown" elsewhere, it might be _slightly_ clearer to say "takeOverAsLeader" and "stepDownToFollower"?

packages/cli/src/scaling/multi-main-setup-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+		}
+	}
+
+	private async checkAreWeStillLeader() {
+		assert(this.instanceSettings.isLeader);
+
+		const renewTtlResult = await this.client.tryRenewLeaderTtl();
+		if (!renewTtlResult.ok) {
+			this.logRedisCommandFailure('Failed to renew leader TTL', renewTtlResult.error);
+			return;
```
mfsiega, line 82: This is the bit I feel least confident about. There are two cases that can occur. 

On one hand, maybe Redis is unavailable for everyone. The leader key will expire, but when it comes back, we'll regain leadership. In this case, by ignoring renewal failure, we avoid an unnecessary swap.

On the other hand, maybe Redis is only unavailable for us. In this case we will stay the leader but some other node might also take over leadership. We could slightly mitigate this by keeping track of the TTL locally, and stepping down when we know we haven't renewed recently enough. (It even gets worse - maybe we DID renew our key, but eg timed out before the response reached us. Or some other weird situations.)

I think on the balance, I'd rather do what you've done and accept some occasional inconsistency for now. Probably worth a comment explaining the situation - "yes this can cause inconsistency, but the alternatives also bring issues, so for now we're going with this assumption".

packages/cli/src/scaling/multi-main-setup-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+		this.logger.warn('[Multi-main setup] Leader failed to renew leader key');
+
+		if (renewalResult.id === 'other-host-is-leader') {
+			this.logger.debug(
+				`[Instance ID ${this.hostId}] Leader is other instance "${renewalResult.currentLeaderId}"`,
+			);
+			this.becomeFollower();
+			return;
+		}
+
```
mfsiega, line 100: IIUC the other possible `renewalResult.id` is `key-is-missing`? If so, we should make it a bit clearer here that that's the case we're handling (maybe just a comment) and explain - the key has expired somehow, so we're going to try to stay the leader.

packages/cli/src/scaling/multi-main-setup-v2.ts
```diff
@@ -0,0 +1,178 @@
...
+				await this.checkAreWeStillLeader();
+			} else {
+				await this.checkCanBecomeLeader();
+			}
+		} finally {
+			this.leaderCheckInProgress = false;
+		}
+	}
+
+	private async checkAreWeStillLeader() {
```
mfsiega, line 71: Would be good to have a small comment explaining what this does: we try to renew our leadership lease. If we've lost the lease, we step down to become a follower.

packages/cli/src/scaling/multi-main-setup-legacy.ts
```diff
@@ -0,0 +1,120 @@
...
+			await this.publisher.setExpiration(this.leaderKey, this.leaderKeyTtl);
+
+			return;
+		}
+
+		if (leaderId && leaderId !== hostId) {
+			this.logger.debug(`[Instance ID ${hostId}] Leader is other instance "${leaderId}"`);
+
+			if (this.instanceSettings.isLeader) {
+				this.instanceSettings.markAsFollower();
```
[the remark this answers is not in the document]
tomi: That's the legacy behaviour we are fixing with the new one

packages/cli/src/scaling/multi-main-setup-v2.ts
tomi: Elaborated the comment

packages/cli/src/scaling/multi-main-setup-v2.ts
tomi: Good point! Renamed

mfsiega (approved):
I think we could probably spend time agonizing on various details here, but I'm still hoping that this code doesn't live too much longer, so I don't think that's a good investment of time. This impl seems slightly cleaner and easier to understand than the previous one, and handle some specific cases slightly better, so LGTM.

packages/cli/src/scaling/__tests__/multi-main-setup.ee.test.ts
```diff
@@ -1,110 +1,382 @@
...
-			publisher.get.mockResolvedValue('main-n8n-main-1');
-			const emit = jest.spyOn(multiMainSetup, 'emit');
+		describe('checkLeader (leader path)', () => {
+			beforeEach(async () => {
+				client.setLeaderIfNotExists.mockResolvedValue(createResultOk(true));
+				await multiMainSetup.init();
+				jest.clearAllMocks();
+			});
+
+			it('should stay leader when TTL renewal succeeds', async () => {
```
mfsiega, line 234: Should we test the case when TTL renewal errors?

tomi: The TTL renewal error case is already covered at line 291: 'should stay leader when TTL renewal Redis command fails'

PR #29698, opened 2026-05-04
feat(core): Add n8n-object-validation ESLint rule for community nodes

garritfra:
## Summary

Adds a new ESLint rule `@n8n/community-nodes/n8n-object-validation` that validates the structure of the `n8n` object in community node `package.json` files.

The rule enforces:
- A top-level `n8n` object must exist
- `n8nNodesApiVersion` must live inside `n8n` (not at the root) and be a positive integer
- `n8n.nodes` must be a non-empty array of strings, each starting with `dist/`
- `n8n.credentials` (optional) must be an array of strings, each starting with `dist/`

The `dist/` prefix is required because community nodes ship compiled JavaScript; n8n cannot load TypeScript sources at runtime. Variants such as `./dist/...` or `DIST/...` are rejected to match the convention used by `@n8n/node-cli` and `@n8n/create-node` templates.

Registered as `error` in both `recommended` and `recommendedWithoutN8nCloudSupport` configs, so it runs both in `@n8n/node-cli` (developer-side linting) and `@n8n/scan-community-package` (Cloud admission scanner).

### Subsumes external rules (prep for [CE-856](https://linear.app/n8n/issue/CE-856))

This rule replaces the following rules from `eslint-plugin-n8n-nodes-base`, so that dependency can be removed in a follow-up PR:

| External rule | Replaced by |
|---|---|
| `community-package-json-n8n-missing` | `missingN8nObject` |
| `community-package-json-n8n-api-version-missing` | `missingNodesApiVersion` |
| `community-package-json-n8n-api-version-not-number` | `invalidNodesApiVersion` |
| `community-package-json-n8n-nodes-missing` | `missingN8nNodes` |
| `community-package-json-n8n-nodes-empty` | `emptyN8nNodes` |

### Behaviour change vs. upstream

The new rule is **stricter** than `community-package-json-n8n-api-version-not-number`: per the ticket spec, `n8nNodesApiVersion` must be a *positive integer*, so values like `0`, `-1`, or `1.5` are now rejected (upstream only checked for non-number). Newly published community packages with a non-integer API version will fail the Cloud scan.

https://linear.app/n8n/issue/CE-741

Unblocks: https://linear.app/n8n/issue/CE-856

## Review / Merge checklist

- [x] Tests included (25 cases covering valid and invalid `package.json` shapes, including each subsumed external rule).
- [x] Docs updated (`docs/rules/n8n-object-validation.md` + README rule table).


garritfra:
@cubic-dev-ai I looked into this and the violation is a **false positive** — no fix needed.

The detection at `n8n-object-validation.ts:61` is bounded by `getTopLevelObjectInJson` (`src/utils/ast-utils.ts`), which guards against exactly this case:

```ts
export function getTopLevelObjectInJson(node) {
    if (node.parent?.type !== AST_NODE_TYPES.ExpressionStatement) {
        return null;
    }
    return node;
}
```

In a JSON file parsed as JS, only the program's sole root object has `ExpressionStatement` as its parent. Nested objects bail out:

- objects inside arrays → parent is `ArrayExpression` → returns `null`
- objects assigned to keys → parent is `Property` → returns `null`

So the `ObjectExpression` visitor fires for every object, but immediately returns unless it's the actual root. There's also a dedicated regression test for this exact scenario:

```ts
{
    name: 'objects nested inside arrays are not treated as the package root',
    filename: 'package.json',
    code: '{ "name": "n8n-nodes-example", "n8n": { ... }, "contributors": [{ "name": "Jane" }, { "name": "John" }] }',
}
```

If the detection were too broad, each `contributors[]` object (none of which contain `n8n`/`nodes`/etc.) would fire `missingN8nObject` and the test would fail. All 26 tests pass.

Joffcom (approved):

PR #29699, opened 2026-05-04
fix(core): Add workflow structure validation

dariacodes:
## Summary

**Problem**: Workflows created via the API could be malformed (e.g., missing node fields, broken connections, or duplicate names) yet still be saved successfully. When opened in the editor, these workflows would crash with the error "Could not find workflow." This failure was effectively invisible to users - the error message disappeared almost immediately, and they were redirected to a new workflow page without any clear indication of what went wrong.

**Solution**: This PR adds a shared structural validator in `n8n-workflow` that is the single source of truth for what constitutes a valid workflow shape. It is intentionally limited to structure only (required fields, correct types, graph integrity) and does not touch activation/semantic validation.

**Backend**:
- rejects malformed payloads on create, update, and import with a `400` and detailed issue description.

**Frontend**:
- **Open existing workflow / import (file/paste/URL)**: non-blocking. Sanitizes malformed data (filters nodes without `type`, fixes bad positions, strips broken connections) and renders the remaining valid nodes. Shows a warning toast. If `initializeWorkspace` still throws, the error is reported to Sentry via `console.error` as a safety net.
- **Template from JSON**: blocking. Rejects invalid data and shows an error toast without rendering.

Additionally, both the open and import paths now sanitize malformed data to prevent canvas rendering crashes:
- Nodes with missing `type` are filtered out (with a toast showing the count of skipped nodes on import).
- Nodes with missing or malformed `position` are defaulted to `[0, 0]` so they still render.
- Malformed connections (non-array buckets, non-array inner entries) are stripped to prevent `TypeError` crashes in `mapConnectionsByDestination` and canvas edge rendering.

## How to test

Paste the helper into the browser console while logged in, then run each test case. All should return `400` with a descriptive error.
To test frontend behavior, create the same workflows on master (requests would pass) and try to access them from the workflow list. All of them should be accessible and for most of them an error toast should be shown.

<details>
<summary>Helper</summary>

```js
const createWorkflow = async (body) => {
  const res = await fetch('/rest/workflows', {
    method: 'POST',
    headers: {
      'browser-id': localStorage.getItem('n8n-browserId'),
    },
    body: JSON.stringify(body),
  });
  const json = await res.json();
  console.log(res.status, json);
  return json;
};
```
</details>

<details>
<summary>1. Node missing <code>type</code></summary>

```js
await createWorkflow({
  name: 'Missing type',
  nodes: [
    { name: 'Start', position: [250, 300], parameters: {} },
  ],
  settings: {},
  connections: {},
});
```
</details>

<details>
<summary>2. Node missing <code>name</code></summary>

```js
await createWorkflow({
  name: 'Missing name',
  nodes: [
    { type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
  ],
  settings: {},
  connections: {},
});
```
</details>

<details>
<summary>3. Node missing <code>position</code></summary>

```js
await createWorkflow({
  name: 'Missing position',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', parameters: {} },
  ],
  settings: {},
  connections: {},
});
```
</details>

<details>
<summary>4 Position is not a 2-number tuple</summary>

```js
// Single number
await createWorkflow({
  name: 'Bad position - 1 number',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250], parameters: {} },
  ],
  settings: {},
  connections: {},
});

// Three numbers (worked before fine as well)
await createWorkflow({
  name: 'Bad position - 3 numbers',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300, 400], parameters: {} },
  ],
  settings: {},
  connections: {},
});

// Strings instead of numbers
await createWorkflow({
  name: 'Bad position - strings',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: ['a', 'b'], parameters: {} },
  ],
  settings: {},
  connections: {},
});
```
</details>

<details>
<summary>5. Duplicate node names</summary>

```js
await createWorkflow({
  name: 'Duplicate node names',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
    { name: 'Start', type: 'n8n-nodes-base.noOp', position: [450, 300], parameters: {} },
  ],
  settings: {},
  connections: {},
});
```
</details>

<details>
<summary>6 Malformed nested connections</summary>

```js
// Inner value is a string instead of the expected array-of-arrays
await createWorkflow({
  name: 'Malformed connections - string value',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
  ],
  settings: {},
  connections: {
    Start: { main: 'not-an-array' },
  },
});

// Inner bucket contains a raw object instead of array of connection entries
await createWorkflow({
  name: 'Malformed connections - object in bucket',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
  ],
  settings: {},
  connections: {
    Start: { main: [{ node: 'Foo', type: 'main', index: 0 }] },
  },
});

// Connection entry missing required fields
await createWorkflow({
  name: 'Malformed connections - incomplete entry',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
    { name: 'NoOp', type: 'n8n-nodes-base.noOp', position: [450, 300], parameters: {} },
  ],
  settings: {},
  connections: {
    Start: { main: [[{ node: 'NoOp' }]] },
  },
});
```
</details>

<details>
<summary>7. Connection source does not match any node</summary>

```js
await createWorkflow({
  name: 'Unknown connection source',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
    { name: 'NoOp', type: 'n8n-nodes-base.noOp', position: [450, 300], parameters: {} },
  ],
  settings: {},
  connections: {
    NonExistentNode: { main: [[{ node: 'NoOp', type: 'main', index: 0 }]] },
  },
});
```
</details>

<details>
<summary>8. Connection target does not match any node</summary>

```js
await createWorkflow({
  name: 'Unknown connection target',
  nodes: [
    { name: 'Start', type: 'n8n-nodes-base.manualTrigger', position: [250, 300], parameters: {} },
  ],
  settings: {},
  connections: {
    Start: { main: [[{ node: 'NonExistentTarget', type: 'main', index: 0 }]] },
  },
});
```
</details>

Closes [ADO-5134](https://linear.app/n8n/issue/ADO-5134/community-issue-workflow-redirects-to-new-workflow)

## Review / Merge checklist

- [x] Tests included. 


packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
```diff
@@ -237,14 +264,33 @@ export function useWorkflowInitialization(workflowState: WorkflowState) {
...
+			currentWorkflowDocumentStore.value = workflowDocumentStore;
+			currentNDVStore.value = useNDVStore(
+				createWorkflowDocumentId(
+					workflowDocumentStore.workflowId,
+					workflowDocumentStore.workflowVersion,
+				),
+			);
+		} catch (error) {
+			// Sending unexpected errors to Sentry
+			console.error('Failed to initialize workspace for workflow', {
```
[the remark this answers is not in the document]
dariacodes: We do have a centralized `Sentry.captureConsoleIntegration({ levels: ['error'] })`, that's why I added a comment about why I chose to "error" and not warning, but the comment could be clearer yeah

packages/cli/src/workflow-helpers.ts
```diff
@@ -121,6 +125,24 @@ export function resolveNodeWebhookIds(workflow: IWorkflowBase, nodeTypes: INodeT
 	}
 }
 
+export function validateWorkflowStructure(workflow: Pick<IWorkflowBase, 'nodes' | 'connections'>) {
```
CharlieKolb, line 128: [nit] Since the function throws `assertValidWorkflowStructure` might be a bit more correct? Not sure if `validate()` has as canonical a meaning

packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
```diff
@@ -150,7 +151,20 @@ export function useWorkflowInitialization(workflowState: WorkflowState) {
 					new Error(i18n.baseText('nodeView.couldntLoadWorkflow.invalidWorkflowObject')),
 					i18n.baseText('nodeView.couldntImportWorkflow'),
 				);
-				await router.replace({ name: VIEWS.NEW_WORKFLOW });
+				return true;
```
CharlieKolb, line 155: Is this a git diff hiccup or are we intentionally no longer replacing the route?

packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
```diff
@@ -2334,7 +2335,13 @@ export function useCanvasOperations() {
 	async function initializeWorkspace(data: IWorkflowDb) {
 		const { workflowDocumentStore: initializedDocumentStore } =
 			await workflowHelpers.initState(data);
-		data.nodes.forEach((node) => {
+
+		// Filter out nodes with missing type to prevent canvas rendering crashes
+		const validNodes = data.nodes
+			.filter((node) => !!node.type)
```
CharlieKolb, line 2341: What will happen to connections from these filtered out nodes? Is something removing these implicitly?

CharlieKolb (commented):
Looks good overall, some sanity checks before merge

packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
dariacodes: Yeah, good point, we currently don't do anything. Of course, after this fix it won't be possible to create such workflows anymore, so it's mostly about the malformed workflows that were created before this fix, but let me see if it's easy to fix this

CharlieKolb: I think I did something like that for sub-workflow extraction, let me check

CharlieKolb: Ah I only did it for a specific case, nevermind me https://github.com/n8n-io/n8n/blob/bdf06fa8ddb9b07c937531b3af22aac5882e1a5f/packages/frontend/editor-ui/src/app/composables/useWorkflowExtraction.ts#L154

packages/frontend/editor-ui/src/app/composables/useWorkflowInitialization.ts
dariacodes: I thought this would suppress the error, but I just checked, in this case it's fine (maybe because there is no console error), so I'll revert it back. Good catch!

packages/frontend/editor-ui/src/app/composables/useCanvasOperations.ts
dariacodes: Fixed in https://github.com/n8n-io/n8n/pull/29699/commits/ae833986cbb671f182a8fbb8fb9f34d063115a19

packages/cli/src/workflow-helpers.ts
dariacodes: It looks like in our codebase we're already using `validate*` for functions that throw errors, but I found better names - "parse" and "safe parse" (like zod)

CharlieKolb (approved):

PR #29700, opened 2026-05-04
fix(core): Defer Instance AI temporary workflow cleanup (no-changelog)

OlegIvaniv:
## Summary
This PR defers Instance AI temporary workflow cleanup while background tasks are still running for the thread.

Previously, an intermediate orchestrator run could finish while workflow-builder tasks were still verifying submitted workflows. The run-finish cleanup then archived workflows that were still marked as AI temporary before their builder tasks had completed and promoted them.

With this change, cleanup returns early when the thread still has running background tasks. A later run-finish after the tasks settle can still clean up workflows that remain marked as temporary.

https://linear.app/n8n/issue/INS-194

## Review / Merge checklist

- [x] Tests included.


aalises (dismissed):
Thanks! for these temporary shenanigans we also have to improve a bit the UX (making sure that the user knows which artifacts are drafts etc...)

aalises (approved):

PR #29701, opened 2026-05-04
fix(core): Improve AI chat file upload handling and error states

mutdmour:
## Summary

Fixes a UX loop in the AI chat where uploading a `.docx` or `.html` file produced an opaque error that then repeated on every follow-up message — even after the file was removed.

Two changes land together:

**1. Boundary validation closes the error loop.** Unsupported attachment MIME types are now rejected at the API boundary (`POST /instance-ai/chat/:threadId` and the chat-hub upload path) with a structured `BadRequestError` *before* anything is persisted. The chat-hub transaction rolls back the human-message row + binary, so nothing reaches LangChain memory and the next turn starts clean. Previously, a bad file rode through validation as `*/*`, was stored, entered conversation memory, and got replayed on every subsequent turn.

**2. Broader format support.** The `parse-file` tool now handles plain text, markdown, HTML, PDF, DOCX, and XLSX in addition to the existing CSV/TSV/JSON. PDFs and DOCX are extracted to text via `pdf-parse` and `mammoth`; XLSX is read by SheetJS and routed through the existing JSON pipeline so column normalization and type inference are reused. HTML is converted to markdown via linkedom + turndown.

A new `@n8n/instance-ai/parsers` subpath export keeps the cli boundary lean (no Mastra ESM contamination in jest).

### How to test

1. Open the AI chat. Attach a `.zip` (or any unsupported binary) → expect a clear toast/inline rejection without a fake AI reply.
2. Send a follow-up text-only message → expect a normal AI response with no echo of the prior failure.
3. Attach `.docx`, `.html`, `.pdf`, `.xlsx`, `.txt`, `.md` → expect the LLM to receive the extracted content and answer.
4. Image uploads (`.png`, `.jpg`) on a vision-capable model still work via multimodal.

https://linear.app/n8n/issue/AI-2417

## Review / Merge checklist

- [x] Tests included. (141 parser/tool tests, 93 cli chat-hub + instance-ai controller tests)


packages/@n8n/instance-ai/src/parsers/html-parser.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Extracts main content from an HTML/XHTML attachment.
+ *
+ * Pipeline:
+ *   linkedom (`parseHTML`) → strip script/style → turndown (markdown)
+ *
+ * We avoid Readability here to keep the type surface small (no DOM typings
+ * pulled in). The body content is converted directly to markdown.
+ */
+export async function extractHtmlContent(
```
Cadiac, line 35: it feels a bit off that we strip the content out of HTML files passed to the AI - I could see myself dropping a HTML file for the AI and asking it to do something with the actual file, like have my workflow return this html page...

But if we're just asking questions about the page contents this makes a lot of sense to parse the file like this 🤔

This same comment kind of applies to all of these file types here and I'm not sure what to do about it - it feels like we need two different ways to handle files here - one for models to look at them and one to store them somewhere and use them later on the sandbox as files, but that's obviously out of scope.

packages/@n8n/instance-ai/package.json
```diff
@@ -45,9 +55,11 @@
     "linkedom": "^0.18.9",
     "luxon": "catalog:",
     "csv-parse": "5.5.0",
+    "mammoth": "1.11.0",
     "nanoid": "catalog:",
     "p-limit": "^3.1.0",
     "pdf-parse": "^1.1.1",
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz",
```
Cadiac, line 64: this doesn't seem right?

packages/@n8n/instance-ai/package.json
```diff
@@ -45,9 +55,11 @@
     "linkedom": "^0.18.9",
     "luxon": "catalog:",
     "csv-parse": "5.5.0",
+    "mammoth": "1.11.0",
     "nanoid": "catalog:",
     "p-limit": "^3.1.0",
     "pdf-parse": "^1.1.1",
```
Cadiac: if https://www.npmjs.com/package/pdf-parse this is the package we're trying to use here it's current version is 2.45.0 🤔

packages/@n8n/instance-ai/package.json
```diff
@@ -45,9 +55,11 @@
     "linkedom": "^0.18.9",
     "luxon": "catalog:",
     "csv-parse": "5.5.0",
+    "mammoth": "1.11.0",
```
Cadiac: There's 1.12.0 of mammoth out already, should we also use `^1.11.0` ?

packages/@n8n/instance-ai/package.json
```diff
@@ -45,9 +55,11 @@
     "linkedom": "^0.18.9",
     "luxon": "catalog:",
     "csv-parse": "5.5.0",
```
Cadiac: csv-parse is also at 6.2.1 😬 https://www.npmjs.com/package/csv-parse

packages/cli/src/modules/chat-hub/chat-hub.constants.ts
```diff
@@ -682,30 +683,40 @@ const TEXT_APPLICATION_MIME_TYPES = [
...
+ * (.docx, .xlsx, .zip). It now resolves to the curated list of formats we
+ * have parsers for, plus media types declared via the other modalities.
+ */
 function resolveAllowedMimeTypes(modalities: ChatHubInputModality[]): string {
+	const mimeTypes = new Set<string>();
+
 	if (modalities.includes('file')) {
-		return '*/*';
+		// Parsers handle these (csv, json, pdf, docx, xlsx, html, etc.)
+		for (const mt of getParseableAttachmentMimeTypes()) mimeTypes.add(mt);
```
Cadiac: Why are we touching this logic / chat hub in general here?
I'm afraid of us introducing some regression here.

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -2139,14 +2139,20 @@ export class InstanceAiService {
 			});
 
 			const enrichedMessage = await this.buildMessageWithRunningTasks(threadId, message);
-			let nonStructuredAttachments: InstanceAiAttachment[] = [];
+			// Parseable formats (csv/tsv/json/xlsx/text/markdown/html/pdf/docx) go
+			// through parse-file; image/* is sent to the model as raw multimodal
+			// content. Anything else has been rejected upstream by the controller —
+			// but we filter defensively here so corrupt requests cannot pollute
+			// LLM memory.
+			let multimodalAttachments: InstanceAiAttachment[] = [];
```
Cadiac, line 2693: I suppose we can't send pdf's as multimodal attachments?

packages/cli/src/modules/chat-hub/chat-hub.constants.ts
mutdmour: it seems we are using common parsers in between for the parsing tool. I can revert this.

packages/@n8n/instance-ai/package.json
mutdmour: I triple checked this. turns out xlsx does not use npm. And that's how we are also using it elsewhere in n8n.

mutdmour: You can see it in nodes-base. https://docs.sheetjs.com/docs/getting-started/installation/standalone
the npm library is not managed anymore and out of date https://www.npmjs.com/package/xlsx

packages/@n8n/instance-ai/src/parsers/html-parser.ts
mutdmour: that's a good point. I could add a parameter that allows llm to control what it wants.

packages/@n8n/instance-ai/package.json
mutdmour: I think CC defaulted to using same version as in nodes-langhcain. I could update both

packages/@n8n/instance-ai/package.json
mutdmour: we are using mammoth for docx parsing

packages/@n8n/instance-ai/package.json
mutdmour: it's the same version as in ai-workflow-builder and nodes-base. I can upgrade this one.

packages/@n8n/instance-ai/package.json
mutdmour: or just upgrade this one.

packages/cli/src/modules/instance-ai/instance-ai.service.ts
mutdmour: depends on the LLM. Anthropic's Claude yes we can. But we need a registry of which models support this. So I outscopped it.

packages/cli/src/modules/chat-hub/chat-hub.constants.ts
mutdmour: Reverted from this PR — scoping changes to instance-ai only. The MIME-type tightening (curated parseable list vs `*/*`) will land as a follow-up so chat-hub regression risk stays out of this PR.

packages/@n8n/instance-ai/package.json
mutdmour: Bumped to `1.12.0` here and in nodes-langchain to keep the monorepo aligned.

packages/@n8n/instance-ai/package.json
mutdmour: Bumped to `6.2.1` here, in nodes-base, and in ai-workflow-builder.ee. v6 narrowed the `onRecord` generic so I split SpreadsheetFile’s CSV branch into header / no-header paths to satisfy the new typing — same runtime behavior.

packages/@n8n/instance-ai/package.json
mutdmour: Bumped to `2.4.5` (latest stable — `2.45.0` was likely a typo). v2 is a class-based API (`new PDFParse({data}).getText()`) so I rewrote `pdf-parser.ts` and `fetch-and-extract.ts` to match. Kept `nodes-langchain` on `1.1.1` because `@langchain/community`’s PDFLoader explicitly throws on v2 ("supports pdf-parse v1 only") — bumping it would break PDF document loaders.

packages/@n8n/instance-ai/src/parsers/html-parser.ts
mutdmour: Agreed there’s a future split between *file-as-content-for-the-model* (current `parse-file` behavior) and *file-as-blob-for-sandbox-storage*. The latter is a bigger design and out of scope here, so leaving the parsing pipeline as-is.

packages/@n8n/instance-ai/package.json
mutdmour: The CDN URL matches `nodes-base/package.json` exactly. SheetJS distributes via CDN because the npm `xlsx` package is unmaintained — that’s their official guidance.

packages/cli/src/modules/instance-ai/instance-ai.service.ts
mutdmour: Anthropic Claude does support PDFs as multimodal content, but other providers don’t. We don’t have a per-model capability registry yet, so I’m scoping multimodal to images for now — will revisit when the model-capability registry lands.

packages/@n8n/instance-ai/package.json
mutdmour: Update: keeping `nodes-base` on `5.5.0` to limit blast radius. Bumped only in `@n8n/instance-ai` and `@n8n/ai-workflow-builder.ee` where I needed it for this PR.

packages/@n8n/instance-ai/src/parsers/xlsx-parser.ts
Cadiac, line 1: Out of scope of this PR, but in the future we might want to have this return an array of sheets, not just the hard coded first sheet? Now we cound all the sheets into the decoded size limits but end up discarding the rest, and user could probably expect us to read data from all sheets

Cadiac (dismissed):
🚀

Cadiac:
Something to fix on the test linting issues still but otherwise LGTM!

Cadiac (approved):

mike12345567 (approved):

Issue #22700, opened 2025-12-03
AWS Bedrock Chat Model does not show "Assume Role" saved credential and documentation does not mention if this is supported or not


ghostpepper108:
### Bug Description

As the title mentions, the agent node when connected to AWS Bedrock Chat Model node does not show saved "Assume Role" credential and only expects access_id, secret_access_key  and session_token. This is not production worthy because these credentials would mostly be temporary. Apparently other AWS nodes like Lambda display the saved "Assume Role" credential.

The documentation [n8n AWS Bedrock](https://docs.n8n.io/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.lmchatawsbedrock/)  does not mention and only has link to Credentials page "You can find authentication information for this node [here](https://docs.n8n.io/integrations/builtin/credentials/aws/).

### To Reproduce

1. Create "AWS Assume Role" Credential
2. Add "AWS Bedrock Chat Model" node
3. Configure Credentials 

### Expected behavior

When configuring credentials in  "AWS Bedrock Chat Model" node, the saved "Assumed Role" does not show up

### Debug Info

NA

### Operating System

Debian GNU/Linux 12 (bookworm)

### n8n Version

1.122.4

### Node.js Version

18.20.4

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

Joffcom:
Hey @ghostpepper108 

Thank you for taking the time to share your suggestion! We’ve reviewed the details, and it seems this is more of a feature or enhancement request rather than a bug report.

As a heads-up, we use GitHub issues exclusively for tracking bugs. For feature requests or support questions, we recommend sharing them on [our forum](https://community.n8n.io), where the community and our team actively discuss and prioritize new ideas.

We’re closing this issue for now to keep our tracker focused, but we’d love to hear more about your idea on the forum! Your feedback helps us improve, and we truly appreciate it. 

Thanks again for contributing, and we hope to see you on the forum!

PR #29717, opened 2026-05-04
feat(awsBedrockChat Node): Add AWS Assume Role support to Bedrock chat and embeddings nodes

rsedletsky:
## Summary

Adds AWS Assume Role authentication support to the **AWS Bedrock Chat Model** and **Embeddings AWS Bedrock** nodes.

Previously both nodes only supported IAM key-based authentication. In environments where workloads run under an IAM role (e.g. EKS with IRSA, ECS task roles), users were forced to create long-lived access keys just to use Bedrock — a security anti-pattern.

This change:
- Adds an **Authentication** option to both nodes (`IAM User` / `Assume Role`)
- When **Assume Role** is selected, the node uses the `awsAssumeRole` credential, calls STS `AssumeRole`, and uses the resulting short-lived credentials for Bedrock API calls
- Extracts shared AWS credential resolution logic into `packages/@n8n/nodes-langchain/utils/aws.ts` to avoid duplication between the two nodes
- Tested in a production EKS cluster using pod-level IAM roles

Docs follow-up: https://github.com/n8n-io/n8n-docs/issues/4564

## Review / Merge checklist

- [x] Tests included.


rsedletsky:
This PR resolves the issue reported in #22700 — AWS Bedrock Chat Model and Embeddings nodes were unable to use the Assume Role credential.

The implementation was tested in production on an EKS cluster using IRSA (IAM Roles for Service Accounts), which is the most common use case for this feature.

PR #29724, opened 2026-05-04
fix: Improve browser-use connection stability (no-changelog)

BerniWittmann:
## Summary

Hardens the connection between the n8n browser-use stack (CDP relay ↔ MV3 extension) and Playwright so credential-setup runs don't hang on idle service workers, never-firing `load` events, or wedged debugger sessions.

Changes:

- **JSON ping/pong heartbeat between the CDP relay and the browser extension.** Receiving an application-level JSON message resets the MV3 service worker's idle-termination timer; protocol-level `ws.ping` is handled by the browser without invoking JS, so it never reset the SW idle timer. Protocol-level ping is kept alongside for backward compatibility with older extensions.
- **Default `page.goto` `waitUntil` from `'load'` to `'domcontentloaded'`** in the Playwright adapter and the public navigation tool. Many SaaS dashboards never fire `load` because of analytics/long-poll connections, which surfaced as spurious navigation timeouts. Not necessary for `agent-browser` as it does not have this option
- **Drop per-CDP-command timeout from 30s to 10s** and add a watchdog in the extension's `forwardCDPCommand` that force-detaches the debugger session on hang. The next call re-attaches fresh via `ensureAttached()` instead of inheriting a wedged session. The watchdog timer is cleared in a `finally` so it never fires for commands that already returned.

### How to test

1. Start an instance-ai credential-setup run that uses the gateway browser tools (`browser_*`) against a SaaS dashboard with heavy analytics (e.g. Notion, Linear, Slack).
2. Leave the relay idle for >30s and confirm the MV3 service worker stays alive (no `heartbeat_timeout` in relay logs).

https://linear.app/n8n/issue/NODE-4976

## Review / Merge checklist

- [ ] Tests included.

🤖 PR Summary generated by AI


packages/@n8n/instance-ai/src/tools/orchestration/browser-credential-setup.tool.ts
```diff
@@ -93,6 +93,30 @@ const browserCredentialSetupToolInputSchema = z.object({
...
 });
 
+/**
+ * Reset the browser session at the end of every orchestration run so the next
+ * run starts with fresh tab/debugger state. Stale state from a previous run
+ * (closed tabs still attached, debugger sessions in a wedged state) can cause
+ * CDP commands to hang on the next run. Best-effort — failures are logged
+ * and swallowed so they never mask the orchestration result.
+ */
+async function resetBrowserSession(
```
dlavrenuek: > Stale state from a previous run (closed tabs still attached, debugger sessions in a wedged state) can cause  CDP commands to hang on the next run.

I am not convinced that disconnecting the browser is the right approach. If something causes the CDP commands to hang then we should find the reason and fix it, but in general I did not experience hanging commands at least with the new implementation

packages/@n8n/instance-ai/src/tools/orchestration/browser-credential-setup.tool.ts
```diff
@@ -93,6 +93,30 @@ const browserCredentialSetupToolInputSchema = z.object({
...
+	context: OrchestrationContext,
+	toolSource: BrowserToolSource,
+): Promise<void> {
+	if (toolSource !== 'gateway' || !context.localMcpServer) return;
+	try {
+		await context.localMcpServer.callTool({
+			name: 'browser_disconnect',
+			arguments: { _confirmation: 'allowOnce' },
+		});
```
dlavrenuek: `_confirmation` property is actually removed from the arguments passed to computer-use, at least if the agent executes the tool call. It is only added on continue after suspense path. Did it work (not asking confirmation without previous "allow for session") when you tested it?

packages/@n8n/mcp-browser-extension/src/relayConnection.ts
```diff
@@ -418,6 +418,15 @@ export class RelayConnection {
 			return;
 		}
 
+		// Application-level ping from the relay. Receiving (and replying to)
+		// these JSON messages resets the MV3 service worker's idle-termination
+		// timer; protocol-level ws.ping is handled by the browser without
+		// invoking any JS handler, so it doesn't keep the SW alive.
+		if (message.method === 'ping' && message.id === undefined) {
+			this.sendMessage({ method: 'pong' });
```
dlavrenuek, line 428: If I see it correctly current implementation sends ping from adapter to extension and pong from extension to computer-use. The new implementation adds pong from relay to adapter, so the extension socket connection could break but the adapter would still be connected. I find it challenging to draw all the edge cases or understand which one is better. When the extension socket dies but adapter stays connected there should at least be some reconnect logic in the extension, but we don't have it (or do we?)

packages/@n8n/mcp-browser/src/tools/navigation.ts
```diff
@@ -8,7 +8,9 @@ import { createConnectedTool, extractDomain, pageIdField, withSnapshotEnvelope }
 const waitUntilField = z
 	.enum(['load', 'domcontentloaded', 'networkidle'])
 	.optional()
-	.describe('When to consider navigation done (default: "load")');
+	.describe(
+		'When to consider navigation done (default: "domcontentloaded"). Use "load" only when you need every subresource to settle.',
```
dlavrenuek: That looks promising, did you observe some performance improvements? Since we now have two adapters it would be worth adding the default here in the schema and not in the adapter itself. I did find that the "default xxx" in the description does not match what the adapter uses in some tools (example scroll or wait?)

packages/@n8n/mcp-browser/src/cdp-relay.ts
```diff
@@ -944,6 +944,17 @@ class ExtensionConnection {
 				return;
 			}
 
+			// Application-level ping: receiving this JSON message in the extension's
+			// MV3 service worker resets its 30s idle-termination timer. Protocol-level
```
dlavrenuek: what is `MV3`?

BerniWittmann: Chrome Manifest Version 3, will adjust that, so its understandable

packages/@n8n/instance-ai/src/tools/orchestration/browser-credential-setup.tool.ts
BerniWittmann: True, removed that part now

packages/@n8n/instance-ai/src/tools/orchestration/browser-credential-setup.tool.ts
BerniWittmann: True, removed now as well

packages/@n8n/mcp-browser-extension/src/relayConnection.ts
BerniWittmann: yeah the ping ponging in between is to keep the Service worker alive. 
but the cdp-relay.ts tracks the `lastPongAt` timestamp and then during the "Heartbeat" it detects this stale connection between extension and relay and also closes the connection to the adapter. 
Also tested this manually (via stopping the service worker mid run)
Bildschirmfoto 2026-05-08 um 10 56 21

packages/@n8n/mcp-browser/src/tools/navigation.ts
BerniWittmann: with playwrigth yes. Moved it up to the default, however, agent-browser doesnt utilize that parameter

dlavrenuek:
@BerniWittmann I could not reproduce any shutdowns or extension handup with the version we have on master - I did follow the instructions in the description and tried multiple notion dashboard. May be you have a concrete example url where the problem happens before the change and does not with the change?

BerniWittmann:
@dlavrenuek Yeah, probably so many changes around that. I will close this PR, if necessary we can still come back to this

PR #29731, opened 2026-05-04
fix(ai-builder): Resolve HitlTool variants to base node in get_node_types

RicardoE105:
## Summary

`get_node_types` failed for every HITL tool variant (`slackHitlTool`, `chatHitlTool`, `emailSendHitlTool`, etc.) because:

1. HITL tools are synthesized at runtime by `createHitlTools()` and have no static type-definition files.
2. The existing `*Tool`-suffix fallback stripped only 4 chars, producing nonsense intermediate names (`slackHitlTool` → `slackHitl`) that don't exist either.
3. The error message reported the *last attempted* (stripped) ID rather than the original request — so users saw an error mentioning a name they never typed.

This change adds a `HitlTool` (8-char) fallback in `findNodeDir` and in `getNodeFilePath`'s outer fallback so HITL variants resolve to their base node (`slackHitlTool` → `slack`). When all fallbacks fail, the original node ID is preserved in the error message.

### How to test

```bash
pushd packages/@n8n/ai-workflow-builder.ee
pnpm test code-builder-get.tool.test
popd
```

The new `tool variant fallback` describe block covers:
- HITL → flat-structure base (`emailSendHitlTool` → `emailSend`)
- HITL → langchain base (`@n8n/n8n-nodes-langchain.chatHitlTool` → `chat`)
- HITL → split-structure base with discriminators (`slackHitlTool` + resource/operation)
- Original ID preserved in error for unresolved `HitlTool` and plain `Tool` variants

End-to-end smoke test against the live MCP server confirmed all four scenarios resolve correctly through `mcp__n8n-mcp__get_node_types`.

https://linear.app/n8n/issue/ADO-5170

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-get.tool.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-get.tool.ts
@@ -62,26 +62,34 @@ function getGeneratedNodesPaths(nodeDefinitionDirs?: string[]): string[] {
 }
 
 /**
+ * Synthetic suffixes appended to base node names when tool variants are generated at runtime.
+ * Listed most-specific-first so "slackHitlTool" strips to "slack" rather than "slackHitl".
+ *  - "HitlTool":  appended by createHitlTools() in @n8n/cli/src/tool-generation/hitl-tools.ts
+ *  - "Tool":      appended by createAiTools() in @n8n/cli/src/tool-generation/ai-tools.ts
+ */
+const TOOL_VARIANT_SUFFIXES = ['HitlTool', 'Tool'] as const;
+
+/**
  * Find the first nodes path that contains the given node directory.
  * Returns { nodesPath, nodeDir } or null if not found in any dir.
+ *
+ * For tool variants (no on-disk type files of their own), falls back to the base node's
+ * directory by stripping known synthetic suffixes.
  */
 function findNodeDir(
 	parsed: { packageName: string; nodeName: string },
 	nodesPaths: string[],
 ): { nodesPath: string; nodeDir: string } | null {
-	for (const nodesPath of nodesPaths) {
-		const nodeDir = join(nodesPath, parsed.packageName, parsed.nodeName);
-		if (existsSync(nodeDir)) {
-			return { nodesPath, nodeDir };
+	const candidateNames: string[] = [parsed.nodeName];
+	for (const suffix of TOOL_VARIANT_SUFFIXES) {
+		if (parsed.nodeName.endsWith(suffix)) {
+			candidateNames.push(parsed.nodeName.slice(0, -suffix.length));
 		}
 	}
 
-	// Tool variant fallback: e.g. "httpRequestTool" -> "httpRequest"
-	// Tool variants share type definitions with their base node
-	if (parsed.nodeName.endsWith('Tool')) {
-		const baseName = parsed.nodeName.slice(0, -4);
+	for (const candidate of candidateNames) {
 		for (const nodesPath of nodesPaths) {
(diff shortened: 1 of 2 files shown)


packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-get.tool.ts
```diff
@@ -64,24 +64,22 @@ function getGeneratedNodesPaths(nodeDefinitionDirs?: string[]): string[] {
...
 ): { nodesPath: string; nodeDir: string } | null {
-	for (const nodesPath of nodesPaths) {
-		const nodeDir = join(nodesPath, parsed.packageName, parsed.nodeName);
-		if (existsSync(nodeDir)) {
-			return { nodesPath, nodeDir };
-		}
-	}
+	const candidates = [parsed.nodeName];
+	const baseName = parsed.nodeName.replace(/(?:HitlTool|Tool)$/, '');
+	if (baseName !== parsed.nodeName) candidates.push(baseName);
```
MiloradFilipovic, line 78: Nit: I know this is unrealistic at the moment, but if `parsed.nodeName` is exactly `Tool` or `HitlTool` it will produce empty strings and path will resolve to root directory. Should we check for that?

```suggestion
	 if (baseName && baseName !== parsed.nodeName) candidates.push(baseName);
```

packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-get.tool.ts
```diff
@@ -64,24 +64,22 @@ function getGeneratedNodesPaths(nodeDefinitionDirs?: string[]): string[] {
 /**
  * Find the first nodes path that contains the given node directory.
  * Returns { nodesPath, nodeDir } or null if not found in any dir.
+ *
+ * Tool variants (e.g. "slackHitlTool", "httpRequestTool") have no on-disk type files;
+ * fall back to the base node by stripping the suffix. The regex tries "HitlTool" before
+ * "Tool" so "slackHitlTool" resolves to "slack", not "slackHitl".
```
MiloradFilipovic, line 70: Nit: Should we add a quick comment to make sure we don't break it when we introduce new sufixes in the future? Something like:
```ts
// Order matters: longer suffixes first
```

packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/test/code-builder-get.tool.test.ts
```diff
@@ -355,6 +355,101 @@ describe('CodeBuilderGetTool', () => {
...
+						operation: 'post',
+					},
+				],
+			});
+
+			expect(result).toContain('SlackV24MessagePostConfig');
+			expect(result).not.toContain('not found');
+		});
+
+		it.each([
```
MiloradFilipovic, line 440: Can we check that old bad names are not in the tripped result?

```suggestion
it.each([
	[
		'n8n-nodes-base.unknownThingHitlTool',
		['n8n-nodes-base.unknownThingHitl', 'n8n-nodes-base.unknownThing'],
	],
	['n8n-nodes-base.unknownThingTool', ['n8n-nodes-base.unknownThing']],
])('should report the original node ID in the error: %s', async (nodeId, strippedNames) => {
	const tool = createCodeBuilderGetTool({ nodeDefinitionDirs: [tempDir] });
	const result = await tool.invoke({ nodeIds: [nodeId] });
	expect(result).toContain(`'${nodeId}'`);
	for (const strippedName of strippedNames) {
		expect(result).not.toContain(`'${strippedName}'`);
	}
});
```

MiloradFilipovic (approved):
Looks good, left a few nits and test improvement suggestions

PR #29733, opened 2026-05-04
feat(editor): Add envFeatureFlag and copyButton property options

guillaumejacquart:
## Summary

Adds two declarative knobs on `INodeProperties` so credential forms can hide fields behind instance-level env feature flags and surface readonly, click-to-copy values without bespoke Vue plumbing:

- **`envFeatureFlag?: Uppercase<string>`** on `INodeProperties` — hides the field on credential forms unless the named `N8N_ENV_FEAT_*` flag is truthy. Honoured by `CredentialInputs.vue` via `useEnvFeatureFlag`. The value is the suffix only (e.g. `'OAUTH2_JWE'`), matching the existing `<EnvFeatureFlag name=\"...\">` component convention.
- **`typeOptions.copyButton?: boolean`** on string properties — renders the value as a `CopyInput` (label + monospace value + click-to-copy with toast) instead of the standard text input + Fixed/Expression toolbar. Useful for system-generated values the user needs to paste somewhere external.

Wired both into `OAuth2Api`: the `jweEnabled` toggle gates on `OAUTH2_JWE` and the `jwksUri` field renders as a copy-button string. `FrontendService.overwriteCredentialsProperties` injects the per-instance JWKS URI into the field's `default` at runtime.

**How to test:** with `N8N_ENV_FEAT_OAUTH2_JWE=true pnpm start`, open an OAuth2 credential — the \"Encrypted Tokens (JWE)\" toggle is visible; toggling it on reveals a JWKS URI field with a click-to-copy affordance and the live instance URL. Without the flag, neither field appears.

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI

### DIFF
--- a/packages/cli/src/services/frontend.service.ts
+++ b/packages/cli/src/services/frontend.service.ts
@@ -695,11 +695,13 @@ export class FrontendService {
 				credential.__skipManagedCreation = true;
 			}
 
-			// Inject the per-instance JWKS URI as the default of `jwksUri` for
-			// `oAuth2Api` and any credential that extends it. The field is hidden
-			// behind the `jweEnabled` toggle (and the toggle itself is gated by
-			// the OAuth2 JWE feature flag on the frontend), so the value only
-			// surfaces when the feature is in use.
+			// Inject the per-instance JWKS URI as the default of `jwksUri` on
+			// `oAuth2Api` and any credential that extends it. The field stays
+			// hidden by the `jweEnabled` toggle's `displayOptions`, which only
+			// resolves on credentials that explicitly re-declare `jweEnabled`
+			// (it carries `doNotInherit: true`). Net effect: bare `oAuth2Api`
+			// surfaces the JWE flow today, future JWE-aware extending
+			// credentials get the URL injected automatically when they opt in.
 			const isOAuth2Credential =
 				credential.name === 'oAuth2Api' ||
 				this.credentialTypes.getParentTypes(credential.name).includes('oAuth2Api');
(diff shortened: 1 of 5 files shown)


packages/cli/src/services/frontend.service.ts
```diff
@@ -695,26 +695,21 @@ export class FrontendService {
...
+			// `doNotInherit: true` on both `jweEnabled` and `jwksUri` so that
+			// extending credentials don't silently inherit half a dependency
+			// pair (which would crash `getParameterResolveOrder`); custom
+			// JWE-aware OAuth2 extensions can re-declare both fields together.
+			const isOAuth2Credential =
+				credential.name === 'oAuth2Api' ||
+				this.credentialTypes.getParentTypes(credential.name).includes('oAuth2Api');
+			if (isOAuth2Credential && credential.properties) {
+				const jwksUri = `${this.urlService.getInstanceBaseUrl()}/${this.globalConfig.endpoints.rest}/.well-known/jwks.json`;
+				credential.properties = credential.properties.map((property) =>
```
afitzek, line 710: Do we store this property in the database next to the credential if we send it as property to the UI like this?

afitzek (commented):
One small comment, I'm not sure on it, maybe you know from the top of your head?

packages/cli/src/services/frontend.service.ts
guillaumejacquart: I don't think so. I did a test by decrypting a credentials saved with jwe enabled, and could not see the jwksUri field inside. I think default values are not automatically added to credential data. And with the copyinput component, we never update `credentialData`

afitzek (approved):

PR #29736, opened 2026-05-04
fix(core): Throw on bare OutputSelector passed to .add()/.to()

RicardoE105:
## Summary

When the workflow SDK received a bare `node.output(n)` (an `OutputSelector`) in `.add()` or `.to()`, it silently dropped the connection. The `OutputSelector` was cast to `NodeInstance`, `_currentNode` was left unset, and the chained `.to(target)` became a no-op because of the `if (this._currentNode)` guard. `validate_workflow` only complained when the target had zero incoming edges, so any target with another incoming wire hid the drop entirely.

This is the LLM-facing failure mode described in the ticket: models write `.add(node.output(n)).to(target)` instead of the correct `.add(node.output(n).to(target))` and the misuse fails silently.

The fix detects `OutputSelector` at the entry of `.add()` and `.to()` (and inside their array paths) and throws a `TypeError` that names the source node, the output index, and the correct shape.

### How to test

```ts
import { workflow, node, trigger } from '@n8n/workflow-sdk';

const t = trigger({ type: 'n8n-nodes-base.webhookTrigger', version: 1, config: {} });
const ifNode = node({ type: 'n8n-nodes-base.if', version: 2, config: { name: 'Branch' } });
const target = node({ type: 'n8n-nodes-base.set', version: 3, config: {} });

// Throws with a clear message instead of silently dropping the connection
workflow('id', 'wf').add(t).add(ifNode.output(0)).to(target);
//                              ^^^^^^^^^^^^^^^^^ TypeError now points at .to() chained the wrong way
```

The documented form keeps working:

```ts
workflow('id', 'wf').add(t).add(ifNode.output(0).to(target));
```

Tests cover:
- `OutputSelector` passed to `.add()` (top-level and inside an array)
- `OutputSelector` passed to `.to()`
- The correct `node.output(n).to(target)` form still produces the expected connection

https://linear.app/n8n/issue/ADO-5161

## Review / Merge checklist

- [x] Tests included.


packages/@n8n/workflow-sdk/src/workflow-builder.ts
```diff
@@ -1162,6 +1173,18 @@ class WorkflowBuilderImpl implements WorkflowBuilder {
 	}
 }
 
+function assertNotOutputSelector(value: unknown, method: 'add' | 'to'): void {
+	if (!isOutputSelector(value)) return;
+	const sourceName = value.node.name;
+	throw new TypeError(
```
MiloradFilipovic, line 1179: In [this PR](https://github.com/n8n-io/n8n/pull/28416/changes), we introduced `WorkflowCodeParseError` and we are adding a hint to use SDK reference if it's thrown. Will this be useful here also?

MiloradFilipovic (approved):
Nice 🙌 Left a comment for a potential improvement so take a look if it makes sense

PR #29739, opened 2026-05-04
feat(core): Add update_partial_workflow MCP tool

RicardoE105:
## Summary

Adds a new MCP tool — `update_partial_workflow` — that updates a workflow by applying a small list of named operations directly to the stored JSON, bypassing the full SDK code → JSON parse + LLM-validate pass that `update_workflow` runs on every edit.

The existing `update_workflow` tool requires the agent to re-emit the full SDK code on every edit, which scales linearly with workflow size. Iterating on a 30-node workflow to tweak one parameter still costs ~5–10 KB of output tokens per call. The partial-update path keeps per-edit payloads small and constant in size — the same parameter tweak ships ~150 chars regardless of workflow size.

Linear: https://linear.app/n8n/issue/ADO-5171

### Operations (v1)

The tool takes an `operations: []` array. Each op is one of:

```
updateNodeParameters   — change parameters on a node (deep-merge or replace)
addNode                — insert a new node
removeNode             — delete a node + all its connections
renameNode             — rename + rewrite connection references
addConnection          — wire two nodes (defaults: type=main, indices=0,0)
removeConnection       — unwire two nodes
setNodeCredential      — attach a credential to a node
setNodePosition        — move a node on the canvas
setNodeDisabled        — toggle a node
setWorkflowMetadata    — update workflow name and/or description
```

The whole batch is **atomic**: if any op fails (missing node, duplicate name, missing connection) the workflow is left unchanged and a `Operation N failed: <reason>` error is returned with the offending op index.

### Why a new tool, not a replacement

The two tools have different signatures and serve different jobs. `update_workflow` (full rewrite) stays for major restructuring; `update_partial_workflow` is preferred for parameter tweaks, single-node insertions, rewiring, rename, and metadata edits. n8n-mcp ships both; their telemetry shows 89.2% of update calls go through the partial path, 10.8% through the full one. The MCP instructions guide agents to prefer partial and fall back to full when the change is too large to express compactly.

### Token economy (measured on the live test in this PR)

3 typical iterative edits, replayed via both tools on the same workflow:

| Edit | `update_workflow` | `update_partial_workflow` | Reduction |
|---|---:|---:|---|
| Set parameters on one node | 610 chars / ~153 tok | 163 chars / ~41 tok | **3.7× smaller** (73.3%) |
| Add a node + connect it | 797 chars / ~200 tok | 232 chars / ~58 tok | **3.4× smaller** (70.9%) |
| Rename a node + set description | 1,129 chars / ~283 tok | 313 chars / ~79 tok | **3.6× smaller** (72.3%) |
| **Total** | **2,536 chars** | **708 chars** | **3.6× smaller (72.1%)** |

This is on a tiny 2–3 node workflow — the smallest possible sample. The gap widens with workflow size: per the n8n-mcp competitive analysis (May 2026), the same edits at 15 nodes show ~15× reduction and at 30 nodes ~22×, with a 41× median across 30K real production mutations.

#### How to replicate the numbers

The benchmark is pure JSON math — no server or auth needed. Save the script below to a file and run with Node ≥ 18 (uses `structuredClone`-equivalent JSON only, no special deps):

```bash
node ./ado-5171-token-bench.mjs
```

Expected output:

```
ADO-5171 — token reduction per edit (live test replication)

  Test 1 — set parameters on one node                    610 chars /     153 tok       163 chars /      41 tok   3.7× smaller (73.3% reduction)
  Test 2 — add a node + connect it                       797 chars /     200 tok       232 chars /      58 tok   3.4× smaller (70.9% reduction)
  Test 3 — rename a node + set description              1129 chars /     283 tok       313 chars /      79 tok   3.6× smaller (72.3% reduction)

Across the 3 edits combined:
  update_workflow         total: 2536 chars / ~634 tok (estimate)
  update_partial_workflow total: 708 chars / ~177 tok (estimate)
  Reduction: 3.6× smaller, 72.1% fewer chars
```

The script defines the same 2-node demo workflow we tested live, evolves it through the 3 edits two ways (full SDK code rewrite vs. partial ops), and prints char/token counts for each. Anyone reviewing the PR can run it and get identical numbers — no setup required.

<details>
<summary>📋 <code>ado-5171-token-bench.mjs</code></summary>

```javascript
// Replicable token-reduction benchmark for ADO-5171.
// Re-runs the 3 edits from the live MCP test, computing payload sizes for both
// the existing `update_workflow` (full SDK rewrite) and the new
// `update_partial_workflow` (small ops). Pure JSON math — no server required.

// ---------- Workflow state, evolved through the same edits as the live test ----------

const wf0 = `import { workflow, node, trigger } from '@n8n/workflow-sdk';

const startTrigger = trigger({
  type: 'n8n-nodes-base.manualTrigger',
  version: 1,
  config: { name: 'Manual Trigger', position: [240, 300] },
  output: [{}]
});

const editFields = node({
  type: 'n8n-nodes-base.set',
  version: 3.4,
  config: { name: 'Edit Fields', parameters: {}, position: [540, 300] },
  output: [{}]
});

export default workflow('ado-5171-demo', 'ADO-5171 Partial Update Demo')
  .add(startTrigger)
  .to(editFields);
`;

// After Test 1 (set parameters on Edit Fields):
const wf1 = wf0.replace(
  "parameters: {}, position: [540, 300]",
  "parameters: { mode: 'manual', includeOtherFields: false }, position: [540, 300]",
);

// After Test 2 (add NoOp + connect):
const wf2 = `import { workflow, node, trigger } from '@n8n/workflow-sdk';

const startTrigger = trigger({
  type: 'n8n-nodes-base.manualTrigger',
  version: 1,
  config: { name: 'Manual Trigger', position: [240, 300] },
  output: [{}]
});

const editFields = node({
  type: 'n8n-nodes-base.set',
  version: 3.4,
  config: { name: 'Edit Fields', parameters: { mode: 'manual', includeOtherFields: false }, position: [540, 300] },
  output: [{}]
});

const afterEdit = node({
  type: 'n8n-nodes-base.noOp',
  version: 1,
  config: { name: 'After Edit', parameters: {}, position: [840, 300] },
  output: [{}]
});

export default workflow('ado-5171-demo', 'ADO-5171 Partial Update Demo')
  .add(startTrigger)
  .to(editFields)
  .to(afterEdit);
`;

// After Test 3 (rename Edit Fields -> Process, set description):
const wf3 = wf2
  .replace(/Edit Fields/g, "Process")
  .replace(/editFields/g, "process_")
  .replace(
    "'ADO-5171 Partial Update Demo')",
    "'ADO-5171 Partial Update Demo', { description: 'Demo workflow created via update_partial_workflow tool — exercises updateNodeParameters, addNode, addConnection, renameNode, and atomic batch behavior.' })",
  );

// ---------- Payloads sent for each tool ----------

const fullWorkflowCall = (workflowId, code, extras = {}) =>
  JSON.stringify({ workflowId, code, ...extras });

const partialWorkflowCall = (workflowId, operations) =>
  JSON.stringify({ workflowId, operations });

const WF_ID = "5qNG3TO1vOdamBlH";

const tests = [
  {
    label: "Test 1 — set parameters on one node",
    full: fullWorkflowCall(WF_ID, wf1),
    partial: partialWorkflowCall(WF_ID, [
      {
        type: "updateNodeParameters",
        nodeName: "Edit Fields",
        parameters: { mode: "manual", includeOtherFields: false },
      },
    ]),
  },
  {
    label: "Test 2 — add a node + connect it",
    full: fullWorkflowCall(WF_ID, wf2),
    partial: partialWorkflowCall(WF_ID, [
      {
        type: "addNode",
        node: {
          name: "After Edit",
          type: "n8n-nodes-base.noOp",
          typeVersion: 1,
          position: [840, 300],
        },
      },
      { type: "addConnection", source: "Edit Fields", target: "After Edit" },
    ]),
  },
  {
    label: "Test 3 — rename a node + set description",
    full: fullWorkflowCall(WF_ID, wf3, {
      description:
        "Demo workflow created via update_partial_workflow tool — exercises updateNodeParameters, addNode, addConnection, renameNode, and atomic batch behavior.",
    }),
    partial: partialWorkflowCall(WF_ID, [
      { type: "renameNode", oldName: "Edit Fields", newName: "Process" },
      {
        type: "setWorkflowMetadata",
        description:
          "Demo workflow created via update_partial_workflow tool — exercises updateNodeParameters, addNode, addConnection, renameNode, and atomic batch behavior.",
      },
    ]),
  },
];

// ---------- Report ----------

const tok = (s) => Math.ceil(s.length / 4); // standard char/4 estimate
const fmt = (n) => n.toString().padStart(7);
const pct = (a, b) => `${(((a - b) / a) * 100).toFixed(1)}%`;
const ratio = (a, b) => `${(a / b).toFixed(1)}×`;

console.log("\nADO-5171 — token reduction per edit (live test replication)\n");

let totalFullChars = 0;
let totalPartialChars = 0;

for (const { label, full, partial } of tests) {
  const fc = full.length;
  const pc = partial.length;
  totalFullChars += fc;
  totalPartialChars += pc;
  console.log(
    `  ${label.padEnd(50)} ${fmt(fc)} chars / ${fmt(tok(full))} tok       ${fmt(pc)} chars / ${fmt(tok(partial))} tok       ${ratio(fc, pc)} smaller (${pct(fc, pc)} reduction)`,
  );
}

console.log(`\nAcross the 3 edits combined:`);
console.log(`  update_workflow         total: ${totalFullChars} chars / ~${tok({ length: totalFullChars })} tok (estimate)`);
console.log(`  update_partial_workflow total: ${totalPartialChars} chars / ~${tok({ length: totalPartialChars })} tok (estimate)`);
console.log(`  Reduction: ${ratio(totalFullChars, totalPartialChars)} smaller, ${pct(totalFullChars, totalPartialChars)} fewer chars\n`);

console.log("Notes:");
console.log("  • Token estimate uses chars / 4 (standard rough heuristic).");
console.log("  • Sizes are JSON request bodies — what the agent emits as output tokens.");
console.log("  • update_workflow re-sends the full SDK code on every edit, so its size");
console.log("    grows linearly with workflow size; partial-update size stays roughly flat.");
console.log("  • For real-world workflows (15-30 nodes with rich parameters), the gap grows:");
console.log("    n8n-mcp's competitive analysis measured 6.5×-22× at 4-30 nodes,");
console.log("    and a 41× median across 30K real production mutations.");
```

</details>

### Scope

The new tool is registered alongside `update_workflow` inside `registerBuilderTools()`, behind the same `N8N_MCP_BUILDER_ENABLED` flag — instances that don't enable the builder MCP surface are unchanged.

### Implementation notes

- `workflow-operations.ts` is the pure ops engine — discriminated zod union, atomic `applyOperations(workflow, ops)`. Easy to unit-test, no I/O.
- The tool handler reuses existing infrastructure: `getMcpWorkflow` for permission/MCP-availability, `collaborationService.ensureWorkflowEditable` for write-lock, `resolveNodeWebhookIds` for added webhook nodes, `workflowService.update` for save, `collaborationService.broadcastWorkflowUpdate` for cross-editor notification.
- Credential auto-assignment runs **only on nodes added in the same batch** (not on untouched existing nodes), so partial updates never silently mutate credential state on nodes the agent didn't touch.

## Test plan

- [x] Unit tests: 28 op-level (`workflow-operations.test.ts`) + 14 handler-level (`update-partial-workflow.tool.test.ts`) — all passing
- [x] Sibling tests still passing (`update-workflow.tool.test.ts`, `mcp.service.test.ts`, `credentials-auto-assign.test.ts`)
- [x] `pnpm typecheck` clean on `packages/cli` and `packages/@n8n/ai-workflow-builder.ee`
- [x] `pnpm lint` clean on changed files
- [x] **End-to-end live test** against a local n8n instance (`N8N_MCP_BUILDER_ENABLED=true`): created a workflow via `create_workflow_from_code`, then exercised `updateNodeParameters`, `addNode`+`addConnection`, `renameNode`+`setWorkflowMetadata`, plus a deliberate atomic-failure case. State verified via `get_workflow_details`. Atomic rollback confirmed (post-failure `versionId` and node positions unchanged).


packages/cli/src/modules/mcp/tools/workflow-builder/update-partial-workflow.tool.ts
```diff
@@ -0,0 +1,235 @@
...
+import type { WorkflowFinderService } from '@/workflows/workflow-finder.service';
+import type { WorkflowService } from '@/workflows/workflow.service';
+
+import { getMcpWorkflow } from '../workflow-validation.utils';
+
+const inputSchema = {
+	workflowId: z.string().describe('The ID of the workflow to update.'),
+	operations: z
+		.array(partialUpdateOperationSchema)
+		.min(1)
```
MiloradFilipovic: Should we also add `max` limit to `operations`?

packages/cli/src/modules/mcp/tools/workflow-builder/update-partial-workflow.tool.ts
```diff
@@ -0,0 +1,235 @@
...
+		.describe(
+			'Graph and JSON validation warnings on the resulting workflow. Use these to self-correct on the next call.',
+		),
+	note: z.string().optional(),
+} satisfies z.ZodRawShape;
+
+/**
+ * MCP tool that updates a workflow by applying a small list of named operations
+ * (addNode, removeNode, updateNodeParameters, addConnection, …) directly to the
+ * stored JSON. Avoids the full code → JSON parse pass of `update_workflow`,
```
MiloradFilipovic: If I saw it correctly, with validation step, we are still doing the `code -> JSON` parsing, so this comment is not correct.

packages/cli/src/modules/mcp/tools/workflow-builder/update-partial-workflow.tool.ts
```diff
@@ -0,0 +1,235 @@
...
+	urlService: UrlService,
+	telemetry: Telemetry,
+	nodeTypes: NodeTypes,
+	credentialsService: CredentialsService,
+	sharedWorkflowRepository: SharedWorkflowRepository,
+	collaborationService: CollaborationService,
+): ToolDefinition<typeof inputSchema> => ({
+	name: MCP_UPDATE_PARTIAL_WORKFLOW_TOOL.toolName,
+	config: {
+		description:
```
MiloradFilipovic: I don't think we need to repeat supported operation types here since they are in the input schema already

packages/cli/src/modules/mcp/mcp.service.ts
```diff
@@ -391,6 +392,23 @@ export class McpService {
 		);
 		server.registerTool(updateTool.name, updateTool.config, updateTool.handler);
```
MiloradFilipovic, line 404: I think we should just go and remove the old `update` tool. It just introduces unnecessary context noise and introduces a chance of miss-using tools.

packages/cli/src/modules/mcp/tools/workflow-builder/workflow-operations.ts
```diff
@@ -0,0 +1,433 @@
...
+	}
+};
+
+const ensureOutputSlot = (
+	connections: IConnections,
+	source: string,
+	connectionType: string,
+	sourceIndex: number,
+): IConnection[] => {
+	const byType = (connections[source] ??= {});
```
MiloradFilipovic, line 302: I think we should limit what can be assigned here as `source`, using something like `connections['__proto__']` will turn `byType` into `Object.prototype` and then the next line can write a property onto it (pollute the prototype).
We should reject unsafe names (like `__proto__` and `constructor`)

packages/cli/src/modules/mcp/tools/workflow-builder/workflow-operations.ts
```diff
@@ -0,0 +1,433 @@
...
+	target: Record<string, unknown>,
+	source: Record<string, unknown>,
+): Record<string, unknown> => {
+	const result: Record<string, unknown> = { ...target };
+	for (const [key, value] of Object.entries(source)) {
+		const existing = result[key];
+		if (isPlainObject(existing) && isPlainObject(value)) {
+			result[key] = deepMerge(existing, value);
+		} else {
+			result[key] = value;
```
MiloradFilipovic: Similar to the issue that I posted below, we should sanitize what's passed when assigning object keys like this:
```json
{
  "type": "updateNodeParameters",
  "nodeName": "Some Node",
  "parameters": {
    "__proto__": {
      "polluted": true
    }
  }
}
```
Can produce a merged `node.parameters` object whose prototype is attacker-controlled.
We should:
1. Reject unsafe values (see comment below)
2. Use `hasOwn` instead of directly checking properties by key to avoid accessing the prototype chain:
```ts
const existing = Object.hasOwn(result, key) ? result[key] : undefined;
```

MiloradFilipovic: Now looking at this, we seem to have `isSafeObjectProperty` method that we can reuse here (and belwo)

MiloradFilipovic (commented):
Looks really good. I left some comments about the potential improvements and a couple of security risks when directly dealing with objects like this so take a look at what makes sense.

Also, I was looking at some numbers and token savings are only 20% compared to the old tool. I was expecting more since the time saved is dramatic. Does it make sense to try using path-based parameter updates instead of sending full `parameters` object? Something like this:
```json
{
  "type": "setNodeParameter",
  "nodeName": "Output Parser",
  "path": "/jsonSchema",
  "value": "{ ... schema ... }"
}
// For deeply nested params:
{
  "type": "setNodeParameter",
  "nodeName": "AI Agent",
  "path": "/options/systemMessage",
  "value": "You are a helpful assistant..."
}
```

packages/cli/src/modules/mcp/mcp.service.ts
RicardoE105: would not mess up out telemetry stuff? should we rename the partial one to this so it does not mess up telemetry?

RicardoE105:
@MiloradFilipovic this one will address in another PR

https://github.com/n8n-io/n8n/pull/29739#pullrequestreview-4251483190

packages/cli/src/modules/mcp/mcp.service.ts
MiloradFilipovic: Yes, let's rename it for consistency but we can also always join two events in the analysis. But let's completely substitute the old one.

packages/cli/src/modules/mcp/tools/workflow-builder/workflow-operations.ts
```diff
@@ -0,0 +1,460 @@
...
+
+			case 'setNodeCredential': {
+				const node = nodeByName.get(op.nodeName);
+				if (!node) return fail(i, `node '${op.nodeName}' not found`);
+				if (!isSafeObjectProperty(op.credentialKey)) {
+					return fail(i, `credential key '${op.credentialKey}' is not allowed`);
+				}
+				node.credentials = {
+					...(node.credentials ?? {}),
+					[op.credentialKey]: { id: op.credentialId, name: op.credentialName },
```
MiloradFilipovic, line 493: If I see it correctly, this assigns whatever credential id comes from the caller. Should we check if the credential exists and is usable in the node?

MiloradFilipovic (commented):
Thanks for addressing the feedback 🙌 I left one more that slipped through the cracks last time so take a look if it makes sense.

MiloradFilipovic (dismissed):
Looks good. Thanks for addressing the feedback!

MiloradFilipovic (approved):

PR #29740, opened 2026-05-04
feat(Crypto Node): Add HMAC-SHA1 support

maxabaumgarten:
## Summary

Adds `SHA1` as a selectable algorithm under the Crypto node's **HMAC** *and* **Hash** actions (both V1 and V2). A number of existing signature schemes and integrity checks still require SHA1 for interoperability with legacy APIs (e.g. legacy webhook signature verification, object-ID / ETag-style integrity checks, SAML/XML-DSig, SRI for older CDN assets). The underlying `createHmac` / `createHash` calls already support it, so this is a UI-visible enum addition only — no changes to credentials, execution logic, or the Sign action's algorithm list.

Why both actions: MD5 (weaker than SHA1 against collisions) is already in both dropdowns, so omitting SHA1 from Hash while permitting it under HMAC would be inconsistent rather than security-preserving. HMAC's security does not depend on collision resistance of the underlying hash (NIST SP 800-107 continues to permit HMAC-SHA1 for interoperability with legacy APIs). SHA256 and above remain the recommended defaults — V2 defaults are unchanged (`SHA256`), V1 defaults are unchanged (`MD5`).

### Changes

- `packages/nodes-base/nodes/Crypto/v1/CryptoV1.node.ts` — add `SHA1` to the Hash `Type` and HMAC `Type` option lists.
- `packages/nodes-base/nodes/Crypto/v2/CryptoV2.node.ts` — same, in V2.
- `packages/nodes-base/nodes/Crypto/v2/test/Crypto.node.test.ts` — unit tests for Hash-SHA1 (hex), HMAC-SHA1 over a string value (hex), and HMAC-SHA1 over in-memory binary data (hex).
- `packages/nodes-base/nodes/Crypto/v1/test/CryptoTest.workflow.json` — workflow-harness fixtures for `Crypto Hash into SHA1`, `Crypto Hmac data with SHA1`, and `Crypto Hmac Binary Data with SHA1`, each with pinned expected output. The existing `createReadStream` mock covers the streamed binary path for SHA1 via the new HMAC binary node.

### Test vectors

| Operation | Key / Input | Encoding | Expected digest |
|---|---|---|---|
| `SHA1("test")` | — | hex | `a94a8fe5ccb19ba61c4c0873d391e987982fbbd3` |
| `HMAC-SHA1("key", "test")` | `key` | hex | `671f54ce0c540f78ffe1e26dcf9c2a047aea4fda` |

### How to test

```
cd packages/nodes-base
pnpm jest nodes/Crypto
```

All existing Crypto tests plus the five added cases pass locally.

Follow-up: docs PR in `n8n-io/n8n-docs` to add `SHA1` to the Crypto node algorithm lists (both Hash and HMAC).

## Review / Merge checklist

- [ ] Docs updated (follow-up PR in `n8n-io/n8n-docs`).
- [x] Tests included.
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (not applicable — feature addition, not an urgent fix).


packages/nodes-base/nodes/Crypto/v2/CryptoV2.node.ts
```diff
@@ -222,6 +222,10 @@ const versionDescription: INodeTypeDescription = {
 					name: 'MD5',
 					value: 'MD5',
 				},
+				{
+					name: 'SHA1',
+					value: 'SHA1',
+				},
```
garritfra, line 284: Worth considering rolling SHA1 into the **Hash** action's `Type` dropdown in this PR too (applies equally to the matching block in `CryptoV1.node.ts`). After this change, the two dropdowns look like:

| Algorithm | `hash` | `hmac` |
|---|---|---|
| MD5 | ✅ | ✅ |
| **SHA1** | ❌ | ✅ (this PR) |
| SHA256 / SHA3-* / SHA384 / SHA512 | ✅ | ✅ |

The "HMAC doesn't depend on collision resistance" rationale in the PR description is correct, but it's an argument for *adding* SHA1 to HMAC, not for *withholding* it from Hash — especially since MD5 (weaker than SHA1 against collisions) is already in the Hash list. The status quo is the incoherent position; this PR partially fixes it.

Most legacy-interop asks I'd expect to drive this kind of request actually land on Hash, not HMAC: Git object IDs, S3 ETag-style integrity checks, recomputing `sha1(payload)` for webhook signature comparison, SAML/XML-DSig, SRI for legacy CDN assets. A user hitting any of those, not finding SHA1, and falling back to a Code node is the friction this PR is meant to remove — but only half-removes.

Cost is trivial: one `{ name: 'SHA1', value: 'SHA1' }` block in each of the two `hash`-scoped `Type` properties (v1 + v2), defaults unchanged, same alphabetic position. Test vector for `SHA1("test")` is `a94a8fe5ccb19ba61c4c0873d391e987982fbbd3`. Rolling it in also keeps the docs follow-up to a single change.

packages/nodes-base/nodes/Crypto/v2/test/Crypto.node.test.ts
```diff
@@ -85,6 +85,29 @@ describe('CryptoV2 Node', () => {
...
+					return params[paramName];
+				});
+				mockExecuteFunctions.getCredentials.mockResolvedValue({
+					hmacSecret: 'key',
+					signPrivateKey: '',
+				});
+
+				const result = await cryptoNode.execute.call(mockExecuteFunctions);
+
+				expect(result[0][0].json.data).toBe('671f54ce0c540f78ffe1e26dcf9c2a047aea4fda');
```
garritfra, line 110: The HMAC execute path has three branches reachable from `(action='hmac', type='SHA1')`:

```ts
const hashOrHmac = createHmac(type, hmacSecret);
if (binaryData) {
  if (binaryData.id) {
    // A: streamed via pipeline()
  } else {
    // B: in-memory Buffer.from(data, BINARY_ENCODING)
  }
} else {
  // C: string value — the only one tested in this PR
}
```

Cryptographically the algorithm string is opaque to the branching, so correctness isn't really at risk. The reasons to cover B (and ideally A) anyway:

- **Parity.** The existing `Crypto Hmac Binary Data` fixture in `CryptoTest.workflow.json` already exercises binary HMAC for MD5. If MD5 binary HMAC is worth a fixture, SHA1 binary HMAC is too — and SHA1 is the algorithm with the explicit legacy-interop justification.
- **Regression net.** The binary plumbing (`BINARY_ENCODING`, `pipeline()`, `getBinaryStream`) is exactly the kind of code a future refactor (e.g. towards `n8n-core`'s binary-data-manager) is likely to touch. Algorithm-name resolution sitting outside the test net there is the kind of thing that silently breaks on refactor.

Cheapest concrete adds:
- **v1**: clone the existing `Crypto Hmac Binary Data` fixture as a SHA1 variant (secret `'key'`, mocked binary contents are already `"test"`, pin `data: "671f54ce0c540f78ffe1e26dcf9c2a047aea4fda"`).
- **v2**: mirror this Jest case with `binaryData: true` and stub `assertBinaryData` to return `{ data: Buffer.from('test').toString(BINARY_ENCODING), id: undefined }`. Same expected hex.

The streamed path (branch A) is harder to mock cleanly in v2 Jest — the v1 fixture's `createReadStream` mock would already cover it for SHA1 if you add the v1 fixture, so I'd defer the v2 stream-path test.

garritfra (changes requested):

packages/nodes-base/nodes/Crypto/v2/CryptoV2.node.ts
maxabaumgarten: Done — the latest commit (6b76483) rolls SHA1 into the Hash Type dropdown in both V1 and V2, with a V2 Jest case for Hash-SHA1 (hex, `a94a8fe5ccb19ba61c4c0873d391e987982fbbd3`) and a `Crypto Hash into SHA1` node in the V1 workflow fixture. PR description updated to reflect the broader scope and the MD5/SHA1-parity argument.

packages/nodes-base/nodes/Crypto/v2/test/Crypto.node.test.ts
maxabaumgarten: Done. Added:

- **V2 Jest** — `should compute HMAC-SHA1 over in-memory binary data with hex encoding`, which mocks `helpers.assertBinaryData` to return `{ data: 'dGVzdA==', ... }` (no `id`) so it hits branch B and exercises the `Buffer.from(..., BINARY_ENCODING)` path. Expected digest `671f54ce0c540f78ffe1e26dcf9c2a047aea4fda`.
- **V1 workflow fixture** — `Crypto Hmac Binary Data with SHA1` node hung off the existing `Read Binary Files` node, which reuses the `createReadStream` mock and therefore also covers the streamed path (branch A) for SHA1. Pinned the same expected digest.

Skipped a standalone V2 Jest case for the streamed path per your suggestion, since the V1 fixture already exercises it.

garritfra (approved):
Lgtm. Thanks @maxabaumgarten!

maxabaumgarten:
@garritfra , I see there is a failing test. Is there something I need to modify to fix this?

garritfra:
@maxabaumgarten nothing to do on your end. E2e tests tend to be a bit flaky sometimes. I'll raise this internally, should hopefully be fixed soon.

Joffcom (approved):

JamesZhang666:
Hey @garritfra / @Joffcom — just checking in on this one. 
Two approvals are in and auto-merge is enabled, but the E2E Shard 3 flakiness has been blocking for ~3 weeks now. 
Any update on when that gets resolved, or is there a way to get this merged manually? Happy to rebase if needed.

garritfra:
@shortstacked can we quarantine this test?

garritfra:
This was a long time coming 😄 Thanks for your patience!

maxabaumgarten:
Thanks Team!

PR #29743, opened 2026-05-05
fix(Extract from File Node): Correct typo in binaryToProperty operation value

arpitshah29:
## Summary

Fixes a typo in the `ExtractFromFile` node where the operation value `'binaryToPropery'` was missing the letter `t` — it should be `'binaryToProperty'`.

**Changes:**
- `ExtractFromFile.node.ts`: Fixed typo in the `options` array value and the `includes()` check
- `moveTo.operation.ts`: Fixed typo in the `displayOptions` show condition and the `!==` guard

**How to test:**
1. Open a workflow and add the **Extract from File** node
2. Select the **Move File to Base64 String** operation
3. Run the node with a binary input — it should execute without errors

## Review / Merge checklist

- [ ] Tests included.

---
🤖 PR Summary generated by AI

### DIFF
--- a/packages/nodes-base/nodes/Files/ExtractFromFile/ExtractFromFile.node.ts
+++ b/packages/nodes-base/nodes/Files/ExtractFromFile/ExtractFromFile.node.ts
@@ -101,7 +101,7 @@ export class ExtractFromFile implements INodeType {
 					},
 					{
 						name: 'Move File to Base64 String',
-						value: 'binaryToPropery',
+						value: 'binaryToProperty',
 						action: 'Move file to base64 string',
 						description: 'Convert a file into a base64-encoded string',
 					},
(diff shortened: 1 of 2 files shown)


Joffcom:
Hi @arpitshah29 — thanks for spotting this! Looking at it more closely though, the misspelling is consistent across all four sites (option value, `includes()` check, `displayOptions` show condition, and the `!==` guard all use `binaryToPropery`), so it isn't actually causing a runtime issue — the node works correctly today because the producer and the consumer agree on the misspelled string.

Per our [contributing guide](../blob/master/CONTRIBUTING.md#community-pr-guidelines), we don't accept typo-only PRs unless they're paired with a logic change. And renaming this internal identifier would actually **break existing workflows** that have the legacy value persisted in their stored JSON (cubic-dev-ai flagged this), so the cost-benefit doesn't work out here.

Closing this one. Thanks for taking the time to look at the codebase — if you spot a typo that does cause a real runtime issue, or a behavioural bug paired with a typo fix, we'd be glad to review that. 🙏

PR #29747, opened 2026-05-05
feat(core): Dynamic 'one year' preset based on date range of data

sandra0503:
## Summary
- Adds `insightsPresetAvailability` in insights module settings (`oneYearRange`) so the editor can hide the **One year** date preset when `insights_by_period` data does not reach a full year of data
- Insights dashboard presets: hide **One year** when `oneYearRange` is `false`;
- Worker instances skip querying period metadata and use the same “no oldest row” path so `settings()` does not touch TypeORM entities that are not loaded there.

**How to test**
1. Instance with **&lt; 365 days** of compacted insights: open **Insights**, open the date range picker → **One year** should not appear (others unchanged).
2. Instance with **≥ 365 days** of history (or edge aligned with `keyRangeToDays.year`): **One year** should appear.
3. Optional: after pruning old periods, reload or refetch module settings and confirm the preset list updates if data span drops below a year.
- https://linear.app/n8n/issue/LIGO-498

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
+++ b/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
@@ -408,6 +408,23 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 		return aggregatedInsightsByTimeParser.parse(rawRows);
 	}
 
+	async getOldestPeriodStart(): Promise<Date | null> {
+		const result = await this.createQueryBuilder('insights')
+			.select('MIN(insights.periodStart)', 'oldest')
+			.getRawOne<{ oldest: Date | string | null }>();
+
+		if (result?.oldest === null || result?.oldest === undefined) {
+			return null;
+		}
+
+		const oldest =
+			result.oldest instanceof Date
+				? result.oldest
+				: new Date(result.oldest.replace(' ', 'T') + 'Z');
+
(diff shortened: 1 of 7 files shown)


packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
```diff
@@ -408,6 +408,23 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 		return aggregatedInsightsByTimeParser.parse(rawRows);
 	}
 
+	async getOldestPeriodStart(): Promise<Date | null> {
+		const result = await this.createQueryBuilder('insights')
+			.select('MIN(insights.periodStart)', 'oldest')
```
geemanjs, line 413: TIL - this does exactly the same as 
```sql
SELECT X FROM insights order by periodStart ASC LIMIT 1
```

geemanjs (commented):
I think whats annoying about this - is we will display 6 months even if there is only 3 months of data. Which makes our behaviour somewhat inconsistent. 

I'm tempted to say we just keep the button and do nothing 👀

sandra0503:
> I think whats annoying about this - is we will display 6 months even if there is only 3 months of data. Which makes our behaviour somewhat inconsistent.
> 
> I'm tempted to say we just keep the button and do nothing 👀

Yeah first I made them all dynamic but then it got too compley quickly. I also feel like we should leave it as is

PR #29750, opened 2026-05-05
fix(core): Tighten instance ai mutation and approval gates (no-changelog)

Cadiac:
## Summary

Collection of fixes to tighten approvals and prevent accidental mutations on instance AI. Addresses issues F4 (comment only, works as it should), F5, F10, F18 and F19, F20 is to be implemented separately as it's rather complex task.

https://linear.app/n8n/issue/INS-191/tighten-instance-ai-mutation-and-approval-gates

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -3255,6 +3291,17 @@ export class InstanceAiService {
 		} = suspended;
```
aalises, line 3349: `resolvePendingConfirmation()` returns before this path, so inline/sub-agent approvals can still proceed after the user is disabled or loses `instanceAi:message`. I'd revalidate before resolving pending confirmations too.

packages/@n8n/instance-ai/src/tools/workflows.tool.ts
```diff
@@ -591,12 +591,56 @@ async function handleRestoreVersion(
...
-		name: input.name,
-		description: input.description,
-	});
-	return { success: true };
+	// Gated under `updateWorkflow` — version metadata edits are workflow-record
+	// mutations, treated the same as live-workflow updates.
+	const resumeData = ctx?.agent?.resumeData as { approved: boolean } | undefined;
+	const suspend = ctx?.agent?.suspend as ((payload: unknown) => Promise<void>) | undefined;
+
+	if (context.permissions?.updateWorkflow === 'blocked') {
```
aalises, line 659: This `updateWorkflow` gate should also cover setup apply. Right now `workflows(action="setup")` can still mutate credentials/parameters via `applyNodeChanges()` when workflow updates are blocked.

packages/cli/src/modules/instance-ai/instance-ai.service.ts
```diff
@@ -2323,6 +2349,16 @@ export class InstanceAiService {
 	}
```
aalises, line 2382: This fetches a fresh user, but the rest of the scheduling/resume flow still uses the captured `user`. I'd pass the revalidated/freshest user onward so permission checks use current scopes.

packages/@n8n/instance-ai/src/tools/workflows.tool.ts
Cadiac: That feels out of the scope of this PR and what I was fixing here (version title/description updates not requiring approval), but I might as well address that here now 👍

aalises (dismissed):

aalises (approved):

PR #29760, opened 2026-05-05
fix: Replay sub-agent conversation on credential-setup nudge (no-changelog)

BerniWittmann:
## Summary

> User asks the assistant to set up a credential (e.g. "Set up a Slack OAuth2 credential for me in n8n").
> The assistant opens Chrome and the user watches it work — navigates to [api.slack.com](http://api.slack.com/), creates a new app, picks a workspace, types the app name, clicks "Create App", lands on the App Credentials page. The Client ID is visible on screen. Everything looks right.
> 
> The assistant pauses for ~5–10 seconds. Nothing visibly happens.
> The assistant abruptly says something like:
> 
> "I apologize for the confusion, but I don't have an active browser session or context about which credential setup we were working on. It appears the conversation may have been interrupted or I'm missing the previous context.

When the credential-setup sub-agent ends without calling `pause-for-user`, the orchestrator nudges it via a second `subAgent.stream(nudgeText)` call. Each Mastra `agent.stream()` invocation starts a fresh conversation, so the nudged sub-agent receives only the nudge string with no memory of what it just did. It apologetically replies *"I don't have an active browser session or context"*, that text becomes the orchestration tool's return value, and the parent agent surfaces the apology back to the user — mid-credential-setup, with the relevant Chrome tab still showing the answers on screen.

**Fix:** replay the prior conversation as input to the nudge stream call by passing `activeStream.messageList.get.all.aiV5.model()` plus the nudge as a final user message. The sub-agent now sees `system → briefing → assistant turns + tool calls + tool results → nudge` and can compose the missing `pause-for-user` from real context. The nudge prompt and input-building helper live in a small Mastra-free module (`browser-credential-setup.nudge.ts`) so they're unit-testable without scaffolding the orchestration.

https://linear.app/n8n/issue/NODE-4977

## Review / Merge checklist

- [x] Tests included.


Cadiac (approved):
🚀 nice fix to this gap

PR #29764, opened 2026-05-05
Instance AI workflow setup nodes grouping

scdekov:
## Summary

Grouping of instance ai setup requests based on:
* Shared credentials
    - 2 nodes having the same credential type are merged only if the second node does not also has parameters to be setup
* Parent/sub nodes eg model/tools/memory of an agent

Review notes:
* Backend change: Added `parentNode` to the `workflowSetupNodeSchema` so that FE doesn't have to fetch the whole workflow in order to group based on parent/sub nodes
* FE
* - Split the Cards rendering to `WorkflowSetupCard` and `WorkflowSetupGroupCard`, both using a new `WorkflowSetupSectionBody` component
* - Extracted `WorkflowSetupWizardFooter` from `WorkflowSetupWizard` as it's now reused for the group and non group card components
* - Think of setup section as an entity containing credential picker and/or parameter inputs for a node (the creds picker could be shared). Then setup step is either a card (single section) or cardGroup (one or more sections). "Cards" are responsible for the visualizing the step data
* Ignore changes on the fixtures .json and .jsonl files

https://linear.app/n8n/issue/ADO-5119/add-cards-grouping

## Review / Merge checklist

- [ ] Tests included. 


packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -785,5 +866,23 @@ export async function analyzeWorkflow(
...
 	);
 
+	// Stamp `parentNode` on every sub-node setup request so the frontend can
+	// render the parent header even when the parent has no setup request of
+	// its own. Sub-node membership is derived from the full workflow graph,
+	// not just the (filtered) setup requests.
+	const subnodeToRootParent = buildSubnodeToRootParentMap(
+		workflowJson.nodes,
+		workflowJson.connections as unknown as IConnections,
+		setupRequests.map((req) => req.node.name),
```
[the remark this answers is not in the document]
scdekov: Since we're not executing nodes in the instance-ai setup flow, it's fine (or even better) to show the subnode as part of the parent which is actionable since that will result in fewer cards and it still make sense to the user.

packages/@n8n/api-types/src/schemas/instance-ai.schema.ts
```diff
@@ -245,6 +245,17 @@ export const workflowSetupNodeSchema = z.object({
 			'Whether this node still requires user intervention. ' +
 				'False when credentials are set and valid, parameters are resolved, etc.',
 		),
+	parentNode: z
```
CharlieKolb: I initially thought this meant predecessor in the graph, though I also struggle with a better name... `associatedMainNode`... meh

packages/@n8n/instance-ai/src/tools/workflows/setup-workflow.service.ts
```diff
@@ -746,6 +748,85 @@ export function buildCompletedReport(
 	return result;
 }
 
+// ── Sub-node grouping ───────────────────────────────────────────────────────
+
+export interface RootParentNode {
```
CharlieKolb: [nit] Should this be `type RootParentNode = Pick<INode, 'name' | 'type' | 'typeVersion' | 'id'>`?

CharlieKolb (approved):
Really nice refactors here! Really like your approach for the groupings also

packages/@n8n/api-types/src/schemas/instance-ai.schema.ts
scdekov: Good point. Wdyt about `subnodeRootNode`? That also hints that the prop is present only for subnodes which is a common term we use for eg model -> agent relation

CharlieKolb (approved):

PR #29766, opened 2026-05-05
fix: Use /form base URL for Form Trigger production links (no-changelog)

BerniWittmann:
## Summary

When a user asks the n8n Instance AI to share or use a Form Trigger production URL, the agent currently builds it from the webhook base URL — producing `<base>/webhook/<id>`, which 404s. Form Triggers serve only at `<base>/form/<id>` after the duplicate `/webhook/` route was removed in [NODE-4871](https://linear.app/n8n/issue/NODE-4871).

This adds a separate `formBaseUrl` plumbed through `OrchestrationContext` and uses it in the system prompt's instance-info section, with an explicit note that `/form/` is a distinct prefix from `/webhook/`.

### How to test

1. Build a workflow with a Form Trigger and activate it.
2. Ask the assistant for the production form URL.
3. Confirm the URL it shares is `<base>/form/<id>`, not `<base>/webhook/<id>`, and that opening it renders the form (no 404).

- Linear: https://linear.app/n8n/issue/NODE-4978
- Caused by: https://linear.app/n8n/issue/NODE-4871 (Form Triggers reachable from both `/form` and `/webhook` — already shipped)

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


packages/@n8n/instance-ai/src/agent/system-prompt.ts
```diff
@@ -33,16 +34,17 @@ The user's current local date and time is: ${isoTime}${tzLabel}.
...
 ## Instance Info
 
 Webhook base URL: ${webhookBaseUrl}
+Form base URL: ${formBaseUrl}
 
 Some trigger nodes expose HTTP endpoints. Always share the full production URL with the user after building a workflow that uses one of these triggers. Each type has a distinct URL pattern:
 
 - **Webhook Trigger**: ${webhookBaseUrl}/{path} (where {path} is the node's webhook path parameter).
-- **Form Trigger**: ${webhookBaseUrl}/{path} (or ${webhookBaseUrl}/{webhookId} if no custom path is set). Same pattern as Webhook — no /chat suffix.
+- **Form Trigger**: ${formBaseUrl}/{path} (or ${formBaseUrl}/{formId} if no custom path is set). The Form Trigger lives under /form/, NOT /webhook/ — they are separate URL prefixes. Do NOT use the Webhook base URL for Form Triggers.
```
aalises: Niceee. One small correction is that I think this should probably be `{webhookId}`, not `{formId}`.

Form Trigger falls back to the node’s `webhookId` when no custom path is set, so `formId` may send the agent looking for a field that does not exist...

We could also add a small `system-prompt.test.ts` case covering the Form Trigger guidance? Something like: when `webhookBaseUrl` and `formBaseUrl` are provided, the prompt includes `/form/{webhookId}` and does not describe Form Trigger as using `/webhook`.

aalises:
Also a broader thought: this prompt fix is useful, but the URL pattern feels a bit hidden here.

Since the Form Trigger node already owns its webhook metadata (`webhooks[].nodeType = 'form'` and the path expression), We could make the production URL shape discoverable through node metadata / the  workflow SDK instead of hard-coding it in the Instance AI system prompt? (see `builderHints` on the workflow sdk).

That would make this easier to maintain when trigger URL behavior changes: the node/SDK stays the source of truth, and Instance AI can render the guidance from there.

Basically moving from having it on `system-prompt` vs. making it discoverable on the SDK when getting the node as a hint

packages/@n8n/instance-ai/src/agent/system-prompt.ts
BerniWittmann: adjusted the `webhookId` and extended the `system-prompt.test.ts` tests around this.

aalises (dismissed):

BerniWittmann:
On your broader thought, yes makes sense for instance-ai to have the url schemese discoverable. 
Yet, I'd like to first land this immediate bug-fix and maybe do the broader rework in a follow up. (I also probably lack much context around this right now :) )

aalises:
> On your broader thought, yes makes sense for instance-ai to have the url schemese discoverable. Yet, I'd like to first land this immediate bug-fix and maybe do the broader rework in a follow up. (I also probably lack much context around this right now :) )

Definitely! we can leave it to some follow-up

BerniWittmann:
@aalises Sorry for bothering again, had to fix some simple merge conflicts and that requires a new review

aalises (approved):

PR #29778, opened 2026-05-05
fix(Schedule Node): Fix hourly intervals that don't divide evenly into 24h

### LINKED ISSUE #20953: Cannot execute the worklfow automaticly :/
### 🧠 **Description of the issue**

The workflow does not execute automatically according to the configured schedule.
I have to manually run it every time.

---

### ⚙️ **Details**

* **Trigger type:** `Schedule Trigger`
* **Configured interval:** Every 18 hours
* **Node configuration:**

  ```json
  {
    "parameters": {
      "rule": {
        "interval": [
          {
            "field": "hours",
            "hoursInterval": 18
          }
        ]
      }
    },
    "type": "n8n-nodes-base.scheduleTrigger",
    "typeVersion": 1.2,
    "position": [416, -64],
    "id": "815b88ea-9656-4507-9b04-791541d72ea0",
    "name": "Schedule Trigger"
  }
  ```

---

### ⚠️ **Observed behavior**

The workflow doesn’t start automatically based on the interval.
It only runs when I manually click **“Execute Workflow”**.

---

### ✅ **Expected behavior**

The workflow should automatically trigger every 18 hours without manual execution.

---

### 🧩 **Additional notes**

* The workflow **is activated** (`Active = true`).
* No error message is shown.
* The issue seems related to the **Schedule Trigger node**, which doesn’t execute automatically.
* Please note that I’m using **[specify whether n8n Cloud or self-hosted (Docker/local)]**.

michael-radency:
## Summary

Schedule Trigger workflows configured with an hourly interval that doesn't divide evenly into 24 (e.g. every 18 hours) never fired after the first execution.

**Root cause:** `toCronExpression` produced `*/18 * * * * *` for `hoursInterval: 18`. Cron's `*/N` syntax matches fixed clock hours divisible by N — so `*/18` only fires at 00:xx and 18:xx. The recurrence check then computed `(18 - 18 + 24) % 24 = 0 < 18`, blocking every subsequent tick permanently.

**Fix:** For hour intervals where `24 % hours !== 0`, emit `* * * * * *` (tick every hour) and let the existing `recurrenceCheck` gate actual emission by elapsed time. Intervals that divide evenly into 24 (1, 2, 3, 4, 6, 8, 12 h) continue using `*/N` as before — their recurrence check always succeeds at the right tick.

**Tests added:**
- `toCronExpression` — verifies correct cron expression is generated for `hoursInterval: 18`
- End-to-end trigger test (no `triggerAtMinute`) — confirms three consecutive emissions 18 h apart
- End-to-end trigger test with `triggerAtMinute: 0` — mirrors the exact UI configuration from the reported issue

https://linear.app/n8n/issue/NODE-3806/community-issue-cannot-execute-the-worklfow-automaticly-schedule
closes #20953

## Review / Merge checklist

- [ ] Tests included. 


DawidMyslak (approved):
LGTM 👌

nihaals:
See https://github.com/n8n-io/n8n/issues/23711#issuecomment-4238332691. Making it always trigger every hour and relying on `recurrenceCheck` could go in this PR and would reduce the number of different cases these functions have which already causes issues.

PR #29786, opened 2026-05-05
ci: Validate required pr quality checks with ci-filter

Matsuuu:
## Summary

Instead of adding a required check for each step in `ci-pr-quality.yml`, combine results behind one `Required PR Quality Checks` job. Uses ci-filter to validate runs and allow skipped runs.

https://linear.app/n8n/issue/CAT-2958/

## Review / Merge checklist

- [ ] Tests included. 

### DIFF
--- a/.github/workflows/ci-pr-quality.yml
+++ b/.github/workflows/ci-pr-quality.yml
@@ -106,6 +106,7 @@ jobs:
     needs: [check-ownership-checkbox, check-pr-size]
     if: always()
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-pr-quality.yml
```diff
@@ -1,6 +1,7 @@
 name: 'CI: PR Quality Checks'
 
 on:
+  merge_group:
```
shortstacked, line 4: Does it need to be on the merge group too?

Matsuuu: If we are going to add a Required status check on branch protection rules, I believe we need to also have this run on merge groups. For merge groups it will just skip all of the checks and return a success from the validation step

shortstacked (approved):

.github/workflows/ci-pr-quality.yml
shortstacked: Yeah you are right, I forgot about that!

PR #29787, opened 2026-05-05
refactor(core): Combine insights by workflow count and page query

sandra0503:
## Summary

**Problem:**

The "Insights by-workflow" pagination was making two heavy traversals over the database: one to count the total grouped results and a second to fetch the actual page data.

**Solution:**
We’re not using COUNT(*) OVER() on the page query. We’ll run the page fetch and a separate total count in parallel instead.

The old code loaded every grouped workflow row just to get a count, then ran the same work again for the page. Now the page query only returns the slice we need, and the count is a lighter grouped count that doesn’t follow the page sort.

**Outdated**

First attempt (commit [1bf01a2](https://github.com/n8n-io/n8n/pull/29787/changes/1bf01a2dbd2704711fd687634a13f9ae42349a90)) was to reduce the call into a single query with including the row count on each row. This got dismissed (see [comment](https://github.com/n8n-io/n8n/pull/29787#discussion_r3204076865)).

Closes [LIGO-504](https://linear.app/n8n/issue/LIGO-504/optimize-insights-pagination-performance)

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
+++ b/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
@@ -307,6 +307,27 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 		return [column, order.toUpperCase() as 'ASC' | 'DESC'];
 	}
 
+	private async countInsightsByWorkflowGroups(
+		groupedQuery: SelectQueryBuilder<InsightsByPeriod>,
+	): Promise<number> {
+		const groupsSubQuery = groupedQuery
+			.clone()
+			.select('metadata.workflowId', 'workflowId')
+			.addSelect('metadata.workflowName', 'workflowName')
+			.addSelect('metadata.projectId', 'projectId')
+			.addSelect('metadata.projectName', 'projectName');
+		groupsSubQuery.orderBy();
+
+		const row = await this.manager
+			.createQueryBuilder()
+			.select('COUNT(*)', 'count')
+			.from(`(${groupsSubQuery.getQuery()})`, 'workflow_groups')
+			.setParameters(groupsSubQuery.getParameters())
+			.getRawOne<{ count: string | number | undefined }>();
+
+		return row?.count !== undefined ? Number(row.count) : 0;
+	}
+
 	async getInsightsByWorkflow({
 		startDate,
 		endDate,
@@ -348,6 +369,7 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 								ELSE 1.0 * SUM(CASE WHEN insights.type = ${TypeToNumber.runtime_ms.toString()} THEN value ELSE 0 END) / ${sumOfExecutions}
 							END AS "averageRunTime"`,
 			])
+			.addSelect('COUNT(*) OVER()', 'totalCount')
 			.innerJoin('insights.metadata', 'metadata')
 			// Use a cross join with the CTE
 			.innerJoin('date_ranges', 'date_ranges', '1=1')
@@ -363,8 +385,11 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 			rawRowsQuery.andWhere('metadata.projectId = :projectId', { projectId });
 		}
 
-		const count = (await rawRowsQuery.getRawMany()).length;
-		const rawRows = await rawRowsQuery.offset(skip).limit(take).getRawMany();
+		const rawRows = await rawRowsQuery.clone().offset(skip).limit(take).getRawMany();
+		const count =
+			rawRows.length > 0
+				? Number(rawRows[0].totalCount as string | number)
+				: await this.countInsightsByWorkflowGroups(rawRowsQuery);
 
 		return { count, rows: aggregatedInsightsByWorkflowParser.parse(rawRows) };
 	}
--- a/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
+++ b/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
@@ -308,24 +308,16 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 	}
 
 	private async countInsightsByWorkflowGroups(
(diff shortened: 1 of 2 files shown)


packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
```diff
@@ -363,8 +385,11 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
 			rawRowsQuery.andWhere('metadata.projectId = :projectId', { projectId });
 		}
 
-		const count = (await rawRowsQuery.getRawMany()).length;
-		const rawRows = await rawRowsQuery.offset(skip).limit(take).getRawMany();
+		const rawRows = await rawRowsQuery.clone().offset(skip).limit(take).getRawMany();
+		const count =
+			rawRows.length > 0
+				? Number(rawRows[0].totalCount as string | number)
+				: await this.countInsightsByWorkflowGroups(rawRowsQuery);
```
konstantintieber: > Empty State Fallback: If a request returns no rows (e.g., paging past the end), we execute a lightweight fallback query to determine if the result is truly zero or just an empty page (last page).

Do we know that this fallback is needed? Which case does it cover?
Adding this fallback seems like a separate code change from solving the original problem.
Is that what the `returns total count when page is past the end` is about?

konstantintieber: Or asking differently: how would I test running into this fallback case manually?

alielkhateeb: Ok, I can answer your question @konstantintieber 

The problem with `COUNT(*) OVER()` is that it only gives us the total count on rows that are actually returned. If the paginated query goes out of bounds, there are no rows, so `total_rows` is inexistent and we can’t get the total count from the result.

For example, with a `users` table like this:

| id | name |
| -- | ---- |
| 1 | Alice |
| 2 | Bob |
| 3 | Carol |

If we run:

```sql
SELECT
    id,
    name,
    COUNT(*) OVER() AS total_rows
FROM users
OFFSET 5;
````

The query returns no rows, even though the actual total count is 3.

That’s why the previous fallback existed: when we couldn’t get a total from the paginated result, we had to run another query to count the items. So to answer your question, you can reach the fallback if you overshoot the last page for example.

This is why I would go with a clear separate query for the count, running the paginated query and the count query together with `Promise.all`, since it avoids this edge case (conditionals), and makes the behaviour more explicit.

I pushed the changes here: bad623500b041efdbf49cdf67aadaac84c4db539

Note: I did not benchmark any performance enhancement but honestly from the code alone it should improve the performance drastically as it achieves eliminating the unnecessary `getRawMany` - as described in the ticket.

Let me know what you think :)

File: packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts [hunk 2]
```diff
@@ -356,15 +369,22 @@ export class InsightsByPeriodRepository extends Repository<InsightsByPeriod> {
...
+		const paginatedQuery = rawRowsQuery
+			.clone()
+			.orderBy(this.escapeField(sortField), sortOrder)
+			.offset(skip)
+			.limit(take);
+
+		const [count, rawRows] = await Promise.all([
+			this.countInsightsByWorkflowGroups(rawRowsQuery),
+			paginatedQuery.getRawMany(),
+		]);
```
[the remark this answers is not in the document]
alielkhateeb: I intentionally decided against that for code clarity - no conditionals, as well as handling large scale, having a separate COUNT query usually means:

- less memory usage
- less sorting/materialization
- faster response times
- better query plans overall.

packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts
alielkhateeb: I also added more objective reasons why separate COUNT query is better at scale:
> I intentionally decided against that for code clarity - no conditionals, as well as handling large scale, having a separate COUNT query usually means:
> - less memory usage
> - less sorting/materialization
> - faster response times
> - better query plans overall.

konstantintieber: Fully agree on the count part, where I was still stumped was the added fallback.

> That’s why the previous fallback existed

Just checking: currently on master, there is [no fallback](https://github.com/n8n-io/n8n/blob/master/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts#L369), so you're talking about [the one Sandra added](https://github.com/n8n-io/n8n/blame/1bf01a2dbd2704711fd687634a13f9ae42349a90/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts#L392), right?

What I'm wondering: can we instead not allow paginated queries with out of bounds parameters?

alielkhateeb: Yes that's the part I'm talking about.
Of course on master we don't have a fallback, because we actually do a FULL table scan without any `limit` or `offset` to get the count https://github.com/n8n-io/n8n/blob/820128196cf550ab8cf371fbebb3457b9fd35d22/packages/cli/src/modules/insights/database/repositories/insights-by-period.repository.ts#L366-L367

This is exactly why performance was really bad. So what Sandra did was not incorrect, the fallback IS needed if we want to handle out-of-bounds pagination parameters with `COUNT(*) OVER`.

I would vote against restricting the user from overshooting the last page, this is not really an "error". I would rather reserve errors to actual errors, clients usually deal with empty results/pages better than an out-of-bounds error, at least ALL paginated API designs I personally built or consumed allowed out of bounds parameters :)

konstantintieber: I'm not going to veto supporting out-of-bounds params 👍 the code looks good.

sandra0503: Thanks for looking into it. I updated the PR description with the final approach taken.

alielkhateeb (approved):

PR #29802, opened 2026-05-05
feat(Microsoft Outlook Node): Add support for recurring event instances

DawidMyslak:
## Summary

The Microsoft Outlook node's "Get Many Events" operation only returns series master events for recurring calendar events, missing individual occurrences. This is because it uses the `/calendars/{id}/events` Graph API endpoint, which by design does not expand recurring events.

This PR adds an **"Include Recurring Event Instances"** toggle to the Get Many Events operation. When enabled, the node switches to Microsoft Graph's [`/calendarView`](https://learn.microsoft.com/en-us/graph/api/calendar-list-calendarview) endpoint, which expands recurring events into individual instances within a user-specified date range. The toggle defaults to `false` for backward compatibility.

**How to test:**
1. Configure a Microsoft Outlook node with Event → Get Many
2. With "Include Recurring Event Instances" **off** — behavior is unchanged, recurring events appear as series masters
3. With "Include Recurring Event Instances" **on** — set a date range, recurring events are expanded into individual occurrences

- https://linear.app/n8n/issue/NODE-3383
- closes #17747

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/nodes-base/nodes/Microsoft/Outlook/v2/actions/event/getAll.operation.ts
+++ b/packages/nodes-base/nodes/Microsoft/Outlook/v2/actions/event/getAll.operation.ts
@@ -21,6 +21,40 @@ export const properties: INodeProperties[] = [
 			},
 		},
 	},
+	{
+		displayName: 'Include Recurring Event Instances',
+		name: 'includeRecurringInstances',
+		type: 'boolean',
+		default: false,
+		description:
+			'Whether to expand recurring events into individual instances within the specified date range. When disabled, recurring events are returned as a single series master event.',
+	},
+	{
+		displayName: 'Start Date',
+		name: 'startDateTime',
+		type: 'dateTime',
+		default: '',
+		required: true,
+		description: 'Start of the date range to retrieve event instances for',
+		displayOptions: {
+			show: {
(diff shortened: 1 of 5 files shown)


packages/nodes-base/nodes/Microsoft/Outlook/v2/actions/event/getAll.operation.ts
```diff
@@ -61,6 +95,11 @@ export const properties: INodeProperties[] = [
 		type: 'collection',
 		placeholder: 'Add Filter',
 		default: {},
+		displayOptions: {
+			show: {
+				includeRecurringInstances: [false],
```
dlavrenuek, line 100: The docs state that some of the `OData` parameters are supported but not which ones. Based on some answers I found in the internet $filter is indeed either not supported at all or does not work the same way as for events endpoint, that's a shame :(

dlavrenuek (approved):
🎸

DawidMyslak:
@dlavrenuek I actually forgot to include my testing video 🤦, just so it's documented:
https://www.loom.com/share/10482dca55f84517a7c8a58e3b9661be

PR #29814, opened 2026-05-05
feat(core): Add observational memory storage foundation

bjorger:
## Summary

- Add SDK observation types, cursors, locks, and BuiltObservationStore.
- Implement observation storage in InMemoryMemory and N8nMemory.
- Add DB entities, repositories, and migrations for observations, cursors, and locks.
- Shape observations as the durable queue that can later feed thread-scoped working memory compaction.
- Use hard-delete semantics for compacted observation inputs.

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/db/src/migrations/common/1784000000000-CreateAgentObservationTables.ts
```diff
@@ -0,0 +1,60 @@
...
+			.withColumns(
+				column('id').varchar(36).primary.notNull,
+				column('scopeKind').varchar(20).notNull.withEnumCheck(['thread', 'resource', 'agent']),
+				column('scopeId').varchar(255).notNull,
+				column('kind').varchar(64).notNull,
+				column('payload').json.notNull,
+				column('durationMs').bigint,
+				column('schemaVersion').int.notNull,
+			)
+			.withIndexOn(['scopeKind', 'scopeId', 'kind', 'createdAt'])
```
mike12345567, line 33: Since this doesn't have a FK to the actual thread itself due to the polymorphic nature of the scopeId, do we correctly clean up this memory when the thread is deleted/removed?

packages/@n8n/agents/src/__tests__/memory-store-observations.test.ts
```diff
@@ -0,0 +1,273 @@
+import { InMemoryMemory } from '../runtime/memory-store';
```
mike12345567, line 1: I see we removed the old integration tests for memory - would it be possible to create an integration test for observational memory? This won't run in CI, but can be used by engineers locally to check given a real model if everything works roughly as expected - they were very useful for us when building out the SDK initially.

packages/@n8n/agents/src/types/sdk/memory.ts
```diff
@@ -140,6 +146,7 @@ export interface MemoryConfig {
 	};
 	semanticRecall?: SemanticRecallConfig;
 	titleGeneration?: TitleGenerationConfig;
+	observationalMemory?: ObservationalMemoryConfig;
```
mike12345567: Should we make this capability gated type? E.g. you can't add observational memory information unless its a memory type that supports it:
```
export type MemoryConfig =
    | (MemoryConfigBase & {
        memory: BuiltMemory;
        observationalMemory?: undefined;
      })
    | (MemoryConfigBase & {
        memory: ObservationCapableMemory;
        observationalMemory: ObservationalMemoryConfig;
      });
```

mike12345567 (commented):
Looks good - just one comment about the migration which I think we need to potentially look at, don't want to leave memory around if the thread is gone.

mike12345567 (approved):

PR #29816, opened 2026-05-05
fix(editor): Refresh node icon when diff sidebar selection changes

MarcL:
## Summary

In the workflow version-history diff view, the sidebar header's node icon stayed locked to whatever was first rendered — clicking a different node would update the title text but not the icon. Root cause was a Vue reactivity bug in `NodeIcon.vue`: prop values were passed to `useNodeIconSource` as plain values rather than getters, so the inner `computed`/`toValue` chain captured a one-time snapshot and never re-tracked. Passing getters restores the reactive contract the composable was designed for.

A second smaller correctness fix in `useWorkflowDiffUI` resolves the selected node from the target workflow first (then source, then the diff entry) so that nodes whose type changed between versions surface the new icon, mirroring the existing pattern in `useReviewChanges`.

To test: open Version History on a workflow with multiple saved versions that differ by node, open the diff, click between nodes — the sidebar icon should now switch with each click.

https://linear.app/n8n/issue/LIGO-521

## Review / Merge checklist

- [ ] Tests included. 

🤖 PR Summary generated by AI

### DIFF
--- a/packages/frontend/editor-ui/src/app/components/NodeIcon.test.ts
+++ b/packages/frontend/editor-ui/src/app/components/NodeIcon.test.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect, vi } from 'vitest';
+import { mount } from '@vue/test-utils';
+import { createTestingPinia } from '@pinia/testing';
+import { ref, computed, defineComponent, h } from 'vue';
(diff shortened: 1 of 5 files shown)


sandra0503 (approved):

sandra0503:
Looks good to me!

PR #29800, opened 2026-05-05
refactor: Extract shared eval helpers (no-changelog)

BerniWittmann:
## Summary

Refactor-only PR. Two no-functional-change extractions that prepare
`@n8n/instance-ai` for the upcoming computer-use eval harness (NODE-4974).
Pulled out of that branch so reviewers can absorb them in isolation.

> **Stacked with #29797.** This PR is the base of a 2-PR stack —
> #29797 (`feat: Computer-use evaluation harness`) targets this branch
> and won't merge cleanly until this lands. Review this one first; once
> it merges, GitHub will auto-retarget #29797 to master.

### 1. Shared chat-loop helpers

The SSE-stream consumer, `WaitConfig` interface, run-finish polling,
background-task draining, confirmation auto-approval, and the small event
helpers move from `evaluations/harness/runner.ts` (~265 lines) into a new
`evaluations/harness/chat-loop.ts`. The upcoming computer-use eval needs
the same primitives end-to-end — keeping a single copy now prevents the
two flows from drifting (e.g. on the `MAX_CONFIRMATION_RETRIES = 5` cap).

### 2. Shared LLM-as-judge utility

Verdict-parsing logic with markdown fallback (`**Verdict:** PASS`,
`Pass: true`, etc.) moves into `src/utils/llm-judge.ts` and is consumed
by the existing `evaluations/binaryChecks/checks/create-llm-check.ts`
factory. The same helper will back the upcoming computer-use LLM grader.
17-test suite covers the parser including the markdown-fallback branches.

### 3. Eval client extensions

`evaluations/clients/n8n-client.ts` gains:
- `versionId: string` and `description?: string` on `WorkflowResponse`
  (already returned by the n8n REST API; just makes the type accurate).
- `deleteThread(threadId)` — DELETE on `/rest/instance-ai/threads/:id`,
  for thread cleanup at the end of an eval run.
- `listWorkflowIds()`, `listCredentialIds()`, `listDataTableIds()`,
  `createWorkflow()` — verification helpers used by the upcoming
  computer-use eval cleanup pass. Each is a thin wrapper over an
  existing endpoint.

## How to test

Refactor-only — covered by the existing test suite:

```bash
cd packages/@n8n/instance-ai
pnpm typecheck
pnpm lint
pnpm test
```

All 1,351 tests pass. The extracted chat-loop is exercised by the
existing `harness/runner.ts` callers; the new `llm-judge.test.ts` covers
the verdict parser including JSON-fenced, bare-JSON, and markdown
[...]

PR #29840, opened 2026-05-06
feat: Add LLM-as-judge grader to computer-use eval (no-changelog)

BerniWittmann:
## Summary

Adds the LLM-as-judge grader (`graders/llm.ts`) to the computer-use eval
harness. Trace-level graders verify the agent called the right tools but
can't read English; this grader closes the gap. Auto-appended to every
scenario by the runner.

> **Stacked on #29797.** This PR targets the
> `feat/instance-ai-computer-use-eval` branch, which ships the
> deterministic harness alone. Review #29797 first; once it merges,
> GitHub will auto-retarget this PR to master.

### What this adds

- `evaluations/computer-use/graders/llm.ts` — the grader. Builds a
  prompt from the user's request, the agent's final text, and a
  compact tool-call summary. Asks a small/cheap LLM (Haiku 4.5 by
  default) "did the assistant succeed?". Parses the verdict via the
  shared `parseJudgeVerdict` helper from #29800.
- `LlmTaskCompletedGrader` interface in `types.ts` — supports
  optional `criteria`, `model`, `timeoutMs`.
- Auto-append in `cli.ts withDefaultGraders` so every scenario gets
  it for free; explicit `llm.taskCompleted` entries in scenario JSON
  override (used by M.1, M.3, M.5 to teach the judge adversarial
  scoring rules).
- Dispatch case + `userPrompt`/`scenarioCategory` plumbing in
  `applyGrader` / `runGraders`.
- 11-test suite covering verdict parsing, prompt construction
  (category/criteria/finalText injection, truncation at 8 000 chars,
  tool-call summarisation with footer), agent error and timeout
  handling.
- Harness realignment: `chat.ts` now calls `ensureThread` before
  sending the first message. Master scoped Instance AI threads to a
  project (#31565) and removed the old auto-create-on-first-message
  path, so without this every run aborts server-side with
  `thread "…" has no bound project`. Mirrors what the workflow eval
  harness already does.
- README docs.

### Why a separate PR

The deterministic harness in #29797 is reviewable on its own — it
runs scenarios end-to-end and scores them mechanically. The LLM judge
is the qualitative layer on top: it adds runtime cost (one Anthropic
call per scenario), introduces a new dependency on an API key during
eval runs, and has its own failure modes (timeouts, unparseable
verdicts) worth reviewing in isolation.

This PR's diff is ~250 counted lines — fits cleanly under the 1 000-line
size limit.

## How to test

After #29797 lands, this branch rebases onto master automatically:

```bash
cd packages/@n8n/instance-ai
pnpm typecheck
pnpm lint
pnpm test
```

End-to-end (smallest scenario, no browser needed), from
`packages/@n8n/instance-ai`:

```bash
pnpm exec dotenvx run -f ../../../.env.local -- \
  pnpm eval:computer-use --filter 6.1
```

- https://linear.app/n8n/issue/NODE-4974
- Stacked on #29797 (which stacks on #29800)

## Review / Merge checklist

- [x] Tests included.

🤖 PR Summary generated by AI


dlavrenuek:
@BerniWittmann I was not able to run the tests with the instructions from PR description.

> pnpm eval:computer-use -- --only 6.1

this command fails with an unknown argument error, looking at cli.ts `--` throws an error and `--only` is also not registered. I guess some commands to run the tests did change in the meanwhile on the main branch?

BerniWittmann:
> @BerniWittmann I was not able to run the tests with the instructions from PR description.
> 
> > pnpm eval:computer-use -- --only 6.1
> 
> this command fails with an unknown argument error, looking at cli.ts `--` throws an error and `--only` is also not registered. I guess some commands to run the tests did change in the meanwhile on the main branch?

@dlavrenuek Ah yes, the harness got consolidated a bit on master, so the new parameter name is `filter`not `only`

`pnpm eval:computer-use -- --filter 6.1`

Updated PR description + README

dlavrenuek:
@BerniWittmann are you sure that the instructions are correct? I still can't make it work with the example command. Looking at https://github.com/n8n-io/n8n/blob/745c9575f9fa0fedd325c1774fbbdc5823db5803/packages/%40n8n/instance-ai/evaluations/computer-use/cli.ts#L248 the email/password are taken from args and not from env, so the instructions on `.env.local` do seem incorrect. the `--` argument also throws an exception, the only way I could start the eval was with `pnpm eval:computer-use --filter 6.1 --email xxx --password xxx` which connects computer use (status changes to connected in AI Assistant UI) but the tests fail with 

> 2026-06-17T11:50:50.539Z [INFO] [6.1-curl-connectivity] start (shell)
2026-06-17T11:50:53.892Z [INFO] [6.1-curl-connectivity] FAIL trace.mustCallMcpServer: agent never invoked any computer-use tool — likely failed to propose it
2026-06-17T11:50:53.892Z [INFO] [6.1-curl-connectivity] FAIL trace.mustCallTool: tool "shell_execute" was never called (saw 0 other calls)
2026-06-17T11:50:53.892Z [INFO] [6.1-curl-connectivity] FAIL trace.finalTextMatches: final text does not match required patterns (got: "...")
2026-06-17T11:50:53.892Z [INFO] [6.1-curl-connectivity] FAIL llm.taskCompleted: The user asked the assistant to run a curl command to test connectivity to the OpenAI API. The assistant made no tool calls and provided no response content. This is a failure to attempt the task — the assistant should have executed a shell command (using the shell execution tool) to test the connection, such as `curl -I https://api.openai.com` or a similar connectivity check. Instead, the assistant appears to have done nothing.

I assume that these tests should be successful?

BerniWittmann:
@dlavrenuek 
Regarding tests not running: 
* had to do a small fix, as master and this PR just drifted in the meantime, which caused the test failures (need to start a thread first before sending a message). Works now again

Regarding the flags/commands you have two options: 
* What i typically do: is run from root of repo with 
`pnpm exec dotenvx run -f .env.local -- pnpm --filter @n8n/instance-ai eval:computer-use --filter M.1`
This assumes cwd is repo root and .env.local is at repo root as well
* Or run from instance-ai package (which i think you tried to do):
`pnpm exec dotenvx run -f ../../../.env.local -- \
  pnpm eval:computer-use --filter 6.1`
 This still assumes .env.local is at repo root. However as you already figured out, you can also use `--email` and `--password` instead. 
 
 dotenvx essentially loads the env file into the args. 
 
 I made some small adjustments to the Readme to make this a bit clearer. 
 
 Also the instance itself needs to be run with `E2E_TESTS=true`
 
 Let me know whether this helps, or is still not working. I am happy to help

dlavrenuek:
@BerniWittmann running with e2e being true did help. When starting the eval for the first time it passes, starting it again sometimes fails. It says that the daemon keeps running after evals and on the second run it is reused and only about 20% of the follow up runs are successful. Without looking at how the evals are setup that does not seem very reliable, can you reproduce it?

BerniWittmann:
@dlavrenuek From an initial suspicion i would suspect this might be related rather to the connection to the relay. But will look into it

BerniWittmann:
@dlavrenuek I think i found the stability issues across test -> retest (with daemon reuse): 
See commit: https://github.com/n8n-io/n8n/pull/29840/changes/f64b660844c6cf3b5ac3afa4d1464a82ccc8b29c

dlavrenuek:
@BerniWittmann ah, then it makes sense that the tests fail. I've fixed the same problem in https://github.com/n8n-io/n8n/pull/32783/changes#diff-ba50e99aa750dc7077f7435ef162149333cced31f7da97223d49a46fd9917766, which will be part of the new extension version to support direct connection to n8n instance, so let's remove the changes to the extension here (these will be included once the other PR is merged and this branch is updated to master)

BerniWittmann:
> @BerniWittmann ah, then it makes sense that the tests fail. I've fixed the same problem in https://github.com/n8n-io/n8n/pull/32783/changes#diff-ba50e99aa750dc7077f7435ef162149333cced31f7da97223d49a46fd9917766, which will be part of the new extension version to support direct connection to n8n instance, so let's remove the changes to the extension here (these will be included once the other PR is merged and this branch is updated to master)

@dlavrenuek ah great :+1: fix has been reverted here

BerniWittmann:
@dlavrenuek is there something still open/blocking this PR from being merged?

dlavrenuek (approved):
Works reliably now 🚀

PR #29841, opened 2026-05-06
feat(editor): Make execution run time column easier to scan

heymynameisrob:
## Summary 
Improve scan-ability of the "Run Time" column by using a single unit and consistent precision so values can be compared visually 

- Right-align numeic values and use tabular numerals so the decimal separator lines up vertically and column-wide comparisons are easier. 
- Format completed run durations as seconds with fixed 3-decimal precision using a small formatter (`durationFormatter` + `formatRuntime`) and render as e.g. `1,800.000s`. 
- Replace the previous `locale.displayTimer(..., true)` output for stopped runs with the new `formatRuntime` value computed from `startedAt/createdAt` and `stoppedAt`. 
- Add a CSS module class `.runTime` to the run time cell that makes font monospace and `font-variant-numeric: tabular-nums` for visual alignment. 
- Update the related unit test expectation in `GlobalExecutionsListItem.test.ts` to match the new formatting. 

| Before | After |
|:---:|:---:|
|  |  |

### Linear ticket
[DS-476](https://linear.app/n8n/issue/DS-476)

## Review / Merge checklist

- [X] PR title and summary are descriptive. ([conventions](<../blob/master/.github/pull_request_title_conventions.md>))
- [ ] [Docs updated](<https://github.com/n8n-io/n8n-docs>) or follow-up ticket created.
- [X] Tests included.

### DIFF
--- a/packages/frontend/editor-ui/src/features/execution/executions/components/global/GlobalExecutionsListItem.test.ts
+++ b/packages/frontend/editor-ui/src/features/execution/executions/components/global/GlobalExecutionsListItem.test.ts
@@ -213,6 +213,6 @@ describe('GlobalExecutionsListItem', () => {
 
 		const executionTimeElement = getByTestId('execution-time');
 		expect(executionTimeElement).toBeVisible();
-		expect(executionTimeElement.textContent).toBe('30m 0s');
+		expect(executionTimeElement.textContent).toBe('1,800.000s');
 	});
 });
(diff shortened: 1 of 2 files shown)


Tuukkaa:
What do you think of right-aligning the numbers? Currently it seems the decimal separators don't align if the values don't have the same amount of numbers before the period. By right aligning, it would be easier to visually see how large the numbers are

Tuukkaa:
Also some durations can be really long (several days) if a workflow has a wait node. I'm wondering how well seconds work for such large durations. Perhaps one additional format (e.g. hours + mins) could still make sense for those edge cases

heymynameisrob:
> What do you think of right-aligning the numbers?

Looks weird unless we do it to all columns (maybe except the first one)

Tuukkaa:
I think right aligning only the durations don't look too bad. It would help to visually see how big are the numbers. Here is an example from Spotify

heymynameisrob:
personally not a fan of it but not got a strong opinion on it

PR #29856, opened 2026-05-06
fix(core): Emit missing auth audit events for OIDC and SSO-restricted login

yuliia-pominchuk:
## Summary

This PR fixes two missing audit log events in SSO login flows when OIDC is enabled.

- Emit n8n.audit.user.login.success for successful OIDC callback logins.
- Emit n8n.audit.user.login.failed when email/password login is blocked by SSO restrictions.
- Add tests covering both event emissions.

Linear ticket: https://linear.app/n8n/issue/IAM-599

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/controllers/auth.controller.ts
```diff
@@ -112,13 +112,18 @@ export class AuthController {
 		}
 	}
 
-	private validateSsoRestrictions(preliminaryUser: User | undefined): void {
+	private validateSsoRestrictions(preliminaryUser: User | undefined, userEmail: string): void {
```
guillaumejacquart, line 115: I'm just wondering, do we need to pass the userEmail here ? Don't we have it already inside `preliminaryUser` ? (maybe not, I'm not sure)

guillaumejacquart (commented):
Looks good ! just a small question / comment

packages/cli/src/controllers/auth.controller.ts
yuliia-pominchuk: You're right to question this. I double-checked this path.
In the failure scenario we care about `preliminaryUser` is `undefined` and we don’t have a User object to rely on. This’s why I pass `emailOrLdapLoginId` explicitly.

guillaumejacquart (approved):

afitzek (approved):
Nice work! 🚀

PR #29860, opened 2026-05-06
fix(core): Preserve AxiosHeaders instance when applying OpenAI vendor defaults

### LINKED ISSUE #29471: HTTP Request node fails with "config.headers.setContentType is not a function" when calling api.openai.com
### Bug Description

The HTTP Request node throws NodeApiError: config.headers.setContentType is not a function when the request URL starts with https://api.openai.com/. The error originates inside the global axios request interceptor in `packages/core/src/execution-engine/node-execution-context/utils/request-helpers/axios-config.ts` at line 30:                

```
  if (config.data === undefined) {
      config.headers.setContentType(false, false);
  } 
```

The interceptor assumes config.headers is an AxiosHeaders instance, but applyVendorHeaders (lines 39–46 in the same file) reassigns config.headers to a plain object literal whenever the URL targets OpenAI:                                                                                                                                  

```
  config.headers = {
      ...Container.get(AiConfig).openAiDefaultHeaders,
      ...(config.headers ?? {}),
  };
```

Plain objects don't expose setContentType, so any downstream invocation of the interceptor on that mutated config (including the path triggered by the V3 HTTP Request node when sending no body) raises the TypeError, which is caught and rewrapped as NodeApiError at `HttpRequestV3.node.js:588`.

### To Reproduce

 1. Create a workflow with an HTTP Request node (typeVersion 4.2).
  2. Set:
    - Method: GET
    - URL: https://api.openai.com/v1/models
    - Authentication: None
    - Send Headers: true, with Authorization: Bearer <key> and OpenAI-Organization: <org> (I think this second header may be optional for reproducibility).
  3. Execute the workflow.
  4. The node fails with NodeApiError: config.headers.setContentType is not a function.

A minimal workflow JSON that reproduces the issue (single "Get models" node hitting https://api.openai.com/v1/models) is enough; no credentials are required because the failure happens before the request leaves the instance. 

### Expected behavior

The HTTP Request node should successfully reach https://api.openai.com/v1/models (or whichever OpenAI endpoint), receive the response from OpenAI, and surface it as the node output. The interceptor should not throw, regardless of whether applyVendorHeaders rewrites the headers as a plain object or an AxiosHeaders instance.

### Debug Info

# Debug info

## core

- n8nVersion: 2.19.0
- nodeJsVersion: 22.21.0
- nodeEnv: production
- database: postgres
- concurrency: 60

## storage

- success: all
- error: all
- progress: false
- manual: false
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 5000 executions

## security

- blockFileAccessToN8nFiles: false

Generated at: 2026-04-29T10:07:45.227Z

### Operating System

N/A

### n8n Version

2.19.0

### Node.js Version

22.21.0

### Database

PostgreSQL

### Execution mode

queue

### Hosting

self hosted

Joffcom:
## Summary

`applyVendorHeaders` was rebuilding `config.headers` as a plain object with a spread (`{ ...vendorDefaults, ...callerHeaders }`), which discarded the `AxiosHeaders` instance that axios expects on `InternalAxiosRequestConfig`. This caused downstream failures when the interceptor tried to call `config.headers.setContentType()` on the mutated plain object.

The fix uses `headers.set(defaults, false)` instead, which:

- preserves the `AxiosHeaders` prototype so the interceptor stays re-runnable,
- merges vendor defaults without overwriting caller-provided headers (`false` = do not overwrite).

The function is typed with `InternalAxiosRequestConfig` since `headers` is guaranteed to be an `AxiosHeaders` instance at the interceptor stage.

### How to test

1. Configure an HTTP Request node targeting `https://api.openai.com/...` with caller-provided headers (e.g. `Authorization: Bearer <key>`).
2. Execute the workflow and verify the request succeeds and caller headers are preserved.

https://linear.app/n8n/issue/NODE-4949
Fixes #29471

## Review / Merge checklist


ShireenMissi (commented):
I tested locally and it works 👏
could you please add some tests to cover the changed functionality 🙏

Joffcom:
@ShireenMissi done, Also ran the test aganst the old code to confirm it would have picked it up as what Claude originally suggested wouldn't have flagged it.

ShireenMissi (approved):
Thanks for adding the tests!
LGTM ✅

PR #29863, opened 2026-05-06
feat: Use agent-browser within Computer Use (no-changelog)

dlavrenuek:
## Summary

- Implements Browser Use functionality with [`agent-browser`](https://github.com/vercel-labs/agent-browser) implementation
- Improves stability
- in relay allow new client websocket connections overriding existing ones instead of rejecting them

The current Playwright adapter is still in place and can potentially be used when passing the right config options. The plan is to remove it in a separate after confirming that the new implementation does the job better.

- https://linear.app/n8n/issue/NODE-4956/explore-usage-of-agent-browser-as-browser-control-implementation
- https://linear.app/n8n/issue/NODE-4769/improve-browser-use-token-usage
- https://linear.app/n8n/issue/NODE-4778/use-interactive-only-snapshots-by-default-to-improve-token-usage

## Review / Merge checklist

- [x] Tests included. 


packages/@n8n/mcp-browser/src/tools/inspection.ts
```diff
@@ -41,7 +47,7 @@ function browserSnapshot(connection: BrowserConnection): ToolDefinition {
 		'Use this tool as your primary way to observe the page. Returns a ref-annotated accessibility tree — a compact text representation of all visible elements. Each interactive element gets a numeric ref for use in subsequent tool calls (browser_click, browser_type, etc.). Snapshots are small and fast. Prefer this over browser_screenshot unless you specifically need visual/layout information.',
 		browserSnapshotSchema,
 		async (state, input, pageId) => {
-			const result = await state.adapter.snapshot(pageId, input.scope);
+			const result = await state.adapter.snapshot(pageId, input.scope, input.interactive);
```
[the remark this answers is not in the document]
dlavrenuek: this is intentional. changes to playwright implementation are not in the scope of this pr

packages/@n8n/computer-use/package.json
```diff
@@ -7,7 +7,7 @@
   },
   "scripts": {
     "clean": "rimraf dist .turbo",
-    "start": "node dist/cli.js http://localhost:5678 --allowed-origins http://localhost:5678",
+    "start": "node dist/cli.js http://localhost:5678 --log-level debug --allowed-origins http://localhost:5678",
```
BerniWittmann, line 10: ni: leftover from debugging but might be good to keep for local development

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
```diff
@@ -0,0 +1,568 @@
...
+	SnapshotResult,
+	TypeOptions,
+	WaitOptions,
+} from '../types';
+
+const log = createLogger('agent-browser');
+
+const execFileAsync = promisify(execFile);
+
+const SESSION_NAME = 'n8n-computer-use';
```
BerniWittmann, line 38: Would this mean that if a user runs both cli and native app at once, aka multiple sessions that they could collide?

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
```diff
@@ -0,0 +1,568 @@
...
+			return readFileSync(path).toString('base64');
+		} finally {
+			rmSync(dir, { recursive: true, force: true });
+		}
+	}
+
+	async getText(pageId: string, target?: ElementTarget): Promise<string> {
+		await this.switchToTab(pageId);
+		const script = target
+			? `document.querySelector('[aria-ref="${this.resolveTarget(target)}"]')?.innerText??''`
```
BerniWittmann: Could this be a potential security issue? e.g. an element with a data attribute `a[data-x='y']` could cause issues with the `'`. 

I think other parts (e.g. select) wrap this in `JSON.stringify(this.resolveTarget(target))` to avoid this

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
```diff
@@ -0,0 +1,568 @@
...
+	// =========================================================================
+	// Private helpers
+	// =========================================================================
+
+	private async run(args: string[], timeoutMs = 30_000): Promise<AgentBrowserResponse> {
+		log.debug('run:', args[0]);
+		const cdpArgs = this.cdpEndpoint ? ['--cdp', this.cdpEndpoint] : [];
+		try {
+			const { stdout } = await execFileAsync(
+				'agent-browser',
```
BerniWittmann, line 79: This expects `agent-browser` to be already installed? I think we need to add it to package.json still, so that on a fresh installed this does not fail

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
```diff
@@ -0,0 +1,568 @@
...
+				['--json', '--session', SESSION_NAME, ...cdpArgs, ...args],
```
BerniWittmann, line 80: do we need to worry about accidentally executing args as flags to the agent-browser command? 

Just thinkiing about `Enter --help into a text field` would then go towards sth like `type_text --help` causing agent-browser to print its help page?

BerniWittmann (changes requested):
Great work 🔥 tested it out locally and worked way better than playwright before :+1: 

My comments are mostly on the smaller side

packages/@n8n/computer-use/package.json
dlavrenuek: I left it in intentionally because I personally always need the debug logs during local development. Otherwise you need to add `--log-level debug` to start command manually, which is a bit bothersome

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
dlavrenuek: I would treat it as possible session separation for the future. Right now the extension does not support multiple concurrent connections anyway, so only one "computer use" instance could use the browser

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
dlavrenuek: good find, will be fixed

dlavrenuek: fixed

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
dlavrenuek: don't know how I missed that, fixed

packages/@n8n/mcp-browser/src/adapters/agent-browser.ts
dlavrenuek: this is a very interesting question. a single argument containing `xxx --help` inside will not be evaluated as help command because we don't use string concatenation to build the command, but pass the args, so `Enter --help into a text field` is valid text. But a single `--help` string as an argument of `open`, `tab new` and other commands will indeed be evaluated as the help command. I did add an assertion to block and potential argument injection. It does allow `-` as a single character because it is not treated as any special argument by `agent-browser` and is a valid argument for key `press` command.

For `type` it's a bit more complicated: it must be allowed to type `--help` or any other string starting with `-`, so we can not block any input, but agent-browser does not provide any encoding/escaping functionality there. Separating the string in multiple parts in a single command will lead to concatenation with an extra ` ` (space) in between, so `agent-browser type e3 "-" "-help"` command will type `- -help` (this is also clear when looking at agent-browser source code for this command). But if you execute the type command multiple times, it will add the text to the end without extra spaces, so `agent-browser type e3 "-"` and then `agent-browser type e3 "-help"` will do the job. In my solution I iterate over the first character of the remaining input while no `-` is found to also exclude shorthand arguments like `-h`

dlavrenuek: fixed

BerniWittmann (approved):
great work 🔥

PR #29871, opened 2026-05-06
fix(ai-builder): Validate MCP tool names and schemas (no-changelog)

aalises:
## Summary

This PR makes Instance AI's MCP tool ingestion deterministic and bounded. MCP connections can contribute tools only when their names and schemas fit the supported contract. Configs/tools outside that contract are skipped with source-aware warning logs, so valid tools can still load.

### What changed

| Area | Previous behavior | New behavior | Example fixed | Why this matters |
| --- | --- | --- | --- | --- |
| MCP server names | External/browser MCP configs were passed to the MCP client as configured. | Server names must be safe identifiers before connection setup; unsafe configs are skipped and logged. | A configured server named `bad server` or `constructor` is skipped before client setup. | Treats MCP connection metadata as untrusted input and avoids unsafe keys in internal maps. |
| MCP tool names | External, browser, and local gateway tools were merged with only limited exact-name filtering. | Tool names must be safe identifiers, must not collide after normalization, and must not overlap with domain/orchestration tools. Unsafe or conflicting tools are skipped and logged. | `search_files` and `search-files` are treated as the same claim; `custom_workflows` is rejected because it claims a reserved suffix. | Prevents MCP-provided names from ambiguously overlapping with trusted Instance AI tools. |
| Toolset composition | Local gateway tools were added inside domain tool construction, making ownership/filtering harder to reason about. | MCP tools are collected centrally in `createInstanceAgent`, then added separately to orchestration context and orchestrator toolsets. | Local gateway tools go through the same validation path before being exposed to the orchestrator or sub-agents. | Keeps the trust boundary in one place, so MCP tools cannot bypass shared validation through a different construction path. |
| MCP schemas | Schema sanitization handled provider-incompatible nulls but had no budget around oversized or unsupported shapes. | Input/output schemas are bounded by depth, node, property, and union limits. Unsupported Zod shapes cause only the offending tool to be skipped. | A tool with 251 input properties or a schema deeper than 32 levels is skipped while other tools continue loading. | Bounds work done on MCP-provided schemas and avoids accepting shapes the provider/tooling cannot safely represent. |
| Local gateway schemas | Raw gateway JSON schemas were converted before size checks. | Gateway JSON schemas are checked before tool construction, then sanitized after conversion. | An oversized local gateway schema is rejected before JSON Schema to Zod conversion. | Applies resource limits before expensive conversion work and keeps local gateway behavior aligned with external MCP behavior. |
| Gateway resource decisions | Resource-decision options were treated as arbitrary strings through the suspend payload, frontend, and confirmation DTO. | The daemon wire payload is parsed as strings, then narrowed to `denyOnce`, `allowOnce`, and `allowForSession`; unsupported options are filtered/rejected. | A daemon option like `alwaysAllow` is not rendered in the confirmation UI and is rejected by the confirmation DTO. | Keeps persistent permission semantics owned by settings/product flows instead of accepting arbitrary daemon-provided actions. |

### Supported MCP identifier contract

- Starts with a letter
- Contains only letters, numbers, `_`, or `-`
- Has a maximum length of 64 characters
- Does not use Unicode compatibility-normalized variants
- Does not use unsafe object keys after lowercasing (`__proto__`, `constructor`, `prototype`)
- Does not collide after normalization, e.g. `search-files`, `search_files`, and `SearchFiles` are treated as the same claim
- Does not claim a reserved domain/orchestration tool name, including reserved suffixes such as `custom_workflows`

### Schema bounds

| Bound | Limit |
| --- | ---: |
| Maximum depth | 32 |
| Maximum total nodes | 1,000 |
| Maximum properties on one object | 250 |
| Maximum union options | 100 |

Unsupported schemas are skipped at the tool level and logged with source, tool name, path, and limit details.

### Review notes

- Name validation is shared across external MCP, browser MCP, and local gateway MCP.
- Valid MCP tools continue to be loaded; invalid tools are skipped instead of failing the whole agent setup.
- Resource confirmation remains limited to once/per-session choices. Persistent permission modes stay settings-owned.

https://linear.app/n8n/issue/INS-193

## Review / Merge checklist


packages/@n8n/instance-ai/src/agent/mcp-tool-name-validation.ts
```diff
@@ -0,0 +1,106 @@
...
+			name,
+			source,
+		);
+	}
+	return normalizeMcpToolName(name);
+}
+
+function findReservedSuffix(name: string, reservedToolNames: Map<string, string>): string | null {
+	for (let index = 0; index < name.length; index++) {
+		const char = name[index];
```
elsmr: Is this check not a bit too strict? if I understand correctly it will block things like `github_workflows` or `..._plan` because those are built in tools

packages/@n8n/instance-ai/src/agent/instance-agent.ts
```diff
@@ -71,54 +79,84 @@ export async function createInstanceAgent(options: CreateInstanceAgentOptions):
...
-	for (const [name, tool] of Object.entries({ ...mcpTools, ...browserMcpTools })) {
-		if (!domainToolNames.has(name)) {
-			allMcpTools[name] = tool;
-		}
-	}
-	if (orchestrationContext && Object.keys(allMcpTools).length > 0) {
-		orchestrationContext.mcpTools = allMcpTools;
-	}
+	const warnSkippedMcpTool = (error: McpToolNameValidationError) => {
+		context.logger?.warn('Skipped MCP tool with unsafe name', {
```
elsmr, line 89: Not for this PR, but probably worth having some feedback for the user on which tools are skipped

packages/@n8n/instance-ai/src/agent/instance-agent.ts
```diff
@@ -71,54 +79,84 @@ export async function createInstanceAgent(options: CreateInstanceAgentOptions):
...
-		? Object.fromEntries(
-				Object.entries(createToolsFromLocalMcpServer(context.localMcpServer)).filter(
-					([name]) => !browserToolNames.has(name),
-				),
-			)
+	const rawLocalMcpTools = context.localMcpServer
+		? createToolsFromLocalMcpServer(context.localMcpServer, context.logger)
 		: {};
+	addSafeMcpTools(allMcpTools, rawLocalMcpTools, {
+		source: 'local gateway MCP',
```
elsmr: Should local gateway be registered before external MCP servers? So that the local gateway tools are not blocked

aalises:
Thanks for the review @elsmr! I removed the suffix-based reserved-name check which was too strict tbf, keeping the check for exact normalized collisions. I also changed registration order so local gateway MCP tools are registered before external/browser MCP tools, meaning gateway tools win collisions for both orchestrator and delegated MCP tool contexts 👍

packages/@n8n/instance-ai/src/agent/mcp-tool-name-validation.ts
aalises: true!

elsmr (approved):

PR #29872, opened 2026-05-06
fix(ai-builder): Use expiring Computer Use setup tokens (no-changelog)

aalises:
## Summary

This PR tightens the Computer Use setup-command flow around the short-lived pairing token.

Security rationale:
- The setup command contains temporary pairing material that is copied into a local terminal and can outlive the immediate setup moment in local user environments.
- That credential should stay short-lived, single-use, and scoped only to initial pairing.
- The setup-command endpoint should not return an existing long-lived `sess_*` session key, because session keys are the durable credential for an already paired Computer Use daemon.

What changes:
- `POST /instance-ai/gateway/create-link` now returns setup-token expiry metadata: `expiresAt` and `ttlSeconds`.
- The Computer Use setup UI shows when the setup token expires and refreshes the command before copying if the previous token has expired.
- Setup command generation now stays on the `gw_*` pairing-token path and does not return an existing active `sess_*` session key.

What was already true / preserved:
- Setup tokens are short-lived and single-use. Expired setup tokens cannot be consumed for pairing.
- Already paired Computer Use sessions continue using their session key and have no time-based expiry. They remain valid until explicit disconnect/revocation, while the server-side session state exists.
- This PR does not add session rotation, daily re-pairing, or timed session expiry.

User impact:
- Users can see that a setup command is time-limited.
- If they copy after expiry, the UI gets a fresh setup command first.
- A running paired Computer Use daemon/service is not invalidated just because the original setup token expired.

https://linear.app/n8n/issue/INS-193

## Review / Merge checklist


BerniWittmann (commented):
PR itself looks good. 

When trying to test this i somehow did not see the token Expiry text, or am i overlooking something that i need to do to show it? 

Image

aalises:
> When trying to test this i somehow did not see the token Expiry text, or am i overlooking something that i need to do to show it?

hmmm it should appear. Have you rebuilt the needed packages and restarted the dev server?  i.e. check if `POST /rest/instance-ai/gateway/create-link` response. If it has expiresAt and ttlSeconds, the backend is good and the issue is frontend-side (needs rebuild)

BerniWittmann (approved):
Nevermind, was on the wrong branch 🙈 
Showing up now, looking good

PR #29877, opened 2026-05-06
refactor(editor): Split Instance AI view into route-driven empty + thread leaves (no-changelog)

r00gm:
## Summary

Make the URL the single source of truth for the active thread id by splitting the monolithic `InstanceAiView.vue` (~960 lines) into a layout shell + two route-children leaf views.

### Layout shell — `InstanceAiView.vue` (~150 lines)

- Sidebar (collapse state via `useSessionStorage`, resize, transition, in-sidebar collapse via `InstanceAiThreadList`'s `@collapse`)
- `<RouterView />` for the leaves
- Page-level lifecycle: `loadThreads`, `fetchCredits`, credits push listener, gateway setup
- `onUnmounted`: `closeSSE`, `stopCreditsPushListener`, `stopGatewayPushListener` (only fires when leaving the namespace, not on empty↔thread)
- `onBeforeRouteLeave`: reset sidebar to collapsed when leaving the chat namespace
- `provide(SidebarStateKey, { collapsed, toggle })` so leaves can render the collapsed-mode toggle in their own header

### Empty leaf — `InstanceAiEmptyView.vue` (`/instance-ai`)

- Header (sidebar toggle when collapsed + credits + settings)
- Empty state UI (suggestions, centered input, status bar, credit warning banner)
- On mount: `store.clearCurrentThread()` — fresh placeholder UUID, SSE closed, state reset
- Submit: awaits `store.syncThread(threadId)` first; on success fires `sendMessage` + `router.replace` in parallel, on failure stays put with a toast. The in-flight stream survives the navigation because the store's single runtime stays mounted across leaf swaps.

### Thread leaf — `InstanceAiThreadView.vue` (`/instance-ai/:threadId`)

- Full chat UI: header (sidebar toggle, thread title, reconnect indicator, credits/settings/debug/artifacts), message list with scroll management, floating input, preview panel
- Watches `props.threadId` with `immediate: true`:
  - already on this thread → reconnect SSE if disconnected
  - known thread → `switchThread`
  - unknown thread → wait for `loadThreads` (idempotent) then validate; redirect to empty if still unknown
- `onUnmounted`: cleanup observers, execution tracking (SSE close stays in the layout)

### Shared chrome

Both leaves consume two reusable pieces so the duplicated header / banner code from the initial split is gone:

- `components/InstanceAiViewHeader.vue` — sidebar toggle (when collapsed) + credits dropdown + settings button + read-only callout, with `title` and `actions` slots for leaf-specific bits (Thread fills both; Empty fills neither). Owns the global `.sidebar-toggle-fade-*` keyframes — single source of truth.
- `composables/useCreditWarningBanner.ts` — owns the `dismissed` ref + watch that resets dismissal on transition into low-credits + `visible` computed. Takes `MaybeRefOrGetter<boolean>` so the composable doesn't depend on the store; leaves call it as `useCreditWarningBanner(isLowCredits)` after `storeToRefs`.

### Module descriptor

Reorganised into nested routes:
- Parent `/instance-ai` → `InstanceAiView` (layout)
- Child `''` → `InstanceAiEmptyView` (named `INSTANCE_AI_VIEW`)
- Child `':threadId'` → `InstanceAiThreadView` (named `INSTANCE_AI_THREAD_VIEW`)

### Why

- Eliminates the `currentThreadId = uuidv4()` placeholder hack — the URL is the only source of truth for which thread is active.
- Sets up clean per-route lifecycle (each leaf owns its own concerns).
- Prerequisite for future per-thread runtimes (`Map<threadId, ThreadRuntime>` + provide/inject), which is where consumer churn happens — that's a follow-up PR, not this one. Components stay on `store.X` here.

### Behaviour tweaks (deliberate)

The PR is **mostly** a structural refactor, but a few intentional behaviour tweaks landed alongside it:

1. **Canvas-preview auto-open simplified.** Dropped `metadata.activePreviewTab` restore, `wasCanvasOpenBeforeSwitch`, and `userSentMessage` flags. New gate: auto-open whenever a build artifact arrives **unless we're hydrating past data**. The auto-open watchers run with `flush: 'sync'` so they observe `isHydratingThread` before `loadHistoricalMessages`' `finally` clears it — re-entering an old thread no longer pops the preview for past artifacts (matches Claude-style "show on creation, don't restore across sessions"). Discussed with @tuukka in Slack.
2. **Thread persistence moved out of `sendMessage`.** The runtime no longer owns persistence — `sendMessage` only ensures SSE, pushes the optimistic user message, fires telemetry, and dispatches. EmptyView's `handleSubmit` awaits `store.syncThread` first; on failure it stays put with a toast, on success it fires `sendMessage` and `router.replace` in parallel. We never land on `/instance-ai/:threadId` for a thread the BE doesn't know about.
3. **Preview width cap.** The canvas preview was capped at 50% of the *window*, which left the chat side with less than 50% of the visible area when the layout sidebar was open. It now caps at 50% of the actual `.threadArea` flex container.

## Review / Merge checklist

- [x] PR title and summary are descriptive. `(no-changelog)` because the `instance-ai` module is experimental.
- [ ] Docs updated or follow-up ticket created. _n/a_
- [x] Tests included. _Old `InstanceAiView.test.ts` split into `InstanceAiEmptyView.test.ts` (suggestions, clearCurrentThread on mount) and `InstanceAiThreadView.test.ts` (no suggestions, reconnect on same-thread re-entry, switch on navigating to known thread)._


packages/frontend/editor-ui/src/features/ai/instanceAi/InstanceAiEmptyView.vue
```diff
@@ -0,0 +1,119 @@
...
+	const threadId = store.currentThreadId;
+	// Kick off the send and the route change in parallel.
+	// `sendMessage` runs its synchronous prelude before its first await:
+	// pushes the optimistic user message into the runtime and opens the
+	// SSE connection. By the time the router resolves the navigation,
+	// ThreadView mounts with that state already in place. Backend
+	// persistence (`syncThread` → `postMessage`) completes in the
+	// background while the user is already on the thread page.
+	void store.sendMessage(message, attachments, rootStore.pushRef);
+	void router.replace({
```
aalises, line 53: We should keep the user on `/instance-ai` until thread creation succeeds? If `ensureThread()` fails, this leaves them on a non-persisted thread URL

packages/frontend/editor-ui/src/features/ai/instanceAi/InstanceAiThreadView.vue
```diff
@@ -0,0 +1,672 @@
...
+// --- Side panels ---
+const showArtifactsPanel = ref(true);
+const showDebugPanel = ref(false);
+const isDebugEnabled = computed(() => localStorage.getItem('instanceAi.debugMode') === 'true');
+
+// --- Preview panel resize (when canvas is visible) ---
+const { width: windowWidth } = useWindowSize();
+const previewPanelWidth = ref(Math.round(windowWidth.value / 2));
+const isResizingPreview = ref(false);
+const previewMaxWidth = computed(() => Math.round(windowWidth.value / 2));
```
aalises: Maybe this should subtract the expanded sidebar width. Otherwise the preview can squeeze the chat area when the sidebar is open?

r00gm: done

packages/frontend/editor-ui/src/features/ai/instanceAi/InstanceAiEmptyView.vue
r00gm: now the empty view, first ensureThread() then in parallel sends the first message and navigate to the detail

packages/frontend/editor-ui/src/features/ai/instanceAi/InstanceAiThreadView.vue
```diff
@@ -0,0 +1,685 @@
...
+
+// --- Execution tracking via push events ---
+const executionTracking = useExecutionPushEvents();
+
+// --- Header title ---
+// Returns the resolved title once we have one, or undefined while we're still
+// figuring out which thread to show. Rendering only on a defined value avoids
+// the "New conversation" → real title flash when resuming a recent thread.
+const currentThreadTitle = computed<string | undefined>(() => {
+	const thread = store.threads.find((t) => t.id === store.currentThreadId);
```
Cadiac, line 78: Did we consider switching the `store.threads` to a Map / object, it seems like we're doing some lookups from this array at different places? Or would that make reactivity difficult?

(but given this is an existing store that's probably out of scope on this PR, just came to my mind reading this code)

packages/frontend/editor-ui/src/features/ai/instanceAi/InstanceAiThreadView.vue
```diff
@@ -0,0 +1,685 @@
...
+	if (thread && thread.title && thread.title !== NEW_CONVERSATION_TITLE) {
+		return thread.title;
+	}
+	const firstUserMsg = store.messages.find((m) => m.role === 'user');
+	if (firstUserMsg?.content) {
+		const text = firstUserMsg.content.trim();
+		return text.length > 60 ? text.slice(0, 60) + '…' : text;
```
Cadiac, line 85: I just happened to see that packages/frontend/editor-ui/src/features/agents/utils/thread-title.ts shipped something very much like this recently that we could consider using here, at our stuff too, but not important right now

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiViewHeader.vue
```diff
@@ -0,0 +1,122 @@
...
+				size="small"
+				icon-size="large"
+				data-test-id="instance-ai-settings-button"
+				@click="goToSettings"
+			/>
+			<slot name="actions" />
+		</div>
+	</div>
+
+	<!-- eslint-disable-next-line vue/no-multiple-template-root -->
```
Cadiac, line 69: is supressing this warning a clean solution here, is this the structure we want?

Cadiac (approved):
Very nice, seemed to work perfectly locally when I tried breaking it.
Left a couple of minor comments, but given how this starts conflicting with other stuff soon LGTM to get this merged as is!

packages/frontend/editor-ui/src/features/ai/instanceAi/InstanceAiThreadView.vue
r00gm: that's coming, yes

packages/frontend/editor-ui/src/features/ai/instanceAi/components/InstanceAiViewHeader.vue
r00gm: we must adjust the linter, that rule is from back on the day!!! when vue did not support this

Cadiac: TIL, I thought we must have a good reason for that rule but I never figured out why we have it :D

PR #29883, opened 2026-05-06
fix(core): Sanitize untrusted input passed to template renderers

shortstacked:
## Summary

- Add shared `escapeHtml` utility and escape user-controlled fields (`message`, `reason`, `error.message`) before passing them to Handlebars `res.render()` calls
- Add `html: false` (explicit) and `sanitize-html` post-processing to the `useMarkdown` composable as defense-in-depth
- Refactor existing OIDC views to reuse the shared `escapeHtml` utility

https://linear.app/n8n/issue/SEC-460

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/cli/src/modules/sso-oidc/views/oidc-test-result.ts
+++ b/packages/cli/src/modules/sso-oidc/views/oidc-test-result.ts
@@ -1,3 +1,5 @@
+import { escapeHtml as escapeHtmlBase } from '@/utils/escape-html';
+
 const PAGE_STYLES = `
 body { background: rgb(251,252,254); font-family: 'Open Sans', sans-serif; padding: 10px; margin: auto; width: 500px; top: 40%; position: relative; }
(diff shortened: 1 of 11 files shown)


packages/cli/src/modules/sso-oidc/views/oidc-test-result.ts
```diff
@@ -8,11 +10,7 @@ li { list-style: none; margin: 0; color: rgb(125, 125, 125); font-size: 12px; }
 `;
 
 function escapeHtml(value: unknown): string {
-	return String(value ?? '(n/a)')
-		.replace(/&/g, '&amp;')
-		.replace(/</g, '&lt;')
-		.replace(/>/g, '&gt;')
-		.replace(/"/g, '&quot;');
+	return escapeHtmlBase(value) || '(n/a)';
```
dlavrenuek, line 13: The new implementation returns `(n/a)` for an empty string after escaping, the old implementation would only do it if `value` was `null`

PR #29884, opened 2026-05-06
feat(core): Generate service-specific OAuth2 credentials for dedicated MCP tools

RomanDavydchuk:
## Summary

Generate service-specific OAuth2 credentials for each dedicated MCP tool that's generated based on the MCP server confgs. The credentials are meant to be used only with these dedicated tools - they can't be used in HTTP request node or generic MCP client. The user doesn't need to provide any extra data, therefore the credentials can be created straight from the NDV using the Quick connect button. Currently only OAuth2 credentials are supported

This PR also adds a mock Linear MCP server to make sure that the feature works properly with multiple different services

image
image
image

### To test
1. Launch n8n with `N8N_ENABLED_MODULES=mcp-registry`
2. Create a workflow
3. Add an AI agent node with a dedicated MCP tool (Notion or Linear)
4. Create a credential from the NDV
5. Verify that the popup window opens, credential is created successfully
6. Verify that the tool include/exclude selector loads the available tools
7. Message an agent to make it invoke the tool, verify that it works successfully
8. Add a different MCP tool and verify the same steps
9. Add a generic MCP tool
10. Set it up to call Notion or Linear
11. Verify that it also works as expected

https://linear.app/n8n/issue/NODE-4962/implement-credentials-for-dedicated-mcp-server-tools

## Review / Merge checklist

- [x] Tests included. 


packages/cli/src/modules/mcp-registry/mcp-registry-node-loader.ts
```diff
@@ -40,6 +42,8 @@ export class McpRegistryNodeLoader implements NodeLoader {
 
 	private nodeTypes: INodeTypeData = {};
 
+	private credentialTypes: ICredentialTypeData = {};
```
RomanDavydchuk, line 45: Turns out we do need this field 😄

packages/frontend/editor-ui/src/features/credentials/composables/useCredentialOAuth.ts
```diff
@@ -84,17 +84,21 @@ export function useCredentialOAuth() {
...
 		);
 
+		if (visibleProperties.length === 0) {
+			return true;
+		}
+
+		if (overwrittenProperties.length === 0) {
+			return false;
+		}
+
```
RomanDavydchuk, line 101: Before, this check would return `false` because for these generated creds we don't have any overwritten props, so I made it first check if there are any visible props and if there aren't any -> return `true` early

If we are going to decide to remove `or setup manually` part from these creds (since there isn't anything to setup) I think it's worth extracting it into a separate var like `hasNoEditableProperties` or something

elsmr: Yep probably a good idea, the name of this fn `hasManagedOAuth` is also a bit outdated, maybe `canQuickConnectOAuth` or similar would be better

packages/@n8n/nodes-langchain/nodes/mcp/shared/types.ts
```diff
@@ -4,9 +4,17 @@ export type McpTool = { name: string; description?: string; inputSchema: JSONSch
 
 export type McpServerTransport = 'sse' | 'httpStreamable';
 
+export type McpOAuth2CredentialType = 'mcpOAuth2Api' | `${string}McpOAuth2Api`;
```
elsmr, line 7: ✨

packages/@n8n/nodes-langchain/nodes/mcp/McpRegistryClientTool/McpRegistryClientTool.node.ts
```diff
@@ -180,3 +186,14 @@ function resolveConfig(
...
+
+function getCredentialType(
+	ctx: Pick<ILoadOptionsFunctions | ISupplyDataFunctions | IExecuteFunctions, 'getNode'>,
+): McpAuthenticationOption {
+	const credentials = ctx.getNode().credentials ?? {};
+	const credentialType = Object.keys(credentials).find(
+		// for now we support only OAuth2
+		(credentialType) => isMcpOAuth2Authentication(credentialType),
+	);
+	return credentialType ?? 'mcpOAuth2Api';
```
elsmr: Should we throw here instead of falling back?

packages/cli/src/modules/mcp-registry/mcp-registry-node-loader.ts
```diff
@@ -59,20 +63,35 @@ export class McpRegistryNodeLoader implements NodeLoader {
 		const { description: baseDescription } = NodeHelpers.getVersionedNodeType(baseNode);
 
 		for (const server of this.registry.getAll({ includeDeprecated: true })) {
-			const description = serverToNodeDescription(server, baseDescription);
-			if (!description) continue;
+			const nodeDescription = serverToNodeDescription(server, baseDescription);
+			const credentialDescription = serverToCredentialDescription(server);
+			if (!nodeDescription) continue;
```
elsmr: Should we also skip if there's no credential, I don't think the node will actually work if the credential is not there? then the `if` on line 82 can also be dropped

packages/@n8n/nodes-langchain/nodes/mcp/McpRegistryClientTool/McpRegistryClientTool.node.ts
RomanDavydchuk: Good point, I think that yeah it'll be better, especially if we support more auth types in the future. Will change

packages/cli/src/modules/mcp-registry/mcp-registry-node-loader.ts
RomanDavydchuk: True, the reason why I decided to allow creating a node with no creds is for the cases when there's no need for auth (analogous to `authentication: 'none'` in generic MCP client tool, but I'm not sure if we'll have such servers to be honest

packages/cli/src/modules/mcp-registry/node-description-transform.ts
```diff
@@ -1,11 +1,80 @@
...
+				displayName: 'Server URL',
+				name: 'serverUrl',
+				type: 'hidden',
+				default: remote.endpointUrl,
+			},
+			{
+				displayName: 'Allowed HTTP Request Domains',
+				name: 'allowedHttpRequestDomains',
+				type: 'hidden',
+				default: 'none',
```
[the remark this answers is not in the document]
RomanDavydchuk: This will be handled separately

packages/cli/src/modules/mcp-registry/mcp-registry-node-loader.ts
elsmr: Aha that makes sense, but oauth2 nodes should still not be included without their credentials

elsmr (approved):

PR #29885, opened 2026-05-06
chore(agents): Stop using `execution_entity` table and avoid setting the `workflowId` to nullable

mike12345567:
## Summary

Removes agent runs from `execution_entity` and gives them their own dedicated schema.

### Why

`AgentExecutionService.recordMessage` was inserting into `execution_entity` with `workflowId = NULL`, requiring the `NOT NULL` constraint to be dropped. ~15 call sites assume `execution.workflowId` is non-null, so a leaked agent row could silently break ownership, redaction, the public API, MCP tools, and the job processor.

Per-message data was also stored as ~13 free-form rows in `execution_metadata` (stringified JSON under arbitrary keys), forcing index-unfriendly `WHERE key='X'` predicates throughout.

### What changed

**Schema** (single migration, replaces unreleased `1783000000001-CreateExecutionThreads`):
- `execution_threads` renamed to `agent_execution_threads`
- New `agent_execution` table with typed columns (`userMessage`, `assistantResponse`, `model`, `promptTokens`, `cost`, `toolCalls`, `timeline`, `hitlStatus`, `source`, etc.)
- Restores `execution_entity.workflowId NOT NULL`, drops the `threadId` column and index
- Deletes any pre-existing `mode='agent'` rows so the constraint can be re-applied

**Backend:**
- New `AgentExecution` + `AgentExecutionThread` entities and repositories
- `AgentExecutionService.recordMessage` is now a single `INSERT` (was 1 + 1 + ~13)
- `backfillSuspendedEx

## Review / Merge checklist

- [ ] Tests included. 


bjorger (approved):
LGTM 🚀

PR #29893, opened 2026-05-06
fix(core): Add support for context establishment hooks in webhook mode

afitzek:
## Summary

In webhook mode the node definitions are not patched at bootup with the discovered extractors. So when a webhook trigger is handled in the webhook mode, the generated workflow definition has its set up extractor hooks removed.

## Review / Merge checklist

- [x] Tests included. 


guillaumejacquart (commented):
Looks good! nice catch :) 
Should we add a test ?

phyllis-noester (approved):

PR #29913, opened 2026-05-06
chore(core): Add @n8n/engine HTTP server and harness (no-changelog)

mfsiega:
## Summary

Adds some basic server infrastructure for engine 2.0. That includes an Express server, plus a Dockerfile. The goal is just to get some of the noise out of the way so subsequent PRs can focus more on the code. Expect to revise details here to align with what we actually need, didn't want to spend too much time tweaking it in a vaccuum.

We intentionally don't re-use the server abstractions from `cli` yet. We also don't run any e2e tests during CI. In general we'll probably want to set it up to only run tests if the package is being touched, since it shouldn't have cross-package dependencies yet.

https://linear.app/n8n/issue/CAT-3020/scaffold-n8nengine-server-package-in-process-e2e-test-harness

## Review / Merge checklist

- [ ] Tests included.


packages/@n8n/engine/src/serve.ts
```diff
@@ -0,0 +1,24 @@
...
+const port = Number.parseInt(process.env.PORT ?? '3000', 10);
+const host = process.env.HOST ?? '0.0.0.0';
+
+const { app } = createEngineServer();
+
+const server = app.listen(port, host, () => {
+	console.log(`engine: listening on http://${host}:${port}`);
+});
+
+const shutdown = (signal: string): void => {
```
tomi, line 15: Should guard for the the callback being called multiple times

packages/@n8n/engine/src/testing/__tests__/start-engine-server.integration.test.ts
tomi, line 1: Can we call these integration tests instead of e2e? We have used e2e solely for playwright tests, where as we have used integration tests for cli specific tests that have the actual server running

packages/@n8n/engine/src/serve.ts
```diff
@@ -0,0 +1,24 @@
+import { createEngineServer } from './server';
+
+const port = Number.parseInt(process.env.PORT ?? '3000', 10);
+const host = process.env.HOST ?? '0.0.0.0';
```
tomi: We have a config mechanism in `@n8n/config`. Should use that for consistency

packages/@n8n/engine/src/serve.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const { app } = createEngineServer();
+
+const server = app.listen(port, host, () => {
+	console.log(`engine: listening on http://${host}:${port}`);
```
tomi: As a followup unless there's already a ticket for that: use our normal logging framework instead of console logs

docker/images/engine/Dockerfile
tomi, line 1: All the rest of our docker files are in `/docker` folder in the repo root. While I would also like to keep this close to the source, I think we should be consistent here

tomi (commented):
Couple comments mainly about conventions

packages/@n8n/engine/src/serve.ts
```diff
@@ -0,0 +1,31 @@
+import 'reflect-metadata';
```
mfsiega: Why do we have this? Do we have this elsewhere in the codebase?

packages/@n8n/config/src/configs/engine.config.ts
mfsiega, line 1: Is this a standard pattern, where a package has its own config? I thought it was mostly all defined in one place under `@n8n/config/src/configs`?

packages/@n8n/engine/README.md
mfsiega, line 1: We don't need this change, the original comment was fine.

packages/@n8n/config/src/configs/engine.config.ts
```diff
@@ -0,0 +1,19 @@
+import { Config, Env } from '../decorators';
+
+@Config
+export class EngineConfig {
+	/** Port the engine HTTP server listens on. */
+	@Env('PORT')
```
Flexicon: Should this be scoped via some sort of prefixing? Something like `N8N_ENGINE_`?

mfsiega: Good call!

docker/images/engine/Dockerfile
mfsiega: Done.

packages/@n8n/engine/src/testing/__tests__/start-engine-server.integration.test.ts
mfsiega: Done.

packages/@n8n/engine/src/serve.ts
```diff
@@ -0,0 +1,29 @@
+import { EngineConfig } from '@n8n/config';
```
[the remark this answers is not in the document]
mfsiega: This package is a common utility, so it makes sense to use it here.

packages/@n8n/engine/src/serve.ts
mfsiega: Done.

packages/@n8n/engine/src/serve.ts
mfsiega: Opened https://linear.app/n8n/issue/CAT-3041.

packages/@n8n/config/src/configs/engine.config.ts
```diff
@@ -0,0 +1,19 @@
...
+	/** Host interface the engine HTTP server binds to. */
+	@Env('N8N_ENGINE_HOST')
+	host: string = '0.0.0.0';
+
+	/**
+	 * Postgres connection URL for the data plane. When empty, the engine boots
+	 * in healthcheck-only mode — workflow execution endpoints are not mounted.
+	 */
+	@Env('N8N_ENGINE_DATABASE_URL')
+	databaseUrl: string = '';
```
tomi: Is this needed in this PR? Probably worth leaving it out if not

packages/@n8n/engine/src/serve.ts
tomi, line 1: Is this package meant to be used as a library or as a standalone app or both? Now it's kinda both, where index exports the server creation, but serve is an entrypoint that spins up the server. Is this intentional?

tomi (dismissed):
Couple questions but nothing blocking 👍

packages/@n8n/config/src/configs/engine.config.ts
mfsiega: Done.

packages/@n8n/engine/src/serve.ts
mfsiega: At the moment it's intentionally both. Eventually I want a package just for the engine as a library, but I don't have a clear place where the server stuff belongs yet, so keeping it here seemed the lowest-friction option until then.

tomi (approved):
Image

Flexicon (approved):
🚀

PR #28017, opened 2026-04-03
feat(core): Agents as first class entities support (no-changelog)

yehorkardash:
## Summary

First version of agents as first class citizens, this is the feature branch which was used to develop the overall feature,  have a look at the `feat(agents)` list of closed PRs to see the work that has gone into this branch: https://github.com/n8n-io/n8n/pulls?q=is%3Apr+is%3Aclosed+feat%28agents%29

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/AGENT-3/feature-implement-agent-json-conversion
https://linear.app/n8n/issue/AGENT-6/feature-allow-running-workflows-as-tools-in-agent

## Review / Merge checklist

- [x] PR title and summary are descriptive. ([conventions](../blob/master/.github/pull_request_title_conventions.md)) 
- [ ] [Docs updated](https://github.com/n8n-io/n8n-docs) or follow-up ticket created.
- [x] Tests included. 
- [ ] PR Labeled with `Backport to Beta`, `Backport to Stable`, or `Backport to v1` (if the PR is an urgent fix that needs to be backported)
- [x] I have seen this code, I have run this code, and I take responsibility for this code.

PR #29915, opened 2026-05-06
fix(ai-workflow-builder): Use OS path separator in MCP node-type path check

### LINKED ISSUE #29902: [MCP] get_node_types always fails with "Invalid path - path traversal detected" while other MCP tools work
### Bug Description

* MCP tool `get_node_types` fails consistently with:
`Error: Invalid path - path traversal detected`

* This happens for valid payloads (both string node IDs and object node IDs with discriminators), and blocks the MCP workflow-building loop that requires fetching type definitions before generating workflow code.

* At the same time, other MCP tools work in the same environment/session (`search_nodes`, `search_workflows`, `validate_workflow`, `create_workflow_from_code`, `execute_workflow`), so this appears isolated to `get_node_types`.

* Calls are made with argument shapes documented in n8n MCP tools reference for get_node_types.

* The impact seem to be high for MCP-based workflow authoring as agents cannot reliably fetch node schemas and must fall back to guessing + validation loops.

* This has been reproduced on both n8n 2.19.2 (beta) and 2.18.7 (stable)

### To Reproduce

1. Connect to n8n via MCP.
2. Call `get_node_types` with a simple node ID:
   ```json
   {"nodeIds":["n8n-nodes-base.manualTrigger"]}
3. Call again with discriminator object:
   ```json
   {"nodeIds":[{"nodeId":"n8n-nodes-base.set","mode":"manual"}]}
4. Call with explicit version:
   ```json
   {"nodeIds":[{"nodeId":"n8n-nodes-base.httpRequest","version":"4.3"}]}

### Expected behavior


`get_node_types` should return a definitions string with TypeScript parameter definitions for the requested nodes.

### Debug Info

# Debug info

## core

- n8nVersion: 2.19.2
- platform: npm
- nodeJsVersion: 22.22.2
- nodeEnv: undefined
- database: postgres
- executionMode: regular
- concurrency: -1
- consumerId: 80397389-d6ab-4e4c-8b90-a79e37657694

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: filesystem

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (windows nt 10.0; win64; x64; rv:150.0) gecko/20100101 firefox/150.0
- isTouchDevice: false

Generated at: 2026-05-06T13:39:31.476Z

### Operating System

Windows Server 2022

### n8n Version

2.19.2 && 2.18.7

### Node.js Version

22.22.2

### Database

PostgreSQL

### Execution mode

main (default)

### Hosting

self hosted

devareddy05:
## Summary

The MCP `get_node_types` tool fails on Windows with `Invalid path - path traversal detected` for every legitimate node ID, blocking the MCP workflow-building loop. Other MCP tools (`search_nodes`, `validate_workflow`, etc.) are unaffected.

### Root cause

The path-containment check in [code-builder-get.tool.ts:49](https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/ai-workflow-builder.ee/src/code-builder/tools/code-builder-get.tool.ts#L49) hardcodes `'/'` as the path separator:

```ts
return resolvedPath.startsWith(resolvedBase + '/') || resolvedPath === resolvedBase;
```

On Windows, `path.resolve()` returns paths with backslash separators (e.g. `C:\Users\foo\nodes\sub`). Comparing them against `resolvedBase + '/'` (e.g. `C:\Users\foo\nodes/`) always fails — every legitimate path gets flagged as path traversal.

n8n already has a canonical implementation of this check in [`@n8n/backend-common`](https://github.com/n8n-io/n8n/blob/master/packages/%40n8n/backend-common/src/utils/path-util.ts#L8-L17) that correctly uses `path.sep`. This PR aligns the tool's local copy with that pattern.

### Fix

Replace the hardcoded `'/'` with `path.sep`. One-character behavioral change on Windows (`\\` instead of `/`), no observable change on POSIX (`sep` is `'/'` there).

### Test

`pnpm --filter @n8n/ai-workflow-builder test:unit code-builder-get` — 36/36 passing (35 existing + 1 new).

The new regression test simulates Windows `resolve` + `sep` behavior using `path.win32` since `path.resolve` on POSIX won't produce Windows-style paths. It asserts:
- `winChild.startsWith(winBase + path.win32.sep)` is `true` (the fix works)
- `winChild.startsWith(winBase + '/')` is `false` (documents the bug)

Closes #29902
[GHC-8182](https://linear.app/n8n/issue/GHC-8182)

## Review / Merge checklist

- [x] Tests included.

### DIFF
--- a/packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/test/code-builder-get.tool.test.ts
+++ b/packages/@n8n/ai-workflow-builder.ee/src/code-builder/tools/test/code-builder-get.tool.test.ts
@@ -520,6 +520,18 @@ describe('CodeBuilderGetTool', () => {
 				const trickyPath = path.join(baseDir, 'package', '..', '..', '..', 'etc', 'passwd');
 				expect(validatePathWithinBase(path.resolve(trickyPath), baseDir)).toBe(false);
 			});
+
+			it('should use the OS-specific path separator (regression for #29902)', () => {
+				// Simulate Windows resolve+sep using path.win32 since path.resolve on
+				// POSIX won't produce Windows paths.
+				const winBase = path.win32.resolve('C:\\users\\foo\\nodes');
+				const winChild = path.win32.resolve(
+					'C:\\users\\foo\\nodes\\n8n-nodes-base\\httpRequest\\v1.ts',
+				);
+
+				expect(winChild.startsWith(winBase + path.win32.sep)).toBe(true);
+				expect(winChild.startsWith(winBase + '/')).toBe(false);
+			});
 		});
 
 		describe('tool invocation with malicious inputs', () => {


Joffcom:
Hi @devareddy05 — thanks for the diligent rebasing on this one! Heads up: looking at the current state, the diff only contains the test file (`code-builder-get.tool.test.ts`), and the production file `code-builder-get.tool.ts` no longer has the path-containment check your fix targets — it was refactored away in #28017 / #30323 between May 6 and May 20.

Could you check whether:
1. The bug still reproduces on current `master` (on Windows, with the latest code-builder), and
2. If yes, where the equivalent check now lives — your fix may need to migrate to a different file

If the bug is gone we can just close this. If it still reproduces, please re-target the fix and let us know. Thanks!

PR #29916, opened 2026-05-06
fix(ai-builder): Preserve collected planning context

aalises:
## Summary

Preserve planning context that was already collected before the planner sub-agent runs, i.e. what answers were already answered and what resources were already discovered so it doesn't ask for a credential name that it already knows of...

Example:
```
## Already-collected answers

- What should this workflow do?: Create a todo list for me based on discussions in Slack
- How often should it run?: Every morning
- Which Slack messages should it read?: All channels I'm a member of
- Where should the output go?: Email it to me
- Which model should it use?: Anthropic
- Credential selected for slackApi: Slack account (slackApi)
- Credential selected for anthropicApi: Anthropic account (anthropicApi)

## Already-discovered resources

- Credentials available: Slack account (slackApi), Anthropic account (anthropicApi)
- Data tables available: Personal todos, Team tasks
```

**Disclaimer:**  It's a bit disorganised, basically assembling the context by hand. https://github.com/n8n-io/n8n/pull/28666 in this (a bit orphaned) PR we improve these briefings by making them fully typed and in a central place... we should move that over the line and use the typed contract for this PR as well...

**What does it do checklist:**
- Adds prior `ask-user` answers plus credential and data table discovery results to the planner briefing as dedicated sections.
- Reads prior tool observations from the current message group when available, so planner follow-up runs can still see context collected earlier in that assistant turn.
- Records selected credential names/types from prior credential setup results when a matching credential list result is available.
- Shares tool-name constants with the tools that emit the observed events, reducing drift if tool IDs change.
- Tells the planner to trust already-collected briefing context and avoid re-asking for those choices.
- Avoids appending the same current user request twice when memory already returned it.

https://linear.app/n8n/issue/INS-203

## Review / Merge checklist


packages/@n8n/instance-ai/src/tools/orchestration/plan-with-agent.tool.ts
```diff
@@ -120,17 +184,343 @@ async function getRecentMessages(
...
+ * context only carries the current run ID and message group ID. This bridges
+ * those two concepts while keeping the current run as a defensive fallback.
+ */
+function getMessageGroupRunIds(context: OrchestrationContext): string[] {
+	const messageGroupId = context.messageGroupId;
+	if (!messageGroupId) return [];
+
+	const runIds = new Set<string>();
+	try {
+		for (const { event } of context.eventBus.getEventsAfter(context.threadId, 0)) {
```
luka-zivkovic, line 292: not blocking, just a possible small nitpick
this loops over every event in the thread just to find a couple of run-starts in the current group. probably fine but maybe worth considering 
breaking out once we hit our own run's run-start

luka-zivkovic (approved):
Left a small nitpick, but not blocking

OlegIvaniv (approved):

packages/@n8n/instance-ai/src/tools/orchestration/plan-with-agent.tool.ts
```diff
@@ -93,6 +125,38 @@ function extractTextParts(parts: unknown[]): string {
 		.join('\n');
 }
 
+function isRecord(value: unknown): value is Record<string, unknown> {
```
OlegIvaniv, line 128: Image

Image

aalises: it totally **is** a record.... of the function that appears most throughout the codebase @OlegIvaniv

PR #29927, opened 2026-05-06
fix(Basic LLM Chain Node): Parse plain object output from structured output parser

### LINKED ISSUE #29903: Basic LLM Chain + Structured Output Parser fails with "Failed to parse agent steps" error
## Bug Description

When using Basic LLM Chain node with Structured Output Parser node, if the node version is >= 1.9, the execution fails with error:

```
Failed to parse agent steps
```

The model output is correctly formatted (e.g., `{"output":{"isValid":false,"message":"..."}}`), but cannot be parsed.

## Steps to Reproduce

1. Create a workflow with Basic LLM Chain node (version 1.9+)
2. Connect Structured Output Parser node to the Basic LLM Chain
3. Configure the Structured Output Parser with a JSON schema
4. Execute the workflow

Expected: Output should be parsed successfully
Actual: Error "Failed to parse agent steps" is thrown

## Root Cause

In `chainExecutor.ts` (line ~185), when `version >= 1.9`, the code uses `getAgentStepsParser(outputParser)` instead of directly using `outputParser`:

```typescript
if (version >= 1.9) {
    chain = promptWithInstructions
        .pipe(model)
        .pipe(getAgentStepsParser(outputParser))
        ...
} else {
    chain = promptWithInstructions
        .pipe(model)
        .pipe(outputParser)
        ...
}
```

The `getAgentStepsParser` function is designed for agentic workflows (handling `AgentFinish`, `AgentAction[]`, `BaseMessage` types). However, Basic LLM Chain outputs plain objects that don't match any of the expected types, causing the function to throw `Error('Failed to parse agent steps')`.

## Expected Behavior

Basic LLM Chain with Structured Output Parser should work correctly regardless of node version.

## Environment

- n8n version: 2.3.6
- Node version: 1.9
- Model: any model that supports structured output

## Additional Context

The issue was introduced when `getAgentStepsParser` was added in commit 3b694a6c79 to support ResponsesApi and OpenAI tools. The function works correctly for AI Agent nodes but causes issues with Basic LLM Chain when using output parsers.

devareddy05:
## Summary

The Basic LLM Chain node (v1.9+) fails with `Failed to parse agent steps` when wired up to a Structured Output Parser. The chain pipes the model output through `getAgentStepsParser()`, which only handles `string`, `BaseMessage`, `AgentAction[]`, and `AgentFinish` shapes. Models that produce structured JSON directly as a plain object (e.g. through tool-bound structured output) fall through to the final `throw`.

This PR adds a fallback branch in `getAgentStepsParser()` that stringifies a plain non-array object and feeds it to the supplied `outputParser`, letting the parser validate against its schema. Existing branches are unchanged, so the regression case for `AgentAction[]` without `format_final_json_response` still throws.

### Root cause

In [`promptUtils.ts`](packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/methods/promptUtils.ts) the type guards run in this order: `string` → `Array` → `BaseMessage` (`instanceof`) → `AgentFinish` (has `returnValues`). A plain object like `{"output":{"isValid":false,"message":"..."}}` matches none of these and reaches `throw new Error('Failed to parse agent steps')`. The wrapper was added in #22936 to support OpenAI Responses API + tools; this PR extends it to also handle plain structured output without removing that support.

### Changes

- `packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/methods/promptUtils.ts` — add a final fallback branch that runs `outputParser.parse(JSON.stringify(steps))` for plain non-array objects.
- `packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/test/promptUtils.test.ts` — two regression tests covering the bug scenario (`{output: {...}}`) and an arbitrary plain object shape.

### How to test

1. Create a workflow with a Basic LLM Chain node (version 1.9 or later).
2. Connect a Structured Output Parser configured with a JSON schema.
3. Execute the workflow.
4. Expected: the parsed object is returned. Previously: `Failed to parse agent steps`.

`pnpm --filter @n8n/n8n-nodes-langchain test ChainLLM` — 85/85 passing (83 existing + 2 new).

Closes #29903
[GHC-8186](https://linear.app/n8n/issue/GHC-8186)

## Review / Merge checklist

- [x] Tests included.
- [ ] PR Labeled with \`Backport to Beta\`, \`Backport to Stable\`, or \`Backport to v1\` (if the PR is an urgent fix that needs to be backported).

### DIFF
--- a/packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/methods/promptUtils.ts
+++ b/packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/methods/promptUtils.ts
@@ -191,5 +191,13 @@ export const getAgentStepsParser =
 			return parsedOutput;
 		}
 
+		if (typeof steps === 'object' && steps !== null && !Array.isArray(steps)) {
+			const parsedOutput = (await outputParser.parse(JSON.stringify(steps))) as Record<
+				string,
+				unknown
+			>;
+			return parsedOutput;
+		}
+
 		throw new Error('Failed to parse agent steps');
 	};
--- a/packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/test/promptUtils.test.ts
+++ b/packages/@n8n/nodes-langchain/nodes/chains/ChainLLM/test/promptUtils.test.ts
@@ -349,5 +349,31 @@ describe('promptUtils', () => {
 
 			await expect(parser(steps)).rejects.toThrow('Failed to parse agent steps');
 		});
+
+		it('should parse plain object via outputParser (regression for #29903)', async () => {
+			const parser = getAgentStepsParser(mockOutputParser);
+			const plainObject = { output: { isValid: false, message: 'error' } };
+
+			const parsedResult = { isValid: false, message: 'error' };
+			mockOutputParser.parse.mockResolvedValue(parsedResult);
+
+			const result = await parser(plainObject as unknown as AgentFinish);
+
+			expect(mockOutputParser.parse).toHaveBeenCalledWith(JSON.stringify(plainObject));
+			expect(result).toEqual(parsedResult);
+		});
+
+		it('should parse plain object with arbitrary shape via outputParser', async () => {
+			const parser = getAgentStepsParser(mockOutputParser);
+			const plainObject = { foo: 'bar', count: 42 };
+
+			const parsedResult = { foo: 'bar', count: 42 };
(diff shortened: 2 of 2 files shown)


yehorkardash:
Hey @devareddy05, thanks for your contribution!
We've investigated and found that the issue was in the `isMessage` guard rather than an incoming message format.
Resolved here https://github.com/n8n-io/n8n/pull/33338

PR #29930, opened 2026-05-06
feat(core): Add get_environment tool for runtime date and timezone (no-changelog)

mike12345567:
**## Summary

Adds a `get_environment` tool that is auto-injected into every agent, returning:

- `now` — current time as an ISO string in the instance timezone (with offset)
- `timezone` — the configured IANA timezone (`GlobalConfig.generic.timezone`)
- `dayOfWeek` — long-form weekday name (e.g. `Tuesday`)

This surfaces information the model cannot know on its own (current date, day of week, instance timezone) so it can reason correctly about "today", deadlines, and scheduling.

Implemented as a tool call rather than baked into the system prompt to avoid busting system-prompt caching on per-request values. Timezone is read at handler time so config changes take effect without rebuilding agents. Injection is wrapped in a try/catch so a failed module load logs a warning and continues rather than breaking agent setup.**
## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/cli/src/modules/agents/agents.service.ts
+++ b/packages/cli/src/modules/agents/agents.service.ts
@@ -636,6 +636,20 @@ export class AgentsService {
 		const { agent, agentId, projectId, credentialProvider, nodeToolsEnabled, integrationType } =
 			params;
 
+		// Inject get_environment unconditionally. It surfaces info the model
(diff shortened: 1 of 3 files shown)


bjorger (approved):
LGTM 🚀

PR #29931, opened 2026-05-06
chore(core): Enable Daytona sandbox in Instance AI evals (no-changelog)

JoseBra:
## Summary

- Wire eval CI containers with `N8N_INSTANCE_AI_SANDBOX_ENABLED=true`, `N8N_INSTANCE_AI_SANDBOX_PROVIDER=daytona`, `DAYTONA_API_URL`, and `DAYTONA_API_KEY` so the workflow builder runs through Daytona instead of the deprecated tool-mode path.
- Add a per-lane assertion step that fails the workflow loudly if any of the 9 lanes isn't actually configured for daytona — guards against env-var / persisted-DB-setting drift that would silently mix sandbox and tool-mode builds.
- Add an always-on post-eval log capture step that dumps non-migration container logs from every lane, so Daytona/builder errors are visible inline in step output.
- Log the effective sandbox config at module init, with a `(DB override)` marker when persisted admin settings disagree with env vars.
- Add `EvalMockedCredentialsHelper` — wraps `additionalData.credentialsHelper` per execution. When `getDecrypted` can't resolve a credential ID, returns a marker payload; `authenticate` / `preAuthentication` no-op for marker payloads. The unauthed request flows into the existing `evalLlmMockHandler` and gets a synthetic response. Provider-agnostic, same per-execution isolation as the existing mock-HTTP layer.
- Tighten the eval verifier prompt: don't cite credential ID values as a root cause of failure under any circumstance — they're auto-substituted by the framework.

## Local validation

Single-scenario smoke against the dev server with sandbox=daytona:

```
5 scenarios   ·  3 PASS / 2 FAIL  ·  60% pass rate  ·  ~3:35 wall-clock
build time    ·  207s (Daytona create + tsc + tsx)
0 build_failure / 0 framework_issue
```

The two fails are the canonical pre-existing `builder_issue`s (Telegram crash with no error handling, Gmail invalid-email validation) — same as previous baselines.

## Test plan

- [ ] CI eval workflow passes with sandbox enabled on every lane
- [ ] Verify step confirms `sandboxEnabled=true sandboxProvider=daytona` for all 9 lanes
- [ ] Pass rate roughly matches the prior baseline (or improves, with the cred-helper)

### DIFF
--- a/.github/workflows/test-evals-instance-ai.yml
+++ b/.github/workflows/test-evals-instance-ai.yml
@@ -206,24 +206,32 @@ jobs:
           N8N_ENCRYPTION_KEY: ${{ secrets.N8N_ENCRYPTION_KEY }}
         run: |
           # Layer 2 — defense in depth: explicitly mask each secret's value.
+          # ::add-mask:: is a single-line workflow command. Multi-line secrets
+          # (e.g. N8N_LICENSE_CERT is PEM-encoded) must be masked one line at
+          # a time, otherwise only the first line is registered.
           for v in "$EVALS_ANTHROPIC_KEY" "$DAYTONA_API_KEY" \
                    "$N8N_LICENSE_ACTIVATION_KEY" "$N8N_LICENSE_CERT" \
                    "$N8N_ENCRYPTION_KEY"; do
-            [ -n "$v" ] && echo "::add-mask::$v"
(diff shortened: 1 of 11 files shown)


mutdmour (approved):
LGTM

PR #29932, opened 2026-05-06
feat(core): Show workflow-triggered runs in agent session history (no-changelog)

mike12345567:
### Summary

Makes `MessageAnAgent` node runs first-class agent sessions: they now appear in the agent's session list alongside chat/Slack/schedule sources, can continue an existing thread across executions, and can be opened directly from the execution log.

### Three changes, one feature

1. **Record workflow-triggered runs as sessions** — `AgentsService.executeForWorkflow` now streams through `ExecutionRecorder` and persists a `MessageRecord` with a new `'workflow'` source type. Previously these runs were invisible in the session list. Recording is fire-and-forget — failures are logged but don't fail the workflow node.

2. **Allow session continuation from the node** — adds an `Advanced > Session ID` option to `MessageAnAgent`. When set, it overrides the default per-call thread id (`${executionId}-${itemIndex}`) so subsequent executions hit the same agent thread and reuse its memory.

3. **Deep link from execution log to session detail** — the node emits a `session: { agentId, projectId, sessionId }` block in its output JSON. A new `useMessageAgentSessionLink` composable resolves this to a route in the session detail view; `LogDetailsPanel` shows a "View session" button when present. Opens in a new tab so the popped-out logs panel stays functional.

### Screenshot

Button to view session in logs:
image

image

## Review / Merge checklist

- [x] Tests included. 

### DIFF
--- a/packages/@n8n/api-types/src/agents.ts
+++ b/packages/@n8n/api-types/src/agents.ts
@@ -48,6 +48,13 @@ export const INCOMPATIBLE_WORKFLOW_TOOL_BODY_NODE_TYPES = [
 
 export const AGENT_SCHEDULE_TRIGGER_TYPE = 'schedule';
 
+/**
+ * Source string recorded on agent executions invoked from a workflow via the
+ * MessageAnAgent node. Mirrors the pattern set by chat/slack/schedule sources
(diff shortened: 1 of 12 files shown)


bjorger (approved):
LGTM 🚀

PR #29933, opened 2026-05-06
fix(Google Sheets Trigger Node): Custom OAuth option not availble when there are managed credentials

RomanDavydchuk:
## Summary

When there is managed OAuth for Google Sheets Trigger credentials, the UI won't display a dropdown to set up credentials manually, like it does for other nodes.

Google Sheets Trigger (bug):
image

Google Sheets Node (proper):
image

This was due to the fact that the FE was checking for `authentication` option with value `oAuth2`, while Google Sheets Trigger has it as `triggerOAuth2`. This PR fixes it by making the check for OAuth2 option more generic.

After the fix:
image

> Note: the reason why the trigger node has it as `triggerOAuth2` is because the regular node and the trigger reuse the same helpers and some of them differentiate between credential types depending on if it's `triggerOAuth2` or `oAuth2`.

### To test
1. Set up credential overwrites: launch n8n with (the client ID and secret don't have to be valid):
  ```
  CREDENTIALS_OVERWRITE_DATA='{"googleOAuth2Api": {"clientId":"client-id","clientSecret":"client-secret"}}'
  ```
2. Add Google Sheets Trigger node
3. Add a new credential (if you have none - click `setup manually`, if you do have some credentials already - create a new one like usual)
4. Verify that you can pick the `Custom OAuth` option (the dropdown wasn't there before the fix)

https://linear.app/n8n/issue/N8N-9975/google-sheets-trigger-node-doesnt-support-generic-credentials

## Review / Merge checklist

- [x] Tests included. 


michael-radency (approved):
LGTM

PR #29935, opened 2026-05-06
fix(editor): Reconcile Data Table column schema on node open

RicardoE105:
## Summary

Fixes the broken UI on Data Table insert/update/upsert nodes whose cached `columns.schema` has drifted from the live table. When an MCP/AI-generated (or hand-authored) workflow ships a schema that was never loaded from the real table, opening the node shows a stale-data warning plus outdated/missing inputs, and the user has to click the small refresh icon for the mappings to render correctly.

This change adds an opt-in `refreshIncompleteSchemaOnOpen` flag to `ResourceMapperTypeOptions` and turns it on for the Data Table node. When set, the resource mapper reconciles the cached schema against the live source on mount, **but only when the cached schema is detected to be structurally incomplete**. A complete-but-drifted schema keeps the existing stale-data warning so the user stays in control of when to refresh.

## How "incomplete" is detected

Real resource mapper loaders always populate the `readOnly` and `removed` properties on every field. An authored schema (e.g. one the AI builder infers from a preceding Set node) omits them. `isResourceMapperSchemaIncomplete` uses that signal: if any field is missing `readOnly`/`removed`, the schema was authored rather than loaded, so it gets reconciled. Otherwise the schema is treated as a real-but-possibly-stale one and the warning is shown.

## Why opt-in per node rather than generic

The detection can't be applied globally. Most loaders (Data Table, Google Sheets, Airtable) populate `readOnly`/`removed`, but the Postgres loader does not, so a legitimately loaded Postgres schema would look "incomplete" and we'd clobber its intentional stale-warning UX. The flag keeps the behavior scoped to nodes whose loader is known to populate those fields. Only the Data Table node opts in for now; the detection helper is reusable if we want it elsewhere later.

## Behavior matrix

| Scenario | Cached schema | Result |
| --- | --- | --- |
| MCP/AI-generated workflow | Incomplete (authored) | Auto-reconciled on open |
| User adds/removes/renames a column after configuring the node | Complete, drifted | Stale warning, user refreshes |
| Imported/duplicated workflow originally built in the UI | Complete | Stale warning, user refreshes |

## Files

- `packages/workflow/src/interfaces.ts` — add `refreshIncompleteSchemaOnOpen?: boolean` to `ResourceMapperTypeOptionsBase`
- `packages/nodes-base/nodes/DataTable/common/addRow.ts` — set the flag on the shared resource mapper config (covers insert / update / upsert)
- `packages/frontend/editor-ui/src/app/utils/nodeTypesUtils.ts` — add `isResourceMapperSchemaIncomplete`
- `packages/frontend/editor-ui/src/features/ndv/parameters/components/ResourceMapper/ResourceMapper.vue` — gate the on-open reconcile on the detector

## Follow-up

Separately worth checking with @mutdmour whether the AI builder should emit `columns.schema` at all, since schemas are editor cache rather than essential node config. That is the upstream half of this issue; this PR is the frontend safety net.

https://linear.app/n8n/issue/ADO-5168

## Review / Merge checklist

- [x] Tests included.

## Test plan

**Unit:**
- `pnpm --filter n8n-editor-ui test src/features/ndv/parameters/components/ResourceMapper` and `src/app/utils/nodeTypeUtils.test.ts` — covers both reconcile-when-incomplete and keep-warning-when-complete, plus the detector helper directly.

**Manual repro (without MCP):**
- Create a Data Table with columns A, B, C. Configure a Data Table Insert node pointing at it. Save.
- Add a new column D directly on the Data Table (via the Data Table UI).
- Reopen the workflow and open the Insert node. Because the cached schema is complete (it was loaded), you get the stale-data warning and refresh on demand. The warning UX is preserved for user-driven drift.

**Manual repro (MCP, original ticket scenario):**
- Use the MCP `create_workflow_from_code` tool to build a workflow that inserts into a Data Table. The cached `columns.schema` is inferred from the preceding Set node and is structurally incomplete. On master: stale warning plus outdated inputs. On this branch: the schema is auto-reconciled on open, no manual refresh needed.

**Regression check:**
- Open a non-Data-Table resource mapper node (e.g. Google Sheets append, or Postgres insert) and confirm the deliberate stale-warning UX is unchanged.


RicardoE105:
@MiloradFilipovic this works but what do you think about the approach? 🤔

MiloradFilipovic:
> @MiloradFilipovic this works but what do you think about the approach? 🤔

I think this only fixes the symptom but we should take a look at the deeper issue here. I feel like the stale warning has it's place in the UX and should be present if schema drift is a result of user action (scenario 1 and 2 in the PR description) - so we shouldn't blindly force refresh. On the other hand, having stale or broken schema for newly generated workflow clearly feels like a bug to me.
IMO, schemas are the editor cache (not essential node config) so we shouldn't instruct our AI builders to generate schemas (like we do [here](https://github.com/n8n-io/n8n/blob/90b215939f364a0b5141e4911efdd6851ca0f697/packages/nodes-base/nodes/DataTable/DataTable.node.ts#L52)) but could be good to check this with @mutdmour since I see schema hints in a few places so there may be a reason he was insisting on these.
On the other side, import/duplicate workflows is also a valid bug scenario that we should tackle. Not sure what's the best here, I guess we can programmatically detect incomplete/invalid schemas and only then refresh (this could be a good safety net for generated workflows as well).

RicardoE105:
@MiloradFilipovic Yeah, totally agree the force-refresh was too blunt. Reworked it so it only auto-refreshes when the cached schema is actually incomplete, i.e. it was authored by the builder rather than loaded from the source (the loader always fills in readOnly/removed, an MCP-generated schema doesn't). So the broken/generated case gets fixed, but if you just added/renamed/removed a column yourself, the schema is still "complete", just drifted, so you keep the normal stale warning and decide when to refresh. No more blindly clobbering user changes.

One thing I found while doing this: I couldn't make the detection fully generic. Postgres's loader doesn't set readOnly/removed at all, so a legit Postgres schema would look "incomplete" and we'd nuke its stale-warning UX. So it stays opt-in per node (refreshIncompleteSchemaOnOpen) and only Data Table turns it on for now, but the detection helper is reusable if we want it as a safety net elsewhere later.

Small scope note: an imported/duplicated workflow that was originally built in the UI (so its schema is complete) will now show the stale warning instead of silently auto-fixing. That felt right given your point about user-driven drift, but lmk if you'd rather still auto-fix imports.

And +1 on checking with @mutdmour about whether we should even be telling the builder to emit columns.schema, that's the upstream half of this. This PR is just the FE safety net. Happy to open a separate ticket for the builder side if it makes sense, didn't want to touch the schema hints blind.

MiloradFilipovic (dismissed):
Works well, thanks for addressing the feedback 🙌
One, probably out-of-scope thing that I noticed [here](https://github.com/n8n-io/n8n/blob/23d9f0bd24745df75b69d39ac23574f25e670a1c/packages/frontend/editor-ui/src/features/ndv/parameters/components/ResourceMapper/ResourceMapper.vue#L402) in `ResourceMapper.vue`, we do:

```ts
field.removed = existingField.removed;
```

I wonder if cached AI-created schemas have the same field ids as live schema but is missing `removed` or `readOnly`, something like:

```
// Cached schema from AI/generated workflow
[
	{
		id: 'name',
		displayName: 'Name',
		type: 'string',
		// missing readOnly
		// missing removed
	}
]
Live loader returns:
[
	{
		id: 'name',
		displayName: 'Name',
		type: 'string',
		readOnly: false,
		removed: false,
	}
]
```

Will this overwrite the exiting `removed: false` to `undefined`?

Should we maybe be a bit stricter about it:

```ts
if (typeof existingField.removed === 'boolean') {
	field.removed = existingField.removed;
} else {
	field.removed = field.removed ?? false;
}
```

MiloradFilipovic (approved):
Thanks ♥️

PR #29953, opened 2026-05-07
ci: Exclude all monorepo packages from safechain minimum age

Matsuuu:
## Summary

Our monorepo has packages that are outside of the @n8n/* matching scope so we need to take those into account too in the safechain exlusion.

Later on we could move to a configuration file but this is just to unblock releases on beta track.

## Review / Merge checklist

- [ ] Tests included. 


tomi (approved):
🚀

PR #29956, opened 2026-05-07
fix(MongoDB Node): Resolve collection parameter per item in write operations

### LINKED ISSUE #19177: Mongodb node operations do not use current item index everywhere
### Bug Description

When using mongodb node operations : `findOneAndReplace`, `findOneAndUpdate`, `insert`, and `update`; with multiple input items, only the first one from the list is used.

This is caused by using 0 as the item index  inside `this.getNodeParameter` (e.g. `this.getNodeParameter('collection', 0) as string` in [MongoDb.node.ts](https://github.com/n8n-io/n8n/blob/346d3b4ba92b15c735508f68a852cd43210a06a7/packages/nodes-base/nodes/MongoDb/MongoDb.node.ts#L351-L353)). 

In the context of an insert operation, this would result in inserting only the first input into the same collection.

### To Reproduce

1. Create a new workflow
2. Add a "Set" node. Configure it to output a simple array (e.g. `["collection-1, "collection-2", "collection-3"]`)
3. Add a "Split" node right after to split the array from step 2 into multiple items
4. Add a mongodb node (mongo version doesn't matter)
5. Configure the mongodb node to insert some arbitrary data (e.g. `{"test": 1}`) into the collections from the split node

### Expected behavior

mongodb should contain 3 collections: `collection-1`, `collection-2`, `collection-3` but only has `collection-1`

### Debug Info

# Debug info

## core

- n8nVersion: 1.107.4
- nodeJsVersion: 22.17.0
- database: postgres
- executionMode: regular
- concurrency: -1
- license: community
- consumerId: unknown

## storage

- success: all
- error: all
- progress: false
- manual: true
- binaryMode: memory

## pruning

- enabled: true
- maxAge: 336 hours
- maxCount: 10000 executions

## client

- userAgent: mozilla/5.0 (macintosh; intel mac os x 10.15; rv:142.0) gecko/20100101 firefox/142.0
- isTouchDevice: false

Generated at: 2025-09-04T08:26:47.526Z

### Operating System

Docker

### n8n Version

1.107.4

### Node.js Version

22.17.0

### Database

SQLite (default)

### Execution mode

main (default)

### Hosting

self hosted

michael-radency:
## Summary

Introduces node version 1.3 of the MongoDB node to fix write operations
(`insert`, `update`, `findOneAndUpdate`, `findOneAndReplace`) incorrectly
reading the `collection` parameter only from the first input item, ignoring
expressions that resolve differently per item.

**What changed in v1.3:**
- All four write operations now evaluate `collection` (and all other
  parameters) against each input item individually
- `insert` groups items by resolved collection name and uses `insertMany`
  per group — preserving atomic batch semantics and minimising round-trips
- Each output item carries an exact `pairedItem` reference to its source
  input item (previously a fallback array covering all inputs was used)
- `continueOnFail` error handling is per-item for all operations

**How to test:**
1. Create a MongoDB node with `insert` or `update`
2. Set the `Collection` field to an expression like `{{ $json.collection }}`
3. Pass multiple items with different `collection` values
4. Confirm each item is written to its respective collection

Previous versions (≤ 1.2) are unaffected — all changes are gated behind
`nodeVersion >= 1.3`.

closes https://github.com/n8n-io/n8n/issues/19177
https://linear.app/n8n/issue/NODE-3583/community-issue-mongodb-node-operations-do-not-use-current-item-index

## Review / Merge checklist

- [ ] Tests included. 


packages/nodes-base/nodes/MongoDb/MongoDb.node.ts
```diff
@@ -343,86 +401,268 @@ export class MongoDb implements INodeType {
...
+								});
+							} else {
+								throw error;
+							}
 						}
+					}
 
-						await mdb
+					// Phase 2: insertMany per collection group
+					for (const [collection, groupItems] of groups) {
```
DawidMyslak, line 512: One thing worth flagging: because we iterate `groups` in `Map` insertion order, the output order no longer matches input order, e.g. input `[A, B, A, B]` yields output `[A0, A2, B1, B3]`.

`pairedItem` is correct, but downstream nodes that pair by position (Merge by position, `$('MongoDB').itemMatching(i)` etc.) would see different items than on v1.2.

WDYT about a `returnData.sort((a, b) => a.pairedItem.item - b.pairedItem.item)` after this loop, so we keep the batching win without the surprise?

packages/nodes-base/nodes/MongoDb/MongoDb.node.ts
```diff
@@ -239,100 +239,158 @@ export class MongoDb implements INodeType {
...
+						const [item] = prepareItems({
+							items: [items[i]],
+							fields,
+							updateKey,
+							useDotNotation,
+							dateFields,
+						});
 
-				const updateKey = ((this.getNodeParameter('updateKey', 0) as string) || '').trim();
+						if (!item) continue;
```
DawidMyslak: When an input item lacks the `updateKey` field, `prepareItems` filters it out and `!item` swallows it, so the item disappears from the output with no `pairedItem` and no error, even with `continueOnFail` on.

Could we push a `continueOnFail` error item here ("Item is missing the updateKey field") or throw a `NodeOperationError` otherwise? Feels friendlier than a silent drop.

Note: we use `if (!item) continue;` in 3 places here 👍, so probably worth covering all of them.

packages/nodes-base/nodes/MongoDb/MongoDb.node.ts
```diff
@@ -343,86 +401,268 @@ export class MongoDb implements INodeType {
...
+								.collection(collection)
+								.insertMany(groupItems.map((g) => g.item));
+
+							for (const [idx, insertedId] of Object.entries(insertedIds)) {
+								const g = groupItems[parseInt(idx, 10)];
+								returnData.push({
+									json: { ...g.item, id: insertedId as unknown as string },
+									pairedItem: { item: g.originalIndex },
+								});
+							}
```
DawidMyslak: Readability nit: since `groupItems` ordering matches the array passed to `insertMany`, you could index directly and skip the `parseInt`:

```ts
for (let i = 0; i < groupItems.length; i++) {
  const g = groupItems[i];
  returnData.push({
    json: { ...g.item, id: insertedIds[i] as unknown as string },
    pairedItem: { item: g.originalIndex },
  });
}
```

DawidMyslak (changes requested):

michael-radency:
Thanks for review @DawidMyslak :pray: , updated as suggested

DawidMyslak (approved):

